PR #1562, opened 2021-12-03
ci: migrated pull requests CI to GitHub Actions

btaillon:
https://coveord.atlassian.net/browse/KIT-1268

https://github.com/coveo/ui-kit/actions/runs/1536954241 👀 

Other things I could do in a separate PR:
* Replace artifacts + tar with github cache
  * Optionally, all builds could be cached separately and builds could be skipped when their cache is hit. Across PRs, this could save some time. Calculating the hashes may be a pain though, since they need to account for *everything* that can affect the build.
* Introduce a JUnit action to display unit tests differently

I will make a separate PR to build & publish on `master`.

PR #1567, opened 2021-12-06
 ci: migrated pull requests CI to GitHub Actions 

btaillon:
https://coveord.atlassian.net/browse/KIT-1268

Follow-up to https://github.com/coveo/ui-kit/pull/1562.

If you wish to compare with the previous PR, here's the reviewable diff: https://github.com/coveo/ui-kit/pull/1567/files/a432ecfef6d52147e7fefe096297d9e6eaa09608..68cdc8d790105ee14afa8be60ff18aa52f488d56


packages/atomic/cypress/integration/result-list/result-components/result-fields-list.cypress.ts
```diff
@@ -1,34 +1,56 @@
...
   fieldsListComponent,
   FieldsListSelectors,
 } from './result-fields-list.selector';
 import {resultTextComponent} from './result-text-selectors';
 
-const addInResultListFixture = (numChild: number) => {
+const rem = 16;
+const separatorWidth = rem * 2;
+
+const addInResultListFixture = (
```
btaillon, line 19: There was some flakiness in the fields list component with these tests, some ResizeObserver loop errors would appear in the console. I adjusted these tests so that there's exactly enough space for all fields to show (minus one pixel when we want to test whether fields are hidden).

Jenkinsfile
```diff
@@ -4,6 +4,10 @@ node('linux && docker') {
   def isBump = !!tag
   def isMaster = env.BRANCH_NAME == 'master'
 
+  if (!isMaster) {
+    return
+  }
+
```
louis-bompart, line 10: why modify Jenkins, is it not dead yet?

louis-bompart, line 10: why modify Jenkins, is it not dead yet?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,35 @@
+name: "Setup & build"
+description: "Setup node, install dependencies & build"
+runs:
+  using: composite
+  steps:
+    - uses: actions/setup-node@v2
+      with:
+        node-version-file: '.nvmrc'
+    - uses: actions/cache@v2
```
btaillon, line 9: This action replaces the whole artifact download/upload and tar compress/extract logic I had in the last PR.

The cache downloads the build if it's available. If it's not available, it builds. The cache action also includes a post-job action which uploads the build to the cache if it wasn't downloaded.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,35 @@
...
+          packages/atomic/loader
+          packages/atomic/storybook-static
+          packages/atomic/www
+        key: ${{ github.sha }}
+    - uses: actions/cache@v2
+      with:
+        path: |
+          node_modules
+          */*/node_modules
+        key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
```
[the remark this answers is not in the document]
louis-bompart: I think `runner.os` is pretty useless here :P

.github/workflows/prbot.yml
```diff
@@ -1,18 +1,90 @@
...
+  unit-test:
+    name: "Run unit tests"
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup
+      - run: npx lerna run copy:staticresources --scope @coveo/quantic
```
louis-bompart, line 56: Just to be on the safe side, if you use lerna through npx, suffix it with a `@4.0.0`, just in case lerna got hijacked.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,35 @@
...
+          packages/bueno/dist
+          packages/headless/dist
+          packages/auth/dist
+          packages/atomic/dist
+          packages/atomic/docs
+          packages/atomic/headless
+          packages/atomic/loader
+          packages/atomic/storybook-static
+          packages/atomic/www
+        key: ${{ github.sha }}
```
btaillon, line 22: I chose the commit SHA as the key for the cache, meaning that we build at-least once per commit.

An alternative would be to try to calculate a hash of the source files and any other file that could potentially affect the build. In this case, we could also divide the cache per-package and build each package individually if their cache isn't hit.

This would likely save a substantial amount of time, but I'm worried that adding new packages would become very difficult and I'm worried that we may forget to include some files in the hash that affect the build (e.g., rollout or esbuild configs).

.github/workflows/prbot.yml
```diff
@@ -1,18 +1,90 @@
 name: PR Robot
-on:
-- pull_request
-
+on: [pull_request]
 jobs:
-  pr-robot:
+  report-size:
```
btaillon, line 4: The report-size job takes a long time, since it builds both the current commit and `master` from scratch. I could try to optimize it to use both the cache from `master` and the cache from the current commit, but since this job runs in parallel with the entire rest of the CI, this may not save time overall.

louis-bompart (commented):
I'll be mostly focusing on my area of expertise, i.e. GHA.

Looks good overall! Some tweaks here and there.

Do you have a run somewhere so we see the result?

.github/actions/setup/action.yml
samisayegh: Some thoughts in no particular order:

- Since we are specifying paths in the action, if we rename folders, add new folders, or add new packages, we would need to maintain this point. Do we need to specify exact dirs, or could we just list the package dirs? `atomic`, `auth`, `bueno` etc.

- If the cache is on the commit SHA, then the benefit is limited to when a step on the build fails (e.g. e2e tests). This seems reasonable since this is the scenario where one is retrying. As you and Louis said, using the files as the source of truth would increase the scope of the benefit.

- I wonder if esbuild will reduce the need for a cache? Caches are helpful when a step is slow. If the step is less slow though, then the cache decreases in usefulness. No cache -> less config (first bullet) or the chance of a file not refreshing. Perhaps we are optimizing unnecessarily?

.github/workflows/prbot.yml
samisayegh: I think we can speed this report with esbuild. I want to revisit the bundle size report, perhaps just comparing the browser esm files which we can generate quickly. Let's not optimize for now,

samisayegh (approved):

ThibodeauJF (approved):
Thank god

.github/actions/setup/action.yml
btaillon: > Since we are specifying paths in the action, if we rename folders, add new folders, or add new packages, we would need to maintain this point. Do we need to specify exact dirs, or could we just list the package dirs? `atomic`, `auth`, `bueno` etc.

I could keep the whole dir and exclude `node_modules`. It just felt a bit weird to cache everything, but logically I don't see much of an issue with it.

> I wonder if esbuild will reduce the need for a cache? Caches are helpful when a step is slow. If the step is less slow though, then the cache decreases in usefulness. No cache -> less config (first bullet) or the chance of a file not refreshing. Perhaps we are optimizing unnecessarily?

Perhaps. However, even without the cache, having a setup action reduces the length of the workflow.

.github/actions/setup/action.yml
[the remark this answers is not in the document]
btaillon: I tried to use `setup-node`'s cache, but it didn't appear to be compatible with `lerna`. It seems like it only cached the root `node_modules`.

.github/actions/setup/action.yml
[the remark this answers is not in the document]
btaillon: Correct.

I addressed this [above](https://github.com/coveo/ui-kit/pull/1567#discussion_r764144387), but it looks like we wrote both of our comments at once. This may be over-optimizing, since builds are gonna become faster with esbuild.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    - uses: actions/cache@v2
+      with:
+        path: |
+          node_modules
+          */*/node_modules
+        key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
+    - run: |
+        npm ci
+        npx lerna bootstrap --ci
```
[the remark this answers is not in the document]
btaillon: > Can't you have a `setup` or `postinstall` script at the top level to do that?

I'm not sure what you mean, every sample I found online used `npm install` the same way., even with setup-node. I also don't see an advantage to moving this elsewhere since it's still only two lines.

> If possible, avoid the use of npx in CI because you then drop the control of the version of the package that you invoke.
> 
> If not possible, fix the version of lerna when invoking it.

I will avoid npx or I will fix the version.

.github/workflows/prbot.yml
```diff
@@ -1,18 +1,90 @@
...
+      - run: npx lerna run copy:staticresources --scope @coveo/quantic
+      - run: npm test
+  e2e-test:
+    name: "Run e2e tests"
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        containers: [1, 2, 3]
```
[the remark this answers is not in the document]
btaillon: Copied it from here: https://github.com/coveo/ui-kit/pull/1306

It splits E2E tests over three containers and runs them in parallel to reduce the overall time it takes to complete E2E tests.

Jenkinsfile
btaillon: it's not dead yet, it will be in the next PR.

louis-bompart: GOOD! 😈

.github/workflows/prbot.yml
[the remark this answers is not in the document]
louis-bompart: 👍 I'll add a comment linking to https://github.com/cypress-io/github-action so it's clear it's some Cypress shenanigans.
Otherwise, future us will forget.

.github/actions/setup/action.yml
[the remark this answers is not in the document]
louis-bompart: For CI/CD, I tend to follow this meme:
image

So, what I mean is that I think, if possible, you should use the human command for the CI too.
I reckon that `npm run setup` on the root of the project should work properly because Lerna bootstrap will acquire the CI flag from the environment variable, same for `npm i`.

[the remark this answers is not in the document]
btaillon: `npm run setup` also runs `npm run lockLernaDependencies`, which is explicitly made to generate `package-lock.json` files. I'd prefer not to generate those.

.github/workflows/prbot.yml
btaillon: I just checked, and running `npx` with a package that isn't installed prompts us to install it.

.github/actions/setup/action.yml
btaillon: update: turns out that trying to save everything and excluding node_modules & src creates a cache of about 427MB. Comparatively, if we're more specific, it's 15MB. Given that there's a limit on how much cache we can have, I will stay with the more specific solution.

btaillon:
Checks passed, but one old check was still showing up despite no longer being required (as proven by the `master` branch PR). I tried to close and reopen the PR, but that didn't fix it, so I force merged.

PR #1568, opened 2021-12-06
chore: activate isolatedModules ts flag

samisayegh:
**Context**
Similar to [Babel](https://www.typescriptlang.org/tsconfig#isolatedModules), esbuild recommends enabling the [isolatedModules](https://esbuild.github.io/content-types/#isolated-modules) flag. When the flag is true, we need to specify when we are exporting a type vs. a value.

**Caveat**
However, this flag prevents importing `const` enums in the typescript code base. It turns out there is a difference between defining a non-const enum as,
```
export enum Animal {} -> generates an object.
```
versus using the `const` keyword
```
export const enum Animal {} -> does not generate an object
``` 
I recommend this [write up](https://ncjamieson.com/dont-export-const-enums/) for more details.

**Issue**
The `api-extractor-model` is [exporting const enums](https://github.com/microsoft/rushstack/blob/master/apps/api-extractor-model/src/items/ApiItem.ts#L18), which we are using in the doc-parser. The `isolatedModules` flag detects these instances and throws errors.

**Solutions**
- Set the `isolatedModules` flag in the `tsconfig.build.json` scoped to just headless (current PR). The problem with this is there are no warnings in the IDE. Errors are caught only at "build" time when generating type declarations, instead of at compile time.
- Adjust doc-parser instances to not use `api-extractor-model`s exported enums (i.e. duplicate/cast where needed). This would be better short-term, but longer-term could introduce issues if `api-extractor-model` adjusts their enums.
- Adjust the `api-extractor-model` project to not use `const` before the `enum`. I will explore with them, since they did this for [another project](https://github.com/microsoft/rushstack/pull/3060) a few days ago.

### DIFF
--- a/packages/headless/esbuild.js
+++ b/packages/headless/esbuild.js
@@ -1,0 +1,148 @@
+const path = require('path');
+const {readFileSync} = require('fs');
+const {build} = require('esbuild');
+const alias = require('esbuild-plugin-alias');
+
+const useCaseEntries = {
+  search: 'src/index.ts',
+  recommendation: 'src/recommendation.index.ts',
+  'product-recommendation': 'src/product-recommendation.index.ts',
+  'product-listing': 'src/product-listing.index.ts',
+  'case-assist': 'src/case-assist.index.ts',
+}
+
+function getGlobalName(useCase) {
+  const map = {
+    search: 'CoveoHeadless',
+    recommendation: 'CoveoHeadlessRecommendation',
+    'product-recommendation': 'CoveoHeadlessProductRecommendation',
+    'product-listing': 'CoveoHeadlessProductListing',
+    'case-assist': 'CoveoHeadlessCaseAssist',
+  }
+
+  const globalName = map[useCase];
+
+  if (globalName) {
+    return globalName;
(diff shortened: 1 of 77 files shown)


btaillon (approved):
I like it when things are explicit. Hopefully this doesn't hurt type checking in any way, though.

olamothe (commented):
How does this affect our "day to day" job ?

Say I try to export a new type from some global file (not an enum).

```typescript
// file foo.ts
export interface Foo {
  bar: number;
}

// file index.ts
export {Foo} from './foo';
// whoopsies, should be export type {Foo} from './foo'
```

Where/when exactly is this going to fail ? That's unclear to me.


edit: I did the experiment locally, and it's when we compile. So, all good 👍

olamothe (approved):

ThibodeauJF (approved):

samisayegh:
> I did the experiment locally, and it's when we compile

Yes, for this PR, it's at compilation, but in the [next one](https://github.com/coveo/ui-kit/pull/1571), the errors will be visible in the IDE.

PR #1573, opened 2021-12-07
fix(quantic): Fix flaky test in E2E facet tests

aelfaleh:
image
image
image
image


packages/quantic/cypress/integration/facets/facet/facet.cypress.ts
```diff
@@ -381,11 +383,14 @@ describe('Facet Test Suite', () => {
...
-        label: defaultLabel,
-        displayValuesAs: 'link',
-      });
+      visitFacetPage(
+        {
+          field: defaultField,
+          label: defaultLabel,
+          displayValuesAs: 'link',
+        },
+        false
```
nathanlb, line 392: Could you help me understand why we explicitely don't want to wait for search to return in this case?

aelfaleh: I have expectations that need to wait for search

nathanlb: The first assertion in the test seems to assume the search has returned though. Could you show me which one requires it please?
```
Expect.displayPlaceholder(false);
```

packages/quantic/cypress/integration/facets/facet/facet.cypress.ts
```diff
@@ -423,6 +431,7 @@ describe('Facet Test Suite', () => {
 
         function collapseFacet() {
           Actions.clickCollapseButton();
+          cy.wait(InterceptAliases.Search);
```
lbergeron, line 434: Collapsing a facet does not trigger a search query. We shouldn't wait here.

lbergeron (approved):
Well done!

lvu285 (approved):

packages/quantic/cypress/integration/facets/facet/facet.cypress.ts
aelfaleh: `    aliasFacetValues();
        Expect.facetValuesEqual(indexFacetValuesAlias);`

` function aliasFacetValues() {
    cy.wait(InterceptAliases.Search).then((interception) => {
      const indexValues = extractFacetValues(interception.response);
      cy.wrap(indexValues).as(indexFacetValuesAlias.substring(1));
    });
  }`

PR #1574, opened 2021-12-07
ci: migrate @coveo/bueno to esbuild

samisayegh:
https://coveord.atlassian.net/browse/KIT-1281


packages/bueno/rollup.config.js
```diff
@@ -23,46 +26,25 @@ function onWarn(warning, warn) {
...
-    },
-    {
-      file: 'dist/browser/bueno.esm.js',
-      format: 'es',
-      sourcemap: true,
-    },
-  ],
+  output: {
+    file: 'dist/browser/bueno.js',
+    format: 'umd',
```
samisayegh, line 33: Since bueno is not being served on CDN,  a UMD bundle should in theory not be necessary.

I kept the bundle to remove any risk of a breaking change, but I think ESM should be enough since consumers of the package must be using a bundler?

Removing UMD support would allow us to remove the rollup config. The motivation would be to simplify the config to a single bundler; the time savings are small in this case.

With `@coveo/auth`, I opted to not provide a UMD as an experiment. We can decide to do the same with bueno at a later stage depending on how it goes,

packages/bueno/esbuild.js
```diff
@@ -0,0 +1,52 @@
...
+
+function browserEsm() {
+  return build({
+    ...base,
+    platform: 'browser',
+    outfile: 'dist/browser/bueno.esm.js',
+    format: 'esm',
+    watch: devMode,
+    minify: true,
+    sourcemap: true,
```
samisayegh, line 44: Since bueno is not served on a CDN, rather consumed via a bundler, in theory these minification options should not be needed (consumer can choose whether or not to minify). I kept them to minimize differences with the bundle generated by rollup, but I wouldn't mind removing them.

Thoughts?

ThibodeauJF: You can remove them

packages/bueno/rollup.config.js
ThibodeauJF: We can also do a major of bueno if that's an issue, I don't think it's that used except internally

ThibodeauJF (approved):

btaillon (approved):

packages/bueno/rollup.config.js
samisayegh: Since we now have a work-around to generate UMDs with esbuild, perhaps the best approach is to keep the UMD to minimize the chance of a breaking change, while still removing the rollup configuration. I will go with this 🤞

PR #1575, opened 2021-12-08
feat(quantic): slots added in quick view

mmitiche:
# Slots Added in Quick view

- ### The button to open the Quick view is now customizable using the following properties  `previewButtonLabel`, `previewButtonIcon` and `previewButtonVariant`.

- ### The ability to have a footer in the quick view where the end user can plug other components using the  slot `footer`.

https://user-images.githubusercontent.com/86681870/145244465-1e7764c8-c7c8-404e-b712-88c87f3e4bc9.mov

## Examples of the  preview buttons that can be displayed: 

- ### Default:
default

- ### Default when there is no preview
defaultDisabled

- ### Brand variant with custom label and custom icon
BrandCustomIconCustomLAbel

- ### Brand variant with custom label and custom icon when there is no preview
brandCustomDisabled

- ### Outline variant with custom label and custom icon
OutlineCustom

- ### Brand variant with custom label and custom icon when there is no preview
OutlineCustomDisabled

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
@@ -18,12 +18,12 @@
   height: 50%;
 }
 
-.quickview__button_base{
+.quickview__button-base{
   border-radius: 0.25rem;
   color: var(--color-text-link-active, #0f2d5d);
 }
 
-.quickview__button_base:hover {
+.quickview__button-base:not([disabled]):hover {
   background-color: #f6f7f9;
   box-shadow: 0 0 0px 0.25rem #f6f7f9;
 }
--- a/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
@@ -18,14 +18,13 @@
   height: 50%;
 }
 
-.quickview__button {
-  color: var(--color-text-link-active, #0f2d5d);
+.quickview__button_base{
   border-radius: 0.25rem;
+  color: var(--color-text-link-active, #0f2d5d);
 }
 
-.quickview__button:hover {
+.quickview__button_base:hover {
   background-color: #f6f7f9;
-  box-shadow: 0 0 0px 0.25rem #f6f7f9;
 }
 
 .slds-modal__footer{
--- a/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
@@ -25,6 +25,7 @@
 
 .quickview__button_base:hover {
   background-color: #f6f7f9;
+  box-shadow: 0 0 0px 0.25rem #f6f7f9;
 }
 
 .slds-modal__footer{
--- a/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
@@ -18,7 +18,7 @@
   height: 50%;
 }
 
-.quickview__button-base{
+.quickview__button-base:not([disabled]){
   border-radius: 0.25rem;
   color: var(--color-text-link-active, #0f2d5d);
 }
--- a/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
@@ -27,3 +27,8 @@
   background-color: #f6f7f9;
   box-shadow: 0 0 0px 0.25rem #f6f7f9;
 }
+
+.slds-modal__footer{
+  background-color: white;
+  text-align: unset;
+}
--- a/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
@@ -169,9 +169,6 @@ export default class QuanticResultQuickview extends LightningElement {
       if (!this.hasNoPreview) {
         classNames.push('quickview__button_base');
       }
-      if (this.previewButtonLabel) {
-        classNames.push('slds-var-p-horizontal_medium');
-      }
     }
     else {
       if (this.previewButtonVariant === 'brand') {
--- a/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
@@ -59,7 +59,7 @@ export default class QuanticResultQuickview extends LightningElement {
    * @type {undefined|'brand'|'outline-brand'}
    * @defaultValue `undefined`
    */
-  @api previewButtonVariant = undefined;
+  @api previewButtonVariant;
 
   /** @type {QuickviewState} */
   @track state;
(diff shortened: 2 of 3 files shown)


packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
```diff
@@ -18,12 +18,17 @@
   height: 50%;
 }
 
-.quickview__button {
-  color: var(--color-text-link-active, #0f2d5d);
+.quickview__button_base{
```
nathanlb, line 21:
```suggestion
.quickview__button-base{
```

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -37,6 +37,30 @@ export default class QuanticResultQuickview extends LightningElement {
...
+   */
+  @api previewButtonLabel = '';
+
+  /**
+   * The variant of the preview button.
+   * @api
+   * @type {''|'brand'|'outline'}
+   * @defaultValue `''`
+   */
+  @api previewButtonVariant = '';
```
nathanlb, line 62: I think it would be better to default the variant and label props to undefined rather than empty strings. and add a conditional in the markup for rendering the label. `''` isn't really a variant and `''` doesn't accurately represent "no label" IMO.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -76,7 +100,7 @@ export default class QuanticResultQuickview extends LightningElement {
       result: this.result,
       maximumPreviewSize: Number(this.maximumPreviewSize),
     }
-    this.quickview = CoveoHeadless.buildQuickview(engine, {options});
+    this.quickview = CoveoHeadless.buildQuickview(engine, { options });
```
nathanlb, line 103: we can revert this prettier change. sidenote: we need to edit the rules because it is indeed annoying that it does this by default. 🙈

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -98,7 +122,7 @@ export default class QuanticResultQuickview extends LightningElement {
 
   addRecentResult() {
     getHeadlessEnginePromise(this.engineId).then((engine) => {
-      const {pushRecentResult} = CoveoHeadless.loadRecentResultsActions(engine);
+      const { pushRecentResult } = CoveoHeadless.loadRecentResultsActions(engine);
```
nathanlb, line 125: can revert this one too

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
```diff
@@ -18,12 +18,17 @@
...
   border-radius: 0.25rem;
+  color: var(--color-text-link-active, #0f2d5d);
 }
 
-.quickview__button:hover {
+.quickview__button_base:hover {
```
nathanlb, line 26:
```suggestion
.quickview__button-base:not([disabled]):hover {
```

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -140,4 +164,27 @@ export default class QuanticResultQuickview extends LightningElement {
   get buttonLabel() {
     return this.hasNoPreview ? this.labels.noPreview : this.labels.openPreview;
   }
+
+  get buttonClass() {
```
nathanlb, line 168: I think this can be simplified if you make the possible variant values: `'brand' | 'outline-brand'`;
```suggestion
  get buttonClass() {
    const variantClass = this.previewButtonVariant
      ? `slds-button_${previewButtonVariant}`
      : 'quickview__button-base';
    return `slds-button ${variantClass}`;
  }
```

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
nathanlb: Variant could also be 'base' I guess too. Just an option.

aelfaleh (approved):

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -140,4 +164,27 @@ export default class QuanticResultQuickview extends LightningElement {
...
+    let classNames = ['slds-button'];
+
+    if (!this.previewButtonVariant || !this.previewButtonLabel) {
```
lbergeron, line 171: I think there is an issue with the condition here. Suppose the component is set like that:

    <c-quantic-result-quickview preview-button-variant="brand" preview-button-label="" ...>

As `previewButtonVariant` is truthy and `previewButtonLabel` is falsy, we'll enter the `if`, and the class `slds-button_brand` would not be added to the button.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -140,4 +164,27 @@ export default class QuanticResultQuickview extends LightningElement {
...
+    else {
+      classNames.push(`slds-button_${this.previewButtonVariant}`);
+    }
+
+    return classNames.join(' ');
+  }
+
+  get buttonIconClass() {
+    return `slds-current-color ${!this.previewButtonLabel ? '' : 'slds-button__icon_right'}`;
+  }
```
lbergeron, line 185: When it comes to handling class names, I often prefer to use an array instead of string interpolation.

```suggestion
  get buttonIconClass() {
    return [
      'slds-current-color',
      this.previewButtonLabel && 'slds-button__icon_right'
    ].join(' ');
  }
```

lbergeron (changes requested):
Please validate the corner case, and fix if needed. The rest looks good to me.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
mmitiche: Actually, I made this on purpose, to avoid having buttons  like this : 

outline

or this  :

brandEmpty


So I made this in  way that when there is no label and we will have just a button containing an icon I force the button to be in the base variant  to have always something like this: 

normalDefault

@lbergeron,  does this make sense or we need to allow having any variant even when there is no variant ?

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
nathanlb: You don't like my suggestion? :P

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
lbergeron: Personally, I prefer to apply the variant when it is set.

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -37,6 +37,30 @@ export default class QuanticResultQuickview extends LightningElement {
...
+   */
+  @api previewButtonIcon = 'utility:preview';
+
+  /**
+   * The label to be shown in the preview button.
+   * @api
+   * @type {string}
+   * @defaultValue `undefined`
+   */
+  @api previewButtonLabel = undefined;
```
nathanlb, line 54:
```suggestion
  @api previewButtonLabel;
```

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -37,6 +37,30 @@ export default class QuanticResultQuickview extends LightningElement {
...
+
+  /**
+   * The variant of the preview button.
+   * @api
+   * @type {undefined|'brand'|'outline-brand'}
+   * @defaultValue `undefined`
+   */
+  @api previewButtonVariant = undefined;
```
nathanlb, line 62:
```suggestion
  @api previewButtonVariant;
```

nathanlb (approved):
No need to explicitely set to 'undefined' :)
Nice work!

PR #1576, opened 2021-12-08
ci: migrated master branch CI to GitHub Actions

btaillon:
https://coveord.atlassian.net/browse/KIT-1278

@coveo/service-integration does this affect your release process?

Jenkins is removed (almost) for good in this PR. It's still run by the deployment pipeline in QA and prod.


.github/workflows/masterbot.yml
```diff
@@ -0,0 +1,90 @@
+name: Master Robot
+on:
+  push:
+    branches:
+      - master
+    tags-ignore: ['*']
+jobs:
+  build:
```
btaillon, line 8: I wanted to extract all of these jobs into [reusable workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows). However, it seems like, unlike [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action), we cannot call a reusable workflow that is part of the HEAD commit.

.github/actions/create-deployment-package/action.yml
```diff
@@ -0,0 +1,31 @@
+name: "Create deployment package"
```
btaillon, line 1: This action contains mostly what I copied from [dev tooling's example](https://github.com/coveo/deployment-pipeline/blob/master/.github/workflows/example-github-actions.yml).

.github/actions/create-deployment-package/action.yml
```diff
@@ -0,0 +1,31 @@
...
+description: "Create a deployment package for Headless and Atomic"
+runs:
+  using: composite
+  steps:
+    - id: parse-version
+      uses: ./.github/actions/parse-version
```
btaillon, line 7: I created this JS action to parse the package version and access it in other steps.

.github/actions/parse-version/action.yml
```diff
@@ -0,0 +1,14 @@
+name: "Parse versions"
```
btaillon, line 1: This is a [javascript action](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action). It uses its own version of node, independent of the job it's running in.

.github/actions/parse-version/index.js
```diff
@@ -0,0 +1,19 @@
+const {setOutput} = require('@actions/core');
```
btaillon, line 1: This code is ran by the `parse-version` action.

.github/workflows/masterbot.yml
```diff
@@ -0,0 +1,90 @@
+name: Master Robot
+on:
+  push:
+    branches:
+      - master
+    tags-ignore: ['*']
```
btaillon, line 6: The intent for this workflow is to run on all commits on `master` except tagged commits, which I assume to always be version bumps.

.github/workflows/publishbot.yml
```diff
@@ -0,0 +1,46 @@
+name: Publish Robot
+on:
+  push:
+    branches:
+      - master
+    tags: ['*']
```
btaillon, line 6: As opposed to `masterbot.yml`, I intend for this workflow to run on tagged commits.

.github/workflows/publishbot.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup
+      - run: npm run npm:publish:alpha
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+  deploy:
+    name: 'Deploy the package'
+    needs: build
```
btaillon, line 30: Should we only deploy if we publish successfully? Should we only publish if we deploy successfully? It looks like in Jenkins we did both regardless, and I don't see much of a reason not to do that here either. Both of these jobs run in parallel.

.github/workflows/publishbot.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: ./.github/actions/setup
+      - run: rm -rf veracode && mkdir veracode
+      - run: mkdir veracode/auth
+      - run: cp -R packages/auth/src packages/auth/package.json packages/auth/package-lock.json veracode/auth
+      - run: mkdir veracode/bueno
+      - run: cp -R packages/bueno/src packages/bueno/package.json packages/bueno/package-lock.json veracode/bueno
+      - run: mkdir veracode/headless
+      - run: cp -R packages/headless/src packages/headless/package.json packages/headless/package-lock.json veracode/headless
+      - run: mkdir veracode/atomic
+      - run: cp -R packages/atomic/src packages/atomic/package.json packages/atomic/package-lock.json veracode/atomic
```
btaillon, line 45: We could:
* create the veracode package here (current solution)
* extract it into a dedicated action
* move it into `create-deployment-package`

I don't feel strongly about any of those.

btaillon, line 45: We could:
* create the veracode package here (current solution)
* create the veracode package in a dedicated action
* create the veracode package in `create-deployment-package`

I don't feel strongly about any of those.

.github/workflows/prbot.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         with:
           working-directory: ./packages/atomic
           parallel: true
-          record: true
```
btaillon, line 78: Forgot to disable recording cypress tests on PRs...

.github/actions/create-deployment-package/action.yml
louis-bompart: I think `@action/core` is pretty much OK.

.github/workflows/masterbot.yml
```diff
@@ -0,0 +1,87 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup
+      - run: npx --no-install lerna run copy:staticresources --scope @coveo/quantic
+      - run: npm test
+  e2e-atomic-test:
+    name: "Run e2e tests on Atomic"
```
btaillon, line 48: I didn't add the matrix/parallel tests to the `master` branch either, since it may be complicated to define a `strategy` depending on whether or not we're on the `master` branch, and I don't want to have too many issues on the `master` branch to fix once this is merged. I can do it in a separate PR, if you think it's important.

.github/workflows/publishbot.yml
samisayegh: We could make a successful publish a prerequisite to creating a deployment package, since the qa and prd steps will be trying to adjust the tag of the package.

.github/workflows/publishbot.yml
samisayegh: Indeed, there are plans to mode to Snyk code next year, but I don't think the deployment pipeline has a certifier for it yet. I think the current solution looks good in the interim,

samisayegh (approved):

.github/actions/parse-version/index.js
```diff
@@ -0,0 +1,19 @@
...
+const headlessPackage = require('../../../packages/headless/package.json');
+const atomicPackage = require('../../../packages/atomic/package.json');
+
+/**
+ * @param {string} fullVersion
+ */
+function getVersions(fullVersion) {
+  const [minor, major] = /^([^\.]*)\.[^\.]*/.exec(fullVersion);
+  return {minor, major};
+}
```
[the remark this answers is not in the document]
btaillon: The split and the alternative regexp would not give the same result as the current solution, since `minor` in my example contains a period (e.g., `2.19`) and that's what we need.

I'll add `semver`.

.github/workflows/publishbot.yml
btaillon: if we switch to Snyk Code, I'd prefer to do it in a separate PR. I'll ask about it.

PR #1579, opened 2021-12-08
ci(headless): create umd bundle using esbuild

samisayegh:
Jenkins build time: 115s -> 22s
See it in action: [salesforce community](https://sandbox-ruby-velocity-5596-dev-ed-17d7c30c6e8.cs77.force.com/examples/s/full-search-example#q=wow&tab=All)

### Context

Generating a `cjs` bundle and wrapping it with `umd` boilerplate works ...
Sourcemaps work too (not in the demo though because the map files are not being uploaded),

One thing that rollup adds are statements that look like this:

```
Object.defineProperty(exports, "__esModule", { value: true })
```

Based on [this SO post](https://stackoverflow.com/questions/50943704/whats-the-purpose-of-object-definepropertyexports-esmodule-value-0), allow importing `default` exports from a UMD bundle.

However, there are no `default` exports in headless.

Also, these statements are primarily used by bundlers, but bundlers typically default to ESM rather than UMD since it has the benefit of being treeshakable.

### Summary

Ideally, esbuild would generate the bundle. It's likely [this will happen](https://github.com/evanw/esbuild/issues/819#:~:text=Once%20esbuild%20is%20more%20mature%20it%20could%20make%20more%20sense%20to%20add%20more%20output%20formats) one day.

The work-around in this PR might be good enough until that day though 🤞. The quantic e2e tests pass, and using esbuild speeds up the build, and simplifies the configuration by removing rollup.

Benefit of using rollup is a smaller bundler size: 243kb (rollup) vs 261kb (esbuild)


scripts/reports/pr-report.js
```diff
@@ -1,4 +1,4 @@
-const {buildBundleSizeReport} = require('./bundle-size/bundle-size');
+// const {buildBundleSizeReport} = require('./bundle-size/bundle-size');
```
olamothe, line 1: Cleanup ?

scripts/bundle/umd.js
```diff
@@ -0,0 +1,12 @@
+function umdWrapper(globalName) {
+  const header = `(function (global, factory) {
+    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
```
olamothe, line 3: Could we use a little less ternary here 😂  ?

I understand it's just some boilerplate, but... if we ever need to debug this or understand why Headless does not load properly in a specific scenario or environment, this would be hell to step through.

olamothe (approved):

btaillon (approved):

PR #1580, opened 2021-12-08
chore(atomic): bump stencil/core & remove warning

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1286

### DIFF
--- a/packages/atomic/stencil.config.ts
+++ b/packages/atomic/stencil.config.ts
@@ -26,6 +26,7 @@ function replace() {
   return replacePlugin({
     'process.env.NODE_ENV': JSON.stringify(env),
     'process.env.VERSION': JSON.stringify(version),
+    preventAssignment: true,
   });
 }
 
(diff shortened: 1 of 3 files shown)


packages/atomic/stencil.config.ts
```diff
@@ -26,6 +26,7 @@ function replace() {
   return replacePlugin({
     'process.env.NODE_ENV': JSON.stringify(env),
     'process.env.VERSION': JSON.stringify(version),
+    preventAssignment: true,
```
ThibodeauJF, line 29: this removes a warning, in an upcoming version, true will be the default

packages/atomic/stencil.config.ts
```diff
@@ -84,6 +85,7 @@ export const config: Config = {
     reloadStrategy: 'pageReload',
   },
   plugins: [
+    // https://github.com/fabriciomendonca/stencil-inline-svg/issues/16
```
ThibodeauJF, line 88: another warning we had, issue was already opened, might keep it here a bit and then just make a PR there if nothing is done

olamothe (approved):

PR #1581, opened 2021-12-08
ci: revert migration of pr CI to GitHub actions

btaillon:
https://coveord.atlassian.net/browse/KIT-1288

Looks like my last change broke the build. Notably, the required verifications still expect one of the jobs I removed to pass. Additionally, I accidentally removed (or rather forgot to add) cypress tests for Quantic, and I made all Cypress tests record in PRs... I will fix all of this soon. For now, this reverses those changes.


btaillon:
Turns out that the check was a required check under https://github.com/coveo/ui-kit/settings/branch_protection_rules/18402434.
I removed it, so this revert is no longer necessary.

PR #1582, opened 2021-12-08
doc(atomic): add sample for using Headless through Atomic

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1289

It was a question asked on the Coveo Community https://coveocommunity.slack.com/archives/C02DY81L6GM/p1638973899000900

Responded and kept the sample so we can reference to it. @dlutzCoveo  we should write a usage article on Using Headless along with Atomic, the html I provide here should help!

### DIFF
--- a/packages/atomic/src/pages/examples/headless.html
+++ b/packages/atomic/src/pages/examples/headless.html
@@ -1,0 +1,54 @@
+<!DOCTYPE html>
+<html dir="ltr" lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />
+    <title>Coveo Atomic: Headless example</title>
+
+    <script type="module" src="/build/atomic.esm.js"></script>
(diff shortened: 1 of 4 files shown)


packages/atomic/src/pages/examples/headless.html
```diff
@@ -0,0 +1,54 @@
...
+        const action = loadAdvancedSearchQueryActions(engine).updateAdvancedSearchQueries({
+          aq: '@author=amoreau',
+        });
+        engine.dispatch(action);
+      }
+
+      function setContext(engine) {
+        const action = loadContextActions(engine).addContext({
+          contextKey: 'headless',
+          contextValue: 'very good',
```
olamothe, line 25: 😂 

While I personally find it funny, I think we should just add something more generic, like `userGroup: 'sales'` or something like that.

olamothe (approved):

packages/atomic/src/pages/examples/headless.html
ThibodeauJF: yeah I had no idea what to put here

PR #1583, opened 2021-12-09
ci(atomic,quantic): added cypress tests to PRs

btaillon:
https://coveord.atlassian.net/browse/KIT-1287

I added Atomic cypress tests and completely neglected to add Quantic cypress tests.

One major change for Atomic's cypress tests in this PR is that they are no longer split over 3 runners (with the `matrix` parameter). Doing so [requires runs to be recorded](https://docs.cypress.io/guides/guides/parallelization#Turning-on-parallelization).

If we eventually have a quota high enough for that, we can re-enable it.


olamothe (approved):

.github/actions/e2e-quantic/action.yml
```diff
@@ -0,0 +1,33 @@
...
+    - run: |
+        ./node_modules/.bin/sfdx force:auth:jwt:grant \
+          --clientid ${{ secrets.SFDX_AUTH_CLIENT_ID }} \
+          --jwtkeyfile ${{ secrets.SFDX_AUTH_JWT_KEY }} \
+          --username rdaccess@coveo.com \
+          --instanceurl https://login.salesforce.com \
+          --setdefaultdevhubusername
+      shell: bash
+      with:
+        working-directory: ./packages/quantic
```
btaillon, line 15: Not sure why syntax highlighting looks wrong for those three lines. In VSCode, they have the correct syntax highlighting.

.github/actions/e2e-quantic/action.yml
```diff
@@ -0,0 +1,33 @@
+name: "Quantic E2E tests"
```
btaillon, line 1: This should be equivalent to what was in the Jenkinsfile.

.github/actions/e2e-quantic/action.yml
```diff
@@ -0,0 +1,33 @@
...
+        install: true
+        config: reporter=cypress/reporters/detailed-reporter.js
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: cypress-screenshots
+        path: packages/quantic/cypress/screenshots
+    - run: ./node_modules/.bin/ts-node scripts/build/delete-org.ts
+      shell: bash
+      if: always()
```
btaillon, line 31: I assume this is a cleanup step, so I added `if: always()` to make sure it runs even if previous steps fail.

.github/actions/e2e-quantic/action.yml
```diff
@@ -0,0 +1,33 @@
+name: "Quantic E2E tests"
+description: "Runs Cypress tests for Quantic"
+runs:
+  using: composite
+  steps:
+    - run: |
+        ./node_modules/.bin/sfdx force:auth:jwt:grant \
```
btaillon, line 7: I could use `npx` for all those commands.

samisayegh (approved):

.github/actions/e2e-quantic/action.yml
louis-bompart: Can you check which version of sfdx got used iif you use `npx` tho?

IIRC, sometimes, @coveo/service-integration wants to use a specific version of the SFDX CLI simply because, well, the latest is brokey.

So I think they'd appreciate that their `package.json` is well respected ;)

btaillon: `npx` looks for the package in the `node_modules/.bin` folder, so it should be equivalent. I will add `--no-install` to every `npx` call to be sure nothing ever gets installed, but when I tested `npx` with a package that isn't installed, it prompted me.

PR #1585, opened 2021-12-10
ci: readd bundle size report

samisayegh:
https://coveord.atlassian.net/browse/KIT-1290

### DIFF
--- a/scripts/reports/bundle-size/command.js
+++ b/scripts/reports/bundle-size/command.js
@@ -24,23 +24,23 @@ function readFileSizes() {
   const dir = resolve(__dirname, '../../../packages/headless/dist/browser');
   const entries = getUseCasesAndFilePaths(dir);
 
-  return entries
+  const sizeEntries = entries
     .map((entry) => {
       const {useCase, filePath} = entry;
       const {size} = statSync(filePath);
-      return {useCase, size};
-    })
-    .reduce((acc, curr) => ({...acc, [curr.useCase]: curr.size}), {});
+      return [useCase, size];
+    });
+  
+  return Object.fromEntries(sizeEntries);
 }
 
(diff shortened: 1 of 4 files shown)


scripts/reports/bundle-size/command.js
```diff
@@ -21,9 +21,50 @@ async function buildFiles() {
...
+  const dir = resolve(__dirname, '../../../packages/headless/dist/browser');
+  const entries = getUseCasesAndFilePaths(dir);
+
+  return entries
+    .map((entry) => {
+      const {useCase, filePath} = entry;
+      const {size} = statSync(filePath);
+      return {useCase, size};
+    })
+    .reduce((acc, curr) => ({...acc, [curr.useCase]: curr.size}), {});
```
btaillon, line 33: I think this would be easier to read using `Object.fromEntries`

scripts/reports/bundle-size/command.js
```diff
@@ -21,9 +21,50 @@ async function buildFiles() {
...
+}
+
+function getUseCasesAndFilePaths(dir) {
+  const filePaths = getEsmFilePaths(dir);
+  return filePaths.map((filePath) => {
+    return {
+      useCase: determineUseCase(dir, filePath),
+      filePath,
+    };
+  });
```
btaillon, line 43: can be shortened with:
```suggestion
  return filePaths.map((filePath) => ({
    useCase: determineUseCase(dir, filePath),
    filePath,
  }));
```

btaillon (approved):

scripts/reports/bundle-size/command.js
samisayegh: Thanks! TIL about this method,

PR #1589, opened 2021-12-13
fix(quantic): Fix the appearance of the preview button of the quick view

mmitiche:
Before: 

- Tooltip appearing only when hovering on the icon.
- When `preview-button-icon` is set to `''` an additional right padding was appearing in the preview button.

before

After:

- Tooltip appearing only when hovering on the whole button.
- When `preview-button-icon` is set to `''` the preview button is now appearing correctly.

fix


nathanlb (approved):
LGTM 👍

lvu285:
Just want to be sure, the background of "Open preview" is not grey?

lvu285 (approved):

aelfaleh (approved):

mmitiche:
> Just want to be sure, the background of "Open preview" is not grey?
The styling of the tooltip is the default styling provided by the browsers since we are not customizing it, so from the screenshots we have I think in windows the background is white and the text color is grey and in MacOS the background is grey and the color is white.

PR #1590, opened 2021-12-13
ci: migrated master branch CI to GitHub Actions 

btaillon:
https://coveord.atlassian.net/browse/KIT-1278

All commits before 715e1fedefecf86611e731fed3e5b4d815eb6da3 were already reviewed.

I reverted the previous PR since it executed both workflows on `master` commits.

I adjusted the conditions for `masterbot.yml` and `publishbot.yml`. I found two forum posts mentioning that `tags` and `branches` are applied as an `OR` condition, and I found in the doc that `branches` and `tags-ignore` are applied as an `AND` condition. I also found out that `*` matches any character except `/`.


.github/workflows/masterbot.yml
```diff
@@ -0,0 +1,86 @@
+name: Master Robot
+on:
+  push:
+    branches:
+      - master
+    tags-ignore:
+      - "@coveo/*"
```
btaillon, line 7: These lines changed in this PR vs the previous PR.

.github/workflows/publishbot.yml
```diff
@@ -0,0 +1,45 @@
+name: Publish Robot
+on:
+  push:
+    tags:
+      - "@coveo/*"
```
btaillon, line 5: These lines changed in this PR vs the previous PR.

samisayegh (approved):

PR #1597, opened 2021-12-14
fix: (headless): Send correct analytics events from Product Listing controllers

pvanool:
A bad request (400) error occurred when sending usage analytics requests from a Product Listing controllers.
After investigation there are 2 problems identified:
1. No SearchUID from product listing response was set.
2. Facet state form Search was queried but should be from Product Listing.

https://coveord.atlassian.net/browse/COM-1341

To validate:
Prerequisite: a page with Product Listing controllers like https://github.com/coveo/saqdemo. 
Open the network tab in chrome dev tools and filter on `search?visitor`.
Do facet related actions.
Previously you would get a 400 response.
Now you get a 200 response when these changes are applied.

### DIFF
--- a/packages/headless/src/features/facets/category-facet-set/category-facet-set-selectors.ts
+++ b/packages/headless/src/features/facets/category-facet-set/category-facet-set-selectors.ts
@@ -1,9 +1,8 @@
+import {CategoryFacetSection} from '../../../state/state-sections';
 import {
-  CategoryFacetSection,
-  ProductListingSection,
-  SearchSection,
-} from '../../../state/state-sections';
-import {baseFacetResponseSelector} from '../facet-set/facet-set-selectors';
+  baseFacetResponseSelector,
+  FacetResponseSection,
+} from '../facet-set/facet-set-selectors';
 import {partitionIntoParentsAndValues} from './category-facet-utils';
 import {CategoryFacetResponse} from './interfaces/response';
 import {AnyFacetResponse} from '../generic/interfaces/generic-facet-response';
@@ -16,7 +15,7 @@ function isCategoryFacetResponse(
 }
 
 export const categoryFacetResponseSelector = (
-  state: CategoryFacetSection & Partial<SearchSection | ProductListingSection>,
+  state: CategoryFacetSection & Partial<FacetResponseSection>,
   facetId: string
 ) => {
   const response = baseFacetResponseSelector(state, facetId);
@@ -35,7 +34,7 @@ export const categoryFacetRequestSelector = (
 };
 
 export const categoryFacetSelectedValuesSelector = (
-  state: CategoryFacetSection & Partial<SearchSection | ProductListingSection>,
+  state: CategoryFacetSection & Partial<FacetResponseSection>,
   facetId: string
 ) => {
   const facetResponse = categoryFacetResponseSelector(state, facetId);
--- a/packages/headless/src/features/facets/category-facet-set/category-facet-set-selectors.ts
+++ b/packages/headless/src/features/facets/category-facet-set/category-facet-set-selectors.ts
@@ -1,5 +1,6 @@
 import {
   CategoryFacetSection,
+  ProductListingSection,
   SearchSection,
 } from '../../../state/state-sections';
 import {baseFacetResponseSelector} from '../facet-set/facet-set-selectors';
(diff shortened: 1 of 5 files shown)


packages/headless/src/features/facets/facet-set/facet-set-selectors.ts
```diff
@@ -7,18 +7,21 @@ import {FacetResponse, FacetValue} from './interfaces/response';
 import {AnyFacetResponse} from '../generic/interfaces/generic-facet-response';
 
 export const baseFacetResponseSelector = (
-  state: SearchSection | ProductListingSection,
+  state: Partial<SearchSection | ProductListingSection>,
   id: string
 ) => {
-  if ('productListing' in state) {
+  if ('productListing' in state && state.productListing) {
```
btaillon, line 13: I think the following would be easier to read:
```suggestion
  if (state?.productListing) {
```

packages/headless/src/features/facets/facet-set/facet-set-selectors.ts
```diff
@@ -7,18 +7,21 @@ import {FacetResponse, FacetValue} from './interfaces/response';
...
     return state.productListing.facets.results.find(
       (response) => response.facetId === id
     );
   }
 
-  return state.search.response.facets.find(
-    (response) => response.facetId === id
-  );
+  if ('search' in state && state.search) {
```
btaillon, line 19: Same as above
```suggestion
  if (state?.search) {
```

packages/headless/src/features/facets/facet-set/facet-set-analytics-actions-utils.ts
```diff
@@ -139,6 +148,13 @@ export const buildFacetStateMetadata = (
   return facetState;
 };
 
+const getFacetResponse = (state: SectionNeededForFacetMetadata) => {
+  if ('productListing' in state && state.productListing)
```
btaillon, line 152: Same as above
```suggestion
  if (state?.productListing)
```

packages/headless/src/features/facets/facet-set/facet-set-analytics-actions-utils.ts
```diff
@@ -139,6 +148,13 @@ export const buildFacetStateMetadata = (
...
+    return state.productListing.facets.results;
+  if ('search' in state && state.search) return state.search.response.facets;
```
btaillon, line 154:
```suggestion
  if (state?.search) return state.search.response.facets;
```

btaillon (approved):

packages/headless/src/features/facets/facet-set/facet-set-selectors.ts
pvanool: It is, but the `in` operator is used by typescript to narrow its type to being a `ProductListingSection` type inside the if. 

I was actually contemplating to use a type predicate function for readability, like `isProductListingSection`.

packages/headless/src/features/facets/category-facet-set/category-facet-set-selectors.ts
```diff
@@ -15,7 +16,7 @@ function isCategoryFacetResponse(
 }
 
 export const categoryFacetResponseSelector = (
-  state: CategoryFacetSection & SearchSection,
+  state: CategoryFacetSection & Partial<SearchSection | ProductListingSection>,
```
samisayegh, line 19: Is this type change needed?

It seems like both `SearchSection` and `ProductListingSection` are optional?
I also don't see the change scaling nicely in the code-base as more use-cases are added.

We could store the union in a type to remove the duplication,

```
type CategoryFacetResponseSection = SearchSection | ProductListingSection;

state: CategoryFacetSection & CategoryFacetResponseSection
```

but if no change is needed, all the better,

samisayegh (approved):
The changes will do 👍  Ideally, the product-listing and search use-cases would use independent analytics clients. However, refactoring now is not worthwhile since analytics will be reworked in the near future.

Spuffynism (approved):

dlafreniere (approved):
with Sami's suggestion, I think its good

packages/headless/src/features/facets/category-facet-set/category-facet-set-selectors.ts
pvanool: I've created a new type to store the union of search and product-listing section.
The type change is needed to make a distinction between a section coming from product-listing or search when coming from the analytics branch. This part can (or better must) be removed when the analytics have independent product-listing or search clients.

PR #1598, opened 2021-12-14
docs(headless): update sortCriteria to describe 'automatic' default

dlutzCoveo:
Updated sortCriteria annotation for facets in Headless to explain the default behaviour of `automatic`. Prompted by [this discussion](https://coveord.atlassian.net/browse/DOC-10113).

### DIFF
--- a/packages/headless/src/controllers/core/facets/facet/headless-core-facet-options.ts
+++ b/packages/headless/src/controllers/core/facets/facet/headless-core-facet-options.ts
@@ -62,7 +62,8 @@ export interface FacetOptions {
   numberOfValues?: number;
 
   /**
-   * The criterion to use for sorting returned facet values. The default value of `automatic` adjusts the sort based on whether the facet is expanded or not, and whether the full list of values is being displayed.
+   * The criterion to use for sorting returned facet values.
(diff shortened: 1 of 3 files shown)


olamothe (approved):

packages/headless/src/controllers/core/facets/facet/headless-core-facet-options.ts
```diff
@@ -62,7 +62,7 @@ export interface FacetOptions {
   numberOfValues?: number;
 
   /**
-   * The criterion to use for sorting returned facet values.
+   * The criterion to use for sorting returned facet values. The default value of `automatic` adjusts the sort based on whether the facet is expanded or not, and whether the full list of values is being displayed.
```
samisayegh, line 65: The intent of explaining inline is good, but I feel the extra text is still vague. If I was trying to understand the behaviour of `automatic`, I still would not understand after reading the additional text.

I think an clear explanation has to include when the values will be ordered alphanumerically vs. by score.

While we could inline this info, I think linking to the [search api docs](https://docs.coveo.com/en/1461/build-a-search-ui/query-parameters#RestFacetRequest-sortCriteria) could be better to keep a single source of truth.

samisayegh (approved):

samisayegh (approved):

PR #1599, opened 2021-12-15
feat(quantic): Case Classification Component

mmitiche:
# Case Classification Component


packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,273 @@
...
+   *  */
+  get filtredOptions() {
+    return this.options.filter((option) => {
+      return (
+        this.classifications.findIndex(
+          (item) => item.value === option.value
+        ) === -1
+      );
+    });
+  }
```
lbergeron, line 182:
```suggestion
  get filteredOptions() {
    return this.options.filter((option) => 
      !this.classifications.some((item) => item.value === option.value)
    );
  }
```

lbergeron (commented):
I didn't review all the changes, mostly did a quick check on the component itself. The Cypress tests are not running on my side though. I guess it is because this task is not yet completed (https://coveord.atlassian.net/browse/SFINT-4327).

mmitiche:
> I didn't review all the changes, mostly did a quick check on the component itself. The Cypress tests are not running on my side though. I guess it is because this task is not yet completed (https://coveord.atlassian.net/browse/SFINT-4327).

Exactly that's why, I'm currently working on that Jira, so the tests will be good to run soon 👌

lbergeron:
> > I didn't review all the changes, mostly did a quick check on the component itself. The Cypress tests are not running on my side though. I guess it is because this task is not yet completed (https://coveord.atlassian.net/browse/SFINT-4327).
> 
> Exactly that's why, I'm currently working on that Jira, so the tests will be good to run soon ok_hand

Perfect, thanks! I updated the ticket accordingly.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,273 @@
...
+  get errorMessage() {
+    return this._errorMessage;
+  }
+
+  /**
+   * Tells if there is more options to show in addition to the suggested options.
+   * @returns {boolean}
+   */
+  get isMoreOptionsVisible() {
+    return this.options.length > Math.max(Number(this.maxChoices), 1);
```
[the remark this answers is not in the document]
mmitiche: Actually, I don't want maxChoices to be equal to 0.  by a choice I mean any kind of option that we can propose to the user, if for example we display only the select dropdown  that's counted as 1 choice. if we display one suggestions and the select dropdown that's counted as 2 choices and so on . so for this reason MaxChoices can be only equal or greater than 1 cause we always at least want to show a choice to the user.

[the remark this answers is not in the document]
lbergeron: Ah OK I see. My understanding was that `maxChoices` was the maximum number of values displayed next to the dropdown.

nathanlb (commented):
I haven't reviewed the test suite yet but did a pass on the rest. This is nice work!

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,285 @@
...
+
+  /**
+   * The ID of the engine instance the component registers to.
+   * @api
+   * @type {string}
+   */
+  @api engineId;
+
+  /**
+   * The field of the case to be classified.
```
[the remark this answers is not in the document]
mmitiche: Nope, I just forgot to add them.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,285 @@
...
+    if (this._isSuggestionsVisible && this.isMoreOptionsVisible) {
+      this._hideSuggestions();
+    }
+  }
+
+  /**
+   * Hide the suggested options.
+   * @returns {void}
+   */
+  _hideSuggestions() {
```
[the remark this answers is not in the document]
mmitiche: I prefixed it that way only cause its a private method used only internally by this component. I think I will just remove it since we are not following such conventions.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,285 @@
...
+    const suggestions = this.template.querySelectorAll('.slds-visual-picker');
+    suggestions.forEach((suggestion) => {
+      // @ts-ignore
+      suggestion.style.width = `${suggestion.clientWidth}px`;
```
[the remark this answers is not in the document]
mmitiche: Yes It would be simpler to perform the hiding using the html conditional templates  and without the animation effect. The UX team preferred to have this animation  effect so that's why it's here,  is there any reason we want to avoid having animations effects? or is it a question of having this component consistant with the other quantic components, like I understand that having all the other Quantic components without any animation effects and this one containing one would be a bit weird.

packages/quantic/cypress/page-objects/search.ts
```diff
@@ -131,6 +139,12 @@ export function interceptSearchIndefinitely(): {sendResponse: () => void} {
   return interceptIndefinitely(routeMatchers.search);
 }
 
+export function interceptClassificationsIndefinitely(): {
```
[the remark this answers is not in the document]
mmitiche: Absolutely👌 I will separate all the new additions in a new file like you suggested.

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
```diff
@@ -0,0 +1,385 @@
...
+
+interface CaseClassificationOptions {
+  maxChoices: number;
+  label: string;
+  required: boolean;
+  selectPlaceholder: string;
+  messageWhenValueMissing: string;
+}
+
+describe('quantic-case-classification', () => {
```
erocheleau, line 21: It might be a stupid test to do, but does it make sense to test that the suggestions appear in the order that they were received from the API?

I only see tests that checks the correct number of suggestions are displayed for example. Would it make sense to also test in the order in which they get displayed?

packages/quantic/force-app/examples/main/lwc/exampleQuanticCaseClassification/exampleQuanticCaseClassification.js
```diff
@@ -0,0 +1,58 @@
+import {LightningElement, track, api} from 'lwc';
+
+export default class ExampleQuanticCaseClassification extends LightningElement {
+  @api engineId = 'case-assist-engine';
+  @api caseAssistId = 'a4fb453a-b1f1-4054-9067-bef117586baa';
```
erocheleau, line 5: What do we think of adding perhaps a utils function to `getDefaultCaseAssistId()` instead of hardcoding a string value directly here?

erocheleau: On second thought it might make sense to keep it there and contained.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.css
```diff
@@ -0,0 +1,37 @@
+lightning-icon{
+  --sds-c-icon-color-foreground-default: white;
+}
+.slds-visual-picker__figure{
+  width: 9.875rem;
```
erocheleau, line 5: Did you take these values somewhere from a component in the SLDS?

Seems weird to have such fractions of rem;

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.css
```diff
@@ -0,0 +1,37 @@
...
+  height: 1.938rem;
+}
+
+.loading-holder{
```
erocheleau, line 9: maybe linter needs to run on this file to have the `{` with a space after the class?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.html
```diff
@@ -0,0 +1,65 @@
...
+        <template if:true={loading}>
+          <div class="loading-holder slds-var-m-top_x-small">
+            <lightning-spinner alternative-text="Loading" size="small"></lightning-spinner>
+          </div>
+        </template>
+        <template if:false={loading}>
+          <template for:each={suggestions} for:item="option">
+            <div key={option.value}
+              class="case-classification-suggestion slds-visual-picker slds-visual-picker_medium slds-m-left_none slds-var-m-right_medium slds-var-m-top_x-small">
+              <input type="radio" data-suggestion-id={option.id} onchange={handleSelectClassification} id={option.value} value={option.value} name='options' />
```
erocheleau, line 17:
```suggestion
              <input type="radio" data-suggestion-id={option.id} onchange={handleSelectClassification} id={option.value} value={option.value} name='suggestions' />
```

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.html
```diff
@@ -0,0 +1,65 @@
...
+    </legend>
+    <div class="slds-form-element__control" aria-describedby="case-classification-error-message">
+      <div class="row">
+        <template if:true={loading}>
+          <div class="loading-holder slds-var-m-top_x-small">
+            <lightning-spinner alternative-text="Loading" size="small"></lightning-spinner>
+          </div>
+        </template>
+        <template if:false={loading}>
+          <template for:each={suggestions} for:item="option">
```
erocheleau, line 14:
```suggestion
          <template for:each={suggestions} for:item="suggestion">
```
Just to be consistent? Do you prefer that name?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,299 @@
...
+ *
+ * @category Case Assist
+ * @example
+ * <c-quantic-case-classification engine-id={engineId} field-name="Priority" required label="Which topic is related to your issue?" select-placeholder="More topics" max-choices="4" message-when-value-missing="Select an option"></c-quantic-case-classification>
+ */
+export default class QuanticCaseClassification extends LightningElement {
+  @wire(getObjectInfo, {objectApiName: CASE_OBJECT})
+  objectInfo;
+
+  @wire(getPicklistValuesByRecordType, {
```
erocheleau, line 29: This gets ALL values for ALL picklist fields?

Is there a way we only get the picklist values for only this specific field?

`fieldName`

erocheleau:
```js
@api(getPicklistValues, { recordTypeId: '$objectInfo.data.defaultRecordTypeId', fieldApiName: CASE_OBJECT[fieldName]})
    picklistValues;
```

I am really not sure if this works, but maybe it does? 😂

erocheleau: It's always been a bit fuzzy the order of execution of things if you modify a variable used in an @wire if the wire refreshes itself correctly

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,285 @@
...
+  /**
+   * The placeholder of the combo box input.
+   * @api
+   * @type {string}
+   * @defaultValue `'More Topics'`
+   */
+  @api selectPlaceholder = this.labels.moreTopics;
+
+  /**
+   * The maximum number of choices to be displayed, a choice can be a suggestion, an inline option or the select dropdown.
```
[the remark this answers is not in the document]
erocheleau: I would like to understand better this option too.
According to the tests, I think `1, 0 and -1` all have the same effect basically?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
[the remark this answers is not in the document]
erocheleau: I have the same understanding as Luc for this option, so perhaps a better name is required?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * The ID of the engine instance the component registers to.
+   * @api
+   * @type {string}
+   */
+  @api engineId;
+
+  /**
+   * The field of the case to be classified.
```
erocheleau, line 49: You will have to be precise here if the fieldName is the salesforce API name, a label, or the Coveo field name

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /**
+   * @param {CaseAssistEngine} engine
+   */
+  initialize = (engine) => {
+    this.engine = engine;
+    this.field = CoveoHeadlessCaseAssist.buildCaseField(engine, {
+      options: {
+        field: `sf${this.fieldName.toLowerCase()}`,
```
erocheleau, line 280: ouf, this is a discussion entirely...

I think it is wrong to assume that the field from the CaseAssist will always have this format. And as mentioned above on the `@api fieldName` property. You are playing with 3 different potential values here.

SF field API name => example `My_Custom_Field__c`

SF field label         => example `My Custom Field`

Coveo field name => example `sfmycustomfield` but can actually be ANYTHING the user wants. This is just the default value


I think I would do this:

```js
@api sfFieldAPIName;
@api coveoFieldName = `sf${this.sfFieldAPIName.toLowerCase()}`;
```

I am not even sure if we don't add a `c` at the end of the `sfmycustomfield` when it is a custom field in Salesforce
`sfmycustomfieldc`.

We used to by default, now I am not sure if it was removed.

erocheleau (changes requested):
We can discuss more the field comment in january

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
```diff
@@ -0,0 +1,385 @@
...
+    options: Partial<CaseClassificationOptions>
+  ) {
+    mockSfPicklistValues(defaultField, allOptions);
+    interceptSearch();
+    cy.visit(pageUrl);
+    configure(options);
+  }
+
+  describe('with default options', () => {
+    it('should work as expected', () => {
```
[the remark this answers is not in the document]
mmitiche: Absolutely👌

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
```diff
@@ -0,0 +1,385 @@
...
+          allOptions.slice(0, suggestionsCount)
+        );
+        fetchClassifications();
+        Expect.displaySelectTitle(false);
+        Expect.numberOfSuggestions(0);
+        Expect.numberOfInlineOptions(0);
+        Expect.displaySelectInput(true);
+      });
+
+      scope('when selecting an option from the select input', () => {
```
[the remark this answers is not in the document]
mmitiche: I just added it👌

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
mmitiche: Absolutely ! I will add such tests 👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.css
mmitiche: Nope, these values came from the design mockups, I used the same width used in the design and  converting it  from px to rem gave such fractions.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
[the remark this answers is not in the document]
mmitiche: Exactly, in other words the minimum value  that `maxChoices` can take is 1.  The logic behind this is that max choices represents the maximum number of choices we propose to the user, and a single choice can be a suggestion, an inline option or the select drop down. so for example if maxChoices is set to 3 we can show 2 suggestions and the select drop down. if maxChoices is set to 1 we only can show the select drop down. Having maxChoices set to 0 means that we don't want anything no suggestion and no inline option and no select drop down and this will make our component kind of useless. so maxChoices is always greater or equal to 1. and in the case where the user gives maxChoices a negative value or 0, we always reset the value to 1.

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
```diff
@@ -0,0 +1,385 @@
...
+        Expect.logUpdatedClassificationFromSelectOption(
+          defaultField,
+          clickedIndex
+        );
+      });
+    });
+  });
+
+  describe('with invalid number of maxChoices', () => {
+    it('should display only the select dropdown when maxChoices is set to 0', () => {
```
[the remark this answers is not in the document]
mmitiche: Yes
> Exactly, in other words the minimum value that `maxChoices` can take is 1. The logic behind this is that max choices represents the maximum number of choices we propose to the user, and a single choice can be a suggestion, an inline option or the select drop down. so for example if maxChoices is set to 3 we can show 2 suggestions and the select drop down. if maxChoices is set to 1 we only can show the select drop down. Having maxChoices set to 0 means that we don't want anything no suggestion and no inline option and no select drop down and this will make our component kind of useless. so maxChoices is always greater or equal to 1. and in the case where the user gives maxChoices a negative value or 0, we always reset the value to 1.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: sfFieldAPIName and coveoFieldName added 👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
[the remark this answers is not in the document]
lbergeron: `maxSuggestions` maybe? Setting it to zero would mean "no suggestions", and thus only the drop-down would be displayed.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,301 @@
...
+  get isMoreOptionsVisible() {
+    return this.options.length > Math.max(Number(this.maxChoices), 1);
+  }
+
+  /**
+   * Whether the select input is visible.
+   * @type {boolean}
+   */
+  get isSelectVisible() {
+    return this._isSelectVisible || 0 >= this.suggestions.length;
```
lbergeron, line 213: Is there a particular reason why you put the constant on the left side of the condition? It is generally easier to read when putting the variable expression on the left and the constant expression on the right.

```suggestion
    return this._isSelectVisible || this.suggestions.length < 0;
```

packages/quantic/quantic-examples-community/profiles/Quantic Examples Profile.profile
```diff
@@ -1,8 +1,483 @@
...
+        <readable>true</readable>
+    </fieldPermissions>
+    <fieldPermissions>
+        <editable>true</editable>
+        <field>Account.AccountSource</field>
+        <readable>true</readable>
+    </fieldPermissions>
+    <fieldPermissions>
+        <editable>true</editable>
+        <field>Account.AnnualRevenue</field>
```
lbergeron, line 104: You are granting way too much accesses through this profile. Please keep the permissions to the absolute minimum for this profile.

packages/quantic/force-app/examples/main/lwc/exampleQuanticCaseClassification/exampleQuanticCaseClassification.js
mmitiche: Is there an API call that I can make to retrieve the case assist config ID ?

nathanlb: [Here ](https://platformdev.cloud.coveo.com/docs/?urls.primaryName=Customer%20Service#/Configurations/getAllCaseAssistConfigurations)are all the case assist API calls. Yes there is one that lists all configs in an org. Keep in mind that the token needs certain privileges to successfully get this info. (It might already)

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.html
mmitiche: The suggestions and the inline options are both displayed using radio inputs and that's why I used the name value `options`, to have all the radio inputs with the same name. otherwise if the suggestions are named `suggestions` and the inline options are named `options` the user will have the possibility to choose more than one option.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Yes exactly, this will get the values  for ALL picklist fields. There is a way to to get only the picklist values for only one specific field but following the doc to achieve this import the field from Salesforce is needed, an import like the following:
`import PRIORITY_FIELD from '@salesforce/schema/Case.Priority';`
and the getting the piclist values like the following:
```
@wire(getPicklistValues, { recordTypeId: RECORD_ID, fieldApiName: PRIORITY_FIELD })
    propertyOrFunction;
```

But since in our case the field can be anything I can't do such import since I don't know the field that the user will want to classify, so I found this workaround of getting all the pickilist values and then using only the values of the field specified by the user.

packages/quantic/force-app/examples/main/lwc/exampleQuanticCaseClassification/exampleQuanticCaseClassification.js
mmitiche: Great! Thank you💪

packages/quantic/cypress/integration/case-classification/case-classification-actions.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  CaseClassificationSelector,
+  CaseClassificationSelectors,
+} from './case-classification-selectors';
+
+function caseClassificationActions(selector: CaseClassificationSelector) {
+  return {
+    clickSuggestion: (idx: number) => selector.suggestedOption(idx).click(),
```
aelfaleh, line 8: if you want you can add `logAction` to make your report more verbal

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,303 @@
...
+      ...CoveoHeadlessCaseAssist.loadCaseFieldActions(engine),
+    };
+  };
+
+  disconnectedCallback() {
+    this.unsubscribeField?.();
+  }
+
+  updateFieldState() {
+    if(this.maxChoices > 1){
```
nathanlb, line 149: space after the if

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,303 @@
...
+   * @api
+   * @type {string}
+   */
+  @api sfFieldApiName;
+    /**
+   * The name of the Coveo field to be classified.
+   * @api
+   * @type {string}
+   */
+    @api coveoFieldName;
```
nathanlb, line 61: rogue space here

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,303 @@
...
+import {
+  registerComponentForInit,
+  initializeWithHeadless,
+} from 'c/quanticHeadlessLoader';
+
+/** @typedef {import("coveo").CaseAssistEngine} CaseAssistEngine */
+/** @typedef {import("coveo").CaseField} CaseField */
+
+/**
+ * The `QuanticCaseClassification` component displays field value suggestions returned by Coveo Case Assist as well as a single-select dropdown containing other available values.
```
nathanlb, line 21: We should add another sentence saying these other values in the dropdown are taken from the org's CASE_OBJECT record type.

nathanlb (changes requested):
This was clearer a lot of work and it's looking real good! It works nicely and the test suite is awesome.
A few comments and some points i'll write here:
1. The mock-ups say the classification with the highest degree of confidence should be selected automatically. This isn't the case here. Was it discussed with UX?
2. The height of the buttons and the select input is different from the mock-ups as well. It's fine as long as it was discussed with UX and the mock-ups gets updated so that QA can validate efficiently.
3. This should be adjusted just a bit so the colors match and the overlay doesn't stick out.
Screen Shot 2022-01-04 at 3 15 29 PM

mmitiche:
> This was clearer a lot of work and it's looking real good! It works nicely and the test suite is awesome. A few comments and some points i'll write here:
> 
> 1. The mock-ups say the classification with the highest degree of confidence should be selected automatically. This isn't the case here. Was it discussed with UX?
> 2. The height of the buttons and the select input is different from the mock-ups as well. It's fine as long as it was discussed with UX and the mock-ups gets updated so that QA can validate efficiently.
> 3. This should be adjusted just a bit so the colors match and the overlay doesn't stick out.
> 
> Screen Shot 2022-01-04 at 3 15 29 PM

1. The feature of having the classification with highest degree of confidence is now added.
2.  Yes the hight of the buttons and the select input was discussed with the UX, I sync with them to update the mockups like you mentioned 👌.
3. The overlay is now adjusted, no more stick out and it's now with the correct color:

https://user-images.githubusercontent.com/86681870/148387852-d64b31a4-6d56-401c-a8e7-e4d9fab4e99a.mov

nathanlb (approved):
👏 👏 👏

lvu285:
Can we have an option to set number of Suggestion display?
for example, suggestion returns 4 values, and I want to display only first 2, the rest will be on the dropdown?

lvu285:
Like our discussion
- Probably we can consider about the `maxChoice` naming, a bit confusing
- According to first discussion during our meeting, we should support user `uncheck` the box and submit it --> option need to be supported from Quantic

*Also, consider to have default value for number of suggestion shows _ maybe 3 suggestions + 1 dropdown (like in the mockups)

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
```diff
@@ -0,0 +1,424 @@
...
+        Actions.clickSelectOption(clickedIndex);
+        Expect.logUpdatedClassificationFromSelectOption(
+          coveoDefaultField,
+          clickedIndex
+        );
+      });
+    });
+  });
+
+  describe('when invalid number of maxChoices is given', () => {
```
lvu285, line 200: would love to have a test to cover the situation when user set `invalid Salesforce field`

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -749,4 +749,32 @@
         <protected>false</protected>
         <shortDescription>Timeframe facet end label</shortDescription>
     </labels>
+    <labels>
```
mmitiche, line 752: @acarpentier01  please tell me if everything looks fine to you in the following labels.

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -400,4 +400,20 @@
         <label>Fin :</label>
         <name>quantic_EndLabel</name>
     </customLabels>
+    <customLabels>
```
mmitiche, line 403: @acarpentier01  french version of the labels.

aelfaleh (approved):

packages/quantic/force-app/examples/main/lwc/utils/utils.js
```diff
@@ -0,0 +1,24 @@
+// @ts-ignore
+import getHeadlessConfiguration from '@salesforce/apex/HeadlessController.getHeadlessConfiguration';
```
erocheleau, line 2: Is it really the best file to have this in?
At least perhaps a caseAssistUtils.js instead of a generic utils.js?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
erocheleau: `import PRIORITY_FIELD from '@salesforce/schema/Case.Priority';`
Inspect what the import gives you at runtime, I *think* it might just be the API name of the field.
It might just be that `PRIORITY_FIELD` is `Priority` as a string :P 

Long time since I checked.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,350 @@
...
+        ) ?? [];
+      this.loading = this.field.state.loading;
+      if (
+        JSON.stringify(this.classifications) !==
+        JSON.stringify(this.previousClassifications)
+      ) {
+        const value = this.classifications.length
+          ? this.classifications[0].value
+          : '';
+        this.setValue(value);
```
erocheleau, line 181: Perhaps a slightly more evocative name? 

`setFieldValue` ?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,350 @@
...
+      .filter(
+        (option) =>
+          !this.classifications.some((item) => item.value === option.value)
+      )
+      .map((option) => {
+        if (option.value === this._value) {
+          return {...option, checked: true};
+        }
+        return {...option, checked: false};
+      });
```
erocheleau, line 280:
```suggestion
      .map((option) => {
        return {...option, checked: (option.value === this._value)};
      });
```

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -0,0 +1,350 @@
...
+   * Returns the suggestions to display.
+   * @returns {Array<object>}
+   */
+  get suggestions() {
+    return this.classifications.map((suggestion) => {
+      if (suggestion.value === this._value) {
+        return {...suggestion, checked: true};
+      }
+      return {...suggestion, checked: false};
+    });
```
erocheleau, line 293:
```suggestion
    return this.classifications.map((suggestion) => {
      return {...suggestion, checked: (suggestion.value === this._value)};
    });
```

packages/quantic/quantic-examples-community/objects/Account.object
```diff
@@ -0,0 +1,395 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
erocheleau, line 1: This probably should not be in this PR or tracked?

packages/quantic/quantic-examples-community/objects/Case.object
```diff
@@ -0,0 +1,421 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
erocheleau, line 1: I don't think we should have that either, it's for giving access to the example community right?

Did you try with a permission set instead?

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
mmitiche: Absolutely! Thank you for your feedback👌

packages/quantic/quantic-examples-community/profiles/Quantic Examples Profile.profile
erocheleau: I think it's either we include it entirely or not at all?

Most of these are set to `false`

erocheleau (approved):
Just probably not necessary to include Account object or Case object in that?

packages/quantic/quantic-examples-community/objects/Case.object
erocheleau: What is the line exactly that grants the permissions? I see the permission on the profile you have included already, don't think we need the object on top of the profile

mmitiche: True! I just retried to setup the community after removing the objects and the community was created correctly with the setting we needed. I will just double check cause I remember trying this previously and it didn't give me the needed results without including the objects. when I confirm that it's working correctly I will just delete the files related to the objects and I'm gonna keep only the profile metadata👌

PR #1013, opened 2021-07-28
ci: fix package-lock files for snyk monitor

btaillon:
https://coveord.atlassian.net/browse/KIT-860

PR #1600, opened 2021-12-15
feat(atomic-react): PoC using atomic via react bindings

samisayegh:
I followed[ this stencil guide](https://stenciljs.com/docs/react) while creating the PR.

There is now an `atomic-react` dir in the monorepo. The structure of the dir was bootstrapped by cloning [this repo](https://github.com/ionic-team/stencil-ds-react-template). A lot of the code is generated by the bootstrap.

Running `build` in `atomic` outputs files to `atomic-react` via the `@stencil/react-output-target` plugin.

Also, here's a [page inside the headless-react](https://github.com/coveo/ui-kit/compare/KIT-1293?expand=1#diff-41973f18ee7553a80f1458bef3f8a0b0275a78a835625718a03d3a40a24ebe2aR1) to test the final output.


package.json
```diff
@@ -20,8 +20,7 @@
     "npm:tag:beta": "npm run npm:tag -- beta",
     "npm:tag:latest": "npm run npm:tag -- latest",
     "notify:docs": "node ./scripts/notify-docs/published-ui-kit.js",
-    "doc:generate": "lerna run doc:generate",
-    "lockLernaDependencies": "lerna exec \"npm install --package-lock-only\""
```
samisayegh, line 24: We have to rethink running this command on `setup` because it fails if a package is not yet published.

Context: https://github.com/coveo/ui-kit/pull/1013#discussion_r678541659

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <AtomicSearchInterface>
+      <AtomicSearchBox />
+      <AtomicResultList />
+    </AtomicSearchInterface>
+  );
+}
+
+async function init() {
+  await customElements.whenDefined('atomic-search-interface');
```
olamothe, line 22: We talked a bit about this this morning. Couple things: 

1- There would be a more idiomatic way to write this in React, with `useRef`, I think. https://reactjs.org/docs/hooks-reference.html#useref

2- In a ideal world, if we could end up with something similar to this:  

```js

const someArbitraryCodeToDetermineWhenToLaunchFirstQuery = ()=> {
 // [...]
}

return (
   <AtomicEngine.Provider provider={myEngine} >
      <AtomicSearchInterface onReady={someArbitraryCodeToDetermineWhenToLaunchFirstQuery} >
        <AtomicSearchBox />
        <AtomicResultList />
      </AtomicSearchInterface>

   </AtomicEngine.Provider>
  );
```

Which means that we would need to both give access to a context provider, and handle that in our react "AtomicSearchInterface", as well as augmenting it to expose a new `onReady` prop.

I'm not sure if that is feasible ? 

Maybe one approach would be to expose a different `AtomicSearchInterface` component in `headless-react` package, that simply wraps the internal "AtomicSearchInterface" that is actually built and shipped with Atomic ?

samisayegh: ✔️ for point 1

For point 2, this would be a nicer api. We could wrap AtomicSearchInterface to intercept the engine form the provider api and handle calling `initialize` (essentially, the code here would move into the wrapped component). I will try it out,

package.json
btaillon: We should check first if the `package-lock.json` files are fixed in the current or the latest version of lerna. If so, we no longer need `lockLernaDependencies`. If not, we must keep it to avoid pushing bad `package-lock.json` files or we must make the CI replace package-lock.json files.

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
btaillon: If we go with solution 2, I think it'd be a cherry on top if we had a `useEngine` instead of `useContext`.

packages/atomic-react/src/components/stencil-generated/react-component-lib/utils/attachProps.ts
```diff
@@ -0,0 +1,114 @@
...
+    eventName,
+    (eventStore[eventName] = function handler(e: Event) {
+      if (newEventHandler) {
+        newEventHandler.call(this, e);
+      }
+    })
+  );
+};
+
+const arrayToMap = (arr: string[] | DOMTokenList) => {
```
btaillon, line 110: Instead of a map, I think you may be looking for a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), which also only contains unique values.

packages/atomic-react/src/components/stencil-generated/react-component-lib/createComponent.tsx
```diff
@@ -0,0 +1,97 @@
...
+export interface HTMLStencilElement extends HTMLElement {
+  componentOnReady(): Promise<this>;
+}
+
+interface StencilReactInternalProps<ElementType> extends React.HTMLAttributes<ElementType> {
+  forwardedRef: React.RefObject<ElementType>;
+  ref?: React.Ref<any>;
+}
+
+export const createReactComponent = <
```
btaillon, line 20: there's a lot of potential here. if we can't wrap Stencil components, maybe calling its lifecycle events 1:1 from React's may work.

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
samisayegh: I encountered an error when using the wrapped component from the atomic-react package.
If the code is inside `headless-react`, it's no problem. Copying the code into `atomic-react` and importing it, I see,

invalid-hook-call

packages/atomic-react/src/components/stencil-generated/react-component-lib/utils/attachProps.ts
samisayegh: All code inside the `stencil-generated` dir is created by the stencil-react plugin. If we adjust, rebuilding atomic will revert the changes,

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <AtomicResultList />
+    </AtomicSearchInterfaceWrapper>
+  );
+}
+
+interface WrapperProps {
+  onReady?: () => void;
+}
+
+function AtomicSearchInterfaceWrapper(
```
samisayegh, line 21: I extracted this wrapper following the comments. However, if the wrapper is imported from `@coveo/atomic-react`, then [this project errors](https://github.com/coveo/ui-kit/pull/1600#discussion_r770809329). I haven't been able to figure out why so far,

packages/atomic-react/src/components/stencil-generated/react-component-lib/createComponent.tsx
samisayegh: Indeed, we can use the stencil code as inspiration, but may want to modify things. I did not find the vanilla stencil-react output a good experience, but I encourage you to try the branch out yourselves,

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ui-kit",
```
samisayegh, line 2: **Pros**
- Easy to setup.
- Works OOTB (except for result templates).
- Autocomplete for components, html attributes.
- Displays an error when a required attribute is missing.

**Cons**
- Initialization is non-idiomatic. Hiding it inside a wrapper gives an invalid hook error (to investigate).
- Custom result templates do not work (condition is using a custom `MapProp` decorator, which the stencil react plugin does not recognize)
- Component options are lost amid a sea of html attributes rather than being shown as props.
- Html attributes are not forwarded (e.g. specifying `placeholder` on `AtomicSearchBox` does not do anything). The autocomplete is therefore misleading.
- No access to the `controller` instance to render a different output.

**Overall impressions**
Working with Atomic components inside a ts instead of a html is a nicer dev experience. However, the components feel like they are not native React components. Certain functionality either does not work, or needs to be done in a non-idiomatic way.

olamothe:
FYI @samisayegh I did some improvements/troubleshooting with the "wrapping" of the search interface in another component.

It had to do with local symlink using two different versions of react.

TLDR: Need a "hack" when doing local dev with a manual symlink, but will work correctly when actually published on npm as a package.

PR #1601, opened 2021-12-15
feat(atomic): added slot inside of atomic-result-badge

btaillon:
https://coveord.atlassian.net/browse/KIT-1092

Allows for the following kind of syntax:
```html
<atomic-result-badge
  icon="https://raw.githubusercontent.com/Rush/Font-Awesome-SVG-PNG/master/black/svg/language.svg"
>
  <atomic-result-multi-value-text field="language"></atomic-result-multi-value-text>
</atomic-result-badge>
```

# Example
Badge with the language field:
image

Badge with a `atomic-result-multi-value-text` slotted element:
image

### DIFF
--- a/packages/atomic/src/components/result-template-components/atomic-result-badge/atomic-result-badge.tsx
+++ b/packages/atomic/src/components/result-template-components/atomic-result-badge/atomic-result-badge.tsx
@@ -3,11 +3,32 @@ import {Component, Element, Prop, h} from '@stencil/core';
 import {ResultContext} from '../result-template-decorators';
 
 /**
- * The `atomic-result-badge` element renders a badge containing a field.
- * @part result-badge-element - The badge element with the background color and text color.
- * @part result-badge-icon - The optional icon displayed in the badge element.
- * @part result-badge-label - The optional icon displayed in the badge element.
- * @slot default - Allow to display alternative content instead of a field or text.
+ * The `atomic-result-badge` element renders a badge to highlight special features of a result.
+ *
+ * A badge can either display:
+ * * Text:
+ * ```html
+ * <atomic-result-badge label="Trending"></atomic-result-badge>
+ * ```
+ * * The contents of a field:
+ * ```html
+ * <atomic-result-badge field="objecttype"></atomic-result-badge>
+ * ```
+ * * An icon alone:
+ * ```html
+ * <atomic-result-badge icon="https://my-website.fake/star.svg"></atomic-result-badge>
+ * ```
+ * * Slotted elements:
+ * ```html
+ * <atomic-result-badge icon="https://my-website.fake/stopwatch.svg">
+ *     Deal ends in <my-dynamic-countdown></my-dynamic-countdown>
+ * </atomic-result-badge>
+ * ```
+ *
+ * @part result-badge-element - The decorative outer-most element with the background color and text color.
+ * @part result-badge-icon - The icon displayed at the left-end of the badge, if present.
+ * @part result-badge-label - A wrapper around the contents at the right-end of the badge. This may be text, a field or slotted elements depending on which was configured.
+ * @slot default - The elements to display inside the badge, instead of a field or label.
  */
 @Component({
   tag: 'atomic-result-badge',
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/result-template-components/atomic-result-badge/atomic-result-badge.tsx
```diff
@@ -47,16 +48,18 @@ export class AtomicResultBadge {
     );
   }
 
+  private getTextContent() {
+    if (this.field !== undefined) {
+      return <atomic-result-text field={this.field}></atomic-result-text>;
+    }
+    if (this.label !== undefined) {
```
btaillon, line 55: Alternatively, we could specify `label` or the slot as a fallback for when the field is missing from a result.

packages/atomic/src/components/result-template-components/atomic-result-badge/atomic-result-badge.tsx
```diff
@@ -7,6 +7,7 @@ import {ResultContext} from '../result-template-decorators';
  * @part result-badge-element - The badge element with the background color and text color.
  * @part result-badge-icon - The optional icon displayed in the badge element.
  * @part result-badge-label - The optional icon displayed in the badge element.
+ * @slot default - Allow to display alternative content instead of a field or text.
```
npushkarskii, line 10: @slot default - Content/text that is to be shown instead of a field.

npushkarskii (changes requested):

packages/atomic/src/components/result-template-components/atomic-result-badge/atomic-result-badge.tsx
olamothe: I think it's better how it is right now versus too much hidden logic about what gets displayed.

olamothe (approved):
I think that's the kind of feature that is neat, but that we don't actually have any way to showcase.

If I am a user and f I just read: 
> * @slot default - Allow to display alternative content instead of a field or text.

I have no idea what that means, how I can use it, and that it's even a possibility for me to do what you explained in the PR... 🤷‍♂️

packages/atomic/src/components/result-template-components/atomic-result-badge/atomic-result-badge.tsx
```diff
@@ -47,16 +48,18 @@ export class AtomicResultBadge {
...
   }
 
+  private getTextContent() {
+    if (this.field !== undefined) {
+      return <atomic-result-text field={this.field}></atomic-result-text>;
+    }
+    if (this.label !== undefined) {
+      return <atomic-text value={this.label}></atomic-text>;
+    }
+    return <slot></slot>;
```
samisayegh, line 58: Should `<slot>` always be rendered, even if e.g. a `label` is specified? I feel this is the more natural behavior.

Also, should all components support a slot? Could this PR open the door to questions like "why am I able to slot here but not there"?

samisayegh (approved):

packages/atomic/src/components/result-template-components/atomic-result-badge/atomic-result-badge.tsx
btaillon: > Should `<slot>` always be rendered, even if e.g. a `label` is specified? I feel this is the more natural behavior?

Agreed

> Also, should all components support a slot? Could this PR open the door to questions like "why am I able to slot here but not there"?

Good point. However, I'm not sure if there's other components where it makes as much sense to add a slot to as badges, since badges are a very small and purely visual component. In a way, from an implementer perspective, badges wouldn't be so different from result sections.

btaillon:
> I think that's the kind of feature that is neat, but that we don't actually have any way to showcase.

I don't really intend to showcase this feature so much as implement a solution for a problem that I expect we'd receive maintenance cases for. Perhaps it's too early, but things such as a multi-value field or formatted numbers would be difficult to support without this feature.

Since this solution is still overall very simple, I prefer to have it than to recommend people make their own badges, which would be more of a hassle than I think is worth, and could introduce weird implementations that have a mix of our badges and their badges.

Another solution could be to remove badges altogether, but I think badges are something that a lot of people will need, and having them ready-made and perfectly compatible with our result sections improves the developer experience. Plus, they serve as a good reference as to what styling people should use for components in the actions section. I think badges also help push our product's design.

olamothe:
> > I think that's the kind of feature that is neat, but that we don't actually have any way to showcase.
> 
> I don't really intend to showcase this feature so much as implement a solution for a problem that I expect we'd receive maintenance cases for. Perhaps it's too early, but things such as a multi-value field or formatted numbers would be difficult to support without this feature.
> 
> Since this solution is still overall very simple, I prefer to have it than to recommend people make their own badges, which would be more of a hassle than I think is worth, and could introduce weird implementations that have a mix of our badges and their badges.
> 
> Another solution could be to remove badges altogether, but I think badges are something that a lot of people will need, and having them ready-made and perfectly compatible with our result sections improves the developer experience. Plus, they serve as a good reference as to what styling people should use for components in the actions section. I think badges also help push our product's design.

I'm not saying not to go ahead with this feature or PR, it is a good idea.

I just want to point out/give you an idea that the work you have done here is "incomplete", and to keep that in the back of your mind ;)

btaillon:
> I'm not saying not to go ahead with this feature or PR, it is a good idea.
> 
> I just want to point out/give you an idea that the work you have done here is "incomplete", and to keep that in the back of your mind ;)

I think I see what you mean. I'll try to put something more descriptive of the slot, and maybe even add an example, and I'll update the description of the component itself.

npushkarskii (approved):

PR #1606, opened 2021-12-17
ci: changed masterbot and publishbot condition to use commit message

btaillon:
https://coveord.atlassian.net/browse/KIT-1299

Ultimately, I believe this is what we're trying to accomplish with tags. Conditions on tags have been pretty unreliable, I believe this approach should work more reliably.

### DIFF
--- a/.github/workflows/masterbot.yml
+++ b/.github/workflows/masterbot.yml
@@ -3,11 +3,16 @@ on:
   push:
     branches:
       - master
-    tags-ignore:
-      - "**"
 jobs:
+  allow-workflow:
+    name: "Allow workflow"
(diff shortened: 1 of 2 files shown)


.github/workflows/masterbot.yml
```diff
@@ -3,11 +3,16 @@ on:
   push:
     branches:
       - master
-    tags-ignore:
-      - "**"
 jobs:
+  allow-workflow:
+    name: "Allow workflow"
+    if: "!startsWith(github.event.head_commit.message, '[Version Bump]')"
```
btaillon, line 9: Unfortunately, we can't put an `if` condition on an entire workflow, and there's no way to target commit messages otherwise.

samisayegh (approved):

PR #1608, opened 2021-12-17
ci: replaced ec2 runner with eks runner

btaillon:
https://coveord.atlassian.net/browse/KIT-1301

### DIFF
--- a/.github/actions/create-deployment-package/action.yml
+++ b/.github/actions/create-deployment-package/action.yml
@@ -3,7 +3,15 @@ description: "Create a deployment package for Headless and Atomic"
 runs:
   using: composite
   steps:
-    - id: parse-version
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+        cache: 'pip'
+    - name: Install deployment package
+      shell: bash
+      run: pip install --index-url https://pypi.dev.cloud.coveo.com/simple 'deployment-package==7.*'
+    - name: Parse Headless & Atomic version
+      id: parse-version
       uses: ./.github/actions/parse-version
     - name: Create & deploy package
       shell: bash
--- a/.github/workflows/publishbot.yml
+++ b/.github/workflows/publishbot.yml
@@ -10,16 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: echo "Workflow allowed to start"
-  build:
-    name: "Build"
-    needs: allow-workflow
-    runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


.github/workflows/publishbot.yml
```diff
@@ -10,16 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: echo "Workflow allowed to start"
-  build:
```
btaillon, line 13: Since the next jobs aren't run in parallel, there's no point building in advance.

.github/workflows/publishbot.yml
```diff
@@ -30,12 +23,7 @@ jobs:
   deploy:
     name: 'Deploy the package'
     needs: npm-publish
-    runs-on: [self-hosted, linux, x64, ec2-alpha]
-    container:
```
btaillon, line 34: It looks like `node` had trouble installing in the container, so I removed it. It's not necessary when installing the deployment package from `pip`.

.github/workflows/publishbot.yml
```diff
@@ -30,12 +23,7 @@ jobs:
...
-      image: 458176070654.dkr.ecr.us-east-1.amazonaws.com/jenkins/deployment_package:stable
-      options: --user root
-      volumes:
-        - /var/run/docker.sock:/var/run/docker.sock
+    runs-on: [self-hosted, linux, x64, eks-poc]
```
btaillon, line 26: The EKS runner is significantly better. Feel free to check the doc.

.github/actions/create-deployment-package/action.yml
```diff
@@ -3,7 +3,15 @@ description: "Create a deployment package for Headless and Atomic"
 runs:
   using: composite
   steps:
-    - id: parse-version
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
```
btaillon, line 8: The example I was given uses python 3.8. I was told that Python 3.9 and 3.10 should work too, but since the only thing we run in python is the deployment package, I don't see a reason to upgrade (yet).

.github/actions/create-deployment-package/action.yml
```diff
@@ -3,7 +3,15 @@ description: "Create a deployment package for Headless and Atomic"
...
+        cache: 'pip'
```
btaillon, line 9: The cache probably won't make a difference, but I see no harm in including it.

samisayegh (approved):

PR #1609, opened 2021-12-17
feat(headless): expose api client on engine

olamothe:
https://coveord.atlassian.net/browse/KIT-1279


packages/headless/src/app/case-assist-engine/case-assist-engine.ts
```diff
@@ -77,7 +77,11 @@ export function buildCaseAssistEngine(
     reducers: caseassistEngineReducers,
   };
 
-  const engine = buildEngine(augmentedOptions, thunkArguments);
+  const engine = buildEngine(
+    augmentedOptions,
+    thunkArguments,
+    caseAssistAPIClient
```
samisayegh, line 83: Why pass a third argument instead of accessing the client from `thunkArguments`?

samisayegh (approved):
Code looks good, but I think we are exposing too much for what is so far a one-off request from PS for a method that was not implemented.

I think asking the developer to implement the call is a reasonable compromise, to keep the api surface small and flexible for future maintainers of the headless project.

olamothe:
> Code looks good, but I think we are exposing too much for what is so far a one-off request from PS for a method that was not implemented.
> 
> I think asking the developer to implement the call is a reasonable compromise, to keep the api surface small and flexible for future maintainers of the headless project.

What do you mean exactly by exposing too much ? The list field values method ? Or something else.

packages/headless/src/api/search/search-api-params.ts
```diff
@@ -89,6 +89,27 @@ export interface TimezoneParam {
...
 
+export interface FieldParam {
+  field: string;
+}
+
+export interface MaximumNumberOfValuesParam {
+  maximumNumberOfValues: number;
+}
+
+export interface ContantQueryOverrideParam {
```
lbergeron, line 100:
```suggestion
export interface ConstantQueryOverrideParam {
```

btaillon (approved):

ThibodeauJF (approved):

samisayegh:
> What do you mean exactly by exposing too much ? The list field values method ? Or something else.

We're exposing the full api client. When headless moves to v4, v5 etc. the shape of the client will change (the methods, the parameters, and the response type). We have a larger api surface to maintain, and if we break, upgrading will be more work for consumers.

Also, ironically in this case, the person wanted an endpoint that headless did not need. We implemented a method that is not used in the project.

We're also setting expectations. All use-cases (e.g. case assist, product listing etc.) will have to expose their api to stay consistent.

> I'm not sure I understand why this is a one-off, I could see other people making custom components and needing to call the API.

Maybe this will happen, but one of the goals behind headless was to abstract the complexity interacting with the api directly. In the past year, only one person needed the api. Exposing it feels like it goes against the ethos of the project for what has so far been a one-off?

> So long as we don't expose too much client logic, this is kind-of a 1:1 of the API, I don't know if it's much to maintain. The way I see it, in theory, this could even have been a separate NPM package maintained by the API team.

Having a separate search-api SDK feels very reasonable. Yes the orgId and token will need to be specified again, but we're talking about power-users building niche UIs.

PR #1614, opened 2021-12-20
ci: added verbose logging to npm publish in GHA

btaillon:
https://coveord.atlassian.net/browse/KIT-1303

It should already be verbose since it was given the `--verbose` parameter, but I guess it's not since it's failing without any details.

### DIFF
--- a/.github/workflows/publishbot.yml
+++ b/.github/workflows/publishbot.yml
@@ -17,6 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/setup
+      - run: npm config set loglevel verbose --global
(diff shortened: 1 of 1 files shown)


olamothe (approved):
Isn't that going to clutter the output of everything in the build job though ?

btaillon:
> Isn't that going to clutter the output of everything in the build job though ?

`run` steps make their output collapsible, and seem to be collapsed by default. Here's what they look like expanded: 
image

btaillon:
> Isn't that going to clutter the output of everything in the build job though ?

Yes, but I don't think it's going to be that significant since every job is separated anyway and this job is only meant for publishing.

PR #1617, opened 2021-12-22
docs(atomic): add default fields to result list docs

dlutzCoveo:
These can be useful to have in docs, so I'm adding them. Figured the most appropriate place was in the fields-to-include description.


packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -70,7 +70,8 @@ export class AtomicResultList implements InitializableComponent {
    */
   private enableInfiniteScroll = false;
   /**
-   * A list of fields to include in the query results, separated by commas.
+   * A list of non-default fields to include in the query results, separated by commas.
+   * The default fields sent in a request are: 'date', 'author', 'source', 'language', 'filetype', 'parents', ‘urihash’, ‘objecttype’, ‘collection’, ‘permanentid’ 'ec_price', 'ec_name', 'ec_description', 'ec_brand', 'ec_category', 'ec_item_group_id', 'ec_shortdesc', 'ec_thumbnails', 'ec_images', 'ec_promo_price', 'ec_in_stock', 'ec_cogs', and 'ec_rating'.
```
btaillon, line 74: I think we should add an example here that shows what kind of value we expect.

Additionally, I would mention that those fields are always included, they aren't only default (maybe test it to double-check, but on my end it appeared to work as such).

btaillon (approved):

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
samisayegh: @btaillon, since Danny is no longer at Coveo, could you make the adjustments and merge this PR?

samisayegh (approved):

ThibodeauJF (approved):

PR #1618, opened 2021-12-23
fix(atomic): maintained keyboard focus on facets for accessibility

btaillon:
https://coveord.atlassian.net/browse/KIT-1075

In addition to E2E test that test focus, I added some other missing E2E tests.

I added a `FocusTarget` decorator, which allows us to define upcoming focus targets that may or may not exist yet. This utility should easily be reusable to improve accessibility in other components such as the pager. I recommend looking at [its usage](https://github.com/coveo/ui-kit/pull/1618/files#r774742155) before looking at its definition.

New behaviours:
* When pressing the "show less" button in any facet, the "show more" button is focused.
* When pressing the "show more" button in any facet, if there's no more "show more" button, the "show less" button is focused.
* When pressing the "clear filters" button in any facets, the header is focused.
* In a category facet, when pressing the "all categories" button, the header is focused.
* In a category facet, when selecting a facet search result, the active facet value is focused.
* ❗️The active category facet value is now displayed as a bold facet value (with the same indentation and font weight)
  * When pressed/clicked, it behaves exactly the same as the "All categories" button.
    * I did this because category facet values (and now also the active category facet value) are represented to screen readers as toggle switches. Logically, unselecting the currently active value should not activate another value.
  * When pressed, the header is focused.

Category facet comparison:
| before | after | after (while hovering active value) |
| --- | --- | --- |
| image | image | image |

@johnnyCoveo in your opinion, is the new look & behaviour of the active facet value okay?

### DIFF
--- a/packages/atomic/cypress/integration/common-assertions.ts
+++ b/packages/atomic/cypress/integration/common-assertions.ts
@@ -1,5 +1,4 @@
 import {TestFixture} from '../fixtures/test-fixture';
-import {AriaLiveSelectors} from './aria-live-selectors';
 import {ComponentErrorSelectors} from './component-error-selectors';
 
 export interface ComponentSelector {
--- a/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
@@ -255,7 +255,7 @@ describe('Category Facet Test Suites', () => {
         function setupSelectFirstParent() {
           setupGoDeeperLastLevel();
           cy.wait(TestFixture.interceptAliases.UA);
-          CategoryFacetSelectors.parentValue().first().click();
+          CategoryFacetSelectors.parentValue().first().focus().click();
           cy.wait(TestFixture.interceptAliases.Search);
         }
 
--- a/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
@@ -17,7 +17,10 @@ import {
   categoryFacetLabel,
 } from './category-facet-actions';
 import {TestFixture} from '../../../fixtures/test-fixture';
-import {typeFacetSearchQuery} from '../facet-common-actions';
+import {
+  pressShowMoreUntilImpossible,
+  typeFacetSearchQuery,
+} from '../facet-common-actions';
 
 describe('Category Facet Test Suites', () => {
   describe('with default settings', () => {
@@ -396,6 +399,27 @@ describe('Category Facet Test Suites', () => {
       describe('verify analytics', () => {
         before(setupShowMore);
         CategoryFacetAssertions.assertLogFacetShowMore();
+      });
+
+      describe('repeatedly until there\'s no more "Show more" button', () => {
+        function setupRepeatShowMore() {
+          setupWithDefaultSettings();
+          pressShowMoreUntilImpossible(CategoryFacetSelectors);
+        }
+
+        describe('verify rendering', () => {
+          before(setupRepeatShowMore);
+
+          CommonFacetAssertions.assertDisplayShowMoreButton(
+            CategoryFacetSelectors,
+            false
+          );
+          CommonFacetAssertions.assertDisplayShowLessButton(
+            CategoryFacetSelectors,
+            true
+          );
+          CommonFacetAssertions.assertFocusShowLess(CategoryFacetSelectors);
+        });
       });
 
       describe('when selecting the "Show less" button', () => {
--- a/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
@@ -255,7 +255,7 @@ describe('Category Facet Test Suites', () => {
         function setupSelectFirstParent() {
           setupGoDeeperLastLevel();
           cy.wait(TestFixture.interceptAliases.UA);
-          CategoryFacetSelectors.parentValue().first().focus().click();
+          CategoryFacetSelectors.parentValue().first().click();
           cy.wait(TestFixture.interceptAliases.Search);
         }
 
@@ -278,6 +278,11 @@ describe('Category Facet Test Suites', () => {
           );
           CategoryFacetAssertions.assertPathInBreadcrumb(selectedPath);
           CategoryFacetAssertions.assertPathInUrl(selectedPath);
+        });
+
+        describe('test accessibility', () => {
+          beforeEach(setupSelectFirstParent);
+
           CategoryFacetAssertions.assertFocusActiveParent();
         });
 
@@ -635,6 +640,7 @@ describe('Category Facet Test Suites', () => {
           CategoryFacetAssertions.assertNumberOfChildValues(1);
           CategoryFacetAssertions.assertNumberOfParentValues(2);
           CommonFacetAssertions.assertSearchInputEmpty(CategoryFacetSelectors);
+          CategoryFacetAssertions.assertFocusActiveParent();
         });
       });
 
--- a/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
@@ -14,6 +14,7 @@ import {
   addCategoryFacet,
   hierarchicalField,
   selectSearchResultAt,
+  categoryFacetLabel,
 } from './category-facet-actions';
 import {TestFixture} from '../../../fixtures/test-fixture';
 import {typeFacetSearchQuery} from '../facet-common-actions';
@@ -55,7 +56,7 @@ describe('Category Facet Test Suites', () => {
       );
       CommonFacetAssertions.assertLabelContains(
         CategoryFacetSelectors,
-        'Atlas'
+        categoryFacetLabel
       );
       CategoryFacetAssertions.assertValuesSortedByOccurences();
     });
@@ -92,6 +93,7 @@ describe('Category Facet Test Suites', () => {
         );
         CategoryFacetAssertions.assertPathInBreadcrumb(selectedPath);
         CategoryFacetAssertions.assertPathInUrl(selectedPath);
+        CategoryFacetAssertions.assertFocusActiveParent();
 
         describe('when collapsing the facet', () => {
           before(() => {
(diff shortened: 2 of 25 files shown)


packages/atomic/cypress/integration/common-assertions.ts
```diff
@@ -1,5 +1,4 @@
 import {TestFixture} from '../fixtures/test-fixture';
-import {AriaLiveSelectors} from './aria-live-selectors';
```
btaillon, line 2: This import was unused

packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
```diff
@@ -251,12 +278,93 @@ describe('Category Facet Test Suites', () => {
...
           CategoryFacetAssertions.assertPathInUrl(selectedPath);
+          CategoryFacetAssertions.assertFocusActiveParent();
         });
 
         describe('verify analytics', () => {
           before(setupSelectFirstParent);
           CategoryFacetAssertions.assertLogFacetSelect(selectedPath);
         });
+
+        describe('when selecting the label button to collapse', () => {
```
btaillon, line 289: Copied from other tests

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.pcss
```diff
@@ -2,17 +2,12 @@
 @import '../facet-common.pcss';
 @import '../facet-search/facet-search.pcss';
 
-.parent-button,
 .parent-active {
-  @apply w-full py-2.5 pr-2 pl-7 text-left;
+  @apply pl-7;
 }
```
btaillon, line 7: `.parent-active` is now a facet value just like the other ones, inheriting all their styling.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.pcss
```diff
@@ -2,17 +2,12 @@
...
 
 .parent-button {
-  @apply relative flex items-center;
-}
-
-.parent-active {
-  @apply truncate font-bold;
+  @apply w-full py-2.5 pr-2 pl-7 text-left relative flex items-center;
 }
```
btaillon, line 11: I tested, and the active value no longer needs this to be truncated.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -156,6 +160,15 @@ export class AtomicCategoryFacet
   @Prop() public injectionDepth = 1000;
   // @Prop() public customSort?: string; TODO: KIT-753 Add customSort option for facet
 
+  @FocusTarget()
+  protected showMoreFocus!: FocusTargetController;
```
btaillon, line 164: This is a good example of what `@FocusTarget` does.

When we press "show less", we know that we want to focus on the "show more" button after. However, if somebody pressed "show more" too many times and all children are shown, then the "show more" button doesn't exist until we press on "show less". Additionally, buttons will only be removed and added after a new search was rendered.

The decorator solves that issue by diving the whole process into two methods:
* setTarget
* focusAfterSearch

Calling `focusAfterSearch` activates a flag that indicates that we want to focus on the target as soon as:
* It's available
* A new search was performed

Calling `setTarget` will set the target to be focused. If the conditions for `focusAfterSearch` are met, it will also focus on the target immediately.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -156,6 +160,15 @@ export class AtomicCategoryFacet
...
+
+  @FocusTarget()
+  protected headerFocus!: FocusTargetController;
```
btaillon, line 167: Even for simpler use cases, such as focusing the header (which always exists), I think the utility can be helpful. I think we are better future-proofed in case some future change re-renders the header after a search is rendered.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -156,6 +160,15 @@ export class AtomicCategoryFacet
...
+
+  @FocusTarget()
+  protected activeValueFocus!: FocusTargetController;
```
btaillon, line 170: The `activeValueFocus` here is a more complex use case that the above two. Depending on the state after the search, we don't necessarily want to focus the same element. I recommend taking a look at where both of its methods are called.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -302,18 +331,35 @@ export class AtomicCategoryFacet
...
+          displayValue={activeParentDisplayValue}
+          numberOfResults={activeParent.numberOfResults}
+          isSelected={true}
+          i18n={this.bindings.i18n}
+          onClick={() => {
+            this.activeValueFocus.focusAfterSearch();
+            this.facet.deselectAll();
+          }}
+          searchQuery={this.facetState.facetSearch.query}
+          part="active-parent"
```
btaillon, line 354: This change is technically slightly breaking, since the part is no longer on a `li` element. I don't think it's significant though.

packages/atomic/src/components/facets/facet-show-more-less/facet-show-more-less.tsx
```diff
@@ -11,6 +11,8 @@ interface FacetShowMoreProps {
   canShowMoreValues: boolean;
   onShowMore(): void;
   onShowLess(): void;
+  showMoreRef?: (element?: HTMLButtonElement) => void;
+  showLessRef?: (element?: HTMLButtonElement) => void;
```
btaillon, line 15: We don't need `showLessRef`, but I think it would be weird to have only one of the two props.

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -17,7 +18,58 @@ export function AriaLiveRegion(regionName: string) {
...
     });
   };
 }
+
+export interface FocusTargetController {
+  setTarget(element: HTMLElement | undefined): void;
+  focusAfterSearch(): void;
+}
+
+export function FocusTarget() {
```
btaillon, line 31: The amount of issues and instabilities I had while developing this component is wild. I may be able to do some small code improvements to this decorator, but I had so many issues that I can't keep track of which was the reason for the various fixes/workarounds I implemented.

I think that having multiple use-cases for FocusTarget in this same PR really paid off, since lifecycle/ref functions were called differently across multiple use-cases.

Now, after all those fixes, the decorator seems very reliable and behaves predictably.

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -17,7 +18,58 @@ export function AriaLiveRegion(regionName: string) {
...
+  return (component: InitializableComponent, setterName: string) => {
+    const {componentWillLoad} = component;
+
+    component.componentWillLoad = function () {
```
btaillon, line 35: The decorator is called on the class itself, not an instance, so I used `componentWillLoad` to scope variables to each instance.

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -17,7 +18,58 @@ export function AriaLiveRegion(regionName: string) {
...
+      componentWillLoad && componentWillLoad.call(this);
+      const {componentDidRender} = this;
+      let focusAfterSearch = false;
+      let lastSearchId: string | undefined = undefined;
+      let element: HTMLElement | undefined = undefined;
+
+      function tryFocusElement(bindings: Bindings) {
```
btaillon, line 42: This function is called at two different places. It's called in `setTarget` (bound to the `ref`) and `componentDidRender`.

The reason for that, is that `setTarget` is only executed when one particular element is re-rendered, while `componentDidRender` is executed after every search (or other state change). If an element already exists and we're waiting for a new search, then this needs to be called from `componentDidRender`. Inversely, I don't remember why, but when a new component is rendered it looks like we have to focus it from `setTarget`/`ref`.

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -17,7 +18,58 @@ export function AriaLiveRegion(regionName: string) {
...
+        if (!element) {
+          return;
+        }
+        const searchId = bindings.engine.state.search.searchResponseId;
+        if (focusAfterSearch && searchId !== lastSearchId) {
+          setTimeout(() => element?.focus());
```
btaillon, line 48: This `setTimeout` is unfortunately necessary. I believe it's because some elements aren't appended to the DOM yet.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -156,6 +160,18 @@ export class AtomicCategoryFacet
   @Prop() public injectionDepth = 1000;
   // @Prop() public customSort?: string; TODO: KIT-753 Add customSort option for facet
 
+  @FocusTarget()
+  protected showMoreFocus!: FocusTargetController;
+
+  @FocusTarget()
+  protected showLessFocus!: FocusTargetController;
```
btaillon, line 167: This is a good example of what `@FocusTarget` does.

When we press "show more", it may or may not be removed. If it is removed, then we want to focus on the "show less" button. Otherwise, we want to leave the focus on the "show more" button. Additionally, these changes will occur only after a search has been completed, which may be multiple `componentDidRender` away.

The decorator solves that issue by diving the whole process into three methods:
* setTarget
* focusAfterSearch
* disableForCurrentSearch

Calling `focusAfterSearch` activates a flag that indicates that we want to focus on the target as soon as:
* It's available
* A new search was performed

Calling `setTarget` will set the target to be focused.

Calling `disableForCurrentSearch` indicates that we shouldn't focus on an element during the current search. In most cases, calling this function isn't necessary, but it's useful when we can't predict whether an element should be focused purely based on whether it's rendered or not.

samisayegh (approved):

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
btaillon: > Why opt for a decorator vs. a class or builder function? e.g.

Because we override some lifecycle methods. We can do it from a class or builder too, but I felt like that kind of behaviour is only expected from a decorator.

> Why use `protected` properties instead of `private` ones?

I don't remember why. I remember that I used to have a minor issue with the code editor that made `protected` more convenient, but switching it back to `private` right now makes no difference. I'll switch them to `private`.

packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
```diff
@@ -212,12 +216,36 @@ describe('Color Facet Test Suites', () => {
...
       before(setupSelectShowMore);
       FacetAssertions.assertLogFacetShowMore(colorFacetField);
     });
 
+    describe('repeatedly until there\'s no more "Show more" button', () => {
+      function setupRepeatShowMore() {
+        new TestFixture()
+          .with(addColorFacet({field: colorFacetField, label: colorFacetLabel}))
+          .init();
+        pressShowMoreUntilImpossible(ColorFacetSelectors);
```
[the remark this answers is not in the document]
btaillon: I was worried it may take a lot of time, but it's surprisingly fast.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -210,7 +226,16 @@ export class AtomicCategoryFacet
...
-        onClearFilters={() => this.facet.deselectAll()}
+        onClearFilters={() => {
+          this.headerFocus.focusAfterSearch();
+          this.facet.deselectAll();
+        }}
+        headerRef={(header) => {
+          this.headerFocus.setTarget(header);
+          if (!this.hasParents) {
+            this.activeValueFocus.setTarget(header);
+          }
```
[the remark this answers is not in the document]
btaillon: Unfortunately, `activeValue` or `header` in this situation may not already exist at the time that we want to call `focus`, and may appear on an upcoming `render` call. We could define some strings and call `focus` on those, but I feel like that would be more complicated than the current approach.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -156,6 +160,18 @@ export class AtomicCategoryFacet
   @Prop() public injectionDepth = 1000;
   // @Prop() public customSort?: string; TODO: KIT-753 Add customSort option for facet
 
+  @FocusTarget()
```
ThibodeauJF, line 163: +1 for decorators 😍

ThibodeauJF (approved):

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
ThibodeauJF: Actually haha this was important my bad!

PR #1621, opened 2022-01-03
ci: added registry to npm publish GHA step

btaillon:
https://coveord.atlassian.net/browse/KIT-1303

I think the publish step may be silently failing because the registry is missing. In the [official documentation](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages), the registry is added. In the old Jenkinsfile, the registry was also added.


ThibodeauJF (approved):
worth a try!

PR #1622, opened 2022-01-03
test(atomic): added accessibility tests for all result components

btaillon:
https://coveord.atlassian.net/browse/KIT-971

### DIFF
--- a/packages/atomic/cypress/integration/common-assertions.ts
+++ b/packages/atomic/cypress/integration/common-assertions.ts
@@ -10,9 +10,23 @@ export function should(should: boolean) {
   return should ? 'should' : 'should not';
 }
 
-export function assertAccessibility(component: string) {
+export function assertAccessibility(
+  component: string | (() => Cypress.Chainable<JQuery<HTMLElement>>)
+) {
+  const rulesToIgnore = ['landmark-one-main', 'page-has-heading-one', 'region'];
+  const rules = rulesToIgnore.reduce(
+    (obj, rule) => ({...obj, [rule]: {enabled: false}}),
+    {}
+  );
+
   it('should pass accessibility tests', () => {
-    cy.checkA11y(component);
+    if (typeof component === 'string') {
+      cy.checkA11y(component, {rules});
+    } else {
+      component().should(([el]) => {
+        cy.checkA11y(el, {rules});
+      });
+    }
   });
 }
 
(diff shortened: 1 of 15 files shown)


packages/atomic/cypress/integration/common-assertions.ts
```diff
@@ -10,9 +10,23 @@ export function should(should: boolean) {
   return should ? 'should' : 'should not';
 }
 
-export function assertAccessibility(component: string) {
+export function assertAccessibility(
+  component: string | (() => Cypress.Chainable<JQuery<HTMLElement>>)
+) {
+  const rulesToIgnore = ['landmark-one-main', 'page-has-heading-one', 'region'];
```
btaillon, line 16: These three tests are relevant for a whole page, not a single component.

When we test accessibility on a single element by specifying a string as a parameter, they appear to already be skipped, but when passing an element they seem to be tested for some reason.

packages/atomic/cypress/integration/common-assertions.ts
```diff
@@ -10,9 +10,23 @@ export function should(should: boolean) {
...
+  const rules = rulesToIgnore.reduce(
+    (obj, rule) => ({...obj, [rule]: {enabled: false}}),
+    {}
+  );
```
btaillon, line 20: we don't have `fromEntries` in cypress tests

packages/atomic/cypress/integration/result-list/result-components/result-icon-selectors.ts
```diff
@@ -5,6 +5,7 @@ export const resultIconComponent = 'atomic-result-icon';
 export const ResultIconSelectors = {
   shadow: () => cy.get(resultIconComponent).shadow(),
   firstInResult: () =>
-    ResultListSelectors.firstResult().find(resultIconComponent).shadow(),
```
btaillon, line 8: Some components had `.shadow()` on `firstInResult`, now none of them do.

packages/atomic/cypress/integration/result-list/result-list-utils.ts
```diff
@@ -5,36 +5,46 @@ import {
...
+}
+
 export function withAnySectionnableResultList(
   assertions: (
     display: ResultDisplayLayout,
     imageSize: ResultDisplayImageSize,
     density: ResultDisplayDensity
-  ) => void
+  ) => void,
+  options?: WithAnySectionnableResultListOptions
```
btaillon, line 21: I wanted to limit `it` blocks to only relevant ones, especially given that each `it` blocks count towards Cypress' limit.

ThibodeauJF (approved):

PR #1625, opened 2022-01-03
fix(atomic): improved labels for rating facet values

btaillon:
https://coveord.atlassian.net/browse/KIT-1082

Examples:
* Rating facet
  * Inclusion filter on 5 out of 5 stars; 6112 results
  * Inclusion filter on 4 out of 5 stars; 19182 results
  * Inclusion filter on 3 out of 5 stars; 3122 results
  * Inclusion filter on 2 out of 5 stars; 619 results
  * Inclusion filter on one out of 5 stars; 572 results
* Rating range facet
  * Inclusion filter on 5 out of 5 stars; 6112 results
  * Inclusion filter on 4 to 5 stars; 24869 results
  * Inclusion filter on 3 to 5 stars; 27714 results
  * Inclusion filter on 2 to 5 stars; 28265 results
  * Inclusion filter on 1 to 5 stars; 28776 results


packages/atomic/src/locales.json
```diff
@@ -1113,5 +1113,17 @@
   "query-suggestions-unavailable": {
     "en": "There are no search suggestions.",
     "fr": "Il n'y a pas de suggestions de recherche."
+  },
+  "stars": {
+    "en": "one out of {{max}} stars",
```
npushkarskii, line 1118:
```suggestion
    "en": "One out of {{max}} stars",
```

packages/atomic/src/locales.json
```diff
@@ -1113,5 +1113,17 @@
...
+    "fr": "une étoile sur {{max}}"
```
npushkarskii, line 1119:
```suggestion
    "fr": "Une étoile sur {{max}}"
```

npushkarskii (approved):
LGTM, just added two minor suggestions ✌

packages/atomic/src/locales.json
npushkarskii: (pointless suggestion)

packages/atomic/src/locales.json
npushkarskii: (pointless suggestion)

npushkarskii (approved):

ThibodeauJF:
It's not always stars, user can change it. Could we say something more generic, such as "rating, 5 out of 5" ?

ThibodeauJF (approved):

packages/atomic/src/locales.json
```diff
@@ -1113,5 +1113,17 @@
...
+  },
+  "stars_plural": {
```
samisayegh, line 1121: This works well for English and French, but some languages have multiple plural forms (a separate word for 1 vs. 2 vs. many).

Could using [i18next's plural](https://www.i18next.com/translation-function/plurals) features be helpful in this case?


Alternatively, could placing `étoile` at the end of the phrase like in the English example remove the need for a plural key?

samisayegh (approved):
Not related to this PR, but there was a maintenance case recently for the JSUI on the "inclusion filter" prefix for facet values. More info available under point 3 in [this jira comment](https://coveord.atlassian.net/browse/JSUI-3315?focusedCommentId=492023).

It seems like it's preferred to simplify to just "1 out of {{max}} stars". We can wait to ask Deque about it though,

packages/atomic/src/locales.json
btaillon: In this case, having star at the beginning sounds best. It should be easy in the future when we want to add a language with multiple plural forms.

PR #1629, opened 2022-01-04
doc(headless): add dateFilter & numericFilter React samples

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-793


packages/samples/headless-react/src/components/date-filter/date-filter.class.tsx
```diff
@@ -0,0 +1,93 @@
+import {Component, ContextType} from 'react';
```
samisayegh, line 1: We'll need to add entries for the date and numeric filter controllers in `packages\headless\doc-parser\use-cases\search.ts` and reference these files, and then they should appear in the generated docs 🤞

samisayegh (approved):

PR #1630, opened 2022-01-04
fix(atomic): allow date input for Safari versions under 14

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1112

Tested with BrowserStacks 
Screen Shot 2022-01-04 at 4 10 34 PM

valueAsDate isn't supported, and I added a placeholder to inform the user of the correct date input format

### DIFF
--- a/packages/atomic/src/components/facets/facet-date-input/facet-date-input.tsx
+++ b/packages/atomic/src/components/facets/facet-date-input/facet-date-input.tsx
@@ -70,6 +70,9 @@ export class FacetDateInput {
     const inputClasses = 'input-primary p-2.5';
     const labelClasses = 'text-neutral-dark self-center';
 
+    const placeholder = 'yyyy-mm-dd';
+    const pattern = '^[12]\\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\\d|3[01])$';
+
     return (
       <form
         class="grid gap-2 grid-cols-min-1fr mt-4 px-2"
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/facets/facet-date-input/facet-date-input.tsx
```diff
@@ -93,13 +93,14 @@ export class FacetDateInput {
...
           aria-label={startAria}
+          placeholder="yyyy-mm-dd"
           required
           // API/Index minimum supported date
           min={this.formattedDateValue('1401-01-01')}
           max={this.formattedDateValue(this.end)}
           value={this.formattedDateValue(this.filterState.range?.start)}
           onInput={(e) =>
-            (this.start = (e.target as HTMLInputElement).valueAsDate!)
+            (this.start = new Date((e.target as HTMLInputElement).value))
```
ThibodeauJF, line 103: dayjs could perhaps be used instead of new Date (if the format is correct, doesn't make much a difference)

samisayegh: What would happen if entering a different format? Would any code throw?

Also, are users alerted somehow that they are using an unsupported format? e.g. the apply button becomes disabled, error message etc.

samisayegh (approved):

packages/atomic/src/components/facets/facet-date-input/facet-date-input.tsx
ThibodeauJF: No code throes, simple a log in the console. I added a pattern which will indicate that the format is incorrect, it acts as a fallback. Otherwise we should stick to the Html validation that the browsers provides in most browsers

btaillon (approved):
I think this is fine. It's probably not accessible, but given that dates work on newer versions of Safari I'm not too concerned.

PR #1632, opened 2022-01-05
fix(atomic): fixed focus when pressing a pager page button

btaillon:
https://coveord.atlassian.net/browse/KIT-1078

If pages "1, 2, 3, 4, 5" are displayed, pressing page 5 shifts the displayed pages to "3, 4, 5, 6, 7", but the focus would stay on the last button (page 7), which is confusing when using a screen reader. With this change, the focus would be moved to page 5, so it stays on the same page they just pressed.

### DIFF
--- a/packages/atomic/cypress/integration/pager-assertions.ts
+++ b/packages/atomic/cypress/integration/pager-assertions.ts
@@ -1,0 +1,7 @@
+import {PagerSelectors} from './pager-selectors';
+
+export function assertFocusActivePage() {
+  it('should focus on the active page', () => {
+    PagerSelectors.buttonActivePage().should('be.focused');
+  });
+}
--- a/packages/atomic/src/components/atomic-pager/atomic-pager.tsx
+++ b/packages/atomic/src/components/atomic-pager/atomic-pager.tsx
@@ -15,6 +15,10 @@ import {
 } from '../../utils/initialization-utils';
 import ArrowRight from '../../images/arrow-right.svg';
 import {Button} from '../common/button';
+import {
+  FocusTarget,
+  FocusTargetController,
+} from '../../utils/accessibility-utils';
 
 /**
  * The `atomic-pager` provides buttons that allow the end user to navigate through the different result pages.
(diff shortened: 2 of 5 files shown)


packages/atomic/src/components/atomic-pager/atomic-pager.tsx
```diff
@@ -112,6 +119,10 @@ export class AtomicPager implements InitializableComponent {
 
   private buildPage(page: number) {
     const isSelected = this.pager.isCurrentPage(page);
+    const parts = ['page-button'];
+    if (isSelected) {
+      parts.push('active-page-button');
+    }
```
btaillon, line 125: `false` was being added as a part

ThibodeauJF: To be tested, but if you have 2 parts, I think selecting a single one won't work with a CSS selector with cypress test, is that true?

ThibodeauJF (approved):

samisayegh (approved):

packages/atomic/cypress/integration/pager-assertions.ts
```diff
@@ -0,0 +1,7 @@
+import {PagerSelectors} from './pager-selectors';
+
+export function assertFocusActivePage() {
```
olamothe, line 3: Why did we put this in a separate file, if it's only used in one place inside one test ?

packages/atomic/src/components/atomic-pager/atomic-pager.tsx
olamothe: If you do `atomic-pager[part*="active-page-button"]` it should be able to select it. 

Or do you mean something different @ThibodeauJF  ?

olamothe (approved):

packages/atomic/src/components/atomic-pager/atomic-pager.tsx
btaillon: Yeah, that's what I assumed. I just tested and we aren't able to query the element when there's two parts. One possible fix would be to introduce a `.part(...)` util for Cypress, but I don't think that's necessary yet.

packages/atomic/cypress/integration/pager-assertions.ts
btaillon: I've been trying to convert new code to the same pattern as facets, where assertions are in a separate file, it does look a bit ridiculous right now though since only one assertion was extracted. I'll extract some more.

btaillon: I've been trying to convert new code to use the same pattern as facets, where assertions are in a separate file. Since most of `pager.cypress.ts` uses the old pattern, it looks a bit weird.

PR #1635, opened 2022-01-05
fix(atomic): prevent ResizeObserver from crashing Safari 12

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1317

Prevent's crashing component and displaying an error to the console. It's quite an old version for the web 2018. The downside will be not adapting the breadcrumb/field list to be on one line.

Side note: css parts selectors don't work in Safari 12 😱


samisayegh (approved):

btaillon (approved):

olamothe:
I'm just wondering why we are testing on Safari 12 and trying to fix bugs.

If CSS selectors don't even work then it's kind of pointless, no ? We want to only support "latest" of all modern browser. And Safari 12 is not latest.

ThibodeauJF:
Yes that's why I didn't adjust to make it work flawlessly, just removed the error with a condition so there no crash in Safari 12.

PR #1637, opened 2022-01-05
fix(atomic): specify use of connectedCallback for resultContext method

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1277

### DIFF
--- a/packages/atomic/src/components/result-template-components/result-template-decorators.tsx
+++ b/packages/atomic/src/components/result-template-components/result-template-decorators.tsx
@@ -68,21 +68,20 @@ const resultContextEventName = 'atomic/resolveResult';
  *
  * This method is useful for building custom result template elements, see [Create a Result List](https://docs.coveo.com/en/atomic/latest/usage/create-a-result-list/) for more information.
  *
+ * You should use the method in the [connectedCallback lifecycle method](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks).
+ *
  * @param element The element that the event is dispatched to, which must be the child of a rendered "atomic-result".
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/result-template-components/result-template-decorators.tsx
```diff
@@ -68,21 +68,20 @@ const resultContextEventName = 'atomic/resolveResult';
  *
  * This method is useful for building custom result template elements, see [Create a Result List](https://docs.coveo.com/en/atomic/latest/usage/create-a-result-list/) for more information.
  *
+ * You should use the method in the [connectedCallback lifecycle method](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks).
```
jpmarceau, line 71:
```suggestion
 * You should use the method in the [`connectedCallback` lifecycle method](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks).
```

jpmarceau (approved):
👍

packages/atomic/src/pages/examples/custom.html
```diff
@@ -11,7 +11,7 @@
     <script>
       (async () => {
         await customElements.whenDefined('atomic-search-interface');
-        const searchInterface = document.querySelector('#search');
```
samisayegh, line 14: Is the `#search` id still needed? Or can it be removed from the html element?

samisayegh (approved):

npushkarskii (approved):

packages/atomic/src/pages/examples/custom.html
ThibodeauJF: Not needed & it wasn't on the html element 😢

PR #1639, opened 2022-01-06
fix(atomic): improved keyboard focus on search box

btaillon:
https://coveord.atlassian.net/browse/KIT-1079

### DIFF
--- a/packages/atomic/cypress/integration/search-box-assertions.ts
+++ b/packages/atomic/cypress/integration/search-box-assertions.ts
@@ -11,3 +11,9 @@ export function assertHasText(text: string) {
     SearchBoxSelectors.inputBox().should('have.value', text);
   });
 }
+
+export function assertHasSuggestionsCount(count: number) {
+  it(`should display ${count} suggestions`, () => {
+    SearchBoxSelectors.querySuggestions().should('have.length', count);
+  });
+}
--- a/packages/atomic/cypress/integration/search-box-assertions.ts
+++ b/packages/atomic/cypress/integration/search-box-assertions.ts
@@ -1,0 +1,13 @@
+import {SearchBoxSelectors} from './search-box-selectors';
+
+export function assertFocusSearchBox() {
+  it('should focus on the search box', () => {
+    SearchBoxSelectors.inputBox().should('be.focused');
+  });
+}
+
+export function assertHasText(text: string) {
+  it(`should contain "${text}"`, () => {
+    SearchBoxSelectors.inputBox().should('have.value', text);
+  });
+}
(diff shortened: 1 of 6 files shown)


packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -116,7 +123,7 @@ export class AtomicSearchBox {
       id: this.id,
       searchBoxController: this.searchBox,
       numberOfQueries: this.numberOfQueries,
-      inputRef: this.inputRef,
```
btaillon, line 119: It took me way too long to notice this, but the reason that focus wasn't coming back to the search box after selecting a suggestion was because it was being explicitly blurred. I removed the blur and introduced a `clearSuggestions` function as a replacement, since I'm guessing this was the intended behaviour.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -231,6 +238,7 @@ export class AtomicSearchBox {
   }
 
   private onInput(value: string) {
+    this.isExpanded = true;
```
btaillon, line 241: After submitting a query, the search box is no longer blurred, but suggestions are hidden. I find it a more natural behaviour if typing again shows new suggestions.

ThibodeauJF: I agree that's better! Small downsides is you have to type something or blur and come back to see suggestions again.

packages/atomic/cypress/integration/search-box-assertions.ts
```diff
@@ -0,0 +1,13 @@
+import {SearchBoxSelectors} from './search-box-selectors';
+
+export function assertFocusSearchBox() {
```
ThibodeauJF, line 3: Can we migrate more assertions, search box cypress tests are a bit dated

ThibodeauJF (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -307,7 +318,7 @@ export class AtomicSearchBox {
         class="h-full outline-none bg-transparent flex-grow px-4 py-3.5 text-neutral-dark placeholder-neutral-dark text-lg"
         value={this.searchBoxState.value}
         onFocus={() => this.onFocus()}
-        onBlur={() => this.onBlur()}
+        onBlur={() => this.clearSuggestions()}
```
samisayegh, line 321: On blur and refocus, would there now be a delay in seeing the suggestions while they are re-fetched?

samisayegh (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
btaillon: Yes, although it's not very noticeable even with a lot of throttling.

PR #1640, opened 2022-01-06
fix(atomic): fixed focus when clearing a breadcrumb

btaillon:
https://coveord.atlassian.net/browse/KIT-1084

Removing a breadcrumb selects the next breadcrumb, or the "clear all" button.

Focus is still lost when removing the last breadcrumb or pressing "clear all". Maintaining focus when that happens would require adding a lot of logic to find what's the next focusable element in the DOM (or adding a package such as [tabbable](https://www.npmjs.com/package/tabbable)). We'll see if Deque deems it important.

### DIFF
--- a/packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
+++ b/packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
@@ -71,7 +71,7 @@ export class AtomicBreadbox implements InitializableComponent {
   @State() private isCollapsed = true;
 
   @FocusTarget()
-  private breadcrumbRemovedFocus!: FocusTargetController;
+  private breadcrumbFocus!: FocusTargetController;
 
   public initialize() {
     this.breadcrumbManager = buildBreadcrumbManager(this.bindings.engine);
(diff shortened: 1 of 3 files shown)


packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -51,6 +55,8 @@ export class AtomicBreadbox implements InitializableComponent {
   private resizeObserver?: ResizeObserver;
   private showMore!: HTMLButtonElement;
   private showLess!: HTMLButtonElement;
+  private lastRemovedBreadcrumbIndex = 0;
+  private numberOfBreadcrumbs = 0;
```
btaillon, line 59: Neither of these values needs an `undefined` state nor needs to be reset, since they are only used after `focusAfterSearch` was called.

ThibodeauJF (approved):

packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -64,6 +70,9 @@ export class AtomicBreadbox implements InitializableComponent {
   @State() public error!: Error;
   @State() private isCollapsed = true;
 
+  @FocusTarget()
+  private breadcrumbRemovedFocus!: FocusTargetController;
```
samisayegh, line 74: Based off the name, this property sounds like a boolean? Could either `focusedBreadcrumb` or `lastFocusedBreadcrumb` be clearer names?

samisayegh (approved):

packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
btaillon: ah, I meant it like an event, like `onBreadcrumbRemoved`. I'll change it to `breadcrumbFocus`.

PR #1641, opened 2022-01-07
feat(quantic): Document Suggestion Component

mmitiche:
# Document Suggestion Component

### Demo:

https://user-images.githubusercontent.com/86681870/148548327-2c1f52aa-00ad-46b2-bf8d-4cb0559bba40.mov


packages/quantic/cypress/integration/document-suggestion/document.suggestion.expectations.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+
+    logClickingSuggestion: (index: number) => {
+      cy.wait(InterceptAliases.UA.SuggestionClick)
+        .then((interception) => {
+          const analyticsBody = interception.request.body;
+          selector
+            .accordionSection(index)
+            .invoke('attr', 'data-id')
+            .should('eq', analyticsBody.svc_action_data.suggestionId);
```
lvu285, line 38: is there anything else that we will log for UA and should be validated? like actionCause, or something else?

packages/quantic/cypress/integration/document-suggestion/document.suggestion.expectations.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+
+    logRatingSuggestion: (index: number) => {
+      cy.wait(InterceptAliases.UA.SuggestionRate)
+        .then((interception) => {
+          const analyticsBody = interception.request.body;
+          selector
+            .accordionSection(index)
+            .invoke('attr', 'data-id')
+            .should('eq', analyticsBody.svc_action_data.suggestionId);
```
lvu285, line 50: same question, is there anything else the system logged? and need to be validated?

packages/quantic/cypress/integration/document-suggestion/document-suggestion.cypress.ts
```diff
@@ -0,0 +1,195 @@
...
+    it('should render the component and all parts', () => {
+      mockDocumentSuggestion(allDocuments);
+      visitDocumentSuggestion();
+
+      scope('when loading the page', () => {
+        Expect.displayAccordion(true);
+        Expect.numberOfSuggestions(defaultMaxDocuments);
+      });
+
+      scope('when clicking on a document suggestion', () => {
```
lvu285, line 87: this test is to click ShowMore or click item title? _ probably we missed 1 of them

packages/quantic/cypress/integration/document-suggestion/document-suggestion.cypress.ts
```diff
@@ -0,0 +1,195 @@
...
+        const clickIndex = 1;
+
+        Actions.clickSuggestion(clickIndex);
+        Expect.logClickingSuggestion(clickIndex);
```
lvu285, line 91: I think probably we should add more `expect` after click on 1 item, like what it's going to show up ...

lvu285:
Small issue:
- Text alignment between title & body
- Is it `Helped 100 users...` another component? as I don't see it in the video
Question:
- What will happen when we click on the title of the item?  (on the main page, not in the quickview model)
- Can user custom number of suggested documents open on first load? I see Salesforce has `Accordion with Multiple Open Sections` ?

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -0,0 +1,149 @@
...
+    this.unsubscribeDocumentSuggestion = this.documentSuggestion.subscribe(() =>
+      this.updateDocumentSuggestionState()
+    );
+
+    this.actions = {
+      ...CoveoHeadlessCaseAssist.loadCaseAssistAnalyticsActions(engine),
+      ...CoveoHeadlessCaseAssist.loadDocumentSuggestionActions(engine),
+    };
+
+    engine.dispatch(this.actions.fetchDocumentSuggestions());
```
nathanlb, line 93: Maybe fetching on init should be optional?

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -0,0 +1,149 @@
...
+
+  connectedCallback() {
+    registerComponentForInit(this, this.engineId);
+    this.template.addEventListener('rating', this.onRating);
+  }
+
+  renderedCallback() {
+    initializeWithHeadless(this, this.engineId, this.initialize);
+    const slots = this.template.querySelectorAll('slot');
+    slots.forEach((slot) => {
```
nathanlb, line 65: If you put this into a well-named function it would be clearer what is happening.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.html
```diff
@@ -0,0 +1,41 @@
+<template>
+  <c-quantic-search-interface engine-id={searchEngineId}>
+    <template if:true={loading}>
+      <div class="loading-holder">
+        <lightning-spinner alternative-text={labels.loading}></lightning-spinner>
+      </div>
+    </template>
+    <template if:false={loading}>
+      <template if:true={hasSuggestions}>
```
nathanlb, line 9: While it wasn't defined in the mock-ups we should render something when no suggestions are returned. @adroletCoveo has been made aware of the situation and added a note in the [figma file](https://www.figma.com/file/mGZNYe9DbqIL1NO4m292PG/Case-Assist-User-experience?node-id=348%3A30935).

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
```
nathanlb, line 3: 53.0 :)

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -91,6 +91,17 @@ export default class QuanticResultQuickview extends LightningElement {
       // eslint-disable-next-line @lwc/lwc/no-inner-html
       this.contentContainer.innerHTML = this.state.content;
     }
+    const slots = this.template.querySelectorAll('slot');
+    slots.forEach((slot) => {
```
nathanlb, line 95: here as well. A named function would make it clearer what is happening with this manipulation.

packages/quantic/cypress/integration/document-suggestion/document-suggestion-actions.ts
```diff
@@ -0,0 +1,20 @@
...
+} from './document-suggestion-selectors';
+
+function documentSuggestionActions(selector: DocumentSuggestionSelector) {
+  return {
+    clickSuggestion: (idx: number) =>
+      selector
+        .sectionTitle(idx)
+        .click()
+        .logAction(
+          'When clicking on the document suggestion at the specified index'
```
nathanlb, line 13: Given you have the `idx` variable you could insert it in the logged text for that extra detail.

packages/quantic/cypress/integration/document-suggestion/document-suggestion.cypress.ts
```diff
@@ -0,0 +1,195 @@
...
+import {sendRating} from '../../page-objects/actions/action-send-rating';
+
+interface DocumentSuggestionOptions {
+  maxDocuments: number;
+}
+
+describe('quantic-document-suggestion', () => {
+  const pageUrl = 's/quantic-document-suggestion';
+
+  const allDocuments = [
```
nathanlb, line 19: There's currently an empty `cypress/fixtures` directory that could be used for large objects like this one.

nathanlb (commented):
Nice!

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -0,0 +1,149 @@
...
+
+  disconnectedCallback() {
+    this.unsubscribeDocumentSuggestion?.();
+    this.template.removeEventListener('rating', this.onRating);
+  }
+
+  updateDocumentSuggestionState() {
+    this.suggestions =
+      this.documentSuggestion.state.documents
+        .map((suggestion) => {
```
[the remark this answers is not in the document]
mmitiche: Cause I had to make the document suggestions returned by Headless Case assist exactly like the the document results returned by Headless search so it will be possible to feed them the the `quanticResultQuickView` and `raw` and `uri` were the missing fields.

packages/quantic/cypress/integration/document-suggestion/document.suggestion.expectations.ts
mmitiche: Actually nope, we have only the suggestions id and the actionCause is in the name of the event so that's all what we are logging for UA.

packages/quantic/cypress/integration/document-suggestion/document-suggestion.cypress.ts
mmitiche: This test is for the whole title including both the showMore and the item title.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
mmitiche: Absolutely !

packages/quantic/cypress/integration/document-suggestion/document.suggestion.expectations.ts
mmitiche: yes there is some additional information about the suggestion like the title and the URI and the position of this suggestion. Just added some tests to make sure we are sending everything correctly.

packages/quantic/cypress/integration/document-suggestion/document.suggestion.expectations.ts
mmitiche: Same here. I improved the tests to check for everything.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -0,0 +1,149 @@
...
+   */
+  @api searchEngineId;
+  /**
+   * The maximum number of document suggesions to display, it's a value between 1 and 5.
+   * @api
+   * @type {number}
+   */
+  @api maxDocuments = 5;
+
+  /** @type {Array<object>} */
```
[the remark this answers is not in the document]
mmitiche: I found type definition for the whole state but not for just the list of suggestions, I will double check to make sure.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.row {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+}
+
+.slds-card {
+  border: 1px solid #dddbda;
```
nathanlb, line 16: to try to avoid hard coded styling have you tried using some of the GA design tokens for border colors?

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.no-suggestion-message {
+  background-color: #f6f7f9;
```
nathanlb, line 20: Same here

packages/quantic/force-app/main/default/lwc/caseAssistUtils/caseAssistUtils.js
```diff
@@ -0,0 +1,24 @@
+// @ts-ignore
+import getHeadlessConfiguration from '@salesforce/apex/HeadlessController.getHeadlessConfiguration';
+
+export async function getCaseAssistId(configName) {
```
nathanlb, line 4: I'm not sure this should be in main/default as I don't think this should be packaged. This util was meant for testing purposes so we could manage configs without needing to make PRs if they change.
Why was it moved here?

nathanlb (commented):
Looks real good! Love the test setup.
In the mockups the suggestion title and body are aligned. Is this possible with minimal css adjustment?
Screen Shot 2022-01-14 at 11 46 41 AM

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -0,0 +1,181 @@
...
+
+  stopPropagation(evt) {
+    evt.stopPropagation();
+  }
+
+  injectIdToSlots() {
+    const slots = this.template.querySelectorAll('slot');
+    slots.forEach((slot) => {
+      let slotContent = slot;
+      while (slotContent?.tagName === 'SLOT') {
```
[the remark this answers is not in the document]
mmitiche: Sure! This manipulation is made to give the slots of this components access to the id of the different documents.
For example we have the slot "rating" made to show the rating of each document suggestion, it can be used this way: 
```
<c-quantic-document-suggestion>
    <c-my-custom-component-to-display-rating slot="rating"></c-my-custom-component-to-display-rating>
</c-quantic-document-suggestion>
```
The component passed as a slot here is meant to be used to display the rating for every document suggestion, the problem is that this component needs the id of the all the respective document suggestions to be able to fetch the data about the rating from somewhere. and since this component is passed as slot it has no access to this data, this data is present in quanticDocumentSuggestion component but not in the component passed as slot.

To surpass this, I'm adding this manipulation where from the quanticDocumentSuggestion I select the slots and I access the content of the slots using the method `assignedNodes` that returns a sequence of  the nodes assigned to this slot, I select the first one of this sequence which is the parent node in the slot content and I add to it a data attribute containing the data that the slot content will need to do the work correctly.
So that's what I'm doing here from the quanticDocumentSuggestion I'm adding a data attribute containing the id of the document suggestions corresponding to that slot. so if we continue with the code example I added above, the component myCustomComponentToDisplayRating will receive a data attribute containing the information needed which is the id of the document suggestions and this will allow the component to fetch the rating needed correctly.

packages/quantic/force-app/main/default/lwc/caseAssistUtils/caseAssistUtils.js
mmitiche: That was a mistake I will move this to be in the right location👌

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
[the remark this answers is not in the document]
nathanlb: Thanks for the detailed explanation! My question should have been a bit more precise because what I didn't really understand was the while loop. I see your iterating until the child is no longer a slot. Do you know if it's common for this loop to run more than once? Just curious at this point.
I guess another way of doing the same could be:
`const slotContent = slot.assignedNodes.find((node) => node.tagName !== 'SLOT')`

packages/quantic/force-app/main/default/lwc/caseAssistUtils/caseAssistUtils.js
nathanlb: All right thanks

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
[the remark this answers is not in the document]
mmitiche: The loop is needed when the slots passed to a component was also a slot passed 
The loop is needed for cases like this one
We have the slot "quickview-footer" which is used to pass a custom component that will be displayed in the footer of the quickview modal, it will be used this way:
```
<c-quantic-document-suggestion>
    <c-my-custom-footer slot="quickview-footer"></c-my-custom-footer>
</c-quantic-document-suggestion>
```
and inside quanticDocumentSuggestion we have something like this:
```
// quanticDocumentSuggestion.js
<template>
...
<c-quantic-result-quickview>
    <slot name="quickview-footer" slot="footer"></slot>
</c-quantic-result-quickview>
...
</template>
```
it's a bit confusing but here myCustomFooter is the component that needs the id information in his data attribute, so here the loop is needed because from quanticResultQuickview when we try access the content of slot "footer" to add the id data attribute we find ourselves access the slot of quanticDocumentSuggestion and here we need another loop to really reach the real custom component that was sent as a slot.
In other words the loop is made for the case when the slot of a component was also a slot coming from a higher component .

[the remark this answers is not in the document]
nathanlb: That makes sense to me. Thanks!

nathanlb:
> Looks real good! Love the test setup. In the mockups the suggestion title and body are aligned. Is this possible with minimal css adjustment? Screen Shot 2022-01-14 at 11 46 41 AM

nvm my adjustment thing i see it was mentioned before.

nathanlb (approved):

aelfaleh (approved):

louis-bompart:
GHA is having trouble, closing reopening hang tight.

PR #1644, opened 2022-01-07
test(atomic): attempt to improve breadbox test stability

btaillon:
https://coveord.atlassian.net/browse/KIT-1322

I don't have an exact idea as to what causes the stability issue of some tests such as Breadbox's. I logged the failed tests from various components in the JIRA above. In this PR, I'm mostly working towards my intuition and trying to solve only one test suite. If this PR gets approved, I will open a second PR to replicate those changes on other test suites.

# My intuition (for breadbox)
I believe the breadbox tests may be failing because of a race condition. I suspect that `cy.wait` on a search event may be calling the next step too quickly before components had the time to re-render, and some checkboxes may be ticked on elements that are being (or about to be) re-rendered.

Part of what's giving me this intuition is the following screenshot from a failed test in the CI, where two checkboxes should be checked but only the first one was actually checked:
<details>
<summary>screenshot</summary>

Breadbox Test Suites -- when selecting a standard facet and a numeric facet -- verify rendering -- should display the selected checkbox facets in the breadcrumbs (failed)
</details>

# My solutions
1. Instead of relying on `cy.wait` finishing after the render, I introduced a new function called `waitUntilNextSearchRendered`. This function looks at the current `searchResponseId`, creates a `MutationObserver`, and resolves after a mutation happened with a different `searchResponseId`.
2. Since analytics also relied on a `cy.wait`, which I doubt would be consistently called after a re-render,I introduced a `AnalyticsTracker` static class, which logs all analytics events and allows us to look at the most recent event. `TestFixture` takes care of resetting it and intercepting analytics events.

### DIFF
--- a/packages/atomic/cypress/integration/common-actions.ts
+++ b/packages/atomic/cypress/integration/common-actions.ts
@@ -1,0 +1,53 @@
+function observeRecursively(
+  element: Node,
+  callback: (mutations: MutationRecord[]) => void,
+  isAlreadyObserved = false
+): () => void {
+  const disconnectCallbacks: (() => void)[] = [];
+  if (!isAlreadyObserved) {
+    const observer = new MutationObserver((mutations) => {
+      mutations.forEach((mutation) => {
+        Array.from(mutation.addedNodes).forEach((node) => {
+          disconnectCallbacks.push(observeRecursively(node, callback, true));
+        });
+        callback(mutations);
+      });
+    });
+    observer.observe(element, {
+      subtree: true,
+      childList: true,
+      attributes: true,
+      characterData: true,
+    });
+    disconnectCallbacks.push(() => observer.disconnect());
+  }
+  if ((element as HTMLElement).shadowRoot) {
+    disconnectCallbacks.push(
+      observeRecursively((element as HTMLElement).shadowRoot!, callback, false)
+    );
+  }
+  Array.from(element.childNodes).forEach((child) =>
+    disconnectCallbacks.push(observeRecursively(child, callback, true))
+  );
+  return () => disconnectCallbacks.forEach((disconnect) => disconnect());
+}
+
+export function waitUntilNextSearchRendered() {
+  cy.wrap(
+    new Promise<void>((resolve) => {
+      cy.get('atomic-search-interface').then(([searchInterface]) => {
+        const lastSearchResponseId =
+          searchInterface.engine?.state.search.searchResponseId;
+        const disconnect = observeRecursively(searchInterface, () => {
+          if (
+            searchInterface.engine?.state.search.searchResponseId !==
+            lastSearchResponseId
+          ) {
+            disconnect();
+            resolve();
+          }
+        });
+      });
+    })
+  );
+}
--- a/packages/atomic/cypress/support/commands.ts
+++ b/packages/atomic/cypress/support/commands.ts
@@ -6,7 +6,7 @@ declare global {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     interface Chainable<Subject> {
       getAnalyticsAt(selector: string, order: number): Chainable<unknown>;
-      getTextOfAllElements(selector: string): Chainable<unknown>;
+      getTextOfAllElements(selector: string): Chainable<string[]>;
       // https://github.com/cypress-io/cypress-documentation/issues/108
       state(key: string): CypressRequest[];
       shouldBeCalled(urlPart: string, timesCalled: number): Chainable<unknown>;
--- a/packages/atomic/cypress/utils/analyticsUtils.ts
+++ b/packages/atomic/cypress/utils/analyticsUtils.ts
@@ -1,0 +1,87 @@
+import {CoveoAnalyticsClient} from '@coveo/headless/node_modules/coveo.analytics';
+import {findLast} from './arrayUtils';
+
+export type SearchEventRequest = Parameters<
+  CoveoAnalyticsClient['sendSearchEvent']
+>[0];
+
+export type ClickEventRequest = Parameters<
+  CoveoAnalyticsClient['sendClickEvent']
+>[0];
+
+export type CustomEventRequest = Parameters<
+  CoveoAnalyticsClient['sendCustomEvent']
+>[0];
+
+export type AnyEventRequest =
+  | SearchEventRequest
+  | ClickEventRequest
+  | CustomEventRequest;
+
+function isSearchEventRequest(
+  request: AnyEventRequest
+): request is SearchEventRequest {
+  return 'actionCause' in request;
+}
+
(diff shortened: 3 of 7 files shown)


packages/atomic/cypress/integration/common-actions.ts
```diff
@@ -0,0 +1,53 @@
+function observeRecursively(
+  element: Node,
+  callback: (mutations: MutationRecord[]) => void,
+  isAlreadyObserved = false
+): () => void {
```
btaillon, line 5: I'm honestly not sure how to simplify this. The problem is that `MutationObserver` doesn't observe the contents of shadowDOMs, so I made this function to recursively observe:
* Changes in the root element (on which this was first called)
* Changes in every shadowRoot (recursively)

packages/atomic/cypress/support/commands.ts
```diff
@@ -6,7 +6,7 @@ declare global {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     interface Chainable<Subject> {
       getAnalyticsAt(selector: string, order: number): Chainable<unknown>;
-      getTextOfAllElements(selector: string): Chainable<unknown>;
+      getTextOfAllElements(selector: string): Chainable<string[]>;
```
btaillon, line 9: There was linter errors due to this type being missing.

packages/atomic/cypress/utils/analyticsUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+  CoveoAnalyticsClient['sendSearchEvent']
+>[0];
+
+export type ClickEventRequest = Parameters<
+  CoveoAnalyticsClient['sendClickEvent']
+>[0];
+
+export type CustomEventRequest = Parameters<
+  CoveoAnalyticsClient['sendCustomEvent']
+>[0];
```
btaillon, line 14: I'm essentially just trying to import these types from `coveo.analytics`. importing from `@coveo/headless/node_modules/coveo.analytics/src/events` wasn't allowed by Cypress.

packages/atomic/cypress/utils/arrayUtils.ts
```diff
@@ -1,3 +1,12 @@
...
 }
+
+export function findLast<T>(values: T[], predicate: (value: T) => boolean) {
+  for (let i = values.length - 1; i >= 0; i--) {
+    if (predicate(values[i])) {
+      return values[i];
+    }
+  }
+  return null;
+}
```
btaillon, line 12: This is not as well typed as lodash's `findLast`, which detects type guards.

btaillon:
First run (successful): https://github.com/coveo/ui-kit/runs/4743259824?check_suite_focus=true

packages/atomic/cypress/integration/common-actions.ts
```diff
@@ -0,0 +1,53 @@
...
+      observeRecursively((element as HTMLElement).shadowRoot!, callback, false)
+    );
+  }
+  Array.from(element.childNodes).forEach((child) =>
+    disconnectCallbacks.push(observeRecursively(child, callback, true))
+  );
+  return () => disconnectCallbacks.forEach((disconnect) => disconnect());
+}
+
+export function waitUntilNextSearchRendered() {
```
samisayegh, line 35: Before going down this path, have we explored using cypress's [built-in retry-ability](https://docs.cypress.io/guides/core-concepts/retry-ability#Commands-vs-assertions) mechanism?

If we expect at a certain point in a test that there will be e.g. two active checkboxes, we can query for them and assert that it is the case.

```
cy.get('.active-checkbox-class')
    .should('have.length', 2)
```

If the assertion fails, cypress will retry until it reaches the timeout. If the assertion passes, then we can proceed with the next step in test.

Could adding these intermediate check-points can avoid the mutation observers?

packages/atomic/cypress/utils/analyticsUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+  return 'eventType' in request;
+}
+
+const analyticsEventsKey = '_analyticsEvents';
+
+interface WindowWithAnalytics extends Window {
+  [analyticsEventsKey]?: AnyEventRequest[];
+}
+
+export class AnalyticsTracker {
```
samisayegh, line 45: It would be ideal if we can continue using Cypress' alias features. If the root cause is the test being too fast for the UI, I'm hoping the check-point approach helps 🤞.

packages/atomic/cypress/integration/common-actions.ts
btaillon: Good point. I will try instead fixing errors on a case-by-case basis with assertions.

However, concerning retry-ability, we use a lot of `before()` calls, and [those cannot be retried](https://docs.cypress.io/guides/guides/test-retries#:~:text=However%2C%20failures%20in%20before%20and%20after%20hooks%20will%20not%20trigger%20a%20retry.).

packages/atomic/cypress/utils/analyticsUtils.ts
btaillon: The checkpoint may help. but I still see a lot of potential for race conditions whenever we wait both for a search and for an analytics call.

IMO, I would prefer to keep that utility. I don't like that some assertions have to wait for a analytics call, and I feel like using this utility would be more straightforward.

packages/atomic/cypress/utils/analyticsUtils.ts
olamothe: Could you import `coveo.analytics` as a direct dev dependencies in this project then ?

olamothe (commented):
If we compare to a few months prior, the Cypress tests were "fairly" stable, no ? Not perfect, but, relatively good.

Could it be the Github actions changes ? Or maybe cypress version ? Can we try to isolate a variable that way ?

I have a feeling that the changes in this PR are not the correct solution, since they add even more complexity instead of trying to leverage what Cypress should give us "for free".

After all, that's what Cypress exists for: Doing automatic retry for us.

btaillon:
> Could it be the Github actions changes ? Or maybe cypress version ? Can we try to isolate a variable that way ?

The test suites that fail always seem to be the same ones (see [JSUI-1322](https://coveord.atlassian.net/browse/KIT-1322)). I think improving those test suites in particular would almost completely resolve the issue.

> I have a feeling that the changes in this PR are not the correct solution, since they add even more complexity instead of trying to leverage what Cypress should give us "for free".
> 
> After all, that's what Cypress exists for: Doing automatic retry for us.

Agreed. I'll close this PR and instead focus on using `should` like Sami mentioned. For now, I also won't include the `AnalyticsTracker` util. If the tests are still flaky after these fixes, I'll make a separate PR where I try `AnalyticsTracker`.

PR #1646, opened 2022-01-10
fix(atomic): update Atomic output target for Stencil, remove exported decorators

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1258

Played a lot with a new Stencil project that imports Atomic and uses it’s tools to create new custom component, piggy backing on the internal decorators I’ve previously exported. I’ve ran into some issues with the idea, this PR removes some exports & fixes the Atomic-Stencil integration.

https://github.com/ThibodeauJF/stencil-app-atomic


packages/atomic/package.json
```diff
@@ -13,22 +13,24 @@
   "es2015": "dist/esm/index.mjs",
   "es2017": "dist/esm/index.mjs",
   "types": "dist/types/index.d.ts",
-  "collection": "dist/collection/collection-manifest.json",
-  "collection:main": "dist/collection/index.js",
+  "collection": "collection-manifest.json",
```
ThibodeauJF, line 16: Stencil is aware that Atomic is a Stencil project, and uses the “collection” output target. The manifest contains the list of Atomic components, which aren’t compiled into Web component. My own Stencil app will then automatically use it’s @stencil/core to compile the Atomic components and lazy load them in my page. This is not compatible with the CDN, and the compilation is invalid as some components don’t render correctly. We should not allow this, Atomic components should only and always be used as Web Component, and not Stencil Components. I haven’t found any specific config to change this, so I wrote my own custom manifest and deleted the generated output in the build to reduce the size. If I just delete the collection/collection:main inside the package.json, the @coveo/atomic import stops working inside my Stencil app.

packages/atomic/src/index.ts
```diff
@@ -6,12 +6,6 @@ export {
...
-  BindStateToController,
-  InitializeBindings,
-  InitializableComponent,
 } from './utils/initialization-utils';
 
-export {
-  ResultContext,
-  resultContext,
-} from './components/result-template-components/result-template-decorators';
+export {resultContext} from './components/result-template-components/result-template-decorators';
```
ThibodeauJF, line 11: I should’ve investigated this more before exporting the decorators. It kind of worked because of the collection output recompiling the Stencil output, but it threw a some errors. Now that the collection output is deleted, the decorators aren’t compatible. They are typescript decorators, compiled into js decorators, which can’t override the component methods such as connected callback. The utils method still work, bellow are linked examples of a custom component and a custom result template component. Not sure if exposing a typescript file with typescript decorators is a good idea, I haven’t many example of packages like this online. We might have to improve upon this later…

- https://github.com/ThibodeauJF/stencil-app-atomic/tree/master/src/components/custom-pager
- https://github.com/ThibodeauJF/stencil-app-atomic/blob/master/src/components/custom-title/custom-title.tsx

packages/atomic/package.json
```diff
@@ -56,6 +58,7 @@
     "i18next": "^20.2.2",
     "i18next-http-backend": "^1.2.2",
     "stencil-inline-svg": "^1.1.0",
+    "@stencil/store": "1.5.0",
```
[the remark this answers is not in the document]
ThibodeauJF: Tested all the bindings in the Stencil app and a non-Stencil app, works well, the store is fully usable

packages/atomic/package.json
btaillon: Why should they not be used as stencil components, is it because of the rendering issues?

samisayegh: Perhaps we should generate the [dist-custom-elements output](https://stenciljs.com/docs/custom-elements)?
There is [this thread comment](https://github.com/ionic-team/stencil/issues/2531#issuecomment-808082875) on Stencil's Github that might be relevant?

ThibodeauJF: There are rendering issues yes, and from our standpoint, I don't see why we would want to output/support Stencil components when it's supposed to simply be a toolkit that builds web component (with our Stencil.config with it's plugins), we would only want to output the resulting web components. From the client's standpoint, they will have a choice of using the CDN or bundling Atomic in their codebase instead of being forced to load and compile our stencil components

packages/atomic/src/index.ts
samisayegh: > Not sure if exposing a typescript file with typescript decorators is a good idea

Exporting only javascript is recommended because it avoids non-typescript consumers (e.g. babel, pure js etc.) needing to transpile our code. However, since decorators are not natively supported in js, then it seems like we have two options. Either:
- We state in our docs that transpiling with tsc or babel is needed.
- We offer a non-decorator approach.

packages/atomic/package.json
ThibodeauJF: @samisayegh it won't affect the collection output, different issue. I'm targeting the `dist` instead of the `collection` and it works fine.

That said, we should perhaps have the `dist-custom-elements` output too, Olivier did it here: https://github.com/coveo/ui-kit/pull/1645#discussion_r781329577

packages/atomic/src/index.ts
ThibodeauJF: In a Stencil dev project, we can expect typescript to be used. My current preferred solution is using the non decorator approaches (initializeBindings, resultContext methods) and in the CLI setup project, which is written in typescript, having an available decorator. I think it's important that the user can see and edit the decorator, as it has a lot of side effects.

olamothe: > and in the CLI setup project, which is written in typescript, having an available decorator.

How do we ensure that when we do a change in our "core" decorator, or that we add a new one, that new project thats gets generated with the CLI also benefit from them ?

And that an old project that want to update is able to get the new updated version of those ?

btaillon: > How do we ensure that when we do a change in our "core" decorator, or that we add a new one, that new project thats gets generated with the CLI also benefit from them ?

Perhaps the CLI should clone a package from our repo. That way, we can test it a bit and ensure that the CLI uses the correct version of Atomic for given files.

btaillon (approved):

packages/atomic/package.json
samisayegh: To target `dist` didn't you have to create an empty `manifest.json` file, adjust the paths in the `package.json`, delete the collection dir etc.? It feels like we're fighting with Stencil's tooling. If we generated `dist-custom-elements`, maybe things would just work?

ThibodeauJF: Generating `dist-custom-elements` was one of the first thing I tried (i'll try again to be 100% sure). The Stencil project will just automatically get whats in `collection`. If we don't set collection/collection:main in package.json, the Stencil app won't see @coveo/atomic as a valid module

packages/atomic/src/index.ts
ThibodeauJF: We use and expose methods like `resultContext` & `initializeBindings` (perhaps more coming). The customer using the CLI would just get a simple decorator using these methods. As long as the api of stencil/core & our methods are still valid, the decorators should always work. 

As to were it comes from, we could have the decorators in the atomic package directly, and it would be copied in the CLI generated project source code. This is important because it has to use their stencil/core instance. It's also better to have it in the source code so the user can see & edit the decorators to his needs.

packages/atomic/package.json
samisayegh: Ok, it sounds like Stencil hasn't considered a stencil project importing another stencil project then,

samisayegh (approved):

packages/atomic/package.json
ThibodeauJF: I managed for collection not to be generated instead of deleting it, here was the error when we remove the collection & collection:main from package json
Screen Shot 2022-01-11 at 6 11 12 PM

ThibodeauJF: > sounds like Stencil hasn't considered a stencil project importing another stencil project 

They did, it's the purpose of the collection output, but they haven't considered disabling it. We don't want the collection output for the reasons listed above

PR #1647, opened 2022-01-10
fix(atomic): replaced atomic-results-per-page button with radio

btaillon:
https://coveord.atlassian.net/browse/KIT-949

This change is meant to improve accessibility. The results-per-page component was already exposed as being a radio button, but did not support arrow keys unlike real radio buttons.

It looks exactly the same as before. I even switched tabs between an old version of the page vs the new one, and nothing changed.


packages/atomic/src/locales.json
```diff
@@ -226,10 +226,6 @@
     "zh-cn": "每一页的结果",
     "zh-tw": "每頁的結果數"
   },
-  "display-results-per-page": {
-    "en": "Display {{results}} results per page",
-    "fr": "Afficher {{results}} résultats par page"
-  },
```
btaillon, line 232: I think this is redundant, since the parent `radiogroup` already describes this as being "Results per page" and there's a label preceding the radiogroup.

ThibodeauJF (approved):

packages/atomic/src/components/atomic-results-per-page/atomic-results-per-page.tsx
```diff
@@ -33,6 +34,7 @@ export class AtomicResultsPerPage implements InitializableComponent {
   private resultPerPage!: ResultsPerPage;
   public searchStatus!: SearchStatus;
   private choices!: number[];
+  private readonly radioGroupName = randomID('atomic-results-per-page-');
```
samisayegh, line 37: With server-side rendering, the radio group name created on the server will differ from the one in browser. If so, could this break accessibility? Or maybe Stencil can recover and resolve the difference?

packages/atomic/src/components/atomic-results-per-page/atomic-results-per-page.tsx
```diff
@@ -86,23 +88,25 @@ export class AtomicResultsPerPage implements InitializableComponent {
...
-        ariaChecked={`${isSelected}`}
-        ariaLabel={this.bindings.i18n.t('display-results-per-page', {
-          results: choice,
-        })}
-        onClick={() => this.resultPerPage.set(choice)}
-        class={`btn-page ${isSelected ? 'selected' : ''}`}
-        part={`button${isSelected && ' active-button'}`}
-        text={choice.toLocaleString(this.bindings.i18n.language)}
-      ></Button>
+      <input
```
samisayegh, line 95: Should the `atomic-pager` use an `input` too to stay consistent?

samisayegh (approved):

packages/atomic/src/components/atomic-results-per-page/atomic-results-per-page.tsx
btaillon: I don't think so, the only reason I added a `name` is so that when we select a radio button, the others are deselected.

packages/atomic/src/components/atomic-results-per-page/atomic-results-per-page.tsx
btaillon: At first, I didn't think so since it has `aria-current`, but it looks like [aria-current is supported for `role="radio"`](https://www.w3.org/TR/wai-aria-1.1/#radio).

I'll create a new issue to make that change too.

olamothe: If so, I recommend extracting a shared functional component for this, similar to the `<Button>` you replaced here.

It can handle any accessibility concern, as well as UI (eg: ripple effect).

olamothe (approved):

PR #1649, opened 2022-01-10
feat(atomic): support result list templating for atomic-react 

olamothe:
I've debated about creating/overriding the AtomicResultList component, in order to create something that would not require `Template` and `ResultTemplate` component to be used in JSX, and might feel more "native" React.

After all, `Template` and `ResultTemplate` mostly exists for when you are dealing with HTML, and allows to have an `if / else if` pattern with simple HTML. This is not needed when you are dealing with JSX, and can simply use normal JavaScript conditions.

For example: 
```JSX
<AtomicResultList>
  <AtomicResultTemplate
    mustMatch={{
      objecttype: ['Account', 'File'],
    }}
  >
    <Template>
      <AtomicResultSectionBadges>
        <AtomicResultBadge field="source"></AtomicResultBadge>
      </AtomicResultSectionBadges>
      <AtomicResultSectionTitle>
        <AtomicResultLink></AtomicResultLink>
      </AtomicResultSectionTitle>
    </Template>
  </AtomicResultTemplate>
</AtomicResultList>
```

Could become: 
```JSX
const myResultTemplate = isObjectTypeAccountOrFile() ? myCarefullyDesignedTemplate() : myGenericTemplate();
return (
<AtomicResultList>
   {myResultTemplate}
</AtomicResultList>
)
```

However, I think in the long run this would have caused more problems than it solved, since it would means any behaviour/logic change we make in the core AtomicResultList component would have to be replicated in all output target.

It also means any example we will create in our documentation will be unique for each output target, each with their own specific recommended way to use them correctly.

So I opted with instead replicating as close as possible the experience you would have with using plain HTML, but with JSX.

This means exposing a component for `<template>` tag that uses `dangerouslySetInnerHTML`.  The explication is given here: https://stackoverflow.com/a/42845196.

We also need to wrap the `mustMatch` and `mustNotMatch` props, since those are not standard Stencil (and HTML) properties.

https://coveord.atlassian.net/browse/KIT-1325


btaillon (approved):
Agreed. If our framework didn't need to support the custom elements / DOM, then I definitely would've preferred not using `template` (and maybe even eliminating `atomic-result-template`).

packages/atomic-react/src/components/index.ts
```diff
@@ -1,6 +1,8 @@
 export * from './stencil-generated/index';
 export * from '@coveo/atomic/headless';
 
-// Important: Re-exporting under the same name "AtomicSearchInterface" shadows the original component
+// Important: Re-exporting under the same name (eg: "AtomicSearchInterface") shadows the original component
 // and should wrap it nicely for users of the library
 export {AtomicSearchInterfaceWrapper as AtomicSearchInterface} from './SearchInterfaceWrapper';
+export {ResultTemplateWrapper as AtomicResultTemplate} from './ResultTemplateWrapper';
+export {TemplateWrapper as Template} from './TemplateWrapper';
```
samisayegh, line 8: Could we combine these two components into one?

samisayegh (approved):
💯

packages/atomic-react/src/components/index.ts
olamothe: Yes we could have only the `ResultTemplateWrapper` automatically output the template tag with inner HTML.

I think that's a great idea.

ThibodeauJF (approved):

PR #1650, opened 2022-01-10
refactor(headless): remove dependency on #q property in querysuggest slice

samisayegh:
There's some code that's needed to ensure the `q` property in the query suggest slice stays in sync with the `q` in the query set. The duplication is not needed though and can be stream-lined.

This PR takes the first step. It removes the dependency on the `q` property and deprecates it. It does not remove the duplicate reducer code and unit tests though to not change the behavior in case anyone is reading the `q` property. We would do this just before headless v2 is released.

https://coveord.atlassian.net/browse/KIT-674

### DIFF
--- a/packages/headless/src/api/search/search-api-client.test.ts
+++ b/packages/headless/src/api/search/search-api-client.test.ts
@@ -241,7 +241,8 @@ describe('search api client', () => {
     it(`when calling SearchAPIClient.querySuggest
     should call PlatformClient.call with the right options`, async () => {
       const id = 'someid123';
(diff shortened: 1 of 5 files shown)


btaillon (approved):
Maybe we should make a "v2 cleanup" epic.

olamothe (approved):

samisayegh:
> Maybe we should make a "v2 cleanup" epic.

Yes, I think we could leave this for when we start the release?

ThibodeauJF (approved):

PR #1651, opened 2022-01-11
fix(atomic): improved accessibility for results

btaillon:
https://coveord.atlassian.net/browse/KIT-1091

# Changes
In this PR, overall I made the following changes:
* ~~Made grid results not be represented as singular links (by giving them a role other than link)~~
  * **EDIT**: this change was removed since Axe reports a [presentation-role-conflict](https://dequeuniversity.com/rules/axe/4.3/presentation-role-conflict?application=axeAPI) and a [aria-allowed-role](https://dequeuniversity.com/rules/axe/4.3/aria-allowed-role?application=axeAPI) warning.
* Made results more easy to interpret & navigate with screen readers by modifying their example template
  * If this PR is approved, I will apply the same modifications to all other sample/test pages we have in the Atomic package
  * The following changes were made:
    * Result sections were reordered to read out the most relevant information first
      * This only affects screen readers. Visually, results look the same
    * The image was hidden, since it provides no additional information
    * Titles were given a heading
      * This allows screen readers to jump between results
    * Roles were given to bottom-metadata to represent it as a grid
    * Badges were given the "note" role

As you can see, these changes put the responsibility of accessibility for results on customers. Perhaps it would be possible to force some of these changes so they aren't part of the result template, but given how subjective accessibility can be, I think this way is more convenient for us and for customers.

# Questions
Question for @npushkarskii: given that most changes need to be implemented by customers, we should document how to make results accessible. IMO, we should wait for Deque's assessment before making any documentation public. Should I write a draft somewhere?

Question for @johnnyCoveo: in which order do you think screen readers should navigate results? Should it be different for commerce and workplace? Keep in mind that screen readers can jump between the title of a result to the title of the next result. Currently, I chose the following order, inspired by Amazon's:
* Title
* Emphasized (price)
* Metadata (rating or printable uri)
* Badges
* Excerpt / description
* Bottom metadata / fields

# Affected issues
These changes also fix [KIT-1088](https://coveord.atlassian.net/browse/KIT-1088), [KIT-1089](https://coveord.atlassian.net/browse/KIT-1089), and [KIT-1148](https://coveord.atlassian.net/browse/KIT-1148).
These changes don't fix [KIT-1076](https://coveord.atlassian.net/browse/KIT-1076).


packages/atomic/src/pages/accessibility/commerce-full.html
```diff
@@ -385,45 +385,47 @@ <h1 class="accessibility-header">Results</h1>
...
+                <atomic-result-badge field="ec_brand" role="note"></atomic-result-badge>
+              </atomic-result-section-badges>
+              <atomic-result-section-visual>
+                <atomic-result-image field="ec_images" aria-hidden="true"></atomic-result-image>
+              </atomic-result-section-visual>
               <atomic-result-section-excerpt
                 ><atomic-result-text field="ec_shortdesc"></atomic-result-text
               ></atomic-result-section-excerpt>
-              <atomic-result-section-bottom-metadata>
+              <atomic-result-section-bottom-metadata role="grid">
```
samisayegh, line 409: Could the role `grid` be coded inside the section? Or would different clients choose different values depending on the structure of the template?

If different, what are the impacts of configuring an incorrect value? e.g. if this role was set to `row` instead of `grid`, what would happen?

samisayegh (approved):

packages/atomic/src/pages/accessibility/commerce-full.html
btaillon: > Could the role `grid` be coded inside the section? Or would different clients choose different values depending on the structure of the template?

It's possible, I'll move it. Clients are gonna be able to structure things however they want, with some exceptions (such as `atomic-result-fields-list` directly containing the children it can hide).

> If different, what are the impacts of configuring an incorrect value? e.g. if this role was set to `row` instead of `grid`, what would happen?

Screen readers are gonna read results wrong. They may:
* Describe elements inaccurately
* Describe individual parts of an element instead of the element as a whole
* Describe an element as a whole instead of its individual pieces
* Describe an element that should be skipped
* Skip an element that should be described
* Etc.

npushkarskii:
> Question for @npushkarskii: given that most changes need to be implemented by customers, we should document how to make results accessible. IMO, we should wait for Deque's assessment before making any documentation public. Should I write a draft somewhere?

Hey Benjamin, thx for a good question. I agree with your assumption, it's a good idea to have some draft in advance. 

You can do the following: 

1. Create a draft in a google doc. The doc can be either in your own Gdrive or some share drive, doesn't matter; the only thing that matters here is that other people should have view/suggesting access to it :)
2. Create a JIRA ticket and put the Gdoc link into it.

@btaillon

npushkarskii (approved):

olamothe (approved):
I really appreciate the pragmatic approach here of trying to do "as much as possible", while acknowledging the fact that we simply can't do magic either and always have the correct role "hardcoded" by us.

ThibodeauJF:
> Result sections were reordered to read out the most relevant information first

Is that a problem? That the visual (DOM) & audio (screen reading) order doesn't match? It might be confusing to some

ThibodeauJF (approved):

btaillon:
> Hey Benjamin, thx for a good question. I agree with your assumption, it's a good idea to have some draft in advance.
> 
> You can do the following:
>
> Create a draft in a google doc. The doc can be either in your own Gdrive or some share drive, doesn't matter; the only thing that matters here is that other people should have view/suggesting access to it :)
Create a JIRA ticket and put the Gdoc link into it.

Thanks! I'll do that.

btaillon:
> > Result sections were reordered to read out the most relevant information first
> 
> Is that a problem? That the visual (DOM) & audio (screen reading) order doesn't match? It might be confusing to some

I don't think so. The visual order already varies when we change image size, density, layout, viewport, etc.

PR #1652, opened 2022-01-11
fix(headless): add updateFacetAutoSelection & stop preventing auto select on new searches

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1333

### DIFF
--- a/packages/headless/src/controllers/search-box/headless-search-box.test.ts
+++ b/packages/headless/src/controllers/search-box/headless-search-box.test.ts
@@ -32,6 +32,7 @@ import {
   search,
 } from '../../app/reducers';
 import {deselectAllBreadcrumbs} from '../../features/breadcrumb/breadcrumb-actions';
+import {updateFacetAutoSelection} from '../../features/facets/generic/facet-actions';
 
 describe('headless searchBox', () => {
   const id = 'search-box-123';
@@ -237,6 +238,14 @@ describe('headless searchBox', () => {
       searchBox.submit();
 
       expect(engine.actions).toContainEqual(deselectAllBreadcrumbs());
+    });
+
+    it('it allows autoSelection', () => {
+      searchBox.submit();
+
+      expect(engine.actions).toContainEqual(
+        updateFacetAutoSelection({allow: true})
+      );
     });
 
     it('dispatches updateQuery with the correct parameters', () => {
--- a/packages/headless/src/controllers/search-box/headless-search-box.test.ts
+++ b/packages/headless/src/controllers/search-box/headless-search-box.test.ts
@@ -248,6 +248,20 @@ describe('headless searchBox', () => {
       );
     });
 
+    it('autoSelection should be allowed after deselecting facets', () => {
+      searchBox.submit();
+
+      const deselectAllBreadcrumbsIndex = engine.actions.findIndex(
+        (action) => action.type === deselectAllBreadcrumbs.type
+      );
+      const updateFacetAutoSelectionIndex = engine.actions.findIndex(
+        (action) => action.type === updateFacetAutoSelection.type
+      );
+      expect(deselectAllBreadcrumbsIndex).toBeLessThanOrEqual(
+        updateFacetAutoSelectionIndex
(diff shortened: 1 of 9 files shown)


packages/headless/src/features/facets/category-facet-set/category-facet-set-actions-loader.ts
```diff
@@ -71,6 +76,15 @@ export interface CategoryFacetSetActionCreators {
   updateCategoryFacetSortCriterion(
     payload: UpdateCategoryFacetSortCriterionActionCreatorPayload
   ): PayloadAction<UpdateCategoryFacetSortCriterionActionCreatorPayload>;
+
+  /**
+   * Updates the preventAutoSelect flag of all facets.
+   * @param payload - The action creator payload.
+   * @returns A dispatchable action.
+   */
+  updateFacetAutoSelection(
```
ThibodeauJF, line 85: Not sure were to export/load this, so I loaded it in the 2 facets that handle the action

btaillon (approved):

packages/headless/src/features/search/search-actions.ts
```diff
@@ -113,6 +114,7 @@ export const prepareForSearchWithQuery = createAsyncThunk<
   });
 
   dispatch(deselectAllBreadcrumbs());
+  dispatch(updateFacetAutoSelection({allow: true}));
```
olamothe, line 117: Is there any other place where we should be dispatching that action ? I think that's only triggered whenever there's a query submit from the search box.

I think it's fine by the way, just asking the question out loud to make sure we dont forget any interaction.

Another approach might have been to have a different action than `deselectAllBreadcrumbs` when we dispatch `prepareForSearchWithQuery` : An action that removes all breadcrumb/deselect all facets, but that do not set prevent auto select (eg: `deselectAllBreadcrumbOnSearchboxSubmit` or whatever...) .

It is a little bit weird to have one action `deselectAllBreadcrumbs` that removes facets selection AND set the auto select flag to true, then right after another action that set the auto select flag to false.

It's not the end of the world, though...

olamothe (approved):

packages/headless/src/features/search/search-actions.ts
ThibodeauJF: Yes we had the choice of making an action that is 90% similar to an existing one (with the flag being the only difference) or making a new action that can be used for possibly other use cases (we had nothing to allow or prevent autoselect directly).

samisayegh: Yeah, there are multiple approaches, each with different pros and cons.

**Current approach**

Pro: flexibility, the action can be composed elsewhere, clear when reading.
Con: the order of dispatch is important, gives direct access to the property

**deselectAllBreadcrumbOnSearchboxSubmit**

Pro: single action, order independent.
Con: less flexible, not clear what the difference(s) is(are) with `deselectAllBreadcrumbs` (higher abstraction could be seen as a pro though)

**deselectAllBreadcrumbs({allowFacetAutoSelect: true})**

Pro: single action, order independent.
Con: possible to get it wrong, when should I set the param to true? (ah ... before a search, didn't know)


The current approach feels like a good balance,
but there are good arguments for taking alternative paths too

packages/headless/src/controllers/search-box/headless-search-box.test.ts
```diff
@@ -239,6 +240,14 @@ describe('headless searchBox', () => {
       expect(engine.actions).toContainEqual(deselectAllBreadcrumbs());
     });
 
+    it('allows autoSelection', () => {
+      searchBox.submit();
+
+      expect(engine.actions).toContainEqual(
+        updateFacetAutoSelection({allow: true})
```
samisayegh, line 247: It would be good to assert the order of the actions, since someone inverting the order would introduce a regression.

e.g. index of `updateFacetAutoSelection` > index of `deselectAllBreadcrumbs` in `engine.actions`

samisayegh (approved):

PR #1653, opened 2022-01-11
ci(atomic): remove non-null assertion to try to improve cypress test stability

samisayegh:
One of the search-box cypress tests failed in a recent run due to calling `getAttribute` from a `null` object. I added a screenshot in the jira, and here are the relevant parts of the error message:

```
  > Cannot read properties of null (reading 'getAttribute')
...
Because this error occurred during a `before each` hook we are skipping the remaining tests in the current suite: `after focusing a suggestion...`
      at g.focusNextValue (http://localhost:3333/build/p-18fc00c4.entry.js:1:25092)
```

I removed the non-null assertions and adjusted the defensive conditions.

Build: https://github.com/coveo/ui-kit/runs/4781578067?check_suite_focus=true#:~:text=1)%20Search%20Box,the%20%60uncaught%3Aexception%60%20event.


packages/atomic/package.json
```diff
@@ -34,9 +34,9 @@
     "test": "stencil test --spec",
     "test:watch": "stencil test --spec --watchAll",
     "generate": "stencil generate",
-    "cypress:test": "./node_modules/cypress/bin/cypress run --record --key 0e9d8bcc-a33a-4562-8604-c04e7bed0c7e --browser chrome",
-    "cypress:test:firefox": "./node_modules/cypress/bin/cypress --record --key 0e9d8bcc-a33a-4562-8604-c04e7bed0c7e --browser firefox",
-    "cypress:open": "./node_modules/cypress/bin/cypress open",
+    "cypress:test": "cypress run --record --key 0e9d8bcc-a33a-4562-8604-c04e7bed0c7e --browser chrome",
+    "cypress:test:firefox": "cypress --record --key 0e9d8bcc-a33a-4562-8604-c04e7bed0c7e --browser firefox",
+    "cypress:open": "cypress open",
```
samisayegh, line 39: I adjusted these commands because they error on my machine otherwise,

current-cypress-command-error

btaillon (approved):

olamothe (approved):

packages/atomic/package.json
ThibodeauJF: Hmm the full path were used for a reason, perhaps jenkins. Now that we are on github, might not be a problem

ThibodeauJF (approved):

PR #1654, opened 2022-01-12
docs(atomic): add full fledge page for AtomicReact in samples

olamothe:
* Allows to showcase a "full" page which uses pretty much all components
* More importantly: Allows me to identify things that are not working :) I've created a couple user story to fix issues while creating this page.

image

https://coveord.atlassian.net/browse/KIT-1329


ThibodeauJF (approved):
Amazing 😍

packages/samples/headless-react/src/pages/AtomicReactPage.css
```diff
@@ -0,0 +1,168 @@
+atomic-color-facet[field='cat_color']::part(value-Black) {
```
samisayegh, line 1: If someone is looking to use Atomic React, would they need to copy this style sheet? Or is it just for our default layout and theme?

samisayegh (approved):

packages/samples/headless-react/src/pages/AtomicReactPage.css
olamothe: Well they would need it if they want to have this specific spacing/grids etc. Which is no different than the basic Atomic lib today.

If you remember there's a thing we want to do in 2022 for Atomic, which is to provide some "layouts" components, for predefined use cases.

So for examples, we would have "default commerce layouts" components, which would drastically remove the need for all this CSS if you want to have something that looks exactly like our default pages.

Does that make sense ?

samisayegh: Sounds good. In the short term, should we extract the layout styles into a CSS file and reference it inside Atomic and here so that the layouts stay in sync? or is it better to keep separate copies?

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
```diff
@@ -4,41 +4,207 @@ import {
...
+            expression="ec_price descending"
+          />
+        </AtomicSortDropdown>
+      </div>
+      <div className="results">
+        <AtomicDidYouMean />
+        <AtomicResultList
+          fields-to-include="ec_price,ec_rating,ec_images,ec_brand,cat_platform,cat_condition,cat_categories,cat_review_count,cat_color"
+          display="grid"
+          image-size="large"
```
btaillon, line 128: I just noticed that this is the only component that uses snake case for its props. Would it be possible to change it to camel case?

Additionally, would it be possible to make `fieldsToInclude` (and ResultsPerPage's choicesDisplayed) take arrays as a parameter?

btaillon (approved):

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
olamothe: Good catch, that was a copy paste mistake. It should be camelCase

As for transforming props that needs an array, from HTML based strings separated by comma, this would require extra preprocessing for every single props like this. And there's quite a lot of them.

It could be cool, but not absolutely mandatory for a first usable release

PR #1655, opened 2022-01-12
chore(linting): altered linting setup to include subpackages with different configs

nathanlb:
This PR aims to:
- Improve linting/prettier for Quantic package
- Allow for specific linting configs in separate packages of the monorepo
- Lint every subpackage's staged files on commit
  - In this PR it is currently set to simply `check` and not automatically `fix`. This can be changed to fit preference.

The gist:
- added linting scripts to package.json files
- .eslintrc converted from .json to .js
- base .eslintrc config defined in project root and extended in sub-packages
- base .prettierrc config defined in project root and extended in sub-packages.
- .prettierignore in subpackages (can't extend that one afaik)
- lint-staged installed and configured in sub-packages to run linting on staged files with sub-package configs
- root lint-staged calls `lerna lint-staged`

My goal was not to change the linting rules for packages other than Quantic.
If this setup is accepted I'll make another PR that lints the whole project so we start from a conforming baseline.


samisayegh (commented):
Would it have been possible to add linting config inside the `quantic` dir that extends and/or overrides the root configuration?

Duplicating certain .json files across all packages feels like it takes away from using a mono-repo?
- Adds overhead to setting up a new package?
- If we ever want to adjust a rule, we have to do it in multiple places?

Or was there a limitation when extending?

louis-bompart:
> Would it have been possible to add linting config inside the `quantic` dir that extends and/or overrides the root configuration?
> 
> Duplicating certain .json files across all packages feels like it takes away from using a mono-repo?
> 
> * Adds overhead to setting up a new package?
> * If we ever want to adjust a rule, we have to do it in multiple places?
> 
> Or was there a limitation when extending?

Maybe I'm not understanding your comment, but ain't that what it's done in this PR? 😕

package.json
```diff
@@ -8,8 +8,11 @@
     "build": "lerna run clean && lerna run build",
     "test": "lerna run --parallel test",
     "test:watch": "lerna run --parallel test:watch --no-prefix",
-    "lint:check": "eslint .",
-    "lint:fix": "eslint --fix .",
+    "lint:staged": "lerna run lint-staged && :",
```
louis-bompart, line 11: What is this `&& :` doing?

samisayegh:
The PR is duplicating config and npm scripts in the different packages atm rather than keeping things central.
See [here](https://github.com/coveo/ui-kit/pull/1655/files#diff-4a3afd444017091b3591979c58a34cab9a68c1499b67209913229d3be36df84fR37), [here](https://github.com/coveo/ui-kit/pull/1655/files#diff-02b0349bdb849c9d46d824831814e69bbd51314badca3c21afac76dcad8b8965R30), [here](https://github.com/coveo/ui-kit/pull/1655/files#diff-79fbe1d958aec3ec2b3d46cdc5b9b34cd834f48550895ce9484f10c0f77a4a8fR23) and so on.

If there quantic needs more control, can't the changes be scoped to that package? Why do all packages need to change?

nathanlb:
> The PR is duplicating config and npm scripts in the different packages atm rather than keeping things central. See [here](https://github.com/coveo/ui-kit/pull/1655/files#diff-4a3afd444017091b3591979c58a34cab9a68c1499b67209913229d3be36df84fR37), [here](https://github.com/coveo/ui-kit/pull/1655/files#diff-02b0349bdb849c9d46d824831814e69bbd51314badca3c21afac76dcad8b8965R30), [here](https://github.com/coveo/ui-kit/pull/1655/files#diff-79fbe1d958aec3ec2b3d46cdc5b9b34cd834f48550895ce9484f10c0f77a4a8fR23) and so on.
> 
> If quantic needs more control, can't the changes be scoped to that package? Why do all packages need to change?

currently the base shared config is defined at the root level and the packages just extend it. So new rules don't need to be replicated. The npm scripts are added everywhere to allow lerna to run them in every package.

Right now linting for quantic as part of a commit-hook isn't working, hence this broader change in how linting is setup. There are limitations that have led to this particular solution as lerna and eslint aren't best friends.

package.json
```diff
@@ -70,11 +73,8 @@
     ]
   },
   "lint-staged": {
-    "**/*.ts?(x)": [
-      "npm run lint:fix"
-    ],
-    "**/*.{scss,css,pcss,html,mdx}": [
-      "prettier --print-width 120 --write"
+    "*": [
+      "npm run lint:staged"
```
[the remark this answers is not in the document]
nathanlb: Yep good point

package.json
nathanlb: supposed to pass the long concatenation of staged file paths into the bash no-op command `:`

nathanlb:
I'm going to put this back into draft because it doesn't seem to be working like I remember and @samisayegh has made me wonder if I went down a rabbit hole based on an incorrect assessment.

PR #1657, opened 2022-01-12
test(atomic): replaced cy.wait with assertions in category-facet cypress tests

btaillon-coveo:
https://coveord.atlassian.net/browse/KIT-1322

If this gets approved, I will expand this fix to all other flaky specs.

# What I did
* Change A: I removed all `cy.wait(search)` and `cy.wait(analytics)` in favor of waiting for the render instead by using various assertions.
* Change B: Since the render may finish after analytics were already sent, I added the `AnalyticsTracker` static class, which stores all analytics sent since the last `TestFixture.init` call and allows us to assert on them.

# Effect on flakiness
* Analytics tests: After change A, all analytics tests started failing every single time. Since those tests were already flaky, I see the consistency as a win. Change B made these tests pass consistently.
* "expected to find element" tests: Since change A, I've never seen these errors again (despite having had to re-run tests **several** times to resolve other issues)
* focus tests: prior to change A, some of these tests consistently failed on the first run regardless of any other change I tried to implement. Since these tests, I've seen them fail occasionally on the first run, but rarely. They always succeeded on the second run so far.

### DIFF
--- a/packages/atomic/package.json
+++ b/packages/atomic/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coveo/atomic",
-  "version": "1.23.2",
+  "version": "1.23.3",
   "description": "A web-component library for building modern UIs interfacing with the Coveo platform",
   "homepage": "https://docs.coveo.com/en/atomic/latest/",
   "repository": {
--- a/packages/atomic-react/CHANGELOG.md
+++ b/packages/atomic-react/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.2.2](https://github.com/coveo/ui-kit/compare/@coveo/atomic-react@0.2.1...@coveo/atomic-react@0.2.2) (2022-01-13)
+
+**Note:** Version bump only for package @coveo/atomic-react
+
+
+
+
+
 ## [0.2.1](https://github.com/coveo/ui-kit/compare/@coveo/atomic-react@0.2.0...@coveo/atomic-react@0.2.1) (2022-01-13)
 
 **Note:** Version bump only for package @coveo/atomic-react
--- a/packages/atomic-react/package-lock.json
+++ b/packages/atomic-react/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@coveo/atomic-react",
-  "version": "0.2.1",
+  "version": "0.2.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/packages/atomic-react/package.json
+++ b/packages/atomic-react/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@coveo/atomic-react",
   "sideEffects": false,
-  "version": "0.2.1",
+  "version": "0.2.2",
   "description": "React specific wrapper for the Atomic component library",
   "repository": {
     "type": "git",
@@ -23,7 +23,7 @@
     "dist/"
   ],
   "dependencies": {
-    "@coveo/atomic": "^1.23.2"
+    "@coveo/atomic": "^1.23.3"
   },
   "devDependencies": {
     "@types/node": "^15.12.2",
--- a/packages/atomic/CHANGELOG.md
+++ b/packages/atomic/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.23.3](https://github.com/coveo/ui-kit/compare/@coveo/atomic@1.23.2...@coveo/atomic@1.23.3) (2022-01-13)
+
+**Note:** Version bump only for package @coveo/atomic
+
+
+
+
+
 ## [1.23.2](https://github.com/coveo/ui-kit/compare/@coveo/atomic@1.23.1...@coveo/atomic@1.23.2) (2022-01-13)
 
 
--- a/packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
@@ -233,7 +233,13 @@ describe('Color Facet Test Suites', () => {
     describe('repeatedly until there\'s no more "Show more" button', () => {
       function setupRepeatShowMore() {
         new TestFixture()
-          .with(addColorFacet({field: colorFacetField, label: colorFacetLabel}))
+          .with(
+            addColorFacet({
+              field: colorFacetField,
+              label: colorFacetLabel,
+              'number-of-values': 25,
+            })
+          )
           .init();
         pressShowMoreUntilImpossible(ColorFacetSelectors);
       }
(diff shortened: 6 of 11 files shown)


packages/atomic/cypress/integration/facets/category-facet/category-facet-actions.ts
```diff
@@ -33,9 +33,72 @@ export const addCategoryFacet =
...
 
 export function selectChildValueAt(index: number) {
-  CategoryFacetSelectors.childValue().eq(index).click();
+  CategoryFacetSelectors.childValue()
+    .find('[part="value-label"]')
+    .eq(index)
+    .then(([childValue]) => {
+      const text = childValue.textContent!;
+      cy.wrap(childValue).click();
+      CategoryFacetSelectors.activeParentValueWithText(text).should('exist');
```
btaillon, line 42: This line is what replaces `cy.wait`. I chose to put them in assertions instead of putting them in specs, because I think it'll make it much easier to add more tests and reduces risks of us forgetting a `should` somewhere.

packages/atomic/cypress/integration/facets/category-facet/category-facet-actions.ts
```diff
@@ -33,9 +33,72 @@ export const addCategoryFacet =
   };
 
 export function selectChildValueAt(index: number) {
-  CategoryFacetSelectors.childValue().eq(index).click();
+  CategoryFacetSelectors.childValue()
+    .find('[part="value-label"]')
+    .eq(index)
+    .then(([childValue]) => {
+      const text = childValue.textContent!;
+      cy.wrap(childValue).click();
```
btaillon, line 41: Doing `childValue.click` directly causes focus tests to fail. My guess is that `childValue.click` calls the DOM's native click function, while `cy.wrap(childValue).click()` calls Selenium's.

packages/atomic/cypress/integration/facets/category-facet/category-facet-assertions.ts
```diff
@@ -54,9 +53,7 @@ export function assertNoPathInUrl() {
 
 export function assertLogFacetSelect(path: string[]) {
   it('should log the facet selection to UA', () => {
-    cy.wait(RouteAlias.analytics).then((intercept) => {
-      const analyticsBody = intercept.request.body;
-      expect(analyticsBody).to.have.property('actionCause', 'facetSelect');
+    cy.expectSearchEvent('facetSelect').then((analyticsBody) => {
```
btaillon, line 56: I added `cy.expectSearchEvent` and `cy.expectClickEvent`, which is defined below in `commands.ts`.

packages/atomic/cypress/support/commands.ts
```diff
@@ -1,26 +1,39 @@
...
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace Cypress {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     interface Chainable<Subject> {
-      getAnalyticsAt(selector: string, order: number): Chainable<unknown>;
       getTextOfAllElements(selector: string): Chainable<unknown>;
       // https://github.com/cypress-io/cypress-documentation/issues/108
       state(key: string): CypressRequest[];
       shouldBeCalled(urlPart: string, timesCalled: number): Chainable<unknown>;
+      expectSearchEvent(actionCause: string): Chainable<SearchEventRequest>;
```
btaillon, line 15: Alternatively, we could take a `Partial<SearchEventRequest>` as a parameter to eliminate the need for `then`. I decided to keep it simple here.

packages/atomic/cypress/utils/analyticsUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class AnalyticsTracker {
+  private static get window() {
+    return window as WindowWithAnalytics;
+  }
+
+  private static get analytics() {
+    return (this.window[analyticsEventsKey] =
+      this.window[analyticsEventsKey] ?? []);
+  }
```
btaillon, line 45: I tried using a static `analytics: AnyEventRequest[]` property instead of saving it in the `window` object, but it looks like `commands.ts` is not using the same static instance of `AnalyticsTracker` as specs do.

packages/atomic/cypress/utils/analyticsUtils.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.window[analyticsEventsKey] = analytics;
+  }
+
+  static push(request: AnyEventRequest) {
+    this.analytics.push(request);
+  }
+
+  static reset() {
+    this.analytics = [];
+  }
```
btaillon, line 57: For now this is only used in `TestFixture`, but I think it could also be useful in some other cases (e.g., clicking the same button multiple times and checking that the last also sent analytics).

packages/atomic/cypress/integration/facets/category-facet/category-facet-assertions.ts
```diff
@@ -66,12 +63,15 @@ export function assertLogFacetSelect(path: string[]) {
         hierarchicalField
       );
 
-      expect(analyticsBody.facetState[0]).to.have.property('state', 'selected');
-      expect(analyticsBody.facetState[0]).to.have.property(
+      expect(analyticsBody.facetState?.[0]).to.have.property(
```
btaillon, line 66: A side effect of `analyticsBody` now being typed is that I needed to make small adjustments like this one.

btaillon:
Runs:
1. https://github.com/coveo/ui-kit/runs/4792792642?check_suite_focus=true

samisayegh:
Hopefully [this PR](https://github.com/coveo/ui-kit/pull/1659) addresses the breadbox error 🤞

packages/atomic/package.json
```diff
@@ -83,6 +83,7 @@
     "autoprefixer": "10.4.0",
     "axe-core": "4.3.5",
     "babel-loader": "8.2.3",
+    "coveo.analytics": "^2.18.53",
```
ThibodeauJF, line 86: How can we ensure this version and this version in Headless is the same? Can we reference the types from the node_modules instead?

ThibodeauJF (approved):

packages/atomic/package.json
btaillon: > How can we ensure this version and this version in Headless is the same?

Since we only need three simple types, this isn't strictly necessary.

> Can we reference the types from the node_modules instead?

If we do so, Cypress seems to prevent us from importing it from `node_modules/coveo.analytics/src`, so we have to extract the types using generic types like `Parameters<>`. I could try again though, since `src` seems to be working when installed directly and importing from this file.

packages/atomic/cypress/integration/facets/category-facet/category-facet-actions.ts
olamothe: Selenium -> Cypress.

But yes :) !

packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
```diff
@@ -224,13 +221,9 @@ describe('Category Facet Test Suites', () => {
       function setupGoDeeperLastLevel() {
         setupWithDefaultSettings();
         selectChildValueAt(canadaHierarchyIndex[0]);
-        cy.wait(TestFixture.interceptAliases.UA);
```
olamothe, line 227: Very nice that you have been able to remove all of those !

It means it's going in the right direction ! (Simpler tests).

olamothe (approved):

PR #1658, opened 2022-01-12
doc(quantic): set default value text on label prop

nathanlb:
The `label` property of the QuanticTimeframe was marked as required by our jsdoc parser because no `@defaultValue` tag was specified.


jpmarceau (approved):
👌

PR #1662, opened 2022-01-13
fix(atomic): improved accessibility for printable uri

btaillon:
https://coveord.atlassian.net/browse/KIT-1076

* Added a label to the printable URI as a whole
* Made the expand button focus on the first newly revealed value
* Added a role to separators

### DIFF
--- a/packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
+++ b/packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
@@ -13,6 +13,10 @@ import {
 import {Schema, NumberValue} from '@coveo/bueno';
 import Arrow from '../../../images/arrow-right.svg';
 import {LinkWithResultAnalytics} from '../../result-link/result-link';
+import {
+  FocusTarget,
+  FocusTargetController,
+} from '../../../utils/accessibility-utils';
 
 /**
  * The `atomic-result-printable-uri` component displays the URI, or path, to access a result.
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
```diff
@@ -50,6 +51,9 @@ export class AtomicResultPrintableUri {
    */
   @Prop() target = '_self';
 
+  @FocusTarget()
+  private expandedPartFocus!: FocusTargetController;
```
olamothe, line 55: Hum.... since the focus can only be in one place at a time in a search page, and that this controller allows to jump to many different place in the page (it's not only local navigation, correct me if I'm wrong), would it then make sense to start thinking this should instead be similar to our other bindings ?

Like i18n, or the engine, etc.
A single instance available to all components in the interface ?

olamothe (approved):

packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
btaillon: Intuitively yes, but this component generally (and especially in this case) waits for the re-render of the component it's attached to. Additionally, in some cases we need multiple instances of this controller to manage different focus "events" and their own expected targets.

samisayegh (approved):

ThibodeauJF (approved):

npushkarskii (approved):

PR #1666, opened 2022-01-14
docs(atomic): add doc for static assets and styling for atomic-react

olamothe:
The original goal of this story was to see if I could make styling result templates component easy with React.

As you might remember, style for result templates component in Atomic need to be injected with either inline style on each component, or with a `<style></style>` tag embedded inside the result list.

I've experimented with some approach, but they were either not very successful, or the end result ended up being more complicated than "doing nothing".

For example, I've thought about using CSS modules that the ResultTemplate React component could start accepting as a prop. This has the drawback of forcing our users to use CSS modules, and potentially install plugin for their module bundler to achieve this. 
I also discovered that, in Webpack at least, the CSS module loader is not a straight up "accept any CSS and output as is", and would not correctly recognize rule such as `atomic-result-badge::part(result-badge-element)`.

I also started playing with https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet , but at the end of the day, you still end up playing with strings to do your styling.

And so, I think that by just recommending a couple different options on how to achieve styling for result templates, that it would be just as good and the least maintenance effort on our part.

Obviously, this will eventually go under any tutorial/documentation we will create on docs.coveo.com.

https://coveord.atlassian.net/browse/KIT-1331


ThibodeauJF (approved):

btaillon (approved):

packages/atomic-react/README.md
```diff
@@ -1,16 +1,104 @@
...
+`npm i @coveo/atomic-react`
 
-- Build your core stencil package.
+## Usage
 
-## Step 3.
+Since Atomic React is built on top of the core [Atomic](https://docs.coveo.com/en/atomic/latest/) web-components library, the vast majority of concept that applies to core Atomic will apply to Atomic React.
 
-- Run build on this package.
+However, there are still some special consideration.
```
jpmarceau, line 15:
```suggestion
However, there are still some special considerations.
```

packages/atomic-react/README.md
```diff
@@ -1,16 +1,104 @@
...
+```
+cp -r node_modules/@coveo/atomic-react/dist/assets public/assets
+cp -r node_modules/@coveo/atomic-react/dist/lang public/lang
+```
+
+It is important to respect the folder hierarchy, with SVG icons under the `assets` subdirectory, and labels and languages under the `lang` subdirectory of the public folder.
+
+## Styling Result Template Components
+
+Due to the way Atomic Web components use [Shadow Dom](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) and [CSS parts](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) to provide encapsulation, it is necessary to follow these guidelines when you wish to style elements inside any result templates.
```
jpmarceau, line 36:
```suggestion
Due to the way Atomic Web components use [Shadow Dom](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) and [CSS parts](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) to provide encapsulation, it is necessary to follow these guidelines when you wish to style elements inside any result template.
```

jpmarceau (approved):
👍

samisayegh (approved):

npushkarskii (approved):

PR #1667, opened 2022-01-14
feat(headless): Add FrequentlyViewedSameCategory and FrequentlyViewedDifferentCategory strategies

thudon-coveo:
Adding to new strategies for viewing products frequently viewed in the same and different categories.

### DIFF
--- a/packages/headless/src/controllers/product-recommendations/headless-frequently-viewed-different-category.ts
+++ b/packages/headless/src/controllers/product-recommendations/headless-frequently-viewed-different-category.ts
@@ -1,0 +1,44 @@
+import {Schema, SchemaValues} from '@coveo/bueno';
+import {
+  baseProductRecommendationsOptionsSchema,
+  buildBaseProductRecommendationsList,
+} from './headless-base-product-recommendations';
+import {validateOptions} from '../../utils/validate-payload';
+import {ProductRecommendationEngine} from '../../app/product-recommendation-engine/product-recommendation-engine';
+
+const optionsSchema = new Schema({
+  ...baseProductRecommendationsOptionsSchema,
+});
+
+export type FrequentlyViewedDifferentCategoryListOptions = SchemaValues<
+  typeof optionsSchema
+>;
+
+export interface FrequentlyViewedDifferentCategoryListProps {
+  options?: FrequentlyViewedDifferentCategoryListOptions;
+}
+
+export type FrequentlyViewedDifferentCategoryList = ReturnType<
+  typeof buildFrequentlyViewedDifferentCategoryList
+>;
+export type FrequentlyViewedDifferentCategoryListState =
+  FrequentlyViewedDifferentCategoryList['state'];
+
+export const buildFrequentlyViewedDifferentCategoryList = (
+  engine: ProductRecommendationEngine,
+  props: FrequentlyViewedDifferentCategoryListProps
+) => {
+  const options = validateOptions(
+    engine,
+    optionsSchema,
+    props.options,
+    'buildFrequentlyViewedDifferentCategoryList'
+  ) as Required<FrequentlyViewedDifferentCategoryListOptions>;
+  return buildBaseProductRecommendationsList(engine, {
+    ...props,
+    options: {
+      ...options,
+      id: 'frequentlyViewedDifferentCategory',
+    },
+  });
+};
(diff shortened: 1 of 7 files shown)


thudon-coveo:
The failing test doesn't seem to be related to these changes

btaillon (approved):

ThibodeauJF:
Branch name should contain a JIRA # so the hooks add the link to the commits.

What is the purpose of those 2 controller, wouldn't one of them be similar to `headless-frequently-viewed-together` ?

thudon-coveo:
> Branch name should contain a JIRA # so the hooks add the link to the commits.
> 
> What is the purpose of those 2 controller, wouldn't one of them be similar to `headless-frequently-viewed-together` ?

There is no Jira ticket open for this change AFAIK. This is for a demo by the commerce product team for https://github.com/coveo/storefront-headless.

It is similar to `headless-frequently-viewed-together`, the difference is that we only want to show one category at a time. For exemple with `FrequentViewedSameCategory` we will have a "Top hats" section on the main page.

packages/headless/src/controllers/product-recommendations/headless-frequently-viewed-different-category.ts
```diff
@@ -0,0 +1,44 @@
+import {Schema, SchemaValues} from '@coveo/bueno';
+import {
+  baseProductRecommendationsOptionsSchema,
+  buildBaseProductRecommendationsList,
+} from './headless-base-product-recommendations';
+import {validateOptions} from '../../utils/validate-payload';
+import {ProductRecommendationEngine} from '../../app/product-recommendation-engine/product-recommendation-engine';
+
+const optionsSchema = new Schema({
+  ...baseProductRecommendationsOptionsSchema,
```
samisayegh, line 10: Should this controller add any options or adjust certain ones to be required? Maybe `category` should be a required option?

packages/headless/src/controllers/product-recommendations/headless-frequently-viewed-same-category.ts
```diff
@@ -0,0 +1,44 @@
...
+  options?: FrequentlyViewedSameCategoryListOptions;
+}
+
+export type FrequentlyViewedSameCategoryList = ReturnType<
+  typeof buildFrequentlyViewedSameCategoryList
+>;
+export type FrequentlyViewedSameCategoryListState =
+  FrequentlyViewedSameCategoryList['state'];
+
+export const buildFrequentlyViewedSameCategoryList = (
```
samisayegh, line 27: What is the difference between this controller and the one above? Aside from the name, it seems like the logic is identical?

Shouldn't there be some logic that makes the controllers behave differently, recommending within vs. in other categories?

packages/headless/src/controllers/product-recommendations/headless-frequently-viewed-different-category.ts
thudon-coveo: The objective was just to use a product's `permanentid` and show popular product from the same category. Exactly like `frequently-viewed-together` but scope to one category.

packages/headless/src/controllers/product-recommendations/headless-frequently-viewed-same-category.ts
thudon-coveo: Indeed they're identical , the only difference is their `id`.

I guess I'm pretty naive but I thought all that matter was the `id` of the strategy on the Coveo platform.

samisayegh: Ok, if the id is enough, then all might be good. Are the strategy string ids documented somewhere? Maybe [this table](https://docs.coveo.com/en/3284/coveo-for-commerce/product-recommendations-deployment) should be updated?

samisayegh (approved):
If you want to make these new controllers available, you will need to export them from `product-recommendation.index.ts`. If you want to keep them private though (e.g. until an api is ready), then the current PR is good,

packages/headless/src/controllers/product-recommendations/headless-frequently-viewed-same-category.ts
thudon-coveo: I was not aware of this table. Thanks for sharing I'll update it !

jdevost:
I feel there should be a way to pass in the `strategy` used for the recommendation rather than hardcoding it in a ID, and then having to duplicate the controller for `frequently-viewed-together` each time we want to try a new strategy to simply change that ID.

PR #1668, opened 2022-01-14
test(atomic): make test stable using the old setup 

samisayegh:
I think I turned all the stones ...

I commented out all code that is clearing suggestions. Still, I saw suggestions being cleared in cypress. So, I think the test code is at fault.

One way to make it work™️ is to slow down the typing. At 200ms, I did not see any failing attempts locally. At 150ms, they still occasionally happened.

Since this solution relies on delays, it is not as stable as the initial PR, so I put the changes in a separate branch,

### DIFF
--- a/packages/atomic/cypress/integration/search-box.cypress.ts
+++ b/packages/atomic/cypress/integration/search-box.cypress.ts
@@ -57,21 +57,6 @@ describe('Search Box Test Suites', () => {
         .init();
     }
 
-    function setupWithRecentQueries() {
-      new TestFixture()
-        .with(setRecentQueries(numOfRecentQueries))
-        .with(
-          addSearchBox({
-            recentQueries: {
-              maxWithoutQuery: maxRecentQueriesWithoutQuery,
-              maxWithQuery: numOfRecentQueries,
-            },
-            props: {'number-of-queries': numOfRecentQueries},
-          })
-        )
-        .init();
-    }
-
     describe('without input', () => {
       const expectedSum =
         maxSuggestionsWithoutQuery + maxRecentQueriesWithoutQuery;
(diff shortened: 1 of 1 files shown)


packages/atomic/cypress/integration/search-box.cypress.ts
```diff
@@ -121,12 +106,10 @@ describe('Search Box Test Suites', () => {
...
         before(() => {
-          setupWithRecentQueries();
-          setInputText();
+          setupWithSuggestionsAndRecentQueries();
 
-          SearchBoxSelectors.inputBox()
-            .focus()
-            .type('{downarrow}{downarrow}{downarrow}');
+          const downKeys = Array(9).fill('{downarrow}').join('');
+          SearchBoxSelectors.inputBox().type(`Rec${downKeys}`, {delay: 200});
```
samisayegh, line 112: The ideal sequence would be:

- Click or focus or type into the input
- Assert that all the suggestions have loaded
- Keydown to the suggestion

However, if two `type` commands s are used, there is a blur in between that clears the suggestions. I tried setting [release](https://docs.cypress.io/api/commands/type#:~:text=true-,Keep%20a%20modifier%20activated%20between%20commands,-scrollBehavior) to `false`, but it did not change the behavior.

Working with the constraint of needing to do everything in one `type`, I set a delay to give time for the suggestions to render in the list before keying down.

I tried other sequences like `focus` -> `wait` -> `type` but the outcome was always the same.

btaillon: > However, if two `type` commands s are used, there is a blur in between that clears the suggestions. I tried setting [release](https://docs.cypress.io/api/commands/type#:~:text=true-,Keep%20a%20modifier%20activated%20between%20commands,-scrollBehavior) to `false`, but it did not change the behavior.

Have you tried setting [force](https://docs.cypress.io/api/commands/type#:~:text=after%20each%20keypress-,force,-false)?

btaillon: Alternatively, looking at Cypress' source code ([keyboard.ts](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cy/keyboard.ts)), it doesn't look like Cypress does anything more "native" than firing keyboard events and changing the `value` of textboxes.

Perhaps we could add our own `pressKey` function (or even somehow import Cypress' `typeSimulatedKey`).

btaillon (approved):

ThibodeauJF (approved):

packages/atomic/cypress/integration/search-box.cypress.ts
samisayegh: I tried the `force` option but did not see a difference. I'm not sure if a `pressKey` function would give different results either. The down arrows register, but clear the recent query suggestions for some reason 🤷

PR #1670, opened 2022-01-17
feat(headless): facet show more/less improvements

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1326

This does not implement the improvements for the other use case(s) such as the product listing page, which aren't using search actions, these still work as before. I could implement similar changes in another story/PR if deemed necessary


jpmarceau (approved):

btaillon (approved):

packages/headless/src/features/facets/category-facet-set/category-facet-set-slice.test.ts
```diff
@@ -786,17 +788,43 @@ describe('category facet slice', () => {
...
+      search.response.facets = facets;
+
+      return fetchFacetValues.fulfilled(
+        search,
+        '',
+        logSearchEvent({evt: 'foo'})
+      );
+    }
+
+    testFulfilledSearchRequest(buildFetchFacetValuesAction);
```
samisayegh, line 828: I think a single test checking the happy path of `fetchFacetValues` would be enough here.

The current approach has the plus of being "extra safe", but running the block of tests twice just means tests will fail in unison on any source change? With a single new test, we'd have the same coverage, but have fewer tests at run-time and avoid duplicate failures.

samisayegh (approved):
I think it would be good to adjust the Product Lising use-case to behave similarly, unless there's a reason not to do so?

ThibodeauJF:
@samisayegh 
For the product listing, I'm not agaisnt changing the behaviour, it's just an entirely separate path unrelated path, takes twice the amount of work, which is why we should split it in another story

PR #1671, opened 2022-01-17
fix(quantic): added escape util for showing query in rich-text

nathanlb:
`lightning-formatted-rich-text` which we use for bolding renders html in a query. I added a util that uses html <textarea> escaping in a clever way so the queries are displayed properly.

Before
Screen Shot 2022-01-17 at 10 26 45 AM
Screen Shot 2022-01-17 at 10 26 31 AM

After
Screen Shot 2022-01-17 at 10 25 36 AM
Screen Shot 2022-01-17 at 10 25 55 AM
Screen Shot 2022-01-17 at 10 25 44 AM

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
@@ -84,7 +84,7 @@ describe('c/quanticUtils', () => {
       });
 
       it('should escape html anchor tags', () => {
-        let htmlWithTags = '<a src="http:.//www.sketchysite.com"></a>';
+        let htmlWithTags = '<a src="http://www.sketchysite.com"></a>';
       
         expect(I18nUtils.escapeHTML(htmlWithTags)).toBe("&lt;a src=\"http:.//www.sketchysite.com\"&gt;&lt;/a&gt;");
       });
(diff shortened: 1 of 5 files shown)


aelfaleh (approved):

packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
```diff
@@ -75,5 +75,19 @@ describe('c/quanticUtils', () => {
...
+
+    describe('escapeHTML', () => {
+      it('should escape html tags', () => {
+        let htmlWithTags = '<div>test string</div>';
+      
+        expect(I18nUtils.escapeHTML(htmlWithTags)).toBe('&lt;div&gt;test string&lt;/div&gt;');
+      });
+
+      it('should escape html anchor tags', () => {
+        let htmlWithTags = '<a src="http:.//www.sketchysite.com"></a>';
```
lbergeron, line 87: Why is there a dot before the two forward slashes?

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
nathanlb: typo

PR #1673, opened 2022-01-18
test(quantic): add standalone searchbox tests

aelfaleh:


packages/quantic/cypress/integration/standalone-search-box/standalone-search-box.cypress.ts
```diff
@@ -0,0 +1,92 @@
...
+        visitStandaloneSearchBox();
+
+        Actions.typeInSearchBox('some query');
+        Expect.displayCloseIcon(true);
+      });
+
+      scope('when setting queries in local storage', () => {
+        typeQueriesListLocalStorage(
+          '["query 1", "query 2", "query 3", "query 4", "query 5", "query 6"]'
+        );
```
nathanlb, line 49: this test has no assertions so I'm not sure what we're testing here.

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box.cypress.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  function visitStandaloneSearchBox(
+    options: Partial<StandaloneSearchBoxOptions> = {}
+  ) {
+    interceptSearch();
+    cy.visit(standaloneSearchBoxUrl);
+    configure(options);
+  }
+
+  describe('with default option', () => {
```
nathanlb, line 29: The main feature of the standalone searchbox is that it redirects to a search page with a given query. So I think we should be testing with option `redirectUrl="/full-search-example"` and test that after the redirect is made to the example search page the correct query is made whether it was a typed queries or a query suggestion selection.

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-expectations.ts
```diff
@@ -0,0 +1,53 @@
...
+        .should(display ? 'exist' : 'not.exist')
+        .logDetail(`${should(display)} display the search icon`);
+    },
+    displaySearchButton: (display: boolean) => {
+      selector
+        .searchButton()
+        .should(display ? 'exist' : 'not.exist')
+        .logDetail(`${should(display)} display the search button`);
+    },
+    displayCloseIcon: (display: boolean) => {
```
nathanlb, line 29: I know technically the icon name is "close" but we're using it to "clear" the text content so I'm wondering if `displayClearIcon` or perhaps `displayClearButton` might make more sense here. What do you think?

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-selectors.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const StandaloneSearchBoxSelectors: StandaloneSearchBoxSelector = {
+  get: () => cy.get(standaloneSearchBoxComponent),
+  input: () => StandaloneSearchBoxSelectors.get().find('input[type="search"]'),
+  suggestionList: () =>
+    StandaloneSearchBoxSelectors.get().find(
+      'c-quantic-search-box-suggestions-list li'
+    ),
+  closeIcon: () =>
```
nathanlb, line 20: maybe here would've been a better place to put my clearButton idea 😅

packages/quantic/cypress/page-objects/actions/action-set-local-storage.ts
```diff
@@ -1,3 +1,11 @@
+export const typeQueriesListLocalStorage = (queries: string) =>
```
nathanlb, line 1: why do we need this?

packages/quantic/force-app/examples/main/lwc/exampleQuanticStandaloneSearchBox/exampleQuanticStandaloneSearchBox.js
```diff
@@ -0,0 +1,45 @@
...
+        attribute: 'numberOfSuggestions',
+        label: 'Number of suggestions',
+        description: 'The maximum number of suggestions to display.',
+        defaultValue: 5,
+    },
+    {
+        attribute: 'redirectUrl',
+        label: 'Redirect url',
+        description: 'The url of the search page to redirect to when a query is made.',
+        defaultValue: '/global-search/%40uri',
```
nathanlb, line 33: I'm a bit torn between having the actual default value like we have for every other components or use `'/full-search-example'` so that it redirects to an existing page rather than be broken by default. What do you think?

nathanlb (commented):
I think we're still missing some tests to cover the functionality of the searchbox. The setup so far is on a good track 👍

packages/quantic/force-app/examples/main/lwc/exampleQuanticStandaloneSearchBox/exampleQuanticStandaloneSearchBox.js
aelfaleh: it changes nothing in my test, I use `/full-search-example` as custom for redirectUrl and it works 😄

packages/quantic/cypress/page-objects/actions/action-set-local-storage.ts
aelfaleh: we don't need it, It s a mistake

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-selectors.ts
aelfaleh: good catch

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-expectations.ts
aelfaleh: done

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box.cypress.ts
aelfaleh: I tested it with both url and it went well

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box.cypress.ts
aelfaleh: I miss a commit 😄

packages/quantic/force-app/examples/main/lwc/exampleQuanticStandaloneSearchBox/exampleQuanticStandaloneSearchBox.js
nathanlb: that's great! My thinking is that if a human visits the example community then this default value won't be very helpful. But  it's not a critical question to answer at the moment.

nathanlb (approved):
Once the tests pass I think this is GG 👍

lvu285:
Test looks good, but I would love that we could check the UA event when search & redirect

nathanlb:
> Test looks good, but I would love that we could check the UA event when search & redirect

Oh yes very good point!

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-actions.ts
```diff
@@ -0,0 +1,27 @@
...
+        .invoke('val', query)
+        .logAction(`when typing "${query}" in search box`);
+    },
+    submitSearch: () => {
+      selector.searchButton().click().logAction('when submitting query');
+    },
+    focusSearchBox: () => {
+      selector.input().trigger('keydown', {keyCode: 27});
+      cy.wait(500);
+      selector.input().trigger('keyup', {keyCode: 27});
```
lbergeron, line 20: Why not use `selector.input().focus()` instead?

Please use a constant for `keyCode: 27`, I didn't remember it was the escape key :) Is it mandatory to use the escape key, or any other key could do?

Also 500 ms is a pretty long delay between `keydown` and `keyup` events. Is this delay even required?

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-selectors.ts
```diff
@@ -0,0 +1,26 @@
...
+  get: () => cy.get(standaloneSearchBoxComponent),
+  input: () => StandaloneSearchBoxSelectors.get().find('input[type="search"]'),
+  suggestionList: () =>
+    StandaloneSearchBoxSelectors.get().find(
+      'c-quantic-search-box-suggestions-list li'
+    ),
+  clearButton: () =>
+    StandaloneSearchBoxSelectors.get().find('.slds-input__icon_right'),
+  searchIcon: () =>
+    StandaloneSearchBoxSelectors.get().find('.slds-input__icon_left'),
```
lbergeron, line 23: These selectors are kind of generic. Is there an attribute we could add to make them more specific and robust?

packages/quantic/force-app/examples/main/lwc/exampleQuanticStandaloneSearchBox/exampleQuanticStandaloneSearchBox.js
```diff
@@ -0,0 +1,45 @@
...
+    {
+        attribute: 'numberOfSuggestions',
+        label: 'Number of suggestions',
+        description: 'The maximum number of suggestions to display.',
+        defaultValue: 5,
+    },
+    {
+        attribute: 'redirectUrl',
+        label: 'Redirect url',
+        description: 'The url of the search page to redirect to when a query is made.',
```
lbergeron, line 32:
```suggestion
        description: 'The URL of the search page to redirect to when a query is performed.',
```

lbergeron (approved):
That looks good! Well done!

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-actions.ts
aelfaleh: I tried `selector.input().focus()` but it didn't work and the test was too flaky
I found that simulating the whole action of escape key works better

lbergeron: Ok, thanks!

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box.cypress.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      scope('with custom #numberOfSuggestions', () => {
+        visitStandaloneSearchBox({
+          numberOfSuggestions: 1,
+        });
+
+        Actions.focusSearchBox();
+        cy.wait(InterceptAliases.QuerySuggestions);
+        Expect.displaySuggestionList(true);
+        Expect.numberOfSuggestions(1);
```
lvu285, line 110: can we cover the test to select the Suggestion in this pr too? if  not, I suggest we create another task to do it later!

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box.cypress.ts
```diff
@@ -0,0 +1,127 @@
...
+      scope('when loading standalone search box', () => {
+        visitStandaloneSearchBox();
+
+        Expect.displayInputSearchBox(true);
+        Expect.displaySearchButton(true);
+        Expect.placeholderContains('Search');
+      });
+
+      scope('when submitting a search', () => {
+        const query = 'test';
```
lvu285, line 59: I suggest have a test to cover few different queries with special character, and script .. if it's not too long for the test to run

lvu285 (approved):

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box.cypress.ts
```diff
@@ -0,0 +1,127 @@
...
+        visitStandaloneSearchBox();
+
+        Actions.typeInSearchBox(query);
+        Expect.displayClearButton(true);
+        Actions.submitSearch();
```
lvu285, line 64: for search box, should cover the part when if we press "ENTER" to submit the search

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
```diff
@@ -116,7 +116,9 @@ export default class QuanticStandaloneSearchBox extends NavigationMixin(
   }
 
   renderedCallback() {
-    initializeWithHeadless(this, this.standaloneEngineId, this.initialize);
+    initializeWithHeadless(this, this.standaloneEngineId, this.initialize).then(() => {
+      this.input.setAttribute('is-initialized', 'true');
```
lbergeron, line 120: Awesome!

lbergeron (approved):
Good work!

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
nathanlb: is this attribute just for the tests?

aelfaleh: yes

nathanlb: Cool thanks. GG!

PR #1677, opened 2022-01-18
feat(atomic): expose idiomatic react templating functionality

olamothe:
As discussed in scrum, the goal is to expose a more natural React pattern to render result templates VS the Web Component/HTML way. 

Changes included in this PR: 

* Remove `atomic-result-template` and `atomic-field-condition` from the output target (These components will no longer appear under `@coveo/atomic-react`
* Add a `setRenderFunction` method to the core Atomic result list web component, which accept a function that receive a result as a parameter, and has to return an HTMLElement
* In atomic-react,  remove the ResultTemplate wrapper, and instead expose a ResultListWrapper, that shadows `AtomicResultList`, with one additional prop: `template`, which accept a result as a parameter and must return a JSX.Element. It then does the necessary work to change this to an HTMLElement, which is what Atomic expects.

Also included in this PR: 
 A patch to fix an issue with the react output plugin, which does not properly `camelToDashCase` HTML attributes.

https://coveord.atlassian.net/browse/KIT-1336


packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -182,7 +197,9 @@ export class AtomicResultList implements InitializableComponent {
 
   private buildListResults() {
     return this.resultListState.results.map((result) => {
-      const template = this.getTemplate(result);
+      const content = this.renderingFunction
```
ThibodeauJF, line 200: More simple than I expected 😄

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -90,6 +99,12 @@ export class AtomicResultList implements InitializableComponent {
    */
   @Prop() image: ResultDisplayImageSize = 'icon';
 
+  private renderingFunction?: (result: Result) => HTMLElement = undefined;
+
+  @Method() public setRenderFunction(render: (result: Result) => HTMLElement) {
```
ThibodeauJF, line 104: Should we set this as `@internal` ?

packages/atomic/stencil.config.ts
```diff
@@ -46,6 +46,7 @@ export const config: Config = {
       componentCorePackage: '@coveo/atomic',
       proxiesFile: '../atomic-react/src/components/stencil-generated/index.ts',
       includeDefineCustomElements: true,
+      excludeComponents: ['atomic-result-template', 'atomic-field-condition'],
```
ThibodeauJF, line 49: A bit unrelated, but maybe the atomic-external should be investigated and perhaps excluded too if it makes thing too complicated (the react context should remove the necessity for it)

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
```diff
@@ -208,3 +138,72 @@ export function AtomicReactPage() {
     </AtomicSearchInterface>
   );
 }
+
+function MyTemplate(result: Result) {
+  return (
+    <>
```
ThibodeauJF, line 144: is this like a Fragment?

ThibodeauJF (approved):

packages/samples/headless-react/src/pages/AtomicReactPage.tsx
olamothe: It's a way to return a "single root" for a JSX function, without the need for any extra DOM node.

https://reactjs.org/docs/fragments.html

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
olamothe: Is this only for documentation ?

samisayegh: Speaking of docs, a line to explain what this function is for would be useful, especially since it's not directly used inside Atomic

packages/atomic/patches/@stencil+react-output-target+0.2.0.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/node_modules/@stencil/react-output-target/react-component-lib/createComponent.tsx b/node_modules/@stencil/react-output-target/react-component-lib/createComponent.tsx
```
samisayegh, line 1: Cool approach to work-around the issue. It would be nice to open a PR for the stencil package too,

samisayegh (approved):

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
btaillon: Whether we make it public/documented or not, this will need to be maintained since we use it in atomic-react and it's likely going to be used in other frameworks. I have a slight preference for documenting it.

btaillon (approved):
clean!

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
olamothe: @jpmarceau / @npushkarskii : Small review for the comment here please !

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -91,6 +100,22 @@ export class AtomicResultList implements InitializableComponent {
    */
   @Prop() image: ResultDisplayImageSize = 'icon';
 
+  private renderingFunction?: (result: Result) => HTMLElement = undefined;
+
+  /**
+   * Set a rendering function, that will bypass the standard HTML template mechanism to render results.
+   * This render function is typically used for integration with web frameworks that do not normally use raw HTML, such as React, Angular or Vue.
+   *
+   * Under normal circumstances, this method should not be used when integrating Atomic in a standard HTML based deployment.
```
npushkarskii, line 109:
```suggestion
   * Sets a rendering function to bypass the standard HTML template mechanism for rendering results.
   * You can use this function while working with web frameworks that don't use plain HTML syntax, e.g., React, Angular or Vue.
   *
   * Do not use this method if you integrate Atomic in a plain HTML deployment.
```

packages/atomic/src/components.d.ts
```diff
@@ -523,6 +523,11 @@ export namespace Components {
           * The expected size of the image displayed in the results.
          */
         "imageSize"?: ResultDisplayImageSize;
+        /**
+          * Set a rendering function, that will bypass the standard HTML template mechanism to render results. This render function is typically used for integration with web frameworks that do not normally use raw HTML, such as React, Angular or Vue.  Under normal circumstances, this method should not be used when integrating Atomic in a standard HTML based deployment.
```
npushkarskii, line 527:
```suggestion
          * Sets a rendering function to bypass the standard HTML template mechanism for rendering results. You can use this function while working with web frameworks that don't use plain HTML syntax, e.g., React, Angular or Vue. Do not use this method if you integrate Atomic in a plain HTML deployment.
```

npushkarskii (approved):
LGTM, just added a couple of edits 👍

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
npushkarskii: @olamothe done! 😉

packages/atomic/src/components.d.ts
npushkarskii: @olamothe  seems like it may have sense to commit this suggestion as well 🤔 Or it may not.

PR #1680, opened 2022-01-19
test(atomic): removed cy.wait from all facet tests

btaillon:
https://coveord.atlassian.net/browse/KIT-1344

I removed `cy.wait` from all facet tests and added relevant assertions after every action to leverage Cypress's [retry mechanism](https://docs.cypress.io/guides/core-concepts/retry-ability).

There was a ton of repetition between different files in this PR (as in, removing the same *kind* of line, changing the same other *kind* of line, etc.). The most important parts to look at are the `*-action` files and the `*-selectors` files. Apart from that, changes to the `*.cypress.ts` files and `*-assertions.ts` files follow the same pattern.


packages/atomic/cypress/integration/breadbox/breadbox-assertions.ts
```diff
@@ -191,9 +191,7 @@ export function assertDeselectColorFacet(index: number) {
```
btaillon, line 191: Some breadcrumb analytics assertions were used in facet tests, so I converted them to the new format. This new format will also wait until `TestFixture.interceptAliases.UA` if it wasn't already intercepted.

packages/atomic/cypress/integration/facets/category-facet/category-facet-actions.ts
```diff
@@ -63,11 +63,6 @@ export function pressParentButton(index: number) {
     });
```
btaillon, line 63: I extracted some of these actions to `facet-common-actions`.

packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
```diff
@@ -133,7 +133,7 @@ describe('Category Facet Test Suites', () => {
       describe('when selecting the "Show more" button', () => {
         function setupShowMore() {
           setupGoDeeperOneLevel();
-          pressShowMore();
+          pressShowMore(CategoryFacetSelectors);
```
btaillon, line 136: Some of these actions are now part of `facet-common-actions`, so they need to take a selector as a parameter.

packages/atomic/cypress/integration/facets/color-facet/color-facet-selectors.ts
```diff
@@ -11,6 +11,12 @@ export const ColorFacetSelectors = {
     ColorFacetSelectors.shadow().find(
       '[part="value-box"][aria-pressed="false"]'
     ),
+  selectedBoxValueWithText: (text: string) =>
+    ColorFacetSelectors.shadow().find(
+      `[part="value-box"][aria-pressed="true"]:contains("${text}")`
+    ),
+  idleBoxValueLabel: () =>
+    ColorFacetSelectors.idleBoxValue().find('[part="value-label"]'),
```
btaillon, line 19: These two selectors (with equivalents for `checkbox` and `link`) were added to every facet. They are used by `selectIdle*ValueAt` tests.

packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
```diff
@@ -575,7 +575,7 @@ describe('Category Facet Test Suites', () => {
       const query = 'mal';
       function setupSearchFor() {
         setupWithFacetSearch();
-        typeFacetSearchQuery(CategoryFacetSelectors, query);
+        typeFacetSearchQuery(CategoryFacetSelectors, query, true);
```
btaillon, line 578: `typeFacetSearchQuery` now has the `shouldFindResults` boolean parameter.

packages/atomic/cypress/integration/facets/facet-common-actions.ts
```diff
@@ -11,31 +11,103 @@ export function selectIdleCheckboxValueAt(
   FacetWithCheckboxSelector: FacetWithCheckboxSelector,
   index: number
 ) {
```
btaillon, line 13: This action is pretty complex. Unfortunately, it kinda needed to be if we want these functions to be self-sufficient (not requiring assertions in `*.cypress.ts` files).

packages/atomic/cypress/integration/facets/facet-common-actions.ts
```diff
@@ -11,31 +11,103 @@ export function selectIdleCheckboxValueAt(
...
-  FacetWithCheckboxSelector.idleCheckboxValue().eq(index).click();
+  FacetWithCheckboxSelector.idleCheckboxValueLabel()
+    .eq(index)
+    .then((idleCheckboxValueLabel) => {
+      const hasAriaLabel = !!idleCheckboxValueLabel.attr('aria-label');
+      const text = hasAriaLabel
+        ? idleCheckboxValueLabel.attr('aria-label')!
+        : idleCheckboxValueLabel.text();
```
btaillon, line 20: Rating facet values don't contain text, but they have an aria-label.

packages/atomic/cypress/integration/facets/facet-common-actions.ts
```diff
@@ -11,31 +11,103 @@ export function selectIdleCheckboxValueAt(
...
+      FacetWithCheckboxSelector.idleCheckboxValueLabel()
+        .filter(
+          hasAriaLabel ? `[aria-label="${text}"]` : `:contains("${text}")`
+        )
+        .its('length')
+        .should(
+          'eq',
+          1,
+          'There should not be any other value similar to this one.'
+        );
```
btaillon, line 30: This assertion is not for the purpose of waiting, I added it because I expect we could get some flakiness with `:contains(TEXT)` if we select multiple values that are similar.

E.g.,: if we selected "sewing table" then selected "table", then I think the last assertion would pass immediately without waiting for re-render.

I prefer having this assertion to prevent even the possibility of introducing a test case that is flaky like that.

packages/atomic/cypress/integration/facets/facet-common-actions.ts
```diff
@@ -11,31 +11,103 @@ export function selectIdleCheckboxValueAt(
...
+      cy.wrap(idleCheckboxValueLabel).click();
```
btaillon, line 31: Like with the actions I introduced last time for category facets, we need to call cypress's `click` instead of the native one. Otherwise, focus is lost.

packages/atomic/cypress/integration/facets/facet/facet.cypress.ts
```diff
@@ -987,9 +963,9 @@ describe('Facet v1 Test Suites', () => {
       const index = [0, 1, 2];
       function setupSelectedMulitpleFacets() {
         setupBreadboxWithFacet();
-        index.forEach((i: number) => {
-          selectIdleCheckboxValueAt(FacetSelectors, i);
-          cy.wait(TestFixture.interceptAliases.Search);
+        index.forEach((position, i) => {
+          selectIdleCheckboxValueAt(FacetSelectors, position);
+          BreadboxSelectors.breadcrumbButton().should('have.length', i + 1);
```
btaillon, line 968: At some point, I had flakiness here due to a bad test and I added this assertion. I didn't remove it after fixing the test, since I think it's still relevant. Although we are selecting facet values, what we're concerned about in this section of the tests is which breadcrumbs are rendered, and I want to make sure breadcrumbs were rendered too.

olamothe (approved):
It's a big PR, so I gotta admit I did not take the time to review every line one by one.

But in general it seems to be going in the right direction. So, good job !

However, just a general warning about trying to share test code across components, because there are such things as "too much abstraction", and it starts to have diminishing return. 

And then it's easy to fall into a trap where we invest too much time into something that is ultimately just "a tool", and not the core thing we should be focusing on 😄

btaillon:
> It's a big PR, so I gotta admit I did not take the time to review every line one by one.

Yes, unfortunately that was my worry as I was writing it. I decided to open a big PR anyway rather than split it because most changes are the same across different files and I want cypress tests to pass ASAP.

> However, just a general warning about trying to share test code across components, because there are such things as "too much abstraction", and it starts to have diminishing return.
>
> And then it's easy to fall into a trap where we invest too much time into something that is ultimately just "a tool", and not the core thing we should be focusing on 😄

Fully agreed. I abstracted a few things because I felt like it made a lot of sense for those and matched the current pattern of our tests, but I don't want to abstract more than this.

PR #1686, opened 2022-01-20
fix(atomic): improved accessibility for pager

btaillon:
https://coveord.atlassian.net/browse/KIT-1328

* I transformed the buttons in the pager into radio buttons
* Since selecting a page moves the focus, I made sure that the page waits for the focus to be changed before scrolling up
* I extracted the radio button logic into a component similar to `<Button>`
* I removed the ul/li in favor of a `radiogroup`.

### DIFF
--- a/packages/atomic/src/components/atomic-pager/atomic-pager.tsx
+++ b/packages/atomic/src/components/atomic-pager/atomic-pager.tsx
@@ -20,7 +20,7 @@ import {
   FocusTargetController,
 } from '../../utils/accessibility-utils';
 import {randomID} from '../../utils/utils';
-import {createRipple} from '../../utils/ripple';
+import {RadioButton} from '../common/radio-button';
 
 /**
  * The `atomic-pager` provides buttons that allow the end user to navigate through the different result pages.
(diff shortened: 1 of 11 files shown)


packages/atomic/src/components/common/radio-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+import {createRipple} from '../../utils/ripple';
+import {
+  ButtonStyle,
+  getClassNameForButtonStyle,
+  getRippleColorForButtonStyle,
+} from './button-style';
+
+export interface RadioButtonProps {
+  style: ButtonStyle;
+  groupName: string;
```
btaillon, line 11: I decided to make `groupName` non-optional, since IMO every radio button should be related to other radio buttons.

packages/atomic/src/components/common/button-style.tsx
```diff
@@ -0,0 +1,46 @@
...
+    case 'outline-primary':
+      return 'neutral';
+    case 'outline-neutral':
+      return 'neutral';
+    case 'outline-bg-neutral':
+      return 'neutral';
+    case 'text-primary':
+      return 'neutral';
+    case 'text-neutral':
+      return 'neutral';
```
samisayegh, line 42: Could these statements be grouped under `default`?

```
case 'primary':
      return 'primary';
case 'text-transparent':
      return 'neutral-light';
default:
    return 'neutral';
```

samisayegh (approved):

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -24,7 +24,7 @@ export function AriaLiveRegion(regionName: string) {
 
 export interface FocusTargetController {
   setTarget(element: HTMLElement | undefined): void;
-  focusAfterSearch(): void;
+  focusAfterSearch(): Promise<void>;
```
btaillon, line 27: It made sense to me that `focusAfterSearch` should return a promise that resolves after something was focused. If you disagree, I can find a different solution.

ThibodeauJF (approved):

packages/atomic/src/components/atomic-pager/atomic-pager.tsx
```diff
@@ -19,11 +19,14 @@ import {
...
   FocusTargetController,
 } from '../../utils/accessibility-utils';
+import {randomID} from '../../utils/utils';
+import {RadioButton} from '../common/radio-button';
 
 /**
  * The `atomic-pager` provides buttons that allow the end user to navigate through the different result pages.
  *
- * @part buttons - The list of buttons.
+ * @part buttons - The list containing the next/previous buttons and page-buttons.
```
npushkarskii, line 28:
```suggestion
 * @part buttons - The list of the next/previous buttons and page-buttons.
```

npushkarskii (approved):
LGTM, I only left one comment 👍

PR #1687, opened 2022-01-20
test(atomic): add smoke test for atomic-react

olamothe:
https://coveord.atlassian.net/browse/KIT-1359


samisayegh (approved):
Why are the tests inside `headless-react` instead of `atomic-react`? I think you mentioned it in the scrum, but I did not fully catch the reason,

ThibodeauJF (approved):

olamothe:
> Why are the tests inside `headless-react` instead of `atomic-react`? I think you mentioned it in the scrum, but I did not fully catch the reason,

Because `atomic-react` is just a package with the components, it does not actually contain a working React project/setup. 

So instead of re-installing/duplicating all the dependencies needed there as dev dependency, we can just piggy back on the React sample package which already has everything needed.

PR #1688, opened 2022-01-20
test(atomic): removed necessity for cy.wait after clearing a breadcrumb

btaillon:
https://coveord.atlassian.net/browse/KIT-1347

### DIFF
--- a/packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
@@ -31,6 +31,7 @@ import {
   breadboxComponent,
   BreadboxSelectors,
 } from '../../breadbox/breadbox-selectors';
+import {AnalyticsTracker} from '../../../utils/analyticsUtils';
 
 describe('Color Facet Test Suites', () => {
   describe('with default setting', () => {
@@ -176,6 +177,7 @@ describe('Color Facet Test Suites', () => {
         describe('when selecting a search result', () => {
           function setupSelectSearchResult() {
             setupSearchFor();
+            AnalyticsTracker.reset();
             selectIdleBoxValueAt(0);
           }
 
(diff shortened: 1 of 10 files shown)


packages/atomic/cypress/support/commands.ts
```diff
@@ -1,6 +1,6 @@
 // Must be declared global to be detected by typescript (allows import/export)
-
-import {AnalyticsTracker, SearchEventRequest} from '../utils/analyticsUtils';
+import {SearchEventRequest} from '@coveo/headless/node_modules/coveo.analytics/src/events';
+import {AnalyticsTracker} from '../utils/analyticsUtils';
```
btaillon, line 3: Not sure why this was working before. `SearchEventRequest` is not exported from `analyticsUtils`, and exporting it causes an error due to it being an import from `src`.

packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
```diff
@@ -176,6 +177,7 @@ describe('Color Facet Test Suites', () => {
         describe('when selecting a search result', () => {
           function setupSelectSearchResult() {
             setupSearchFor();
+            AnalyticsTracker.reset();
```
btaillon, line 180: In some cases where two analytics of the same type are sent in a row, tests are gonna fail since they will assert on the previously logged analytics rather than the upcoming one. There's two ways to prevent this.
1. Clearing old recorded analytics
2. Making AnalyticsTracker more specific to leverage Cypress's retry mechanism

packages/atomic/cypress/support/commands.ts
```diff
@@ -20,18 +20,18 @@ declare global {
 
 Cypress.Commands.add('expectSearchEvent', (actionCause) => {
   cy.wrap(AnalyticsTracker)
-    .invoke('getLastSearchEvent')
+    .invoke('getLastSearchEvent', actionCause)
```
btaillon, line 23: As stated [above](https://github.com/coveo/ui-kit/pull/1688/files#r789146098), I also decided to make AnalyticsTracker a bit more specific and take the `actionCause` or `eventType` as a parameter.

I could make it fully specific to even further reduce flakiness by making `getLastSearchEvent`, `getLastCustomEvent` and `getLastClickEvent` take an object as a parameter and assert on every key-value of the object. I decided not to do this since this is a bigger change and would require some more code to recursively compare objects, but I can do it if you prefer that.

ThibodeauJF (approved):

olamothe (approved):

samisayegh (approved):

PR #1689, opened 2022-01-21
feat(atomic): allow atomic-search-box to be standalone

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-937

Follows this guide https://docs.coveo.com/en/headless/latest/usage/standalone-search-box-usage/

### DIFF
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -642,6 +642,10 @@ export namespace Components {
           * The amount of queries displayed when the user interacts with the search box. By default, a mix of query suggestions and recent queries will be shown. You can configure those settings using the following components as children:  - atomic-search-box-query-suggestions  - atomic-search-box-recent-queries
          */
         "numberOfQueries": number;
+        /**
+          * Defining this options makes the search box standalone.  The default Url the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that Url instead.
+         */
+        "redirectionUrl"?: string;
     }
     interface AtomicSearchBoxQuerySuggestions {
         /**
@@ -1873,6 +1877,10 @@ declare namespace LocalJSX {
           * The amount of queries displayed when the user interacts with the search box. By default, a mix of query suggestions and recent queries will be shown. You can configure those settings using the following components as children:  - atomic-search-box-query-suggestions  - atomic-search-box-recent-queries
          */
         "numberOfQueries"?: number;
+        /**
+          * Defining this options makes the search box standalone.  The default Url the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that Url instead.
+         */
+        "redirectionUrl"?: string;
     }
     interface AtomicSearchBoxQuerySuggestions {
         /**
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -13,6 +13,7 @@ import { ResultDisplayDensity, ResultDisplayImageSize, ResultDisplayLayout } fro
 import { TemplateContent } from "./components/atomic-result-template/atomic-result-template";
 import { i18n } from "i18next";
 import { InitializationOptions } from "./components/atomic-search-interface/atomic-search-interface";
+import { StandaloneSearchBoxData } from "./utils/local-storage-utils";
 export namespace Components {
     interface AtomicAriaLive {
     }
@@ -643,7 +644,7 @@ export namespace Components {
          */
         "numberOfQueries": number;
         /**
-          * Defining this option makes the search box standalone. This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
+          * Defining this option makes the search box standalone.  This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
          */
         "redirectionUrl"?: string;
     }
@@ -1878,7 +1879,7 @@ declare namespace LocalJSX {
          */
         "numberOfQueries"?: number;
         /**
-          * Defining this option makes the search box standalone. This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
+          * Defining this option makes the search box standalone.  This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
          */
         "redirectionUrl"?: string;
     }
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -643,7 +643,7 @@ export namespace Components {
          */
         "numberOfQueries": number;
         /**
-          * Defining this options makes the search box standalone.  The default Url the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that Url instead.
+          * Defining this option makes the search box standalone. This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
          */
         "redirectionUrl"?: string;
     }
@@ -1878,7 +1878,7 @@ declare namespace LocalJSX {
          */
         "numberOfQueries"?: number;
         /**
-          * Defining this options makes the search box standalone.  The default Url the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that Url instead.
+          * Defining this option makes the search box standalone. This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
          */
         "redirectionUrl"?: string;
     }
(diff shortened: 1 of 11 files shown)


packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
```diff
@@ -144,7 +143,10 @@ export function buildStandaloneSearchBox(
       return {
         ...searchBox.state,
         isLoading: standaloneSearchBoxState.isLoading,
-        redirectTo: standaloneSearchBoxState.redirectTo,
+        redirectTo:
```
ThibodeauJF, line 146: The type is string | null, but in the slice's state, it's only a string. This poses a bit of a comprehension problem.

packages/atomic/src/pages/examples/standalone.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
```
ThibodeauJF, line 1: Test it out! It works pretty well, run dev & http://localhost:3333/examples/standalone.html

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -68,28 +72,43 @@ export class AtomicSearchBox {
    */
   @Prop() public numberOfQueries = 8;
 
+  /**
+   * Defining this options makes the search box standalone.
+   *
+   * The default Url the user should be redirected to, when a query is submitted.
+   * If a query pipeline redirect is triggered, it will redirect to that Url instead.
+   */
+  @Prop() public redirectionUrl?: string;
```
ThibodeauJF, line 81: I decided to adapt the current search box, it doesn't add much complexity since the controllers have the same interface. The complexity would be greater in copying & sharing code in between 2 component. 

For a user of the library, the setup is basically the same for both and the search box behaves the same, setting a simple prop handles everything in an elegant manner.

packages/atomic/src/components.d.ts
```diff
@@ -642,6 +642,10 @@ export namespace Components {
           * The amount of queries displayed when the user interacts with the search box. By default, a mix of query suggestions and recent queries will be shown. You can configure those settings using the following components as children:  - atomic-search-box-query-suggestions  - atomic-search-box-recent-queries
          */
         "numberOfQueries": number;
+        /**
+          * Defining this options makes the search box standalone.  The default Url the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that Url instead.
```
jpmarceau, line 646:
```suggestion
          * Defining this option makes the search box standalone. This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
```

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -99,6 +118,25 @@ export class AtomicSearchBox {
...
+    if (redirectTo) {
+      const data = {value, analytics};
+      const storage = new SafeStorage();
+      storage.setItem(
+        StorageItems.STANDALONE_SEARCH_BOX_DATA,
+        JSON.stringify(data)
+      );
+
+      window.location.href = redirectTo;
+    }
```
btaillon, line 137: I slightly prefer not having a `return` in a lifecycle method (or a constructor). Perhaps this should be extracted into a separate methods?

E.g.:
```ts
if (this.isStandaloneSearchBox()) {
  const {redirectTo, value, analytics} = this.searchBoxState;
  if (redirectTo) {
    this.redirectTo(redirectTo, value, analytics);
  }
}
```

jpmarceau (approved):

btaillon (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
ThibodeauJF: I would agree if we used the lifecycle method for other purposes, but here what would it change?
```
componentDidUpdate() {
 redirect();
}
```

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -235,7 +242,34 @@ export class AtomicSearchInterface {
...
+    );
+
+    const {value, analytics} = standaloneSearchBoxData;
+    const {cause, metadata} = analytics;
+
+    this.engine!.dispatch(updateQuery({q: value}));
+
+    return cause === 'searchFromLink'
+      ? logSearchFromLink()
+      : logOmniboxFromLink(metadata!);
```
samisayegh, line 272: This code can be removed if using `executeFirstSearchAfterStandaloneSearchBoxRedirect`. @jpmarceau, perhaps we can update the guide too?

packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
samisayegh: We could adjust the type to only be a `string` here too? Starting to send `null` could be breaking if someone is checking for an empty string.

e.g.

```
if (redirectTo === '') {
  // do nothing
}
```

samisayegh (approved):

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
ThibodeauJF: Oh right forgot about that method! :D

packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
ThibodeauJF: Im afraid that's breaking to change the type as well, or misleading? Before the new slice, I think we actually returned null?

ThibodeauJF: We would also need to change the usage doc to check of an empty string

samisayegh: The old slice was indeed sending both values, whereas the new one just returns `string`. The null type is removed inside `fetchRedirectUrl`.

Constraining a return type is not breaking because a caller is expected to be handling all return types. Ofc, if someone is handling `null` and not empty string, there would be an impact, but I think most are doing:

```
if (redirectionUrl) {
  // do nothing
}
```

Some reasons to remove `null`:
- The less return types to check, the easier it is for a consumer.
- If someone configures the `redirectionUrl` option an empty string, it feels more correct to return an empty string rather than `null`.
- One less ternary.

> We would also need to change the usage doc to check of an empty string

Which part of the doc would need to change?

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
olamothe: I guess that's fine ... 🤔   The internal complexity is not really what interest me though.

My main concern is if someone searches for "how to create a standalone search box atomic", will they discover/stumble on that prop on their own ?

I think it's a fair assumption to say that they would look for properties in the search box component and  ultimately discover this one relatively easily.

If we had a unique component for this, however, I think it might be a bit more easily discoverable overall.

If you scanned the list of all components in our doc, and saw two different search box, it might be more easy to understand that there's two different "type" of search box that behave a bit differently.

However with all that said, I think it's fine to leave it like you have done it here.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -68,28 +72,44 @@ export class AtomicSearchBox {
...
-          notMatchDelimiters: {
-            open: '<span class="font-bold">',
-            close: '</span>',
-          },
-          correctionDelimiters: {
-            open: '<span class="font-normal">',
-            close: '</span>',
-          },
+
+    const searchBoxOptions = {
```
olamothe, line 91: One problem with this is for SinglePageApp. 

Users might want to have the same component on their home page as their full search page (eg: in the header of the website), but changing (adding/removing) the `redirectionUrl` property depending on which page is currently displayed, so that the search box behave differently.

By using the `initialize()` function I think we forfeit that possibility ?
Could we add a `@Watch()` decorator, which rebuild the controller ?

packages/atomic/src/utils/local-storage-utils.tsx
```diff
@@ -1,5 +1,13 @@
+import {StandaloneSearchBoxAnalytics} from '@coveo/headless';
+
 export enum StorageItems {
   RECENT_QUERIES = 'coveo-recent-queries',
+  STANDALONE_SEARCH_BOX_DATA = 'coveo_standalone_search_box_data',
```
olamothe, line 5: Use `-` instead of `_` to use the same pattern as the other entry.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -99,6 +119,25 @@ export class AtomicSearchBox {
...
+    if (!('redirectTo' in this.searchBoxState)) {
+      return;
+    }
+
+    const {redirectTo, value, analytics} = this.searchBoxState;
+
+    if (redirectTo) {
+      const data = {value, analytics};
+      const storage = new SafeStorage();
+      storage.setItem(
```
olamothe, line 132: Use `setJSON`, and let the storage handle the `JSON.stringify` (because any `JSON.stringify` can throw).

olamothe (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
ThibodeauJF: I'd add a usage page with more details on how to set it up too

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
ThibodeauJF: ✅

PR #1690, opened 2022-01-21
fix(atomic): made screen readers skip colon in "filters" and "sort by"

btaillon:
https://coveord.atlassian.net/browse/KIT-1087

In the case of "filter:" at-least, the colon (`:`) symbol being inside a `::after` caused screen readers to select it individually. In the first commit, I extracted the colon symbol out of `::after` blocks.

In the second commit (so I can revert it if you guys think it's too out-of-scope), I did the same thing for parentheses (which are notably used in every facet) and quotation marks. I didn't spot accessibility issues with those, these changes are mostly preventive. As a bonus, those are all localized, now.

### DIFF
--- a/packages/atomic/src/locales.json
+++ b/packages/atomic/src/locales.json
@@ -1129,5 +1129,13 @@
   "with-colon": {
     "en": "{{text}}:",
     "fr": "{{text}}\u00A0:"
+  },
+  "between-quotations": {
+    "en": "\"{{text}}\"",
+    "fr": "«\u00A0{{text}}\u00A0»"
+  },
+  "between-parentheses": {
+    "en": "({{text}})",
+    "fr": "(\u00A0{{text}}\u00A0)"
   }
 }
--- a/packages/atomic/src/locales.json
+++ b/packages/atomic/src/locales.json
@@ -1125,5 +1125,9 @@
   "stars-range": {
     "en": "{{value}} to {{count}} stars",
     "fr": "{{value}} à {{count}} étoiles"
+  },
+  "with-colon": {
+    "en": "{{text}}:",
+    "fr": "{{text}}\u00A0:"
   }
 }
--- a/packages/atomic/cypress/integration/breadbox/breadbox-assertions.ts
+++ b/packages/atomic/cypress/integration/breadbox/breadbox-assertions.ts
@@ -77,7 +77,7 @@ export function assertCategoryPathInBreadcrumb(path: string[]) {
 
   it(`should display the selected path "${joinedPath}" in the breadcrumbs`, () => {
     BreadboxSelectors.breadcrumbButton().contains(
-      `${categoryFacetLabel}${joinedPath}`
+      `${categoryFacetLabel}:${joinedPath}`
     );
   });
 }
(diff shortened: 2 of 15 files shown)


packages/atomic/src/locales.json
```diff
@@ -1125,5 +1125,17 @@
   "stars-range": {
     "en": "{{value}} to {{count}} stars",
     "fr": "{{value}} à {{count}} étoiles"
+  },
+  "with-colon": {
+    "en": "{{text}}:",
+    "fr": "{{text}}\u00A0:"
```
btaillon, line 1131: `\u00A0` is a non-breaking space (`&nbsp;`).

I added them because I recall seeing spaces being added before colons, quotation marks and parentheses often in French, but I'm uncertain whether they belong here.

npushkarskii: @btaillon that's a quite intriguing thing to discuss :) 

As I know, there's no such a rule in Canadian French. 
On the contrary, there IS a rule about non-breaking spaces in European French. The rule says there should be a non-breaking-space BEFORE and AFTER the following punctuation marks:
- exclamation marks
- question marks
- semicolons
- colons
- the percentage mark
- currency symbols
- the hash
- the guillemet

So, I'd like to summon @jpmarceau here so he can ask my question: do we use Canadian French or European French in our public texts (UI, docs, etc.)? (_I know for sure that we use the American version of English, but not aware about French, unfortunately._)

If we are using Canadian French, we don't need non-breaking spaces at all.
If we are using European French, we do need non-breaking spaces.

jpmarceau: I'm not sure about which French we use exactly, but I did a bit of research and it seems that even here in Quebec we put an insecable whitespace before colons http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?id=2039
So I think this is fine 👍

btaillon: Wow! I'm saving that link, very useful table.

npushkarskii: @jpmarceau thx for research! Doesn't it mean that we need adding the second non-breaking space after the colon?

packages/atomic/src/locales.json
```diff
@@ -1125,5 +1125,17 @@
...
+  },
+  "between-quotations": {
+    "en": "\"{{text}}\"",
+    "fr": "«\u00A0{{text}}\u00A0»"
+  },
+  "between-parentheses": {
+    "en": "({{text}})",
+    "fr": "(\u00A0{{text}}\u00A0)"
```
npushkarskii, line 1139:
```suggestion
    "fr": "{{text}}\u00A0: "
  },
  "between-quotations": {
    "en": "\"{{text}}\"",
    "fr": " «\u00A0{{text}}\u00A0» "
  },
  "between-parentheses": {
    "en": "({{text}})",
    "fr": " ({{text}}) "
```

According to the link that @jpmarceau shared with us: http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?id=2039

npushkarskii (approved):
LGTM, just a bit of edits on spaces and non-breaking spaces 👍

packages/atomic/src/locales.json
btaillon: No need, the extra spaces are already present in the UI. But thanks!

samisayegh (approved):

ThibodeauJF (approved):

PR #1693, opened 2022-01-21
fix(atomic): gave main role to refine modal's content

btaillon:
https://coveord.atlassian.net/browse/KIT-1212

When focusing on the "Sort" heading with TalkBack, it would read out "Sort, InIn". You can see a video of it in the JIRA above.

### DIFF
--- a/packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
+++ b/packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
@@ -271,10 +271,10 @@ export class AtomicRefineModal implements InitializableComponent {
           <hr class="border-neutral"></hr>
           <div class="overflow-auto px-6 flex-grow">
             <div class="adjust-for-scroll-bar">
-              <main class="centered">
+              <aside class="centered">
                 {this.renderSort()}
                 {this.renderFilters()}
-              </main>
+              </aside>
             </div>
           </div>
           {this.renderFooter()}
--- a/packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
+++ b/packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
@@ -271,10 +271,10 @@ export class AtomicRefineModal implements InitializableComponent {
           <hr class="border-neutral"></hr>
           <div class="overflow-auto px-6 flex-grow">
             <div class="adjust-for-scroll-bar">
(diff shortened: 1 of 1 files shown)


ThibodeauJF:
> A document mustn't have more than one <main> element that doesn't have the hidden attribute specified.

Is there a possible incompatibility with some customer's pages?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

I think it's a similar problem that we had to face with landmark

btaillon:
> > A document mustn't have more than one
> > 
> > element that doesn't have the hidden attribute specified.
> 
> Is there a possible incompatibility with some customer's pages?
> [developer.mozilla.org/en-US/docs/Web/HTML/Element/main](developer.mozilla.org/en-US/docs/Web/HTML/Element/main)
> 
> I think it's a similar problem that we had to face with landmark

The modal inherently has a risk of being incompatible with a customer's page since it must hide every element outside of it from the tab order and hide them all from the accessibility DOM by giving them `aria-hidden`.

We are already doing this, and since `aria-hidden` removes an element and its descendants from the accessibility DOM, there should not, in theory, be any other `main` element in the page.

olamothe:
I don't think it's a good idea to use `main` here... If you read the description from MDN : 

> The main HTML element represents the dominant content of the <body> of a document. The main content area consists of content that is directly related to or expands upon the central topic of a document, or the central functionality of an application.

I don't think it's fair to say that the refine modal of the search page is "the central functionality of an application"...

Also, even if everything else is hidden, it still causes problem for the validity of the document, if there's more than one main tag present in the page.

packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -271,10 +271,10 @@ export class AtomicRefineModal implements InitializableComponent {
           <hr class="border-neutral"></hr>
           <div class="overflow-auto px-6 flex-grow">
             <div class="adjust-for-scroll-bar">
-              <div class="centered">
+              <main class="centered">
                 {this.renderSort()}
                 {this.renderFilters()}
-              </div>
+              </main>
```
samisayegh, line 277: As an alternative, [this post](https://stackoverflow.com/questions/3072597/semantically-accurate-html5-element-for-a-modal-dialog) suggests using `aside` elements for modals.

samisayegh (approved):

PR #1695, opened 2022-01-24
chore: generate atomic files and commit on PR

olamothe:
Adds a new action `commit-generated-files` which will run on each PR, after the build step is performed.

It will run only if git status shows there are file changes after building (`git status --porcelain`), and then only commit matching the pattern for some Atomic files (definitions + Atomic React generated files).

I've opted to only add and commit on files matching a specific pattern since it seems less dangerous that it would start to commit random junk.

We will be able to add more pattern match there if needed in the future.

https://coveord.atlassian.net/browse/KIT-1368


ThibodeauJF (approved):

.github/workflows/prbot.yml
```diff
@@ -13,25 +13,27 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-node@v2
         with:
-          node-version-file: '.nvmrc'
+          node-version-file: '.n\vmrc'
```
samisayegh, line 16: Is this back-slash intended?

.github/actions/commit-generated-files/action.yml
```diff
@@ -0,0 +1,18 @@
...
+  using: composite
+  steps:
+    - run: |
+        if [ ! -z "$(git status --porcelain)" ]; then  
+          git fetch --depth=1
+          git checkout ${{ github.head_ref }}
+          git add 'packages/atomic/**/*.d.ts' 'packages/atomic-react/src/components/stencil-generated/*'
+
+          git config --global user.name "GitHub Actions Bot"
+          git config --global user.email "<>"
```
samisayegh, line 13: Is this git configuration happening elsewhere? If so, can it be centralized in one place?

.github/actions/commit-generated-files/action.yml
```diff
@@ -0,0 +1,18 @@
+name: 'Commit generated files'
+description: 'Automatically commit generated files'
+runs:
+  using: composite
+  steps:
+    - run: |
+        if [ ! -z "$(git status --porcelain)" ]; then  
+          git fetch --depth=1
+          git checkout ${{ github.head_ref }}
+          git add 'packages/atomic/**/*.d.ts' 'packages/atomic-react/src/components/stencil-generated/*'
```
samisayegh, line 10: What would happen if `git status --porcelain` returned files, but they did not match these paths? Would this lead to empty commits being pushed? errors?

samisayegh (approved):

.github/actions/commit-generated-files/action.yml
olamothe: Pretty much "nothing".

If you looked at the output in github action you would see something like `No staged files found, nothing to commit working tree clean`

But no errors, and no empty commit.

.github/actions/commit-generated-files/action.yml
olamothe: I just copied what is done for bump version on master.

What is in that config is not really important, just needs to be evident to a dev looking at their PR that "some github bot did commit some stuff for me".

PR #1696, opened 2022-01-25
chore(atomic): add publish official script

olamothe:
https://coveord.atlassian.net/browse/KIT-1374

### DIFF
--- a/packages/atomic-react/package.json
+++ b/packages/atomic-react/package.json
@@ -14,7 +14,6 @@
     "clean": "rimraf -rf dist",
     "compile": "npm run tsc",
     "tsc": "tsc -p .",
-    "npm:publish": "node ../../scripts/deploy/publish.js",
     "npm:publish:alpha": "node ../../scripts/deploy/publish.js alpha",
     "copy:assets": "ncp node_modules/@coveo/atomic/dist/atomic/assets dist/assets && ncp node_modules/@coveo/atomic/dist/atomic/lang dist/lang "
   },
--- a/packages/atomic-react/package.json
+++ b/packages/atomic-react/package.json
@@ -14,6 +14,7 @@
     "clean": "rimraf -rf dist",
     "compile": "npm run tsc",
     "tsc": "tsc -p .",
+    "npm:publish": "node ../../scripts/deploy/publish.js",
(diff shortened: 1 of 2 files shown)


ThibodeauJF (approved):

packages/atomic-react/package.json
```diff
@@ -14,6 +14,7 @@
     "clean": "rimraf -rf dist",
     "compile": "npm run tsc",
     "tsc": "tsc -p .",
+    "npm:publish": "node ../../scripts/deploy/publish.js",
```
btaillon, line 17: don't forget to add `atomic-react` to `update-npm-tag.js`

btaillon, line 17: This publishes as `latest`, is this the intended effect? If you want to publish as alpha, you can make the below change and also add `atomic-react` to `update-npm-tag.js`:
```suggestion
    "npm:publish": "node ../../scripts/deploy/publish.js alpha",
```

samisayegh: Indeed, though I do see the [package on npm](https://www.npmjs.com/package/@coveo/atomic-react) tagged as alpha? I wonder how that's happening,

Also, I think you'll need to update the array in the `update-npm-tag.js` script to change the tag to latest,

samisayegh (approved):

packages/atomic-react/package.json
olamothe: There's the alpha script just below.

Here, I want to have the release "official", or latest.

samisayegh: Why not promote the `alpha` version to `latest`? I think if you publish a version a second time, npm will error and say it's already published.

olamothe: Isn't this the exact same mechanism as all other packages, or I'm just terribly confused here ?

samisayegh: It is not the same mechanism (please correct me if wrong). I believe publishing directly as latest is an artifact from the early release process that used multiple versions to move a package from dev -> prd (prerelease -> graduate).

The current mechanism publishes as alpha, and then promotes to `beta` in qa and to `latest` in prd.

olamothe: Alright, I'll remove the publish script and just add the tag in `update-npm-tag.ts`

PR #1697, opened 2022-01-25
feat(atomic-angular): initial project setup for atomic angular 

olamothe:
So, yeah, a bunch of generated files mostly. 
I followed the guide on Stencil website, which suggested using `ng cli` to bootstrap the project, and then add `@stencil/angular-output-target`, and create all the generated files in `packages/atomic-angular`.

I had to create a patch (`patch-package`) for a problem with the `atomic/scrollToTop` event in the pager: The generated file contains an interface with an invalid property otherwise.

After this initial setup, I'll start working on testing all the components, similar to what I did with a "full commerce page" for React output.

https://coveord.atlassian.net/browse/KIT-1376


.nvmrc
```diff
@@ -1 +1 @@
-v16.5.0
```
olamothe, line 1: Needed by `ng cli`, which builds the project.

samisayegh (approved):

ThibodeauJF (approved):

PR #1698, opened 2022-01-26
feat(atomic): layouts POC

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1381

View the [document](https://docs.google.com/document/d/19_0rVe7txEDiExVS80vPMYPkiIm0rPZsskgEhOki2O8/edit) for more details!


packages/atomic/src/components/atomic-layout/atomic-layout.pcss
```diff
@@ -0,0 +1,27 @@
+@import './search-layout.pcss';
```
ThibodeauJF, line 1: Here is all the CSS that is "static" meaning that does not need to adapt whether some sections are present are not.
In our current search layout, 3 things are dynamic:
- the facets, we don't display an empty column if they are not defined. In the case where some facets are defined but return no results, I think it's better to keep the column not to have a jittery interface
- the mobile media query, we can't use a CSS variable for it unfortunately, which is why it's dynamically generated
- the refine toggle/modal which hides the sort dropdown when in mobile view

packages/atomic/src/components/atomic-layout/atomic-layout.tsx
```diff
@@ -0,0 +1,33 @@
+import {Component, Element, Prop} from '@stencil/core';
```
ThibodeauJF, line 1: I made a reusable component here too, so we could switch the layout dynamically using the Prop.
I am also open to make an atomic-search-layout, atomic-agent-panel-layout, etc. 🤷

packages/atomic/src/components/atomic-layout/search-layout.ts
```diff
@@ -0,0 +1,76 @@
+import {findSection, sectionSelector} from '../atomic-layout-section/sections';
```
ThibodeauJF, line 1: Probably a controversial file, it's still POC, so it's not necessary to review the details now, but I think it's necessary to use some JS to generate some of the dynamic conditional CSS. 

Notice the unique id that is being generated & used, to allow for potential multiple layout in the same page.

Here's example of what is being generated:
```css
@media only screen and (min-width: 1024px) {
  atomic-layout[layout='search']#atomic-layout-ztr8e {
    grid-template-areas:
      '. .                     atomic-section-search-box .'
      '. atomic-section-facets atomic-section-main       .'
      '. atomic-section-facets .                         .';
  }
  atomic-layout[layout='search']#atomic-layout-ztr8e {
    grid-template-columns: 1fr minmax(272px, 350px) minmax(656px, 1100px) 1fr;
  }
  atomic-layout[layout='search']#atomic-layout-ztr8e {
    column-gap: var(--atomic-layout-spacing-x);
  }
  atomic-layout[layout='search']#atomic-layout-ztr8e
    atomic-layout-section[section='facets'] {
    display: block;
  }
}
atomic-layout[layout='search']#atomic-layout-ztr8e
  atomic-layout-section[section='top-bar']
  atomic-sort-dropdown {
  display: none;
}
@media only screen and (min-width: 1024px) {
  atomic-layout[layout='search']#atomic-layout-ztr8e
    atomic-layout-section[section='top-bar']
    atomic-sort-dropdown {
    display: block;
  }
}
@media only screen and (min-width: 1024px) {
  atomic-layout[layout='search']#atomic-layout-ztr8e
    atomic-layout-section[section='top-bar']
    atomic-refine-toggle {
    display: none;
  }
}
```

packages/atomic/src/components/atomic-query-summary/atomic-query-summary.tsx
```diff
@@ -133,10 +142,7 @@ export class AtomicQuerySummary implements InitializableComponent {
 
     return (
       <div class="text-on-background" part="container">
-        {this.querySummaryState.hasResults && [
-          this.renderHasResults(),
-          this.renderDuration(),
-        ]}
+        {this.querySummaryState.hasResults && this.renderHasResults()}
```
ThibodeauJF, line 145: Visual glitch I encountered trying to render long queries with an ellipsis (I'll probably open an official PR for this)

packages/atomic/src/components/atomic-layout/atomic-layout.pcss
```diff
@@ -0,0 +1,27 @@
...
+
+atomic-layout {
+  width: 100%;
+  height: 100%;
+  display: grid;
+}
+
+atomic-layout-section[section='search-box'] {
```
btaillon, line 9: Although I like the `section=` approach too, I would prefer if we used `atomic-layout-section-search-box` and so on instead, since this is the pattern we use for result template sections.

packages/atomic/src/components/atomic-layout/atomic-layout.tsx
btaillon: I like having the `layout` prop here. Being able to change it dynamically means I'll be able to test locally using a simple tampermonkey script.

packages/atomic/src/components/atomic-layout/search-layout.ts
olamothe: So, I think there's a way to make this "clean", as far as the code is concerned. 

After all it's mostly string templating, and there's many way it could be done that it would be more easily maintainable/readable. So, following your comment, not focusing on the implementation details.

This will create a dynamic style tag, is downloaded at the same time as the component JS code (lazy mode).

Does this mean that the layout component will end up "flickering" the page a lot when the end user is on a slow connection, and it takes 0.5 sec to download the styles from a JS file, instead of having them pre-applied at page load like a "normal" style tag ?

How do you see we can tackle this kind of problem, and force the layout to be applied initially correctly on page load ?

packages/atomic/src/components/atomic-layout/atomic-layout.tsx
```diff
@@ -0,0 +1,33 @@
...
+@Component({
+  tag: 'atomic-layout',
+  styleUrl: 'atomic-layout.pcss',
+  shadow: false,
+})
+export class AtomicLayout {
+  @Element() private host!: HTMLElement;
+
+  @Prop({reflect: true}) public layout: Layout = 'search';
+  @Prop() public mobileBreakpoint = '1024px';
```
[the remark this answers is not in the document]
btaillon: If we do this here, we should probably also modify result templates, since they use `@screen desktop-only` and `@screen mobile-only`.

packages/atomic/src/components/atomic-query-summary/atomic-query-summary.tsx
```diff
@@ -66,6 +67,10 @@ export class AtomicQuerySummary implements InitializableComponent {
     return `<span class="font-bold" part="highlight">${content}</span>`;
   }
 
+  private wrapQuery(query: string) {
+    return `<span title="${query}" class="font-bold truncate" part="highlight query">${query}</span>`;
```
btaillon, line 71: Make sure to test with a screen reader (or multiple) that this `title` doesn't reduce accessibility.

packages/atomic/src/components/atomic-layout/atomic-layout.pcss
ThibodeauJF: alternative is to make the no results image take both columns, but the search box shouldn't move

packages/atomic/src/components/atomic-layout/atomic-layout.tsx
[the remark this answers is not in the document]
ThibodeauJF: it's fine if both aren't exactly the same initially, but eventually making the result template have dynamic breakpoints following the layout would be pretty cool

packages/atomic/src/components/atomic-layout/atomic-layout.pcss
ThibodeauJF: result template ones are more opinionated and have different features/behaviour. Here it's only to identify a section with common/generic options. It would be a lot more components for few benefits.

packages/atomic/src/components/atomic-layout/search-layout.ts
ThibodeauJF: I don't think there will be a race condition issue because of the lifecycle methods:
-connectedCallback (here is where out layout component injects CSS
-componentWillLoad (bindings are being fetched)
-render (blocked until bindings are available)

I tested it will a throttled connection and prod mode with `serve` and it works fine

ThibodeauJF: maybe if customers put something that is "not Stencil" in their layout as well. In this case, like we have the `atomic-search-interface-no-results` or the `hydrated` class on the search interface, we could have a new class. But I don't think it will be necessary

PR #1699, opened 2022-01-27
fix(auth): send organizationId as query string param during handshake

samisayegh:
In collaboration with the searchapi team, we wanted to remove the need to send an organizationId during the SAML handshake.
The goal was to make the handshake token necessary and sufficient.

However, [it turns out](https://coveord.atlassian.net/browse/SEARCHAPI-6571) that this is complex to change in the service.

This PR adjusts to send the org id during handshake.

https://coveord.atlassian.net/browse/KIT-1382

### DIFF
--- a/packages/auth/src/saml/saml-flow.ts
+++ b/packages/auth/src/saml/saml-flow.ts
@@ -33,18 +33,21 @@ export function buildSamlFlow(config: SamlFlowOptions): SamlFlow {
     },
 
     async exchangeHandshakeToken() {
-      const {location, history, request} = options;
+      const {organizationId, location, history, request} = options;
       const handshakeToken = getHandshakeToken(location);
       removeHandshakeToken(location, history);
 
       try {
-        const response = await request(`${api}/handshake/token`, {
-          method: 'POST',
-          body: JSON.stringify({handshakeToken}),
-          headers: {
-            'content-type': 'application/json; charset=UTF-8',
-          },
-        });
+        const response = await request(
+          `${api}/handshake/token?organizationId=${organizationId}`,
+          {
+            method: 'POST',
+            body: JSON.stringify({handshakeToken}),
+            headers: {
+              'content-type': 'application/json; charset=UTF-8',
+            },
+          }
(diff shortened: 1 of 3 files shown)


ThibodeauJF (approved):

dreisch-coveo:
thanks (and sorry). what is the impact for the ongoing migration? do we want to backport this to the older patch of jsui we did or is there some other way? or are there not any more such clients remaining?

packages/auth/src/saml/saml-flow.ts
```diff
@@ -33,18 +33,21 @@ export function buildSamlFlow(config: SamlFlowOptions): SamlFlow {
...
       try {
-        const response = await request(`${api}/handshake/token`, {
-          method: 'POST',
-          body: JSON.stringify({handshakeToken}),
-          headers: {
-            'content-type': 'application/json; charset=UTF-8',
-          },
-        });
+        const response = await request(
+          `${api}/handshake/token?organizationId=${organizationId}`,
```
olamothe, line 42: You could `encodeURIComponent` the orgId for good measure.

However I think all org ID are url safe characters anyway, so that's kind of a belt and suspenders situation 🤷 

It's more a reflex at this point hehe.

olamothe (approved):

btaillon (approved):

samisayegh:
> thanks (and sorry). what is the impact for the ongoing migration? do we want to backport this to the older patch of jsui we did or is there some other way? or are there not any more such clients remaining?

There's no impact on the migration because the JSUI sends the `organizationId`. This package is to help headless/atomic SAML implementations!

dreisch-coveo:
ahhh so you're forced to keep moar legacy in the shiny new thing :(

samisayegh:
Yea, when writing new stuff, I start by avoiding porting legacy things if possible, opting to try to improve at the api level so it's available to everyone. If it's too complex though, then at least we can say we tried :). In this case, most people should be happy to to use an api client, so these nuances will be abstracted away.

PR #1700, opened 2022-01-27
feat(atomic-angular): work on minimal output needed for a local test project

olamothe:
Very small PR so far, but this is what I've been using to debug/troubleshoot a random local project.

I'll eventually add a full project in the sample package, like discussed this morning.

https://coveord.atlassian.net/browse/KIT-1384


samisayegh (approved):

packages/atomic-angular/projects/atomic-angular/src/lib/atomic-angular.module.ts
```diff
@@ -0,0 +1,34 @@
...
+import {ModuleWithProviders, NgModule} from '@angular/core';
+
+import {
+  AtomicSearchInterface,
+  AtomicSearchBox,
+  AtomicResultList,
+  AtomicLoadMoreResults,
+} from './stencil-generated/components';
+
+import {defineCustomElements} from '@coveo/atomic/loader';
```
olamothe, line 11: This is to work around an issue with the stencil angular output, as described here https://github.com/ionic-team/stencil-ds-output-targets

packages/atomic-angular/projects/atomic-angular/src/lib/atomic-angular.module.ts
```diff
@@ -0,0 +1,34 @@
...
+defineCustomElements(window);
+
+const DECLARATIONS = [
+  AtomicSearchInterface,
+  AtomicSearchBox,
+  AtomicResultList,
+  AtomicLoadMoreResults,
+];
+
+@NgModule({
```
olamothe, line 21: So that's the thing that can workaround the problem with angular ivy:.

For example doing this in the consuming app: 
```
import {AtomicSearchInterface, AtomicSearchBox} from '@coveo/atomic-angular';

@NgModule({
  declarations: [AppComponent],
  imports: [BrowserModule, AppRoutingModule, AtomicSearchInterface, AtomicSearchBox],
  providers: [],
  bootstrap: [AppComponent],
})
export class AppModule {}
```
Then we get errors like this at runtime: 

```
Error: ../ui-kit/packages/atomic-angular/dist/atomic-angular/lib/stencil-generated/components.d.ts:478:22 - error NG6002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class.

This likely means that the dependency (@coveo/atomic-angular) which declares AtomicSearchInterface has not been processed correctly by ngcc.

478 export declare class AtomicSearchInterface {
```

However, in the consuming app if we do this: 
```
import { AtomicAngularModule } from '@coveo/atomic-angular';
@NgModule({
  declarations: [AppComponent],
  imports: [BrowserModule, AppRoutingModule, AtomicAngularModule],
  providers: [],
  bootstrap: [AppComponent],
})
export class AppModule {}
```

Then it works correctly. The catch is that we need to have a static declaration in the `AtomicAngularModule`. 

I think we can workaround this by changing the angular output plugin to generate that array at compile time, so that we don't need to manually update it.


And then, apparently the best practice in angular world is a Single Component Angular Module (SCAM), which means the best would be for us to have one module for each component, instead of one module that contains every component : https://angular-training-guide.rangle.io/modules/module-scam-pattern#what-is-the-scam-single-component-angular-module-pattern-in-angular


With that optic, I'm thinking of submitting a PR to the stencil team to add the boiler plate needed to the generated files.

packages/atomic-angular/projects/atomic-angular/tsconfig.lib.json
```diff
@@ -6,10 +6,8 @@
     "declaration": true,
     "declarationMap": true,
     "inlineSources": true,
-    "types": []
+    "types": [],
+    "paths": {"@coveo/atomic-angular": ["../atomic-angular/*"]}
```
olamothe, line 10: That's to workaround weird symlink issue in local.

packages/atomic-angular/projects/atomic-angular/src/lib/atomic-angular.module.ts
olamothe: I'm also suspecting a weird symlink issue with Angular, which might not require us to go a too complicated approach 🤞

btaillon (approved):

PR #1701, opened 2022-01-28
fix(headless, quantic): Prevent search analytics event when using QV in case assist

mmitiche:
[SFINT-4371](https://coveord.atlassian.net/browse/SFINT-4371)
### Headless side:
- New Core controller `headless-core-quickview` created,  this controller contain all the logic for building a quickview, It just doesn't contain any analytics logic.
-  This core controller is then wrapped by headless-quickview, which is the the controller for the quickview for search,  adding to this core the analytics logic needed.
- In the Case assist use case, the quickview component will use the core controller to give us the ability so send the right analytics using the case assist engine.
- New case assist analytics action created `logQuickviewDocumentSuggestionClick`, this action will send a `suggestion_click` event but with an additional information indicating that the click was made to quickview the document suggestion.

### Quantic side:
- New property `preventAnalytics` added to `quanticResultQuickview` in Quantic. a new property that when set to true will  prevent sending the  document quickview click event and this by using the `headless-core-quickview` controller to build the quickview.
- New event `quickview_opened` dispatched  in `quanticResultQuickview` when the quickview is opened.
- New event listener added in `quanticDocumentSuggestion` that will listen to `quickview_opened`.  when receiving  such event `quanticDocumentSuggestion` will call `logDocumentSuggestionClick` with the parameter `fromQuickview` set true in order to send the correct `suggestion_click` event 

https://user-images.githubusercontent.com/86681870/151577487-76cbda84-6f4a-4d64-8506-fc2b9aaed2f2.mov


packages/headless/src/controllers/quickview/headless-quickview.test.ts
```diff
@@ -47,15 +47,15 @@ describe('Quickview', () => {
     expect(quickview.subscribe).toBeTruthy();
```
samisayegh, line 47: > In order to prevent sending analytics search events when using quanticResultQuickview is a case assist use case which needs different analytics events to be sent I did the following things:

Hi Mic, when a use-case needs a different behaviour, it's best to create a dedicated controller. If there is a lot of common logic, the core of the current quickview controller can be extracted and reused.

The issue with adding a parameter is it shifts the burden to implementers, who need to know/read documentation to know what to pass. We want to minimize the chance of errors, where someone sets `preventAnalytics` to `false` for a case-assist implementation, or `true` for a `search` implementation.

packages/headless/src/controllers/quickview/headless-quickview.ts
```diff
@@ -33,7 +33,7 @@ export interface Quickview extends Controller {
   /**
    * Retrieves the preview content for the configured result.
    */
-  fetchResultContent(): void;
+  fetchResultContent(preventAnalytics: boolean): void;
```
samisayegh, line 37: We cannot add a required param because it will be a breaking change to devs upgrading.

Suggestion:
- Rename this controller to`headless-core-quickview` and remove the analytics
- Create a wrapper for search that sends the search analytics.
- Create a wrapper for case-assist that sends the case-assist analytics.

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -68,12 +68,19 @@ export const logClassificationClick = (classificationId: string) =>
       })
   )();
 
-export const logDocumentSuggestionClick = (suggestionId: string) =>
+export const logDocumentSuggestionClick = (
+  suggestionId: string,
+  fromQuickview = false
```
samisayegh, line 73: Curious why do we need to distinguish between fromQuickview or otherwise? I'm guessing from an analytics/ml perspective, this detail is important?

If it is, could it be clearer if we had a separate action? e.g. `logQuickviewDocumentSuggestionClick`

I'm also not clear on when this should be used:
- Is it when some opens the quickview?
- Is it when some clicks the document title inside an open quickview?

Separate action would allow for dedicated documentation to explain,

packages/headless/src/controllers/quickview/headless-quickview.test.ts
nathanlb: That's a good point. [headless-core-interactive-result](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/controllers/core/interactive-result/headless-core-interactive-result.ts) is an example of that pattern.

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
mmitiche: Basically we want to distinguish between when the the user clicks on the title of a document suggestion and when the user clicks on the quickview button of a document suggestion, you guessed it right from an analytics/ml perspective we want to differentiate  the two events, a differentiation which was considered important in a case assist use case but you absolutely right having a separate action for the event related to the quickview will be much better 👌

packages/headless/src/controllers/quickview/headless-quickview.ts
mmitiche: Thanks for the suggestion, I'll update my code accordingly👌

mmitiche: I applied your suggestion and created the headless-core-quickview and the wrapper that will be used when building a quickview for search and that will take care of sending the search analytics, but I wasn't able to create a same wrapper for case-assist cause to send the case-assist analytics we need the case assist engine and in the quickview controller we build the controller using a search engine and not a case-assist one.
The reason why the case assist analytic actions need to be dispatched using a case assist engine is because in this actions we need to gather some data from the state of the case assist engine. for example when we want to send a suggestion_click event, which is sent when clicking on a document suggestion, we dispatch the action `logDocumentSuggestionClick` with one parameter which is the id of this suggestion,  in the body of this action we need to retrieve all the data of this suggestion specified with this id parameter in order to send the data of this suggestion in the analytics, and we retrieve this data from the state of the case assist engine. so when dispatching this same action with a search engine the document suggestion will never be found since the state of the search engine doesn't contain any document suggestions.
So now when using the quickview in a case-assist use case I'm building the quickview with the headless-core-quickview, to avoid sending any analytics, and then I'm handling sending the right analytics using the case assist engine from the quantic side.
What do u think about that @samisayegh ?

samisayegh: What if we adjust the type of the core quickview controller to accept a `CoreEngine`? The dev facing wrapper controller would continue to use `CaseAssistEngine`. Logging this event under the hood would be nice 🤞

packages/quantic/cypress/integration/document-suggestion/document-suggestion-selectors.ts
```diff
@@ -12,6 +12,7 @@ export interface DocumentSuggestionSelector extends ComponentSelector {
   noSuggestionsMessage: () => CypressSelector;
   quickviewButton: (idx: number) => CypressSelector;
   quickviews: () => CypressSelector;
+  closeButtonQv: () => CypressSelector;
```
lbergeron, line 15: I'd suggest `quickviewCloseButton`, it seems more consistent to me.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -122,6 +124,12 @@ export default class QuanticDocumentSuggestion extends LightningElement {
     );
   };
 
+  onQvOpened = (evt) => {
```
lbergeron, line 127: I'd suggest `onQuickviewOpened` just to make it clearer.

packages/headless/src/controllers/core/quickview/headless-core-quickview.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      quickview.fetchResultContent();
+    });
+
+    it('dispatches a #fetchResultContent action with the result uniqueId', () => {
+      const action = engine.findAsyncAction(fetchResultContent.pending);
+      expect(action?.meta.arg).toEqual({uniqueId, requestedOutputSize});
+    });
+  });
+
+  it(`when configured result uniqueId matches the uniqueId in state,
```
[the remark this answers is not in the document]
mmitiche: The back-ticks  are used here because it's a multiline string.

packages/headless/src/api/search/html/html-api-client.ts
```diff
@@ -0,0 +1,62 @@
...
+import {SearchAPIErrorWithStatusCode} from '../search-api-error-response';
+import {baseSearchRequest} from '../search-api-params';
+import {PlatformClient} from '../../platform-client';
+import {pickNonBaseParams, unwrapError} from '../search-api-client';
+import {findEncoding} from '../encoding-finder';
+import {TextDecoder} from 'web-encoding';
+import {Logger} from 'pino';
+import {PreprocessRequest} from '../../preprocess-request';
+
+export interface HtmlApiClient {
```
olamothe, line 11: Curious why you decided to go with the approach of an interface `HtmlApiClient` with `getHtml` exported as a pure function, instead of the pattern used for other API clients (a static class) ?

It's not a comment about necessarily changing it by the way, just curious.

packages/headless/src/controllers/quickview/case-assist-headless-quickview.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import {
+  buildQuickviewCore,
+  QuickviewCoreOptions,
+  QuickviewCoreState,
+  QuickviewCoreProps,
+  QuickviewCore,
+} from '../core/quickview/headless-core-quickview';
+
+export interface CaseAssistQuickviewProps extends QuickviewCoreProps {}
```
olamothe, line 12: @samisayegh are those going to be documented properly by the tool ? I think so, but just making sure.

packages/headless/src/api/search/html/html-api-client.ts
olamothe: Okay, the more I'm reading, the more I'm not certain why this interface is needed.

The correct types should already  be available on `SearchApiClient` and `CaseAssistApiClient` 🤔 

That is, `CaseAssistApiClient.html()` and `SearchApiClient.html()` should already be typed properly with the return types on error and success.

And so the only the function `getHtml` present in this file and used by both client should be enough.

olamothe (approved):
Great job on the PR, that was not an easy one for a newcomer in the project !

packages/headless/src/api/search/html/html-api-client.ts
samisayegh: The motivator behind this interface was [this interface here](https://github.com/coveo/ui-kit/pull/1701/files#diff-ada24d9636c2eb702fbde2a92daf14b08c60ac06b4212b5c4ce159d8c994b437L20) being tightly coupled to the `SearchApiClient`.

The action errored when used with a `CaseAssistEngine`, which has a `CaseAssistApiClient`.

We could have solved the error with a union: `SearchApiClient | CaseAssistApiClient`
But this union will need to continue growing as new use-cases/engines are added.

If the type is scoped html client though, then the api clients of new use-cases will just need to make sure the `html` method is implemented.

One thing though, @MehdiMitiche did you explore if you could make typescript error when passing an engine with an api-client without the `html` method implemented?

packages/headless/src/api/search/search-api-client.ts
```diff
@@ -402,7 +369,7 @@ function isSearchAPIException(
   );
 }
 
-function pickNonBaseParams<Params extends BaseParam>(req: Params) {
+export function pickNonBaseParams<Params extends BaseParam>(req: Params) {
```
samisayegh, line 372: It would be safer to place `pickNonBaseParams` and `unwrapError` in their own file, because atm we have a circular dependency, where:

- `search-api-client.ts` is importing from `html-api-client.ts`
- `html-api-client.ts` is importing from `search-api-client.ts`

This can lead to errors when bundling the files together,

It's nice to avoid `utils` in file names, but in this case, I can only think of `api-client-utils` for the file name 😄

packages/headless/src/controllers/quickview/case-assist-headless-quickview.ts
samisayegh: I think the tooling will give an error, saying these interfaces need to be annotated with jsdoc.

I don't see a `case-assist` use-case defined in `doc-parser/use-cases` though. I'd suggest tackling the doc task for all case-assist controllers in a separate PR.

samisayegh (approved):
💯

packages/headless/src/api/search/html/html-api-client.ts
mmitiche: I'm currently  exploring it, my attempts are not successful yet🧐

PR #1702, opened 2022-01-28
test(atomic): add hosted search page test

wmannard:
https://coveord.atlassian.net/browse/QA-918

Adapt Atomic cypress tests to run against Hosted search page

The test runs a script that adds a current (prod) hosted search page to the www directory. It replaces the js and css links to use ones from the repo.

The test validates that the version being served is actually the latest version from the atomic package (if you're running locally you might need to do a build for it to pass)

For now it's run with:
`cypress-hsp:open`
`cypress-hsp:test`
...but we can decide if we want to add it to the regular suite of tests eventually.

The next step will be adding basic tests to all components. I was wondering if we can import the existing component tests.
We could also use image change detection to highlight changes.


packages/atomic/cypress/scripts/get_hosted_search_page.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
```
olamothe, line 1: Any possibility to use just a node script for this ?

If we add a python dependency, it needs to be for a good reason, ie: it's actually impossible to perform the needed task with a technology already used in the repo.

The train of thought is simply to make sure that if a newcomer arrives in the team, and we ask them to setup to work on this project, it's as straightforward as it can be.

wmannard: That make sense. I re-wrote in js.

packages/atomic/cypress/scripts/getHostedSearchPage.js
```diff
@@ -0,0 +1,88 @@
...
+    'js/SearchInterfaces.bundle.min.js',
+  ];
+
+  links.forEach((link) => {
+    const filepath = PUBLIC_FOLDER + link;
+    try {
+      fs.mkdirSync(path.dirname(filepath), {recursive: true});
+    } catch (err) {
+      console.error(err);
+    }
```
wmannard, line 64: I'm not very good with error handling in js, please comment if it can be done better, here and elsewhere. I tried to use the synchronous methods where possible to make things simpler.

samisayegh: It looks good, one question though, when an error occurs (the dir is not created), should we exit the process? or continue with the file downloads below?

Same comment for other catch statements (e.g. line 86)

packages/atomic/package.json
```diff
@@ -36,8 +36,11 @@
     "test:watch": "stencil test --spec --watchAll",
     "generate": "stencil generate",
     "cypress:test": "cypress run --record --key 0e9d8bcc-a33a-4562-8604-c04e7bed0c7e --browser chrome",
-    "cypress:test:firefox": "cypress --record --key 0e9d8bcc-a33a-4562-8604-c04e7bed0c7e --browser firefox",
+    "cypress:test:firefox": "cypress run --record --key 0e9d8bcc-a33a-4562-8604-c04e7bed0c7e --browser firefox",
     "cypress:open": "cypress open",
+    "cypress-hsp:test": "cypress run --config-file cypress-hsp.json --browser chrome",
```
samisayegh, line 41: Should we add a line in github actions to call this script? Or will that be done in a separate PR/different way?

samisayegh (approved):

packages/atomic/cypress/scripts/getHostedSearchPage.js
wmannard: If any of the steps fail, the script might as well exit with failure immediately

packages/atomic/cypress/integration-hsp/hosted-search-page.cypress.ts
```diff
@@ -0,0 +1,20 @@
+const atomicPackage = require('../../package.json');
+
+/*
+ *  This test is run under the separate configuraton "atomic/cypress-hsp.json"
+ */
+
+describe('Validate Atomic components in Hosted Search Page', () => {
+  before(() => {
+    cy.exec('node cypress/scripts/getHostedSearchPage.js');
```
wmannard, line 9: ^ The script is called here

packages/atomic/package.json
wmannard: The script is triggered at line 9 of [hosted-search-page.cypress.ts](https://github.com/coveo/ui-kit/pull/1702/files#diff-a39dff766e5636dbd454266545e1cb6ed65cc003bd68a950e68d3f5d8edf6238)
using cy.exec()

samisayegh: Oh, I thought we were going to run them as a separate step. Including it as part of the atomic test suite work stoo ✔️

packages/atomic/cypress/scripts/getHostedSearchPage.js
btaillon: I think doing `process.exit` or throwing a new error are both ideal.

btaillon (approved):
Looks good!

packages/atomic/package.json
wmannard: To be clear, the flow now is like this:
npm run cypress:test --> run the usual tests, not the hosted search page test
npm run cypress-hsp:test --> run only the hosted search page test, with the download script being launched at the beginning of the test

wmannard: Just updated the README

packages/atomic/cypress/scripts/getHostedSearchPage.js
wmannard: ok, done ✔️

packages/atomic/package.json
```diff
@@ -36,8 +36,11 @@
...
+    "cypress-hsp:test:firefox": "cypress run --config-file cypress-hsp.json --browser firefox",
```
ThibodeauJF, line 42: Why firefox too?

ThibodeauJF (approved):
:100:

packages/atomic/package.json
wmannard: Since Cypress supports Firefox and this comes at no extra cost, I was thinking we might want to have our tests run again both FF and the default Chrome-based test. I had added the Firefox command to the package.json a while back for the regular tests, but didn't follow up about integrating them into our pipelines.

packages/atomic/package.json
samisayegh: Ok, so we're back to the start of the thread :) Where does `npm run cypress-hsp:test` get called? Should it be run as part of ui-kit's continuous integration?

PR #1703, opened 2022-02-01
test(atomic-react): improve stability of atomic-react smoke tests

olamothe:
* Add a script to `npm link` local react instance when running the samples in dev
* Add a larger viewport for the smoke tests, I can't be bothered with making the sample responsive on all devices :D 
* The regex check on the resize observer, from the google search I've done, seems to be what many people recommend, as it is apparently a benign error that is reported by Chrome. And cypress will automatically fail on any console error being thrown.

Now, I'm not sure what actually causes this error to happen. I suspect a deeper performance issue in Atomic as a whole (nothing to do with Atomic React), that simply (sometimes) gets triggered in the "full commerce page sample" because it uses many components. But I don't think the performance issue is critical. Just not as good as it could be in a perfect world, I guess.

So, I propose just shoving it under the rug for now.

https://coveord.atlassian.net/browse/KIT-1389


ThibodeauJF (approved):

btaillon (approved):

samisayegh (approved):
Thanks 💯

PR #1704, opened 2022-02-01
feat(headless): allow toggleSelectFacetValue to add new values

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1388

To try it out with Atomic:

```html
<script type="module">
  import {loadFacetSetActions} from '/build/headless/headless.esm.js';
  async function main() {
    const searchInterface = document.querySelector(
      'atomic-search-interface'
    );
      accessToken: 'xx564559b1-0045-48e1-953c-3addd1ee4457',
    });

    const engine = searchInterface.engine;
    const action = loadFacetSetActions(engine).toggleSelectFacetValue({
      facetId: 'author',
      selection: {state: 'selected', value: 'helloooo'},
    });
    engine.dispatch(action);

  }
  main();
</script>
```

### DIFF
--- a/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
+++ b/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
@@ -183,26 +183,33 @@ describe('facet-set slice', () => {
       expect(targetValue?.state).toBe('idle');
     });
 
-    it('adds new values', () => {
+    it('adds a new value before idle values when it does not exist', () => {
       const id = '1';
 
-      const facetValue = buildMockFacetValue({value: 'TED'});
+      const newFacetValue = buildMockFacetValue({
+        value: 'TED',
+        state: 'selected',
+      });
 
-      state[id] = buildMockFacetRequest({currentValues: []});
+      state[id] = buildMockFacetRequest({
+        currentValues: [
+          buildMockFacetValue({value: 'selected1', state: 'selected'}),
+          buildMockFacetValue({value: 'selected2', state: 'selected'}),
+          buildMockFacetValue({value: 'idle1', state: 'idle'}),
+          buildMockFacetValue({value: 'idle2', state: 'idle'}),
+        ],
+      });
 
       const action = toggleSelectFacetValue({
         facetId: id,
-        selection: facetValue,
+        selection: newFacetValue,
       });
-      const finalState = facetSetReducer(state, action);
 
-      const targetValue = finalState[id].currentValues.find(
-        (req) => req.value === facetValue.value
-      );
-      expect(targetValue).toEqual(facetValue);
+      const finalState = facetSetReducer(state, action);
+      expect(finalState[id].currentValues.indexOf(newFacetValue)).toBe(2);
     });
 
-    it('sets #freezeCurrentValues to true', () => {
+    it('sets #freezeCurrentValues to true when toggled value exists', () => {
       const id = '1';
 
       const facetValue = buildMockFacetValue({value: 'TED'});
@@ -217,6 +224,20 @@ describe('facet-set slice', () => {
       const finalState = facetSetReducer(state, action);
 
       expect(finalState[id].freezeCurrentValues).toBe(true);
+    });
+
+    it('does not #freezeCurrentValues to true when toggled value exists', () => {
+      const id = '1';
+
+      state[id] = buildMockFacetRequest({currentValues: []});
+
+      const action = toggleSelectFacetValue({
+        facetId: id,
+        selection: buildMockFacetValue({value: 'TED'}),
+      });
+      const finalState = facetSetReducer(state, action);
+
+      expect(finalState[id].freezeCurrentValues).toBe(false);
     });
 
     it('sets #preventAutoSelect to true', () => {
--- a/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
+++ b/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
@@ -226,7 +226,7 @@ describe('facet-set slice', () => {
       expect(finalState[id].freezeCurrentValues).toBe(true);
     });
 
-    it('does not set #freezeCurrentValues to true when toggled value does not exists , () => {
+    it('does not set #freezeCurrentValues to true when toggled value does not exists', () => {
       const id = '1';
 
       state[id] = buildMockFacetRequest({currentValues: []});
--- a/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
+++ b/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
@@ -226,7 +226,7 @@ describe('facet-set slice', () => {
       expect(finalState[id].freezeCurrentValues).toBe(true);
     });
 
-    it('does not #freezeCurrentValues to true when toggled value exists', () => {
+    it('does not set #freezeCurrentValues to true when toggled value exists', () => {
       const id = '1';
 
       state[id] = buildMockFacetRequest({currentValues: []});
--- a/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
+++ b/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
@@ -226,7 +226,7 @@ describe('facet-set slice', () => {
       expect(finalState[id].freezeCurrentValues).toBe(true);
     });
 
-    it('does not set #freezeCurrentValues to true when toggled value exists does not', () => {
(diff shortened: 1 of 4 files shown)


packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -79,17 +79,17 @@ export const facetSetReducer = createReducer(
...
           (req) => req.value === selection.value
         );
 
-        if (!targetValue) {
-          return;
+        if (existingValue) {
+          const isSelected = existingValue.state === 'selected';
+          existingValue.state = isSelected ? 'idle' : 'selected';
+        } else {
+          facetRequest.currentValues.push(selection);
```
ThibodeauJF, line 90: Two things if slightly unsure of:
- Should we care about the order the value is inserted? Or should we let the API take care of it, by removing freezeCurrentValues
- Now it takes a full value with selection, should we toggle the state on a new value as well as an existing one?

btaillon: > Should we care about the order the value is inserted? Or should we let the API take care of it, by removing freezeCurrentValues

I don't remember why `freezeCurrentValues` is necessary here. I would prefer to let the API choose the order.

> Now it takes a full value with selection, should we toggle the state on a new value as well as an existing one?

I feel like this issue stems from `toggleSelectFacetValue` being meant to be called with a value we read directly in the state, vs here we are expecting someone to create a value from scratch. I feel like the least ambiguous method would be to instead introduce a new action & function (e.g., `setFacetValueInitialState` or just `setFacetValueState`). I understand we took a decision to go with a `toggle` function, but we may be uncovering a better reason not to.

samisayegh: First off, Benjamin's suggestion of a dedicated action could be a good option. If we take it, I would suggest `updateFacetCurrentValues` for the name.

Counter argument might be `toggleSelectCategoryFacetValue` creates new paths, so adjusting to create a missing value may make the behaviour more consistent with the category facet.


For creating a value inside toggleSelect, those are great questions.

1. We could treat it similar to selecting a facet search result (no control over position). There's logic in that reducer for inserting the value at the first idle index. It also adjusts the number of values if adding the new value takes the array over `numberOfValues`, otherwise the api will error. We should probably extract and reuse this code.

The freeze current values is tricky, but I think it _needs_ to be `false`, otherwise the facet will not "fill up" to the number of values? My reasoning being when `freeze` is `true`, it's for UX in response to a client interaction. However, an end-user cannot click on a non-existant value => if toggleSelect is adding, then it is most likely programmatic => no need to freeze?


2. When a toggleSelect value is being added, setting the initial state to what is given (i.e. not toggling) feels the most natural imo.

ThibodeauJF: Okay I think I'll go for being close to the behaviour of toggleSelectCategoryFacetValue & the facet search selection

When it doesn't exist
1. insert before the first idle value
2. don't touch freezeCurrentValues
3. don't toggle the state of the new value, take it as it is

packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
```diff
@@ -200,6 +226,20 @@ describe('facet-set slice', () => {
       expect(finalState[id].freezeCurrentValues).toBe(true);
     });
 
+    it('does not set #freezeCurrentValues to true when toggled value exists', () => {
```
ThibodeauJF, line 229: It does not touch the freezeCurrentValues, it does not set it to false

packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
```diff
@@ -200,6 +226,20 @@ describe('facet-set slice', () => {
       expect(finalState[id].freezeCurrentValues).toBe(true);
     });
 
+    it('does not set #freezeCurrentValues to true when toggled value exists does not', () => {
```
ThibodeauJF, line 229:
```suggestion
    it('does not set #freezeCurrentValues to true when toggled value does not exists , () => {
```

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -189,27 +189,29 @@ export const facetSetReducer = createReducer(
...
       });
   }
 );
 
+function insertValueBeforeIdleValues(
+  facetRequest: FacetRequest,
+  facetValue: FacetValueRequest
+) {
+  const {currentValues} = facetRequest;
+  const firstIdleIndex = currentValues.findIndex((v) => v.state === 'idle');
```
olamothe, line 204: There might be one flaw here if the `selected` are not all grouped together at the beginning of the array, and there's gap where an idle value is inserted before.

Normally it "should not" happen, but... 
What if instead we used the last index for `selected` state ? So `insertValueBeforeIdleValues` -> `insertValueAfterSelectedValues`

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -79,19 +79,19 @@ export const facetSetReducer = createReducer(
           return;
         }
 
-        const targetValue = facetRequest.currentValues.find(
+        facetRequest.preventAutoSelect = true;
```
olamothe, line 82: Since we expect people to use this mechanism to bootstrap their UI with preselected values on load, I guess it makes sense that we would not also allow auto selection from the API in that scenario.

olamothe (approved):

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
ThibodeauJF: So before
```
[✓] Selected
[✓] New
[.  ] Idle
[✓] Selected
[.  ] Idle
```

with your proposition

```
[✓] Selected
[.  ] Idle
[✓] Selected
[✓] New
[.  ] Idle
```

is that it?

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -189,27 +189,29 @@ export const facetSetReducer = createReducer(
...
-        ];
-        facetRequest.numberOfValues = facetRequest.currentValues.length;
+        insertValueBeforeIdleValues(facetRequest, searchResultValue);
         facetRequest.freezeCurrentValues = true;
         facetRequest.preventAutoSelect = true;
       });
   }
 );
 
+function insertValueBeforeIdleValues(
```
samisayegh, line 199: The `before` in this function name is slightly misleading. The insert happens "in place" of the first idle value, or at the end of the array if there are no idle values.

Suggestion: simplify the name  to `insertValue` or `addNewValue` etc. I don't think explaining in the name the details of how the insertion is happening is important.

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
olamothe: Yeah I mean it's still weird anyway 🤷

samisayegh: Hmm, the logic should insert correctly when the selected values are not grouped at the beginning. It's possible to test out the UX by selecting a facet search result with different facet values pre-selected.

Note that it _replaces_ the first idle value, not insert before the idle value.

packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
```diff
@@ -200,6 +226,20 @@ describe('facet-set slice', () => {
       expect(finalState[id].freezeCurrentValues).toBe(true);
     });
 
+    it('does not set #freezeCurrentValues to true when toggled value does not exists', () => {
```
samisayegh, line 229:
```suggestion
    it('does not set #freezeCurrentValues to true when toggled value does not exist', () => {
```

samisayegh: It would be helpful to group the tests for adding a new value together in a describe block:

```
describe('when the toggled value does not exist', () => {
  ...
})
```

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
samisayegh: Adding a test check that it is set to true when a new value is added would be nice.

Said differently, if this line was moved to the bottom, I don't think any test atm would fail? But it would lead to a change in behavior,

samisayegh (approved):

PR #1705, opened 2022-02-02
feat(atomic-react): add stencil plugin to generate angular module

olamothe:
I've spent quite a bit of time trying to see if we could do without this plugin, and by simply exporting the components straight up (if you recall my previous PR about Angular IVY/ngcc).

I wasn't able to come up with something that works.

So instead, I've created a small stencil plugin that generate one angular module, which automatically generate the declarations for all components.

Which now means that a consumer of the lib should be able to declare an import on `AtomicAngularModule`, and have access to all atomic components automatically.

https://coveord.atlassian.net/browse/KIT-1398


btaillon (approved):
Looks good! I just hope this doesn't affect treeshake-ability though.

samisayegh (approved):
Pretty cool 💯

PR #1707, opened 2022-02-03
ci: changed NPM token for updating tags

btaillon:
https://coveord.atlassian.net/browse/KIT-1372

I'm gonna try to discuss with CSEC, IT and/or DT about updating this token for one that isn't a personal access token.

### DIFF
--- a/JenkinsfileProductionRelease
+++ b/JenkinsfileProductionRelease
@@ -15,8 +15,8 @@ node('linux && docker') {
     
     stage('Npm publish') {
       withCredentials([
-      string(credentialsId: 'npmjs_com_token', variable: 'NPM_TOKEN')]) {
-        sh "echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > ~/.npmrc"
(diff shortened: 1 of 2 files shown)


JenkinsfileProductionRelease
```diff
@@ -15,8 +15,8 @@ node('linux && docker') {
     
     stage('Npm publish') {
       withCredentials([
-      string(credentialsId: 'npmjs_com_token', variable: 'NPM_TOKEN')]) {
-        sh "echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > ~/.npmrc"
```
btaillon, line 19: Jenkins gave some warnings about using string interpolation with double quotes.

According to [this doc](https://www.jenkins.io/doc/pipeline/steps/credentials-binding/), it's better to use single quotes.

samisayegh: Hmm, are you sure we can remove the braces? I did not see an example without them,

samisayegh (approved):

PR #1712, opened 2022-02-04
docs(angular): create angular sample package

olamothe:
Bootstrapping a `samples/angular` package using `ng` CLI.

Include a `atomic-angular-searchpage` component, which reproduce the "full commerce" page, similar to atomic react (allows to tests all components).

So there's quite a bunch of boilerplate that's just there to allow angular to compile and run the project.

Main files to look at is how we include the `AtomicAngularModule` in `AppModule`, and then how the configuration, initialization and rendering of the search interface happens in `AtomicAngularPageComponent`.

https://coveord.atlassian.net/browse/KIT-1402


packages/samples/angular/src/app/atomic-angular-page/atomic-angular-page.component.css
```diff
@@ -0,0 +1,163 @@
+atomic-color-facet[field='cat_color']::part(value-Black) {
```
olamothe, line 1: Pretty much everything here would be replaced by the layouts being currently worked on by @ThibodeauJF

lerna.json
```diff
@@ -1,7 +1,8 @@
 {
   "packages": [
     "packages/*",
-    "packages/samples/**"
+    "packages/samples/**",
+    "packages/atomic-angular/projects/**"
```
samisayegh, line 5: Why the double `*` at the end? Are there multiple `package.json`?

samisayegh (approved):

btaillon (approved):

lerna.json
olamothe: Indeed, `/**` not needed

PR #1714, opened 2022-02-07
Fix(Quantic): New properties and slot in the QunaticCaseAssistInterface  and QuanticDocumentSuggestion

mmitiche:
### QuanticDocumentSuggestion:
- Default value of  maxDocuments updated to be 3 instead of 5.
- New slot added named no-suggestions to give the user the ability to customize what the component will display when there is no suggestions to display.
- Fixed the behaviour of the component when given invalid properties.

### QuanticCaseAssistInterface:
- Sending the case start event only in the first execution of the initialized  callback.


packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
```diff
@@ -26,13 +26,26 @@ export default class QuanticCaseAssistInterface extends LightningElement {
...
    * @type {string}
    */
   @api caseAssistId;
+  /**
+   * Whether or not we want to log a case start analytics event when initializing this component.
+   * @api
+   * @type {boolean}
+   * @defaultValue `false`
+   */
+  @api preventLogCaseStartOnInit = false;
```
erocheleau, line 41: Why would we prevent this exactly?

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -181,8 +185,9 @@ export default class QuanticDocumentSuggestion extends LightningElement {
       console.warn(
         'Please enter a valid number of automatically opened documents.'
       );
+    }else{
```
erocheleau, line 188: Seems like linting errors there?

erocheleau (approved):
Some of these settings are a bit weird but oh well.
I don't get why we would let people hide the quickview or prevent logging the ticket_create_start event?

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
mmitiche: Yes you are right! I'll fix that.

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
mmitiche: Mainly I did this since the action to log a case start is dispatched in the initialize method which is a callback executed each time  all the components linked to the same engine  are registered. when having a page that contains all the components this callback will be called just once, but when being in the case of a flow and the components are in different screens, the components registers only when the screen containing them displays, so on each screen appearing the components in it registers triggering the initialize  callback once again and sending the log case start again, and we don't want that. so that was the case where preventing this logCaseStart would make sense.
But now thinking about that I think I can improve that by just changing where this logCaseStart is dispatched bringing it out of this callback.

mmitiche: In that case assist flow I encountered this problem and the solution was to prevent sending this caseStart from the quanticCaseAssistInterface and instead sending it manually using Headless from the first screen of the flow

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -39,14 +41,16 @@ export default class QuanticDocumentSuggestion extends LightningElement {
...
+  @api hideQuickview = false;
   /**
-   * Whether or not we want to fetch suggestions when initializing this component.
+   * Whether or not we want to prevent fetching suggestions when initializing this component.
    * @api
    * @type {boolean}
+   * @defaultValue `false`
    */
-  @api fetchOnInit = false;
+  @api preventFetchOnInit = false;
```
lbergeron, line 53: Removing `fetchOnInit` is a breaking change. I suggest we manage to honor the now deprecated property.

lbergeron (changes requested):
We need to decide how we handle the breaking changes.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
mmitiche: What are the possible options we can do to handle such cases when we need to make breaking change?

lbergeron: Unfortunately, I'm not yet sure what is the best way to handle breaking changes with LWC and unmanaged packages.

mmitiche: Okay, so I think I'll just revert the changes about #fetchOnInit and #showQuickview just so I dont block this PR needed by the cookbook, and I will come back and update these two properties once we figure out how we will handle the breaking changes in quantic, what do u think?

lbergeron: I agree, reverting those breaking changes seems to be the way to go for now.

nathanlb:
> 

I don't think we should be saying "oh well ¯\_(ツ)_/¯ " while approving a PR in this repo. Let's try to be as rigorous as we can to avoid future headaches. 😅

aelfaleh (approved):

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
nathanlb: Wait a second why is the event retriggering when new components register? Are you using more than one case assist interface?

nathanlb: The initialized callback is for the engine so it should only trigger once when the engine is initialized and not if new components are added after. I think the solution here is a modification in the quanticHeadlessLoader that makes sure it's only executed once after the engine instance is created.

erocheleau:
As far as the breaking changes go, I would give the go ahead on it since officially the quantic case assist components are not "released" right?

I mean they have been merged into master, but the case assist cookbook is not out yet and AFAIK nobody uses JUST the quantic components to do something by themselves. So I would be okay renaming attributes until we have our first version of the cookbook released and "published" for customers to use.

@nathanlb What do you think?

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -198,8 +203,9 @@ export default class QuanticDocumentSuggestion extends LightningElement {
       console.warn(
         'Please enter a valid maximum number of document suggesions.'
       );
+    } else {
+      this._maxDocuments = Number(value);
```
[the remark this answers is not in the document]
mmitiche: I just fixed this, it was just the way I used to be able to log errors when these proprieties have invalid values and fall back to the default behaviour in such case. but I just updated this following your comment yesterday about how we should handle invalid values by just rendering an error message.
So the properties now are exactly like the others at the top, and we staying consistent across all the quantic components on how we handle invalid props values.

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
erocheleau: That's quite a big change :D Are you comfortable changing the behavior of the quantic headless loader at this point @nathanlb ?

mmitiche: @nathanlb Nope I'm not using more than one case assist interface, just one.  
when loading the first screen and all the components in it are initialized with headless we call the initializedCallback.
when going to a next screen with a new quantic component and we initialize this component we face the following code when executing the `initializeWithHeadless` function: 
  component.initialized = true;
  if (window.coveoHeadless[engineId].initializedCallback && areAllComponentsInitialized(engineId)) {
    debounceInitializedCallback(engineId);
  }
  and since the initializedCallback is there and all the components are indeed initialized  we call the initializedCallback again.

nathanlb: @erocheleau yes you're right it would be tricker considering the search use-case in community relies on the re-execution of the initialized callback.

Right so maybe I just don't understand how you're planning on using this @mmitiche. If you only have one caseAssistInterface then you either preventLogCaseStartOnInit or you don't. so if you set this option to true the event never gets sent?

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
```diff
@@ -70,7 +83,13 @@ export default class QuanticCaseAssistInterface extends LightningElement {
...
     this.actions = {
       ...CoveoHeadlessCaseAssist.loadCaseAssistAnalyticsActions(engine),
+      ...CoveoHeadlessCaseAssist.loadCaseFieldActions(engine),
     };
-    this.engine.dispatch(this.actions.logCaseStart());
+    if (!this.preventLogCaseStartOnInit) {
+      this.engine.dispatch(this.actions.logCaseStart());
+    }
+    if (!this.preventFetchClassificationOnInit) {
+      this.engine.dispatch(this.actions.fetchCaseClassifications());
```
nathanlb, line 92: I think classifications should not be fetched from the interface component. Classifications are meant to be fetched by the case input controller.

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
nathanlb: Maybe we should remove the initialized callback for this interface and just send the case start analytics event from a cookbook component at the start of the flow.

mmitiche: Exactly, in the only caseAssistInterface I'm setting the  preventLogCaseStartOnInit option to true. and in the first screen of the flow I'm sending the analytic event manually like so: https://github.com/coveooss/sf-case-assist-cookbook/blob/a64589489c2edb53e9440bdc02dda6cee13b1011/src/main/default/lwc/describeProblemScreen/describeProblemScreen.js#L73

mmitiche: Removing the initialized callback for this interface can be a solution, but what if a user want to use the Case Assist Components but he decides to not use them in a  flow. He will just create one page containing all the needed components, in this case he will not face the problem of wanting to prevent the log case start event. so in my opinion it'll make sense to keep this feature of choosing if we want to prevent these events or not. If working with flows like in the cookbook preventing the events is necessary and the user should send the events he wants  from his screens. and if the user for any reason doesn't want to work with flows and he builds everything in one page he will find the right events sent for him.

lbergeron:
> As far as the breaking changes go, I would give the go ahead on it since officially the quantic case assist components are not "released" right?
> 
> I mean they have been merged into master, but the case assist cookbook is not out yet and AFAIK nobody uses JUST the quantic components to do something by themselves. So I would be okay renaming attributes until we have our first version of the cookbook released and "published" for customers to use.
> 
> @nathanlb What do you think?

@erocheleau Unfortunately, you can't presume a component is not used as part of an implementation because it is not used in the cookbook.

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
mmitiche: Yes I agree, that's a simple effective solution, I just implemented it and got rid of the previous boolean properties.

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
mmitiche: Fixed 👌

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -101,6 +122,15 @@ export default class QuanticDocumentSuggestion extends LightningElement {
     this.template.removeEventListener('rating', this.onRating);
   }
 
+  validateProps() {
+    if (!(Number(this.maxDocuments) > 0)) {
+      this.renderingError = `${this.maxDocuments} is an invalid maximum number of document suggestions. A integer greater than 0 was expected.`;
```
lbergeron, line 127: I'd suggest to surround the interpolated value with double quotes as it makes it easier to spot leading or trailing white space issues in the error message.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -39,14 +41,30 @@ export default class QuanticDocumentSuggestion extends LightningElement {
...
   @api showQuickview = false;
   /**
    * Whether or not we want to fetch suggestions when initializing this component.
    * @api
    * @type {boolean}
+   * @defaultValue `false`
    */
   @api fetchOnInit = false;
+  /**
+   * The maximum number of document suggesions to display.
```
lbergeron, line 55:
```suggestion
   * The maximum number of document suggestions to display.
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticDocumentSuggestion/exampleQuanticDocumentSuggestion.js
```diff
@@ -28,7 +28,7 @@ export default class ExampleQuanticDocumentSuggestion extends LightningElement {
       label: 'Max documents',
       description:
         "The maximum number of document suggesions to display, it's a value between 1 and 5.",
```
lbergeron, line 30:
```suggestion
        "The maximum number of document suggestions to display. The value must be between 1 and 5, inclusively.",
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticDocumentSuggestion/exampleQuanticDocumentSuggestion.js
```diff
@@ -28,7 +28,7 @@ export default class ExampleQuanticDocumentSuggestion extends LightningElement {
...
-      defaultValue: 5,
+      defaultValue: 3,
     },
     {
       attribute: 'fetchOnInit',
```
lbergeron, line 34: There is a typo in the `showQuickview` description.

> Whether or not we want to displ**a**y the quick view for the document suggestions.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.html
```diff
@@ -1,55 +1,60 @@
...
+    <template if:false={loading}>
+      <template if:true={hasSuggestions}>
+        <div class="slds-card">
+          <lightning-accordion active-section-name={openedDocuments} allow-multiple-sections-open
+            class="slds-p-vertical_xx-small">
+            <template iterator:it={suggestions}>
+              <lightning-accordion-section data-id={it.value.uniqueId} onclick={handleSectionClick}
+                key={it.value.uniqueId} name={it.value.uniqueId} label={it.value.title}>
+                <div slot="actions" class="slds-var-p-horizontal_x-large">
+                  <slot name="rating" data-doc-id={it.value.uniqueId}> </slot>
```
lbergeron, line 24:
```suggestion
                  <slot name="rating" data-doc-id={it.value.uniqueId}></slot>
```

lbergeron (approved):
Found a few typos here and there, but nothing major.

nathanlb (approved):
Thanks for the constructive back-and-forth and and nice work! 👍

PR #1715, opened 2022-02-07
fix(atomic): improved aria-live accessibility

btaillon:
https://coveord.atlassian.net/browse/KIT-1237

This took me a long time to figure out, primarily because I made the mistake of testing VoiceOver for Mac OS with Firefox, but also because I didn't write down the various issues I encountered and what device + screen reader they affected, meaning I wasted a lot of time coming up with complex solutions to simple problems.

The flakiness issues with aria-live mostly come to this, and possibly more (from my memory):
1. Having multiple aria-live in the page was very flaky with some screen readers
    * Some aria-live were getting read, and some were getting ignored. It seems like one aria-live would be consistently ignored while the others weren't, and which one was getting ignored seemed random.
2. VoiceOver for iOS doesn't (or rarely?) detects `aria-live` when it's inside a ShadowDOM
    * This caused VoiceOver for iOS to randomly start/stop detecting aria-live in the page on each reload.
3. NVDA (and JAWS?) occasionally lose track of `aria-live` when the element is removed/reappended to the DOM
    * This caused NVDA and JAWS to never pronounce "No results". Adding a `key=` to `atomic-aria-live`, surrounding it with `<Host></Host>` instead of an array or appending it manually in `initialize` did not seem to fix this issue though.
4. Some screen readers strictly read everything that ever appears in the `atomic-aria-live` element (e.g., TalkBack) while some will stop reading whenever the user does an action and only read the next updates to `atomic-aria-live` (e.g., VoiceOver)

The solutions I came up with were:
* Recommend developers to add an `atomic-aria-live` outside the search interface.
  * This fixes issue 2 and 3
* Use the `atomic-aria-live` element itself and give it `aria-live` instead of rendering multiple elements in a ShadowDOM with each their own `aria-live`
  * This fixes issue 1 and 2
* Append a default `atomic-aria-live` to the search interface instead of rendering it in its ShadowDOM
  * When somebody forgets to add an `atomic-aria-live` to their page (or accidentally removes it), this fixes issue 2 and in theory would be supposed to fix issue 3 (but doesn't, as described in the bullet point under issue 3)

# Important
I also removed 3 redundant accessibility testing pages. I don't think they are necessary anymore, given that we have `commerce-full.html`.


packages/atomic/cypress/integration/common-assertions.ts
```diff
@@ -57,19 +58,8 @@ export function assertRemovesComponent(
...
-  selector: () => Cypress.Chainable<JQuery<HTMLElement>>,
-  message: string
-) {
+export function assertAriaLiveMessage(message: string) {
   it(`screen readers should read out "${message}".`, () => {
-    selector().should('contain.text', message);
-  });
-}
-
-export function assertNoAriaLiveMessage(
```
btaillon, line 69: I removed the `assertNoAriaLiveMessage` tests since I'm afraid they would become flaky due to a variety of elements that can update the `aria-live` text.

packages/atomic/src/components.d.ts
```diff
@@ -16,6 +16,11 @@ import { InitializationOptions } from "./components/atomic-search-interface/atom
 import { StandaloneSearchBoxData } from "./utils/local-storage-utils";
 export namespace Components {
     interface AtomicAriaLive {
+        /**
+          * Only the `atomic-aria-live` element with the greatest priority will be used to announce changes in the search interface.
```
btaillon, line 20: I tried to add `@internal` here but it didn't stick for some reason?

The reason I went for this solution is because I'm afraid that developers using JS frameworks may dynamically add or remove `aria-live`, and I wanted a solution that works as consistently as possible in those cases.

packages/atomic/src/components/atomic-search-interface/atomic-aria-live.tsx
```diff
@@ -14,27 +13,64 @@ export interface UpdateLiveRegionEventArgs {
...
+    if (wouldOverwriteAnotherRegion) {
+      this.lastUpdatedRegion = region;
+      if (!message) {
+        return;
+      }
+    }
+    this.message = message;
+  }
+
+  public connectedCallback() {
```
btaillon, line 48: I spent an embarrassing amount of time trying to figure out a way to detect all the `aria-live` elements in the page and pick out the one with the highest priority. Unfortunately, what caused me the longest delay, was that Stencil's `@Listen` seems to somehow provide only a clone of the event's arguments to the method, which meant updating said arguments gave nothing to the event dispatcher.

packages/atomic/cypress/integration/no-results-selectors.ts
```diff
@@ -1,5 +1 @@
-import {AriaLiveSelectors} from './aria-live-selectors';
-
-export const NoResultsSelectors = {
-  liveRegion: () => AriaLiveSelectors.region('no-results'),
-};
+export const NoResultsSelectors = {};
```
ThibodeauJF, line 1: Can we delete the file

packages/atomic/src/components/atomic-search-interface/atomic-aria-live.tsx
```diff
@@ -1,40 +1,80 @@
...
+  @Element() private host!: HTMLAtomicAriaLiveElement;
+  @State() private message = '';
 
-  @Listen('atomic/accessibility/updateLiveRegion', {target: 'document'})
-  protected updateMessage({
-    detail: {message, region},
-  }: CustomEvent<UpdateLiveRegionEventArgs>) {
-    this.regions = {...this.regions, [region]: message};
+  /**
+   * Only the `atomic-aria-live` element with the greatest priority will be used to announce changes in the search interface.
```
ThibodeauJF, line 21: The priority system might be a bit overkill, at least at first.

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -342,6 +343,12 @@ export class AtomicSearchInterface {
     window.addEventListener('hashchange', this.onHashChange);
   }
 
+  private initAriaLive() {
```
ThibodeauJF, line 346: What if we either:
- not create a default aria live component and document adding one properly in a accessibility article. Especially if we know that it doesn't work super well inside the shadow dom for some tools.
- check with a query selector if one exists before adding a second one

packages/atomic/src/components/atomic-search-interface/atomic-aria-live.tsx
btaillon: it's definitely overkill, I agree, which is why I made it internal. I added it pretty much exclusively to differentiate between an `atomic-aria-live` inside the all search interfaces and one inside. Should I use a different mechanism?

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
btaillon: > not create a default aria live component and document adding one properly in a accessibility article. Especially if we know that it doesn't work super well inside the shadow dom for some tools.

I don't mind that, but I think the current solution might deflect some maintenance cases.

> check with a query selector if one exists before adding a second one

I prefer not to do that in case an `atomic-aria-live` element is added/removed dynamically.

npushkarskii (approved):
LGTM 👍 
I only left one minor comment.

packages/atomic/src/components.d.ts
olamothe: That's generated code, from the component itself.

And even if you add things to the definition file, that would be wrong since it would not match the implementation.

packages/atomic/src/components/atomic-search-interface/atomic-aria-live.tsx
olamothe: If it's not needed or overkill, I propose to remove it.

We should not add props and option because we think we might possibly need it in in the future. 
Only when it's necessary.

packages/atomic/src/components/atomic-search-interface/atomic-aria-live.tsx
```diff
@@ -1,40 +1,80 @@
...
+
+  /**
+   * @internal
+   */
+  @Method()
+  public async updateMessage(region: string, message: string) {
+    const wouldOverwriteAnotherRegion = region !== this.lastUpdatedRegion;
+    if (wouldOverwriteAnotherRegion) {
+      this.lastUpdatedRegion = region;
+      if (!message) {
```
olamothe, line 45: Shouldn't that be at the top of the function, as an early check and return ?

Is there a scenario where we would want to update the message to "nothing" ? That's the same as not changing it at all, I think ?

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
olamothe: > check with a query selector if one exists before adding a second one

> I prefer not to do that in case an atomic-aria-live element is added/removed dynamically.

I think that's also overkill. 

I think, either it exists at init time (and we don't expect it to be removed, because why would someone need to do that?), or it doesn't and then we need to add one.

packages/atomic/src/components/atomic-search-interface/atomic-aria-live.tsx
btaillon: Changing the message to nothing may cancel the message on some screen readers. The purpose of returning here is to allow a region to cancel its own message, but not another region's.

packages/atomic/src/components/atomic-search-interface/atomic-aria-live.tsx
```diff
@@ -1,40 +1,83 @@
...
+      }
+    }
+    this.message = message;
+  }
+
+  public connectedCallback() {
+    const eventListener = (ev: Event) =>
+      this.onFindAriaLive((ev as CustomEvent<FindAriaLiveEventArgs>).detail);
+    document.addEventListener(
+      'atomic/accessibility/findAriaLive',
```
ThibodeauJF, line 59: now that your not using a decorator, you can put the event value inside a variable & reuse it directly

ThibodeauJF (approved):

PR #1716, opened 2022-02-08
chore(headless): document each specific "TestUtils"

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1410

### DIFF
--- a/packages/headless/src/test/index.ts
+++ b/packages/headless/src/test/index.ts
@@ -1,4 +1,17 @@
-export * from './mock-engine';
-export * from './mock-state';
-export * from './mock-result';
-export * from './mock-raw';
+export {
+  buildMockSearchAppEngine,
+  buildMockCaseAssistEngine,
+  buildMockProductListingEngine,
+  buildMockProductRecommendationsAppEngine,
+  buildMockRecommendationAppEngine,
+} from './mock-engine';
+export type {
+  MockSearchEngine,
+  MockCaseAssistEngine,
+  MockProductListingEngine,
+  MockProductRecommendationEngine,
+  MockRecommendationEngine,
+} from './mock-engine';
+export {createMockState} from './mock-state';
+export {buildMockResult} from './mock-result';
+export {buildMockRaw} from './mock-raw';
--- a/packages/headless/src/test/mock-engine.ts
+++ b/packages/headless/src/test/mock-engine.ts
@@ -116,6 +116,8 @@ export interface MockProductListingEngine
     MockEngine {}
 
 /**
+ * For internal use only.
+ *
  * Returns a non-functionnal `ProductListingEngine`.
  * To be used only for unit testing controllers, not for functionnal tests.
  */
--- a/packages/headless/src/test/mock-engine.ts
+++ b/packages/headless/src/test/mock-engine.ts
@@ -61,6 +61,8 @@ export interface MockSearchEngine
     MockEngine {}
 
 /**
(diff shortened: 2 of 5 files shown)


packages/headless/src/test/mock-engine.ts
```diff
@@ -60,6 +60,10 @@ export interface MockSearchEngine
   extends SearchEngine<SearchAppState>,
     MockEngine {}
 
+/**
+ * Returns non-functionnal `SearchEngine`.
+ * To be used only for unit testing controllers, not for functionnal tests.
+ */
```
ThibodeauJF, line 66: Could also add "deprecate" if we want to remove them in the future

btaillon: IMO, if we deprecate them I don't think we should also document them. So far I prefer the idea of deprecating all of them.

btaillon (approved):

packages/headless/src/test/mock-raw.ts
```diff
@@ -1,5 +1,10 @@
 import {Raw} from '../api/search/search/raw';
 
+/**
+ * Returns the `Raw` property of a `Result`.
+ * @param config partial `Raw`.
+ * @returns the new `Raw`.
```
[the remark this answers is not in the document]
jpmarceau: I tried something there, but I don't really understand what's `@param config partial 'Raw'.` though. Anything someone show know when passing this param?

jpmarceau (commented):
I wasn't exactly sure whether this was for a component that would get documented. I tried generating the parsed_doc.json and it did not include these strings, so I guess not?

Anyways, I tried to review like I would normal strings :)

packages/headless/src/test/mock-engine.ts
olamothe: I would add a disclaimer: "For internal use only". 

Like making it obvious that we don't encourage people to use this...

packages/headless/src/test/index.ts
```diff
@@ -1,4 +1,17 @@
-export * from './mock-engine';
-export * from './mock-state';
-export * from './mock-result';
-export * from './mock-raw';
+export {
```
olamothe, line 1: That's something we should have caught on earlier I guess.

It seems we use that for Atomic, so we are guilty of the same thing we should not encourage clients to do... 

That is, treating internal stuff from a package (Headless) as something that can be seen as "public" and backward compatible.

packages/headless/src/test/mock-raw.ts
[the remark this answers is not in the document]
ThibodeauJF: Raw is a property or Result which contains all the non standard fields

ThibodeauJF:
@jpmarceau We would probably have to update the parser to gather this doc, but we don't want it made public too much, so I guess it's fine.

packages/headless/src/test/index.ts
ThibodeauJF: Glad not too much stuff was exposed in the end 😅

packages/headless/src/test/mock-engine.ts
```diff
@@ -97,6 +115,10 @@ export interface MockProductListingEngine
   extends ProductListingEngine<ProductListingAppState>,
     MockEngine {}
 
+/**
```
jpmarceau, line 118: Should this one be for internal use only as wel?

jpmarceau (approved):

PR #1717, opened 2022-02-08
SVCC-423

mmitiche:


lvu285:
I would suggest if we fall back to default value, we should add the message on the console as well, so the developer will know what's wrong, why it says invalid number, but it still displays something

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -145,12 +137,20 @@ export default class QuanticCaseClassification extends LightningElement {
...
+      this.field = CoveoHeadlessCaseAssist.buildCaseField(engine, {
+        options: {
+          field: this.coveoFieldName,
+        },
+      });
+      this.unsubscribeField = this.field?.subscribe(() =>
+        this.updateFieldState()
+      );
+    } catch (e) {
+      if (!this.coveoFieldName) {
```
erocheleau, line 150: Why is that if in a try catch and not just before that?
This seems like a required parameter no? I would assert in the constructor and then display a warning and not render anything if the coveoFieldName is not provided.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -338,7 +338,26 @@ export default class QuanticCaseClassification extends LightningElement {
...
+  @api get maxSuggestions() {
+    return this._maxSuggestions;
+  }
+  /**
+   * Set the maximum number of suggestions to be displayed.
+   * @param {number} value - the value to be set.
+   * @returns {void}
+   */
+  set maxSuggestions(value) {
+    if (Number(value) >= 0) {
```
erocheleau, line 357: JavaScript is funny that way ;) 
image

It might be okay lol, this resolves to `0`

erocheleau (approved):
Approved, but fail earlier than this if the coveoFieldValue should be required for your component.

olamothe:
FYI, for this generic problem of "validate properties/input and display warning", there's `@coveo/bueno` in this repo, that we use extensively in Headless and Atomic.

You don't have to use it, but it does some nice validation for types, required arguments, maximum and minimum values for number, etc.

mmitiche:
> FYI, for this generic problem of "validate properties/input and display warning", there's `@coveo/bueno` in this repo, that we use extensively in Headless and Atomic.
> 
> You don't have to use it, but it does some nice validation for types, required arguments, maximum and minimum values for number, etc.

Thanks! I'll take a look👌

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
```diff
@@ -248,8 +249,15 @@ describe('quantic-case-classification', () => {
       });
     });
 
-    it('should display the select dropdown when maxSuggestion is inferior to 0', () => {
-      visitCaseClassification({
+    it('should log a warning in the console and fall back to the default behavour when maxSuggestion is inferior to 0 ', () => {
```
lbergeron, line 252:
```suggestion
    it('should log a warning in the console and fall back to the default behavior when maxSuggestion is inferior to 0 ', () => {
```

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
lbergeron: ...and why are you suppressing all other possible errors?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -338,7 +338,26 @@ export default class QuanticCaseClassification extends LightningElement {
    * @returns {void}
    */
   setFieldValue(value) {
-    this.field.update(value);
+    this.field?.update(value);
```
lbergeron, line 341: The component can't work without its controller. I understand we might not want the component to break the whole page, but the error must be visible somewhere.

lbergeron (changes requested):
Please fix the error handling.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Yes it's a required parameter, I think we can't access the props in the constructor🧐 or am I mistaken?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -123,34 +123,36 @@ export default class QuanticCaseClassification extends LightningElement {
...
+  /** @type {boolean} */
+  _renderingError = false;
 
   connectedCallback() {
     registerComponentForInit(this, this.engineId);
+    if (!this.coveoFieldName) {
+      this._renderingError = true;
+      console.error(
+        'the coveoFieldName is required, please introduce its value'
+      );
```
lbergeron, line 135: I would suggest

> coveoFieldName is required, please set its value

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -357,7 +359,9 @@ export default class QuanticCaseClassification extends LightningElement {
     if (Number(value) >= 0) {
       this._maxSuggestions = Number(value);
     } else {
-      console.warn('Please enter a valid number of maximum suggestions');
+      console.warn(
+        'Invalid number of maximum suggestions introduced, the default value is being used'
```
lbergeron, line 363: The part `the default value is being used` can be misleading as the value is not updated with the default, it is simply not updated when the specified value is smaller than zero.

I would suggest this

    `"${value}" is an invalid maximum number of suggestions. A positive integer was expected. "${this._maxSuggestions}" will be used instead.`

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -131,13 +121,25 @@ export default class QuanticCaseClassification extends LightningElement {
   _value = '';
   /** @type {boolean} */
   _isSuggestionsVisible = true;
+  /** @type {number} */
+  _maxSuggestions = 3;
+  /** @type {boolean} */
+  _renderingError = false;
 
   connectedCallback() {
     registerComponentForInit(this, this.engineId);
```
nathanlb, line 130: Why bother registering the component if it won't be initialized due to the missing attribute?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
nathanlb: FYI we already do [this](https://github.com/coveo/ui-kit/blob/d1a3f3eab52b54837c3f8cce3f1cbca147669c4c/packages/quantic/force-app/main/default/lwc/quanticResultLabel/quanticResultLabel.js#L67) in the result template components so perhaps it would be good to remain consistent.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
nathanlb: I think defaulting back to another value might be too subtle and just unnecessary. This error should only happen when the developer is setting up the page so I think letting the component fail gently while rendering a red error message should be enough to allow them to fix it.

nathanlb (commented):
TBH this error should only happen while the developer is putting the page together so I think I disagree with defaulting back to the other value. The pattern we currently use in the result template components is just to render a red error message where the component is that's obvious and let the dev fix it.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: I agree I'll update this to have something consistent with other components.

nathanlb (approved):
Nice work!

PR #1718, opened 2022-02-08
chore(atomic): migrate to Tailwindcss v3

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1397

Followed https://tailwindcss.com/docs/upgrade-guide

### DIFF
--- a/packages/atomic/package.json
+++ b/packages/atomic/package.json
@@ -105,6 +105,7 @@
     "postcss-focus-visible": "6.0.4",
     "postcss-import": "14.0.2",
     "postcss-mixins": "9.0.1",
+    "postcss-nested": "5.0.6",
     "puppeteer": "9.1.1",
     "react": "16.14.0",
     "rollup-plugin-html": "0.2.1",
--- a/packages/atomic/package.json
+++ b/packages/atomic/package.json
@@ -61,17 +61,15 @@
     "focus-visible": "^5.2.0",
     "i18next": "^20.2.2",
     "i18next-http-backend": "^1.2.2",
-    "stencil-inline-svg": "^1.1.0",
-    "tailwindcss": "3.0.19"
+    "stencil-inline-svg": "^1.1.0"
   },
   "devDependencies": {
     "@babel/core": "7.15.8",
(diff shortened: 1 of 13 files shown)


packages/atomic/stencil.config.ts
```diff
@@ -113,9 +113,9 @@ export const config: Config = {
       plugins: [
         atImport(),
         mixins(),
+        tailwindNesting(),
```
ThibodeauJF, line 116: We have to use `tailwindcss/nesting` & put it before `tailwind()` for it to work, we get errors in the other case

ThibodeauJF: With this, we are getting an issue with mixins & nesting, the following doesn't work (same pattern as used in result templates):
```css
/* Mixin with nesting */
@define-mixin test-mixin {
  input {
    background-color: red;
  }
}

div {
  @mixin test-mixin;
}
```

The fix is to reuse `postcssNesting()` after `tailwind()` like we did before.

packages/atomic/package.json
```diff
@@ -104,13 +102,14 @@
...
+    "postcss-focus-visible": "6.0.4",
     "postcss-import": "14.0.2",
-    "postcss-mixins": "8.1.0",
+    "postcss-mixins": "9.0.1",
     "postcss-nested": "5.0.6",
     "puppeteer": "9.1.1",
     "react": "16.14.0",
-    "rollup-plugin-html": "0.2.1"
+    "rollup-plugin-html": "0.2.1",
+    "tailwindcss": "3.0.19"
```
ThibodeauJF, line 112: not sure why it was a dependency ?

btaillon (approved):

packages/atomic/package.json
olamothe: Well, hum, I think it can indeed be seen as a dev dependency, because it's not actually "shipped" to our clients, like they won't actually download and install tailwind when they `npm install @coveo/atomic`

olamothe (approved):
Really nice !

PR #1722, opened 2022-02-10
docs(atomic-angular): add README for atomic angular 

olamothe:
Add README for atomic angular, with the specific framework "gotcha".

I still need to figure out styling for result templates, which probably means I'll add more to this README in an upcoming PR.

@jpmarceau : Very similar to Atomic React, this will be the things we will want to resurface on our doc website !

https://coveord.atlassian.net/browse/KIT-1406


packages/atomic-angular/README.md
```diff
@@ -1,27 +1,117 @@
...
+## Installation
 
-## Code scaffolding
+`npm i @coveo/atomic-angular`
 
-Run `ng generate component component-name` to generate a new component. You can also use `ng generate directive|pipe|service|class|guard|interface|enum|module`.
+## Usage
 
-## Build
+Since Atomic React is built on top of the core [Atomic](https://docs.coveo.com/en/atomic/latest/) web-components library, the vast majority of concepts that apply to core Atomic will apply to Atomic Angular.
```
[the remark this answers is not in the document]
olamothe: You found my copy paste ;)

btaillon (approved):
Looks good, but I wish we didn't have to pass props as native attributes like this. I wonder if some other Stencil component libraries managed to get around the issue.

Would there be a way to wrap components differently than what Stencil generated, which wouldn't have this issue?

packages/atomic-angular/README.md
```diff
@@ -1,27 +1,117 @@
...
+
+<app-field-label label="My label"> </app-field-label>
+```
+
+In the above example, we can see that we need to annotate `get label()` and `set label()` with the `@Input()` angular decorator.
+
+When the label is set in the app component (`app.component.html`), we arrange for it to be passed down and propagated to `<atomic-text>` using `atomicText.el.setAttribute`. This ensure that that web component will properly receive the property without any modification of the property by the Angular rendering engine.
+
+The reference to `atomicText` is then obtained with the `ViewChild('atomictext')` decorator.
+`atomictext` is the reference set in the HTML template (`<atomic-text #atomictext></atomic-text>`).
```
jpmarceau, line 111:
```suggestion
`atomictext` is the reference set in the `<atomic-text #atomictext></atomic-text>` HTML template.
```

packages/atomic-angular/README.md
```diff
@@ -1,27 +1,117 @@
...
+
+Since that reference will only exists on `ngAfterViewInit`, we need to code defensively against undefined reference.
```
jpmarceau, line 113:
```suggestion
Since that reference will only exist on `ngAfterViewInit`, we code defensively against undefined references.
```

packages/atomic-angular/README.md
```diff
@@ -1,27 +1,142 @@
...
-## Further help
+For example, for any project created with [Angular CLI](https://angular.io/cli), this would mean copying language and icon assets to the root of the source directory
 
-To get more help on the Angular CLI use `ng help` or go check out the [Angular CLI Overview and Command Reference](https://angular.io/cli) page.
+```
+cp -r node_modules/@coveo/atomic-angular/assets src/assets src/assets
+cp -r node_modules/@coveo/atomic-angular/lang src/lang
+```
+
+It is important to respect the folder hierarchy, with SVG icons under the `assets` subdirectory, and labels and languages under the `lang` subdirectory.
```
olamothe, line 32: I've added this section and the `Including default Coveo Theme` section since I had forgotten to talk about those concepts @jpmarceau

packages/atomic-angular/README.md
```diff
@@ -1,27 +1,142 @@
...
+
+Once this is done, these folders must be configured as assets folder in the application.
```
jpmarceau, line 34:
```suggestion
Once this is done, these folders must be configured as asset folders in the application.
```

packages/atomic-angular/README.md
```diff
@@ -1,27 +1,142 @@
...
+
+We can do so using the `angular.json` configuration file.
```
jpmarceau, line 36:
```suggestion
You can do so using the `angular.json` configuration file.
```

packages/atomic-angular/README.md
```diff
@@ -1,27 +1,142 @@
...
+```json
+"build": {
+  // [...] omitted for brevity
+  "options": {
+    "assets": ["src/favicon.ico", "src/assets", "src/lang"],
+  },
+},
+```
+
+## Including default Coveo Theme
```
jpmarceau, line 47:
```suggestion
## Including the Default Coveo Theme
```

packages/atomic-angular/README.md
jpmarceau: question: is using `angular.json` the only way to do this? If so, then I'd just write:
```suggestion
Do so using the `angular.json` configuration file.
```

packages/atomic-angular/README.md
```diff
@@ -1,27 +1,142 @@
...
+
+To include the default Coveo theme, you can use the `angular.json` configuration file.
```
jpmarceau, line 49: Similarly to above, if using the `angular.json` file is the only way to do this, then:
```suggestion
To include the default Coveo theme, use the `angular.json` configuration file.
```

jpmarceau (approved):
👍

PR #1726, opened 2022-02-11
feat(headless): add support for analytics parameter in search API

olamothe:
This will be a larger task than I had originally hoped.

From the JIRA + linked epic, we also need to send the `actionCause` and `customData` to search API.

However, this is all currently internal/blackbox information, sent and manage by the coveo.analytics lib. 

I'll need to do changes to resurface that information in Headless... 😢 

For this PR, as I don't want it to become too large to review, I've left that part as "TODO" (👀   `fromAnalyticsStateToAnalyticsParams` in `configuration-state.ts`).

https://coveord.atlassian.net/browse/KIT-1414


ThibodeauJF (approved):
If only the search api and ua api could talk to each other...

btaillon (approved):

PR #1730, opened 2022-02-14
feat(atomic): add search layout

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1396
design3


packages/atomic/src/components/atomic-layout/__snapshots__/search-layout.spec.ts.snap
```diff
@@ -0,0 +1,43 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ThibodeauJF, line 1: I think that snapshots have a good use case here, to validate the main paths of CSS generation.

packages/atomic/src/components.d.ts
```diff
@@ -248,6 +250,26 @@ export namespace Components {
          */
         "icon": string;
     }
+    interface AtomicLayout {
+        /**
+          * Name of the layout.
```
jpmarceau, line 255:
```suggestion
          * The name of the layout.
```

packages/atomic/src/components.d.ts
```diff
@@ -248,6 +250,26 @@ export namespace Components {
...
+        /**
+          * For column sections, the maximum horizontal space it should take. E.g. '300px'
+         */
+        "maxWidth"?: string;
+        /**
+          * For column sections, the minimum horizontal space it should take. E.g. '300px'
+         */
+        "minWidth"?: string;
+        /**
+          * Name of the layout's section.
```
jpmarceau, line 269:
```suggestion
          * The name of the layout section.
```

packages/atomic/src/components.d.ts
```diff
@@ -1492,6 +1528,26 @@ declare namespace LocalJSX {
          */
         "icon": string;
     }
+    interface AtomicLayout {
+        /**
+          * Name of the layout.
```
jpmarceau, line 1533:
```suggestion
          * The name of the layout.
```

packages/atomic/src/components.d.ts
```diff
@@ -1492,6 +1528,26 @@ declare namespace LocalJSX {
...
+        /**
+          * For column sections, the maximum horizontal space it should take. E.g. '300px'
+         */
+        "maxWidth"?: string;
+        /**
+          * For column sections, the minimum horizontal space it should take. E.g. '300px'
+         */
+        "minWidth"?: string;
+        /**
+          * Name of the layout's section.
```
jpmarceau, line 1547:
```suggestion
          * The name of the layout section.
```

packages/atomic/src/components/atomic-layout-section/atomic-layout-section.tsx
```diff
@@ -0,0 +1,26 @@
+import {Component, Prop} from '@stencil/core';
+import {Section} from './sections';
+
+/**
+ * The `atomic-layout-section` allows to identify various sections for the related `atomic-layout`.
```
jpmarceau, line 5:
```suggestion
 * The `atomic-layout-section` lets you identify various sections for the related `atomic-layout` component.
```

packages/atomic/src/components/atomic-layout-section/atomic-layout-section.tsx
```diff
@@ -0,0 +1,26 @@
...
+ */
+@Component({
+  tag: 'atomic-layout-section',
+  shadow: false,
+})
+export class AtomicLayoutSection {
+  /**
+   * Name of the layout's section.
```
jpmarceau, line 13:
```suggestion
   * The name of the layout section.
```

jpmarceau (approved):
LGTM 👍 
I suppose this will impact https://docs.coveo.com/en/atomic/latest/usage/create-a-result-list ?

packages/atomic/src/components/atomic-layout/__snapshots__/search-layout.spec.ts.snap
olamothe: Instead of comparing the generated CSS, could we try a small POC with a visual comparison plugin for Cypress ?

https://docs.cypress.io/guides/tooling/visual-testing

Nothing too complex, I'm mostly very curious to see if it's not too complicated to setup, and if it's stable enough.

Because that could be the best way to make sure there's no weird CSS interaction.

ThibodeauJF:
@jpmarceau No it's layouts for the components in the page, not the templates which already had layouts.

I'll write a draft doc for it

packages/atomic/src/components/atomic-layout/atomic-layout.pcss
```diff
@@ -0,0 +1,27 @@
+@import './search-layout.pcss';
```
olamothe, line 1: If I understand correctly, we will eventually import all layouts heres (commerce, service, etc), and then the "correct" one will be applied with the layout name.

But all styling for all layouts will always be downloaded by the end user ?

olamothe (approved):

packages/atomic/src/components/atomic-layout/__snapshots__/search-layout.spec.ts.snap
ThibodeauJF: I'll look into it!

packages/atomic/src/components/atomic-layout/atomic-layout.pcss
ThibodeauJF: We could go the separate component's route to scope that out

ThibodeauJF: @olamothe @btaillon perhaps we should relaunch that discussion. Having a separate layout component would only the necessary styles but lose the possibility of quickly switching style with the prop. Keeping it as is (should stay very light has it's simply layout but it scales a lot it might become an issue) would be a big heavier, or perhaps we can dynamically load the styles in another manner than the way Stencil does

btaillon: I'm fine with losing the quick layout switching, I don't think it's worth having to include unused CSS or having to find a workaround.

olamothe: Yeah, it's not mandatory that we can "insta switch" layout without reloading. So it's fine to go with a per component approach

PR #1736, opened 2022-02-15
fix(Quantic): issue managing two document suggestions with the same title fixed

mmitiche:
[SVCC-449](https://coveord.atlassian.net/browse/SVCC-449)

Bug: When having two document suggestions with the same title, opening the second one will open the first one like in the video recording below.

https://user-images.githubusercontent.com/86681870/154108550-f71525f0-6b5a-4767-843f-6d9f19776a0e.mov

Bug Fixed: When having  two document suggestions with the same title, opening the second one will only open the second one and opening the first one will open only the first one. Which is the expected behaviour:

https://user-images.githubusercontent.com/86681870/154109699-501f7b97-60df-4f77-92ca-68436e734b44.mov


packages/quantic/cypress/integration/document-suggestion/document-suggestion.cypress.ts
```diff
@@ -420,4 +421,41 @@ describe('quantic-document-suggestion', () => {
...
+        Expect.displayAccordionSectionContent(true, 0);
+      });
+
+      scope('when clicking on a document suggestion', () => {
+        const clickIndex = 1;
+
+        Actions.clickSuggestion(clickIndex);
+        Expect.logClickingSuggestion(clickIndex, similarDocuments);
+        Expect.displayAccordionSectionContent(true, 0);
+        Expect.displayAccordionSectionContent(true, clickIndex);
```
lbergeron, line 450: Am I understanding correctly that, when many suggestions have the same title, clicking a suggestion automatically unfold all suggestions sharing the same title?

lbergeron:
From the PR description:

> We should handle this case, even if 2 items with the same title, it should work properly

"it should work properly" is not clear to me. Can you please define what is the expected behavior?

packages/quantic/cypress/integration/document-suggestion/document-suggestion.cypress.ts
mmitiche: I just updated the PR description with two video recordings to make it clear what's the bug here and what's the expected behaviour. Also made a small change on the e2e test to make it more clear.

packages/quantic/cypress/integration/document-suggestion/document-suggestion.cypress.ts
```diff
@@ -420,4 +421,43 @@ describe('quantic-document-suggestion', () => {
...
+      scope('when fetching suggestions', () => {
+        mockDocumentSuggestion(similarDocuments);
+        fetchSuggestions();
+        Expect.displayNoSuggestions(false);
+        Expect.displayAccordion(true);
+        Expect.numberOfSuggestions(similarDocuments.length);
+        Expect.displayAccordionSectionContent(false, 0);
+      });
+
+      scope('when clicking on a second document suggestion', () => {
```
lbergeron, line 446:
```suggestion
      scope('when clicking on the second document suggestion', () => {
```

lbergeron (approved):
Thank you for improving the issue description. It makes it easier to validate that your PR fixes the right thing.

aelfaleh (approved):

PR #1737, opened 2022-02-15
feat(atomic): cypress visual testing for layouts

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1419

I used a [free solution](https://www.npmjs.com/package/cypress-image-diff-js) which is functional and based on ["pixelmatch"](https://www.npmjs.com/package/pixelmatch).

@lvu285 told me that the admin ui team is investigating [Percy](https://docs.cypress.io/guides/tooling/visual-testing#Percy) we might move to this paid solution in the future. Something to keep in mind (if you wish, we can try it right away as they have a [free version](https://percy.io/pricing) with up to 5k screenshots per month)


packages/atomic/src/pages/layouts/search.html
```diff
@@ -0,0 +1,69 @@
...
+        const searchInterface = document.querySelector('atomic-search-interface');
+
+        await searchInterface.initialize({
+          accessToken: 'xx564559b1-0045-48e1-953c-3addd1ee4457',
+          organizationId: 'searchuisamples',
+          search: {
+            preprocessSearchResponseMiddleware: (response) => {
+              response.body.totalCountFiltered = 100;
+              const firstResult = response.body.results[0];
+              firstResult.title = 'First result';
```
btaillon, line 24: Since this file mocks results, perhaps we should move it into a sub-directory that contains `test` or `mock`.

btaillon (approved):

packages/atomic/src/pages/layouts/search.html
```diff
@@ -0,0 +1,69 @@
...
+              firstResult.clickUri = 'https://docs.coveo.com/#' + Math.random();
```
olamothe, line 25: Why the math.random() ? 

Isn't this defeating the purpose of always having the exact same result on every run ? It does not seem to serve any purpose at all.

packages/atomic/cypress/integration/layout.cypress.ts
```diff
@@ -0,0 +1,25 @@
+import {RouteAlias, setupIntercept} from '../utils/setupComponent';
+
+describe('Layouts', () => {
+  it('should compare screenshot of a search layout on desktop', () => {
+    cy.viewport('macbook-13');
+    setupIntercept();
+    cy.visit('/layouts/search.html');
+    cy.wait(RouteAlias.analytics);
+    cy.get('atomic-search-interface').compareSnapshot(
```
olamothe, line 9: Have you tried to purposely "make it fail" ?

What does it take for the test to "correctly" catch errors ? It's very important to make sure that the test will correctly catch regression.

What happens if we set the threshold to "pixel perfect" ?

Since we mock results, we mock the viewport, I would pretty much expect every screenshot comparison to be near identical.

packages/atomic/src/pages/layouts/search.html
ThibodeauJF: When testing locally vs the CI, the link can sometimes be "visited" which is a slightly different style (purple text)

olamothe: I see ! 

Alright, can you add a comment to this effect ?

packages/atomic/cypress/integration/layout.cypress.ts
ThibodeauJF: Yes, it's quite easy to make fail, for example just running it on the CI vs locally the slightly different font will make the test fail cause it will change the height of the elements a bit. Like Lucille mentioned to me we might have to rerun the scans if the browser version on the CI changes.

These little things are why perhaps going with the paid Percy might be work it.

I won't merge this PR before trying out the other technology.

ThibodeauJF:
Closing this for now as I'm not entirely sure it won't just be annoying more than anything else.
Maybe when we have multiple & more complex layouts?

PR #1739, opened 2022-02-15
fix(quantic): fix searchbox button style

nathanlb:
Before
Screen Shot 2022-02-15 at 2 42 25 PM

After
Screen Shot 2022-02-15 at 2 42 46 PM

Changed to make custom button as custom css doesn't collaborate well with SLDS LWC. This should keep us from getting messed up from changes introduced by salesforce.
Also added a CSS instruction to keep the styling on iOS, preventing the search input from getting the extra-rounded style.


aelfaleh (approved):
good job

PR #1740, opened 2022-02-15
fix(headless, quantic):  Update analytics events sent when auto selecting the classification with the highest confidence

mmitiche:
[SVCC-456](https://coveord.atlassian.net/browse/SVCC-456)
### What is wrong?
On the first time loading the case classification component (no interaction by user yet), this component already send 2 UA events : `svc_action: "ticket_classification_click"`  & `svc_action: "ticket_field_update"`.
This is happening because from the Quantic side we automatically select the classification with highest confidence when fetching classifications. .

### What is the correct behaviour?

When a value is auto-selected because it has the highest confidence, we should send a "ticket_classification_click" with metadata identifying it as an automatic selection.

#### The Quantic Case Classification component:

https://user-images.githubusercontent.com/86681870/154517620-3da46c16-fb86-4606-9ee9-f951a9c6e330.mov

### DIFF
--- a/packages/headless/src/controllers/case-field/headless-case-field.test.ts
+++ b/packages/headless/src/controllers/case-field/headless-case-field.test.ts
@@ -81,7 +81,7 @@ describe('Case Field', () => {
       };
     });
 
-    it('dispatches a #logClassificationClick action', () => {
+    it('dispatches a #logClassificationClick action when value is a suggestion', () => {
       field.update('suggested value');
 
       expect(engine.actions).toContainEqual(
--- a/packages/headless/src/controllers/case-field/headless-case-field.test.ts
+++ b/packages/headless/src/controllers/case-field/headless-case-field.test.ts
@@ -119,7 +119,7 @@ describe('Case Field', () => {
       );
     });
 
-    it('does not dispatche a #logCaseFieldUpdate analytics action when the autoSelection parameter is set to true', () => {
+    it('does not dispatch a #logCaseFieldUpdate analytics action when the autoSelection parameter is set to true', () => {
       field.update(testValue, undefined, true);
 
       expect(engine.actions).not.toContainEqual(
--- a/packages/headless/src/controllers/case-field/headless-case-field.test.ts
+++ b/packages/headless/src/controllers/case-field/headless-case-field.test.ts
@@ -81,7 +81,7 @@ describe('Case Field', () => {
       };
     });
 
-    it('dispatches a #logClassificationClick action when value is a suggestion', () => {
+    it('dispatches a #logClassificationClick action when value is a suggestion and autoSelection is falsy ', () => {
       field.update('suggested value');
 
       expect(engine.actions).toContainEqual(
@@ -101,6 +101,16 @@ describe('Case Field', () => {
       );
     });
 
+    it('does not dispatch a #logClassificationClick action when value is a suggestion but autoSelection is truthy', () => {
+      field.update(testValue, undefined, true);
+
+      expect(engine.actions).not.toContainEqual(
+        expect.objectContaining({
+          type: 'analytics/caseAssist/classification/click/pending',
+        })
+      );
+    });
+
     it('dispatches a #updateCaseField action with the passed field value', () => {
       field.update(testValue);
 
@@ -109,10 +119,20 @@ describe('Case Field', () => {
       );
     });
 
-    it('dispatches a #logCaseFieldUpdate analytics action', () => {
+    it('dispatches a #logCaseFieldUpdate analytics action when autoSelection is falsy', () => {
       field.update(testValue);
 
       expect(engine.actions).toContainEqual(
+        expect.objectContaining({
+          type: 'analytics/caseAssist/case/field/update/pending',
+        })
+      );
+    });
+
+    it('does not dispatch a #logCaseFieldUpdate analytics action when autoSelection is truthy', () => {
+      field.update(testValue, undefined, true);
+
+      expect(engine.actions).not.toContainEqual(
         expect.objectContaining({
           type: 'analytics/caseAssist/case/field/update/pending',
         })
--- a/packages/headless/src/controllers/case-field/headless-case-field.test.ts
+++ b/packages/headless/src/controllers/case-field/headless-case-field.test.ts
@@ -81,7 +81,7 @@ describe('Case Field', () => {
       };
     });
 
-    it('dispatches a #logClassificationClick action when value is a suggestion and autoSelection is falsy', () => {
+    it('dispatches a #logClassificationClick action', () => {
       field.update('suggested value');
 
       expect(engine.actions).toContainEqual(
@@ -101,16 +101,6 @@ describe('Case Field', () => {
       );
     });
 
-    it('does not dispatch a #logClassificationClick action when value is a suggestion but autoSelection is truthy', () => {
-      field.update(testValue, undefined, true);
-
-      expect(engine.actions).not.toContainEqual(
-        expect.objectContaining({
-          type: 'analytics/caseAssist/classification/click/pending',
-        })
-      );
-    });
-
     it('dispatches a #updateCaseField action with the passed field value', () => {
       field.update(testValue);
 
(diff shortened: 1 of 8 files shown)


erocheleau:
I don't remember the whole discussions around this @ebouchard2 ? Do you remember that we are certain we don't want to log analytics on auto-selection of case field value?
Perhaps a selection event but with metadata around the auto-selection? something like autoSelected: true and confidenceScore: X ?

mmitiche:
> I don't remember the whole discussions around this @ebouchard2 ? Do you remember that we are certain we don't want to log analytics on auto-selection of case field value?
> Perhaps a selection event but with metadata around the auto-selection? something like autoSelected: true and confidenceScore: X ?

I should've mentioned you in this thread: https://coveo.slack.com/archives/GB3EW5BGQ/p1644008004109199
I like your suggestion, sending the  `svc_action: "ticket_classification_click"` will not make a lot of sense since no suggestion was clicked but we can still send the `svc_action: "ticket_field_update"` with metadata around the auto-selection like you suggested.

nathanlb:
> I don't remember the whole discussions around this @ebouchard2 ? Do you remember that we are certain we don't want to log analytics on auto-selection of case field value? Perhaps a selection event but with metadata around the auto-selection? something like autoSelected: true and confidenceScore: X ?

I believe the reasoning behind it was that ML is already be able to tell if the selected result was the one with highest confidence, regardless of auto-selection and this auto-selection event will happen every time at least one classification is returned so there's no value to it that I can see.

ebouchard2:
> > I don't remember the whole discussions around this @ebouchard2 ? Do you remember that we are certain we don't want to log analytics on auto-selection of case field value?
> > Perhaps a selection event but with metadata around the auto-selection? something like autoSelected: true and confidenceScore: X ?
> 
> I should've mentioned you in this thread: https://coveo.slack.com/archives/GB3EW5BGQ/p1644008004109199 I like your suggestion, sending the `svc_action: "ticket_classification_click"` will not make a lot of sense since no suggestion was clicked but we can still send the `svc_action: "ticket_field_update"` with metadata around the auto-selection like you suggested.

I would keep it as a ticket_field_update .

mmitiche:
> > > I don't remember the whole discussions around this @ebouchard2 ? Do you remember that we are certain we don't want to log analytics on auto-selection of case field value?
> > > Perhaps a selection event but with metadata around the auto-selection? something like autoSelected: true and confidenceScore: X ?
> > 
> > 
> > I should've mentioned you in this thread: https://coveo.slack.com/archives/GB3EW5BGQ/p1644008004109199 I like your suggestion, sending the `svc_action: "ticket_classification_click"` will not make a lot of sense since no suggestion was clicked but we can still send the `svc_action: "ticket_field_update"` with metadata around the auto-selection like you suggested.
> 
> I would keep it as a ticket_field_update .

a normal ticket_field_update event or with some metadata around the auto selection?

nathanlb:
> > > > I don't remember the whole discussions around this @ebouchard2 ? Do you remember that we are certain we don't want to log analytics on auto-selection of case field value?
> > > > Perhaps a selection event but with metadata around the auto-selection? something like autoSelected: true and confidenceScore: X ?
> > > 
> > > 
> > > I should've mentioned you in this thread: https://coveo.slack.com/archives/GB3EW5BGQ/p1644008004109199 I like your suggestion, sending the `svc_action: "ticket_classification_click"` will not make a lot of sense since no suggestion was clicked but we can still send the `svc_action: "ticket_field_update"` with metadata around the auto-selection like you suggested.
> > 
> > 
> > I would keep it as a ticket_field_update .
> 
> a normal ticket_field_update event or with some metadata around the auto selection?

What's the potential value provided by this event?

ebouchard2:
> Exactly what @ebouchard2 said, we want to know if it was auto-selected as much as if it was manually selected since auto-selection is a client-side decision.
> 
> > this auto-selection event will happen every time at least one classification is returned so there's no value to it that I can see.
> 
> This is false by the way, it's not because we do it in the cookbook or by default that all customers WANT auto-selection of values. We should be able to know if an auto-selection occurred so I would also change that to a `ticket_field_update` with some metadata saying it was an auto-selection and also maybe the confidence etc.
> 
> @ebouchard2 we would need to update this documentation with our recommendation: https://docs.coveo.com/en/3437/service/log-case-assist-events#fill-a-field
> 
> as well as a clear definition on what the metadata is for an auto-selection:
> 
> If I look at the `ticket_classification_click` we send a ton of useful metadata:
> 
> ```
> coveoua("svc:setAction", "ticket_classification_click", {
>     classificationId: "d8cn3d8-d9n3-s7cb-4knd-d9cnal3njd8c",
>     responseId: "123e4567-e89b-12d3-a456-426614174000",
>     fieldName: "category",
>     classification: {
>         value: "Electrical",
>         confidence: 0.76
>     }
> });
> ```
> 
> Can we take a decision quickly onto how we want to handle this? Just having custom metadata on a `ticket_field_update` is useful for reporting, but potentially not so much for future ML seeing as we would not have the `classificationId`, `responseId`, etc.

When a value is auto-selected because it has the highest confidence, we should send a "ticket_classification_click" with metadata identifying it as an automatic selection. Please also open a doc request to update the documentation page https://docs.coveo.com/en/3437/service/log-case-assist-events .

packages/headless/src/controllers/case-field/headless-case-field.test.ts
```diff
@@ -119,6 +119,16 @@ describe('Case Field', () => {
       );
     });
 
+    it('does not dispatche a #logCaseFieldUpdate analytics action when the autoSelection parameter is set to true', () => {
```
lbergeron, line 122:
```suggestion
    it('does not dispatch a #logCaseFieldUpdate analytics action when the autoSelection parameter is set to true', () => {
```

packages/headless/src/features/case-assist/case-assist-analytics-selectors.test.ts
```diff
@@ -217,6 +217,25 @@ describe('case assist analytics selectors', () => {
...
 
+    it('should return the classification matching the specified ID with the field autoSelection when the autoSelection parameter is set to true', () => {
+      const classification = caseAssistCaseClassificationSelector(
+        buildStateWithCaseField(),
+        'service-suggestion-id',
+        true
+      );
+
+      expect(classification).toMatchObject({
+        autoSelection: true,
```
lbergeron, line 228: Before you can add `autoSelection` here, you'll need to update the [CaseAssistClient](https://github.com/coveo/coveo.analytics.js/blob/master/src/caseAssist/caseAssistClient.ts#L53) in coveo.analytics.js in order to correctly forward the value to the collect event.

lbergeron (changes requested):
The code look good, but `coveo.analytics.js` has to be modified first.

packages/headless/src/features/case-assist/case-assist-analytics-selectors.test.ts
mmitiche: I'm using the  logSelectFieldSuggestion from coveo.analytics.js not logUpdateCaseField.  you mean updating the  SelectFieldSuggestionMetadata interface in the [logSelectFieldSuggestion](https://github.com/coveo/coveo.analytics.js/blob/88676968178236ddbc1365487d96f1ccfcc7e715/src/caseAssist/caseAssistClient.ts#L61) right? 
I'm currently adding the autoSelection information into this metadata.

lbergeron: My bad, I pointed the wrong Case Assist Client method. `logSelectFieldSuggestion` is indeed the method that needs to be modified.

mmitiche: > The code look good, but `coveo.analytics.js` has to be modified first.
coveo.analytics.js Updated👌

lbergeron (approved):
It seems correct to me.

jeremierobert-coveo (approved):

ThibodeauJF (approved):

PR #1742, opened 2022-02-17
feat(atomic): add configurable breakpoint for layouts

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1420


packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -377,7 +379,10 @@ export class AtomicResultList implements InitializableComponent {
     }
   }
 
+  private updateBreakpoints = once(() => updateBreakpoints(this.host));
+
   public render() {
+    this.updateBreakpoints();
```
ThibodeauJF, line 385: During the render is the right time, looking at Stencil's source code, the style is generated & injected right before the render. If we replace it on `componentDidLoad` there's a slight glitch.

packages/atomic/src/components/atomic-result-list/replace-breakpoint.ts
```diff
@@ -0,0 +1,47 @@
...
+  const layout = element.closest('atomic-search-layout');
+  if (!layout) {
+    return;
+  }
+
+  if (layout.breakpoint === DEFAULT_BREAKPOINT) {
+    return;
+  }
+
+  replaceStyleSheet(element, layout.breakpoint);
```
ThibodeauJF, line 45: Depending on the browser's support of Constructed Stylesheets, a different method will be used to inject the styles.

packages/atomic/src/components/atomic-layout/atomic-search-layout.tsx
```diff
@@ -12,8 +12,10 @@ import {buildSearchLayout} from './search-layout';
 })
 export class AtomicSearchLayout {
   @Element() private host!: HTMLElement;
-  // TODO: make breakpoint configurable & reuse in templates
-  private breakpoint = '1024px';
+  /**
+   * CSS value, in px, that defines where the layout goes from mobile to desktop.
+   */
+  @Prop() public breakpoint = 1024;
```
btaillon, line 18: I prefer giving people the option to use the unit they want by passing a string (in case they use `em`/`rem`, for instance).

btaillon (approved):

packages/atomic/src/components/atomic-layout/atomic-search-layout.tsx
olamothe: Is a single breakpoint enough 🤔 

Should we leave room to expand on it in the future for layouts other than search ?

So something like an array, or a complex object might be more appropriate ?

packages/atomic/src/components/atomic-result-list/replace-breakpoint.ts
```diff
@@ -0,0 +1,47 @@
+function replaceMediaQuery(style: string, breakpoint: number) {
+  return style.replace(
+    /\(min-width: 1024px\)/g,
+    `(min-width: ${breakpoint}px)`
+  );
+}
+
+function replaceStyleSheet(element: HTMLElement, breakpoint: number) {
+  // Not adopted by all browsers, not part of Typescript yet
```
olamothe, line 9: You can augment the base `HTMLElement` interface, though. Don't need to cast as `any`.

https://basarat.gitbook.io/typescript/type-system/lib.d.ts#modifying-native-types

packages/atomic/src/components/atomic-result-list/replace-breakpoint.ts
```diff
@@ -0,0 +1,47 @@
...
+  // https://caniuse.com/mdn-api_document_adoptedstylesheets
+  // https://github.com/microsoft/TypeScript/issues/30022
+  const stylesheets = (element.shadowRoot as any).adoptedStyleSheets;
+  if (!stylesheets || !stylesheets.length) {
+    return;
+  }
+
+  const stylesheet = stylesheets[0];
+  const style = Object.values(stylesheet.cssRules)
+    .map((rule: any) => rule.cssText)
```
olamothe, line 19: Same as above, if we augment we base interface we won't need to cast here.

olamothe (approved):

packages/atomic/src/components/atomic-result-list/replace-breakpoint.ts
```diff
@@ -0,0 +1,47 @@
+function replaceMediaQuery(style: string, breakpoint: number) {
+  return style.replace(
+    /\(min-width: 1024px\)/g,
```
olamothe, line 3: Instead of targeting `1024px`, would it be more robust to use a placeholder/variable name in `atomic-result-list.pcss` ?

packages/atomic/src/components/atomic-layout/atomic-search-layout.tsx
ThibodeauJF: since it's now a specific atomic-search-layout component, it won't be a problem to add more for other layouts, I've renamed it to mobile-breakpoint to be sure

PR #1749, opened 2022-02-21
Fix(Headless): prevent caseField and caseInput from reseting when being already registered 

mmitiche:
### The Problem: 
- Calling the **buildCaseField** function of a field that was already registered  will reset its value to an empty string and its suggestions to an empty array.
- Calling the **buildCaseInput** function of a field that was already registered  will reset its value to an empty string.

### The Fix:
- Conditions added to check if the field is registered or not before registering it again.

### DIFF
--- a/packages/headless/src/controllers/case-field/headless-case-field.ts
+++ b/packages/headless/src/controllers/case-field/headless-case-field.ts
@@ -116,7 +116,9 @@ export function buildCaseField(
     return engine.state;
   };
 
-  if (!getState().caseField?.fields?.[options.field]) {
+  const isRegistered = getState().caseField?.fields?.[options.field];
+
+  if (!isRegistered) {
     dispatch(
       registerCaseField({
         fieldName: options.field,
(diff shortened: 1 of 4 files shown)


packages/headless/src/controllers/case-field/headless-case-field.ts
```diff
@@ -112,17 +112,19 @@ export function buildCaseField(
...
-      fieldName: options.field,
-      fieldValue: '',
-    })
-  );
-
   const getState = () => {
     return engine.state;
   };
 
+  if (!getState().caseField?.fields?.[options.field]) {
```
ThibodeauJF, line 119: I'd put this condition in a well name variable like `isRegistered` so we know why this check is there

packages/headless/src/controllers/case-input/headless-case-input.ts
```diff
@@ -95,12 +95,14 @@ export function buildCaseInput(
 
   const fieldName = props.options.field;
 
-  dispatch(
-    updateCaseInput({
-      fieldName: fieldName,
-      fieldValue: '',
-    })
-  );
+  if (!getState().caseInput?.[fieldName]) {
```
ThibodeauJF, line 98: same comment

ThibodeauJF (approved):

lbergeron (approved):

nathanlb (approved):

PR #1750, opened 2022-02-21
feat(atomic): added depends-on feature to facets

btaillon:
https://coveord.atlassian.net/browse/KIT-1411

This PR adds support for facet depends-on. It doesn't contain cypress tests (yet), but I will add them before merging the PR.

# Usage
## Defining in the HTML
We depend on the id of a facet (which defaults to its field). Defining an id isn't mandatory, I'm including it for the sake of the example.

```html
<!-- Will always be rendered. Referred to as the "parent". -->
<atomic-facet facet-id="abc" field="filetype"></atomic-facet>

<!-- Will be rendered if any value is selected in filetype. Referred to as a dependent. -->
<atomic-facet field="objecttype" depends-on-abc></atomic-facet> 

<!-- Will be rendered if "YoutubeVideo" is selected in filetype. Referred to as a dependent. -->
<atomic-numeric-facet field="ytlikecount" depends-on-abc="YoutubeVideo"></atomic-numeric-facet>
```

## Defining in the store (not possible yet)
When we register a dependency with HTML, what Atomic is doing (as of this PR) is adding an object with the following interface to the Atomic store:
```ts
interface FacetDependency {
  parentFacetId: string;
  dependantFacetId: string;
  isDependencyMet: (state: FacetRequestWithType) => boolean;
}

interface FacetRequestWithType {
  type: 'facets' | 'categoryFacets' | 'numericFacets' | 'dateFacets';
  request: FacetRequest | CategoryFacetRequest | NumericFacetRequest | DateFacetRequest;
}
```

We could easily expose a way in `atomic-search-interface` that allows implementers to add a custom `FacetDependency`. 

However, I experimented a bit and found that we would may need dependencies to be present on initialization. If that's the case, we could add support for dependencies in `initialize` like so:
```js
searchInterface.initialize({
  accessToken: '...',
  organizationId: '...',
  customFacetDependencies: [{
    parentFacetId: 'abc',
    dependantFacetId: 'ytviewcount',
    isDependencyMet: ({request}) =>
      request.currentValues.some(({state, value}) => value === 'YouTubeVideo' && state === 'selected')
  }]
});
```

Otherwise, I would prefer to introduce a method that works like so:
```js
searchInterface.addFacetDependency(
  'abc',
  'ytviewcount',
  ({request}) =>
    request.currentValues.some(({state, value}) => value === 'YouTubeVideo' && state === 'selected')
);
```

Opinion? If you like it, I'll make that change in a separate PR.

# What the changes in this PR do
## When a facet is initialized
We add `FacetDependency` objects to the store

## When url manager is initialized
We check which facets should be enabled based on facet dependencies, we enable/disable facets based on that.

This happens even when `reflectStateInUrl` is `false`.

## When url manager's listener is called
We check which facets should be enabled based on facet dependencies, we enable/disable facets based on that.

We need to do this here rather than on any other listener, because we need to prevent invalid hashes from being added to the browser's history.

# Caveats
* Two searches are executed when facets are enabled or disabled by dependsOn
  * The second search uses **the wrong analytics event**
    * IMO, we should only have one analytics event, and it should correspond with the user's initial intent
  * This often causes one search action to be rejected by headless. The outcome in the UI is as desired, but we see a red message in the console.
  * Will be fixed in a separate PR by replacing every function that deselects any amount of facet values by their respective 
* We cannot depend on a numeric facet or date facet using the attribute
  * Could be done with custom facet dependencies.
* We can depend on a category facet, but we depend on a value by name, not path.
  * Could be done with custom facet dependencies.redux action

# Other changes
* I updated `MapProp`. Prior to this PR, it always split attribute values on commas. With this new change, it only splits attribute values when explicitly told to.


packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -52,6 +56,7 @@ export class AtomicSearchInterface {
   private unsubscribeSearchStatus: Unsubscribe = () => {};
   private hangingComponentsInitialization: InitializeEvent[] = [];
   private initialized = false;
+  private ignoreUrlManagerStateChanges = false;
```
btaillon, line 59: While fixing facet dependencies, we don't want the url manager's listener to be called.

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -327,18 +332,18 @@ export class AtomicSearchInterface {
   }
 
   private initUrlManager() {
-    if (!this.reflectStateInUrl) {
-      return;
-    }
```
btaillon, line 332: We now always initialize the url manager.

I could make it so we update facet dependencies on a different listener (e.g., breadcrumb manager) when `reflectStateInUrl` is `false`, but I prefer to consistently use the same listener regardless of the use case.

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -357,7 +362,22 @@ export class AtomicSearchInterface {
...
+      fixFacetDependencies(this.engine!, this.store);
+      this.ignoreUrlManagerStateChanges = false;
+    }
+  }
+
   private updateHash() {
+    if (this.ignoreUrlManagerStateChanges) {
+      return;
+    }
+    this.fixFacetDependencies();
```
btaillon, line 377: Notice here that we don't `return` when facet dependencies are invalid, we just activate the `ignoreUrlManagerStateChanges` flag while fixing dependencies and then we continue.

The reason for this is because fixing invalid facet dependencies doesn't always call `updateHash` a second time. For instance, when we enable facets.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -160,6 +161,8 @@ export class AtomicCategoryFacet
   @Prop() public injectionDepth = 1000;
   // @Prop() public customSort?: string; TODO: KIT-753 Add customSort option for facet
 
+  @MapProp() public dependsOn: Record<string, string> = {};
```
btaillon, line 164: Although logically, changes happen when we change something on the parent facet, not the dependant facet, I still find it more intuitive to define dependencies from the dependant facet (just like JSUI).

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -190,11 +193,17 @@ export class AtomicCategoryFacet
...
-      facetId: this.facetId!,
-      element: this.host,
-    });
+    registerFacetToStore(
+      this.bindings.store,
+      'categoryFacets',
+      {
+        label: this.label,
+        facetId: this.facetId!,
+        facet: this.facet,
```
btaillon, line 202: We now need to pass the facet directly, so we can call `enable`/`disable` on it.

At first, I added this because I also wanted to read the facet's current state, but this is no longer needed since I changed my mind and decided to read the facet request state on the engine instead of the facet controller state.

If you prefer, I can remove this requirement and instead dispatch the `enable`/`disable` actions directly.

If you want to know why I decided to read the facet request state instead of the facet controller state, it's because the state of the url manager corresponds with the state of the request, not the state of the response, so it's much easier to keep in sync this way.

packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -190,11 +193,17 @@ export class AtomicCategoryFacet
...
+        element: this.host,
+      },
+      parseDependsOn(this.dependsOn)
```
btaillon, line 205: A `Record<string, string>` is a pretty generic type, so I decided to parse it here.

I decided not to add dependencies in the object above, because this object directly replaces the facet state in the Atomic store, and I wanted dependencies to be in a separate array.

packages/atomic/src/components/facets/facet-depends-on/facet-dependency-resolver.ts
```diff
@@ -0,0 +1,108 @@
...
+  });
+  return incorrectFacetStates;
+}
+
+export function fixFacetDependencies(
+  engine: SearchEngine,
+  store: ObservableMap<AtomicStore>
+) {
+  const state = store.state;
+  let facetStatesToUpdate = getIncorrectFacetStates(engine, state);
```
btaillon, line 85: Notice here that `facetStatesToUpdate` is not a `const`.

packages/atomic/src/components/facets/facet-depends-on/facet-dependency-resolver.ts
```diff
@@ -0,0 +1,108 @@
...
+        facetValuesWereCleared ||= facetHasSelectedValue(facetWithType);
+      }
+      facetStatesToUpdate = getIncorrectFacetStates(engine, state);
+    });
+  }
+  if (facetWasEnabled || facetValuesWereCleared) {
+    thawAllFacets(engine);
+    engine.dispatch(
+      loadSearchActions(engine).executeSearch(
+        loadSearchAnalyticsActions(engine).logClearBreadcrumbs()
```
olamothe, line 104: 🤔  Why are we executing a search here ? 
Won't it be executed by the selection action performed by the parent/owner facet ?

packages/atomic/src/utils/store.ts
```diff
@@ -2,21 +2,44 @@ import {
...
 }
 
 interface FacetValueFormat<ValueType> {
   format(facetValue: ValueType): string;
   content?(facetValue: ValueType): VNode;
 }
 
+export interface FacetDependency {
+  parentFacetId: string;
+  dependantFacetId: string;
```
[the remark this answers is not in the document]
btaillon: Ah, that's because when I searched them, I thought `dependant` was a noun, but it looks like that's specific to british English. I'll use `dependent` across the board, it's easier.

PR #1752, opened 2022-02-22
feat(headless): add support for analytics parameter in search API

olamothe:
This is a PR I had closed as I had left "TODOs" to support action cause and custom data for analytics.

After an extensive investigation, I was not able to come up with a pattern that would make this a non breaking change in the library.

I've discussed with G.Laporte, and we will instead add support for action cause and custom data in the next major release.

We can still however support other analytics parameters without any breaking changes. So, I've re-opened this one, and removed the TODOs.

https://coveord.atlassian.net/browse/KIT-1414


packages/headless/src/api/analytics/analytics.ts
```diff
@@ -194,6 +194,17 @@ export const getVisitorID = () =>
 
 export const historyStore = new history.HistoryStore();
 
+export const getPageID = () => {
+  const actions = historyStore.getHistory();
+  const lastPageView = actions.reverse().find((action) => {
+    return action.name === 'PageView';
+  });
```
btaillon, line 201: Since this function will be called fairly often and I'm not sure how many pageviews some clients may have logged, I think it may be worth using `lastIndexOf` (or something similar) and skipping `reverse()` for performance.

btaillon (approved):

packages/headless/src/api/analytics/analytics.ts
olamothe: Max page views is 10.

ThibodeauJF: +1 for lastIndexOf

packages/headless/src/features/configuration/configuration-slice.ts
```diff
@@ -99,6 +99,15 @@ export const configurationReducer = createReducer(
         if (!isNullOrUndefined(action.payload.anonymous)) {
           state.analytics.anonymous = action.payload.anonymous;
         }
+        if (!isNullOrUndefined(action.payload.deviceId)) {
+          state.analytics.deviceId = action.payload.deviceId;
+        }
+        if (!isNullOrUndefined(action.payload.deviceId)) {
```
ThibodeauJF, line 105: duplicate

ThibodeauJF (approved):

packages/headless/src/api/analytics/analytics.ts
ThibodeauJF: Oh thought it took a method as a second arg, bad memory.

Good as is then

PR #1753, opened 2022-02-22
ci(atomic-angular): add deploy alpha script

olamothe:
Angular is a bit special, since the `ng cli` tool generates a "project distribution" folder (`packages/atomic-angular/projects/atomic-angular/dist/**`) , which is what needs to be published (not the "root" package.json file at `packages/atomic-angular`)

So I've added a small script to build then modify the generated `package.json` with the needed script to `npm:publish:alpha` like other packages.

https://coveord.atlassian.net/browse/KIT-1427

### DIFF
--- a/packages/atomic-angular/README.md
+++ b/packages/atomic-angular/README.md
@@ -14,6 +14,28 @@ Since Atomic Angular is built on top of the core [Atomic](https://docs.coveo.com
 
 However, there are still some special considerations.
 
+## Importing AtomicAngularModule
+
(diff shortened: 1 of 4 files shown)


packages/atomic-angular/README.md
```diff
@@ -14,6 +14,28 @@ Since Atomic Angular is built on top of the core [Atomic](https://docs.coveo.com
 
 However, there are still some special considerations.
 
+## Importing AtomicAngularModule
```
olamothe, line 17: Some more precision on the usage of the package

btaillon (approved):

jpmarceau (approved):
👌

PR #1755, opened 2022-02-23
fix(atomic): search layout optimization for table results & more

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1429

The table now has an horizontal scroll bar when layout get's smaller than the table

### DIFF
--- a/packages/atomic/src/pages/index.html
+++ b/packages/atomic/src/pages/index.html
@@ -108,7 +108,6 @@
         background-color: transparent;
       }
 
-      /* Layout customization */
       body {
         margin: 0;
       }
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/atomic-layout/__snapshots__/search-layout.spec.ts.snap
```diff
@@ -2,15 +2,17 @@
 
 exports[`Search Layout #buildSearchLayout when the atomic-search-layout component contains a main & facets sections
     should match snapshot 1`] = `
-"atomic-search-layout#abc { display: grid }@media only screen and (min-width: 1024px) {
+"atomic-search-layout#abc { display: grid }
```
ThibodeauJF, line 5: From a maintainer's perpective, I found the snapshot to be much more useful than any visual comparison.

PR #1756, opened 2022-02-23
fix(quantic): fix standalone searchbox init attribute and timeframe facet tests

nathanlb:
The logic to set the initialized attribute was flawed in that it wasn't executing after the rendering of the input element. The proposed solution follows the [LWC recommendation](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_lifecycle_hooks_rendered) for executing code once on a rendering cycle.

Running the tests locally also showed consistent failures in the timeframe facet tests. Curious why it isn't failing elsewhere but it runs successfully locally with these changes

### DIFF
--- a/packages/quantic/cypress/integration/facets/timeframe-facet/timeframe-facet.cypress.ts
+++ b/packages/quantic/cypress/integration/facets/timeframe-facet/timeframe-facet.cypress.ts
@@ -21,13 +21,13 @@ describe('quantic-timeframe-facet', () => {
   const pageUrl = 's/quantic-timeframe-facet';
 
   const validRange = {
-    start: '2000-01-01',
-    end: '2000-12-31',
+    start: '1/1/2000',
+    end: '12/31/2000',
     filter: '2000/01/01@00:00:00..2000/12/31@23:59:59',
   };
   const invalidRange = {
-    start: '2000-12-31',
-    end: '2000-01-01',
+    start: '12/31/2000',
+    end: '1/1/2000',
     filter: '2000/12/31@00:00:00..2000/01/01@23:59:59',
   };
 
@@ -308,7 +308,7 @@ describe('quantic-timeframe-facet', () => {
 
           Expect.numberOfValidationErrors(1);
           Expect.validationError(
-            'Your entry does not match the allowed format yyyy-MM-dd.'
+            'Your entry does not match the allowed format M/d/yyyy.'
           );
         });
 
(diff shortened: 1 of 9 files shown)


packages/quantic/force-app/examples/main/lwc/exampleLayout/exampleLayout.html
```diff
@@ -4,11 +4,11 @@
 
   <div class="slds-grid slds-wrap slds-gutters_direct slds-m-top_large">
     <div class="slds-col slds-size_1-of-1 slds-medium-size_1-of-2 slds-large-size_1-of-2">
-      <lightning-card title="Configuration">
+      <c-quantic-card-container title="Configuration">
```
nathanlb, line 7: The SLDS changed their css for the lightning-card so our example layout was lacking some padding.

packages/quantic/cypress/integration/facets/timeframe-facet/timeframe-facet.cypress.ts
```diff
@@ -21,13 +21,13 @@ describe('quantic-timeframe-facet', () => {
   const pageUrl = 's/quantic-timeframe-facet';
 
   const validRange = {
-    start: '2000-01-01',
-    end: '2000-12-31',
+    start: '1/1/2000',
```
ThibodeauJF, line 24: Maybe it's the timezone on the CI vs locally?

nathanlb: I'm thinking the locale would be the difference yes given that's the only parameter we pass to the salesforce utility for formatting. Kind of annoying.

nathanlb: I'll revert that part of the changeset so we can get the build unblocked and investigate on the side.

nathanlb: I made an addition to the github actions script to use an action from the marketplace to set the timezone. Let's see.

nathanlb: Damn it looks like it's against the rules.

olamothe: Does this work ? https://github.com/cypress-io/cypress/issues/1043#issuecomment-623293020

ThibodeauJF (approved):
👍

mmitiche (approved):

PR #1759, opened 2022-02-23
fix(Quantic): hide error message when fetching new classifications

mmitiche:
## Whats Wrong ?

When the value in the Case Classification Component is required and the user lets the input empty, calling the method reportValidaty  will correctly display the "value missing" error message under the input. The problem is that when the user updates the case information and a fetchClassification call is made returning some suggestions and triggering the auto selection of the suggestion with the highest confidence, the error message is still displayed under the input.

We should hide  the “value missing” error message when auto-selecting  the suggestion with most confidence because the input is no longer empty.

### Before
https://user-images.githubusercontent.com/86681870/155406643-23d61035-277a-456d-b634-1308d7ba643c.mov

### After
https://user-images.githubusercontent.com/86681870/155406511-ce4c27fe-a49d-4102-8850-e6fc307b3090.mov


packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -352,5 +350,8 @@ export default class QuanticCaseClassification extends LightningElement {
   setFieldValue(value) {
     this.field.update(value);
     this._value = value;
+    if(this._errorMessage && value){
```
erocheleau, line 353: Shouldn't you instead re-trigger the reportValidity? Does that make sense?

erocheleau (approved):
Just a different way of doing it maybe?

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Yeah! better 👌

nathanlb (approved):

PR #1760, opened 2022-02-23
chore(quantic): fixing github actions quantic package test workflow

nathanlb:
The github actions workflow that creates an **UNPUBLISHED** package has been broken since 1746. This workflow runs daily and is only to test that the current state of quantic in master can successfully be packaged so we can catch errors before real releases.

Anyway, this is an attempt to fix the issue. From what I can read [here](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) it seems manual or scheduled workflows can only be run when in the default branch so I don't think I can test this before merge. If i'm wrong please let me know.

### DIFF
--- a/.github/workflows/create-quantic-package.yml
+++ b/.github/workflows/create-quantic-package.yml
@@ -48,9 +48,6 @@ jobs:
             })
       - name: Setup
         uses: ./.github/actions/setup
-      - name: Build Quantic project
-        working-directory: ./packages/quantic
-        run: npm run build
       - name: Create package version
         working-directory: ./packages/quantic
         run: |
--- a/.github/workflows/create-quantic-package.yml
+++ b/.github/workflows/create-quantic-package.yml
@@ -46,20 +46,14 @@ jobs:
               repo: context.repo.repo,
               run_id: context.runId
             })
-      - name: Install NPM dependencies
-        working-directory: ./packages/quantic
-        run: |
-          npm i -g typescript
-          npm i -g sfdx-cli
-          npm i
+      - name: Setup
+        uses: ./.github/actions/setup
       - name: Build Quantic project
         working-directory: ./packages/quantic
         run: npm run build
-      - name: Copy static resources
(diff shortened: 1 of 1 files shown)


.github/workflows/create-quantic-package.yml
```diff
@@ -46,20 +46,14 @@ jobs:
...
-        working-directory: ./packages/quantic
-        run: |
-          npm i -g typescript
-          npm i -g sfdx-cli
-          npm i
+      - name: Setup
+        uses: ./.github/actions/setup
       - name: Build Quantic project
         working-directory: ./packages/quantic
         run: npm run build
```
btaillon, line 53: You should no longer need the build step, since it's included in the `setup` action

btaillon (approved):

.github/workflows/create-quantic-package.yml
```diff
@@ -46,20 +46,14 @@ jobs:
...
               run_id: context.runId
             })
-      - name: Install NPM dependencies
-        working-directory: ./packages/quantic
-        run: |
-          npm i -g typescript
-          npm i -g sfdx-cli
-          npm i
+      - name: Setup
+        uses: ./.github/actions/setup
```
btaillon, line 50: One small warning about `setup`: it will install the version of node contained in `.nvmrc`, which is currently the version we tested for Atomic (v16.13.2)

nathanlb: Ok thanks for the heads up. We haven't determined a version that is problematic for Quantic so I would go with this and correct if it causes an issue.

.github/workflows/create-quantic-package.yml
nathanlb: Ah yes perfect

olamothe (approved):

ThibodeauJF (approved):

erocheleau (approved):

PR #1763, opened 2022-02-24
chore(quantic): remove unnecessary check for github token

nathanlb:
https://github.com/coveo/ui-kit/runs/5321152179?check_suite_focus=true
We got farther but this should fix it.. for real this time.

The GITHUB_TOKEN env variable is not necessary for creating unpublished packages.
It's only to create Discussion posts after a successful publish so I deem it non-critical.
If the env variable is missing the final step will fail but the release won't be cancelled outright.


mmitiche (approved):
👌

PR #1765, opened 2022-02-24
chore(quantic): fixing an error with the token key file

nathanlb:
Reverted to a syntax that seemed to work previously.
Still a step forward.


lbergeron (approved):

jeremierobert-coveo (approved):

.github/workflows/create-quantic-package.yml
```diff
@@ -51,6 +51,6 @@ jobs:
       - name: Create package version
         working-directory: ./packages/quantic
         run: |
-          echo "${{ env.SFDX_AUTH_JWT_KEY }}" > server.key
+          echo "$SFDX_AUTH_JWT_KEY" > server.key
```
erocheleau, line 54: It doesn't work with that syntax? `${{ env.SFDX_AUTH_JWT_KEY }}` Just for my info, why was that not working?

erocheleau (approved):

ThibodeauJF (approved):

.github/workflows/create-quantic-package.yml
nathanlb: Good question. I don't know. I know the sfdx command was giving the error because the contents of the server.key file weren't adequate. So I reverted it to something I'm pretty sure worked before. But i don't know what this outputted and I don't need to find out 😅

PR #1766, opened 2022-02-25
chore(quantic): set the key path as the env variable value

nathanlb:
image

The key arg was the file path all along!

My previous failures have led me here. I looked more closely at the code like I should've done the first time.
The env variable needed by the ts deployment script is not the content of the key but the path to the key.
This is the one, folks. This is the one.


.github/workflows/create-quantic-package.yml
```diff
@@ -2,7 +2,7 @@ name: Create Quantic Package
 
 env:
   SFDX_AUTH_JWT_USERNAME: sfdc.integration.devv2.hub@coveo.com
-  SFDX_AUTH_JWT_KEY: ${{ secrets.SFDX_AUTH_PACKAGE_JWT_KEY }}
+  SFDX_AUTH_JWT_KEY: server.key
```
btaillon, line 5: If only one step (or combinable steps) requires those environment variables, then I suggest moving them to that step (similarly to `e2e-quantic/action.yml`).

btaillon (approved):

.github/workflows/create-quantic-package.yml
nathanlb: Yep good tip. thanks

mmitiche (approved):

ThibodeauJF (approved):

PR #1767, opened 2022-02-25
feat(quantic): updating behaviour of QuanticCaseClassification when fetching suggestions

mmitiche:
### Expected behaviour 

1. When changes are made to the description/subject:

- If the default value of the suggestion is changed:  we keep the selection made by the user. (no matter if it’s selected from a suggestion or the dropdown) If they selected it, it was probably good enough.

- If the classifications are still the ones by default : we show the new suggestions based on the changes in the description.

2. When NO changes are made to the description/subject:

- keep what users had selected whether in the dropdown or suggestions.

### Demo:

- #### when selecting an option from the select input and then fetching suggestions, we keep the option selected by the user.
After fetching new suggestions, the loading spinner does not appear.

https://user-images.githubusercontent.com/86681870/155992237-74846d2e-dde9-4934-a616-6397d8884273.mov

- #### when selecting a suggestion and then receiving new suggestions, we keep the suggestions selected by the user.

https://user-images.githubusercontent.com/86681870/155727099-0972c9a5-989d-4bda-bba7-dfb7f5d5f9ab.mov


packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -183,13 +184,14 @@ export default class QuanticCaseClassification extends LightningElement {
           Math.max(Number(this.maxSuggestions), 0)
         ) ?? [];
       this.loading = this.field.state.loading;
```
nathanlb, line 186: The loading spinner still appears even though the value is fixed and won't update. Seems like an unecessary and potentially confusing visual feedback. What do you think about altering the logic so it doesn't appear?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -350,7 +352,40 @@ export default class QuanticCaseClassification extends LightningElement {
...
     this._value = value;
   }
+
+  /**
+   * Indicates whether a value is one of the proposed suggestions.
+   * @param {string} value
+   * @returns {boolean}
+   */
+  isSuggestion(value) {
+    return this.suggestions.find((suggestion) => suggestion.value === value);
```
nathanlb, line 367: `.find()` returns the array element if found. If we're just looking for a boolean value using `.includes()` may be more appropriate.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -326,15 +328,15 @@ export default class QuanticCaseClassification extends LightningElement {
...
+      this.animateHideSuggestions();
     }
   }
 
   /**
    * Hide the suggested options.
    * @returns {void}
    */
-  hideSuggestions() {
+  animateHideSuggestions() {
```
nathanlb, line 339: Why change the name?

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.html
```diff
@@ -18,30 +20,32 @@
...
+                  ">
+                  <input type="checkbox" checked={suggestion.checked} data-suggestion-id={suggestion.id}
+                    onchange={handleSelectSuggestion} id={suggestion.value} value={suggestion.value} name="options" />
+                  <label for={suggestion.value} title={suggestion.value}>
+                    <span class="
+                        slds-visual-picker__figure slds-visual-picker__text
+                        slds-var-p-horizontal_medium
+                        content-center
+                      ">
+                      <span class="slds-truncate">{suggestion.value}</span>
```
nathanlb, line 40: I think this might be slightly too indented.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Absolutely 👌 I'll fix that

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -350,7 +352,40 @@ export default class QuanticCaseClassification extends LightningElement {
...
+   */
+  isSuggestion(value) {
+    return this.suggestions.find((suggestion) => suggestion.value === value);
+  }
+
+  /**
+   * Indicates whether new suggestions have been received.
+   * @returns {boolean}
+   */
+  get newSuggestionsRecieved() {
```
lbergeron, line 374:
```suggestion
  get newSuggestionsReceived() {
```

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
lbergeron: [Array.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) only works with value types. In this case [Array.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) should be used.

lbergeron (commented):
I have not much to add to Nathan's comments.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Fixed👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Fixed👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.html
mmitiche: Fixed👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Fixed👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
mmitiche: Just because I needed to create a property which contains a boolean value to indicate if the suggestions must be hidden or not, the appropriate name for this new property was `hideSuggestions`, so I updated the name of this method to differentiate  them, and `animateHideSuggestions` is the appropriate name of this method since its the method responsible of doing the animation to hide the suggestions.

lbergeron (approved):

Issue #1754, opened 2022-02-23
Issue with labels / values not being interpolated resulting in malformed / missing data


dolphub:
Hi!
Info:

`@coveo/atomic-react": "^0.4.11"` (have tried many version after v3+, including beta)

### Issue
**Seeing non interpolated variables**
I see results and the templating for results is working ok for now, but the facet and the query summary / info has some of the labels / values not interpolated.

![image](https://user-images.githubusercontent.com/3790394/155257122-ca9a3154-d52a-46da-9e5b-7afbb85b81f4.png)

code:

```

<AtomicSearchInterface engine={engine}>
        <Row
          marginTop={5}
          marginBottom={5}
          justifyContent={'center'}
          flexGrow={1}
        >
          <AtomicSearchBox>
            <AtomicSearchBoxQuerySuggestions />
          </AtomicSearchBox>
        </Row>
        <Row display="flex">
          <Col lg={3}>
            <AtomicFacetManager>
              <AtomicFacet field="customer_facing_source" label="Source" />
              <AtomicFacet field="language" label="Language" />
            </AtomicFacetManager>
          </Col>
          <Col lg={9}>
            <Row>
              <Col lg={8}>
                <AtomicQuerySummary enableDuration></AtomicQuerySummary>
                <AtomicBreadbox></AtomicBreadbox>
              </Col>
              <Col justifyContent={'end'} lg={4}>
                <AtomicSortDropdown style={{ justifyContent: 'end' }}>
                  <AtomicSortExpression
                    label="Relevance"
                    expression="relevancy"
                  />
                  <AtomicSortExpression
                    label="Date"
                    expression="date descending"
                  />
                </AtomicSortDropdown>
              </Col>
            </Row>
            <Row justifyContent={'center'}>
              <AtomicNoResults></AtomicNoResults>
              <Divider marginTop={2} />
              <AtomicResultList
                template={(result: any) => <ResultTemplate result={result} />}
              ></AtomicResultList>
            </Row>
            <Row marginTop={5}>
              <Col>
                <AtomicResultsPerPage />
              </Col>
              <Col>
                <AtomicPager />
              </Col>
            </Row>
          </Col>
        </Row>
      </AtomicSearchInterface>

```

### Expected Behavior
**values / labels to be filled in appropriately**

olamothe:
Hello, this is due to the way static assets needs to be made available in the public directory of your app.

[...]

PR #1768, opened 2022-02-25
feat(atomic): allow to specify assets path on search interface

olamothe:
This is in relation with the issue reported here: https://github.com/coveo/ui-kit/issues/1754

This allows to specify the assets (icons) and language path on the search interface.

https://coveord.atlassian.net/browse/KIT-1431


packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -314,6 +330,7 @@ export class AtomicSearchInterface {
       i18n: this.i18n,
       store: this.store,
       interfaceElement: this.host,
+      interfaceComponent: this,
```
olamothe, line 333: This adds a reference to the search interface as a component, instead of only as an HTMLElement.

ThibodeauJF: Why not adding the asset paths to the bindings?

olamothe: > Why not adding the asset paths to the bindings?

That seems too narrow, though ?

As in, I can see being able to reference the search interface component (and all it's props) as more generically useful for a lot more components in the future, than just the asset path themselves.

ThibodeauJF: I'm just worried that would be too "wide". In the future we might have other types of parent interface component with not all the same props/methods.

You could technically just make the bindings the component and have all properties be accessible

olamothe: Good point. It's indeed not generic enough. 

Alright, I can change it to assets path...

What about adding it only to `<atomic-icon asset-path="./foo">` though ? 

You would need to set it as a prop on every icon you use in your interface, but it would be more direct/scoped.

btaillon: Alternatively, it could be in the store.

> What about adding it only to `<atomic-icon asset-path="./foo">` though ?
> 
> You would need to set it as a prop on every icon you use in your interface, but it would be more direct/scoped.

I don't expect someone to use `atomic-icon` outside of a search interface, I think it's more convenient and looks less cluttered to read bindings from `atomic-icon`.

btaillon (approved):
Some components like `atomic-breadbox` use the `import ??? from '.../???.svg'` syntax. Are those replaced with their respective svg content on transpilation?

ThibodeauJF (approved):

olamothe:
> Some components like `atomic-breadbox` use the `import ??? from '.../???.svg'` syntax. Are those replaced with their respective svg content on transpilation?

That's okay since that's a direct import, so it's included when Atomic is built.

What's causing problem is loading assets dynamically (anything that uses `getAssetPath` from Stencil)

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
olamothe: Moved to store

PR #1770, opened 2022-02-25
ci(ci): add manual workflow for package creation 

olamothe:
image

Add a manual workflow that we can trigger to bump a version.

You can decide to skip integration (and unit tests) if needed.

https://coveord.atlassian.net/browse/KIT-1432

### DIFF
--- a/.github/workflows/manual-package.yml
+++ b/.github/workflows/manual-package.yml
@@ -31,7 +31,7 @@ jobs:
       - run: npm run lint:check
   unit-test:
     name: 'Run unit tests'
-    if: inputs.skipUnitTests != 'true'
+    if: ${{ github.event.inputs.skipUnitTests != 'true' }}
     needs: build
     runs-on: ubuntu-latest
     steps:
@@ -41,7 +41,7 @@ jobs:
       - run: npm test
   e2e-atomic-test:
     name: 'Run e2e tests on Atomic'
-    if: inputs.skipAtomicIntegrationTests != 'true'
+    if: ${{ github.event.inputs.skipAtomicIntegrationTests != 'true' }}
     needs: build
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/manual-package.yml
```diff
@@ -0,0 +1,98 @@
...
+          jwtkey: ${{ secrets.SFDX_AUTH_JWT_KEY }}
+  bump-version:
+    name: 'Bump version'
+    needs:
+      - 'lint-check'
+      - 'unit-test'
+      - 'e2e-atomic-test'
+      - 'e2e-quantic-test'
+    runs-on: ubuntu-latest
+    if: always()
```
olamothe, line 76: You cannot have conditional `needs`, so instead, I went with this approach: 

 * `if: always()` so that `bump-version` is always ran, even if a part previous job is skipped.
 * Add steps that verify that each `needs` was not a failure.
 * Net net is that you can skip a job (end to end test) and still bump a version and start a release, but if a job is not skipped, it must be a success.

ThibodeauJF (approved):

.github/workflows/manual-package.yml
```diff
@@ -66,4 +65,34 @@ jobs:
...
+          token: ${{ secrets.ADMIN_GITHUB_TOKEN }}
+          fetch-depth: 0
+      - uses: ./.github/actions/setup
+      - name: Clean directory
+        run: git clean -xfd -e node_modules/ -e .husky
+      - name: Commit & push
+        run: |
+          git config --global user.name "GitHub Actions Bot"
+          git config --global user.email "<>"
+          npm run bump:version
```
btaillon, line 98: I think this should be extracted into a composite action to avoid having to remember two places to update the version bump steps.

Careful, composite actions don't support environment variables, so the GitHub token would need to be passed as an input just like `e2e-atomic`.

btaillon (approved):

PR #1771, opened 2022-02-25
feat(headless): added FacetDependenciesManager

btaillon:
https://coveord.atlassian.net/browse/KIT-1411

Edit: Opening this as a PR, since I think there's a lot of important discussion around this and comments are very constructive.

This PR adds the `FacetDependenciesManager`, which manages dependencies for a single dependent facet.

I chose to manage dependencies for a dependent facet instead of a parent facet because we can't really manage a facet with multiple parents otherwise (there would be a series of enable/disable as opposed to a OR condition applied on each of them). While having multiple parents isn't something we necessarily want to support, I don't think it's too costly to take this approach. If you prefer, I can swap the order and remove support for multiple parents.

I will not merge this PR until UTs and doc are added.


packages/samples/headless-react/src/pages/DependentFacetPage.tsx
```diff
@@ -61,87 +33,47 @@ export class DependentFacetPage extends Component<
...
       options: {field: 'filetype'},
     });
-    this.state = {currentExample: 'SingleParentSingleDependent'};
+  }
+
+  establishDependency(
+    dependentFacet: HeadlessFacet,
+    parentFacet: HeadlessFacet
+  ) {
+    this.dependenciesManagers.push(
```
olamothe, line 42: The second we start introducing "facet dependency" in Headless, then by the very definition of what this means, we must properly support and expose building blocks that correctly represent what users will want to do. 

If what we expose does not exactly represent what our users might want to achieve, then it means we must start adding more options, more corner case so that it works for "everyone".

The fact that you need to use/build multiple facet dependency manager, and that user can only do "one parent to one child, and then loop multiple time and push that in an array" is * not * particularly intuitive. 

As a user, I would think that there's something I must not be doing correctly, or that Headless does not support what I want to achieve.
I would expect it to accept some kind of complex tree as input upfront.

That's the original problem I wanted to solve by not having any code in Headless related to this feature.

Similar to "give a man a fish VS teach a man to fish" :)

The helper function approach can work and we can keep it.

Just that it could also be done in a "sample/look at this demo/copy and adapt this generic function".

packages/samples/headless-react/src/pages/DependentFacetPage.tsx
```diff
@@ -1,46 +1,18 @@
...
   SearchEngine,
 } from '@coveo/headless';
 import {Component} from 'react';
-import {MultilevelDependentFacet} from '../components/dependent-facet/multi-level-dependent-facet';
-import {SingleParentMultipleDependentFacet} from '../components/dependent-facet/single-parent-multiple-dependent.fn';
-import {SingleParentSingleDependentFacet} from '../components/dependent-facet/single-parent-single-dependent.fn';
-import {SingleValueDependentFacet} from '../components/dependent-facet/single-value-dependent.fn';
+import {Facet} from '../components/facet/facet.fn';
 
-const Examples = {
```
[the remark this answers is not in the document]
btaillon: I'm not sure what you mean. With a controller, this is pretty simple, and I don't see an advantage to having multiple example sub-pages. If everything gets collapsed into a single controller/helper however, then I agree and I will definitely restore all those examples.

packages/samples/headless-react/src/pages/DependentFacetPage.tsx
btaillon: > The second we start introducing "facet dependency" in Headless, then by the very definition of what this means, we must properly support and expose building blocks that correctly represent what users will want to do.
>
> If what we expose does not exactly represent what our users might want to achieve, then it means we must start adding more options, more corner case so that it works for "everyone".
> [...]
> That's the original problem I wanted to solve by not having any code in Headless related to this feature.
> [...]
> it could also be done in a "sample/look at this demo/copy and adapt this generic function".

Honestly, after seeing all the potential issues with the url manager and with analytics that dependsOn can have, I think we should make the commitment of providing a helper/controller. Otherwise, I'm very worried the amount of support cases & broken analytics we would get would give us a greater amount of work than the helper/controller would.

> The fact that you need to use/build multiple facet dependency manager, and that user can only do "one parent to one child, and then loop multiple time and push that in an array to achieve complex dependencies" is * not * particularly intuitive.
>
> As a user, I would think that there's something I must not be doing correctly, or that Headless does not support what I want to achieve.
I would expect it to accept some kind of complex tree as input upfront.

I'm not sure I fully agree, maybe what's wrong here is partly the name of the controller. In this PR, the controller represents the dependencies of one facet. To me that's fairly intuitive, I'd like to hear more opinions to know if that may just be me. If we collapse everything into a single helper/controller, then managing something like a facet being removed/replaced becomes complex.

packages/headless/src/controllers/core/facets/facet-dependencies-manager/headless-facet-dependencies-manager.ts
```diff
@@ -0,0 +1,146 @@
...
+export interface FacetDependenciesManagerProps {
+  dependentFacetId: string;
+  dependencies: AnyFacetDependency<AnyFacetValueRequest>[];
+}
+
+export interface FacetDependenciesManager {
+  stopWatching(): void;
+}
+
+export function buildFacetDependenciesManager(
```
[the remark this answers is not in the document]
btaillon: > This is not a controller though, since it does not use `buildController`, does not return the same shape as a controller (for example: a `subscribe` function), even if it follows the `build` pattern, with the `engine` and `props`.
> 
> That's kind of misleading. It's more like a helper function to help you write less code... 🤷

Good point. I guess we shouldn't document it as a controller, and maybe we should even rename it to avoid confusion. Perhaps replacing "build" with "create", or renaming it something like `bindFacetToDependencies`/`unbind()`.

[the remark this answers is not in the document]
btaillon: > This is not a controller though, since it does not use `buildController`, does not return the same shape as a controller (for example: a `subscribe` function), even if it follows the `build` pattern, with the `engine` and `props`.
> 
> That's kind of misleading. It's more like a helper function to help you write less code... 🤷

Good point. I guess we shouldn't document it as a controller, and maybe we should even rename it to avoid confusion. Perhaps replacing "build" with "create", or renaming it something like `bindFacetToDependencies`/`unbind()`.

> Other than that, I also don't see any code in this function that makes it so it * needs * to reside in Headless 🤔 Is there something I'm missing ?
>
> Said differently, what part of this helper function could have not been copy pasted pretty much exactly as is in Atomic, and that would not have worked just as well ?

Yes, that's something I forgot to mention in this PR. Subscribing directly to the engine is something I hadn't tried, and I hadn't thought about the fact that it would be called before the url manager, my bad.

However, as mentioned in my other comment, I think we should still commit to providing some kind of helper.

packages/headless/src/controllers/core/facets/facet-dependencies-manager/headless-facet-dependencies-manager.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (engine.state.facetSet) {
+      Object.entries(engine.state.facetSet).forEach(
+        ([facetId, request]) =>
+          request.freezeCurrentValues &&
+          engine.dispatch(
+            updateFreezeCurrentValues({facetId, freezeCurrentValues: false})
+          )
+      );
+    }
+  };
```
btaillon, line 108: Without thawing facets, some facets won't fetch new values upon being enabled or upon another facet getting disabled/getting its values cleared.

ThibodeauJF: why is thawing only for specific facets?

btaillon: because they're the only ones that can (and need to) be thawed

PR #1773, opened 2022-02-28
fix(headless): JS Doc description added

mmitiche:
JS Doc description added for autoSelect parameter.


olamothe (approved):
Merge this when ready, it is indeed failing all builds on master 😉

PR #1775, opened 2022-02-28
ci(headless,atomic): use github artifacts to store build output on bump

olamothe:
Tentative to remove the "setup" ( `./.github/actions/setup` ) from executing in self hosted runner.

Instead of trying to clone, install dependencies (disk space failure here), and rebuild, use artifacts from the built packages.

I'm not 100% sure this will work 🤷 🎲 

Because of this, I added a manual workflow dispatch to be able to test more easily.

I will remove it once I can confirm that it works properly.

https://coveord.atlassian.net/browse/KIT-1436

### DIFF
--- a/.github/workflows/publishbot.yml
+++ b/.github/workflows/publishbot.yml
@@ -3,10 +3,17 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
+    inputs:
+      deployWithNoVersionBump:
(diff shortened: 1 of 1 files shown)


ThibodeauJF (approved):

btaillon (approved):
All looks good

olamothe:
For the time being, we will stay with Jenkins release process.

PR #1779, opened 2022-03-01
ci(ci): move from veracode to snyk code 

olamothe:
https://coveord.atlassian.net/browse/KIT-1443

### DIFF
--- a/.dcignore
+++ b/.dcignore
@@ -1,0 +1,7 @@
+*.spec.*
+*.test.*
+*.stories.tsx
+scripts
+packages/samples
+.storybook
+cypress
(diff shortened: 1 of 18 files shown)


ThibodeauJF (approved):

btaillon (approved):
Looks like the problem with most of these `ReactSetInnerHtml` errors is mostly just that it doesn't like us using `innerHTML` at all? Maybe we should still look into eliminating `innerHTML` from most of these places

olamothe:
> Looks like the problem with most of these `ReactSetInnerHtml` errors is mostly just that it doesn't like us using `innerHTML` at all? Maybe we should still look into eliminating `innerHTML` from most of these places (even though they are all escaped)

Every place where innerHTML is used is for a good reason, and removing them would mean breaking a useful end user feature.

PR #1784, opened 2022-03-03
feat(atomic): added dependsOn

btaillon:
https://coveord.atlassian.net/browse/KIT-1448

# Diff
The diff for this PR is huge, mostly because a lot of the changes were copy and pasted across all facets (and all their cypress tests), and because this PR is maybe 75%-90% cypress tests.

Here's a list of what I recommend checking (I use a `/` when files are basically copy and pasted from eachother):
* Cypress tests
  * `test-fixture.ts`
  * One of the facet selector files (e.g., `category-facet-selectors.ts`)
  * `facet.cypress.ts`
  * `category-facet.cypress.ts`
  * `color-facet.cypress.ts` / `rating-facet.cypress.ts` / `rating-range-facet.cypress.ts`
  * `numeric-facet.cypress.ts` / `timeframe-facet.cypress.ts`
* MapProp changes (so `MapProp` doesn't always split values on commas)
  * `props-utils.ts`
  * `props-utils.spec.ts`
  * `atomic-field-condition.tsx`
  * `atomic-result-template.tsx`
* Actual code changes (and doc)
  * `index.html`
  * `facet-common.ts` (pretty much contains all the dependsOn logic for Atomic)
  * `atomic-facet.tsx` / `atomic-category-facet.tsx` / `atomic-color-facet.tsx` / `atomic-rating-facet.tsx` / `atomic-range-facet.tsx`
  * `atomic-numeric-facet.tsx` / `atomic-timeframe-facet.tsx`

# Feature
Supported:
* Depending on a facet
  * E.g.: `<atomic-facet depends-on-filetype></atomic-facet>`
* Depending on a facet value
  * E.g.: `<atomic-facet depends-on-filetype="doc"></atomic-facet>`
* Depending on a category facet (the end value, not the full path)
  * E.g.: `<atomic-facet depends-on-geographicalhierarchy="Canada"></atomic-facet>`
* Using an input-only facet as a dependent
  * E.g.,: `<atomic-numeric-facet with-input="integer" number-of-values="0" depends-on-filetype></atomic-numeric-facet>`
* Multi-level dependencies
  * E.g.: `facet-3` depends on `facet-2` and `facet-2` depends on `facet-1`.
  * Deselecting a root facet doesn't cause any additional search nor analytics to be dispatched.

Unsupported:
* Depending on a numeric facet
  * E.g.: `atomic-numeric-facet`, `atomic-rating-facet`, etc.
* Custom conditions
  * Will be supported later, in a new PR.
[...]

PR #1786, opened 2022-03-03
feat(atomic): added dependsOn

btaillon:
https://coveord.atlassian.net/browse/KIT-1448

Follow-up to https://github.com/coveo/ui-kit/pull/1784 because of a rebase. @jpmarceau already reviewed it.

# Diff
The diff for this PR is huge, mostly because a lot of the changes were copy and pasted across all facets (and all their cypress tests), and because this PR is maybe 75%-90% cypress tests.

Here's a list of what I recommend checking (I use a `/` when files are basically copy and pasted from eachother):
* Cypress tests
  * `test-fixture.ts`
  * One of the facet selector files (e.g., `category-facet-selectors.ts`)
  * `facet-depends-on.cypress.ts`
  * `facet.cypress.ts`
  * `category-facet.cypress.ts`
  * `color-facet.cypress.ts` / `rating-facet.cypress.ts` / `rating-range-facet.cypress.ts`
  * `numeric-facet.cypress.ts` / `timeframe-facet.cypress.ts`
* MapProp changes (so `MapProp` doesn't always split values on commas)
  * `props-utils.ts`
  * `props-utils.spec.ts`
  * `atomic-field-condition.tsx`
  * `atomic-result-template.tsx`
* Actual code changes (and doc)
  * `index.html`
  * `facet-common.ts` (pretty much contains all the dependsOn logic for Atomic)
  * `atomic-facet.tsx` / `atomic-category-facet.tsx` / `atomic-color-facet.tsx` / `atomic-rating-facet.tsx` / `atomic-range-facet.tsx`
  * `atomic-numeric-facet.tsx` / `atomic-timeframe-facet.tsx`

# Feature
Supported:
* Depending on a facet
  * E.g.: `<atomic-facet depends-on-filetype></atomic-facet>`
* Depending on a facet value
  * E.g.: `<atomic-facet depends-on-filetype="doc"></atomic-facet>`
* Depending on a category facet (the end value, not the full path)
  * E.g.: `<atomic-facet depends-on-geographicalhierarchy="Canada"></atomic-facet>`
* Using an input-only facet as a dependent
  * E.g.,: `<atomic-numeric-facet with-input="integer" number-of-values="0" depends-on-filetype></atomic-numeric-facet>`
* Multi-level dependencies
  * E.g.: `facet-3` depends on `facet-2` and `facet-2` depends on `facet-1`.
  * Deselecting a root facet doesn't cause any additional search nor analytics to be dispatched.

Unsupported:
* Depending on a numeric facet
  * E.g.: `atomic-numeric-facet`, `atomic-rating-facet`, etc.
* Custom conditions
  * Will be supported later, in a new PR.
* Depending on multiple facets and/or values
  * This would require more changes to `MapProp`.
  * Nothing at the headless level prevents this from working.)


packages/atomic/cypress/integration/facets/category-facet/category-facet-selectors.ts
```diff
@@ -8,49 +8,84 @@ export const BreadcrumbSelectors = {
...
-  valueHighlight: () =>
-    CategoryFacetSelectors.shadow().find('[part="search-highlight"]'),
+  withId(id: string) {
+    return {
+      ...this,
+      shadow() {
+        return cy.get(`${categoryFacetComponent}[facet-id="${id}"]`).shadow();
+      },
+    };
+  },
```
btaillon, line 18: I changed every selector so they can be called either like `CategoryFacetSelectors.shadow()` or `CategoryFacetSelectors.withId('facet-a').shadow()`. This is very important to test dependsOn, since we insert multiple facets. Alternatively, I could make every selector optionally take an `id` as a parameter and pass is down to `shadow`.

packages/atomic/src/components/facets/facet-common.ts
```diff
@@ -23,3 +30,68 @@ export interface FacetValueProps {
...
+
+function isCategoryFacetValueRequest(
+  value: AnyFacetValueRequest
+): value is CategoryFacetValueRequest {
+  return 'children' in value && Array.isArray(value.children);
+}
+
+function isFacetValueRequest(
+  value: AnyFacetValueRequest
+): value is FacetValueRequest | CategoryFacetValueRequest {
```
ThibodeauJF, line 42: Why `FacetValueRequest | CategoryFacetValueRequest` looks like it filters out values with `children`
```suggestion
): value is FacetValueRequest {
```

packages/atomic/src/components/facets/facet-common.ts
```diff
@@ -23,3 +30,68 @@ export interface FacetValueProps {
...
+  return (
+    'value' in value &&
+    typeof value.value === 'string' &&
+    !('children' in value)
+  );
+}
+
+function getSelectedCategoryFacetValueRequest(
+  ancestor: CategoryFacetValueRequest
```
ThibodeauJF, line 51: `ancestor` might be a bit confusing here, when it's recursively calling the method w/ the children. Simply value would be fine.

packages/atomic/src/components/facets/facet-common.ts
```diff
@@ -23,3 +30,68 @@ export interface FacetValueProps {
...
+): CategoryFacetValueRequest | null {
+  if (ancestor.state === 'selected') {
```
ThibodeauJF, line 53: For all `=== 'selected'` checks, maybe `!== 'idle'` is best if we ever expect an `excluded` value in a near/distant future.

packages/atomic/src/components/facets/facet-common.ts
```diff
@@ -23,3 +30,68 @@ export interface FacetValueProps {
...
+        return values.some((value) => {
+          if (isCategoryFacetValueRequest(value)) {
+            const selectedValue = getSelectedCategoryFacetValueRequest(value);
+            if (!selectedValue) {
+              return false;
+            }
+            if (!expectedValue) {
+              return true;
+            }
+            return selectedValue.value === expectedValue;
```
ThibodeauJF, line 81: This logic is repeated for both value types. You could just get the selectedValue/activeValue of either category (recursive) or specific facet (which would return null if idle).

packages/atomic/src/components/facets/atomic-facet/atomic-facet.tsx
```diff
@@ -164,6 +189,9 @@ export class AtomicFacet
     }).validate({
       displayValuesAs: this.displayValuesAs,
     });
+    if (Object.keys(this.dependsOn).length > 1) {
+      throw "Depending on multiple facets isn't supported";
+    }
```
ThibodeauJF, line 194: Could be extracted & reused

packages/atomic/src/utils/props-utils.ts
```diff
@@ -15,20 +17,28 @@ export function MapProp(opts?: MapPropOptions) {
...
     component.componentWillLoad = function () {
-      const prefix = (opts && opts.attributePrefix) || variableName;
-      const variable = this[variableName];
+      const prefix = opts?.attributePrefix ?? variableName;
       const attributes = getElement(this).attributes;
-      mapAttributesToProp(prefix, variable, Array.from(attributes));
+      const map = attributesToStringMap(prefix, Array.from(attributes));
+      this[variableName] = opts?.splitValues
+        ? stringMapToStringArrayMap(map)
+        : map;
```
ThibodeauJF, line 25: Rather than exporting 2 methods for tests purposes, stringMapToStringArrayMap & attributesToStringMap, this piece of logic could be kept in the exported mapAttributesToProp which would be able to handle slipping values or not. stringMapToStringArrayMap & attributesToStringMap would still be as is but not exported

packages/atomic/cypress/integration/facets/category-facet/category-facet-selectors.ts
ThibodeauJF: Adding the feature + the cypress tests + a cypress test refactor in the same PR makes this a pretty big review! 😄

packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
```diff
@@ -566,4 +569,124 @@ describe('Color Facet Test Suites', () => {
...
+          true
+        );
+        CommonFacetAssertions.assertDisplayFacet(
+          ColorFacetSelectors.withId(facetId),
+          true
+        );
+      });
+    });
+
+    describe('with two dependencies', () => {
```
ThibodeauJF, line 669: I think testing for multiple level of dependency is also important

ThibodeauJF (approved):
Pretty nice!

ThibodeauJF:
Did the PR with the dependency manager disappear ?

btaillon:
> Did the PR with the dependency manager disappear ?

Yes, I have a bunch of changes to make to it, I'll reopen it ASAP.

packages/atomic/src/components/facets/facet-common.ts
btaillon: Woops, you're reight.

packages/atomic/src/components/facets/facet-common.ts
btaillon: if `excluded` gets added, then I believe we still want this to only return the selected value or nothing, not an excluded value.

packages/atomic/src/components/facets/facet-common.ts
btaillon: I could, but in this case, I don't think it would be easier to read, only less repetitive.

packages/atomic/cypress/integration/facets/color-facet/color-facet.cypress.ts
btaillon: I forgot to include in the list to review that I added `facet-depends-on.cypress.ts`, which covers multiple level.

PR #1791, opened 2022-03-04
fix(quantic): not displaying suggestions when they are not among all the available options

mmitiche:
[SVCC-466](https://coveord.atlassian.net/browse/SVCC-466)

### The problem:
Currently we always display the suggestions returned by our ML models even if the returned suggestions are not among all the available options retrieved from Salesforce. 

If a user for Example introduces `Priority` as the SF field name and `sfreason` as the Coveo field name, like that we having non corresponding SF field name and Coveo field name. in this case we dont want to display the suggestions cause they are not corresponding to the right field and to inform the user of this problem we want to log warnings to inform him.

### The fix:
Ignore suggestions that aren't part of the possible values retrieved from Salesforce and raise a warning when it happens.

### DIFF
--- a/packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
+++ b/packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
@@ -31,8 +31,12 @@ const invalidMaxSuggestionsError = (value: string | number) => {
 };
 const missingCoveoFieldNameError =
   'coveoFieldName is required, please set its value.';
-const nonCorrespondingSuggestionWarning = (value: string) => {
-  return `The value "${value}" was not found among all the options retrieved from Salesforce. Please confirm that the coveoFieldName corresponds to the correct sfFieldApiName.`;
+const nonCorrespondingSuggestionWarning = (
+  value: string,
+  coveoFieldName: string,
+  sfFieldApiName: string
+) => {
+  return `The value "${value}" was not found among all the options retrieved from Salesforce. Ensure that the Coveo field name "${coveoFieldName}" corresponds to the correct Salesforce field name "${sfFieldApiName}".`;
 };
 
 describe('quantic-case-classification', () => {
@@ -40,6 +44,7 @@ describe('quantic-case-classification', () => {
 
   const sfDefaultField = 'Priority';
   const coveoDefaultField = 'sfpriority';
+  const nonCorrespondingCoveoField = 'sforigin';
   const allOptions = [
     {id: '0', value: 'Very low', label: 'Very low'},
     {id: '1', value: 'Low', label: 'Low'},
@@ -825,7 +830,9 @@ describe('quantic-case-classification', () => {
           stubConsoleWarning(win);
         },
       });
-      configure();
+      configure({
+        coveoFieldName: nonCorrespondingCoveoField,
+      });
 
       scope('when loading the page', () => {
         Expect.numberOfInlineOptions(0);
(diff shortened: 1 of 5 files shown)


packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
```diff
@@ -821,4 +829,42 @@ describe('quantic-case-classification', () => {
...
+      interceptCaseAssist();
+      mockSfPicklistValues(sfDefaultField, allOptions);
+
+      cy.visit(pageUrl, {
+        onBeforeLoad(win) {
+          stubConsoleWarning(win);
+        },
+      });
+      configure({
+        maxSuggestions: defaultMaxSuggestions,
```
nathanlb, line 844: if we pass no argument here would it not default to the same thing?

nathanlb (approved):
Looks good for the most part but a few things to discuss.

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -298,9 +305,24 @@ export default class QuanticCaseClassification extends LightningElement {
...
+      return [];
+    }
+    return this.classifications
+      .filter((suggestion) => {
+        const suggestionIncludedInOptions = this.options.some(
+          (option) => option.value === suggestion.value
+        );
+        if (!suggestionIncludedInOptions) {
+          this.logWarningOnce(
+            `The value "${suggestion.value}" was not found among all the options retrieved from Salesforce. Please confirm that the coveoFieldName corresponds to the correct sfFieldApiName.`
```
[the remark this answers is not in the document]
mmitiche: @acarpentier01 Can you please review this message we want to log in the console when the error described in the description of this PR appears?

aelfaleh (approved):

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -44,8 +44,11 @@ export default class QuanticCaseClassification extends LightningElement {
       console.error('Error getting the picklist values');
     } else {
       this.picklistValues = data;
```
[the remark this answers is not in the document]
mmitiche: Fixed👌

packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
mmitiche: you are right! Fixed👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -392,4 +414,11 @@ export default class QuanticCaseClassification extends LightningElement {
       this.previousClassifications?.[0]?.value === this.field.state.value
     );
   }
+
+  logWarningOnce(message) {
+    if (!this.loggedWarnings[message]) {
```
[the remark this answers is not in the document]
mmitiche: Nice one! thanks👌

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
acarpentier01, line 318:
```suggestion
            `The value "${suggestion.value}" was not found. Ensure that the Coveo field name (`coveoFieldName`) corresponds to the correct Salesforce field name (`sfFieldApiName`).`
```
Is the "among all options retrieved from Salesforce" necessary?

mmitiche: Thank you! Yes the "among all options retrieved from Salesforce" is necessary, it's  the main hint to help the user understand the error.

PR #1797, opened 2022-03-07
feat(headless): reconcile search token config with engine config

olamothe:
Context:

It is possible, through search token configuration ( https://docs.coveo.com/en/56/build-a-search-ui/search-token-authentication ) to enforce a search hub, pipeline, userDisplay name.

Since search token authentication is not mandatory with Coveo, then we also need to expose those configuration with Headless, because those properties are still very important even when using something like SAML, API keys, OAuth token, etc.

Then the problem becomes that if an implementers use search token, configure those properties in the search token, and does not do the same in Headless (aka: make sure the properties match), then you get a weird/confusing behaviour.

What you see in the redux store/what is sent in the body of the search request does not match the authentication header. So you just have to "know" that the Search API will use the Authentication header over what's sent in the body of the request.

This can be even worse for other services, like Analytics, which might straight up reject the event. Not all services in the platform respect the same contract of Authentication header content > Body parameters. So all in all it's very confusing and error prone.

This PR tries to fix that by synchronizing everything, using the content of the JSON web token over what's configured in Headless, and tries to warn when it sees there's a "intentional" mismatch.

And since this is touching a lot of different part of the state (aka: not doable in a single slice reducer), I had to implement the pattern described here for a "root" or "cross slice" reducer: https://redux.js.org/usage/structuring-reducers/beyond-combinereducers

https://coveord.atlassian.net/browse/KIT-1446


packages/headless/src/app/search-engine/jwt-reducer.ts
```diff
@@ -0,0 +1,223 @@
...
+  if (stateValue === tokenValue) {
+    return false;
+  }
+
+  return true;
+};
+
+const decodeJSONWebToken = (token: string): CoveoJSONWebToken | false => {
+  try {
+    const atobImplementation = typeof atob !== 'undefined' ? atob : atobshim;
```
olamothe, line 79: `atob` does not exists in Node environment, you have to use Buffer module.

So instead I just added a new package to shim the implementation when it's needed.

packages/headless/src/app/search-engine/jwt-reducer.ts
```diff
@@ -0,0 +1,219 @@
...
+  return state;
+};
+
+export const jwtReducer: (logger: P.Logger) => Reducer = (logger) => {
+  return (state: SearchAppState, action: AnyAction) => {
+    const jwt = decodeJSONWebToken(state.configuration.accessToken);
+    if (!jwt) {
+      return state;
+    }
+    switch (action.type) {
```
btaillon, line 185: Should we handle (or is it already handled?) when the search token is updated last?

btaillon (approved):

packages/headless/src/app/engine.ts
```diff
@@ -93,6 +94,12 @@ export interface EngineOptions<Reducers extends ReducersMapObject>
    * [Redux documentation on reducers.](https://redux.js.org/glossary#reducer)
    */
   reducers: Reducers;
+  /**
+   * An optional cross slice (or root) reducer that can be combined with the slice reducers.
+   *
+   * [Redux documentation on root reducers.](https://redux.js.org/usage/structuring-reducers/beyond-combinereducers)
+   */
+  crossSlice?: Reducer;
```
ThibodeauJF, line 102: Picky: i'd keep the name reducer, rather than slice for the option. `crossReducer` `rootReducer` is good. We only use slice in the file name, but reducer is the correct term. A slice is the combination or reducer & it's related actions

packages/headless/src/app/reducer-manager.ts
```diff
@@ -1,19 +1,39 @@
...
 }
 
 export function createReducerManager(
   initialReducers: ReducersMapObject
 ): ReducerManager {
   const reducers = {...initialReducers};
+  let crossSlice: Reducer;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const rootReducer: (combined: Reducer<any, AnyAction>) => Reducer = (
```
ThibodeauJF, line 22: Would this work?
```suggestion
  const rootReducer: (combined: ReducersMapObject) => Reducer = (
```

packages/headless/src/app/search-engine/jwt-reducer.ts
```diff
@@ -0,0 +1,219 @@
...
+          ...state.configuration.analytics,
+          userDisplayName: jwt.userDisplayName!,
+        },
+      },
+    };
+  }
+  return state;
+};
+
+export const jwtReducer: (logger: P.Logger) => Reducer = (logger) => {
```
ThibodeauJF, line 179: Why not using the `createReducer` toolkit method like for other reducers?

packages/headless/src/app/search-engine/jwt-reducer.ts
```diff
@@ -0,0 +1,219 @@
...
+    return {
+      ...state,
+      configuration: {
+        ...state.configuration,
+        analytics: {
+          ...state.configuration.analytics,
+          userDisplayName: jwt.userDisplayName!,
+        },
+      },
+    };
```
ThibodeauJF, line 174: The massive spreading would not be necessary if `createReducer`, which leverages immer, was used

ThibodeauJF (approved):

packages/headless/src/app/search-engine/jwt-reducer.ts
olamothe: I'm not sure I understand ? 🤔 Do you mean when the token is renewed ?

packages/headless/src/app/search-engine/jwt-reducer.ts
olamothe: The main one was that we don't really have "the initial search app state" (that is needed by createReducer), just the initial state for every individual slice.

I think it would still work if I just use an empty object and cast to SearchAppState however. Or I could create a initial global search app state function.

I'll try that approach.

packages/headless/src/app/reducer-manager.ts
olamothe: Humm indeed, it would be: 
```
Reducer<StateFromReducersMapObject<ReducersMapObject>, AnyAction>
```

PR #1800, opened 2022-03-08
feat(headless): cap execution count of query at 2

liviarett:
https://coveord.atlassian.net/browse/KIT-1387

### DIFF
--- a/packages/headless/src/features/query/query-actions.ts
+++ b/packages/headless/src/features/query/query-actions.ts
@@ -1,6 +1,6 @@
 import {createAction} from '@reduxjs/toolkit';
 import {validatePayload} from '../../utils/validate-payload';
-import {BooleanValue, StringValue} from '@coveo/bueno';
+import {BooleanValue, NumberValue, StringValue} from '@coveo/bueno';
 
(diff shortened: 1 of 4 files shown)


packages/headless/src/features/query/query-actions.ts
```diff
@@ -12,6 +12,10 @@ export interface UpdateQueryActionCreatorPayload {
    * Whether to interpret advanced [Coveo Cloud query syntax](https://docs.coveo.com/en/1814/searching-with-coveo/search-prefixes-and-operators) in the query.
    */
   enableQuerySyntax?: boolean;
+  /**
+   * Number of times the query has been executed in a row
+   */
+  calls?: number;
```
btaillon, line 18: This feels like a state more than a parameter, I would prefer if it was stores in a the state (e.g.,  in `did-you-mean-state.ts`). Another reason I would rather not have this as a parameter is that actions in Headless (including this one) are publicly exposed.

Additionally, instead of storing the number of calls, I think it may be easier to store "lastQueryWasCorrectedTo?: string" or something similar to that.

PR #1803, opened 2022-03-08
feat(quantic): added searchhub and pipeline params to ssb

nathanlb:
The SSB doesn't permit to set a `searchhub` or `pipeline` to the engine via arguments. A user should be able to set those values to configure query suggestions within the SSB and send analytics to right hub.

Related [Discuss](https://discuss.coveo.com/t/qs-model-and-v2-call-using-seperate-search-hubs-in-standalone-searchbox/7429)


packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
```diff
@@ -76,6 +72,24 @@ export default class QuanticStandaloneSearchBox extends NavigationMixin(
    * @defaultValue '/global-search/%40uri'
    */
   @api redirectUrl = '/global-search/%40uri';
+  /**
+   * The [search hub](https://docs.coveo.com/en/1342/) to use for this Standalone Search Box.
```
nathanlb, line 76: @jpmarceau could you please review these strings? Thanks!

jpmarceau (approved):
Strings look great, I have nothing to say 🥇 👌

olamothe:
@nathanlb : This might interest you in relation to this PR

https://github.com/coveo/ui-kit/pull/1797

nathanlb:
> @nathanlb : This might interest you in relation to this PR
> 
> #1797

Oh yes that is interesting. Thanks for pointing it out!

erocheleau:
@nathanlb What happens if I set the same engineId on both the SSB and the search page component?
Whose searchhub/pipeline wins?
Is there any point in getting the searchhub from the data what was maybe specified in the Apex token generation class?

nathanlb:
> @nathanlb What happens if I set the same engineId on both the SSB and the search page component? Whose searchhub/pipeline wins? Is there any point in getting the searchhub from the data what was maybe specified in the Apex token generation class?

The engineId given to the SSB gets appended with `_standalone` before being used in the SSB's dedicated search interface so you'd really have to want to do it for them to have the same ID. We did that to make the whole redirection work so they're supposed to be two distinct engines. If a user did make both IDs the same on purposeI believe the searchhub and pipeline values given to the SSB would take precedence but I also think the whole thing would work funky or not at all. I would say that's undefined and unsupported territory there.

If there are conflicts between values specified and the markup and the given JWT then you can see in Olivier's linked PR that it would be highlighted.

Does that answer your question?

packages/quantic/cypress/integration/standalone-search-box/standalone-search-box-expectations.ts
```diff
@@ -75,6 +75,13 @@ function standaloneSearchBoxExpectations(
         })
         .logDetail('should log the "searchFromLink" UA event');
     },
+    fetchQuerySuggestWithParams: (
+      params: Record<string, string | number | boolean>,
+      alias: string
+    ) => {
+      cy.wait(alias);
+      cy.logDetail(`should fetch query suggestions with ${params}`);
```
lbergeron, line 83: You should serialize `params` if you want the content to be displayed in the log.

```suggestion
      cy.logDetail(`should fetch query suggestions with ${JSON.stringify(params)}`);
```

packages/quantic/cypress/page-objects/search.ts
```diff
@@ -6,10 +6,18 @@ import {
   // eslint-disable-next-line node/no-unpublished-import
 } from 'cypress/types/net-stubbing';
 
+type RequestParams = Record<string, string | number | boolean | undefined>;
+
 function uaAlias(eventName: string) {
   return `@UA-${eventName}`;
 }
 
+function compare(superset: RequestParams, subset: RequestParams) {
```
lbergeron, line 15: I guess `includes` might be a better name for this function. Typically, `compare` refers to the equality operator.

lbergeron (approved):
It looks pretty nice!

nathanlb:
> Nice 👌 Would it make sense to add the two new params in our [exampleStandaloneSearchbox.html](https://github.com/coveo/ui-kit/blob/91302445dcce770cfff43598d32a69e5984e7e29/packages/quantic/force-app/main/default/lwc/exampleStandaloneSearchbox/exampleStandaloneSearchbox.html#L6-L10) ?

Thanks! Technically yes but frankly I think we should remove that component. It isn't really an example component the same way the exampleSearch is as it isn't a recommended way of using it. It only serves to try the SSB in a localdevserver setting so I don't think we should be packaging it tbh. @erocheleau I think you're the one who added it a while ago. What do you think about removing it?

erocheleau (approved):

PR #1805, opened 2022-03-09
docs(vuejs): Bootstrap vue.js project in samples folder

liviarett:
https://coveord.atlassian.net/browse/KIT-1464?atlOrigin=eyJpIjoiOWI5MDAzZGZhMzMxNDU1M2IwNWE1NWI3ZTg4YzYwNzQiLCJwIjoiaiJ9


packages/samples/vuejs/.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["johnsoncodehk.volar", "johnsoncodehk.vscode-typescript-vue-plugin"]
```
btaillon, line 2: Are those the default recommendations from vue?

packages/samples/vuejs/src/assets/base.css
```diff
@@ -0,0 +1,74 @@
+/* color palette from <https://github.com/vuejs/theme> */
```
olamothe, line 1: It will be important to clean up the base css, icons, logo, fonts, etc that comes "pre installed" with VueJS cli tooling as much as possible before marking this project as "complete".

The goal is to provide a sample with as less clutter as possible so that someone that wants to learn will only see the absolute necessary code that's needed for Atomic + VueJS setup (and not everything that would be needed for a full fledge website).

This can be done later (just a reminder) !

olamothe (approved):

ThibodeauJF:
I think we should start up the project using the official Vue CLI https://cli.vuejs.org/

packages/samples/vuejs/.eslintrc.cjs
```diff
@@ -0,0 +1,15 @@
+/* eslint-env node */
+require('@rushstack/eslint-patch/modern-module-resolution');
+
+module.exports = {
+  root: true,
+  extends: [
+    'plugin:vue/vue3-essential',
+    'eslint:recommended',
+    '@vue/eslint-config-typescript/recommended',
+    '@vue/eslint-config-prettier',
```
btaillon, line 10: Since this uses prettier and people may clone this, I think it would be neat to add prettier to the recommended extensions and to add format on save by default.

btaillon:
> I think we should start up the project using the official Vue CLI [cli.vuejs.org](https://cli.vuejs.org/)
I'd see using something else only if the cli doesn't offer typescript support

If this project wasn't created with the default CLI, I agree.

packages/samples/vuejs/.vscode/extensions.json
liviarett: yeah!

packages/samples/vuejs/.eslintrc.cjs
liviarett: I'm adding that in the next PR!

PR #1808, opened 2022-03-09
docs(vuejs): create simple search page KIT-1465

liviarett:
Created an initial page building on Jean Felix's POC https://coveo.slack.com/archives/G7HD9P48H/p1645546164930599

Screenshot 2022-03-09 at 18 22 28

Added atomic from CDN due to incompatibilities with Rollup as stated here https://coveo.slack.com/archives/G7HD9P48H/p1646844904800519

UPDATE:
The page now looks like this. There are still a few things that don't work the same as the angular one, like some suggestions or facet filters
Screenshot 2022-03-11 at 12 40 44


packages/samples/vuejs/src/components/SearchPage.vue
```diff
@@ -0,0 +1,52 @@
+<script setup lang="ts">
+import {onMounted} from 'vue';
+
+interface InitializationOptions {
+  organizationId: string;
+  accessToken: string;
+}
+
+interface HTMLAtomicSearchInterfaceElement extends HTMLElement {
```
olamothe, line 9: There should be a way to import that type from Atomic package 🤔 

Something like: 
```
import {Components} from '@coveo/atomic';

// [...]

const searchInterface = document.querySelector('atomic-search-interface') as Components.AtomicSearchInterface;
```

olamothe (approved):

packages/samples/vuejs/src/components/SearchPage.vue
liviarett: I'm currently using the CDN version, any ideas how to use it this way?

packages/samples/vuejs/src/main.ts
```diff
@@ -1,4 +1,24 @@
 import {createApp} from 'vue';
 import App from './App.vue';
+import * as components from './components/index.ts';
```
liviarett, line 3: this gives me an error `An import path cannot end with a '.ts' extension. Consider importing './components/index.js' instead.` and if I remove the `.ts` it complains about the imported components, 
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'typeof import("/Users/liviarett/coveo/ui-kit/packages/samples/vuejs/src/components/index")'.
  No index signature with a parameter of type 'string' was found on type 'typeof import("/Users/liviarett/coveo/ui-kit/packages/samples/vuejs/src/components/index")'.
```
what should I do?

packages/samples/vuejs/vite.config.ts
```diff
@@ -6,7 +6,20 @@ import vueJsx from '@vitejs/plugin-vue-jsx';
...
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [vue(), vueJsx()],
+  plugins: [
+    vue({
+      template: {
+        compilerOptions: {
+          isCustomElement: (tag) => {
+            return (
+              tag.startsWith('atomic-') || tag === 'app-label-and-field-value'
```
liviarett, line 15: I was wondering why `app-label-and-field-value` doesn't start with `atomic`?

packages/samples/vuejs/src/components/ResultsList.vue
```diff
@@ -0,0 +1,47 @@
+<script setup lang="ts">
+import {onMounted} from 'vue';
+import resultTemplate from '../templates/result-template.html?raw';
```
liviarett, line 3: I tried sooo many different things, how do you guys feel about this approach?

liviarett: it's the only one that worked, other than what Jean Felix had done before, which is setting the `v-html` of the component to a string - I guess we could do that too to avoid the boilerplate?
```
<atomic-result-list
      fields-to-include="ec_price,ec_rating,ec_images,ec_brand,cat_platform,cat_condition,cat_categories,cat_review_count,cat_color"
      display="grid"
      image-size="large"
      v-html={{resultTemplate}}
    >
    </atomic-result-list>
```

packages/samples/vuejs/src/components/FacetManager.vue
```diff
@@ -0,0 +1,67 @@
+<template>
```
btaillon, line 1: Based on this component's name, I would expect it to take facets as a parameter.

Instead of naming it `FacetManager`, I think we should name it something like `Facets`.

packages/samples/vuejs/src/components/ResultsList.vue
olamothe: So it works if we do this ?:
```
import resultTemplate from '../templates/result-template.html?raw';
[...]
<atomic-result-list v-html={{resultTemplate}} > </atomic-result-list>
```

If so... that does not look too bad ? 🤔 

You have to know the "recipe" of course, but we can document and recommend this approach so that our clients don't have to go through the same process and have to figure it out on their own.

packages/samples/vuejs/src/assets/styles.css
```diff
@@ -0,0 +1,184 @@
+:root {
```
ThibodeauJF, line 1: We added layouts recently https://docs.coveo.com/en/atomic/latest/usage/layouts/ which allows us to remove most of the styling here.

Like in https://github.com/coveo/ui-kit/blob/master/packages/atomic/src/pages/index.html

packages/samples/vuejs/src/components/ResultsList.vue
ThibodeauJF: I think both are different way of doing the same thing: we can't put the template as JSX, we need to insert it as html. I didn't know about the `?raw`, using `v-html` seemed to be the way to go with Vue when I played with it

packages/samples/vuejs/vite.config.ts
olamothe: So the `app-label-and-field-value` that exists in the Angular sample is there to showcase how you can wrap existing Atomic component with "your own", and how you can come up with a pattern to pass properties down to atomic (web) components.

They are not part of Atomic, and we don't necessarily need to replicate them for VueJS.

In the case of the VueJS project, what we should do is see how it looks like if someone wants to create a higher level order component using Atomic ones.

If there's really nothing special to do apart from simply passing down props exactly as is, we don't need to document anything.

For Angular, that part was particularly tricky so we wanted to document and showcase a pattern that works.

packages/samples/vuejs/src/components/ResultsList.vue
btaillon: I like it!

packages/samples/vuejs/src/main.ts
```diff
@@ -1,4 +1,22 @@
...
 
-createApp(App).mount('#app');
+function kebabCase(str: string): string {
```
olamothe, line 5: Given the comment below related to the Angular specific `app-and-field-label`, I'm guessing all this logic is not really needed ?

packages/samples/vuejs/src/assets/styles.css
olamothe: Yeah we'll need to do the job and clean up Atomic React and Atomic Angular samples.

packages/samples/vuejs/src/main.ts
btaillon: Does this work?
```suggestion
import * as components from './components';
```

liviarett: no, it gives me the same error as using index without `.ts`

liviarett: actually, so does changing it to `.js` 😩

packages/samples/vuejs/vite.config.ts
liviarett: aah alright! thanks for the explanation

packages/samples/vuejs/src/components/FacetManager.vue
liviarett: it complains when it's one word only, that's why I added the 'Manager'. But I'll change to `SearchPageFacets`

packages/samples/vuejs/src/assets/styles.css
liviarett: I'll add a separate Jira ticket for this!

btaillon (approved):

ThibodeauJF (approved):

PR #1809, opened 2022-03-10
feat(atomic): add initWithSearchEngine method to be leveraged by atomic-react

olamothe:
The original problem reported was that there was some console errors when initializing atomic-react with "partial" configuration (eg: without specifying all the analytics properties).

Since the way it was done was that `engine.state.configuration` was passed to Atomic, we would get an incorrect engine initialization inside Atomic. 

This PR adds a way to initialize Atomic with an existing engine, instead of always creating a new one internally.

https://coveord.atlassian.net/browse/KIT-1468

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,7 @@
   "ignorePatterns": [
     "node_modules",
     "stencil-generated",
-    "atomic-angular/dist"
+    "dist"
   ],
   "env": {
     "jest": true,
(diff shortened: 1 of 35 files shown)


jpmarceau (approved):
Looks good 👌

ThibodeauJF (approved):
Love this!

We need to talk about that feature in Atomic's usage page as well, it's very useful

PR #1813, opened 2022-03-11
fix(quantic): persisting case classification values

mmitiche:
[SFINT-4425](https://coveord.atlassian.net/browse/SFINT-4425)

### The Problem:
When QuanticCaseClassification doesn't return any suggestions and the user goes back & forth in the flow, the values  selected by the user are not persisted.

https://user-images.githubusercontent.com/86681870/157897575-8f601e6f-16ce-42b6-b4ee-661100c2b54f.mov

### The Fix:
The values of the user are now always persisted when going back and forth in the flow.

https://user-images.githubusercontent.com/86681870/157897746-308255ef-705b-4ed0-a394-513d8cb9bd57.mov


packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -202,6 +202,9 @@ export default class QuanticCaseClassification extends LightningElement {
           this.showSelect();
         }
         this.setFieldValue(value, true);
+      } else {
+        const value = this.field.state.value;
```
nathanlb, line 206:
```suggestion
        this.setFieldValue(this.field.state.value);
```

nathanlb (approved):

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
lbergeron, line 206: The fix itself seems correct, however I don't really understand why hitting the `back` button ends up executing this `else` block.

It seems many small fixes were made in `updateFieldState` and the function could now use some refactoring. My understanding is that, most of that logic depends on whether the field has already been selected. So maybe we should detect this condition first, and then do the rest of the logic.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -185,27 +185,17 @@ export default class QuanticCaseClassification extends LightningElement {
...
-      this.previousClassifications = this.classifications;
-      this.classifications =
-        this.field.state.suggestions.slice(
-          0,
-          Math.max(Number(this.maxSuggestions), 0)
-        ) ?? [];
+    if (!this.lockedState) {
       this.loading = this.field.state.loading;
-      if (this.newSuggestionsReceived) {
+      if (this.maxSuggestions > 0 && this.newSuggestionsReceived) {
```
erocheleau, line 190: You could simplify a bit I think?

```suggestion
const hasNewSuggestions = this.maxSuggestions > 0 && this.newSuggestionsReceived;
const value = (hasNewSuggestions && this.isAutoSelectionNeeded) ? this.classifications[0].value : this.field.state.value;
this.setFieldValue(value, hasNewSuggestions);
this.handleSuggestionsDisplayBehaviour();
```

I am not sure what the 2nd parameter for the `setFieldValue` does, here I am assuming you want to pass it false if there was no new suggestion.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -432,4 +432,22 @@ export default class QuanticCaseClassification extends LightningElement {
       );
     }
   }
+
+  /**
+   * Handles the display behaviour of the suggestions.
+   * @returns {void}
+   */
+  handleSuggestionsDisplayBehaviour() {
```
erocheleau, line 440:
```suggestion
  handleDisplaySuggestions() {
```
Maybe?

erocheleau (approved):
Few comments, also is it possible to add a test for this?

lbergeron (approved):

PR #1815, opened 2022-03-11
fix(quantic): improvement of the display behaviour of the suggestions

mmitiche:
[SFINT-4428](https://coveord.atlassian.net/browse/SFINT-4428)
## What is wrong?
Steps to reproduce:

1. The user types the subject/description of his case.
2. Some suggestions appears and the user selects one of the suggestions.
3. The user updates the subject/description triggering new suggestions that do not include the value selected by the user.
4. The suggestions are not displayed and the value selected by the user is persisted and displayed in the select dropdown.
5. The user updates the subject/description triggering new suggestions that do include the value selected by the user.
6. The value suggestions are still not displayed and the the value selected by the user is still displayed in the select input which is not the behaviour we want.

https://user-images.githubusercontent.com/86681870/158168302-3344522e-2003-4e22-a37e-8fdea69f6690.mov

## What is the correct behaviour?
In the step 6 we want to display again the suggestions and display the value selected by the user as a selected suggestion and not in the select dropdown.

https://user-images.githubusercontent.com/86681870/158168886-103b0a1d-9f85-4d2a-b717-02a7bd7ff9f1.mov

### DIFF
--- a/packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
+++ b/packages/quantic/cypress/integration/case-classification/case-classification.cypress.ts
@@ -634,23 +634,26 @@ describe('quantic-case-classification', () => {
         Expect.displaySelectInput(true);
       });
 
-      scope('when fetching suggestions', () => {
-        const suggestionsCount = 3;
-        const firstSuggestionIndex = 0;
+      scope(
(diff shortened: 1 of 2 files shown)


nathanlb (approved):
LGTM

lbergeron (approved):

PR #1817, opened 2022-03-11
fix(atomic): make layout & refine modal compatible with Stencil's JSX

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1472


packages/atomic/src/components/atomic-in-stencil/atomic-in-stencil.tsx
```diff
@@ -0,0 +1,68 @@
+import {Component, Element, h} from '@stencil/core';
```
ThibodeauJF, line 1: We don't have to keep this, but it was useful to test

packages/atomic/src/utils/store.ts
```diff
@@ -57,8 +57,21 @@ export const registerFacetToStore = <T extends FacetType, U extends string>(
...
+    }
+  }
+  return false;
+}
+
 export const getFacetElements = (store: ObservableMap<AtomicStore>) => {
-  return store.state.facetElements.filter((element) =>
-    document.contains(element)
-  );
+  return store.state.facetElements.filter((element) => isInDocument(element));
```
ThibodeauJF, line 76: Didn't work w/ shadow DOM (prevented possible duplicate facets)
Found an alternative

packages/atomic/src/components/atomic-layout/atomic-search-layout.tsx
```diff
@@ -18,7 +18,7 @@ export class AtomicSearchLayout {
    */
   @Prop({reflect: true}) public mobileBreakpoint = '1024px';
 
-  public connectedCallback() {
+  public componentDidLoad() {
```
ThibodeauJF, line 21: On connectedCallback, the props of the components in Stencil's JSX are not readable even when reflected

btaillon (approved):

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -374,7 +374,7 @@ export class AtomicResultList implements InitializableComponent {
   }
 
   public componentDidRender() {
-    if (this.resultListState.firstSearchExecuted) {
+    if (this.resultListState?.firstSearchExecuted) {
```
olamothe, line 377: Why this change ?

packages/atomic/src/components/atomic-in-stencil/atomic-in-stencil.tsx
```diff
@@ -0,0 +1,87 @@
...
+const template = '<h1>LINK<h1><atomic-result-link></atomic-result-link>';
+
+/**
+ * @internal
+ */
+@Component({
+  tag: 'atomic-in-stencil',
+  shadow: true,
+})
+export class AtomicInStencil {
```
olamothe, line 13: I'm not sure I understand perfectly what the goal is of this component ?

I see it's marked as internal, so it's for "testing" ? 

And also, the sample is put in `packages/atomic/src/pages/examples/stencil.html`. If we want to follow the logic of the rest of the samples/"output target", should we consider Stencil just like React or Angular or VueJS ?

So we could have a `packages/samples/stencil` package, where we showcase implementation of Atomic inside Stencil, with README, code samples, cypress smoke tests, etc.

olamothe, line 13: I'm not sure I understand perfectly what the goal is of this component ?

I see it's marked as internal, so it's for "testing" ? 

And also, the sample is put in `packages/atomic/src/pages/examples/stencil.html`. If we want to follow the logic of the rest of the samples/"output target", should we consider Stencil just like React or Angular or VueJS ?

So we could have a `packages/samples/stencil` package, where we showcase implementation of Atomic inside Stencil, with README, code samples, cypress smoke tests, etc.

ThibodeauJF: Good idea, I'll do that

olamothe (approved):

PR #1821, opened 2022-03-14
fix(atomic): flaky breadcrumb tests

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1479

### DIFF
--- a/packages/atomic/cypress/integration/breadbox/breadbox-assertions.ts
+++ b/packages/atomic/cypress/integration/breadbox/breadbox-assertions.ts
@@ -144,13 +144,6 @@ export function assertDeselectCheckboxFacet(
   });
 }
 
-export function assertDeselectCategoryFacet(index: number) {
-  it('should be deselected after removing from breadcrumb list', () => {
-    deselectBreadcrumbAtIndex(index);
-    CategoryFacetSelectors.activeParentValue().should('not.exist');
-  });
-}
-
 export function assertDeselectLinkFacet(
   FacetWithLinkSelector: FacetWithLinkSelector,
   index: number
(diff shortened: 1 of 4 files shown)


packages/atomic/cypress/integration/breadbox/breadbox-assertions.ts
```diff
@@ -144,13 +144,6 @@ export function assertDeselectCheckboxFacet(
   });
 }
 
-export function assertDeselectCategoryFacet(index: number) {
```
ThibodeauJF, line 147: These checks were kind of duplicates, we already had assertions for validating that the facets were deselected correctly

btaillon (approved):

packages/atomic/cypress/integration/common-assertions.ts
```diff
@@ -3,7 +3,7 @@ import {AriaLiveSelectors} from './aria-live-selectors';
 import {ComponentErrorSelectors} from './component-error-selectors';
 
 export interface ComponentSelector {
-  shadow: () => Cypress.Chainable<JQuery<HTMLElement>>;
+  shadow: () => Cypress.Chainable<JQuery<any>>;
```
olamothe, line 6: Why the change to any ?

olamothe (approved):

packages/atomic/cypress/integration/common-assertions.ts
ThibodeauJF: which is weird since HTMLAtomicCategoryFacetElement extends HTMLStencilElement which in turn extends HTMLElement

PR #1823, opened 2022-03-15
docs(vuejs): import atomic as a package and remove vite

liviarett:
https://coveord.atlassian.net/browse/KIT-1480
this is what it currently looks like:
Screenshot 2022-03-15 at 18 03 18


packages/atomic/stencil.config.ts
```diff
@@ -81,6 +81,8 @@ export const config: Config = {
     }),
     {
       type: 'dist-custom-elements',
+      autoDefineCustomElements: true,
```
olamothe, line 84: I don't think those will be needed 🤔

packages/samples/vuejswithcli/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "vuejswithcli",
```
olamothe, line 2: I created a new package so as to not pollute any changes you might have. I created it using vue cli: https://cli.vuejs.org/

I added a script to `copy:assets` ie: get the proper labels support in the app.

packages/samples/vuejswithcli/public/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
olamothe, line 1: I copy pasted all the `App.vue` / `index.html` , `components` etc from the `packages/samples/vuejs` folder.

packages/samples/vuejswithcli/src/main.ts
```diff
@@ -0,0 +1,15 @@
+import {applyPolyfills, defineCustomElements} from '@coveo/atomic/loader';
+import '@coveo/atomic/dist/atomic/themes/coveo.css';
```
olamothe, line 2: This is nice to add as an example, since this will load the default theme.

An alternative is for people to define all the variables in `style.css`.

liviarett: nice, I guess it was coming automatically via the cdn!

liviarett:
@olamothe thanks very much for your help.
I though vite was the recommended way of building vue, since I followed this guide https://vuejs.org/guide/quick-start.html#with-build-tools. I find it a bit weird that they would have that guide and don't explain the difference between using that and the cli https://cli.vuejs.org/guide/creating-a-project.html.
I will make a note in the readme file about this too.
This should be ready to go now that you fixed it 😊

btaillon (approved):

olamothe (approved):

PR #1833, opened 2022-03-17
docs(vuejs): add smoke tests

liviarett:


.github/workflows/prbot.yml
```diff
@@ -76,6 +76,14 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/setup
       - uses: ./.github/actions/e2e-atomic-angular
+  e2e-vuejs-test:
```
btaillon, line 79: I think we should also add this to `masterbot.yml` and `manual-package.yml`

btaillon (approved):

packages/samples/vuejs/package.json
```diff
@@ -17,8 +17,9 @@
     "@vue/cli-plugin-babel": "~5.0.0",
     "@vue/cli-plugin-typescript": "~5.0.0",
     "@vue/cli-service": "~5.0.0",
-    "typescript": "~4.5.5",
+    "cypress": "^9.5.2",
+    "html-loader": "^3.1.0",
     "ncp": "^2.0.0",
-    "html-loader": "^3.1.0"
+    "typescript": "~4.5.5"
```
ThibodeauJF, line 23: We have typescript in the root package.json devDependencies, we might not need it here I think

ThibodeauJF (approved):

PR #1834, opened 2022-03-17
fix(quantic): showQuickview property marked as deprecated

mmitiche:
### - New hideQuickview property created.

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
@@ -1,4 +1,4 @@
-import {LightningElement, api, track} from 'lwc';
+import { LightningElement, api, track } from 'lwc';
 import {
   registerComponentForInit,
   initializeWithHeadless,
@@ -31,12 +31,16 @@ export default class QuanticDocumentSuggestion extends LightningElement {
    */
   @api engineId;
   /**
-   * Whether or not we want to display the quick view for the document suggestions.
-   * @api
-   * @type {boolean}
-   * @defaultValue `false`
+   * @deprecated
    */
   @api showQuickview = false;
+  /**
+ * Whether or not we want to hide the quick view for the document suggestions.
+ * @api
+ * @type {boolean}
+ * @defaultValue `false`
+ */
+  @api hideQuickview = false;
   /**
    * Whether or not we want to fetch suggestions when initializing this component.
    * @api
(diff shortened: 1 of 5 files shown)


aelfaleh (approved):

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -31,12 +31,16 @@ export default class QuanticDocumentSuggestion extends LightningElement {
    */
   @api engineId;
   /**
-   * Whether or not we want to display the quick view for the document suggestions.
-   * @api
-   * @type {boolean}
-   * @defaultValue `false`
+   * @deprecated
```
nathanlb, line 34: Even if it's marked as deprecated, it still has to work until we remove it. Is that the case?

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
```diff
@@ -31,12 +31,16 @@ export default class QuanticDocumentSuggestion extends LightningElement {
...
    */
   @api showQuickview = false;
+  /**
+ * Whether or not we want to hide the quick view for the document suggestions.
+ * @api
+ * @type {boolean}
+ * @defaultValue `false`
+ */
+  @api hideQuickview = false;
```
lbergeron, line 43: Unfortunately, this is introducing a behavior change for an implementation where the customer does not want the quick view to be shown. Since the `show-quickview` attribute would not be specified, the quickview will appear after the upgrade.

This will need to be clearly documented in the release notes so, hopefully, a customer in this situation would know that the `hide-quickview` attribute needs to be added to preserve the desired behavior.

I'm not sure if there is an easy way to make this change non-breaking.

packages/quantic/force-app/main/default/lwc/quanticDocumentSuggestion/quanticDocumentSuggestion.js
lbergeron: Even though the `show-quickview` attribute is not considered in the code anymore, I think it is covered.

Suppose a customer wants the quickview to be shown (before the change). He would declare the component like this.

```html
<c-quantic-document-suggestion show-quickview />
```
After the upgrade, quickview will be shown because `hide-quickview` is not specified.

mmitiche: that's how it's working right now:
how_it's_working

There is one case where the customer could suffer from a behaviour change and it's the case you mentioned @lbergeron: 
> Unfortunately, this is introducing a behavior change for an implementation where the customer does not want the quick view to be shown. Since the `show-quickview` attribute would not be specified, the quickview will appear after the upgrade.

I don't see a solution to prevent this🧐 the customer will need to set the hide-quickview attribute to get  back to his usual behaviour.
Can we afford to do this change and rely on documentation to make this easy for customers?

PR #1835, opened 2022-03-17
chore(headless): update esbuild

olamothe:
Last time I tried to update esbuild, this broke the integration with Quantic, since there was some changes in how esbuild generates the common js module.

I've modified the "umd" output (which is itself not supported by esbuild) to use an IIFE + global name (CoveoHeadless) + a footer to mimic what UMD would do.

I also had to specify es6 for the esm module, otherwise this would break in Atomic storybook integration.

https://coveord.atlassian.net/browse/KIT-1450

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "concurrently": "6.5.1",
     "cross-fetch": "3.1.5",
     "cz-conventional-changelog": "^3.0.1",
-    "esbuild": "0.14.2",
+    "esbuild": "^0.14.27",
     "esbuild-plugin-alias": "0.2.1",
     "eslint": "7.32.0",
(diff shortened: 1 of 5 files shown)


btaillon (approved):

package.json
```diff
@@ -39,7 +39,7 @@
     "concurrently": "6.5.1",
     "cross-fetch": "3.1.5",
     "cz-conventional-changelog": "^3.0.1",
-    "esbuild": "0.14.2",
+    "esbuild": "^0.14.27",
```
ThibodeauJF, line 42:
```suggestion
    "esbuild": "0.14.27",
```

ThibodeauJF (approved):

PR #1837, opened 2022-03-18
fix(headless): prevent empty query values in actionHistory

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1496

Analytics PR https://github.com/coveo/coveo.analytics.js/pull/330


packages/headless/src/features/search/search-actions.ts
```diff
@@ -342,7 +342,9 @@ const addEntryInActionsHistory = (state: StateNeededByExecuteSearch) => {
   if (state.configuration.analytics.enabled) {
     historyStore.addElement({
       name: 'Query',
-      value: state.query?.q || getQueryInitialState().q,
+      ...(state.query?.q && {
```
ThibodeauJF, line 345: We don't really have to do this anymore

packages/headless/src/api/analytics/analytics.ts
```diff
@@ -199,7 +199,7 @@ export const getPageID = () => {
   const lastPageView = actions.reverse().find((action) => {
     return action.name === 'PageView';
   });
-  if (!lastPageView) {
+  if (!lastPageView?.value) {
     return '';
   }
```
btaillon, line 204: I'm not too sure what's the context for this PR.

Are you just trying to ignore page views which have no value? If that's the case, then I would suggest the following instead:
```suggestion
  const lastPageView = actions.reverse().find((action) => {
    return action.name === 'PageView' && action.value;
  });
  if (!lastPageView) {
    return '';
  }
```

btaillon, line 204: In what context is an empty pageview sent? If an invalid pageview is sent, do we prefer not to include a pageId with requests at all, or do we prefer to send the last valid pageId?

If it's the latter, then I would suggest this:
```suggestion
  const lastPageView = actions.reverse().find((action) => {
    return action.name === 'PageView' && action.value;
  });
  if (!lastPageView) {
    return '';
  }
```

packages/headless/src/features/search/search-actions.ts
btaillon: In this case, I would find it easier to read if we wrote the function like this:
```ts
const addEntryInActionsHistory = (state: StateNeededByExecuteSearch) => {
  if (state.configuration.analytics.enabled) {
    const element: TYPEHERE = {
      name: 'Query',
      time: JSON.stringify(new Date()),
    };
    if (state.query?.q) {
      element.value = state.query.q;
    }
    historyStore.addElement(element);
  }
};
```

olamothe: Or:

```
if (state.configuration.analytics.enabled && state.query.q) {
  historyStore.addElement({...})
}
```

olamothe (approved):

packages/headless/src/features/search/search-actions.ts
ThibodeauJF: No we still need to send the Query even if the value is empty

PR #1843, opened 2022-03-21
feat(headless): added feedback modal logic & analytics

btaillon:
https://coveord.atlassian.net/browse/KIT-1486


packages/headless/src/controllers/smart-snippet/headless-smart-snippet.ts
```diff
@@ -99,13 +128,14 @@ export function buildSmartSnippet(engine: SearchEngine): SmartSnippet {
     get state() {
       const state = getState();
       return {
-        disliked: state.questionAnswering.disliked,
-        liked: state.questionAnswering.liked,
-        expanded: state.questionAnswering.expanded,
-        answerFound: state.search.response.questionAnswer.answerSnippet !== '',
```
btaillon, line 105: These parts of the state weren't removed, I reordered them to match the order of their definition in the SmartSnippetState interface.

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -1,10 +1,31 @@
...
 import {validatePayload} from '../../utils/validate-payload';
 import {AnalyticsType, makeAnalyticsAction} from '../analytics/analytics-utils';
 import {
   documentIdentifierPayloadDefinition,
   QuestionAnsweringDocumentIdActionCreatorPayload,
 } from './question-answering-document-id';
 
+type AnalyticsSmartSnippetFeedbackReason = Parameters<
+  CoveoSearchPageClient['logSmartSnippetFeedbackReason']
+>[0];
```
btaillon, line 11: Had lots of trouble with this. The `SmartSnippetFeedbackReason` enum isn't exported from `coveo.analytics`. I'll open a PR to export it in `coveo.analytics`.

olamothe: I think another solution would have been to return the string values `as const`.

I think TypeScript would have correctly detected that since they are constant, they match the enum.

olamothe (changes requested):

packages/headless/src/controllers/smart-snippet/headless-smart-snippet.ts
```diff
@@ -125,6 +155,22 @@ export function buildSmartSnippet(engine: SearchEngine): SmartSnippet {
       engine.dispatch(logDislikeSmartSnippet());
       engine.dispatch(dislikeSmartSnippet());
     },
+    openFeedbackModal() {
```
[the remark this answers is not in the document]
btaillon: What's the difference between this and the `like`/`dislike` state?

If we remove the state, do we keep this method for the purpose of logging analytics, or just export the actions?

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
btaillon: I already tried something similar. Unfortunately, it doesn't. Here's what it shows with `as const`:
image

packages/headless/src/controllers/smart-snippet/headless-smart-snippet.ts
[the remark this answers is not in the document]
olamothe: Good point about the like/dislike...

I went into the old design doc for smart snippet, and here's what I found.

We added it to encourage/"force" people to do it to log analytics so we get some raw material for ML. So that users read the controller API and think about implementing it in their own UI.

So with that aspect in mind, then I guess it makes sense to have the same logic for the feedback mechanism... Even if it's purely UI. 

So, I take back my comment then, and we can leave it in Headless. It's good that it's there if it helps encourage best practices.

jpmarceau (approved):
💯 🥇

ThibodeauJF (approved):
🔥

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -1,10 +1,31 @@
...
+  | 'was_not_a_question';
+
+function parseSmartSnippetFeedbackReason(
+  feedback: SmartSnippetFeedback
+): AnalyticsSmartSnippetFeedbackReason {
+  switch (feedback) {
+    case 'does_not_answer':
+    case 'partially_answers':
+    case 'was_not_a_question':
+      return feedback as AnalyticsSmartSnippetFeedbackReason;
```
liviarett, line 25: shouldn't there be a default case here? and if not, what's the difference between doing this and simply 
```js
function parseSmartSnippetFeedbackReason(
  feedback: SmartSnippetFeedback
): AnalyticsSmartSnippetFeedbackReason {
      return feedback as AnalyticsSmartSnippetFeedbackReason;
}
```

packages/samples/headless-react/src/components/smart-snippet/smart-snippet.class.tsx
```diff
@@ -50,13 +51,78 @@ export class SmartSnippet extends Component<{}, SmartSnippetState> {
...
+      feedbackModalOpened,
+    } = this.state;
 
     if (!answerFound) {
       return <div>Sorry, no answer has been found for this query.</div>;
     }
 
+    if (feedbackModalOpened) {
+      return (
+        <div role="dialog">
```
liviarett, line 70: how do we do the styling of these?

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
btaillon: It's equivalent, yes. I forgot to leave a comment about this.

It's a leftover from when I tried to directly use the enum from `coveo.analytics.js`, but I got an error doing that since I was importing it from the `dist` folder. Upon changing course to instead return the string directly, I decided to keep this function for when the enum would be exported (which it is now, so I'll have to update this PR).

packages/samples/headless-react/src/components/smart-snippet/smart-snippet.class.tsx
btaillon: we don't do styling on the `headless-react` package, it's purely meant to be code that is read by developers.

jpmarceau (approved):
Docs is perfect 👌 💯

PR #1844, opened 2022-03-22
fix(headless): made question answer reset its state when changed

btaillon:
https://coveord.atlassian.net/browse/KIT-1503

Prior to this change, you could press "like" on a smart snippet, and after doing another search and getting a new snippet, the new snippet would also be marked as "liked".


olamothe (approved):
I would just remove the "hash" phrasing, personal preference.

I think `identifier` is a better description of what it represent.

PR #1848, opened 2022-03-23
feat(atomic): added SmartSnippetFeedbackBanner

btaillon:
https://coveord.atlassian.net/browse/KIT-1488

The "explain why" modal will be added in a separate PR.

Demo: https://staticdev.cloud.coveo.com/atomic/pr/KIT-1488-b-storybook/index.html?path=/story/atomic-smart-snippet--default-smart-snippet


packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-feedback-banner.tsx
```diff
@@ -0,0 +1,105 @@
...
+      class="flex flex-wrap items-center gap-x-4 text-sm leading-4 h-4 overflow-hidden"
+    >
+      <div
+        part="feedback-inquiry-and-buttons"
+        role="radiogroup"
+        aria-labelledby={inquiryId}
+        aria-controls={thankYouId}
+        class="inline-flex gap-x-4"
+      >
+        <Inquiry></Inquiry>
```
btaillon, line 97: I saw lots of example where the radiogroup contains its title (the inquiry, in this case).

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.pcss
```diff
@@ -20,3 +20,13 @@ footer {
...
+
+[part='feedback-thank-you'] {
+  &:before {
+    content: ' ';
+    display: inline-block;
+    width: 1px;
+    background-color: var(--atomic-neutral);
+    margin-right: 1rem;
+  }
+}
```
btaillon, line 32: If you prefer, I can create a dedicated file for the feedback banner styling

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-feedback-banner.tsx
```diff
@@ -0,0 +1,105 @@
...
+  onLike(): void;
+  onDislike(): void;
+}
+
+export const SmartSnippetFeedbackBanner: FunctionalComponent<
+  SmartSnippetFeedbackBannerProps
+> = (props) => {
+  const inquiryId = 'feedback-inquiry-' + props.id;
+  const thankYouId = 'feedback-thank-you-' + props.id;
+  const radioGroupName = 'feedback-options-' + props.id;
```
btaillon, line 21: I decided to make this a functional component since it doesn't really have a state of its own, and unlike the expandable answer, doesn't need to render multiple times.

However, a side effect of this, is that `randomID()` would be executed multiple times and would keep changing the ID & group name of various components, which I fear could affect screen readers. My solution was to take a `id` as a prop.

jpmarceau (approved):
Strings are excellent 👌

ThibodeauJF (approved):
LGTM!

PR #1852, opened 2022-03-23
doc(atomic): add @disableStorybook tag which allows to disable props in Storybook, disable "conditions"

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1509


olamothe (approved):

npushkarskii (approved):
LGTM! 👍

packages/atomic/src/components.d.ts
```diff
@@ -188,7 +188,8 @@ export namespace Components {
     }
     interface AtomicFieldCondition {
         /**
-          * A list of conditions that must be fulfilled for this template to be selected.
+          * A function that must return true on results for the result template to apply.  For example, a template with the following condition only applies to results whose `title` contains `singapore`: `[(result) => /singapore/i.test(result.title)]`
+          * @disableStorybook
```
btaillon, line 192: Is this doc tag visible from the jsdoc/tooltips when using React, Angular or Vue?

If so, I would prefer to either add a script to remove this doc tag from those exports, or find an alternative way of declaring that a prop should not be included in storybook.

olamothe (approved):

packages/atomic/src/components.d.ts
ThibodeauJF: Ah crap it's true... maybe looking at the type is a better solution

btaillon (approved):
I like this solution

PR #1853, opened 2022-03-24
feat(atomic): add folded results component

liviarett:
This is what the example will look like currently
Screenshot 2022-03-28 at 15 40 51

This is what it looks like without the folded component (same as before)
Screenshot 2022-03-28 at 15 58 49

This is what it looks like if you add a children component in a non-folded result list
Screenshot 2022-03-28 at 16 06 33


packages/atomic/src/components/atomic-result-template/atomic-result-template.tsx
```diff
@@ -116,7 +118,37 @@ export class AtomicResultTemplate {
...
   private getContent() {
-    return this.getTemplateElement().content;
+    // TODO: reorganise
+    const childTemplatesBySection: HTMLAtomicResultChildrenTemplateElement[][] =
+      [];
+    const content = this.getTemplateElement().content;
+    const childrenTemplates = content.querySelectorAll(
+      'atomic-result-children-template'
+    );
+    childrenTemplates.forEach((template) => {
```
liviarett, line 128: this might have to change to
```js
    const childrenTemplates = Array.from(
      content.querySelectorAll('atomic-result-children-template')
    ).filter((template) => {
      return !elementHasAncestorTag(
        template,
        'atomic-result-children-template'
      );
    });
```
to make sure we don't add nested children templates to be used on this level of children

packages/atomic/src/components/atomic-children-result-list/atomic-children-result-list.tsx
```diff
@@ -0,0 +1,64 @@
...
+  private constructor() {
+    this.host.removeAttribute('result-children');
+  }
+
+  public async initialize() {
+    this.childResultsTemplatesManager = buildResultTemplatesManager(
+      this.bindings.engine
+    );
+    await this.registerChildrenResultTemplates();
+    this.ready = true;
```
btaillon, line 37: Instead of rendering twice using a state attribute, perhaps we could make `initialization-utils` await the `initialize` function. `initialization-utils` executes `initialize` inside [componentWillLoad](https://stenciljs.com/docs/component-lifecycle#componentwillload), which supports promises.

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,427 @@
...
+  private enableInfiniteScroll = false;
+  /**
+   * A list of non-default fields to include in the query results, separated by commas.
+   * The default fields sent in a request are: 'date', 'author', 'source', 'language', 'filetype', 'parents', ‘urihash’, ‘objecttype’, ‘collection’, ‘permanentid’ 'ec_price', 'ec_name', 'ec_description', 'ec_brand', 'ec_category', 'ec_item_group_id', 'ec_shortdesc', 'ec_thumbnails', 'ec_images', 'ec_promo_price', 'ec_in_stock', 'ec_cogs', and 'ec_rating'.
+   */
+  @Prop({reflect: true}) public fieldsToInclude = '';
+  /**
+   * The desired layout to use when displaying results. Layouts affect how many results to display per row and how visually distinct they are from each other.
+   */
+  @Prop({reflect: true}) display: ResultDisplayLayout = 'list';
```
btaillon, line 106: Gonna have to check with Johnny, but I imagine table and grid won't be supported, so you'll probably be able to remove this prop.

packages/atomic/src/components/result-template-components/atomic-result-sections/atomic-result-section-children.tsx
```diff
@@ -0,0 +1,17 @@
...
+import {hideEmptySection} from '../../../utils/result-section-utils';
+
+/**
+  TODO:
+ */
+@Component({
+  tag: 'atomic-result-section-children',
+  shadow: false,
+})
+export class AtomicResultChildrenEmphasized {
```
btaillon, line 11: Looks like the class name is wrong here

packages/atomic/src/components/atomic-child-result/atomic-child-result.tsx
```diff
@@ -0,0 +1,29 @@
+import {Result} from '@coveo/headless';
+import {Component, Listen, Prop, h} from '@stencil/core';
+import {ResultContextEvent} from '../result-template-components/result-template-decorators';
+@Component({
+  tag: 'atomic-child-result',
+  shadow: true,
+})
+export class AtomicChildResult {
```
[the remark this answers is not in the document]
liviarett: very good point, I think so!

packages/atomic/src/components/result-template-components/atomic-result-sections/atomic-result-section-children.tsx
liviarett: oops

packages/atomic/src/components/result-template-components/result-template-decorators.tsx
```diff
@@ -85,3 +85,79 @@ export const resultContext = (element: Element) =>
...
+ * A [StencilJS property decorator](https://stenciljs.com/) to be used for folded result template components.
+ * This allows the Stencil component to fetch the current folded result from its rendered parent, the `atomic-result` component.
+ *
+ *
+ * @example
+ * @FoldedResultContext() private foldedResult!: FoldedResult;
+ *
+ * For more information and examples, view the "Utilities" section of the readme.
+ */
+export function FoldedResultContext() {
```
olamothe, line 99: Similar to my above comment, it might be possible to modify `ResultContext` decorator to always resolve a `FoldedResult`, with an empty `children` array when there is no folding active.

So that we would not need two different decorator.

I'm not 100% certain it's possible, just a suggestion

packages/atomic/src/components/atomic-result-template/atomic-result-template.tsx
```diff
@@ -118,37 +136,25 @@ export class AtomicResultTemplate {
...
-    // TODO: reorganise
-    const childTemplatesBySection: HTMLAtomicResultChildrenTemplateElement[][] =
-      [];
-    const content = this.getTemplateElement().content;
-    const childrenTemplates = content.querySelectorAll(
-      'atomic-result-children-template'
+    return this.getTemplateElement().content;
+  }
+
+  public async initialize() {
```
btaillon, line 142: in `initialization-utils.ts`, you will need to edit the function that calls `initialize` to await it.

btaillon (commented):
looks good!

packages/atomic/src/utils/utils.ts
```diff
@@ -84,3 +84,14 @@ export function parseAssetURL(url: string, assetPath = './assets') {
   }
   return null;
 }
+
+// TODO: add tests
+export function elementHasAncestorTag(
```
ThibodeauJF, line 89: Could the following method be used instead https://developer.mozilla.org/en-US/docs/Web/API/Element/closest ?

liviarett: I tried that but it didn't work, can't remember why now! will have another look

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  private listWrapperRef?: HTMLDivElement;
+
+  private get fields() {
+    if (this.fieldsToInclude.trim() === '') return [];
+    return this.fieldsToInclude.split(',').map((field) => field.trim());
+  }
+
+  private get defaultFieldsToInclude() {
```
olamothe, line 137: This is no directly related to this PR, but this list should probably be defined and exported from Headless (and then used as source of truth), instead of duplicating it in many place in the codebase.

Anyway, it's fine to leave like this for now !

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,316 @@
...
+    this.resultsPerPage = buildResultsPerPage(this.bindings.engine);
+    this.registerDefaultResultTemplates();
+    this.registerChildrenResultTemplates();
+  }
+
+  private initFolding(options = {}) {
+    const opts: FoldedResultListProps = {};
+    opts.options = {...options};
+    opts.options.folding = {};
+    if (this.foldingCollection)
```
olamothe, line 177: This should enforced with the linter, but there should be curly braces for all if statements.

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,316 @@
...
+  @State()
+  private resultsPerPageState!: ResultsPerPageState;
+
+  @State() public error!: Error;
+  @State() private templateHasError = false;
+
+  /**
+   * TODO:
+   */
+  @Prop({reflect: true}) public foldingCollection?: string;
```
olamothe, line 80: Since the wording in Headless is `collectionField`, `parentField`, `childField`, we should respect the same nomenclature in Atomic.

That way, if someone reads or search our documentation, they will always find the same name for everything, versus subtle variations between each projects.

ThibodeauJF (approved):
Great start!

PR #1867, opened 2022-03-25
feat(atomic): added a way to style the contents of the result list

btaillon:
https://coveord.atlassian.net/browse/KIT-1513

### DIFF
--- a/packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
+++ b/packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
@@ -36,6 +36,7 @@ import {TemplateContent} from '../atomic-result-template/atomic-result-template'
 import {LinkWithResultAnalytics} from '../result-link/result-link';
 import {updateBreakpoints} from './replace-breakpoint';
 import {once} from '../../utils/utils';
+import {StyleFromTemplate} from '../../utils/style-utils';
 
 /**
  * The `atomic-result-list` component is responsible for displaying query results by applying one or more result templates.
@@ -103,6 +104,7 @@ export class AtomicResultList implements InitializableComponent {
   @Prop({reflect: true}) image: ResultDisplayImageSize = 'icon';
 
   private renderingFunction?: (result: Result) => HTMLElement = undefined;
+  @State() private style?: string;
 
   /**
    * Sets a rendering function to bypass the standard HTML template mechanism for rendering results.
@@ -116,6 +118,13 @@ export class AtomicResultList implements InitializableComponent {
     render: (result: Result) => HTMLElement
   ) {
     this.renderingFunction = render;
+  }
+
+  /**
+   * Sets the contents of a style tag to use inside the result list.
+   */
+  @Method() public setStyle(style: string) {
+    this.style = style;
   }
 
   private listWrapperRef?: HTMLDivElement;
@@ -389,6 +398,12 @@ export class AtomicResultList implements InitializableComponent {
 
     return (
       <Host>
+        {
+          <StyleFromTemplate
+            host={this.host}
+            fallback={this.style}
+          ></StyleFromTemplate>
+        }
         {this.templateHasError && <slot></slot>}
         {this.buildResultWrapper()}
       </Host>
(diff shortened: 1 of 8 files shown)


packages/samples/vuejs/src/components/ResultsList.vue
```diff
@@ -1,5 +1,15 @@
...
 import resultTemplate from '../templates/result-template.html';
+
+const resultList = ref<HTMLAtomicResultListElement | null>(null);
+
+onMounted(async () => {
+  await nextTick();
+  resultList.value!.setStyle(`
+    /* Some style here */
+  `);
+});
```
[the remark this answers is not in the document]
liviarett: will that add flicker to the page? should it have a loading or hidden state until that's done?

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -103,6 +104,7 @@ export class AtomicResultList implements InitializableComponent {
   @Prop({reflect: true}) image: ResultDisplayImageSize = 'icon';
 
   private renderingFunction?: (result: Result) => HTMLElement = undefined;
+  @State() private style?: string;
```
ThibodeauJF, line 107: My main concern is this is going to be a slippery slope for all components. Is there a less drastic way of accomplishing the dev's request?

With this method, we cannot ensure that we won't break the result list styling for customers in the future. It is perhaps our component with the most styling as well.

btaillon: > My main concern is this is going to be a slippery slope for all components.

I don't think it's necessarily a slippery slope for all components, since the result list and facets are some of the only components that are complex enough that they may warrant this, and I don't think facets would ever need this since, as far as I'm aware, all their styling can be done with parts.

> Is there a less drastic way of accomplishing the dev's request?

The only ones I can think of are:
* Telling them to make their own result list.
* Adding parts for every even and every uneven row in the table, solving this one use case. I'm not against it, but the compromise is that other structure-based stylings may be expected to be supported the same way in the future.

> With this method, we cannot ensure that we won't break the result list styling for customers in the future. It is perhaps our component with the most styling as well.

That's right. I think we should document that parts are officially supported, and that this is a fallback, but we don't guarantee all changes will be non-breaking for this particular area.

btaillon:
@jpmarceau @npushkarskii Added you guys as reviewers, not because we have doc or strings in this PR, but in case you have suggestions concerning the feature or the comments.

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
ThibodeauJF: Adding parts is a good solution for the dev but I realize it seems a bit silly here, but a stripped table is a common style.

> I don't think it's necessarily a slippery slope for all components, since the result list and facets are some of the only components that are complex enough that they may warrant this, and I don't think facets would ever need this since, as far as I'm aware, all their styling can be done with parts.

I could want my facets values to be stripped with `:nth-child(even)` or something similar. These will probably show up eventually and we can't open a back door too quickly. Inserting a style tag inside the shadow DOM is already something users can "hack" anyways, we are just making the hack official. 

Users will use it very often even when not necessary I'm afraid.

olamothe: Lets say we want to fix the current problem precisely, could there be a css variable for `table-result-list-even-color` and `table-result-list-odd-color` ? And by default it's ... "inherit" for both ?

If we think about it differently, we can try to list what are the frequent things you might need when styling a table : 
* header background + text color
* rows and columns color
* margins for rows and columns
* footer background + text color
* Being able to sort by columns
* ??

Then we make these things easy to do (with css variable/parts/props when needed). And if you need something more than that, then yeah, we say "sorry, you will have to roll your own component).

Because it's true that by adding the `style` prop, we make the hack "official".

btaillon: In this case, I'll add support for alternating rows instead.

PR #1874, opened 2022-03-29
feat(headless): export SearchDefaultFieldsToInclude

liviarett:
https://coveord.atlassian.net/browse/KIT-1518

### DIFF
--- a/packages/headless/src/api/search/search/search-request.ts
+++ b/packages/headless/src/api/search/search/search-request.ts
@@ -59,3 +59,26 @@ export type SearchRequest = BaseParam &
   AnalyticsParam &
   ExcerptLength &
   ActionsHistoryParam;
+
+// Change this list when changing the fields exposed by `FieldsToIncludeParam`
+export const SearchDefaultFieldsToInclude: string[] = [
+  'date',
+  'author',
+  'source',
+  'language',
+  'filetype',
+  'parents',
+  'ec_price',
+  'ec_name',
+  'ec_description',
+  'ec_brand',
+  'ec_category',
+  'ec_item_group_id',
+  'ec_shortdesc',
+  'ec_thumbnails',
+  'ec_images',
+  'ec_promo_price',
+  'ec_in_stock',
+  'ec_cogs',
+  'ec_rating',
+];
--- a/packages/headless/src/api/search/search/search-request.ts
+++ b/packages/headless/src/api/search/search/search-request.ts
@@ -59,26 +59,3 @@ export type SearchRequest = BaseParam &
   AnalyticsParam &
   ExcerptLength &
   ActionsHistoryParam;
-
-// Change this list when changing the fields exposed by `FieldsToIncludeParam`
-export const SearchDefaultFieldsToInclude: string[] = [
-  'date',
-  'author',
-  'source',
-  'language',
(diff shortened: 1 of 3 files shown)


packages/headless/src/api/search/search/search-request.ts
```diff
@@ -59,3 +59,26 @@ export type SearchRequest = BaseParam &
   AnalyticsParam &
   ExcerptLength &
   ActionsHistoryParam;
+
+// Change this list when changing the fields exposed by `FieldsToIncludeParam`
+export const SearchDefaultFieldsToInclude: string[] = [
```
olamothe, line 64: It might be a bit more complicated than that.

There's a partial overlap in `fields-state.ts`, where we already have a list of the bare minimum fields that are needed in order for various analytics events to work properly. These should be considered "mandatory in all scenarios". 

Then you have a couple of them here which are really useful to power any result list (date, parents), but not strictly mandatory for all use cases.

Then all those `ec_${something}` fields are commerce specific, and we would only need them when dealing with commerce pages. So they should not have been put in `atomic-result-list` in the first place, in my opinion.

So... all in all, just to say I'm not 100% certain how we should subdivide that list to ensure that there's just the strict minimum requested depending on the use case (ie: which type of page the end user is trying to build).

If we request too much all the time, it will work correctly, but: 

* In the best case, the worst that will happen is that it pollute the request with useless fields that are absent from the index anyway.

* In the worst case, this will have negative performance impact.

So yeah, I'm not sure how we should arrange this... maybe the simplest would be various exported list with explicit naming that describe when they should be used ? 🤔

liviarett: I think it makes sense to have a 'bare minimum' and then a couple of others like 'defaults for ecommerce', etc. who would be best to help me come up with these different lists?

packages/headless/src/features/fields/fields-state.ts
```diff
@@ -19,16 +19,51 @@ export interface FieldsState {
   fieldsDescription: FieldDescription[];
 }
 
+/**
+ * A list of the minimum fields required in a query.
+ */
+export const MinimumFieldsToInclude: string[] = [
```
btaillon, line 25:
```suggestion
export const MinimumFieldsToInclude = [
```

packages/headless/src/features/fields/fields-state.ts
```diff
@@ -19,16 +19,51 @@ export interface FieldsState {
...
+  ...MinimumFieldsToInclude,
+  'date',
+  'filetype',
+  'parents',
+];
+
+/**
+ * A list of the recommended fields to be used in a query for an e-commerce page.
+ */
+export const EcommerceDefaultFieldsToInclude: string[] = [
```
btaillon, line 48:
```suggestion
export const EcommerceDefaultFieldsToInclude = [
```

olamothe (approved):
I think the list is good !

For the commerce fields, to double check, I've pinged the guys working in commerce to verify if everything is there.

However, just exporting from this file won't make it globally available. We'll need to re-export the lists from index.ts

olamothe:
@liviarett After discussing with commerce team, we can remove `e_cogs`, but the rest is good !

PR #1876, opened 2022-03-29
feat(headless,atomic): added smart snippet source analytics

btaillon:
https://coveord.atlassian.net/browse/KIT-1498

I decided to make the `SmartSnippet` component responsible for logging click analytics, similarly to the InteractiveResult controller.

It can be called like so:
```ts
const smartSnippet = buildSmartSnippet(...);
...
smartSnippet.source.select();
```

React samples here: https://github.com/coveo/ui-kit/pull/1880

PR #1880, opened 2022-03-29
docs(headless): added react sample for smart snippet source analytics

btaillon:
https://coveord.atlassian.net/browse/KIT-1498

Split from: https://github.com/coveo/ui-kit/pull/1876


packages/samples/headless-react/src/components/smart-snippet/smart-snippet.class.tsx
```diff
@@ -90,6 +91,23 @@ export class SmartSnippet extends Component<{}, SmartSnippetState> {
...
+                onMouseUp={() => this.controller.source?.select()}
+                onTouchStart={() =>
+                  this.controller.source?.beginDelayedSelect()
+                }
+                onTouchEnd={() => this.controller.source?.cancelPendingSelect()}
+              >
+                Source
+              </a>
+            ) : (
+              []
```
liviarett, line 109: how come you return an empty array here?
isn't it cleaned to do 
`Boolean(source) && <Component />` ?

btaillon: in a few instances, I saw `false` being rendered. So now, I'm avoiding using the `&&` pattern in JSX.

liviarett: odd! I've seen 0 before, but never false.
fair enough

btaillon: I may have seen it in attributes

olamothe: I personally prefer extracting any logic from the JSX code, and just have a `renderSource()` function which itself contains the `if/else` code.

Just seems easier to read ?

olamothe (approved):

packages/samples/headless-react/src/components/smart-snippet/smart-snippet.class.tsx
btaillon: In Atomic that's generally what I do, I didn't do it here for the sake of having the whole thing in the same function, but I'll extract it.

ThibodeauJF (approved):

PR #1881, opened 2022-03-29
fix(atomic): query suggestion & recent queries, wrap long query up to 2 lines

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1321

Without highlight
Screen Shot 2022-03-29 at 2 56 55 PM

With highlight
Screen Shot 2022-03-29 at 2 57 07 PM

### DIFF
--- a/packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
+++ b/packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
@@ -141,12 +141,13 @@ export class AtomicSearchBoxRecentQueries {
       key: `recent-${value}`,
       query: value,
       content: (
-        <div class="flex items-center">
-          <atomic-icon icon={Clock} class="w-5 h-5 mr-2 -ml-1"></atomic-icon>
+        <div class="flex items-center break-all">
+          <atomic-icon icon={Clock} class="w-4 h-4 mr-2 shrink-0"></atomic-icon>
           {query === '' ? (
-            <span>{value}</span>
+            <span class="break-all line-clamp-2">{value}</span>
           ) : (
             <span
+              class="break-all line-clamp-2"
               // deepcode ignore DOMXSS: Value escaped in upstream code, deepcode ignore ReactSetInnerHtml: This is not React code.
               innerHTML={HighlightUtils.highlightString({
                 content: value,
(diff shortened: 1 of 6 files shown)


packages/atomic/tailwind.config.js
```diff
@@ -84,5 +83,6 @@ module.exports = {
         variants('outlineColor')
       );
     }),
+    require('@tailwindcss/line-clamp'),
```
btaillon, line 86: we could replace the line-clamp mixin in polyfills.pcss with this

btaillon (approved):

liviarett (approved):
that's some sorcery right there 👏🏻

packages/atomic/src/images/clock.svg
```diff
@@ -1 +1 @@
-<svg fill="none" stroke="currentColor" stroke-linejoin="round" stroke-linecap="round"  viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" r="5.5"/><path d="m8.5 4.5v4"/><path d="m10.3066 10.1387-1.80932-1.5768"/></svg>
\ No newline at end of file
+<svg fill="none" stroke="currentColor" stroke-linejoin="round" stroke-linecap="round"  viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" r="7.5"/><path d="m8.5 4.5v4"/><path d="m10.3066 10.1387-1.80932-1.5768"/></svg>
```
olamothe, line 1: What was wrong with the SVG ?

packages/atomic/tailwind.config.js
olamothe: We definitely should replace it, and not have two different manner to support line clamp ;)

packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -141,12 +141,13 @@ export class AtomicSearchBoxRecentQueries {
       key: `recent-${value}`,
       query: value,
       content: (
-        <div class="flex items-center">
-          <atomic-icon icon={Clock} class="w-5 h-5 mr-2 -ml-1"></atomic-icon>
+        <div class="flex items-center break-all">
```
olamothe, line 144: I was gonna comment on using `break-word` instead, but yeah, google uses break-all also. Which is a bit weird... 

But it's true that if we have one suggestions that's just one massively long word, we'd have problems.

So 👍

olamothe (approved):

packages/atomic/tailwind.config.js
ThibodeauJF: The line clamp mixin is used where there's no Tailwind @btaillon right? The result list

packages/atomic/src/images/clock.svg
ThibodeauJF: Too small vs the parent atomic-icon. Changed `r="7.5"` which allowed me to remove the `-ml-1` and use the same size as the other icon (magnifier).

PR #1883, opened 2022-03-30
refactor(atomic): extract reusable code for results lists and templates

liviarett:
I hope it's not too hard to understand what I've done, but if anyone wants a walkthrough, let me know!


packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  const hasReachedEndOfElement =
+    window.innerHeight + window.scrollY >= host.offsetHeight;
+
+  if (hasReachedEndOfElement) {
+    resultList.fetchMoreResults();
+  }
+}
+
+export async function registerResultTemplates(
```
olamothe, line 135: I'm not a huge fan of these functions (registerResultTemplates, initializeResultList, postRenderCleanup), which receives a reference to the component and then "augment it" in a non explicit way.

Said differently, I think we will have to jump around quite a bit in the code, across different files, to follow what's going on.

I think these could be refactored in a single class (lets say `ResultListCommon`). That class itself would contain the properties it needs to do its job (ie: the headless controller).

Then, in `atomic-result-list` we could do `this.resultListCommon = new ResultListCommon({...whatever is needed...})` when initializing.

Afterwards, let's say that `atomic-result-list` needs access to the resultsPerPage headless controller, it could do `this.resultListCommon.resultsPerPage`.

I think with that pattern we'd be able to look at `ResultListCommon` as one entity which contains everything it needs to do its job properly.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,204 @@
...
+    | 'resultListState'
+    | 'bindings'
+    | 'imageSize'
+    | 'image'
+    | 'getContentOfResultTemplate'
+  > {
+  classes: string;
+}
+
+export function getClasses({
```
olamothe, line 64: It seems `getClasses` function is only used locally in `result-list.tsx`. Why is it exported here and not localized close to where it is used ?

packages/atomic/src/components/result-template/result-template-common.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  return {
+    conditions: this.conditions.concat(this.matchConditions),
+    content: getTemplateElement(this.host).content!,
+    priority: 1,
+  };
+}
+
+export function renderIfError(this: TemplateComponent) {
```
olamothe, line 85: Similar to my other comment about extracting classes, we could try to subdivide the problems into "what are the functionalities that need to be shared between both result lists".

ie: 
* Things that need to hook into the rendering lifecycle (renderIfError, postRenderCleanup)
* Things that need to instantiate headless controllers (initializeResultList, initializeResultsPerPage)
* Things that need to manage templates (getTemplate, addMatchConditions, validateTemplate)
* Things that need to handle user events (handleInfiniteScroll)

What if for each of those generic concerns, there was a class that handle only that ? And atomic-result-list and atomic-folded-result-list would instantiate these different classes at initialization.

Do you think that's a pattern that might work ?

packages/atomic/src/components.d.ts
```diff
@@ -540,7 +538,7 @@ export namespace Components {
         /**
           * Gets the appropriate result template based on conditions applied.
          */
-        "getTemplate": () => Promise<ResultTemplate<TemplateContent> | null>;
+        "getTemplate": () => Promise<import("/Users/liviarett/coveo/ui-kit/packages/headless/dist/definitions/index").ResultTemplate<DocumentFragment> | null>;
```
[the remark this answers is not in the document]
liviarett: https://media.giphy.com/media/xThuW7icYBlQZxjnyg/giphy.gif

packages/atomic/src/components/result-lists/atomic-result-list/atomic-result-list.tsx
```diff
@@ -0,0 +1,152 @@
...
+  InitializableComponent,
+  InitializeBindings,
+} from '../../../utils/initialization-utils';
+import {
+  ResultDisplayLayout,
+  ResultDisplayDensity,
+  ResultDisplayImageSize,
+} from '../../atomic-result/atomic-result-display-options';
+import {TemplateContent} from '../../result-template/atomic-result-template/atomic-result-template';
+import {ResultList as ResultListComponent} from '../result-list';
```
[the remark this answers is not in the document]
liviarett: I'm calling it `BaseResultList` as it's more in line with what we do elsewhere!

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: ok, that makes sense! I had a feeling that this wasn't great, but I'm not really used to working with classes, so didn't have better ideas! I'll try to implement your suggestion

packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,169 @@
...
+   */
+  @Prop({reflect: true}) density: ResultDisplayDensity = 'normal';
+  /**
+   * The expected size of the image displayed in the results.
+   */
+  @Prop({reflect: true}) imageSize?: ResultDisplayImageSize;
+  /**
+   * @deprecated use `imageSize` instead.
+   */
+  @Prop({reflect: true}) image: ResultDisplayImageSize = 'icon';
```
ThibodeauJF, line 97: I wouldn't add this one

packages/atomic/src/components/result-lists/atomic-result-list/atomic-result-list.tsx
```diff
@@ -0,0 +1,152 @@
...
+  public resultsPerPageState!: ResultsPerPageState;
+
+  @State() public ready = false;
+
+  @State() public error!: Error;
+  @State() public templateHasError = false;
+
+  /**
+   * A list of non-default fields to include in the query results, separated by commas.
+   * The default fields sent in a request are: 'date', 'author', 'source', 'language', 'filetype', 'parents', ‘urihash’, ‘objecttype’, ‘collection’, ‘permanentid’ 'ec_price', 'ec_name', 'ec_description', 'ec_brand', 'ec_category', 'ec_item_group_id', 'ec_shortdesc', 'ec_thumbnails', 'ec_images', 'ec_promo_price', 'ec_in_stock', 'ec_cogs', and 'ec_rating'.
```
ThibodeauJF, line 89: Changing the default fields in Headless won't be reflected here and we probably won't remember to keep both in sync 😄

packages/atomic/src/components/result-lists/atomic-result-list/atomic-result-list.tsx
```diff
@@ -0,0 +1,152 @@
...
+    this.renderingFunction = render;
+  }
+
+  @Listen('scroll', {target: 'window'})
+  handleInfiniteScroll() {
+    handleInfiniteScroll(this.enableInfiniteScroll, this.host, this.resultList);
+  }
+
+  public async initialize() {
+    if (this.host.innerHTML.includes('<atomic-result-children')) {
```
ThibodeauJF, line 129:
```suggestion
    if (this.host.innerHTML.includes('<atomic-result-children>')) {
```
or without brackets

Or you could use `querySelector()` ?

ThibodeauJF (commented):
Moving files + extracting logic makes the PR a bit too large, some code was already "reviewed" but we can't be sure what is was. Maybe we can extract what we need without moving the files, and move the files in another PR later if it makes more sense.

packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * The `atomic-folded-result-list` component is responsible for displaying query results by applying one or more result templates.
+ *
+ * @internal
+ * @part result-list - The element containing every result of a result list
+ */
+@Component({
+  tag: 'atomic-folded-result-list',
+  styleUrl: 'atomic-folded-result-list.pcss',
```
btaillon, line 48: I would extract all the contents of `atomic-result-list.pcss` into a `atomic-common-result-list.pcss` file or something similar, and I would make both `atomic-folded-result-list.pcss` and `atomic-result-list.pcss` extend it.

This way, when we make changed to `atomic-common-result-list.pcss`, we know we're affecting more than one component.

packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,169 @@
...
+    this.renderingFunction = render;
+  }
+
+  @Listen('scroll', {target: 'window'})
+  handleInfiniteScroll() {
+    handleInfiniteScroll(this.enableInfiniteScroll, this.host, this.resultList);
+  }
+
+  public async initialize() {
+    await initializeResultList.call(this, this.initFolding.bind(this));
```
btaillon, line 131: I would avoid the use of `bind` in general and instead use an arrow function, like so:
```ts
    await initializeResultList.call(this, () => this.initFolding());
```

Additionally, as Oli already pointed out, I would also avoid `call` and instead extend a class.

packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  private initFolding(_: SearchEngine<{}>, options: FoldedResultListProps) {
```
btaillon, line 134: Would it be possible here to take result list options as a parameter instead of folded result list options?

packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -0,0 +1,169 @@
...
+      opts.options.folding.collectionField = this.collectionField;
+    }
+    if (this.parentField) {
+      opts.options.folding.parentField = this.parentField;
+    }
+    if (this.childField) {
+      opts.options.folding.childField = this.childField;
+    }
+
+    return buildFoldedResultList(this.bindings.engine, opts);
```
btaillon, line 151: Instead of all the above logic for options, I would prefer a bit of a one-liner like this:
```ts
    return buildFoldedResultList(
      this.bindings.engine,
      {
        ...options,
        collectionField: this.collectionField ?? 'foldingcollection',
        parentField: this.parentField ?? 'foldingparent',
        childField: this.childField ?? 'foldingchild',
      }
    );
```

packages/atomic/src/components/result-lists/result-list.tsx
```diff
@@ -0,0 +1,82 @@
...
+};
+
+const Results: FunctionalComponent<ResultsProps> = (props) => {
+  if (!props.resultListState.results.length) {
+    return null;
+  }
+  if (props.display === 'table') {
+    return <TableDisplayResults {...props} />;
+  }
+  return <ListDisplayResults {...props} />;
```
btaillon, line 81: I like this, I think you should go even further and:
* Divide `ListDisplayResults` into `ListDisplayResults` and `GridDisplayResults`, and use a switch case here for those 3 cases
* Divide the placeholder into three corresponding components: `ListDisplayResultsPlaceholder`, `GridDisplayResultsPlaceholder` and `TableDisplayResultsPlaceholder`.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+export function getId(
+  result: Result | FoldedCollection,
+  resultListState: FoldedResultListState | ResultListState
+) {
+  return unfolded(result).uniqueId + resultListState.searchResponseId;
+}
+
+export function unfolded(result: Result | FoldedResult): Result {
+  return (result as FoldedResult).result || result;
```
btaillon, line 87: I prefer `??` over `||`:
```suggestion
  return (result as FoldedResult).result ?? result;
```

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,204 @@
...
+  resultsPerPageState: ResultsPerPageState;
+
+  resultList: FoldedResultList | ResultList;
+  resultListState: FoldedResultListState | ResultListState;
+  getContentOfResultTemplate(
+    result: Result | FoldedResult
+  ): HTMLElement | DocumentFragment;
+}
+
+export interface ResultsProps
```
btaillon, line 49: If possible, I would like to avoid a pattern that requires extracting a set of properties like this. I would prefer to extend a base interface.

packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
liviarett: thanks! didn't notice it was deprecated

packages/atomic/src/components/result-lists/atomic-result-list/atomic-result-list.tsx
liviarett: do we reference headless in atomic? could I say check `EcommerceDefaultFieldsToInclude` in headless for a list of the fields included by default?

packages/atomic/src/components/result-lists/atomic-result-list/atomic-result-list.tsx
ThibodeauJF: this.host.shadowRoot.querySelector should work

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -107,11 +88,17 @@ export class AtomicFoldedResultList implements InitializableComponent {
...
-  private renderingFunction?: (result: FoldedResult) => HTMLElement = undefined;
+  @Prop({reflect: true}) public collectionField?: string;
+  /**
+   * TODO:
+   */
+  @Prop({reflect: true}) public parentField?: string;
+  /**
+   * TODO:
+   */
+  @Prop({reflect: true}) public childField?: string;
```
ThibodeauJF, line 101: Is `display` omitted on purpose? I think grid displays could still benefit from folded results?

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+  templateHasError: boolean;
+  listWrapperRef?: HTMLDivElement;
+  resultsPerPageState: ResultsPerPageState;
+
+  resultListCommon?: ResultListCommon;
+
+  resultList: FoldedResultList | ResultList;
+}
+
+export function getId(
```
ThibodeauJF, line 55: All the methods/variables that exported here should have a more specific name e.g. getAtomicResultListId.
If it's too generic, using the IDE you'll get completions that can be a bit unclear,

packages/atomic/src/components/atomic-result-placeholder/atomic-result-placeholder.tsx
```diff
@@ -20,7 +20,7 @@ const placeholderClasses = 'block bg-neutral w-full h-full rounded';
 export class AtomicResultPlaceholder {
   @Prop() display!: ResultDisplayLayout;
   @Prop() density!: ResultDisplayDensity;
-  @Prop() imageSize!: ResultDisplayImageSize;
+  @Prop() imageSize?: ResultDisplayImageSize;
```
ThibodeauJF, line 23: Why make it optional now? or maybe why aren't they all optional as they have no default assigned values

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -59,38 +56,43 @@ import {updateBreakpoints} from '../../utils/replace-breakpoint';
 })
 export class AtomicResultList implements InitializableComponent {
```
ThibodeauJF, line 57: Maybe both result list classes could implement a common interface
```suggestion
export class AtomicResultList implements InitializableComponent, AtomicBaseResultList {
```

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -59,38 +56,43 @@ import {updateBreakpoints} from '../../utils/replace-breakpoint';
...
-  /**
-   * TODO: KIT-452 Infinite scroll feature
-   * Whether to automatically retrieve an additional page of results and append it to the
-   * current results when the user scrolls down to the bottom of element
-   */
-  private enableInfiniteScroll = false;
   /**
    * A list of non-default fields to include in the query results, separated by commas.
-   * The default fields sent in a request are: 'date', 'author', 'source', 'language', 'filetype', 'parents', ‘urihash’, ‘objecttype’, ‘collection’, ‘permanentid’ 'ec_price', 'ec_name', 'ec_description', 'ec_brand', 'ec_category', 'ec_item_group_id', 'ec_shortdesc', 'ec_thumbnails', 'ec_images', 'ec_promo_price', 'ec_in_stock', 'ec_cogs', and 'ec_rating'.
+   * TODO: add a note about headless?.
```
ThibodeauJF, line 89: This todo would appear on the website reference doc

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -117,296 +117,53 @@ export class AtomicResultList implements InitializableComponent {
   @Method() public async setRenderFunction(
     render: (result: Result) => HTMLElement
```
ThibodeauJF, line 118:
```suggestion
    render: RenderingFunc
```

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -117,296 +117,53 @@ export class AtomicResultList implements InitializableComponent {
...
   ) {
-    this.renderingFunction = render;
-  }
-
-  private listWrapperRef?: HTMLDivElement;
-
-  private get fields() {
-    if (this.fieldsToInclude.trim() === '') return [];
-    return this.fieldsToInclude.split(',').map((field) => field.trim());
+    this.resultListCommon.renderingFunction = render as RenderingFunc;
```
ThibodeauJF, line 120:
```suggestion
    this.resultListCommon.renderingFunction = render;
```

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
btaillon: I recall suggesting that we remove `display`. I think whether we want to include folded results in grid displays or not is probably gonna depend on Johnny's design, so I think adding support for it would be better in a separate PR.

packages/atomic/src/components/result-lists/atomic-result-list/atomic-result-list.tsx
liviarett: the problem is that it's nested within atomic-result's shadow dom too. so I could do `this.host?.shadowRoot?.querySelector('atomic-result').shadowRoot.querySelector('atomic-result-children')` but it doesn't seem to have rendered anything at that point, so `shadowRoot?.querySelector('atomic-result')` is actually null

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
liviarett: I wasn't planning on leaving it, I just wanted to know what I should do there instead!

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
liviarett: sooo my idea here was to make it clear what kind of result you are receiving (if you are in `AtomicResultList`, you will only get `Result`, so you need to handle that in your rendering function. If you are in the folded one, you only get `FoldedResult` and therefore can render children, etc

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export function getFields(fieldsToInclude: string): string[] {
+  if (fieldsToInclude.trim() === '')
+    return [...EcommerceDefaultFieldsToInclude];
+  return EcommerceDefaultFieldsToInclude.concat(
+    fieldsToInclude.split(',').map((field) => field.trim())
+  );
+}
+
+export function handleInfiniteScroll(
```
btaillon, line 74: If possible, I would move this into the common result list class.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+  onReady(): void;
+  onError(): void;
+}
+
+export class ResultListCommon {
+  private bindings: Bindings;
+  private render?: RenderingFunc;
+
+  public resultTemplatesManager!: ResultTemplatesManager<TemplateContent>;
+  public listOpts?: ResultListProps;
```
btaillon, line 119: I would probably rename this something like `resultListControllerProps`

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+    resultListState: FoldedResultListState | ResultListState,
+    listWrapperRef?: HTMLDivElement
+  ) {
+    if (resultListState.firstSearchExecuted) {
+      listWrapperRef?.classList.remove('placeholder');
+    }
+  }
+}
+
+export type RenderingFunc = (result: Result | FoldedResult) => HTMLElement;
```
btaillon, line 194: I would rename this `RenderingFunction` or `ResultRenderingFunction`

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const hasReachedEndOfElement =
+    window.innerHeight + window.scrollY >= host.offsetHeight;
+
+  if (hasReachedEndOfElement) {
+    resultList.fetchMoreResults();
+  }
+}
+
+function makeDefaultTemplate(): ResultTemplate<DocumentFragment> {
```
btaillon, line 91: I would move this inside the class below

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+  resultListState: FoldedResultListState | ResultListState
+) {
+  return unfolded(result).uniqueId + resultListState.searchResponseId;
+}
+
+export function unfolded(result: Result | FoldedResult): Result {
+  return (result as FoldedResult).result ?? result;
+}
+
+export function getFields(fieldsToInclude: string): string[] {
```
btaillon, line 66: I would rename this to something like `determineFieldsToInclude` and rename the parameter to something like `configuredFieldsToInclude` or something similar

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export function getId(
+  result: Result | FoldedCollection,
+  resultListState: FoldedResultListState | ResultListState
+) {
+  return unfolded(result).uniqueId + resultListState.searchResponseId;
+}
+
+export function unfolded(result: Result | FoldedResult): Result {
```
btaillon, line 62: If this is exported, I would rename it something like `getUnfoldedResult`

packages/atomic/src/components/result-lists/result-list-common.tsx
btaillon: I would also move as many of these as possible inside the class below, even if they have no parameters.

packages/atomic/src/components/result-lists/result-list.tsx
```diff
@@ -0,0 +1,120 @@
...
+import {TableDisplayResults} from './table-display-results';
+import {ListDisplayResults} from './list-display-results';
+import {GridDisplayResults} from './grid-display-results';
+import {GridDisplayResultsPlaceholder} from './grid-display-results-placeholder';
+import {ListDisplayResultsPlaceholder} from './list-display-results-placeholder';
+interface ResultListProps {
+  parent: AtomicResultListBase;
+}
+
+export const BaseResultList: FunctionalComponent<ResultListProps> = (props) => {
```
btaillon, line 23: I would probably rename this file to match the class: `base-result-list.tsx`

packages/atomic/src/components/result-lists/result-list.tsx
```diff
@@ -0,0 +1,120 @@
...
+  ResultPlaceholderProps,
+  ResultsProps,
+} from './result-list-common';
+import {TableDisplayResultsPlaceholder} from './table-display-results-placeholder';
+import {TableDisplayResults} from './table-display-results';
+import {ListDisplayResults} from './list-display-results';
+import {GridDisplayResults} from './grid-display-results';
+import {GridDisplayResultsPlaceholder} from './grid-display-results-placeholder';
+import {ListDisplayResultsPlaceholder} from './list-display-results-placeholder';
+interface ResultListProps {
```
btaillon, line 19: would probably rename this `BaseResultListProps` and export it

packages/atomic/src/components/result-lists/result-list.tsx
```diff
@@ -0,0 +1,120 @@
...
+  parent: AtomicResultListBase;
+}
+
+export const BaseResultList: FunctionalComponent<ResultListProps> = (props) => {
+  const parent = props.parent;
```
btaillon, line 24: you can omit this line by replacing `(props) => {` with `({parent}) => {`

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,194 @@
...
+  bindings: Bindings;
+  host: HTMLElement;
+  getContentOfResultTemplate(
+    result: Result | FoldedResult
+  ): HTMLElement | DocumentFragment;
+
+  classes: string;
+}
+
+export interface AtomicResultListBase extends Omit<ResultsProps, 'classes'> {
```
btaillon, line 45: Instead of having `AtomicResultListBase`, `ResultsProps`, etc. I think `ResultListCommon` should expose all these attributes & functions somehow. I would prefer not to have both a base interface and a base class, essentially.

packages/atomic/src/components/atomic-result/atomic-result-display-options.ts
```diff
@@ -26,7 +26,7 @@ function getDensityClass(density: ResultDisplayDensity) {
   }
 }
 
-function getImageClass(image: ResultDisplayImageSize) {
+function getImageClass(image?: ResultDisplayImageSize) {
```
btaillon, line 29: I would not make `image` optional here

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -107,11 +88,17 @@ export class AtomicFoldedResultList implements InitializableComponent {
    */
   @Prop({reflect: true}) imageSize?: ResultDisplayImageSize;
```
btaillon, line 89: `imageSize` should no longer be optional. It was optional in `atomic-result-list` because `image` was deprecated.

packages/atomic/src/components/result-lists/result-list.tsx
liviarett: why export it? I don't think it's being used anywhere else

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
liviarett: aah ok

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: I've done everything else you guys suggested but I'm not sure how to do this. `AtomicResultListBase` has a few things that need to be in the component itself (and not in the common class), like `bindings`,  `State`s, etc. 
do you have a suggestion of how to achieve this?

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -124,197 +112,66 @@ export class AtomicFoldedResultList implements InitializableComponent {
...
-    const linkEl = document.createElement('atomic-result-link');
-    content.appendChild(linkEl);
-    this.resultTemplatesManager.registerTemplates({
-      content,
-      conditions: [],
+  public async initialize() {
+    this.resultListCommon = new ResultListCommon({
+      bindings: this.bindings,
+      fieldsToInclude: this.fieldsToInclude,
+      templateElements: this.host.querySelectorAll('atomic-result-template'),
```
ThibodeauJF, line 131: The `host` could be passed to the ResultListCommon and the template resolution code would be done there once

packages/atomic/src/components/result-lists/base-result-list.tsx
```diff
@@ -0,0 +1,127 @@
...
+import {once} from '../../utils/utils';
+import {
+  getResultDisplayClasses,
+  ResultDisplayLayout,
+} from '../atomic-result/atomic-result-display-options';
+import {
+  AtomicResultListBase,
+  ResultPlaceholderProps,
+  ResultsProps,
+} from './result-list-common';
```
ThibodeauJF, line 12: Having both a common result list and a base result list was a tad confusing

packages/atomic/src/components/result-lists/result-list-common.tsx
ThibodeauJF: Maybe `ResultListCommon` doesn't need to be a separate component. `AtomicResultListBase` could perhaps have a render method?

packages/atomic/src/components/result-lists/base-result-list.tsx
```diff
@@ -0,0 +1,127 @@
...
+import {ListDisplayResultsPlaceholder} from './list-display-results-placeholder';
+
+interface BaseResultListProps {
+  parent: AtomicResultListBase;
+}
+
+export const BaseResultList: FunctionalComponent<BaseResultListProps> = ({
+  parent,
+}) => {
+  const updateBreakpointsOnce = once(() => updateBreakpoints(parent.host));
```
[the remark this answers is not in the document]
liviarett: yeah, probably!!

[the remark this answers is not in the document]
liviarett: @btaillon can't find your comment about this now, so don't know if this is what you meant, but it does seem to be useful (when defined outside of the component)
Screenshot 2022-04-05 at 10 10 41

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: @ThibodeauJF done, what do you think?

ThibodeauJF: It's much better! I still find some things a bit confusing:

1. `AtomicResultListBase` contains `resultListCommon: ResultListCommon;` meanwhile `ResultListCommon` has the method `public renderList(props: AtomicResultListBase)`. It's a circular reference

2. There is already common props for  `ResultListCommon`, `ResultListCommonProps`, so the `props: AtomicResultListBase` are also common props?

3. `AtomicResultListBase` starts off by extending `ResultsProps` (`AtomicResultListBase extends Omit<ResultsProps, 'classes'>`) which is another common props interface?

Maybe the naming is problematic for me 🤔  I would keep everything with Prop associated to the user facing `Prop()` in Stencil. For the "jsx" props or just common properties/options/params, we can remove the Props and use Options, or something similar

ThibodeauJF (approved):

liviarett:
@ThibodeauJF ok, so if I understand correctly from your comment, I should change `DisplayProps` to `DisplayOptions`, `ResultPlaceholderProps` and `ResultsProps` can stay as they are, since they are props for functional components.
`ResultListCommonProps` can become `ResultListCommonOptions`, since they are used in the constructor.
And I'll also change `constructor(props: ResultListCommonProps)` to `constructor(opts: ResultListCommonOptions)`.

about the `AtomicResultListBase` extending `ResultsProps`, it's because it's `AtomicResultListBase` that passes down the props to the functional components, so I'm just reusing it instead of declaring all of the props again..
and the reason for there being both `AtomicResultListBase` and `ResultListCommonOptions` is that the list base includes the stuff that needs to be in a stencil component, while the common options are things that I managed to extract to the instance of `ResultListCommon`. I wonder if part of the confusion is because they are very close to one another on the file, making you wonder if they are connected? Maybe adding a `Component` to the end will make it clearer?

also, I think that that recursiveness you mentioned makes sense, since `renderList` then needs access to some stuff that comes from the component itself, so the same as doing `this.resultCommonList.renderList.call(this)` from the component's render function... I can't think of a better way!

packages/atomic/src/components/result-lists/base-result-list.tsx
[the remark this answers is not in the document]
btaillon: I did not comment about `updateBreakpoints`, as far as I recall.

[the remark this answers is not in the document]
btaillon: now I remember. yes, as Olivier pointed out, the issue was that it was being called at the same place where it was defined.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,316 @@
...
+      density,
+      (imageSize ?? image)!
+    );
+    if (resultListState.firstSearchExecuted && resultList.state.isLoading) {
+      classes.push('loading');
+    }
+    return classes.join(' ');
+  }
+
+  public renderList(props: AtomicResultListBaseComponent) {
```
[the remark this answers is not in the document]
liviarett: do you mean, instead of doing `this.resultListCommon.renderList(this);`, do `this.resultListCommon.renderList({...anObjectWithEverything);`?
I don't get how we could take a `ResultListCommonProps` if it doesn't have any of the stuff we need for rendering 🤔  I think I need a bit more help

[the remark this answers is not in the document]
liviarett: perhaps I could `Pick<AtomicResultListBaseComponent, '....'>` ?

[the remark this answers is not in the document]
btaillon: I basically mean removing `AtomicResultListBaseComponent` altogether, and calling this function like:
```ts
return this.resultListCommon.renderList({
    templateHasError: this.templateHasError,
    listWrapperRef: this.listWrapperRef,
    /* etc */
});
```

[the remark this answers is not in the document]
liviarett: ah, ok I see. but then, in practice, how do I declare the parameters of `renderList` ? do I then need a new `Props` interface? and then do I copy and paste, like
```ts
interface RenderListProps {
  resultListState: FoldedResultListState | ResultListState;
  resultListCommon: ResultListCommon;
  host: HTMLElement;
  etc...
}
```
even if we have most of that stuff in other interfaces already, like
```ts
export interface ResultsProps extends DisplayOptions {
  resultListState: FoldedResultListState | ResultListState;
  resultListCommon: ResultListCommon;
  bindings: Bindings;
  host: HTMLElement;
  getContentOfResultTemplate(
    result: Result | FoldedResult
  ): HTMLElement | DocumentFragment;

  classes: string;
}

interface DisplayOptions {
  density: ResultDisplayDensity;
  imageSize?: ResultDisplayImageSize;
  image?: ResultDisplayImageSize;
  display?: ResultDisplayLayout;
}
```

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,316 @@
...
+    const hasReachedEndOfElement =
+      window.innerHeight + window.scrollY >= host.offsetHeight;
+
+    if (hasReachedEndOfElement) {
+      resultList.fetchMoreResults();
+    }
+  }
+
+  public componentDidRender(
+    resultListState: FoldedResultListState | ResultListState,
```
btaillon, line 205: Similarly to below, instead of taking the state here, I think we should take `firstSearchExecuted` directly.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,316 @@
...
+    return (
+      this.getUnfoldedResult(result).uniqueId + resultListState.searchResponseId
+    );
+  }
+
+  public getUnfoldedResult(result: Result | FoldedResult): Result {
+    return (result as FoldedResult).result ?? result;
+  }
+
+  public handleInfiniteScroll(
```
btaillon, line 187: I think it would be simpler to understand if this became a getter read by both result lists.
Example:
```ts
public get hasReachedEndOfList() {
    return window.innerHeight + window.scrollY >= host.offsetHeight;
}
```

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,316 @@
...
+      includeDefaultTemplate ? [this.makeDefaultTemplate()] : []
+    ).concat(
+      customTemplates.filter(identity) as ResultTemplate<DocumentFragment>[]
+    );
+
+    this.resultTemplatesManager.registerTemplates(...templates);
+    onReady();
+  }
+
+  public getResultId(
```
btaillon, line 174: if this isn't used outside of this component, I think it should be `private`.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,316 @@
...
+    result: Result | FoldedCollection,
+    resultListState: FoldedResultListState | ResultListState
+  ) {
+    return (
+      this.getUnfoldedResult(result).uniqueId + resultListState.searchResponseId
+    );
+  }
+
+  public getUnfoldedResult(result: Result | FoldedResult): Result {
```
btaillon, line 183: if this isn't used outside of this component, I think it should be `private`.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,316 @@
...
+    return {
+      content,
+      conditions: [],
+    };
+  }
+
+  public async registerResultTemplates(
+    elements: NodeListOf<TemplateElement>,
+    includeDefaultTemplate = true,
+    onReady: () => void,
```
btaillon, line 148: Instead of taking `onReady` as a param, the caller could just `await` this function.

Instead of taking `onError` as a param, perhaps we could `throw` here and `catch` from the caller.

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: yeah, that makes sense

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: it is used outside

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: also used outside!

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -0,0 +1,317 @@
...
+      this.resultListControllerProps = {
+        options: {
+          fieldsToInclude: this.determineAllFieldsToInclude(
+            opts.fieldsToInclude
+          ),
+        },
+      };
+    }
+
+    this.registerResultTemplates(
```
btaillon, line 102: since `registerResultTemplates` is async and public, the class which initializes `ResultListCommon` could call and await `registerResultTemplates` instead of passing `onReady` and `onError`

packages/atomic/src/components/result-lists/result-list-common.tsx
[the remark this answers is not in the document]
btaillon: yes, I think a new prop interface would be good. Basically, I want to eliminate `AtomicResultListBaseComponent` and I don't want to `implement` an interface in addition to initializing `ResultListCommon`.

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: has this ever worked? I did some logging and noticed that it was always saying that it was true, because `host.offsetHeight === 0`. host only has a shadow dom, so I'm changing it to `this.host.shadowRoot?.firstElementChild.offsetHeight` which seems to work. lmk if there are any issues with it

btaillon: as far as I'm aware, this was supposed to work 🤔 
If this fixes it, them your change sounds good

packages/atomic/src/components/result-lists/result-list-common.tsx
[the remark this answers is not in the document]
liviarett: alright, I think I got it!

packages/atomic/src/components/result-lists/result-list-common.tsx
liviarett: the `onReady` function has the problem of initialize not being awaited either. I'm looking into it in a separate task, but it's looking very, very tricky. I'll leave it as is for now.
about the `onError`, I personally think it looks neater like this instead of then implementing the catch on each component. having it as pros makes it more obvious, I think!
I'll merge like this anyway and can always change after!

PR #1884, opened 2022-03-30
feat(headless, quantic): search hub added to payload of case assist analytics"

mmitiche:
[SFINT-4444](https://coveord.atlassian.net/browse/SFINT-4444)

Added the possibility to specify the searcHub when sending Case Assist analytics events.
The following changes were made:

- CaseAssistClientProvider added to the CaseAssistClient, the CaseAssistClientProvider is the responsible of providing the CaseAssistClient with the value of the search hub.

- Search hub added to the case assist Engine Configuration.

- Search hub added to the case assist Engine state.

- new attribute searchHub added to the QuanticCaseAssistInterface.

SearchHub


ThibodeauJF (approved):

packages/headless/src/app/case-assist-engine/case-assist-engine-configuration.ts
```diff
@@ -24,6 +24,14 @@ export interface CaseAssistEngineConfiguration extends EngineConfiguration {
    *  Moreover, this information can be referred to in query expressions and QPL statements by using the $locale object.
    */
   locale?: string;
+  /**
+   * The first level of origin of the request, typically the identifier of the graphical search interface from which the request originates.
```
nathanlb, line 28: Given it isn't for a search interface I think we should rewrite this.

packages/quantic/force-app/main/default/lwc/quanticCaseAssistInterface/quanticCaseAssistInterface.js
```diff
@@ -32,6 +32,13 @@ export default class QuanticCaseAssistInterface extends LightningElement {
    * @type {string}
    */
   @api caseAssistId;
+  /**
+   * The search interface [search hub](https://docs.coveo.com/en/1342/).
```
nathanlb, line 36: not a search interface so we should change it to something like "case assist interface"

nathanlb (commented):
Looks good generally. You're clearly getting comfortable in this project and it's awesome. 👏 
However I think some comments/names copied from the search use-case should be rewritten to fit the case assist use-case.

packages/headless/src/api/analytics/analytics.ts
```diff
@@ -139,6 +141,14 @@ export class AnalyticsProvider implements SearchPageClientProvider {
   }
 }
 
+export class CaseAssistAnalyticsProvider implements CaseAssistClientProvider {
+  constructor(private state: StateNeededByCaseAssistAnalytics) {}
+
+  public getOriginLevel1() {
```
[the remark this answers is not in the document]
mmitiche: Isn't the originLevel1 and searchHub two sides for the same coin? and the names `originLevel1` and `searchHub` can be interchangeable?
Also Changing the name of this method will oblige us to also change the name of this method in the `CaseAssistClientProvider` interface in coveo.analytics.js which I'm afraid  will introduce a breaking change to this library 😕

packages/headless/src/app/case-assist-engine/case-assist-engine-configuration.ts
```diff
@@ -24,6 +24,14 @@ export interface CaseAssistEngineConfiguration extends EngineConfiguration {
...
+   * Coveo Machine Learning models use this information to provide contextually relevant output.
```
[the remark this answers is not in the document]
mmitiche: Absolutely correct👌 I'll rewrite this thank you!

packages/headless/src/api/analytics/analytics.ts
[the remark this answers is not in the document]
nathanlb: Ah darn. Well to avoid the breaking change let's leave it then. But while the two terms are used to indicate the same thing, I don't think we should regard them as interchangeable. We use the searchHub as the value for originLevel1 but they have different meanings. In the platform and the doc we don't say "assign your query pipeline to a given originLevel1". Do you know what I mean? The concept of searchHub is already often misunderstood so giving it an alias can be even more confusing for users. But that's just my feeling and the breaking change argument is strong so no worries.

packages/quantic/force-app/examples/main/lwc/exampleQuanticDocumentSuggestion/exampleQuanticDocumentSuggestion.js
```diff
@@ -17,6 +19,12 @@ export default class ExampleQuanticDocumentSuggestion extends LightningElement {
       description: 'The id of the case assit engine associated.',
```
lbergeron, line 19:
```suggestion
      description: 'The id of the case assist engine associated.',
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticDocumentSuggestion/exampleQuanticDocumentSuggestion.js
```diff
@@ -17,6 +19,12 @@ export default class ExampleQuanticDocumentSuggestion extends LightningElement {
...
       defaultValue: 'case-assist-engine',
     },
+    {
+      attribute: 'searchHub',
```
lbergeron, line 23: I think we should not include the `searchHub` option in this form. The reason is that the option is being set on the interface component, and not on the tested component (`QuanticDocumentSuggestion` in this case).

Why not simply affect the `searchHub` attribute a static value?

mmitiche: You are right! I just updated this thank you👌

lbergeron (approved):

PR #1854, opened 2022-03-24
feat(atomic): added custom styling to smart snippet

btaillon:
https://coveord.atlassian.net/browse/KIT-1499

Example:
```html
<atomic-smart-snippet>
  <template>
    <style>
      b { color: red; }
    </style>
  </template>
</atomic-smart-snippet>
```

PR #1887, opened 2022-03-31
feat(atomic): added custom styling to smart snippet answer

btaillon:
https://coveord.atlassian.net/browse/KIT-1499

Follow-up to https://github.com/coveo/ui-kit/pull/1854/files, but improved.

This adds custom styling to smart snippet answers by doing the following:
```html
<atomic-smart-snippet>
  <template>
    <style>
      b { color: red; }
    </style>
  </template>
</atomic-smart-snippet>
```

This also adds a prop which can be used as an alternative in frameworks like React, Angular or Vue.

### DIFF
--- a/packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
+++ b/packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
@@ -1,5 +1,6 @@
-import {Component, h, Prop, Event, EventEmitter} from '@stencil/core';
+import {Component, h, Prop, Event, EventEmitter, Host} from '@stencil/core';
 import {sanitize} from 'dompurify';
+import {sanitizeStyle} from '../../utils/utils';
 
 /**
  * @internal
@@ -11,6 +12,7 @@ import {sanitize} from 'dompurify';
 })
 export class AtomicSmartSnippetAnswer {
   @Prop({reflect: true}) htmlContent!: string;
+  @Prop({reflect: true}) innerStyle?: string;
 
   @Event({
     eventName: 'atomic/smartSnippet/answerRendered',
@@ -22,7 +24,23 @@ export class AtomicSmartSnippetAnswer {
     this.answerRendered.emit({height: this.wrapperElement!.scrollHeight});
   }
 
-  public render() {
+  private get sanitizedStyle() {
+    if (!this.innerStyle) {
+      return undefined;
+    }
+    return sanitizeStyle(this.innerStyle);
+  }
+
+  private renderStyle() {
+    const style = this.sanitizedStyle;
+    if (!style) {
+      return;
+    }
+    // deepcode ignore ReactSetInnerHtml: Defined by implementer and sanitized by dompurify
+    return <style innerHTML={style}></style>;
+  }
+
+  private renderContent() {
     return (
       <div class="wrapper" ref={(element) => (this.wrapperElement = element)}>
         {/* deepcode ignore ReactSetInnerHtml: Sanitized by back-end + dompurify */}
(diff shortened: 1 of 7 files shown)


packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -68,12 +69,38 @@ export class AtomicSmartSnippet implements InitializableComponent {
...
+   * Example:
+   * ```ts
+   * smartSnippet.snippetStyle = `
+   *   b {
+   *     color: blue;
+   *   }
+   * `;
+   * ```
+   *
+   * Alternatively, you can style the snippet by inserting a template element like this:
```
btaillon, line 84: Although using a `template` with a `style` element is pretty much the "primary way" of doing it, I think it's easier to document like this.

packages/atomic/src/utils/utils.ts
```diff
@@ -96,3 +97,15 @@ export function elementHasAncestorTag(
   if (parentElement.tagName === tagName.toUpperCase()) return true;
   return elementHasAncestorTag(parentElement, tagName);
 }
+
+export function sanitizeStyle(style: string) {
+  const purifiedOuterHTML = sanitize(`<style>${style}</style>`, {
+    ALLOWED_TAGS: ['style'],
+    ALLOWED_ATTR: [],
+    FORCE_BODY: true,
```
btaillon, line 105: `FORCE_BODY` is necessary to sanitize this. Otherwise, an empty string is returned.

Excerpt from their doc:
```js
// glue elements like style, script or others to document.body and prevent unintuitive browser behavior in several edge-cases (default is false)
var clean = DOMPurify.sanitize(dirty, {FORCE_BODY: true});
```

jpmarceau (approved):
👍

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
```diff
@@ -22,7 +24,23 @@ export class AtomicSmartSnippetAnswer {
     this.answerRendered.emit({height: this.wrapperElement!.scrollHeight});
   }
 
-  public render() {
+  private get sanitizedStyle() {
```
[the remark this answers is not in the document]
ThibodeauJF: I'm not sure this should apply here just like the result templates style tags. A customer could put any <script> tag anywhere in the page.

On the other end, what if it removes some valuable css ?

ThibodeauJF (approved):

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
olamothe: Why do you find this easier ? 🤔  

My first instinct would have been to document the recommended/primary way first, with explicit keywords that explains/reassure readers that "this is the recommended way".

And not leave them to figure it out on their own.

olamothe (approved):

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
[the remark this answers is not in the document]
btaillon: > I'm not sure this should apply here just like the result templates style tags.

My worry is mostly somebody managing to escape out of the `style` tag. I'm not sure if that's possible in the first place, I couldn't find much information online, but I see that as being especially unlikely for result template style tags since they are already part of results, which are a more "obvious" thing to proof against XSS.

> A customer could put any <script> tag anywhere in the page.

my concern is mostly string interpolation / passing values through from a user, similar to storybook.

PR #1891, opened 2022-03-31
ci(cdn): deploy patch version to CDN

olamothe:
To allow [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) capabilities to our clients, we need to deploy patch versions to our CDN.

That way a link on a specific patch version can be "fixed", have its SHA-512 computed, and will not change over time.

As opposed to minor version, which could contain hotfix in the future.

@npushkarskii & @jpmarceau : Eventually (really not needed short term/not urgent), we should arrange for CDN links to be auto generated in our doc website for each release we do, similar to JSUI. 

For example with last release:
 
https://docs.coveo.com/en/m16b2503/javascript-search-framework/march-2022-release

https://coveord.atlassian.net/browse/KIT-1512

### DIFF
--- a/.deployment.config.json
+++ b/.deployment.config.json
@@ -16,33 +16,11 @@
   },
   "ordered_phases": [
     {
-      "id": "deploy-headless-to-s3-latest",
+      "id": "deploy-headless-patch-to-s3-version",
       "s3": {
         "bucket": "{terraform.infra.infra.bucket_binaries}",
-        "directory": "proda/StaticCDN/headless/latest",
+        "directory": "proda/StaticCDN/headless/v$[HEADLESS_PATCH_VERSION]",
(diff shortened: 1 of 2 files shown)


.deployment.config.json
```diff
@@ -16,55 +16,55 @@
   },
   "ordered_phases": [
     {
-      "id": "deploy-headless-to-s3-latest",
+      "id": "deploy-headless-patch-to-s3-version",
```
olamothe, line 19: I re-ordered the phases in this step so that it's : 
* headless patch -> minor -> major -> latest
* atomic patch -> minor -> major -> latest
* atomic/headless patch -> minor -> latest

ThibodeauJF (approved):

PR #1894, opened 2022-04-01
feat(atomic): allow to add children inside atomic-result-children

liviarett:
https://coveord.atlassian.net/browse/KIT-1535

so we could add something like
Screenshot 2022-04-01 at 17 59 22
only when there are children
Screenshot 2022-04-01 at 17 59 29


btaillon (approved):
sweet!

olamothe (approved):

PR #1895, opened 2022-04-01
fix(atomic): fix error console when state is undefined

aelfaleh:
This change fix: https://coveord.atlassian.net/browse/SVCC-532

### DIFF
--- a/packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
+++ b/packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
@@ -389,7 +389,7 @@ export class AtomicResultList implements InitializableComponent {
   }
 
   public componentDidRender() {
-    if (this.resultListState.firstSearchExecuted) {
+    if (this.resultListState?.firstSearchExecuted) {
       this.listWrapperRef?.classList.remove('placeholder');
     }
   }


nathanlb (approved):
Do we know why the state would be undefined? is there a lifecycle loop that runs before it's initialized?

aelfaleh:
> Do we know why the state would be undefined? is there a lifecycle loop that runs before it's initialized?
 
I don't know 😉

btaillon (commented):
this should never have been undefined after render in the first place, there must be a deeper issue

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
```diff
@@ -389,7 +389,7 @@ export class AtomicResultList implements InitializableComponent {
   }
 
   public componentDidRender() {
-    if (this.resultListState.firstSearchExecuted) {
+    if (this.resultListState?.firstSearchExecuted) {
```
lbergeron, line 392: Is it possible to add a test to avoid getting this bug again in the future?

lbergeron (approved):

packages/atomic/src/components/atomic-result-list/atomic-result-list.tsx
aelfaleh: working on it

liviarett:
> this should never have been undefined after render in the first place, there must be a deeper issue
@btaillon I think it's because the `initalize` function is not being waited for, so sometimes it hasn't finished when `componentDidRender` happens. I expect this task https://coveord.atlassian.net/jira/software/c/projects/KIT/boards/764?modal=detail&selectedIssue=KIT-1522&quickFilter=2211 to fix it

ThibodeauJF (changes requested):
You can close this PR. The bug has been fixed here  https://github.com/coveo/ui-kit/pull/1817/files (`packages/atomic/src/utils/initialization-utils.tsx`)

The console error was a side effect.

aelfaleh:
> You can close this PR. The bug has been fixed here https://github.com/coveo/ui-kit/pull/1817/files (`packages/atomic/src/utils/initialization-utils.tsx`)
> 
> The console error was a side effect.

Thank you

PR #1898, opened 2022-04-05
docs(atomic): add cypress tests for folded component

liviarett:
https://coveord.atlassian.net/browse/KIT-1521


packages/atomic/cypress/integration/result-list/result-list.cypress.ts
```diff
@@ -1,158 +1,208 @@
...
+  [
+    typeof ResultListSelectors,
+    string,
+    (template?: HTMLElement | undefined) => (fixture: TestFixture) => void,
+    string,
+    boolean
+  ]
+> = [foldedResultListConfig, resultListConfig];
+
+configs.forEach(
```
olamothe, line 42: Nice ! 👍

packages/atomic/cypress/integration/result-list/result-list.cypress.ts
```diff
@@ -1,158 +1,208 @@
...
+];
+const resultListConfig = [
+  ResultListSelectors,
+  resultListComponent,
+  addResultList,
+  'Result List Component',
+  false,
+];
+
+const configs: Array<
```
olamothe, line 32: Kinda nitpicky, but instead of this being an array for configuration, it could have been an object.

Purely so there's key (with names) to describe what each property represent. So the next time we need to modify this, it's a bit more straightforward versus having to map on line 43 what each value represent.

```
const foldedResultListConfig = {
  componentSelectors: FoldedResultListSelectors,
  componentTag: foldedResultListComponent,
  addResultFn: addFoldedResultList,
  title: 'Folded result List Component',
  isFolded: true,
};
```

packages/atomic/cypress/integration/result-list/result-list.cypress.ts
```diff
@@ -1,158 +1,208 @@
...
+    typeof ResultListSelectors,
+    string,
+    (template?: HTMLElement | undefined) => (fixture: TestFixture) => void,
+    string,
+    boolean
+  ]
+> = [foldedResultListConfig, resultListConfig];
+
+configs.forEach(
+  ([componentSelectors, componentTag, addResultFn, title, isFolded]) => {
```
olamothe, line 43: I think `isFolded` could be omitted from the config, and just derived from `componentTag === 'atomic-folded-result-list'`

olamothe (approved):

packages/atomic/cypress/integration/result-list/result-list.cypress.ts
liviarett: coolz, will change

packages/atomic/cypress/integration/result-list/folded-result-list-utils.ts
```diff
@@ -0,0 +1,173 @@
...
+export function withFoldedResult(
+  response: SearchFoldedResponseSuccess
+): SearchFoldedResponseSuccess {
+  response.results[0] = elfCupsAndAllies(opercs, [
+    elfCups(elfCupsAndAllies(), [stubbleLichens(pyrenomycetes)]),
+  ]);
+  return response;
+}
+
+const reusableFields = {
```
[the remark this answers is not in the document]
olamothe: Yes, I think we can tweak the tests a bit so that instead of looking for specific we just verify that "there's any string".

Also, that source should pretty much be static. It's a push source, so it won't change for no reason over time, it will require someone to repush/rebuild manually.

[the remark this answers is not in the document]
liviarett: I changed it to request the animals, but I don't think there is a way to guarantee that we are checking that a result with children renders correctly and that a result without children also renders correctly without knowing what the data look like. I though about accessing the `results` array and deciding from that (getting the index of the top level result that matches the criteria, for example), but that seems like a bigger job.
if the source shouldnt change, shall we try to leave it like this and see if it maintains?

packages/atomic/cypress/integration/result-list/folded-result-list.cypress.ts
```diff
@@ -0,0 +1,134 @@
...
+        )
+      )
+      .init();
+
+    FoldedResultListSelectors.firstResult()
+      .contains('Pyrenomycetes')
+      .should('be.visible');
+
+    getAtomicResultAtIndex(FoldedResultListSelectors, 0)
+      .find('atomic-result-link')
```
lvu285, line 62: can we have all selectors in a separate file? so it's will be easy to manage them in the future

packages/atomic/cypress/integration/result-list/folded-result-list-utils.ts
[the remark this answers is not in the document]
olamothe: Yep !

packages/atomic/cypress/fixtures/test-fixture.ts
```diff
@@ -37,6 +50,7 @@ export class TestFixture {
   private responseModifiers: SearchResponseModifierPredicate[] = [];
   private returnError = false;
   private reflectStateInUrl = true;
+  private redirected = false;
```
ThibodeauJF, line 53: 👀

packages/atomic/cypress/integration/result-list/folded-result-list-utils.ts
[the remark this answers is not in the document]
ThibodeauJF: You can pick a result out of the one on the org using the `permanentid` field and using the AQ

```
loadAdvancedSearchQueryActions(engine).updateAdvancedSearchQueries({ 
  aq: "@permanentid=5570a7c57da54559b2dde4a1db26c1a9f2b05f277fbb91db49bda76fc8cb",
})
```
Screen Shot 2022-04-05 at 4 37 27 PM

packages/atomic/cypress/fixtures/test-fixture.ts
liviarett: that is because typescript was complaining! it should start as `false`

packages/atomic/cypress/integration/result-list/folded-result-list.cypress.ts
liviarett: @lvu285 can you please let me know if my latest commit is good enough?

packages/atomic/cypress/fixtures/test-fixture.ts
```diff
@@ -1,9 +1,22 @@
+import {Result} from '@coveo/headless';
 import {CyHttpMessages} from 'cypress/types/net-stubbing';
 import {i18n} from 'i18next';
 import {SearchResponseSuccess} from '../../../headless/dist/definitions/api/search/search/search-response';
 import {AnalyticsTracker, AnyEventRequest} from '../utils/analyticsUtils';
 import {buildTestUrl} from '../utils/setupComponent';
 
+export interface ResultWithFolding extends Result {
```
btaillon, line 8: I would not add any interface that relates to results inside `test-fixture.ts`

packages/atomic/cypress/integration/result-list/folded-result-list-selectors.ts
```diff
@@ -0,0 +1,70 @@
...
+    .shadow();
+}
+
+export function getAtomicGrandChildResultAtIndex(
+  grandChildIndex: number,
+  childResultIndex = 0,
+  resultIndex = 0
+) {
+  return getAtomicChildResultAtIndex(childResultIndex, resultIndex)
+    .find('atomic-result-children')
```
lvu285, line 65: how about variable for  `atomic-result-children` & `atomic-result`?

packages/atomic/cypress/integration/result-list/folded-result-list-utils.ts
```diff
@@ -0,0 +1,173 @@
+import {
+  generateComponentHTML,
+  ResultWithFolding,
+  SearchFoldedResponseSuccess,
+} from '../../fixtures/test-fixture';
+
+export function withFoldedResult(
```
btaillon, line 7: this could be moved into `folded-result-list-actions.ts` along with `addFoldedResultList`

packages/atomic/cypress/integration/result-list/folded-result-list-utils.ts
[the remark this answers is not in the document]
btaillon: you can also filter on the `@foldedcollection`.

packages/atomic/cypress/integration/result-list/folded-result-list.cypress.ts
```diff
@@ -0,0 +1,128 @@
...
+          ])
+        )
+      )
+      .init();
+
+    getAtomicResultChildren()
+      .children()
+      .eq(0)
+      .contains('Before children!')
+      .should('have.attr', 'slot', 'before-children');
```
lvu285, line 104: I would suggest not to repeat the input value "before-children" & "after-children"

packages/atomic/cypress/fixtures/test-fixture.ts
liviarett: this actually comes from headless, but doesn't seem to be exported there. would it make sense to change it in headless? I don't see anywhere else that it would fit in the current structure, would need to add a separate file maybe?

packages/atomic/cypress/integration/result-list/folded-result-list.cypress.ts
```diff
@@ -0,0 +1,128 @@
...
+          ])
+        )
+      )
+      .init();
+
+    FoldedResultListSelectors.firstResult()
+      .contains('Pyrenomycetes')
+      .should('be.visible');
+
+    getAtomicChildResultAtIndex(0)
```
lvu285, line 38: we also can create an assert function for 
```
 getAtomicChildResultAtIndex(0)
      .find('atomic-result-link')
      .contains('Stubble Lichens');
```

packages/atomic/cypress/integration/result-list/folded-result-list.cypress.ts
```diff
@@ -0,0 +1,128 @@
...
+      .find('atomic-result-link')
```
btaillon, line 64: could be replaced with `find(resultLinkComponent)` (defined in `result-link-selectors.ts`)

packages/atomic/cypress/integration/result-list/folded-result-list-utils.ts
liviarett: I forgot to remove this, I'm not using it anymore!

packages/atomic/cypress/fixtures/test-fixture.ts
btaillon: I wouldn't export it from Headless, I think it could fit in `folded-result-list-actions.ts` (alongside the methods that need it).

packages/atomic/cypress/integration/result-list/folded-result-list.cypress.ts
btaillon: you can find examples of what Lucille is talking about in files like `smart-snippet-assertions.ts`

packages/atomic/cypress/integration/result-list/result-list.cypress.ts
```diff
@@ -5,154 +5,198 @@ import {
...
+};
+
+const resultListConfig = {
+  componentSelectors: ResultListSelectors,
+  componentTag: resultListComponent,
+  addResultFn: addResultList,
+  title: 'Result List Component',
+};
+
+const configs: Array<typeof foldedResultListConfig> = [
```
btaillon, line 33: instead of adding the type `Array<typeof foldedResultListConfig>`, you may be able to use `as const`:
```ts
const configs = [
  ...
] as const;
```

btaillon (approved):
very good idea to re-use the `result-list` tests!

lvu285:
good job @liviarett! Thanks for the tests!

lvu285 (approved):

packages/atomic/cypress/integration/result-list/result-list.cypress.ts
liviarett: silly typescript https://stackoverflow.com/a/66993654/8454914

liviarett: and great suggestion

PR #1899, opened 2022-04-05
docs(atomic): add folded result list to storybook

liviarett:
https://coveord.atlassian.net/browse/KIT-1531


olamothe (approved):
Were you able to run storybook locally ? The command is a bit hidden IIRC...

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.stories.tsx
```diff
@@ -0,0 +1,402 @@
...
+          response.body.results[0] = resultWithChildren;
+          return response;
+        },
+      },
+    },
+    additionalChildMarkup: () => html`
+      <atomic-result-template>
+        <template>
+          <atomic-result-link></atomic-result-link>
+          <div style="padding-left: 15px; margin-top: 20px;">
```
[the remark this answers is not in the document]
liviarett: yes! I have that component already, but because we are waiting for Johnny's designs, it doesn't do anything... I'll make sure I update storybook when I do them. will add to the jira ticket

liviarett:
yes, I managed to run it without any problems with `npm run storybook`

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.stories.tsx
```diff
@@ -0,0 +1,402 @@
+import {html} from 'lit-html';
+import defaultStory from '../../../.storybook/default-story';
+
+const {defaultModuleExport, exportedStory} = defaultStory(
+  'Atomic/ResultList',
```
olamothe, line 5: This should be `Atomic/FoldedResultList` or there will be a "collision" in the page ID that gets generated.

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.stories.tsx
```diff
@@ -0,0 +1,402 @@
...
+              </atomic-result-children-template>
+            </atomic-result-children>
+          </div>
+        </template>
+      </atomic-result-template>
+    `,
+  }
+);
+
+const resultWithChildren = {
```
ThibodeauJF, line 37: I think what Benjamin indexed recently is a more appropriate content

ThibodeauJF (approved):

packages/atomic/src/components/atomic-folded-result-list/atomic-folded-result-list.stories.tsx
ThibodeauJF: Could be accessed with an `aq` instead of mocking the response

liviarett: took me my time to find out what `aq` meant, but seems like a much better approach, thanks!

PR #1900, opened 2022-04-05
chore(atomic): bump output target package to 1.x

olamothe:
https://coveord.atlassian.net/browse/KIT-1540


ThibodeauJF (approved):

olamothe:
> I'm slightly tempted to wait to have more demos before bumping to 1.0, even if that may take a very long time.

Problem is quite a bit of people are using it/looking into using it, and the 0.x is not inspiring confidence.

It is documented officially on the website, after all.

liviarett (approved):

PR #1902, opened 2022-04-07
feat(atomic): allow grandchild children to inherit templates from chi…

liviarett:
…ldren

beautiful 
Screenshot 2022-04-07 at 16 24 51

Screenshot 2022-04-07 at 16 25 16

https://coveord.atlassian.net/browse/KIT-1527

### DIFF
--- a/packages/atomic/src/components/atomic-result-children/atomic-result-children.tsx
+++ b/packages/atomic/src/components/atomic-result-children/atomic-result-children.tsx
@@ -1,11 +1,18 @@
-import {FoldedResult} from '@coveo/headless';
-import {Component, Element, State, h, Host} from '@stencil/core';
+import {FoldedResult, Result, ResultTemplatesManager} from '@coveo/headless';
+import {Component, Element, State, h, Host, Listen, Prop} from '@stencil/core';
 import {Bindings, InitializeBindings} from '../../utils/initialization-utils';
 import {elementHasAncestorTag} from '../../utils/utils';
-import {ResultContext} from '../result-template-components/result-template-decorators';
+import {
+  ResultContext,
+  ChildTemplatesContext,
+  ChildTemplatesContextEvent,
+} from '../result-template-components/result-template-decorators';
 import {ResultListCommon} from '../result-lists/result-list-common';
+import {TemplateContent} from '../result-templates/result-template-common';
 
 const childTemplateComponent = 'atomic-result-children-template';
+
+const componentTag = 'atomic-result-children';
 
 /**
  * TODO:
@@ -17,6 +24,8 @@ const childTemplateComponent = 'atomic-result-children-template';
 })
 export class AtomicResultChildren {
   @InitializeBindings() public bindings!: Bindings;
+  @ChildTemplatesContext()
+  public templatesManager!: ResultTemplatesManager<TemplateContent>;
   @ResultContext({folded: true}) private result!: FoldedResult;
 
   public resultListCommon?: ResultListCommon;
@@ -26,15 +35,23 @@ export class AtomicResultChildren {
   @State() public ready = false;
   @State() public templateHasError = false;
 
+  @Prop() inheritTemplates = false;
+
+  @Listen('atomic/resolveChildTemplates')
+  public resolveChildTemplates(event: ChildTemplatesContextEvent) {
+    event.preventDefault();
+    event.detail(this.resultListCommon?.resultTemplatesManager);
+  }
+
   public async initialize() {
     const childrenTemplates = Array.from(
       this.host.querySelectorAll(childTemplateComponent)
     ).filter((template) => {
       return !elementHasAncestorTag(template, childTemplateComponent);
     });
-    if (!childrenTemplates.length) {
+    if (!childrenTemplates.length && !this.inheritTemplates) {
       this.error = new Error(
-        `The "atomic-result-children" component requires at least one "${childTemplateComponent}" component.`
+        `The "${componentTag}" component requires at least one "${childTemplateComponent}" component.`
       );
       return;
     }
@@ -44,6 +61,7 @@ export class AtomicResultChildren {
         host: this.host,
         bindings: this.bindings,
         templateElements: this.host.querySelectorAll(childTemplateComponent),
+        includeDefaultTemplate: false,
         onReady: () => {
           this.ready = true;
         },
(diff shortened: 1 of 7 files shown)


packages/atomic/src/pages/index.html
```diff
@@ -347,16 +348,14 @@
...
                           <atomic-result-link></atomic-result-link>
-                          <atomic-result-children>
-                            <atomic-result-children-template>
-                              <template>
-                                <atomic-result-link></atomic-result-link>
-                              </template>
-                            </atomic-result-children-template>
-                          </atomic-result-children>
+                          <div style="padding-left: 15px">
+                            <atomic-result-children inherit-templates> <!-- GRANDCHILD --> </atomic-result-children>
```
ThibodeauJF, line 357: Ohh interesting, what does `inherit-templates` do?

packages/atomic/src/components/atomic-result-children/atomic-result-children.tsx
```diff
@@ -54,30 +72,51 @@ export class AtomicResultChildren {
     }
   }
 
+  private selectInheritedTemplate(result: Result) {
+    const content = this.templatesManager?.selectTemplate(result);
+    if (content) {
```
ThibodeauJF, line 77: Nitpick: early return `if (!content)` in this method and the next

liviarett: no prob!

packages/atomic/src/pages/index.html
liviarett: so, usually if you add a `atomic-result-children` component without any `atomic-result-children-template`, you get a warning (because we don't have default templates for children results, so you need to specify them.
howeeeverrr if you then add `inherit-templates` to it, you are specifying that you want your grandchild result to use the same template as you specified for the child results (so it only works when you have nested `atomic-result-children`)

a few different cases for structure + the template options that the results will have:
```
  atomic-result-children
    template 1
    template 2
    atomic-result-children with inherit-templates

parent 
  child: template 1, template 2
    grandchild: template 1, template 2
```

```
  atomic-result-children
    template 1
    template 2
    atomic-result-children with inherit-templates
      template 3
      template 4

parent 
  child: template 1, template 2
    grandchild: template 3, template 4, template 1, template 2 (in this order)
```

```
  atomic-result-children
    template 1
    template 2
    atomic-result-children WITHOUT inherit-templates
      template 3
      template 4

parent 
  child: template 1, template 2
    grandchild: template 3, template 4
```

packages/atomic/src/components/atomic-result-children/atomic-result-children.tsx
```diff
@@ -26,15 +34,23 @@ export class AtomicResultChildren {
   @State() public ready = false;
   @State() public templateHasError = false;
 
+  @Prop() inheritTemplates = false;
```
liviarett, line 37: add documentation for this prop in this PR

packages/atomic/src/components/result-template-components/result-template-decorators.tsx
```diff
@@ -99,3 +100,34 @@ export const resultContext = (element: Element) =>
...
+
+export function ChildTemplatesContext() {
+  return (component: ComponentInterface, resultVariable: string) => {
+    const {componentWillRender} = component;
+    component.componentWillRender = function () {
+      const element = getElement(this);
+      const event = buildCustomEvent(
+        childTemplatesContextEventName,
+        (result: ResultTemplatesManager<TemplateContent> | undefined) => {
+          const resultListManager = (this as any).resultListCommon
```
liviarett, line 119: oops, will fix this

packages/atomic/src/components/result-template-components/result-template-decorators.tsx
```diff
@@ -99,3 +100,34 @@ export const resultContext = (element: Element) =>
...
 }
+
+type ChildTemplatesContextEventHandler = (
+  result: ResultTemplatesManager<TemplateContent> | undefined
+) => void;
+export type ChildTemplatesContextEvent =
+  CustomEvent<ChildTemplatesContextEventHandler>;
+const childTemplatesContextEventName = 'atomic/resolveChildTemplates';
+
+export function ChildTemplatesContext() {
```
liviarett, line 111: document this here too

packages/atomic/src/components/atomic-result-children/atomic-result-children.tsx
olamothe: How about we make this true by default ? 🤔

What would be the cons ?

olamothe (approved):
👍 Nice !

packages/atomic/src/components/atomic-result-children/atomic-result-children.tsx
liviarett: what I can think of is: the top level one (that can't really inherit the templates) will have that setting by default, so will never get this warning
```js
    if (!childrenTemplates.length && !this.inheritTemplates) {
      this.error = new Error(
        `The "${componentTag}" component requires at least one "${childTemplateComponent}" component.`
      );
      return;
    }
```
unless we do some sort of extra checking later on when all events have been fired, for example in the render method, but I'm assuming that's not ideal... or maybe in `ChildTemplatesContext` we can send the config, like `ChildTemplatesContext(inheritsTemplate)` and if it does but there is nothing to inherit, like 
```ts
      const event = buildCustomEvent(
        childTemplatesContextEventName,
        (result: ResultTemplatesManager<TemplateContent> | undefined) => {
          const resultListManager = (this as AtomicResultChildren)
            .resultListCommon?.resultTemplatesManager !== result ? result : null;
          if (inheritTemplate && !resultListManager) {
            this.error = new Error(
             `The "${componentTag}" component requires at least one "${childTemplateComponent}" component.`
            );
          } else {
             this[resultVariable] = resultListManager;
          }
        }
      );
```
I'll do that and see how it looks!

liviarett: actually, this doesn't really work.. we can't validate this unless we look at the parents..
I think leaving it as false makes the behaviour a bit more explicit, and makes it so that we don't need to distinguish between the child and the grandchild in the code for the validation (the user could still set `inherit-templates=true` for the top level children and not get the warning, but at least they will have read the docs about this feature and that should tell them that it only works for grandchildren.

btaillon (approved):

PR #1906, opened 2022-04-08
test(atomic): add Stencil samples & smoke tests

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1490


packages/samples/stencil/src/components/sample-component/sample-component.tsx
```diff
@@ -0,0 +1,123 @@
+import {Bindings, initializeBindings} from '@coveo/atomic';
```
ThibodeauJF, line 1: Very similar to what the CLI generates

packages/samples/stencil/src/components/sample-result-component/sample-result-component.tsx
```diff
@@ -0,0 +1,35 @@
+import {Result} from '@coveo/atomic/headless';
```
ThibodeauJF, line 1: Very similar to what the CLI generates

packages/samples/stencil/src/components/results-manager/results-manager.tsx
```diff
@@ -0,0 +1,18 @@
+import {Component, h} from '@stencil/core';
```
ThibodeauJF, line 1: Very similar to what the CLI generates

.github/workflows/prbot.yml
```diff
@@ -84,7 +84,14 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/setup
       - uses: ./.github/actions/e2e-vuejs
-
+  e2e-atomic-stencil-test:
```
olamothe, line 87: Would we want to run it also on main branch ?

This will only run on pull request

olamothe (approved):

.github/workflows/prbot.yml
ThibodeauJF: Oh yes, I'll add it to the other workflows!

packages/samples/stencil/src/components/sample-component/sample-component.tsx
```diff
@@ -0,0 +1,123 @@
...
+  private i18nUnsubscribe = () => {};
+
+  @Element() private host!: Element;
+
+  @State() private pagerState!: PagerState;
+  @State() private statusState!: SearchStatusState;
+
+  public async connectedCallback() {
+    try {
+      this.bindings = await initializeBindings(this.host);
```
liviarett, line 33: out of curiosity, why don't we export the decorators for this?

packages/samples/stencil/src/style/index.css
```diff
@@ -0,0 +1,8 @@
+:root {
+  --atomic-primary: orange;
+  --atomic-primary-light: magenta;
```
liviarett, line 3: love the colour palette

liviarett (approved):

packages/samples/stencil/src/components/sample-component/sample-component.tsx
```diff
@@ -0,0 +1,123 @@
...
+  private statusUnsubscribe: Unsubscribe = () => {};
+  private i18nUnsubscribe = () => {};
+
+  @Element() private host!: Element;
+
+  @State() private pagerState!: PagerState;
+  @State() private statusState!: SearchStatusState;
+
+  public async connectedCallback() {
+    try {
```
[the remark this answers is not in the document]
ThibodeauJF: This is an automatically generated sample by the Atomic CLI.

It's necessary when building a custom component using Headless and X Framework (here Stencil) to create the controller and register to its state. The try/catch is there because controller can throw errors on creation if the options are invalid (something that's perhaps not handled in other samples)

packages/samples/stencil/src/components/sample-component/sample-component.tsx
ThibodeauJF: The decorators `InitializeBindings()` (e.g. do a lot of work under the hood. If we export them users might wonder what's happening with other lifecycle methods, they'll have to look at the implementation to understand. Better to have a single method.

Additionally, the decorators would only work for Stencil projects, and we prefer having one way for all frameworks. Also the decorators are Typescript (pre compilation) and we bundle Javascript in the project.

packages/samples/stencil/src/style/index.css
ThibodeauJF: Thanks! 😄

packages/samples/stencil/cypress.json
```diff
@@ -0,0 +1,20 @@
+{
+    "baseUrl": "http://localhost:3666",
```
btaillon, line 2: nice

btaillon (approved):

PR #1909, opened 2022-04-11
fix(atomic): fix atomic react rendering function

olamothe:
Couple seems to have went wrong:

* With the bump to atomic-react@1.0, I also bumped some `cra` dependencies. This caused subtle changes to jest setup in headless-react samples, which means we now have to add extra config for Jest
* `setRenderingFunction` in AtomicResultList could be undefined if called before `initialize`. I modified it so that `ResultListCommon` is always created first, no matter the entry point.

🤞 

https://coveord.atlassian.net/browse/KIT-1557

### DIFF
--- a/packages/atomic-react/package-lock.json
+++ b/packages/atomic-react/package-lock.json
@@ -6,9 +6,9 @@
   "packages": {
     "": {
       "name": "@coveo/atomic-react",
-      "version": "0.7.4",
+      "version": "1.0.0",
       "dependencies": {
-        "@coveo/atomic": "^1.37.0"
+        "@coveo/atomic": "1.42.4"
       },
       "devDependencies": {
         "@types/node": "15.14.9",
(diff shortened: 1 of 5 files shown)


btaillon (approved):

olamothe:
🤔  I might need to perform a "manual" publish of some packages to fix everything

ThibodeauJF (approved):

liviarett:
I'm just a bit confused by this, because `ResultListCommon` needs `this.bindings`, which only becomes available after `initialize`, so how can it be done independently?

olamothe:
> I'm just a bit confused by this, because ResultListCommon needs this.bindings, which only becomes available after initialize, so how can it be done independently?

The initialize function is part of Atomic lifecyle (ie: we control when it's called, with the search interface).

Which means it is possible for any code interacting with our components to call any public function on any component. before we internally have a chance to call `initialize`.

PR #1911, opened 2022-04-12
feat(atomic): attempt to make initialize awaitable

liviarett:
https://coveord.atlassian.net/browse/KIT-1522

I thought I had opened this a while ago and forgot about it... this was an attempt at making the component wait for the initialize function, but it doesn't really work


packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -112,6 +112,7 @@ export function InitializeBindings() {
       componentDidLoad,
       disconnectedCallback,
     } = component;
+    // console.log('element:', element);
```
olamothe, line 115: 👀

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -183,6 +184,7 @@ export function InitializeBindings() {
     };
 
     component.componentDidRender = function () {
+      // console.log('componentDidRender:');
```
olamothe, line 187: 👀

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -195,12 +195,12 @@ export class AtomicSearchInterface {
   }
 
   @Listen('atomic/initializeComponent')
-  public handleInitialization(event: InitializeEvent) {
+  public async handleInitialization(event: InitializeEvent) {
```
olamothe, line 198: A change from sync to async on a public component method would in theory be a breaking change. The method signature has changed, and the function does not return the same type as before.


I don't think this change in particular will necessarily cause problems for actual implementation, but just a heads up that we have to be generally careful with those.

olamothe (approved):
Looks good

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -41,7 +41,7 @@ const initializableElements = ['atomic-search-interface', 'atomic-external'];
 
 /**
  * Retrieves `Bindings` on a configured parent search interface.
- * @param event Element on which to dispatch the event, which must be the child of a configured "atomic-search-interface" or "atomic-external" element.
+ * @param element Element on which to dispatch the event, which must be the child of a configured "atomic-search-interface" or "atomic-external" element.
  * @returns A promise that resolves on initialization of the parent "atomic-search-interface" or "atomic-external" element, and rejects when it's not the case.
  */
 export const initializeBindings = (element: Element) =>
```
btaillon, line 47: I would also update the type of `initialize` in `InitializableComponent` from `initialize?: () => void;` to `initialize?: () => void | Promise<void>;`

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -122,36 +123,36 @@ export function InitializeBindings() {
...
+            const updateLanguage = () => forceUpdate(this);
+            this.bindings.i18n.on('languageChanged', updateLanguage);
+            unsubscribeLanguage = () =>
+              this.bindings.i18n.off('languageChanged', updateLanguage);
+
+            try {
+              // When no controller is initialized, updating a property with a State() decorator, there will be no re-render.
+              // In this case, we have to manually trigger it.
+              await (this.initialize ? this.initialize() : forceUpdate(this));
+              resolve(componentWillLoad && componentWillLoad.call(this));
```
btaillon, line 142: I wouldn't pass anything as a parameter to `resolve`. Do you mean to `await` componentWillLoad?
```suggestion
              componentWillLoad && await componentWillLoad.call(this);
              resolve();
```

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -122,36 +123,36 @@ export function InitializeBindings() {
...
-      }
-
-      return componentWillLoad && componentWillLoad.call(this);
+        const canceled = element.dispatchEvent(event);
+        if (canceled) {
+          this.error = new MissingInterfaceParentError(
+            element.nodeName.toLowerCase()
+          );
+        }
+        resolve();
```
btaillon, line 154: Unless I'm mistaken, I believe this will cause the promise to resolve immediately.

Perhaps this is what you meant to do?
```suggestion
          resolve();
        }
```

packages/atomic/src/utils/initialization-utils.tsx
btaillon: actually, I also think that `resolve` should be moved outside of the `try catch`, just below it. Even when we error, we want to resolve.

PR #1914, opened 2022-04-12
docs(atomic-react): move atomic-react sample to dedicated package

olamothe:
@jpmarceau / @npushkarskii 

This change will make it so that the link we use here: https://docs.coveo.com/en/atomic/latest/usage/frameworks/atomic-react-wrapper/ ( ie: See this "Atomic React Page" link) will no longer work.

Is it complicated to change it to https://github.com/coveo/ui-kit/blob/master/packages/samples/atomic-react/src/App.tsx when this gets merged ? (if you would like to see it before the PR gets merged, it's available at: https://github.com/coveo/ui-kit/blob/KIT-1560/packages/samples/atomic-react/src/App.tsx )

https://coveord.atlassian.net/browse/KIT-1560


packages/samples/atomic-react/public/app.css
```diff
@@ -0,0 +1,168 @@
...
+  --spacing: 16px;
+  --spacing-col: 32px;
+  --search-box-width: 648px;
+}
+
+atomic-search-box {
+  margin-top: 1rem;
+}
+
+atomic-search-interface {
```
ThibodeauJF, line 51: What's the support status of layouts for Atomic React ?

olamothe: Should work... I could do the change in this PR 👍

liviarett:
@olamothe I don't know if I'm being super dumb here, but I am on your branch trying to test your react page with the folded component, and..... the changes I make don't reflect on the page! the webpack console pretends it's done its job, but it doesn't seem to change anything in `public/dist`. any steps I might have missed?

liviarett:
I think my comment above is related to the reason why the test is failing

olamothe:
> @olamothe I don't know if I'm being super dumb here, but I am on your branch trying to test your react page with the folded component, and..... the changes I make don't reflect on the page! the webpack console pretends it's done its job, but it doesn't seem to change anything in `public/dist`. any steps I might have missed?

Yep, was missing a build step in the `dev` command

PR #1916, opened 2022-04-13
feat(atomic-react): add folded result list to atomic react

liviarett:
We'll need to remove the @internal tag first
https://coveord.atlassian.net/browse/KIT-1564


btaillon (approved):
I would also add a demo/test in the `headless-react` samples

olamothe:
I think this will need a pull/merge from master

packages/samples/atomic-react/public/index.html
```diff
@@ -10,5 +10,5 @@
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
   </body>
-  <script src="dist/app.js"></script>
+  <script src="dist/App.js"></script>
```
liviarett, line 13: @olamothe should it be app or App? it seems to work like this, but I haven't changed anything so I'm confused

PR #1917, opened 2022-04-13
feat(commerce): add a context controller for the product listings engine

pvanool:
Add a context controller to give custom context information through the product listings engine.

[COM-1459](https://coveord.atlassian.net/browse/COM-1459)

### DIFF
--- a/packages/headless/src/controllers/product-listing/context/headless-product-listing-context.ts
+++ b/packages/headless/src/controllers/product-listing/context/headless-product-listing-context.ts
@@ -1,13 +1,13 @@
 import {
   ContextPayload,
   ContextValue,
-} from '../../features/context/context-state';
-import {SearchEngine} from '../../app/search-engine/search-engine';
+} from '../../../features/context/context-state';
+import {ProductListingEngine} from '../../../app/product-listing-engine/product-listing-engine';
 import {
   buildCoreContext,
   Context,
   ContextState,
-} from '../core/context/headless-core-context';
+} from '../../core/context/headless-core-context';
 
 export type {Context, ContextState, ContextPayload, ContextValue};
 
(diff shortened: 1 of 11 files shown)


packages/headless/src/controllers/product-listing/context/headless-product-listing-context.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export type {Context, ContextState, ContextPayload, ContextValue};
+
+/**
+ * Creates a `Context` controller instance.
+ *
+ * @param engine - The headless engine.
+ * @returns A `Context` controller instance.
+ */
+export function buildContext(engine: ProductListingEngine): Context {
```
olamothe, line 20: I think in this particular context (pun intended :) ) we could simply re-export `coreContext` as is, and that we would not need to return a merged object with new augmented props.

Since there is no difference between `coreContext` and specific search or product listing context controller, I don't think the specific `get state()`, set(contextPayload)` etc serves any purpose.

And if they need to diverge in behaviour in the future, it will be possible to do what you just did here, which is to overwrite specific props when needed, without any breaking change in the component interface.

IE: I think this would do:
```
**
 * Creates a `Context` controller instance.
 *
 * @returns A `Context` controller instance.
 */
export function buildContext(engine: ProductListingEngine): Context {
  return buildCoreContext(engine);
}
```

Similar thing for  search context controller.

olamothe (approved):

packages/headless/src/controllers/product-listing/context/headless-product-listing-context.ts
btaillon: Don't forget to update the export in `product-listing.index`

btaillon (approved):

packages/headless/src/controllers/product-listing/context/headless-product-listing-context.ts
pvanool: Agreed. I was already on the fence of doing that. Not sure why I didn't do it in the end.

PR #1918, opened 2022-04-13
feat(atomic): add hrefTemplate prop to atomic-result-link

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1558

Works pretty much exactly like the JSUI so users won't be as lost when using this pretty complex and powerful prop

### DIFF
--- a/packages/atomic/src/utils/result-utils.spec.ts
+++ b/packages/atomic/src/utils/result-utils.spec.ts
@@ -21,24 +21,19 @@ describe('buildStringTemplateFromResult', () => {
   });
 
   it('should create string templates', () => {
-    expect(buildStringTemplateFromResult('abc', result, bindings)).toBe('abc');
-    expect(buildStringTemplateFromResult('${title}bar', result, bindings)).toBe(
-      'foobar'
-    );
-    expect(
-      buildStringTemplateFromResult('${raw.source}', result, bindings)
-    ).toBe('the source');
-    expect(buildStringTemplateFromResult('${uri}/abc', result, bindings)).toBe(
-      'http://uri.foo.com/abc'
-    );
+    const templates = [
+      {in: 'abc', out: 'abc'},
+      {in: '${title}bar', out: 'foobar'},
+      {in: '${raw.source}', out: 'the source'},
+      {in: '${uri}/abc', out: 'http://uri.foo.com/abc'},
+      {in: '${window.location.hostname}', out: 'testing.stenciljs.com'},
+    ];
 
-    expect(
-      buildStringTemplateFromResult(
-        '${window.location.hostname}',
-        result,
-        bindings
+    templates.forEach((template) =>
+      expect(buildStringTemplateFromResult(template.in, result, bindings)).toBe(
+        template.out
       )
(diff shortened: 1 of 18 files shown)


packages/atomic/src/utils/result-utils.ts
```diff
@@ -42,3 +43,36 @@ export function bindLogDocumentOpenOnResult(
...
+    let newValue = readFromObject(result, key);
+    if (!newValue) {
+      newValue = readFromObject(window, key);
+    }
+
+    if (!newValue) {
+      bindings.engine.logger.warn(
+        `${key} used in the href template is undefined for this result: ${result.title}`
+      );
+      return '';
```
ThibodeauJF, line 63: Except this early return with an empty string, this is taken entirely from the JSUI. Investigated other modern solutions and was getting stuff similar to that or eval,Function stuff which is way riskier

packages/atomic/src/utils/result-utils.ts
```diff
@@ -42,3 +43,36 @@ export function bindLogDocumentOpenOnResult(
...
+  return template.replace(/\${(.*?)}/g, (value: string) => {
+    const key = value.substring(2, value.length - 1);
+    let newValue = readFromObject(result, key);
+    if (!newValue) {
+      newValue = readFromObject(window, key);
+    }
+
+    if (!newValue) {
+      bindings.engine.logger.warn(
+        `${key} used in the href template is undefined for this result: ${result.title}`
```
liviarett, line 61: maybe add the result id here too just in case they have the same title?

packages/atomic/src/utils/result-utils.ts
```diff
@@ -42,3 +43,36 @@ export function bindLogDocumentOpenOnResult(
...
 }
+
+export function buildStringTemplateFromResult(
+  template: string,
+  result: Result,
+  bindings: Bindings
+) {
+  return template.replace(/\${(.*?)}/g, (value: string) => {
+    const key = value.substring(2, value.length - 1);
+    let newValue = readFromObject(result, key);
```
liviarett, line 54: we use lodash in this project, right? how about `_.get(result, key)` which then translates to, say, `get(result, 'raw.title')` instead of this new func?

olamothe: We use lodash only as a dev dependencies for tests IIRC

olamothe (approved):

packages/atomic/src/utils/result-utils.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    title: 'foo',
+    uri: 'http://uri.foo.com',
+    raw: {
+      ...TestUtils.buildMockResult().raw,
+      source: 'the source',
+    },
+  });
+
+  it('should create string templates', () => {
+    expect(buildStringTemplateFromResult('abc', result, bindings)).toBe('abc');
```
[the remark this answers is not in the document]
ThibodeauJF: XSS is handled in `LinkWithResultAnalytics`

ThibodeauJF:
For the doc, I have another PR coming for the same story & component. I can make any changes then!

PR #1920, opened 2022-04-13
feat(atomic-react): support react18 as peer dependency

olamothe:
https://coveord.atlassian.net/browse/KIT-1554

### DIFF
--- a/packages/atomic-react/package.json
+++ b/packages/atomic-react/package.json
@@ -31,11 +31,11 @@
     "@types/react": "17.0.37",
     "@types/react-dom": "17.0.11",
     "ncp": "2.0.0",
-    "react": "17.0.2",
-    "react-dom": "17.0.2"
+    "react": "18.0.0",
+    "react-dom": "18.0.0"
   },
   "peerDependencies": {
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react": "^17.0.2 || ^18.0.0",
+    "react-dom": "^17.0.2 || ^18.0.0"
   }
 }
--- a/packages/atomic-react/src/components/SearchInterfaceWrapper.tsx
+++ b/packages/atomic-react/src/components/SearchInterfaceWrapper.tsx
@@ -18,15 +18,10 @@ interface WrapperProps extends Omit<JSX.AtomicSearchInterface, 'i18n'> {
    */
   onReady?: (executeFirstSearch: ExecuteSearch) => Promise<void>;
   /**
-   * An optional `theme` property can be set in order to load one of the premade Coveo themes.
-   *
-   * Possible values are:
-   *
-   * - `coveo`: the default theme, used if no value is provided. It consists of a set of colors that match the Coveo brand.
-   * - `accessible`: a high contrast theme, best suited for implementations where web accessibility is important.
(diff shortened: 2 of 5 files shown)


packages/atomic-react/package.json
```diff
@@ -31,11 +31,11 @@
...
     "ncp": "2.0.0",
-    "react": "17.0.2",
-    "react-dom": "17.0.2"
+    "react": "18.0.0",
+    "react-dom": "18.0.0"
   },
   "peerDependencies": {
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react": "^17.0.2 || ^18.0.0",
```
olamothe, line 38: That's an important change, means it no longer complains when you `npm i @coveo/atomic-react` with a recent React project.

Otherwise you had to `--force` install.

packages/atomic-react/src/components/SearchInterfaceWrapper.tsx
```diff
@@ -62,21 +57,21 @@ export const SearchInterfaceWrapper = (
...
   useEffect(() => {
     const searchInterfaceAtomic = searchInterfaceRef.current!;
-
-    searchInterfaceAtomic.initializeWithSearchEngine(engine).then(() => {
-      localization(searchInterfaceAtomic.i18n);
-      onReady(
-        searchInterfaceAtomic.executeFirstSearch.bind(searchInterfaceAtomic)
-      );
-    });
+    if (!initialization) {
```
olamothe, line 66: In dev mode, React will often re-run hook multiple times to see if there's side effect.

We only want to initialize the interface once: It does not crash, it's handled well by Atomic, but Atomic will (correctly) print warning in the console when that happens.

packages/atomic-react/src/components/SearchInterfaceWrapper.tsx
```diff
@@ -18,15 +18,10 @@ interface WrapperProps extends Omit<JSX.AtomicSearchInterface, 'i18n'> {
...
   onReady?: (executeFirstSearch: ExecuteSearch) => Promise<void>;
   /**
-   * An optional `theme` property can be set in order to load one of the premade Coveo themes.
-   *
-   * Possible values are:
-   *
-   * - `coveo`: the default theme, used if no value is provided. It consists of a set of colors that match the Coveo brand.
-   * - `accessible`: a high contrast theme, best suited for implementations where web accessibility is important.
-   * - `none`: no premade theme will be loaded. You will have to provide the theme yourself.
+   * @deprecated
```
olamothe, line 21: That `theme` props was a mistake. It complicates things with many build tool when we try to do dynamic import.

ThibodeauJF: We should explain why or offer an alternative after the `deprecated` if possible

ThibodeauJF (approved):

packages/atomic-react/src/components/SearchInterfaceWrapper.tsx
olamothe: Yeah, I think the link to the doc that explains how to do theming + the existing doc for Atomic-react should do.

PR #1922, opened 2022-04-14
feat(atomic): add "attributes" slot for the atomic-result-link to allow copying attributes

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1558

What do you think about this pattern? Let me know and I'll add cypress tests!


btaillon:
Which attributes do we generally want to customize?

packages/atomic/src/components/result-template-components/atomic-result-link/atomic-result-link.tsx
```diff
@@ -56,16 +59,48 @@ export class AtomicResultLink implements InitializableComponent {
...
+  private linkAttributes?: Attr[];
 
   public initialize() {
     this.interactiveResult = buildInteractiveResult(this.bindings.engine, {
       options: {result: this.result},
     });
   }
 
+  private isAttributeSlot(element: Element): element is HTMLAnchorElement {
+    return element.slot === 'attributes' && element.nodeName === 'A';
```
olamothe, line 72: Should there be some warning if someone is trying to use a `slot="attributes"` with something other than an anchor ?

olamothe (approved):
I think it's a neat pattern !

However, one thing we'll need to make sure of eventually is that every component that displays a link to a result (ie: smart snippets, quickviews, printable uri, etc.) should all use the same mechanism to allow to customize the link.

All of them should expose `href-template`, `slot="attributes"`, `field` etc.

So, some kind of inheritance or composition on anything that can be considered a "result link" component.

VS here, where it's localized and only usable for one component.

packages/atomic/src/components/result-template-components/atomic-result-link/atomic-result-link.tsx
```diff
@@ -56,16 +59,48 @@ export class AtomicResultLink implements InitializableComponent {
   @Prop({reflect: true}) hrefTemplate?: string;
```
ThibodeauJF, line 59: Docs: could I please get a review on this one as well please? Thanks!

packages/atomic/src/components/result-template-components/atomic-result-link/atomic-result-link.tsx
```diff
@@ -17,6 +17,7 @@ import {buildStringTemplateFromResult} from '../../../utils/result-utils';
 /**
  * The `atomic-result-link` component automatically transforms a search result title into a clickable link that points to the original item.
  * @slot default - Allow to display alternative content inside the link
+ * @slot attributes - Allows to pass [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) down to the link element, overriding other attributes, to be used exclusively with an "a" tag such as `<a slot="attributes" target="_blank" download></a>`.
```
jpmarceau, line 20:
```suggestion
 * @slot default - Lets you display alternative content inside the link
 * @slot attributes - Lets you pass [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) down to the link element, overriding other attributes, to be used exclusively with an "a" tag such as `<a slot="attributes" target="_blank" download></a>`.
```

jpmarceau (approved):
👍

packages/atomic/src/utils/slot-utils.tsx
```diff
@@ -1,54 +0,0 @@
-import {ComponentInterface} from '@stencil/core';
```
ThibodeauJF, line 1: unused dead code

olamothe (approved):

packages/atomic/src/utils/slot-utils.ts
```diff
@@ -0,0 +1,23 @@
+export function getSlotFromHost(host: HTMLElement, slotName: string) {
+  const children = Array.from(host.children);
+  const slots = children.filter((child) => {
+    if (slotName) {
+      return child.getAttribute('slot') === slotName;
+    }
+
+    return child.setAttribute('slot', '');
```
btaillon, line 8: Why would `slotName` be falsy/empty? I would avoid setting an attribute in a `get*` function.

ThibodeauJF: 🤔

ThibodeauJF: me neither! 🤣

PR #1932, opened 2022-04-19
docs(atomic): add header navigation to sample page

liviarett:
https://coveord.atlassian.net/browse/KIT-1571

Screenshot 2022-04-19 at 14 19 35
Screenshot 2022-04-19 at 14 19 51

### DIFF
--- a/packages/atomic/src/pages/header.js
+++ b/packages/atomic/src/pages/header.js
@@ -23,11 +23,11 @@ const makeLinks = () => {
     )
     .join('');
 };
-
+const example = getCurrentExample();
 header.innerHTML = `
   <nav style="padding: 10px 20px; font-family: var(--atomic-font-family);">
     <span style="font-weight: var(--atomic-font-bold);">${
-      getCurrentExample().label
+      example ? example.label : ''
     } example</span>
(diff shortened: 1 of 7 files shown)


packages/atomic/src/pages/header.js
```diff
@@ -0,0 +1,36 @@
...
+
+const getCurrentExample = () =>
+  links.find((link) => link.href === window.location.pathname);
+
+const makeLinks = () => {
+  return links
+    .filter((link) => link.href !== window.location.pathname)
+    .map(
+      (link) =>
+        `<a href="${link.href}" style="margin-right: 10px; color: var(--atomic-primary); text-decoration: none;">${link.label}</a>`
```
btaillon, line 20: I would wrap the `a` element in a `li` (even though accessibility doesn't matter that much here)

btaillon (approved):
neat!

packages/atomic/src/pages/header.js
liviarett: ah yeah, that's what I meant to do! thanks for catching it

ThibodeauJF (approved):
Hey that's pretty cool! Thank you for that 😄

PR #1933, opened 2022-04-19
feat(headless): add methods for finding results by id and by collection

liviarett:
https://coveord.atlassian.net/browse/KIT-1572

### DIFF
--- a/packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
+++ b/packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
@@ -2,6 +2,7 @@ import {
   FoldedResultList,
   FoldedResultListOptions,
   buildFoldedResultList,
+  FoldedCollection,
 } from './headless-folded-result-list';
 import {
   buildMockSearchAppEngine,
@@ -180,11 +181,14 @@ describe('FoldedResultList', () => {
   });
 });
 
-function makeFoldedResult(result: Result, children: Result[] = []) {
+function makeFoldedResult(
+  result: Partial<Result>,
+  children: FoldedCollection[] = []
+) {
   return {
     isLoadingMoreResults: true,
     moreResultsAvailable: true,
     result: result as Result,
-    children,
-  };
+    children: children as FoldedCollection[],
+  } as FoldedCollection;
 }
--- a/packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
+++ b/packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
@@ -180,7 +180,7 @@ describe('FoldedResultList', () => {
   });
 });
 
-function makeFoldedResult(result: any, children: any[] = []) {
+function makeFoldedResult(result: Result, children: Result[] = []) {
   return {
     isLoadingMoreResults: true,
     moreResultsAvailable: true,
--- a/packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
+++ b/packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
@@ -51,10 +51,12 @@ describe('FoldedResultList', () => {
   describe('with a result and two collections', () => {
     beforeEach(() => {
       const threadResult = buildMockResult({
+        uniqueId: 'thread-result',
         title: 'thread-result',
         raw: {urihash: '', foldingcollection: 'thread'},
       });
       const peopleResult = buildMockResult({
+        uniqueId: 'people-result',
         title: 'people-result',
         raw: {urihash: '', foldingcollection: 'people'},
       });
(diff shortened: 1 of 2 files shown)


packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -160,6 +172,16 @@ export function buildFoldedResultList(
...
 
+    findResultByUniqueID(id) {
+      return this.state.results.find((r) => r.result.uniqueId === id);
+    },
+
+    findResultByCollection(collection) {
+      return this.state.results.find(
+        (r) => r.result.raw.foldingcollection === collection
+      );
+    },
```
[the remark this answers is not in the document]
liviarett: @olamothe suggested we add these, the idea was to pass a result instead of the strings initially but I thought it made more sense to pass the id/collection.
what do you think, Olivier?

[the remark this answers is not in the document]
olamothe: Passing in the id or collection as a string does not seem really useful

As an end user of the lib, with no extensive knowledge of Coveo, how would you know which string property to use ? There's many of them available on each individual result.

You would have to read the doc, which is not that awesome. It puts burden on the user.

Where it starts to show value is if it accept a generic object (the whole result), and then internally know which property of that result it should use out of those available. 

That way, using TypeScript, you also get re-assurance that you are doing "the right thing", VS a string type that is non specific, so you are never sure which property you should pass in.

It should also recursively search through all descendants, as Benjamin said.

With all that, it starts being pretty useful.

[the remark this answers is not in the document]
liviarett: ok I see, that all makes sense!

liviarett:
@olamothe @btaillon let me know what you think! if you think it won't be useful I can close!

ThibodeauJF (approved):
👍  The branch name length is scaring me a little bit haha!

olamothe (approved):

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -86,6 +86,20 @@ export interface FoldedResultList extends Controller {
    * @param collection - The collection for which to load more results.
    */
   loadCollection(collection: FoldedCollection): void;
+  /**
+   * Finds a folded result by its unique id.
```
npushkarskii, line 90:
```suggestion
   * Finds a folded result by its unique ID.
```

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -86,6 +86,20 @@ export interface FoldedResultList extends Controller {
...
+   *
+   * @param collection - The folded collection whose id will be used to find a collection in the results.
```
npushkarskii, line 92:
```suggestion
   * @param collection - The folded collection whose ID will be used to find a collection in the results.
```

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -86,6 +86,20 @@ export interface FoldedResultList extends Controller {
...
+   */
+  findResultById(collection: FoldedCollection): FoldedCollection | undefined;
+  /**
+   * Finds a folded result by its collection.
+   *
+   * @param collection - The folded collection whose collection name will be used to find a collection in the results
```
npushkarskii, line 98:
```suggestion
   * @param collection - The folded collection whose collection name will be used to find a collection in the results.
```

npushkarskii (approved):
LGTM! 🎉

I only left a pinch of minor suggestions 🙏

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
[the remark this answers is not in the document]
liviarett: interesting! the command is `npm run doc:generate`

liviarett:
@btaillon would you mind having a final look at the recursion so I can merge?

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
```diff
@@ -106,5 +110,81 @@ describe('FoldedResultList', () => {
...
+        [expected]
+      );
+      foldedResultList.state.results[2].children.push(firstChild);
+      const result = foldedResultList.findResultByCollection(expected);
+      expect(result).toEqual(expected);
+    });
   });
 });
+
+function makeFoldedResult(result: any, children: any[] = []) {
```
btaillon, line 183: I would avoid `any`.

```suggestion
function makeFoldedResult(result: Result, children: Result[] = []) {
```

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -192,3 +222,22 @@ function loadFoldingReducer(
   engine.addReducers({search, configuration, folding, query});
   return true;
 }
+
+function searchForResult(
+  results: FoldedCollection[] | FoldedResult[],
+  compareCb: (result: FoldedResult) => Boolean
```
btaillon, line 228: typically `boolean` is lowercase in Typescript:
```suggestion
  compareCb: (result: FoldedResult) => boolean
```

btaillon (approved):
looks good!

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.test.ts
liviarett: oopsie

liviarett: I thought we allowed it in tests! (not that anyone told me we did, I just assumed)

PR #1936, opened 2022-04-20
fix(headless): set root result as main result instead of collection p…

liviarett:
…arent

I still don't understand why in some cases, like the video below, the result that is returned from the collections call seems to have no children

https://user-images.githubusercontent.com/30511433/164240170-14ab9761-7f05-4e7a-9cf3-0680691fbeab.mov

### DIFF
--- a/packages/headless/src/features/folding/folding-actions-loader.ts
+++ b/packages/headless/src/features/folding/folding-actions-loader.ts
@@ -9,6 +9,7 @@ import {
   LoadCollectionFulfilledReturn,
   StateNeededByLoadCollection,
 } from './folding-actions';
+import {ResultWithFolding} from './folding-slice';
 
 export type {RegisterFoldingActionCreatorPayload};
 
@@ -33,10 +34,10 @@ export interface FoldingActionCreators {
    * @returns A dispatchable action.
    */
   loadCollection(
-    collectionId: string
+    result: ResultWithFolding
   ): AsyncThunkAction<
     LoadCollectionFulfilledReturn,
-    string,
+    ResultWithFolding,
     AsyncThunkSearchOptions<StateNeededByLoadCollection>
   >;
 }
--- a/packages/headless/src/features/folding/folding-actions-loader.ts
+++ b/packages/headless/src/features/folding/folding-actions-loader.ts
@@ -9,7 +9,6 @@ import {
   LoadCollectionFulfilledReturn,
   StateNeededByLoadCollection,
 } from './folding-actions';
-import {ResultWithFolding} from './folding-slice';
 
 export type {RegisterFoldingActionCreatorPayload};
 
@@ -34,10 +33,10 @@ export interface FoldingActionCreators {
    * @returns A dispatchable action.
    */
   loadCollection(
-    result: ResultWithFolding
+    collectionId: string
   ): AsyncThunkAction<
     LoadCollectionFulfilledReturn,
-    ResultWithFolding,
+    string,
     AsyncThunkSearchOptions<StateNeededByLoadCollection>
   >;
 }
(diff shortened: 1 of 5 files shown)


packages/headless/src/features/folding/folding-actions-loader.ts
```diff
@@ -33,10 +34,10 @@ export interface FoldingActionCreators {
    * @returns A dispatchable action.
    */
   loadCollection(
-    collectionId: string
+    result: ResultWithFolding
```
olamothe, line 37: This is a breaking change on the public API: a publicly exported function from the package which used to accept a `string` as a parameter now accept a `ResultWithFolding`. 

User code importing and trying to use this function, on upgrade, would either no longer compile (if they are using TypeScript), or would crash at runtime if using JavaScript.

We unfortunately cannot do this.

We can either:
* Add a new function with a different name (and potentially mark the old one as `@deprecated`, but still functional)
* expand the accepted parameter type to be `string | ResultWithFolding` and have logic to detect which it is
* Or see if there's another solution that does not involve modifying the public API

liviarett: hmm I was thinking this would then be a breaking change that would bump the major version...? but I guess that's not how it works!
@btaillon since you understand a bit more what's going on with these results, I'll let you give your opinion on whether this is the way to go before I make any changes!

ThibodeauJF: > hmm I was thinking this would then be a breaking change that would bump the major version

Yes that's what we do for breaking changes, but we don't bump major versions often, because it's annoying to the lib users to upgrade and it's hard for us to maintain. We are planning a major bump this summer I think? There will be a laundry list of improvements to create and we can do them all at once when going to v2!

packages/headless/src/features/folding/folding-slice.ts
```diff
@@ -86,28 +86,31 @@ function resolveRootFromFields(
...
+  rootResult?: ResultWithFolding
 ): FoldedCollection {
   const resultsInCollection = getAllIncludedResultsFrom(relevantResult);
 
-  const rootResult =
+  const resolvedRoot =
     resolveRootFromFields(resultsInCollection, fields) ?? relevantResult;
 
+  const children = resolveChildrenFromFields(
+    resolvedRoot,
```
btaillon, line 98: `resolvedRoot` here should be replaced with `rootResult ?? resolvedRoot`

packages/headless/src/features/folding/folding-actions.ts
```diff
@@ -67,18 +69,20 @@ export type StateNeededByLoadCollection = ConfigurationSection &
...
 export const loadCollection = createAsyncThunk<
   LoadCollectionFulfilledReturn,
-  CollectionId,
+  ResultWithFolding,
   AsyncThunkSearchOptions<StateNeededByLoadCollection>
 >(
   'folding/loadCollection',
   async (
-    collectionId: CollectionId,
+    result: ResultWithFolding,
```
btaillon, line 77: Instead of taking a `result` as a parameter, I think resolving the result as the following might work:
```
const result = state.folding.collections[collectionId]!.result;
```

packages/headless/src/features/folding/folding-slice.ts
```diff
@@ -86,28 +86,31 @@ function resolveRootFromFields(
...
   relevantResult: ResultWithFolding,
-  fields: FoldingFields
+  fields: FoldingFields,
+  rootResult?: ResultWithFolding
 ): FoldedCollection {
   const resultsInCollection = getAllIncludedResultsFrom(relevantResult);
 
-  const rootResult =
+  const resolvedRoot =
     resolveRootFromFields(resultsInCollection, fields) ?? relevantResult;
```
btaillon, line 95: I just realized now that it's not necessary to call `resolveRootFromFields` when `rootResult` is specified.

All in all, the code could look something like this:
```ts
const resultsInCollection = getAllIncludedResultsFrom(relevantResult);

const resultToUseAsRoot = rootResult
  ?? resolveRootFromFields(resultsInCollection, fields)
  ?? relevantResult;

return {
  result: resultToUseAsRoot,
  children: resolveChildrenFromFields(
    resultToUseAsRoot,
    resultsInCollection,
    fields
  ),
  moreResultsAvailable: true,
  isLoadingMoreResults: false
};
```

btaillon (approved):

PR #1938, opened 2022-04-21
fix(headless): rework pagination logic to reuse existing searchResponseId

olamothe:
Problem was:

1. Execute a search, get search response ID, log an analytics search event.
2. Click on a result, analytics click event is linked to search response ID generated at step # 1, everything is good so far.
3. Use the pager to go to next page, execute a new query, get a new search response ID, update the results set/response/search ID. Log a custom event (not a search, since from the end user perspective, it's not a new search, it's just a continuation of the same query performed on step # 1)
4. Click on a new result, analytics click event is linked to search response ID generated at step # 3. Since at step # 3 we never log a search event, but a custom event, this means that click is flagged as incoherent by Usage Analytics, since there is no accompanying search event.

Now, this problem was actually also present for everything that loaded more result: "Load more" button, infinite scrolling, pagination, and facet load more values because in `AnalyticsClient` we were not using the correct `searchResponseId` property when available.

For load more/infinite scrolling/facet fetch more, there was nothing more to fix than using the correct property in `AnalyticsClient`.

For pagination, we used the standard `executeSearch` action, which means we had no way of differentiating between a "new search" and "fetching a new page" from an end user perspective.

For this, I added a `fetchPage` action, which does the same thing as `executeSearch`, but without the code needed to handle did you mean feature.

Hope that makes sense !

https://coveord.atlassian.net/browse/KIT-1573

### DIFF
--- a/packages/headless/src/api/analytics/analytics.test.ts
+++ b/packages/headless/src/api/analytics/analytics.test.ts
@@ -138,7 +138,22 @@ describe('analytics', () => {
       );
     });
 
-    it('when search is provided, #getSearchUID returns the correct id', () => {
+    it('when #searchResponseId is provided, #getSearchUID returns the correct id', () => {
+      const searchUid = 'test';
+      const state = {
+        ...baseState,
+        search: buildMockSearchState({
+          searchResponseId: searchUid,
+          response: buildMockSearchResponse({
+            searchUid: 'another_id',
+          }),
+        }),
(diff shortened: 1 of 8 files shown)


liviarett (approved):
I think it makes sense!

btaillon (approved):

ThibodeauJF (approved):

PR #1939, opened 2022-04-21
feat(atomic): add styling and states for children results and loading

liviarett:
Couple pics:
Screenshot 2022-04-21 at 17 22 07
Screenshot 2022-04-21 at 17 22 36
Screenshot 2022-04-21 at 17 23 12

Vid: (I changed the copy from show/hide all to collapse/expand, but it needed a server restart)
https://user-images.githubusercontent.com/30511433/164505605-ab8fb516-d22a-45f7-b626-1cc0efea45cc.mov

### DIFF
--- a/packages/atomic/src/components/atomic-result-placeholder/atomic-result-placeholder.tsx
+++ b/packages/atomic/src/components/atomic-result-placeholder/atomic-result-placeholder.tsx
@@ -21,9 +21,14 @@ export class AtomicResultPlaceholder {
   @Prop() display!: ResultDisplayLayout;
   @Prop() density!: ResultDisplayDensity;
   @Prop() imageSize?: ResultDisplayImageSize;
+  @Prop() classes = '';
 
   private getClasses() {
-    return getResultDisplayClasses(this.display, this.density, this.imageSize!);
+    return getResultDisplayClasses(
+      this.display,
+      this.density,
+      this.imageSize!
+    ).concat(this.classes);
   }
 
   private renderExcerptLine(width: string) {
@@ -45,9 +50,9 @@ export class AtomicResultPlaceholder {
   public render() {
     return (
       <div
-        class={`result-root with-sections animate-pulse ${this.getClasses().join(
-          ' '
-        )}`}
+        class={`result-root with-sections animate-pulse ${this.getClasses()
+          .join(' ')
+          .trim()}`}
       >
         <atomic-result-section-visual>
           <div class={placeholderClasses}></div>
--- a/packages/atomic/src/components/atomic-result-placeholder/atomic-result-placeholder.tsx
+++ b/packages/atomic/src/components/atomic-result-placeholder/atomic-result-placeholder.tsx
@@ -21,14 +21,14 @@ export class AtomicResultPlaceholder {
   @Prop() display!: ResultDisplayLayout;
   @Prop() density!: ResultDisplayDensity;
   @Prop() imageSize?: ResultDisplayImageSize;
-  @Prop() classes = '';
+  @Prop() isChild = false;
 
   private getClasses() {
     return getResultDisplayClasses(
       this.display,
       this.density,
       this.imageSize!
-    ).concat(this.classes);
+    ).concat(this.isChild ? 'child-result' : '');
   }
 
   private renderExcerptLine(width: string) {
--- a/packages/atomic/src/components/atomic-result/atomic-result-row-desktop.pcss
+++ b/packages/atomic/src/components/atomic-result/atomic-result-row-desktop.pcss
@@ -1,7 +1,36 @@
+@define-mixin atomic-children-desktop $max, $min {
+  &.child-result {
+    margin-bottom: $max;
+  }
+
+  atomic-result-section-children {
+    atomic-result-children::part(children-root) {
+      margin-top: $min;
+      padding: $max;
+    }
+
+    atomic-result-children::part(show-hide-button) {
+      margin-top: $min;
+      margin-bottom: $min;
+    }
+  }
+
+  atomic-result-placeholder {
+    margin-bottom: $max;
+  }
+}
+
 @define-mixin row-result-desktop {
   &.image-large.density-compact atomic-result-section-visual {
     width: 10.25rem;
     height: 10.25rem;
+  }
+
+  atomic-result-section-children {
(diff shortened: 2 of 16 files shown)


packages/atomic/src/components/atomic-result/atomic-result-row-desktop.pcss
```diff
@@ -1,9 +1,38 @@
+@define-mixin atomic-children-desktop $max, $min {
```
btaillon, line 1: I would rename $max and $min to something like $margin-small and $margin-large

packages/atomic/src/components/atomic-result-placeholder/atomic-result-placeholder.tsx
```diff
@@ -21,9 +21,14 @@ export class AtomicResultPlaceholder {
   @Prop() display!: ResultDisplayLayout;
   @Prop() density!: ResultDisplayDensity;
   @Prop() imageSize?: ResultDisplayImageSize;
+  @Prop() classes = '';
```
btaillon, line 24: Instead of having both display/density/imageSize and `classes`, I would prefer to only have one of the two patterns. `classes` could be replaced with `isChild?: boolean` or `display/density/imageSize` could be removed and `classes` could be used instead

packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
```diff
@@ -140,31 +162,91 @@ export class AtomicResultChildren {
...
-const Loading: FunctionalComponent = () => {
-  // TODO:
-  return <p>Loading...</p>;
-};
+  public render() {
+    if (!this.ready) return null;
+    if (this.templateHasError) return <slot></slot>;
+    return (
+      <Host>
+        {this.showCollapseButton() && (
```
btaillon, line 234: Instead of having extracting the condition and rendering directly in `render`, I would prefer to extract the whole thing and follow a pattern like this:
```tsx
private renderCollapseButton() {
  if (!someConditions) {
    return;
  }
  return <Button ...>...</Button>;
}

public render() {
  ...
  return (
    <Host>
      {this.renderCollapseButton()}
    </Host>
  );
}
```

packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
```diff
@@ -140,31 +162,91 @@ export class AtomicResultChildren {
...
-        </Host>
-      );
+    return (
+      Boolean(result.children.length) &&
+      !result.isLoadingMoreResults &&
+      !result.moreResultsAvailable
+    );
+  }
+
+  private showChildrenWrapper() {
```
btaillon, line 174: The name of this function makes me instinctively think it's an action rather than a condition/getter.

I would either make this a getter (`private get showChildrenWrapper()`) or prefix it with a verb like `should`

btaillon (approved):
very nice!

packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
liviarett: okie

packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
liviarett: yeah, I agree. could you explain why making it a getter would make it better? and also in general when it is that you decide to use getters?

packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
```diff
@@ -140,31 +162,94 @@ export class AtomicResultChildren {
...
+  }
+
+  private getComponents() {
+    const components = [];
+    const result = this.getResult();
+    const hasResults = Boolean(result.children.length);
+    if (hasResults) {
+      components.push(<slot name="before-children"></slot>);
+    }
+    if (result.isLoadingMoreResults) {
```
olamothe, line 191: I think this might be easier to read and reason about if it was structured into different scenarios with dedicated functions to handle them, with all the boolean check aligned in a "flat" manner.

For example:
```typescript
[... etc ...]
if(result.isLoadingMoreResult()) {
   return this.getComponentsForLoadingState()
}

if(this.hasResults) {
   return this.getComponentsForHasResultsState()
}

if(this.showAllResults) {
   return this.getComponentsForNoResultsState()
}
[... etc ...]


private get hasResults() {
  return this.getResult().children.length !== 0;
}

private getComponentsForLoadingState() {
   return [ 
           this.hasResult ? <slot name="before-children"></slot> : null, 
          <ListDisplayResultsPlaceholder
              resultsPerPageState={{
                numberOfResults: this.getResult().children.length || 2,
              }}
              density={this.displayConfig.density}
              imageSize={this.imageSize || this.displayConfig.imageSize}
              isChild
          />
          this.hasResult ? , <slot name="after-children"></slot>) : null
].filter(component=> component !== null)

}

[... etc..]
```

I think this would help to see each state as individual switch cases that needs to independently render different components, VS all intermixed.

olamothe (approved):

packages/atomic/src/components/atomic-result/atomic-result.tsx
```diff
@@ -68,13 +71,25 @@ export class AtomicResult {
...
   @Listen('atomic/resolveResult')
   public resolveResult(event: ResultContextEvent<FoldedResult | Result>) {
     event.preventDefault();
     event.stopPropagation();
     event.detail(this.result);
   }
 
+  @Listen('atomic/resolveResultDisplayConfig')
+  public resolveResultDisplayConfig(event: ResultContextEvent<DisplayConfig>) {
+    event.preventDefault();
```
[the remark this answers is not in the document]
liviarett: I think it's better to leave it bc it returns the closest parent - if we let it bubble up it could get the data from the grandparent atomic-result (which is not the end of the world, but I think this makes more sense)

packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
liviarett: I agree, what I did isn't great... I'm updating!

PR #1940, opened 2022-04-21
fix(quantic): fixed double search on timeframe facet clear

nathanlb:
Clearing the filters on the Quantic Timeframe Facet would trigger two search requests due to the filter.clear action dispatching a search as well as the facet.deselectAll.
Now we only execute one or the other depending on if the filter is set using the filter or a selected timeframe.

[SVCC-550](https://coveord.atlassian.net/browse/SVCC-550)


lbergeron (approved):

mmitiche (approved):
👌

PR #1941, opened 2022-04-21
fix(quantic): added optional chaining to searchbox

nathanlb:
Clicking the search boxes before the engine is initialized causes an error that breaks the components.
Simple optional chaining fixes it.


mmitiche (approved):

aelfaleh:
Can we test this with cypress?

PR #1942, opened 2022-04-21
feat(quantic): new Quantic Modal component created

mmitiche:
### Quantic Modal
The QuanticModal component provides the logic to easily implement a modal in Quantic. This component handles the animation logic, exposes methods to open and close the modal, offers the option to open the modal in full screen or just to cover the search interface, and exposes a set of slots to fully customize the modal content.

[SFINT-4493](https://coveord.atlassian.net/browse/SFINT-4493)

### Note:
The full screen option is still present in the quantic modal api for the following reason:

- Having the Quantic Modal always covering the search interface means that the height of the modal will always be equal to the height of the search interface. Which is a bit problematic because for example in the Example Search Page the height of the search interface  it pretty big, so if the modal covers the whole search interface the user will need to scroll down to see the footer of the modal which is present in the bottom of the modal.  For this particular reason having the modal on full screen can be more useful and more natural. up to debate of course!

### Demo

https://user-images.githubusercontent.com/86681870/164706654-2d92b614-0f15-4396-b2fe-7c446847ad04.mov

https://user-images.githubusercontent.com/86681870/164709216-09022c09-7f09-4b46-be5e-7beeead8a1fe.mp4


lbergeron:
@MehdiMitiche In the Full Screen section of the PR description you say this:
> Having the Quantic Modal always covering the search interface means that the height of the modal will always be equal to the height of the search interface.

Is it possible to make the height of the modal match the viewport's height? Or is that what the "full screen" option is doing?
This might be tricky to implement though, especially when other components are rendered on top of the Insight Panel.

mmitiche:
> @MehdiMitiche In the Full Screen section of the PR description you say this:
> 
> > Having the Quantic Modal always covering the search interface means that the height of the modal will always be equal to the height of the search interface.
> 
> Is it possible to make the height of the modal match the viewport's height? Or is that what the "full screen" option is doing? This might be tricky to implement though, especially when other components are rendered on top of the Insight Panel.

Yes that's what the "full screen" option is doing, giving the modal a height equal to the viewport's height and a width equal to the viewport's width with a big value of z-index to make the modal appear above of all the other components.

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.css
```diff
@@ -0,0 +1,60 @@
+.modal {
+  background-color: white;
```
lbergeron, line 2: If possible, I think we should use the `$color-background-alt` design token.
See [Design Tokens](https://www.lightningdesignsystem.com/design-tokens/#category-background-color)

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.css
```diff
@@ -0,0 +1,60 @@
...
+  clip-path: inset(100% 0 0 0);
+}
+
+.hidden-modal_slide-to-left {
+  -webkit-clip-path: inset(0 0 0 100%);
+  clip-path: inset(0 0 0 100%);
+}
+
+.modal__header {
+  background-color: #F3F3F3;
```
lbergeron, line 29: We should use the `$color-background` design token.

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.css
```diff
@@ -0,0 +1,60 @@
...
+  overflow-y: hidden;
+}
+
+.modal__content {
+  overflow-y: auto;
+  flex: 1;
+}
+
+.modal__footer {
+  background-color: #F3F3F3;
```
lbergeron, line 45: We should use the `$color-background` design token.

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -0,0 +1,60 @@
...
+   * @type {'slideToTop'|'slideToLeft'}
+   * @defaultValue `slideToTop`
+   */
+  @api animation = 'slideToTop';
+
+  /** @type {boolean} */
+  visible = false;
+
+  /**
+   * Set the visible attribute to true to open the modal.
```
lbergeron, line 30: This comment should explain the "what" more than the "how".

```suggestion
   * Opens the modal container.
```

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -0,0 +1,60 @@
...
+   * @returns {void}
+   */
+  @api openModal() {
+    this.visible = true;
+  }
+
+  /**
+   * Set the visible attribute to false to close the modal.
```
lbergeron, line 38:
```suggestion
   * Closes the modal container.
```

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -0,0 +1,60 @@
...
+   * @returns {void}
+   */
+  @api closeModal() {
+    this.visible = false;
+  }
+
+  /**
+   * Returns the CSS class of the modal.
```
lbergeron, line 46:
```suggestion
   * Returns the modal container CSS classes.
```

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -0,0 +1,60 @@
+import {LightningElement, api} from 'lwc';
+
+/**
+ * The `QuanticModal` component provides the logic to easily implement a modal in Quantic. This component handles the animation logic, exposes methods to open and close the modal, offers the option to open the modal in full screen or just to cover the search interface, and exposes a set of slots to fully customize the modal content.
```
lbergeron, line 4: I see "modal" is used as a noun in this file. But since it is an adjective, it should be followed by a noun (e.g., modal container, modal box, modal window).

I personally like `modal container`.

@jpmarceau Do you think of a better term?

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -0,0 +1,60 @@
...
+    return `modal ${this.fullScreen ? 'full-screen' : 'part-screen'} ${
+      this.visible
+        ? ''
+        : `modal_hidden ${
+            this.animation === 'slideToLeft'
+              ? 'hidden-modal_slide-to-left'
+              : 'hidden-modal_slide-to-top'
+          }`
+    }`;
+  }
```
lbergeron, line 59: In my opinion, this string interpolation expression is too complex and should be split.

```suggestion
  get modalCssClass() {
    const displayAreaClass = this.fullScreen ? 'full-screen' : 'part-screen';
    const animationClass = this.animation === 'slideToLeft' ? 'hidden-modal_slide-to-left' : 'hidden-modal_slide-to-top';
    const visibilityClass = this.visible ? '' : `modal_hidden ${animationClass}`;
  
    return `modal ${displayAreaClass} ${visibilityClass}`;
  }
```

lbergeron (commented):
It looks pretty good! Well done!

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
jpmarceau: Actually in the context of front-end development, it's OK to use 'modal' as a noun too :) 
Examples:
https://blog.hubspot.com/website/modal-web-design
https://www.w3schools.com/w3css/w3css_modal.asp

nathanlb:
> 

In the case of the refine modal with facets as content. The content will often be longer than the container and the button in the footer to go back to the search results should always be visible and clickable.

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Indicates the type of animation to use to open the modal.
+   * @api
+   * @type {'slideToTop'|'slideToLeft'}
+   * @defaultValue `slideToTop`
+   */
+  @api animation = 'slideToTop';
+
+  /** @type {boolean} */
+  visible = false;
```
nathanlb, line 27: For boolean variables `isVisible` is a good pattern generally. It's a style choice though

nathanlb:
What does it take to achieve the responsiveness you show in the RefineModalExampleSearch.online-video-cutter.com.mp4 ? If it can be done in a slick way using template conditional markup for example then could we add it to the exampleSearch component while we're at it?

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
lbergeron: Ah! Good to know. Thank you!

lbergeron:
> > > > @MehdiMitiche In the Full Screen section of the PR description you say this:
> > > > > Having the Quantic Modal always covering the search interface means that the height of the modal will always be equal to the height of the search interface.
> > > > 
> > > > 
> > > > Is it possible to make the height of the modal match the viewport's height? Or is that what the "full screen" option is doing? This might be tricky to implement though, especially when other components are rendered on top of the Insight Panel.
> > > 
> > > 
> > > Yes that's what the "full screen" option is doing, giving the modal a height equal to the viewport's height and a width equal to the viewport's width with a big value of z-index to make the modal appear above of all the other components.
> > 
> > 
> > The thing that worries me is having to scroll down to reach the modal footer when the search page is longer than the viewport. I think that when `full-screen` is `false`, the modal height should still match the viewport height so the footer is always visible. We'd have to decide what to do when the modal content is longer than the viewport. My guess is that the footer should remain visible and the content section should scroll, but pushing the footer down after the content could also make sense.
> 
> I'm a bit confused, why would we want to the modal height to match the viewport when 'full-screen' is set to false? For example in the Insight Panel shown below: InsightPAnel
> 
> It's the case where we need the modal to only cover the search interface so 'full-screen' is set to false here. Why would we want the modal height to match the viewport height here?

From what I see in the screenshot, it already behaves as I would expect. I mean that the footer is visible and the content is scrolling, so it seems to be all good already.

mmitiche:
> > The thing that worries me is having to scroll down to reach the modal footer when the search page is longer than the viewport. I think that when `full-screen` is `false`, the modal height should still match the viewport height so the footer is always visible. We'd have to decide what to do when the modal content is longer than the viewport. My guess is that the footer should remain visible and the content section should scroll, but pushing the footer down after the content could also make sense.
>  I understand that we don't want to make the user scroll down to reach the modal footer,  but I can't think of a way that will allow us to always show the  modal footer and in the same time have the modal covering only the search interface when the height of the search interface is longer than the viewport.
The logic I followed is like the following:

- If the height of your search interface is greater than the viewport's height, set the value of 'full-screen' to true to display the modal in full screen and avoid scrolling the page to reach the footer of the modal.
- if the height of your search interface is lower than the viewport's height, you can set 'full-screen' to true or false depending on the use case. In both cases, you will not need to scroll to reach the footer of the modal.

mmitiche:
> What does it take to achieve the responsiveness you show in the RefineModalExampleSearch.online-video-cutter.com.mp4 ? If it can be done in a slick way using template conditional markup for example then could we add it to the exampleSearch component while we're at it?

Yes this can be done in a slick way👌 We just need the components needed to achieve this which are the QuanticRefineToggle and the QuanticRefineModal. I'll open PRs for these two components after merging this one.
Then I'll do a separate  PR to update the responsiveness of the exampleSearch component.

packages/quantic/cypress/integration/modal/modal.cypress.ts
```diff
@@ -0,0 +1,99 @@
...
+        Expect.openModal(true);
+        Expect.displayHeader(true);
+        Expect.displayContent(true);
+        Expect.displayFooter(true);
+        Expect.displayFullScreen(false);
+      });
+
+      scope('when closing the modal', () => {
+        closeModal();
+        Expect.openModal(false);
```
erocheleau, line 41: Should you expect display* to be false?

packages/quantic/force-app/examples/main/lwc/exampleQuanticModal/exampleQuanticModal.js-meta.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>54.0</apiVersion>
+    <isExposed>true</isExposed>
+    <targets>
+      <target>lightning__RecordPage</target>
+      <target>lightning__AppPage</target>
+      <target>lightning__HomePage</target>
+      <target>lightningCommunity__Page</target>
+      <target>lightningCommunity__Default</target>
```
erocheleau, line 10: Do we really want to be exposed in all of these places?

erocheleau (approved):
Everything looks good 👍

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -0,0 +1,56 @@
+import {LightningElement, api} from 'lwc';
+
+/**
+ * The `QuanticModal` component provides the logic to easily implement a modal in Quantic. This component handles the animation logic, exposes methods to open and close the modal, offers the option to open the modal in full screen or just to cover the search interface, and exposes a set of slots to fully customize the modal content.
+ *
+ * @category Search
+ * @example
+ * <c-quantic-modal full-screen animation="slideToLeft"></c-quantic-modal>
```
nathanlb, line 8: I think this example might be a good place to show usage of the available slots.

mmitiche: Sure thing👌 I'll add that.

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
nathanlb: "QuanticX component provides the logic to easily implement a X in Quantic" i feel like this could apply to any component so I'm not sure this first sentence is very useful.
What about "The `QuanticModal` is a container component that displays slotted content in a modal."

jpmarceau: That sounds like a good idea to me 👍

nathanlb (approved):
Pretty darn awesome 👏

PR #1943, opened 2022-04-22
fix(quantic): initialized callback executing more than once fixed

mmitiche:
### Bug:
[SFINT-4494](https://coveord.atlassian.net/browse/SFINT-4494)
Adding a new Quantic component to the page after the first initialization of the search interface is triggering the execution of the the initialization loop for a second time causing by this a weird behaviour by the Quantic Facets: the facet values are deleted after selecting one value.

### Expected behaviour:
The initialization loop should run only once and the facets should work as expected after adding them after the first execution of the the initialization loop.

### Source of the problem:
`buildUrlManager` headless controller is called in each initialization loop, building this controller is triggering the dispatch of the following action [restoreSearchParameters](https://github.com/coveo/ui-kit/blob/7640744168e5c46d5b2e5bd578e9ac899d975602/packages/headless/src/features/facets/facet-set/facet-set-slice.ts#L59) , which is causing the deletion of the facet values.

https://user-images.githubusercontent.com/86681870/164709896-866fa84d-ceec-43db-817e-e33da3909d3e.mov

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
@@ -76,6 +76,9 @@ export default class QuanticSearchInterface extends LightningElement {
   /** @type {Function} */
   unsubscribeUrlManager;
 
+  /** @type {boolean} */
+  initialized = false;
+
   connectedCallback() {
     loadDependencies(this).then(() => {
       if (!getHeadlessBindings(this.engineId)?.engine) {
@@ -116,25 +119,29 @@ export default class QuanticSearchInterface extends LightningElement {
    * @param {SearchEngine} engine
    */
   initialize = (engine) => {
-    const {updateQuery} = CoveoHeadless.loadQueryActions(engine);
+    if (!this.initialized) {
+      const {updateQuery} = CoveoHeadless.loadQueryActions(engine);
 
-    if (!this.disableStateInUrl) {
-      this.initUrlManager(engine);
-    }
-
-    if (!this.skipFirstSearch) {
-      const redirectData = window.localStorage.getItem(
-        STANDALONE_SEARCH_BOX_STORAGE_KEY
-      );
-      if (!redirectData) {
-        engine.executeFirstSearch();
-        return;
+      if (!this.disableStateInUrl) {
+        this.initUrlManager(engine);
       }
-      window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
-      const {value, analytics} = JSON.parse(redirectData);
+  
+      if (!this.skipFirstSearch) {
+        const redirectData = window.localStorage.getItem(
+          STANDALONE_SEARCH_BOX_STORAGE_KEY
+        );
+        if (!redirectData) {
+          engine.executeFirstSearch();
+          return;
+        }
+        window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
+        const {value, analytics} = JSON.parse(redirectData);
+  
+        engine.dispatch(updateQuery({q: value}));
+        engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
 
-      engine.dispatch(updateQuery({q: value}));
-      engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
+        this.initialized = true;
+      }
     }
   };
 
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
@@ -132,13 +132,13 @@ export default class QuanticSearchInterface extends LightningElement {
         );
         if (!redirectData) {
           engine.executeFirstSearch();
-          return;
-        }
-        window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
-        const {value, analytics} = JSON.parse(redirectData);
+        } else {
+          window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
+          const {value, analytics} = JSON.parse(redirectData);
   
-        engine.dispatch(updateQuery({q: value}));
-        engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
+          engine.dispatch(updateQuery({q: value}));
+          engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
+        }
       }
       this.initialized = true;
     }
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
@@ -119,29 +119,30 @@ export default class QuanticSearchInterface extends LightningElement {
    * @param {SearchEngine} engine
    */
   initialize = (engine) => {
-    if (!this.initialized) {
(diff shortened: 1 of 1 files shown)


packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -116,25 +119,29 @@ export default class QuanticSearchInterface extends LightningElement {
...
+        }
+        window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
+        const {value, analytics} = JSON.parse(redirectData);
+  
+        engine.dispatch(updateQuery({q: value}));
+        engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
 
-      engine.dispatch(updateQuery({q: value}));
-      engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
+        this.initialized = true;
```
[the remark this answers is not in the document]
mmitiche: The initialize method of the Quantic Search Interface is executed only when all the components are initialized, this is handled by the [QuanticHeadlessLoader](https://github.com/coveo/ui-kit/blob/e200e1dcac1c619498bd23da821b7bd7c36dc11d/packages/quantic/force-app/main/default/lwc/quanticHeadlessLoader/quanticHeadlessLoader.js#L202-L203).
The only thing I'm doing here is preventing this initialize method to be executed more than once.

nathanlb:
Have you verified that this change works well with the standalone searchbox in a community setting? The SPA navigation interacts with the engine initialization cycles in a tricky way sometimes so we should test with a SSB in the header.

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
lbergeron, line 143: I guess this line should be outside and after the `if (!this.skipFirstSearch)` block because if `skipFirstSearch` is set to `true`, then `initialized` is always `false`.

lbergeron:
Did you try loading a search page with selected facets directly from the URL (like a bookmarked search page)?

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
mmitiche: True thank you👌 I accidentally added it inside the if statement.

mmitiche:
> Have you verified that this change works well with the standalone searchbox in a community setting? The SPA navigation interacts with the engine initialization cycles in a tricky way sometimes so we should test with a SSB in the header.

I will check that thanks👌 I just have one question, what does SPA navigation mean?

nathanlb:
> > Have you verified that this change works well with the standalone searchbox in a community setting? The SPA navigation interacts with the engine initialization cycles in a tricky way sometimes so we should test with a SSB in the header.
> 
> I will check that thanks👌 I just have one question, what does SPA navigation mean?

SPA = Single Page App 😅 sorry about that.
Because the community is a single page app, the header remains while the page content changes, hence the sometimes tricky behaviour regarding component lifecycles.

mmitiche:
> Did you try loading a search page with selected facets directly from the URL (like a bookmarked search page)?

Yes I did, and it's working correctly 👌

mmitiche:
> > > Have you verified that this change works well with the standalone searchbox in a community setting? The SPA navigation interacts with the engine initialization cycles in a tricky way sometimes so we should test with a SSB in the header.
> > 
> > 
> > I will check that thanks👌 I just have one question, what does SPA navigation mean?
> 
> SPA = Single Page App 😅 sorry about that. Because the community is a single page app, the header remains while the page content changes, hence the sometimes tricky behaviour regarding component lifecycles.

Seems to work just fine with the standalone search box👌

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -116,25 +119,28 @@ export default class QuanticSearchInterface extends LightningElement {
    * @param {SearchEngine} engine
    */
   initialize = (engine) => {
-    const {updateQuery} = CoveoHeadless.loadQueryActions(engine);
-
-    if (!this.disableStateInUrl) {
-      this.initUrlManager(engine);
-    }
+    if (!this.initialized) {
```
erocheleau, line 122:
```suggestion
    if (this.initialized) {
      return;
    }
```
Do we prefer this instead of a big if?

erocheleau (approved):
As long as you test it everywhere, looks good to me :)

nathanlb (approved):

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -116,25 +119,28 @@ export default class QuanticSearchInterface extends LightningElement {
...
-      engine.dispatch(updateQuery({q: value}));
-      engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
+  
+      if (!this.skipFirstSearch) {
+        const redirectData = window.localStorage.getItem(
+          STANDALONE_SEARCH_BOX_STORAGE_KEY
+        );
+        if (!redirectData) {
+          engine.executeFirstSearch();
+          return;
```
liviarett, line 135: do we not want to set `this.initialized = true` here?

if so, could not return and just do the following:
```ts
        if (!redirectData) {
          engine.executeFirstSearch();
        } else {
          window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
          const {value, analytics} = JSON.parse(redirectData);

          engine.dispatch(updateQuery({q: value}));
          engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
        }
        this.initialized = true;
```

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -116,25 +119,28 @@ export default class QuanticSearchInterface extends LightningElement {
...
+        }
+        window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
+        const {value, analytics} = JSON.parse(redirectData);
+  
+        engine.dispatch(updateQuery({q: value}));
+        engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
```
lbergeron, line 141: If we reach the `return` statement on line 135, `this.initialized` will never be set to `true`. I'd use an `if/else` instead of an `if/return`.

```suggestion
        if (!redirectData) {
          engine.executeFirstSearch();
        } else {
          window.localStorage.removeItem(STANDALONE_SEARCH_BOX_STORAGE_KEY);
          const {value, analytics} = JSON.parse(redirectData);
  
          engine.dispatch(updateQuery({q: value}));
          engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
        }
```

lbergeron (changes requested):
`this.initialized` is not always set.

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
mmitiche: Good catch! Thank you 👌

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
mmitiche: Thanks! I'll update this 👌

lbergeron (approved):

PR #1945, opened 2022-04-25
feat(atomic,headless): add analytics to the show more/show less button

liviarett:
https://coveord.atlassian.net/browse/KIT-1576?atlOrigin=eyJpIjoiZmQxZDZmNjMwOTQwNGUzNWE1NzJmMjM3Y2ZjNTk1MTQiLCJwIjoiaiJ9

### DIFF
--- a/packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
+++ b/packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
@@ -1,6 +1,7 @@
 import {
   FoldedCollection,
   FoldedResult,
+  FoldedResultList,
   FoldedResultListState,
   Result,
   ResultTemplatesManager,
@@ -20,7 +21,10 @@ import {
 } from '../../result-template-components/result-template-decorators';
 import {ResultListCommon} from '../result-list-common';
 import {TemplateContent} from '../../result-templates/result-template-common';
-import {FoldedResultListStateContext} from '../result-list-decorators';
+import {
+  FoldedResultListContext,
+  FoldedResultListStateContext,
+} from '../result-list-decorators';
 import {ResultDisplayImageSize} from '../../atomic-result/atomic-result-display-options';
 import {ListDisplayResultsPlaceholder} from '../list-display-results-placeholder';
 import {Button} from '../../common/button';
@@ -56,6 +60,9 @@ export class AtomicResultChildren {
   @State() public error!: Error;
   @State() public ready = false;
   @State() public templateHasError = false;
+
+  @FoldedResultListContext()
+  private foldedResultList!: FoldedResultList;
 
   @FoldedResultListStateContext()
   @State()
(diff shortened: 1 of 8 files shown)


packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -86,6 +89,16 @@ export interface FoldedResultList extends Controller {
    * @param collection - The collection for which to load more results.
    */
   loadCollection(collection: FoldedCollection): void;
```
olamothe, line 91: Hum why not bake in the `dispatch(logShowMoreFoldedResults(result))` inside the `loadCollection` function ?

We want to encourage users as much as possible to log analytics on every interaction, and having it automatically done in the function that you need to call anyway to "show more" seems more foolproof.

Having proper analytics recording is a big pain point for any Coveo deployment, and so we must make sure it's as automatic/transparent as possible for implementers, and that they don't have to go out of their way to remember to call a specific function at the right time.

liviarett: makes sense!

btaillon (approved):

packages/atomic/src/components/result-lists/atomic-result-children/atomic-result-children.tsx
```diff
@@ -230,7 +237,16 @@ export class AtomicResultChildren {
           part="show-hide-button"
           class="show-hide-button"
           style="text-primary"
-          onClick={() => (this.hideResults = !this.hideResults)}
+          onClick={() => {
+            if (this.hideResults) {
+              this.foldedResultList.logShowMoreFoldedResults(
```
olamothe, line 242: If the event is called when loading a collection, I don't understand why we re-send the event here ?

Doesn't it get sent two times to the API then ?

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -172,6 +185,13 @@ export function buildFoldedResultList(
           ] as string
         )
       );
+      dispatch(logShowMoreFoldedResults(collection.result));
+    },
+    logShowMoreFoldedResults: (result) => {
```
olamothe, line 190: Related to my above comment, if the event is now dispatched when loading the collection, this function now seems superfluous and could be removed.

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -86,6 +89,16 @@ export interface FoldedResultList extends Controller {
...
    */
   loadCollection(collection: FoldedCollection): void;
+  /**
+   * Logs a Click Event for when user loads/shows more folded results.
+   *
+   * @param result - The result for which to send analytics.
+   */
+  logShowMoreFoldedResults(result: Result): void;
+  /**
+   * Logs a Custom Event for when user shows less folded results.
```
npushkarskii, line 99:
```suggestion
   * Logs a click event when a user loads/shows more folded results.
   *
   * @param result - The result to send analytics for.
   */
  logShowMoreFoldedResults(result: Result): void;
  /**
   * Logs a custom event when a user shows less folded results.
```

npushkarskii (approved):
LGTM 👍

ThibodeauJF (approved):

PR #1947, opened 2022-04-25
feat(atomic): make grid result <a> a sibling instead of parent

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1574

Why? https://html.spec.whatwg.org/#the-a-element

> [Transparent](https://html.spec.whatwg.org/#transparent), but there must be no [interactive content](https://html.spec.whatwg.org/#interactive-content-2) descend[a](https://html.spec.whatwg.org/#the-a-element)nt, a element descendant, or descendant with the [tabindex](https://html.spec.whatwg.org/#attr-tabindex) attribute specified.

Inspired by https://css-tricks.com/nested-links/ & https://inclusive-components.design/cards/

### DIFF
--- a/packages/atomic/src/components/result-lists/grid-display-results.tsx
+++ b/packages/atomic/src/components/result-lists/grid-display-results.tsx
@@ -12,7 +12,7 @@ export const GridDisplayResults: FunctionalComponent<ResultsProps> = (
     });
 
     return (
-      <div part="result-list-grid-parent">
+      <div part="result-list-grid-clickable-container">
         <LinkWithResultAnalytics
           part="result-list-grid-clickable"
           onSelect={() => interactiveResult.select()}
--- a/packages/atomic/src/components/result-lists/grid-display-results.tsx
+++ b/packages/atomic/src/components/result-lists/grid-display-results.tsx
@@ -12,14 +12,16 @@ export const GridDisplayResults: FunctionalComponent<ResultsProps> = (
     });
 
     return (
-      <LinkWithResultAnalytics
-        part="result-list-grid-clickable"
-        onSelect={() => interactiveResult.select()}
-        onBeginDelayedSelect={() => interactiveResult.beginDelayedSelect()}
-        onCancelPendingSelect={() => interactiveResult.cancelPendingSelect()}
-        href={props.resultListCommon.getUnfoldedResult(result).clickUri}
-        target="_self"
-      >
+      <div part="result-list-grid-parent">
+        <LinkWithResultAnalytics
+          part="result-list-grid-clickable"
+          onSelect={() => interactiveResult.select()}
+          onBeginDelayedSelect={() => interactiveResult.beginDelayedSelect()}
+          onCancelPendingSelect={() => interactiveResult.cancelPendingSelect()}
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/result-lists/result-list-common.pcss
```diff
@@ -73,9 +73,18 @@
...
+  cursor: pointer;
+}
+
+[part='result-list-grid-clickable'] {
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  pointer-events: none;
```
liviarett, line 87: doesn't this prevent the click event from being dispatched?

ThibodeauJF: 🤦 🤦 🤦 🤦 🤦

packages/atomic/src/components/result-lists/grid-display-results.tsx
```diff
@@ -12,14 +12,16 @@ export const GridDisplayResults: FunctionalComponent<ResultsProps> = (
...
     return (
-      <LinkWithResultAnalytics
-        part="result-list-grid-clickable"
-        onSelect={() => interactiveResult.select()}
-        onBeginDelayedSelect={() => interactiveResult.beginDelayedSelect()}
-        onCancelPendingSelect={() => interactiveResult.cancelPendingSelect()}
-        href={props.resultListCommon.getUnfoldedResult(result).clickUri}
-        target="_self"
-      >
+      <div part="result-list-grid-parent">
```
olamothe, line 15: This part does not seem properly documented

packages/atomic/src/components/result-lists/result-list-common.pcss
ThibodeauJF: Reworking lol

olamothe (approved):

ThibodeauJF:
To disable would just be 

```
atomic-result-list::part(result-list-grid-clickable) {
  display: none;
}
```

packages/atomic/src/components/result-lists/grid-display-results.tsx
olamothe, line 15: Suggestion: 
```suggestion
      <div part="result-list-grid-clickable-container">
```

olamothe (approved):

packages/atomic/src/components/result-lists/grid-display-results.tsx
ThibodeauJF: hmm it contains the clickable link & the result too? maybe `result-list-grid-container`

liviarett (approved):
Looks good, but looking at the description you added made me wonder if a screen reader wouldn't consider this a clickable link like any other... and then if you actually select the result with it, it won't send the analytics? I don't know if that makes sense, but just a thought!

ThibodeauJF:
Good question!

Yes it is a link taking the whole space, if you click anywhere (except a child a or button) it redirects you & logs a click. Other links that we have such as result link already log analytics so it's good. In non-grid results, that's already how it works

jpmarceau (approved):
👍

PR #1948, opened 2022-04-25
feat(quantic): Create QuanticRefineModalContent component

mmitiche:
[SFINT-4471](https://coveord.atlassian.net/browse/SFINT-4471)
The `QuanticRefineModalContent` component duplicates and displays the facets present in the search interface in addition to the sort component. This component is primarily  made to be displayed inside a modal to assure the responsiveness  when the search interface is displayed in smaller screens.

#### Demo

https://user-images.githubusercontent.com/86681870/165296127-f7d82921-cff9-4345-a5fe-1e151431e2e0.mp4


packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -0,0 +1,204 @@
...
+   * @param {HTMLElement} element
+   * @param {Array<string>} properties
+   * @returns {object}
+   */
+  extractFacetDataFromElement(element, properties) {
+    const data = {};
+    properties.forEach((property) => {
+      data[property] = element[property];
+    });
+    return data;
```
lbergeron, line 194: You could use `reduce` here to copy a subset of the object properties.

```suggestion
    return properties.reduce((data, property) => {
      data[property] = element[property];
      return data;
    }, {});
```

packages/quantic/cypress/integration/refine-modal-content/refine-modal-content-expectations.ts
```diff
@@ -0,0 +1,223 @@
...
+        );
+    },
+
+    correctFacetsOrder: () => {
+      selector
+        .duplicatedFacetManagerItems()
+        .should('exist')
+        .then((duplicatedFacetManagerItems) => {
+          const duplicatedFacetsOrder = duplicatedFacetManagerItems.map(
+            (index) => duplicatedFacetManagerItems[index].firstChild.tagName
```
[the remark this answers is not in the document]
mmitiche: Updated 👌

packages/quantic/cypress/integration/refine-modal-content/refine-modal-content-expectations.ts
```diff
@@ -0,0 +1,223 @@
...
+            .then((facetManagerItems) => {
+              const facetsOrder = facetManagerItems.map(
+                (index) => facetManagerItems[index].firstChild.tagName
+              );
+              expect(facetsOrder.length).to.be.eq(duplicatedFacetsOrder.length);
+              for (let index = 0; index < facetsOrder.length; index++) {
+                expect(facetsOrder[index]).to.be.eq(
+                  duplicatedFacetsOrder[index]
+                );
+              }
```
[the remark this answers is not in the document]
mmitiche: Yes I was able to do that using the [eql method](https://www.chaijs.com/api/bdd/#method_eql) 👌

packages/quantic/cypress/integration/refine-modal-content/refine-modal-content-selectors.ts
```diff
@@ -0,0 +1,110 @@
+import {ComponentSelector, CypressSelector} from '../common-selectors';
+
+const refineContentComponent = 'c-quantic-refine-modal-content';
+const refineContentContainer = '.facets-container';
+
+export interface RefineContentSelector extends ComponentSelector {
```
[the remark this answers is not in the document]
mmitiche: Neat idea! I followed the logic you described to implement it, thank you!

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Returns facet data.
+   * @returns {Array<object>}
+   */
+  get facets() {
+    if (!this.data) {
+      return [];
+    }
+    const facetData = Object.keys(this.data).map((facetId) => {
```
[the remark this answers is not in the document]
mmitiche: updated 👌

packages/quantic/cypress/integration/refine-modal-content/refine-modal-content-selectors.ts
```diff
@@ -0,0 +1,96 @@
...
+  timeframeFacetClearFiltersButton: () => CypressSelector;
+  facetClearFiltersButton: () => CypressSelector;
+  timeframeFacetFirstOption: () => CypressSelector;
+  facetFirstOption: () => CypressSelector;
+}
+
+const getCommonFacetSelectors = (container: boolean): FacetSelector => {
+  return {
+    get: () =>
+      cy.get(container ? refineContentContainer : refineContentComponent),
```
lbergeron, line 32: You could get rid of the ternary on each selector by passing a Cypress selector to `getCommonFacetSelectors` and chaining the other calls from there.

For example:
```
const getCommonFacetSelectors = (rootSelector): FacetSelector => {
  return {
    get: () => rootSelector(),
    facet: () => rootSelector().find('c-quantic-facet'),
    ...
  };
}
```

and you would invoke `getCommonFacetSelectors` this way:
```
getCommonFacetContainer(() => cy.get(refineContentComponent));
getCommonFacetContainer(() => cy.get(refineContentContainer));
```

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -0,0 +1,219 @@
...
+  clearAllFilters() {
+    this.breadcrumbManager.deselectAll();
+  }
+
+  /**
+   * Returns the data needed to create a copy of the numeric facet.
+   * @param {HTMLElement} facetElement
+   * @returns {object}
+   */
+  static toNumericFacet(facetElement) {
```
lbergeron, line 120: I think these functions don't have to be `static`. It is best to keep them as private instance functions at this moment.

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineModalContent/exampleQuanticRefineModalContent.js-meta.xml
```diff
@@ -0,0 +1,12 @@
...
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
+    <isExposed>true</isExposed>
+    <targets>
+      <target>lightning__RecordPage</target>
+      <target>lightning__AppPage</target>
+      <target>lightning__HomePage</target>
+      <target>lightningCommunity__Page</target>
+      <target>lightningCommunity__Default</target>
+  </targets>
```
erocheleau, line 11: Do we really want to expose in all of these targets? keep in mind this adds this example component in all the page builder for these environment

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -196,7 +196,8 @@ export default class QuanticCategoryFacet extends LightningElement {
     this.unsubscribe = this.facet.subscribe(() => this.updateState());
     registerToStore(this.engineId, Store.facetTypes.CATEGORYFACETS, {
       label: this.label,
-      facetId: this.facet.state.facetId
+      facetId: this.facet.state.facetId,
+      element: this.template.host
```
erocheleau, line 200:
```suggestion
      facetId: this.facet?.state?.facetId,
      element: this.template.host
```

maybe?

erocheleau: oh... why do we need a reference towards the template.host now?

packages/quantic/force-app/main/default/lwc/quanticHeadlessLoader/quanticHeadlessLoader.js
```diff
@@ -267,7 +267,7 @@ function destroyEngine(engineId) {
  * Register a facet in the store.
  * @param {string} engineId The engine ID.
  * @param {string} facetType
- * @param {{ label: string; facetId: string; format?: function}} data
```
erocheleau, line 270: This seems more or less like a weird workaround no?

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.css
```diff
@@ -0,0 +1,5 @@
+.filters-header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+}
```
erocheleau, line 5: Just FYI, I don't know what we've decided to do here, but there is already slds classes to do that afaik

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.html
```diff
@@ -0,0 +1,86 @@
...
+            sort-criteria={facet.sortCriteria}
+            no-search={facet.noSearch}
+            display-values-as={facet.displayValuesAs}
+            no-filter-facet-count={facet.noFilterFacetCount}
+            injection-depth={facet.injectionDepth}
+            key={facet.field}
+            is-collapsed
+          ></c-quantic-facet>
+        </template>
+        <template if:true={facet.isTimeframe}>
```
erocheleau, line 70: Just for readability, can we put timeFrameFacet higher than the default one?

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -0,0 +1,219 @@
...
+  'delimitingCharacter',
+  'numberOfValues',
+  'sortCriteria',
+  'withSearch',
+];
+const TIMEFRAME_FACET_PROPERTIES = [
+  'withDatePicker',
+  'noFilterFacetCount',
+  'injectionDepth',
+];
```
erocheleau, line 43: Is there a way other facets could export those? Would we want to handle it like that instead?

erocheleau (approved):
There are a couple of fundamental questions raised I think and I'll let @nathanlb team decide if they are okay with the global changes, example of passing the this.template.host as a prop, but otherwise all the code looks great! 👍

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineModalContent/exampleQuanticRefineModalContent.js-meta.xml
mmitiche: probably only the community page is needed here true.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
mmitiche: I'm passing the template.host here to save all the attribute of the facet in the Quantic store which give us the ability to duplicate the facets in the modal with the same values for all attributes to get the exact same facets.

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.html
mmitiche: Sure!

packages/quantic/force-app/main/default/lwc/quanticHeadlessLoader/quanticHeadlessLoader.js
mmitiche: This was the only workaround I found to be able to create the exact duplicates of each facet, any ideas for a better alternative solution?

lvu285:
Are you sure we still want to display "SORT component" as a dropdown inside the model?

nathanlb:
> Are you sure we still want to display "SORT component" as a dropdown without formatting it to have similar style with other Filters inside the model? image

the mock-ups show a good solution for this. adding css to lwc can be tough due to the shadow dom but hopefully it won't be too bad.

mmitiche:
> Are you sure we still want to display "SORT component" as a dropdown without formatting it to have similar style with other Filters inside the model?
> image
Good point I think it will be great to format the sort component to have a similar style to fit with the other filters👌
I'm just wondering if we need to do that now or we need to do it in another iteration since our main focus now is in the Hosted Insight Panel  which will have a QuanticRefineModalContent without the sort component.
I would like to hear the team's opinion on this.

nathanlb:
> Good point I think it will be great to format the sort component to have a similar style to fit with the other filters👌 I'm just wondering if we need to do that now or we need to do it in another iteration since our main focus now is in the Hosted Insight Panel which will have a QuanticRefineModalContent without the sort component. I would like to hear the team's opinion on this.

It can be done separately just create a user story jira and you can add it to this epic https://coveord.atlassian.net/browse/SVCC-76

mmitiche:
> It can be done separately just create a user story jira and you can add it to this epic https://coveord.atlassian.net/browse/SVCC-76

Perfect! Thank you!

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineModalContent/exampleQuanticRefineModalContent.html
```diff
@@ -0,0 +1,81 @@
...
+            <c-quantic-timeframe unit="year"></c-quantic-timeframe>
+            <c-quantic-timeframe
+              amount="10"
+              unit="year"
+              label="Last decade"
+            ></c-quantic-timeframe>
+          </c-quantic-timeframe-facet>
+        </c-quantic-facet-manager>
+      </div>
+      <c-quantic-refine-modal-content engine-id={engineId} hide-sort={config.hideSort}></c-quantic-refine-modal-content>
```
lvu285, line 78: for testing, I suggest to setup the page where `The Quantic-refine-modal is passed in the content slot of the Quantic-modal Component` _ like the way you show me, it's safer, as these 2 components should be used together (?)

lvu285:
general question:
do we plan to make all new Quantic components 'Accessibility' compliant?

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
mmitiche: We can the properties as static properties in each facets to get something like this:
static
and then then we can use them in the QuanticRefineModalContent component as follows:
`const NUMERIC_FACET_PROPERTIES = QuanticNumericFacet.attributes;`

What do you think?

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineModalContent/exampleQuanticRefineModalContent.html
mmitiche: Actually the two components will be tested together in the tests of the [QuanticRefineToggle component](https://github.com/coveo/ui-kit/blob/055987fec3f4354a4143f7e8410514da89a65c40/packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.html#L17-L40). this component is the button that will be used to open the refine modal.
I will open[ the PR of the QuanticRefineToggle](https://github.com/coveo/ui-kit/pull/1970) after merging this one.

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * The `QuanticRefineModalContent` component displays a copy of the search interface facets and sort components. This component is intended to be displayed inside a modal to assure the responsiveness when the search interface is displayed on smaller screens.
```
nathanlb, line 46: instead of "a modal" i think it would be better to directly mention the QuanticModal component you built for this purpose :)

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -0,0 +1,219 @@
...
+  initialize = (engine) => {
+    this.searchStatus = CoveoHeadless.buildSearchStatus(engine);
+    this.breadcrumbManager = CoveoHeadless.buildBreadcrumbManager(engine);
+    this.unsubscribeSearchStatus = this.searchStatus.subscribe(() =>
+      this.gatherFacets()
+    );
+  };
+
+  /**
+   * Gather all facets registred from the Quantic store.
```
nathanlb, line 98:
```suggestion
   * Gather all facets registered in the Quantic store.
```

or 

```suggestion
   * Gather all registered facets from the Quantic store.
```

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.html
```diff
@@ -0,0 +1,86 @@
...
+        variant="base"
+        label={labels.clearAllFilters}
+        title={labels.clearAllFilters}
+        onclick={clearAllFilters}
+      ></lightning-button>
+    </div>
+    <c-quantic-facet-manager engine-id={engineId}>
+      <template for:each={facets} for:item="facet">
+        <template if:true={facet.isNumeric}>
+          <c-quantic-numeric-facet
```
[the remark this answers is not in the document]
mmitiche: No, we're not removing the other facets. For example, in ExampleSearch, we'll just hide the original facets with a bit of CSS when the search interface is open in a narrow screen, and instead show the facets of refined modal.
but both sets of facets are always there at all time and both communicate with the engine, not sure how we can improve that :/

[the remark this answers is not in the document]
mmitiche: Yes, from an accessibility standpoint the originals are hidden and the screen reader can't see them.
Anything having the the  CSS property : `display: none` will be hidden from an Accessibility standpoint.

nathanlb (approved):
GG!

PR #1949, opened 2022-04-26
feat(atomic): allow for custom query suggestions + example

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1582

### DIFF
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -744,6 +744,10 @@ export namespace Components {
           * Defining this option makes the search box standalone.  This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
          */
         "redirectionUrl"?: string;
+        /**
+          * This option controls the timeout for suggestion queries. If a query times out, only the suggestions from that particular query won't be shown.
+         */
+        "suggestionTimeout": number;
     }
     interface AtomicSearchBoxQuerySuggestions {
         /**
@@ -2200,6 +2204,10 @@ declare namespace LocalJSX {
           * Defining this option makes the search box standalone.  This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
          */
         "redirectionUrl"?: string;
+        /**
+          * This option controls the timeout for suggestion queries. If a query times out, only the suggestions from that particular query won't be shown.
+         */
+        "suggestionTimeout"?: number;
     }
     interface AtomicSearchBoxQuerySuggestions {
         /**
--- a/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
@@ -82,6 +82,12 @@ export class AtomicSearchBox {
    */
   @Prop({reflect: true}) public redirectionUrl?: string;
 
+  /**
+   * This option controls the timeout for suggestion queries.
+   * If a query times out, only the suggestions from that particular query won't be shown.
+   */
+  @Prop() public suggestionTimeout = 400;
+
   @AriaLiveRegion('search-box')
   protected ariaMessage!: string;
 
@@ -260,11 +266,24 @@ export class AtomicSearchBox {
 
   private async triggerSuggestions() {
     const settled = await Promise.allSettled(
-      this.suggestions.map((suggestion) => promiseTimeout(suggestion.onInput()))
+      this.suggestions.map((suggestion) =>
+        promiseTimeout(suggestion.onInput(), this.suggestionTimeout)
+      )
     );
 
-    const suggestionElements = this.suggestions
-      .filter((_, i) => settled[i].status === 'fulfilled')
+    const fulfilledSuggestions: SearchBoxSuggestions[] = [];
+
+    settled.forEach((prom, j) => {
+      if (prom.status === 'fulfilled') {
+        fulfilledSuggestions.push(this.suggestions[j]);
+      } else {
+        console.error(
+          'Some query suggestions are not being shown because the promise timed out.'
+        );
+      }
+    });
+
+    const suggestionElements = fulfilledSuggestions
       .sort((a, b) => a.position - b.position)
       .map((suggestion) => suggestion.renderItems())
       .flat();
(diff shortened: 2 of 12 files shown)


packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -416,8 +416,13 @@ export class AtomicSearchBox {
         }`}
         onMouseDown={(e) => e.preventDefault()}
         onClick={() => suggestion.onSelect()}
+        ref={(e) => {
```
liviarett, line 419: I didn't quite understand what's going on with this! you said that we could get reference to the other children, how is that working here?

ThibodeauJF: For getting reference to the other children, that already exists, you can look at the `getSuggestions` method of the `SearchBoxSuggestionsBindings`. It is used in `AtomicSearchBoxQuerySuggestions` to check whether to add an icon or not

packages/atomic/src/components/search-box-suggestions/suggestions-common.ts
```diff
@@ -32,12 +33,17 @@ export interface SearchBoxSuggestionsBindings extends Bindings {
...
-  element: Element
+  element: HTMLElement
 ) => {
-  const canceled = element.dispatchEvent(
+  element.dispatchEvent(
     buildCustomEvent('atomic/searchBoxSuggestion/register', event)
   );
-  if (canceled) {
+
+  const hasParentSearchBox = elementHasAncestorTag(
```
liviarett, line 42: how comes you changed this here? from a quick test, it seems to work with the `canceled` pattern but not with this new one for the old components (because of initialisation order)

ThibodeauJF: Benjamin made this recently, it allows to traverse the Shadow DOM

liviarett: I made `elementHasAncestorTag` a while ago, I can see that Benjamin added a `closest` method that checks for shadow DOM too. I'll see if I can replace with that

ThibodeauJF: oh my bad I meant to use the closest method then 😂 

Seems that we can merge these methods?

btaillon: woops, my bad!

liviarett: I actually didn't replace the other place that's using my method yet, because I'm not sure if traversing shadow doms will keep the behaviour that I want! so don't worry about it, I'll look at it

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -258,10 +259,13 @@ export class AtomicSearchBox {
   }
 
   private async triggerSuggestions() {
-    await Promise.all(
-      this.suggestions.map((suggestion) => suggestion.onInput())
+    // eslint-disable-next-line node/no-unsupported-features/es-builtins
+    const settled = await Promise.allSettled(
```
btaillon, line 263: I knew about `Promise.all` but not `Promise.allSettled`, neat! will try to keep in mind from now on.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -258,10 +259,13 @@ export class AtomicSearchBox {
...
+      this.suggestions.map((suggestion) => promiseTimeout(suggestion.onInput()))
```
btaillon, line 264: I think I would prefer not to implicitly add a time out, especially if they time out silently, since it may be difficult to debug why some particular component works *sometimes* but not always.

However, I would keep the `Promise.allSettled` and I would try to log whenever a promise rejected.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -415,9 +419,17 @@ export class AtomicSearchBox {
           isSelected ? 'bg-neutral-light' : ''
         }`}
         onMouseDown={(e) => e.preventDefault()}
-        onClick={() => suggestion.onSelect()}
+        onClick={() => {
+          suggestion.onSelect();
+          this.clearSuggestions();
```
btaillon, line 424: you should double-check that this doesn't notify screen readers.

btaillon (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: so just let it take as long as it takes? I wonder if there should be a loading state, as currently it will just pop up out of nowhere when it's ready, and if it takes like 2s, it's quite a while

olamothe: In this case I think we should prioritize end user experience over dev experience.

There should be some limit to how long we wait, so that end user can benefit from "some" suggestion if one of the provider does not return properly.

When one provider fails to do its job, we should definitely be very verbose in the console as to why something was not shown so that a dev can at least quickly understand what's happening.

This timeout should definitely be configurable (an option on the search box), and have a sane default of ... couple hundred ms... ?

olamothe (approved):

packages/atomic/src/components.d.ts
```diff
@@ -2200,6 +2204,10 @@ declare namespace LocalJSX {
           * Defining this option makes the search box standalone.  This option defines the default URL the user should be redirected to, when a query is submitted. If a query pipeline redirect is triggered, it will redirect to that URL instead (see [query pipeline triggers](https://docs.coveo.com/en/1458)).
          */
         "redirectionUrl"?: string;
+        /**
+          * This option controls the timeout for suggestion queries. If a query times out, only the suggestions from that particular query won't be shown.
```
liviarett, line 2208: @jpmarceau any suggestions on this beautiful sentence?

jpmarceau: Only minor ones :) 
```suggestion
          * The timeout for suggestion queries, in milliseconds. If a query times out, the suggestions from that particular query won't be shown.
```

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -81,6 +82,12 @@ export class AtomicSearchBox {
    */
   @Prop({reflect: true}) public redirectionUrl?: string;
 
+  /**
+   * This option controls the timeout for suggestion queries.
+   * If a query times out, only the suggestions from that particular query won't be shown.
```
jpmarceau, line 87:
```suggestion
   * The timeout for suggestion queries, in milliseconds.
   * If a query times out, the suggestions from that particular query won't be shown.
```

PR #1954, opened 2022-04-29
feat(atomic,headless): added atomic-smart-snippet-suggestions

btaillon:
https://coveord.atlassian.net/browse/KIT-1489

Missing:
* Show more / Show less (& analytics)
* Analytics when clicking on the source
* A unique ID for related questions
  * Currently, the "atlas data lake" query returns multiple related questions that each have a different question/answer but share the same source. The expand/collapse headless action takes the source as a parameter, meaning that we can only expand/collapse all of them at once.
  * The ID could be exposed and used as a key in JSX frameworks
* Analytics when expanding/collapsing a related question

Live demo: https://staticdev.cloud.coveo.com/atomic/pr/KIT-1489-storybook/index.html?path=/story/atomic-smart-snippet-suggestions--default-smart-snippet-suggestions


packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+ * @part heading - The heading above the list of related question.
+ * @part questions - The list of related questions.
+ * @part question-answer-expanded - An expanded related question.
+ * @part question-answer-collapsed - A collapsed related question.
+ * @part question-expanded-button - The button to collapse a related question.
+ * @part question-collapsed-button - The button to expand a related question.
+ * @part question-expanded-icon - The caret of an expanded related question.
+ * @part question-collapsed-icon - The caret of a collapsed related question.
+ * @part question-expanded-text - The title of an expanded related question.
+ * @part question-collapsed-text - The title of a collapsed related question.
```
btaillon, line 49: I added a `-expanded` and `-collapsed` version of most parts that are related to a relatedQuestion because I don't see any other way that those could be styled if a client wants a special expanded or collapsed style. We could also be more opinionated here and simplify the parts. What do you think?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+ * @part answer-and-source - The wrapper around the answer and source of a related question.
+ * @part answer - The container displaying the full document excerpt of a related question's answer.
+ * @part footer - The wrapper around the source of a related question's answer.
+ * @part source-url - The URL to the document a related question's answer is extracted from.
+ * @part source-title - The title of the document a related question's answer is extracted from.
+ *
+ * @internal
```
btaillon, line 56: I added `@internal` until the bug with multiple snippets expanding/collapsed at once is fixed in Headless.

packages/atomic/cypress/integration/smart-snippet-suggestions.cypress.ts
```diff
@@ -0,0 +1,319 @@
...
+import {SearchBoxSelectors} from './search-box-selectors';
+import {
+  addSmartSnippetSuggestions,
+  addSmartSnippetSuggestionsDefaultOptions,
+} from './smart-snippet-suggestions-actions';
+
+const {remSize, relatedQuestions: defaultRelatedQuestions} =
+  addSmartSnippetSuggestionsDefaultOptions;
+
+describe('Smart Snippet Suggestions Test Suites', () => {
```
btaillon, line 18: I copied most of the tests from the smart snippet.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -59,19 +72,6 @@ export class AtomicSmartSnippet implements InitializableComponent {
    * The [heading level](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) to use for the question at the top of the snippet, from 1 to 5.
    *
    * We recommend setting this property in order to improve accessibility.
-   *
-   * You can style the snippet by inserting a template element like this:
-   * ```html
```
btaillon, line 64: For whatever reason, I put that doc in this prop which has nothing to do with this example. I have no idea how I made that mistake...

packages/atomic/src/locales.json
```diff
@@ -1213,5 +1213,9 @@
   "smart-snippet-feedback-reason-other": {
     "en": "Other",
     "fr": "Autre"
+  },
+  "smart-snippet-people-also-ask": {
+    "en": "People also ask",
+    "fr": "Autres questions posées"
```
btaillon, line 1219: Copied from JSUI

packages/atomic/src/pages/index.html
```diff
@@ -206,6 +206,7 @@
           </atomic-layout-section>
           <atomic-layout-section section="results">
             <atomic-smart-snippet></atomic-smart-snippet>
+            <atomic-smart-snippet-suggestions></atomic-smart-snippet-suggestions>
```
btaillon, line 209: Eventually, that's likely to be moved between the first and second result in `index.html`. I'm likely going to introduce slots in `atomic-result-list` and `atomic-folded-result-list`.

packages/headless/src/controllers/smart-snippet-questions-list/headless-smart-snippet-questions-list.ts
```diff
@@ -74,6 +76,10 @@ export interface SmartSnippetRelatedQuestion {
    * Determines if the snippet is currently expanded.
    */
   expanded: boolean;
+  /**
+   * Provides the source of the smart snippet.
+   */
+  source?: Result;
```
btaillon, line 82: I copied this from the normal question answer controller.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+ *       b {
+ *         color: blue;
+ *       }
+ *     </style>
+ *   </template>
+ * </atomic-smart-snippet-suggestions>
+ * ```
+ *
+ * @part container - The wrapper with a border around the entire component.
+ * @part heading - The heading above the list of related question.
```
jpmarceau, line 40:
```suggestion
 * @part heading - The heading above the list of related questions.
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+  @State()
+  public smartSnippetQuestionsListState!: SmartSnippetQuestionsListState;
+  public error!: Error;
+  @Element() public host!: HTMLElement;
+
+  /**
+   * The [heading level](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) to use for the "People also ask" heading over the snippets, from 1 to 5.
+   *
+   * We recommend setting this property in order to improve accessibility.
+   * ```
```
jpmarceau, line 76:
```suggestion
   * 
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -19,6 +19,19 @@ import {randomID} from '../../utils/utils';
 /**
  * The `atomic-smart-snippet` component displays the excerpt of a document that would be most likely to answer a particular query.
  *
+ * You can style the snippet by inserting a template element like this:
```
jpmarceau, line 22:
```suggestion
 * You can style the snippet by inserting a template element as follows:
```

jpmarceau: (just to be closer to the rest of our documentation)

jpmarceau (approved):
👍 🥇

packages/atomic/cypress/integration/smart-snippet-suggestions-assertions.ts
```diff
@@ -0,0 +1,30 @@
...
+export function assertAnswerBottomMargin(margin: number) {
+  it(`has a ${margin}px gap between the snippet and the footer`, () => {
+    SmartSnippetSuggestionsSelectors.answer()
+      .distanceTo(SmartSnippetSuggestionsSelectors.footer)
+      .should('have.property', 'vertical', margin);
+  });
+}
+
+export function assertlogOpenSmartSnippetSuggestionsSource(log: boolean) {
+  it.skip(`${should(log)} log a openSmartSnippetSource click event`, () => {
```
liviarett, line 22: did you mean to leave this skipped?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+import {LinkWithResultAnalytics} from '../result-link/result-link';
+import {Button} from '../common/button';
+import {randomID} from '../../utils/utils';
+
+/**
+ * The `atomic-smart-snippet-suggestions-suggestions` component displays an accordion of questions related to the query with their corresponding answers.
+ *
+ * You can style the snippets by inserting a template element like this:
+ * ```html
+ * <atomic-smart-snippet-suggestions>
```
liviarett, line 28: what is the difference between this approach and using `snippetStyle`?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
liviarett: how about splitting the parts, so instead of `question-collapsed-text` you could have `question-text collapsed`. this way you could target either both at the same time or them separately, and it's easier to read. Could possibly have less descriptions either and explain about this feature in one place? something like
```
@part question-button - The button to collapse a related question. Also appends 'expanded' or 'collapsed' to its part for more granular control.
```

I wasn't sure if this suggestion would work, so I made a snippet
Screenshot 2022-05-03 at 10 39 44

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
liviarett: I just saw the code above, and I'd like to rephrase the question - why are we allowing users to do both?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+              : 'question-collapsed-icon'
+          }
+          class="w-2.5 mr-3 stroke-[1.25]"
+        ></atomic-icon>
+        <Heading
+          level={this.headingLevel ? this.headingLevel + 1 : 0}
+          class="text-left text-xl font-bold py-4"
+          part={
+            relatedQuestion.expanded
+              ? 'question-expanded-text'
```
liviarett, line 153: especially if you follow my suggestion, I'd create a `makePart` function to do this, like `this.makePart('question-text')`

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+      <Button
+        style="text-neutral"
+        part={
+          relatedQuestion.expanded
+            ? 'question-expanded-button'
+            : 'question-collapsed-button'
+        }
+        onClick={toggleRelatedQuestion}
+        class="flex items-center px-4"
+        ariaExpanded={`${relatedQuestion.expanded}`}
```
liviarett, line 136: out of curiosity, does this not work with the boolean without the back ticks?

liviarett, line 136: out of curiosity, does this not work with the boolean without the back ticks?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+          icon={relatedQuestion.expanded ? ArrowDown : ArrowRight}
+          part={
+            relatedQuestion.expanded
+              ? 'question-expanded-icon'
+              : 'question-collapsed-icon'
+          }
+          class="w-2.5 mr-3 stroke-[1.25]"
+        ></atomic-icon>
+        <Heading
+          level={this.headingLevel ? this.headingLevel + 1 : 0}
```
liviarett, line 149: is this right? haven't looked into `Heading` component, but it seems weird that you would add one to what the person sets it to.. but if this is right, I'd suggest `level={this.headingLevel ? this.headingLevel + 1 : this.headingLevel}` so you can control the default in one place only (you currently set the 0 in two different places)

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,258 @@
...
+      return <Hidden></Hidden>;
+    }
+
+    return (
+      <aside
+        part="container"
+        class="bg-background border border-neutral rounded-lg text-on-background mb-6 overflow-hidden"
+      >
+        <Heading
+          level={this.headingLevel ?? 0}
```
liviarett, line 244: if the default is 0, isn't this check necessary?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
liviarett: ok, I saw the code down there and saw that it's probably just a smaller title, right?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
liviarett: or - may be overkill, but.... - since you have so many ternaries with the expanded, something like `this.ifExpanded('question-text expanded', 'question-text collapsed')`, 
```
    const toggleRelatedQuestion = () =>
      this.ifExpanded(
        this.smartSnippetQuestionsList.collapse(relatedQuestion.documentId),
        this.smartSnippetQuestionsList.expand(relatedQuestion.documentId)
      )
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: oh right, that makes more sense! I'll change it.

packages/atomic/cypress/integration/smart-snippet-suggestions-assertions.ts
btaillon: yes. Unfortunately, it doesn't log the correct click event yet.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: This was copied from SmartSnippet, which works the same way.

The first method (with a template element) is easier to define from a static/vanilla page.

The attribute is easier to define from JSX, and may be the only option in some frameworks.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: Sure! I'll add `this.suffixQuestionPart(relatedQuestion, 'question-text')`

btaillon: I'll extract a method to toggle a related question, like `this.toggleRelatedQuestion(relatedQuestion)`

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: It doesn't, because Button is a functional component which takes a string as a parameter
image

as for the `button` element rendered by it, I'm not sure, I recall it not working so I didn't really try.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: Exactly, it's a smaller title. Sure, I'll replace the ternary.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: It is, you're right. I'll remove it.

lvu285:
Small note, related to ShowMore ShowLess arrow, can you adjust the alignment of the icons?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
ThibodeauJF: yeah all aria-X attribute have to be strings

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,257 @@
...
+        <LinkWithResultAnalytics
+          title={source.title}
+          href={source.clickUri}
+          target="_self"
+          part="source-title"
+          onSelect={() => interactiveResult.select()}
+          onBeginDelayedSelect={() => interactiveResult.beginDelayedSelect()}
+          onCancelPendingSelect={() => interactiveResult.cancelPendingSelect()}
+        >
+          {source.title}
```
ThibodeauJF, line 204: Can you add a todo for Simo here? https://coveord.atlassian.net/browse/KIT-1523

ThibodeauJF: Maybe the footer can be extracted to a functional component as well 🤔

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,257 @@
...
+    });
+    return (
+      <footer
+        part="footer"
+        aria-label={this.bindings.i18n.t('smart-snippet-source')}
+      >
+        <LinkWithResultAnalytics
+          title={source.clickUri}
+          href={source.clickUri}
+          target="_self"
```
olamothe, line 187: Why is the target forced to self ? It's because it's required in LinkWithResultAnalytics ?

If so, we will need to make this configurable for everything that's a result link. That was used quite a lot in JSUI.

For example, people want to configure all links in the page to always open in a new window/tab. And so I don't think it should be a "per component" option, but a "per interface" global setting.

This should be another story though.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -0,0 +1,257 @@
...
+      options: {result: source},
+    });
+    return (
+      <footer
+        part="footer"
+        aria-label={this.bindings.i18n.t('smart-snippet-source')}
+      >
+        <LinkWithResultAnalytics
+          title={source.clickUri}
+          href={source.clickUri}
```
olamothe, line 186: What about the support for `hrefTemplate` ?

olamothe (approved):

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: I copy and pasted this from smart snippet. I think it should be fixed along with highlighting by Simo

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: same as above

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: Updated the doc of the JIRA instead

PR #1955, opened 2022-05-02
feat(atomic): refactored atomic-modal mobile/desktop look to fullscreen true/false prop

btaillon:
https://coveord.atlassian.net/browse/KIT-1587

Prior to this PR, the `atomic-modal` had a "desktop" look (which is what we generally saw for the `atomic-smart-snippet-feedback-modal` component), and it had a mobile look (which is what we always saw for the `atomic-refine-modal` component).

In this PR, a boolean `fullscreen` prop was added to `atomic-modal`. The desktop look was mapped to the value `false` and the mobile look was mapped to the value `true`.

# `atomic-smart-snippet-feedback-modal`
Uses `<atomic-modal fullscreen={false} ...>...</atomic-modal>`

## Desktop
image

## Mobile
image

# `atomic-refine-modal` (mobile)
Uses `<atomic-modal fullscreen={true} ...>...</atomic-modal>`

image

### DIFF
--- a/packages/atomic/src/components/atomic-modal/atomic-modal.pcss
+++ b/packages/atomic/src/components/atomic-modal/atomic-modal.pcss
@@ -13,13 +13,13 @@ atomic-focus-trap {
   grid-area: modal;
 }
 
-:host(.opened) {
+:host(.open) {
   [part='backdrop'] {
     @apply pointer-events-auto;
   }
 }
 
-:host(.opened.dialog) {
+:host(.open.dialog) {
   [part='backdrop'] {
     background-color: rgba(40, 40, 40, 0.8);
   }
--- a/packages/atomic/src/components/atomic-modal/atomic-modal.pcss
+++ b/packages/atomic/src/components/atomic-modal/atomic-modal.pcss
@@ -35,23 +35,12 @@ atomic-focus-trap {
       '. modal .'
       '. .     .';
 
-    grid-template-columns: 1fr auto 1fr;
-    @screen desktop-only {
-      grid-template-rows: 1fr auto 3fr;
-    }
-    @screen mobile-only {
-      grid-template-rows: 0 auto 0;
-    }
+    grid-template-columns: 1fr min(30rem, 100%) 1fr;
+    grid-template-rows: 1fr auto 3fr;
   }
 
   [part='container'] {
     @apply rounded;
-
-    @screen desktop-only {
-      min-width: 30rem;
-      max-width: 100vw;
-      max-height: 100vh;
-    }
   }
 
   [part='header-wrapper'] {
--- a/packages/atomic/src/components/atomic-modal/atomic-modal.pcss
+++ b/packages/atomic/src/components/atomic-modal/atomic-modal.pcss
@@ -40,7 +40,7 @@ atomic-focus-trap {
   }
 
   [part='container'] {
-    @apply rounded;
+    @apply rounded shadow;
   }
 
   [part='header-wrapper'] {
@@ -64,7 +64,6 @@ atomic-focus-trap {
 :host(.fullscreen) {
   [part='container'] {
     @apply absolute inset-0;
-    box-shadow: 0 2px 3px rgba(0, 0, 0, 0.1);
   }
 
   [part='header-wrapper'] {
@@ -76,6 +75,6 @@ atomic-focus-trap {
   }
 
   [part='footer-wrapper'] {
-    @apply shadow-lg px-6 py-4;
+    @apply shadow-t-lg px-6 py-4;
   }
 }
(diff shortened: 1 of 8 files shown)


packages/atomic/src/components/atomic-modal/atomic-modal.pcss
```diff
@@ -2,25 +2,6 @@
```
btaillon, line 2: the diff in this file is all messed up because I went from this layout:
```scss
some-component {
   // some generic style

  desktop {
    // some desktop style
  }

  mobile {
    // some mobile style
  }
}
```
to:
```scss
some-component {
  // some generic style
}

dialog {
  some-component {
    // some non-fullscreen style
  }
}

fullscreen {
  some-component {
    // some fullscreen style
  }
}
```

the only changes are that I added `p-6` to the backdrop and I changed `grid-template-columns` for the backdrop.

packages/atomic/src/components/atomic-modal/atomic-modal.tsx
```diff
@@ -90,55 +105,55 @@ export class AtomicModal implements InitializableComponent {
...
-          this.isOpen ? 'opened' : ''
-        }`}
-        onClick={(e) => e.target === e.currentTarget && this.close()}
-      >
-        <atomic-focus-trap
-          active={this.isOpen}
-          role="dialog"
-          aria-modal={this.isOpen.toString()}
-          aria-labelledby={this.headerId}
+      <Host class={this.getClasses().join(' ')}>
```
btaillon, line 108: I recommend that you ensure that whitespace changes are hidden to review this section, otherwise you will see a huge diff.
image

I moved the `opened` class to the `Host`.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-feedback-modal.pcss
```diff
@@ -1,29 +1,18 @@
 @import '../../global/global.pcss';
 
-[part='reason-title'],
```
btaillon, line 3: I removed the mobile look for the feedback modal (it had expanded buttons and titles) and I increased the gap between the buttons on mobile.

packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -243,6 +243,7 @@ export class AtomicRefineModal implements InitializableComponent {
   public render() {
     return (
       <atomic-modal
+        fullscreen={true}
```
liviarett, line 246: doesn't `<atomic-modal fullscreen` suffice?

packages/atomic/src/components/atomic-modal/atomic-modal.pcss
```diff
@@ -30,53 +11,71 @@ atomic-focus-trap {
...
+  }
 }
 
-[part='body-wrapper'] {
-  @screen mobile-only {
-    @apply px-6 pt-8 pb-5;
+:host(.fullscreen) {
+  [part='container'] {
+    @apply absolute inset-0;
+    box-shadow: 0 2px 3px rgba(0, 0, 0, 0.1);
```
ThibodeauJF, line 67: nothing in tailwind for this?

ThibodeauJF (approved):
Looks nice!

packages/atomic/src/components/atomic-modal/atomic-modal.pcss
btaillon: not really, I can define them in the theme but we're not re-using it and it seems pretty specific to this use case, so I don't see a good reason to do that yet.

olamothe: I'm not sure I understand why we can't just add the class `shadow` or `shadow-md` on the container ?

What do you need to define exactly ? The exact values from the mockup ?

Because if that is so, it's much better to just use tailwind default values, even if it's a bit different than the mockups.

packages/atomic/src/components/atomic-modal/atomic-modal.tsx
```diff
@@ -90,55 +105,55 @@ export class AtomicModal implements InitializableComponent {
...
+        <div
+          part="backdrop"
+          class="fixed left-0 top-0 right-0 bottom-0 z-10"
```
olamothe, line 111: Why not go "all in" on a really huge z-index value ? So that it has more chance to always be on top of whatever UI we're implementing.

olamothe (approved):

packages/atomic/src/components/atomic-modal/atomic-modal.pcss
btaillon: Good point, I'll try a tailwind value instead.

packages/atomic/src/components/atomic-modal/atomic-modal.tsx
btaillon: Sure, will do.

PR #1957, opened 2022-05-03
fix(atomic): graciously handle prop changes before initialization and log warning

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1605

### DIFF
--- a/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
+++ b/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
@@ -135,11 +135,7 @@ export class AtomicSearchInterface {
   @Watch('searchHub')
   @Watch('pipeline')
   public updateSearchConfiguration() {
-    if (!this.engine) {
-      console.warn(
-        'You have to call "initialize" on the atomic-search-interface component before modifying the "searchHub" or "pipeline" props.',
-        this.host
-      );
+    if (!this.engineIsCreated(this.engine)) {
       return;
     }
 
@@ -156,11 +152,7 @@ export class AtomicSearchInterface {
 
   @Watch('analytics')
   public toggleAnalytics() {
-    if (!this.engine) {
-      console.warn(
-        'You have to call "initialize" on the atomic-search-interface component before modifying the "analytics" prop.',
-        this.host
-      );
+    if (!this.engineIsCreated(this.engine)) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -135,10 +135,18 @@ export class AtomicSearchInterface {
   @Watch('searchHub')
   @Watch('pipeline')
   public updateSearchConfiguration() {
+    if (!this.engine) {
```
olamothe, line 138: Could we wrap `this.engine` in a getter that perform that check in that class ? Meaning we won't forget to add those checks in the future when accessing the engine.

And as for the warning, just a generic `You have to call "initialize" on the atomic-search-interface component before modifying any property` could do.

ThibodeauJF: Good idea!

ThibodeauJF: Hmm since engine is a Prop and is used for Atomic React I guess we can't break that interface by renaming "engine"
But I'll had a method

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -275,6 +283,18 @@ export class AtomicSearchInterface {
     this.engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
   }
 
+  private engineIsCreated(engine?: SearchEngine): engine is SearchEngine {
+    if (!engine) {
+      console.error(
```
ThibodeauJF, line 288: This error is much more clear than the previous Headless undefined engine error 😄

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -275,6 +283,18 @@ export class AtomicSearchInterface {
     this.engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
   }
 
+  private engineIsCreated(engine?: SearchEngine): engine is SearchEngine {
```
olamothe, line 286: The `engine` argument is superfluous, the method could just check `if(!this.engine)`

olamothe (approved):

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
ThibodeauJF: I know but the type is better after, dont need `this.engine!`

btaillon (approved):

PR #1962, opened 2022-05-04
feat(atomic): Highlights smart snippet and smart snippet suggestion titles

mbenkirane-coveo:
https://coveord.atlassian.net/browse/KIT-1523

Now highlights smart snippet and smart snippet suggestion titles through the use of "resolveResult" listeners in the atomic-smart-snippet and atomic-smart-snippet suggestions components that propagate data to the atomic-result-text components.


CODEOWNERS
```diff
@@ -1,5 +1,6 @@
 # Default owners for everything in the repo. Later matches take precedence.
-* @coveo/search
+* @coveo/search @mbenkirane-coveo
+ 
 
 # Owners of the Quantic package
 /packages/quantic/ @coveo/service-integrations
```
ThibodeauJF, line 6:
```suggestion
/packages/quantic/ @coveo/service-integration
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -91,6 +93,16 @@ export class AtomicSmartSnippetSuggestions implements InitializableComponent {
...
   @Prop({reflect: true}) snippetStyle?: string;
 
+  @Listen('atomic/resolveResult')
+  public resolveResult(event: ResultContextEvent<Result>) {
+    event.preventDefault();
+    event.stopPropagation();
+
+    this.smartSnippetQuestionsListState.questions.forEach((question) => {
+      if (question.source !== undefined) event.detail(question.source);
+    });
```
ThibodeauJF, line 103: This probably won't work all the time, since it's responding to 1 event multiple times

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -100,6 +102,14 @@ export class AtomicSmartSnippet implements InitializableComponent {
 
   @State() feedbackSent = false;
 
+  @Listen('atomic/resolveResult')
```
ThibodeauJF, line 105: Perfect!

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
btaillon: I think we should listen on each suggestion separately to only provide them with the correct result. This could be done either by using `ref=` on each question or by creating a new component that would wrap the source.

mbenkirane-coveo: Will do !

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,104 @@
...
+import {buildInteractiveResult, Result, SmartSnippet} from '@coveo/headless';
+import {LinkWithResultAnalytics} from '../result-link/result-link';
+import {
+  Bindings,
+  InitializableComponent,
+  InitializeBindings,
+} from '../../utils/initialization-utils';
+import {ResultContextEvent} from '../result-template-components/result-template-decorators';
+
+@Component({
```
ThibodeauJF, line 11: We use a JSDoc tag (internal) so the documentation for this component is not generated in the reference doc https://docs.coveo.com/en/atomic/latest/reference/
```suggestion
/**
 * @internal
 */
@Component({
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,104 @@
...
+        onSelect={() => interactiveResult.select()}
+        onBeginDelayedSelect={() => interactiveResult.beginDelayedSelect()}
+        onCancelPendingSelect={() => interactiveResult.cancelPendingSelect()}
+      >
+        <atomic-result-text
+          field="title"
+          default="no-title"
+        ></atomic-result-text>
+      </LinkWithResultAnalytics>,
+    ];
```
ThibodeauJF, line 102: Yeah... now looking at both I can see they're a bit different... Using a separate component for the `atomic-smart-snippet.tsx` might be overkill, just handling the event directly like it was should be enough.

That means we would only use this new component for `atomic-smart-snippet-suggestions.tsx`

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -222,7 +180,16 @@ export class AtomicSmartSnippetSuggestions implements InitializableComponent {
...
+
+            {relatedQuestion.source ? (
+              <atomic-smart-snippet-source
+                source={relatedQuestion.source}
+                smartSnippet={undefined}
+                isSuggestion={true}
+              ></atomic-smart-snippet-source>
+            ) : (
+              []
+            )}
```
ThibodeauJF, line 192: Alternative pattern common with JSX
```suggestion
            {relatedQuestion.source && (
              <atomic-smart-snippet-source
                source={relatedQuestion.source}
                smartSnippet={undefined}
                isSuggestion={true}
              ></atomic-smart-snippet-source>
            )}
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,104 @@
...
+  public resolveResult(event: ResultContextEvent<Result>) {
+    event.preventDefault();
+    event.stopPropagation();
+    if (this.source) event.detail(this.source);
+  }
+
+  render() {
+    const interactiveResult = buildInteractiveResult(this.bindings.engine!, {
+      options: {result: this.source},
+    });
```
btaillon, line 32: I would build interactiveResult in the `initialize` function, which is called by default by `@InitializeBindings()`.

E.g.:
```ts
public initialize() {
  this.interactiveResult = buildInteractiveResult(...);
}
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
btaillon: actually, I think we should do the opposite: only use it for `smart-snippet`. The problem here is that smart snippet suggestions don't support `select`/`beginDelayedSelect`/`cancelPendingSelect` yet, which they will in the future.

When I add the missing functions to smart snippet suggestions, I will replace the source in smart snippet suggestions to use this component instead.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,104 @@
...
+import {ResultContextEvent} from '../result-template-components/result-template-decorators';
+
+@Component({
+  tag: 'atomic-smart-snippet-source',
+  shadow: false,
+})
+export class AtomicSmartSnippetSource implements InitializableComponent {
+  @InitializeBindings() public bindings!: Bindings;
+  @Prop() source!: Result;
+  @Prop() smartSnippet: SmartSnippet | undefined;
```
btaillon, line 18: I think we should replace the `smartSnippet` prop with either three callback parameters, or preferably three event emitters.

You can copy the event emitter pattern from `atomic-modal` (which emits `animationEnded`) and `atomic-refine-modal` (which listens to it through `onAnimationEnded`).

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,104 @@
...
+} from '../../utils/initialization-utils';
+import {ResultContextEvent} from '../result-template-components/result-template-decorators';
+
+@Component({
+  tag: 'atomic-smart-snippet-source',
+  shadow: false,
+})
+export class AtomicSmartSnippetSource implements InitializableComponent {
+  @InitializeBindings() public bindings!: Bindings;
+  @Prop() source!: Result;
```
btaillon, line 17: We currently use a pattern where all props should have `reflect:true`. Additionally, I expect that some props like `source` and `smartSnippet` should be mutable.

Example:
```suggestion
  @Prop({reflect: true, mutable: true}) source!: Result;
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,104 @@
...
+  @Prop() smartSnippet: SmartSnippet | undefined;
+  @Prop() isSuggestion!: boolean;
```
btaillon, line 19: As discussed in [this thread](https://github.com/coveo/ui-kit/pull/1962/files#r865404099), I would recommend removing this prop only using the component in `smart-snippet` for now. I'll add it in `smart-snippet-suggestions` when the headless controller is ready.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  render() {
+    const interactiveResult = buildInteractiveResult(this.bindings.engine!, {
+      options: {result: this.source},
+    });
+
+    if (this.smartSnippet && !this.isSuggestion) {
+      return (
+        <section aria-label={this.bindings.i18n.t('smart-snippet-source')}>
+          <div part="source-url">
```
btaillon, line 37: I would remove the wrapping divs (`div[part="source-url"]` and `div[part="source-title"]`), and I would add their parts to their respective `LinkWithResultAnalytics` (similarly to what `atomic-smart-snippet-suggestions` does).

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -227,7 +182,15 @@ export class AtomicSmartSnippet implements InitializableComponent {
...
-            {this.renderSource()}
+            {this.smartSnippetState.source ? (
+              <atomic-smart-snippet-source
+                source={this.smartSnippetState.source}
+                smartSnippet={this.smartSnippet}
+                isSuggestion={false}
+              ></atomic-smart-snippet-source>
+            ) : (
+              []
+            )}
```
btaillon, line 193: IMO this should stay inside a `renderSource` function. I don't feel strongly about it though.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -227,7 +182,15 @@ export class AtomicSmartSnippet implements InitializableComponent {
           {this.renderQuestion()}
           {this.renderContent()}
           <footer part="footer">
-            {this.renderSource()}
+            {this.smartSnippetState.source ? (
+              <atomic-smart-snippet-source
+                source={this.smartSnippetState.source}
+                smartSnippet={this.smartSnippet}
+                isSuggestion={false}
```
btaillon, line 189: Parts must be exported here, otherwise they won't be customizable from outside `atomic-smart-snippet`:
```suggestion
                isSuggestion={false}
                exportparts="source-url,source-title"
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
btaillon: One thing I just noticed I forgot: the `@Event()` should have `bubbles: false` like so: `@Event({bubbles: false})`

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
ThibodeauJF: Even without shadow dom?

btaillon: ah, if there's no ShadowDOM, I don't think it's necessary

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
btaillon: edit: I am wrong about `mutable`, I forgot that it only means that the prop is mutable from the inside. You don't need that option.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.pcss
```diff
@@ -6,7 +6,7 @@
   @mixin set-font-size var(--atomic-text-base);
 }
 
-[part='source-title'] a {
+a[part='source-title'] {
```
ThibodeauJF, line 9:
```suggestion
[part='source-title'] {
```
just using the part itself should be ok. Allow users to more easily update the styling

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -227,7 +182,20 @@ export class AtomicSmartSnippet implements InitializableComponent {
...
+                beginDelayedSelectSource={() =>
+                  this.smartSnippet.beginDelayedSelectSource()
+                }
+                cancelPendingSelectSource={() =>
+                  this.smartSnippet.cancelPendingSelectSource()
+                }
+              ></atomic-smart-snippet-source>
+            ) : (
+              []
+            )}
```
ThibodeauJF, line 198:
```suggestion
            {this.smartSnippetState.source && (
              <atomic-smart-snippet-source
                source={this.smartSnippetState.source}
                selectSource={() => this.smartSnippet.selectSource()}
                beginDelayedSelectSource={() =>
                  this.smartSnippet.beginDelayedSelectSource()
                }
                cancelPendingSelectSource={() =>
                  this.smartSnippet.cancelPendingSelectSource()
                }
              ></atomic-smart-snippet-source>
            )}
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -16,9 +16,9 @@ import {
 } from '@coveo/headless';
 import {Hidden} from '../common/hidden';
 import {Heading} from '../common/heading';
-import {LinkWithResultAnalytics} from '../result-link/result-link';
 import {Button} from '../common/button';
 import {randomID} from '../../utils/utils';
+import {LinkWithResultAnalytics} from '../result-link/result-link';
```
ThibodeauJF, line 21: handling of that component is still in progress? or will it be in an upcoming PR?

mbenkirane-coveo: Benjamin will be handling it in an upcoming PR using the new atomic-smart-snippet-source component

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,72 @@
...
+  tag: 'atomic-smart-snippet-source',
+  shadow: false,
+})
+export class AtomicSmartSnippetSource implements InitializableComponent {
+  @InitializeBindings() public bindings!: Bindings;
+  @Prop({reflect: true, mutable: true}) source!: Result;
+
+  @Prop() selectSource!: () => void;
+  @Prop() beginDelayedSelectSource!: () => void;
+  @Prop() cancelPendingSelectSource!: () => void;
```
btaillon, line 26: Do you mean for those to be events or is this intentional?

```suggestion
  @Event() selectSource!: EventEmitter;
  @Event() beginDelayedSelectSource!: EventEmitter;
  @Event() cancelPendingSelectSource!: EventEmitter;
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <section aria-label={this.bindings.i18n.t('smart-snippet-source')}>
+        <LinkWithResultAnalytics
+          title={this.source.clickUri}
+          href={this.source.clickUri}
+          target="_self"
+          className="block"
+          part="source-url"
+          onSelect={this.selectSource}
+          onBeginDelayedSelect={() => this.beginDelayedSelectSource}
+          onCancelPendingSelect={() => this.cancelPendingSelectSource}
```
btaillon, line 50: Looks like you made a typo here, I'm not sure all of those functions are being executed. We probably need more cypress tests to verify that.

I think you mean:
```suggestion
          onSelect={() => this.selectSource()}
          onBeginDelayedSelect={() => this.beginDelayedSelectSource()}
          onCancelPendingSelect={() => this.cancelPendingSelectSource()}
```
or, for events:
```suggestion
          onSelect={() => this.selectSource.emit()}
          onBeginDelayedSelect={() => this.beginDelayedSelectSource.emit()}
          onCancelPendingSelect={() => this.cancelPendingSelectSource.emit()}
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -0,0 +1,72 @@
...
+        </LinkWithResultAnalytics>
+        <LinkWithResultAnalytics
+          title={this.source.title}
+          href={this.source.clickUri}
+          target="_self"
+          className="block mb-6"
+          part="source-title"
+          onSelect={this.selectSource}
+          onBeginDelayedSelect={() => this.beginDelayedSelectSource}
+          onCancelPendingSelect={() => this.cancelPendingSelectSource}
```
btaillon, line 62: Same as above:
```suggestion
          onSelect={() => this.selectSource()}
          onBeginDelayedSelect={() => this.beginDelayedSelectSource()}
          onCancelPendingSelect={() => this.cancelPendingSelectSource()}
```
or, for events:
```suggestion
          onSelect={() => this.selectSource.emit()}
          onBeginDelayedSelect={() => this.beginDelayedSelectSource.emit()}
          onCancelPendingSelect={() => this.cancelPendingSelectSource.emit()}
```

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -227,7 +182,18 @@ export class AtomicSmartSnippet implements InitializableComponent {
...
+            {this.smartSnippetState.source && (
+              <atomic-smart-snippet-source
+                source={this.smartSnippetState.source}
+                selectSource={() => this.smartSnippet.selectSource()}
+                beginDelayedSelectSource={() =>
+                  this.smartSnippet.beginDelayedSelectSource()
+                }
+                cancelPendingSelectSource={() =>
+                  this.smartSnippet.cancelPendingSelectSource()
+                }
```
btaillon, line 194: If you decide to use events, these will be automatically renamed. Otherwise, looks good!
```suggestion
                onSelectSource={() => this.smartSnippet.selectSource()}
                onBeginDelayedSelectSource={() =>
                  this.smartSnippet.beginDelayedSelectSource()
                }
                onCancelPendingSelectSource={() =>
                  this.smartSnippet.cancelPendingSelectSource()
                }
```

btaillon (approved):
looks good! just gotta fix these callbacks since I doubt they're getting called

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
mbenkirane-coveo: oh yes, I'll fix those ! I think this.selectSource and () => this.selectSource() are equivalent from my understanding. I'll switch to () => this.selectSource() if that's the standard.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-source.tsx
ThibodeauJF: Since it's an internal component & it's a direct child, using props is fine I think, like if it was a functional component

PR #1963, opened 2022-05-05
fix(atomic): fixed smart snippet show more/show less issue when rotating screen

btaillon:
https://coveord.atlassian.net/browse/KIT-1604

The "show more" button in smart snippet only appears when there's not enough room vertically to display the snippet. 

Prior to this PR, the height of the content of the snippet was only evaluated when the snippet was initially rendered. However, the height varies when the size of the screen changes, such as when a mobile device's orientation is changed.

### DIFF
--- a/packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
+++ b/packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
@@ -1,4 +1,12 @@
-import {Component, h, Prop, Event, EventEmitter, Host} from '@stencil/core';
+import {
+  Component,
+  h,
+  Prop,
+  Event,
+  EventEmitter,
+  Host,
+  Listen,
+} from '@stencil/core';
 import {sanitize} from 'dompurify';
 import {sanitizeStyle} from '../../utils/utils';
 
@@ -15,14 +23,23 @@ export class AtomicSmartSnippetAnswer {
   @Prop({reflect: true}) htmlContent!: string;
   @Prop({reflect: true}) innerStyle?: string;
 
-  @Event({
-    eventName: 'atomic/smartSnippet/answerRendered',
-  })
+  @Event({bubbles: false})
   private answerRendered!: EventEmitter<{height: number}>;
   private wrapperElement?: HTMLElement;
+  private isRendering = true;
+
+  public componentWillRender() {
+    this.isRendering = true;
+  }
 
   public componentDidRender() {
-    this.answerRendered.emit({height: this.wrapperElement!.scrollHeight});
+    this.isRendering = false;
+    this.emitCurrentHeight();
+  }
+
+  @Listen('resize', {target: 'window'})
+  public windowResized() {
+    this.emitCurrentHeight();
   }
 
   private get sanitizedStyle() {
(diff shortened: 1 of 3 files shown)


packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
```diff
@@ -15,14 +15,38 @@ export class AtomicSmartSnippetAnswer {
...
 
-  @Event({
-    eventName: 'atomic/smartSnippet/answerRendered',
-  })
-  private answerRendered!: EventEmitter<{height: number}>;
+  @Event({bubbles: false})
+  private answerSizeUpdated!: EventEmitter<{height: number}>;
   private wrapperElement?: HTMLElement;
+  private isRendering = true;
+  private resizeObserver: ResizeObserver | undefined;
```
btaillon, line 22: I decided to use a `ResizeObserver` instead of the `window.onresize` event or the `ScreenOrientation.onchange` event since technically the size of the smart snippet could also change if a parent element is resized, which I imagine could easily be the case in some clients' cases.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
```diff
@@ -15,14 +15,38 @@ export class AtomicSmartSnippetAnswer {
...
+
+  public componentWillRender() {
+    this.isRendering = true;
+  }
 
   public componentDidRender() {
-    this.answerRendered.emit({height: this.wrapperElement!.scrollHeight});
+    this.isRendering = false;
+    this.emitCurrentHeight();
```
btaillon, line 30: In practice, I saw that the resize observer what calling its callback immediately on observe, which means that this call to `emitCurrentHeight` is redundant.

However, I haven't found anything in the spec that implies that the callback *must* be called immediately on observe. I prefer to keep this call here to avoid variations between different browsers that may interpret the spec differently.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
```diff
@@ -15,14 +15,38 @@ export class AtomicSmartSnippetAnswer {
...
+  }
+
+  public connectedCallback() {
+    if (this.wrapperElement && this.resizeObserver) {
+      this.resizeObserver.observe(this.wrapperElement);
+    }
+  }
+
+  public disconnectedCallback() {
+    this.resizeObserver?.disconnect();
```
btaillon, line 40: I saw a mention on stackoverflow of somebody getting a memory leak due to forgetting to call `disconnect`.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
```diff
@@ -32,6 +56,13 @@ export class AtomicSmartSnippetAnswer {
     return sanitizeStyle(this.innerStyle);
   }
 
+  private emitCurrentHeight() {
+    if (this.isRendering) {
```
olamothe, line 60: What is this check for (as well as the flag that accompanies it) ?

What would happen if it's removed ? Infinite rendering loop, or just some minor inefficient rendering loop ?

btaillon: Inefficient. I recall, when initially making the `collapsible-answer` component, that for a moment before `componentDidRender`, the element had the wrong height. This is to prevent that height from rendering first.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
ThibodeauJF: yes it's important

ThibodeauJF (approved):

PR #1966, opened 2022-05-05
feat(atomic): added show more/show less to smart snippet suggestions

btaillon:
https://coveord.atlassian.net/browse/KIT-1595

Most of the doc and tests were copy and pasted from the smart snippet, then modified slightly to fit with the suggestions use case.

This required some relatively big changes to the `atomic-smart-snippet-expandable-answer` component. Prior to this PR, it took very few parameters and it initialized its own smart snippet controller. After this PR, it receives the answer & expanded state as a prop and emits an `expand` and `collapse` event.

Screenshot (with maximumHeight set to 120px and collapsedHeight to 80px):
image


jpmarceau (approved):
💯 🏅

PR #1967, opened 2022-05-06
fix(atomic): made smart snippet wait for answer to be rendered

btaillon:
https://coveord.atlassian.net/browse/KIT-1525

The answer part of the snippet takes a bit longer to render than the rest (about 120ms more on my laptop). Originally, I assumed this was due to it rendering twice in order to estimate the height of the snippet, but adding logic to only render once did not fix the issue, so I stashed that fix. My guess is that setting `innerHTML` causes the delay.

My solution was to wait for `componentDidRender` to be called before revealing the smart snippet component. I expected the `componentDidRender` lifecycle event to be called too early due to the above issue with the answer rendering twice, but that didn't turn out to be the case.

### DIFF
--- a/packages/atomic/src/utils/initialization-utils.tsx
+++ b/packages/atomic/src/utils/initialization-utils.tsx
@@ -290,4 +290,26 @@ export function DeferUntilRender() {
   };
 }
 
+export function RemainHiddenUntilDescendantsRendered() {
+  return (component: ComponentInterface, methodName: string) => {
+    const {componentDidRender} = component;
+
+    component[methodName] = function () {
+      const element = getElement(this);
+      if (!element.classList.contains('atomic-hidden')) {
+        return;
+      }
+      element.style.visibility = 'hidden';
+      element.style.position = 'absolute';
+    };
+
+    component.componentDidRender = function () {
+      const element = getElement(this);
+      element.style.visibility = '';
+      element.style.position = '';
+      return componentDidRender && componentDidRender.call(this);
+    };
+  };
+}
+
 export type I18nState = Record<string, (variables?: TOptions) => string>;
--- a/packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
+++ b/packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
@@ -4,6 +4,7 @@ import {
   InitializeBindings,
   Bindings,
   BindStateToController,
+  RemainHiddenUntilDescendantsRendered,
 } from '../../utils/initialization-utils';
 import {
   buildSmartSnippet,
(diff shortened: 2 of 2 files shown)


ThibodeauJF (approved):
Cypress test would be appreciated for these tricky decorators

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -290,4 +290,26 @@ export function DeferUntilRender() {
...
+    component[methodName] = function () {
+      const element = getElement(this);
+      if (!element.classList.contains('atomic-hidden')) {
+        return;
+      }
+      element.style.visibility = 'hidden';
+      element.style.position = 'absolute';
+    };
+
+    component.componentDidRender = function () {
```
liviarett, line 306: with these decorators, I'm just wondering if setting lifecycle methods like this isn't dangerous, and it might override stuff defined in the component or even in other decorators?
or is that what `return componentDidRender && componentDidRender.call(this);` does?

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -290,4 +290,26 @@ export function DeferUntilRender() {
   };
 }
 
+export function RemainHiddenUntilDescendantsRendered() {
+  return (component: ComponentInterface, methodName: string) => {
+    const {componentDidRender} = component;
+
+    component[methodName] = function () {
+      const element = getElement(this);
+      if (!element.classList.contains('atomic-hidden')) {
```
liviarett, line 299: I'm a little confused, couldnt this be achieved with css only?

packages/atomic/src/utils/initialization-utils.tsx
btaillon: > or is that what return componentDidRender && componentDidRender.call(this); does?

Yes, it "overrides" them but then calls the original. The biggest risk is that the order in which these are applied is ambiguous.

packages/atomic/src/utils/initialization-utils.tsx
btaillon: No, because the element loses `atomic-hidden`. We're basically extending the life of `atomic-hidden`, but using `visibility: hidden` instead so we can estimate the size of children.

btaillon:
> Cypress test would be appreciated for these tricky decorators

I wanted to add cypress tests, but I did a lot of exploration and so far it seems too complicated for what it's worth.

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -290,4 +290,26 @@ export function DeferUntilRender() {
   };
 }
 
+export function RemainHiddenUntilDescendantsRendered() {
```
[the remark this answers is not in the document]
btaillon: The reason was primarily because I expected some other components may need it too in the future, but now I think i prefer moving this inside the component. We can make it a decorator later if we need to re-use it.

PR #1970, opened 2022-05-09
feat(quantic): Creation of the Quantic Refine Toggle component

mmitiche:
[SFINT-4470](https://coveord.atlassian.net/browse/SFINT-4470)
The `QuanticRefineToggle` component displays a button that is used to open the refine modal.

https://user-images.githubusercontent.com/86681870/168607601-afc74e7a-921c-4b73-aed4-bc1cf3cebc87.mov


packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.html
```diff
@@ -0,0 +1,41 @@
+<template>
+  <button class="slds-button slds-button_neutral refine-button slds-is-relative" onclick={openModal}
+    title={buttonLabel}>
+    <template if:true={hasButtonLabel}> {buttonLabel} </template>
```
lbergeron, line 4:
```suggestion
    <template if:true={hasButtonLabel}>{buttonLabel}</template>
```

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.html
```diff
@@ -0,0 +1,41 @@
...
+      slot="header">
+      <slot name="refine-title">
+        <div class="slds-text-title_bold">{labels.sortAndFilters}</div>
+      </slot>
+      <lightning-button class="slds-is-absolute slds-align_absolute-center slds-var-p-right_large refine-modal__action"
+        onclick={closeModal} variant="base" title="Close" icon-name="utility:close">
+      </lightning-button>
+    </div>
+    <div class="slds-align_absolute-center" slot="content">
+      <div class="refine-modal__content">
```
lbergeron, line 29: Out of curiosity, why did you put this element (with the `refine-modal__content` class) inside the slot? If custom content is inserted without specifying the class, it could break the styling, no?

Could you put the `slot` attribute on this element instead?

```suggestion
      <div class="refine-modal__content" slot="content">
```

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.html
```diff
@@ -0,0 +1,41 @@
+<template>
+  <button class="slds-button slds-button_neutral refine-button slds-is-relative" onclick={openModal}
```
lbergeron, line 2: Instead of having attributes to customize and enable the button text and icon, why don't you define a slot for the button content instead? This way, anyone could customize the button as needed. We might want to add a property to allow customizing the button CSS classes though.

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -0,0 +1,207 @@
...
+      'numericFacetBreadcrumbs',
+      'categoryFacetBreadcrumbs',
+      'dateFacetBreadcrumbs',
+      'staticFilterBreadcrumbs',
+    ];
+    return facetBreadcrumbKeys.reduce(
+      (count, facetKey) =>
+        count +
+        this.getFiltersCountFromFacetBreadcrumb(
+          this.breadcrumbManager.state[facetKey]
```
lbergeron, line 140: If, for some reason, some keys are not defined in the state, the running total would become `NaN`. You should include a fallback to add `0` in such case.

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.html
mmitiche: I tried to follow the same logic we currently  using with the Quickview button but having a slot for the button content is indeed a better idea. I'll try to implement this logic 👌

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
mmitiche: sure 👌

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -0,0 +1,207 @@
...
+  openModal() {
+    /** @type {QuanticModalElement} */
+    const modal = this.template.querySelector(`[data-id=${this.modalId}]`);
+    modal.openModal();
+  }
+
+  /**
+   * Closes the refine modal.
+   * @returns {void}
+   *
```
nathanlb, line 170: extra line here

nathanlb (commented):
Pretty awesome

packages/quantic/force-app/main/default/lwc/exampleSearch/exampleSearch.css
```diff
@@ -4,3 +4,24 @@
...
 }
+
+.refine-toggle_container {
+  display: none;
+  width: 100%;
+  justify-content: flex-end;
+}
+
+
+@media only screen and (max-width: 1023px) {
```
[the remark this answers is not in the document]
mmitiche: Absolutely  💯 I just updated this!

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -0,0 +1,207 @@
...
+    this.unsubscribeQuerySummary?.();
+    this.unsubscribeBreadcrumbManager?.();
+  }
+
+  /**
+   * Updates total number of results.
+   * @returns {void}
+   */
+  updateTotalResults() {
+    this.total = this.querySummary.state.total;
```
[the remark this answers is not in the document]
mmitiche: Because with this updateTotalResults method, I'm always listening to the state changes of the Query summary state and updating the right value of `total`, not sure if with a getter that returns this.querySummary.state.total, the component will re-render when the `total` value in the state changes 🧐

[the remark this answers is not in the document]
nathanlb: getters are reactive in LWC afaik but you make valid points so keep as is 👍

nathanlb (approved):

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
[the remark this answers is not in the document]
lbergeron: Yes getters are reactive, but it means LWC would need to track the whole QuerySummary state, which I don't think it would do automatically. That's the reason why the `@track` annotation has to be added when the attribute is an object or an array.

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.css
```diff
@@ -0,0 +1,24 @@
...
+  width: 100%;
+  max-width: 40rem;
+}
+
+.refine-button_filters-badge {
+  background: var(--lwc-brandAccessible, #0176d3);
+  color: var(--lwc-colorBackgroundAlt, #ffffff);
+  border-radius: 50%;
+  top: -.5rem;
+  right: -.5rem;
```
erocheleau, line 20: Why do we need negative stuff?

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -0,0 +1,182 @@
...
+      'numericFacetBreadcrumbs',
+      'categoryFacetBreadcrumbs',
+      'dateFacetBreadcrumbs',
+      'staticFilterBreadcrumbs',
+    ];
+    return facetBreadcrumbKeys.reduce(
+      (count, facetKey) =>
+        count +
+        this.getFiltersCountFromFacetBreadcrumb(
+          this.breadcrumbManager?.state?.[facetKey]
```
erocheleau, line 133: I think @lbergeron already mentioned it, but make sure this is fine because in JS, adding a number and `undefined` will result in `NaN`.

```
let count = 0;
count = count + undefined;
// NaN
```

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -0,0 +1,182 @@
...
+    );
+  }
+
+  /**
+   * Opens the refine modal.
+   * @returns {void}
+   */
+  openModal() {
+    /** @type {QuanticModalElement} */
+    const modal = this.template.querySelector(`[data-id=${this.modalId}]`);
```
erocheleau, line 159: Might want to have a `this.getModal()` function?

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.css
mmitiche: We need it for the positioning of the currently active filters badge:
Badge

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
mmitiche: sure 👌

PR #1971, opened 2022-05-09
fix(atomic): made smart snippet wait for answer to be rendered

btaillon:
https://coveord.atlassian.net/browse/KIT-1525

Follow-up to https://github.com/coveo/ui-kit/pull/1967

The answer part of the snippet takes a bit longer to render than the rest (about 120ms more on my laptop).

In the previous PR, my solution was only to keep the `atomic-smart-snippet` element hidden longer until `componentDidRender` is called. However, I noticed a new more subtle issue while debugging, which is that the "show more" button was showing up and then getting removed (or the other way around, I don't remember). The fix for that issue requires evaluating the height of the snippet before actually rendering the component.


packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-expandable-answer.tsx
```diff
@@ -64,28 +65,52 @@ export class AtomicSmartSnippetExpandableAnswer {
...
+      );
+      answerElement.htmlContent = this.smartSnippetState.answer;
+      answerElement.innerStyle = this.snippetStyle;
+      answerElement.style.visibility = 'hidden';
+      answerElement.style.position = 'absolute';
+      listenOnce(answerElement, 'answerSizeUpdated', (event) => {
+        answerElement.remove();
+        resolve((event as CustomEvent<{height: number}>).detail.height);
+      });
+      this.host.parentElement!.appendChild(answerElement);
```
btaillon, line 90: By appending the temporary `answerElement` to the parent of the host, we ensure that it inherits the same part styling, which affects the height.

packages/atomic/src/utils/utils.ts
```diff
@@ -16,6 +16,33 @@ export function once<T extends unknown[]>(fn: (...args: T) => unknown) {
...
+  type: K,
+  listener: (this: HTMLElement, ev: HTMLElementEventMap[K]) => any,
+  options?: boolean | AddEventListenerOptions
+): void;
+export function listenOnce(
+  element: HTMLElement,
+  type: string,
+  listener: EventListenerOrEventListenerObject,
+  options?: boolean | AddEventListenerOptions
+): void;
```
btaillon, line 30: copied from `addEventListener`

ThibodeauJF: File is getting a lot of variety, maybe this should be moved to "event-utils"

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -121,6 +121,11 @@ export class AtomicSmartSnippet implements InitializableComponent {
     return styleTag.innerHTML;
   }
 
+  private set isRenderingFromHidden(shouldHide: boolean) {
```
ThibodeauJF, line 124: method name is unclear

ThibodeauJF (approved):

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
btaillon: I'll try `hideDuringRender`

packages/atomic/src/utils/event-utils.ts
```diff
@@ -9,3 +9,30 @@ export function buildCustomEvent<T>(name: string, detail: T) {
     composed: true,
   });
 }
+
+export function listenOnce<K extends keyof HTMLElementEventMap>(
+  element: HTMLElement,
+  type: K,
+  listener: (this: HTMLElement, ev: HTMLElementEventMap[K]) => any,
```
liviarett, line 16: could we change this to something else to avoid the `any` warning?

packages/atomic/src/utils/event-utils.ts
```diff
@@ -9,3 +9,30 @@ export function buildCustomEvent<T>(name: string, detail: T) {
...
+  options?: boolean | AddEventListenerOptions
+): void;
+export function listenOnce(
+  element: HTMLElement,
+  type: string,
+  listener: EventListenerOrEventListenerObject,
+  options?: boolean | AddEventListenerOptions
+): void;
+export function listenOnce(
```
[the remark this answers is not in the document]
btaillon: This is a Typescript feature called [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads).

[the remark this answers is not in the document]
liviarett: very nice 👏

PR #1972, opened 2022-05-10
fix(atomic): send correct result to rendering func and assign it at r…

liviarett:
…ight time

https://coveord.atlassian.net/browse/KIT-1621

### DIFF
--- a/packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
+++ b/packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
@@ -55,6 +55,8 @@ export class AtomicFoldedResultList implements InitializableComponent {
   @State() public templateHasError = false;
 
   public resultListCommon!: ResultListCommon;
+  private renderingFunction: ((res: FoldedResult) => HTMLElement) | null = null;
+
   /**
    * Whether to automatically retrieve an additional page of results and append it to the
    * current results when the user scrolls down to the bottom of element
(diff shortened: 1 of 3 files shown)


packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -138,6 +141,7 @@ export class AtomicFoldedResultList implements InitializableComponent {
       templateElements: this.host.querySelectorAll('atomic-result-template'),
       onReady: () => {
         this.ready = true;
+        this.assignRenderingFunctionIfPossible();
```
ThibodeauJF, line 144: why is this change necessary? it loses the rendering method reference when scrolling?

Might be better to improve the solution with a slightly different architecture rather than have this kind of patch

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -132,7 +132,9 @@ export class ResultListCommon {
     resultOrFolded: Result | FoldedResult
   ): HTMLElement | DocumentFragment {
     const result = (resultOrFolded as FoldedResult).result || resultOrFolded;
-    return this.render ? this.render(result) : this.getTemplate(result)!;
+    return this.render
+      ? this.render(resultOrFolded)
```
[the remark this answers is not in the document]
liviarett: because `result` will always be of type `Result`, i.e `resultOrFolded.result`, and I want the whole `FoldedResult` when available

packages/atomic/src/components/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
liviarett: I copied what @olamothe did for the other result list - it's not related to scrolling, it's related to trying to set rendering function before `initialize` happens. currently throws an error saying that `resultListCommon` doesn't exist

olamothe: Yeah, it's mostly about not having to deal with "are components properly initialized yet?" in Atomic react code, because that's kind of tricky to deal with.

So you can assign the render function at any time, and it will be picked up by the Atomic result list when it's ready to do its thing.

olamothe (approved):

PR #1973, opened 2022-05-10
docs(atomic-react): add examples for folding component

liviarett:
https://coveord.atlassian.net/browse/KIT-1622


packages/samples/atomic-react/src/App.tsx
```diff
@@ -1,218 +1,40 @@
...
+          <HeaderLink
+            page={FOLDED_CUSTOM_CHILDREN_PAGE}
+            currentPage={page}
+            setPage={setPage}
+          />
+        </ul>
+      </header>
+      {page === LIST_PAGE && <List />}
+      {page === FOLDED_LIST_PAGE && <Folded />}
+      {page === FOLDED_CUSTOM_CHILDREN_PAGE && <CustomChildren />}
```
ThibodeauJF, line 35: Is the main difference between these pages the result list? If so having multiple fully fledged apps makes it harder to maintain VS having something like

```
<AtomicApp>
   <AtomicFoldedResultList ...  />
 </AtomicApp>
```

Which would render the slotted result list component in the right spot of the app.

liviarett: yeah, I actually also changed the engine, but you're right. let me suggest something

packages/samples/atomic-react/src/App.tsx
```diff
@@ -1,218 +1,40 @@
-import React, {FunctionComponent} from 'react';
+import React, {FunctionComponent, useState} from 'react';
```
olamothe, line 1: It was there before your PR, but if you can find the magic recipe for the linter config to suppress that warning, it would be awesome.

packages/samples/atomic-react/src/App.tsx
```diff
@@ -1,218 +1,40 @@
...
+            page={FOLDED_CUSTOM_CHILDREN_PAGE}
+            currentPage={page}
+            setPage={setPage}
+          />
+        </ul>
+      </header>
+      {page === LIST_PAGE && <List />}
+      {page === FOLDED_LIST_PAGE && <Folded />}
+      {page === FOLDED_CUSTOM_CHILDREN_PAGE && <CustomChildren />}
     </>
```
btaillon, line 36: When just looking at the file names, I find it difficult to figure out which components are pages.

I think it would be useful to rename them so their names contain `Page` (e.g., `ResultListPage`, `FoldedResultListPage`, `CustomFoldedResultListPage`).

Additionally, I think it would be useful to nest all three components in a `src/pages` folder.

packages/samples/atomic-react/src/HeaderLink.tsx
```diff
@@ -0,0 +1,23 @@
+import React, {FunctionComponent} from 'react';
+
+type Props = {
+  page: string;
+  currentPage: string;
+  setPage(page: string): void;
+};
+export const HeaderLink: FunctionComponent<Props> = ({
```
btaillon, line 8: I would nest this component in a directory like `src/utils` to avoid too much clutter at the base `src` directory, where developers may first look.

packages/samples/atomic-react/src/App.tsx
liviarett: you're right! I didn't give it too much thought because I thought it was more for us to test, but I guess external people might use it too.. and also we should keep it tidy anyway! will follow your suggestions

PR #1977, opened 2022-05-10
feat(atomic,headless): add more out of the box analytics configuration for atomic

olamothe:
Most notably `documentLocation`, `documentReferrer`, which Headless cannot determine itself (needs to be in a browser context).

Added auto resolving of Facet title (Headless does not contain the notion of facet label: it's purely a UI option).

Some minor changes in Headless for analytics parameters: Instead of sending an empty string when a value is un-available, simply don't send anything.

https://coveord.atlassian.net/browse/KIT-1623

### DIFF
--- a/packages/atomic/src/components/atomic-search-interface/analytics-config.spec.ts
+++ b/packages/atomic/src/components/atomic-search-interface/analytics-config.spec.ts
@@ -1,8 +1,8 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import {
   getSampleSearchEngineConfiguration,
   SearchEngineConfiguration,
 } from '@coveo/headless';
-import {registerFacet} from '@coveo/headless/dist/definitions/features/facets/facet-set/facet-set-actions';
 import {createStore, ObservableMap} from '@stencil/store';
 import {
   AtomicStore,
--- a/packages/atomic/src/components/atomic-search-interface/analytics-config.spec.ts
+++ b/packages/atomic/src/components/atomic-search-interface/analytics-config.spec.ts
@@ -1,0 +1,108 @@
+import {
+  getSampleSearchEngineConfiguration,
+  SearchEngineConfiguration,
+} from '@coveo/headless';
+import {registerFacet} from '@coveo/headless/dist/definitions/features/facets/facet-set/facet-set-actions';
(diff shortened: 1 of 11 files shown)


ThibodeauJF (approved):

packages/atomic/src/components/atomic-search-interface/analytics-config.spec.ts
```diff
@@ -0,0 +1,108 @@
+import {
```
liviarett, line 1: there are quite a few warnings in this file, could they be avoided?

packages/atomic/src/components/atomic-search-interface/analytics-config.ts
```diff
@@ -4,40 +4,99 @@ import {
...
-  enabled: boolean
+  enabled: boolean,
+  store: ObservableMap<AtomicStore>
 ): AnalyticsConfiguration {
+  const analyticsClientMiddleware = (
+    event: string,
+    payload: AnalyticsPayload
+  ) => augmentAnalytics(event, payload, store, searchEngineConfig);
+
+  const outOfTheBoxConfig: AnalyticsConfiguration = {
```
liviarett, line 22: do you guys always use 'out of the box' in names, or would it be better to change to `default`?

liviarett (approved):

packages/atomic/src/components/atomic-search-interface/analytics-config.ts
btaillon: I would also prefer `default` or `fallback`.

btaillon (approved):

packages/atomic/src/components/atomic-search-interface/analytics-config.spec.ts
olamothe: Well add an eslint disable for `any`. Unfortunately, it's needed in this file. The types coming from `coveo.analytics` for the analytics middleware uses `any`.

packages/atomic/src/components/atomic-search-interface/analytics-config.ts
liviarett: yeah sorry, I meant replacing `outOfTheBox` with `default`, so `defaultConfig`

PR #1981, opened 2022-05-11
feat(atomic): network improvements

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1620

First set of performance improvements

1. Start loading i18n configured language sooner, before `initialize` is called
2. Only load configured language for i18n, don't load fallback language (e.g. if "fr", "en.json" was still loaded)
3. Remove lodash imports (should only be used as a dev dep), saves ~97kb
4. Use loading="lazy" on `img` tags

### DIFF
--- a/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
+++ b/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
@@ -342,7 +342,6 @@ export class AtomicSearchInterface {
       lng: this.language,
       fallbackLng: false,
       backend: this.i18nBackendOptions,
-      initImmediate: false,
     });
   }
 
@@ -446,9 +445,9 @@ export class AtomicSearchInterface {
       );
       return;
     }
-    await this.i18nPromise;
     this.updateIconAssetsPath();
     initEngine();
+    await this.i18nPromise;
     this.initComponents();
     this.initSearchStatus();
     this.initUrlManager();
--- a/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
+++ b/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
@@ -337,10 +337,11 @@ export class AtomicSearchInterface {
   }
 
   private initI18n() {
+    const isLanguageComplete = ['en', 'fr'].includes(this.language);
     return this.i18n.use(Backend).init({
       debug: this.logLevel === 'debug',
       lng: this.language,
-      fallbackLng: false,
+      fallbackLng: isLanguageComplete ? false : 'en',
       backend: this.i18nBackendOptions,
     });
   }
--- a/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
+++ b/packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
@@ -23,7 +23,7 @@ import {
   loadQueryActions,
 } from '@coveo/headless';
 import {Bindings, InitializeEvent} from '../../utils/initialization-utils';
-import i18next, {i18n} from 'i18next';
+import i18next, {i18n, TFunction} from 'i18next';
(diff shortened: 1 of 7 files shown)


packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -166,19 +171,17 @@ export class AtomicSearchInterface {
 
   @Watch('language')
   public updateLanguage() {
-    if (!this.engineIsCreated(this.engine)) {
-      return;
+    if (this.engineIsCreated(this.engine)) {
```
olamothe, line 174: Why the change in this function ? I personally like the early return pattern, helps readability.

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -235,7 +238,7 @@ export class AtomicSearchInterface {
   /**
    * Initializes the connection with the headless search engine using options for `accessToken` (required), `organizationId` (required), `renewAccessToken`, and `platformUrl`.
    */
-  @Method() public initialize(options: InitializationOptions) {
+  @Method() public async initialize(options: InitializationOptions) {
```
olamothe, line 241: Why the async keyword if we do not use await in the body of the function ? 

It was returning a promise anyway before, no ?

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -167,7 +166,9 @@ export class ResultListCommon {
     const templates = (
       includeDefaultTemplate ? [this.makeDefaultTemplate()] : []
     ).concat(
-      customTemplates.filter(identity) as ResultTemplate<DocumentFragment>[]
+      customTemplates.filter(
```
olamothe, line 169: What is this change for ?

packages/atomic/src/components/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -53,6 +53,12 @@ export class AtomicResultImage implements InitializableComponent {
       return;
     }
 
-    return <img alt={`${this.field} image`} src={filterProtocol(url)} />;
+    return (
+      <img
+        alt={`${this.field} image`}
+        src={filterProtocol(url)}
+        loading="lazy"
```
olamothe, line 60: Great idea !

olamothe (approved):
GG !

Any idea of much loading time this shaves off ? Relatively speaking (%)

ThibodeauJF:
> Any idea of much loading time this shaves off ? Relatively speaking (%)

Depends on the connection (removing lodash improves a lot on 3G) and when the user trigger initialize, then the i18n changes have a slight difference.

On a very good connection, these network improvements don't do much. More improvements to come...

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -337,7 +340,7 @@ export class AtomicSearchInterface {
     return this.i18n.use(Backend).init({
       debug: this.logLevel === 'debug',
       lng: this.language,
-      fallbackLng: ['en'],
+      fallbackLng: false,
```
btaillon, line 343: If some strings are missing from the language set by the user, are english strings going to be used?

ThibodeauJF: Good point! let me check...

btaillon (approved):

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
ThibodeauJF: This indeed should be an fr/en improvement

PR #1982, opened 2022-05-11
feat(atomic): add part and hideIf to suggestions

liviarett:
https://coveord.atlassian.net/browse/KIT-162

This adds parts to the recent queries header to make it more customisable. I also noticed that this was happening:
Screenshot 2022-05-11 at 12 43 35
7
so had an idea to add a `hideIf` function to the suggestion, which we can later build upon if we need to

### DIFF
--- a/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
@@ -419,12 +419,27 @@ export class AtomicSearchBox {
     this.ariaMessage = '';
   }
 
+  private makeSuggestionPart(isSelected: boolean, customPart = '') {
+    let part = 'suggestion';
+    if (isSelected) {
+      part += ' active-suggestion';
+    }
+    if (customPart) {
+      part += ` ${customPart}`;
+    }
+    return part;
+  }
+
   private renderSuggestion(
     suggestion: SearchBoxSuggestionElement,
-    index: number
+    index: number,
+    lastIndex: number
   ) {
     const id = `${this.id}-suggestion-${index}`;
     const isSelected = id === this.activeDescendant;
+    if (suggestion.hideIf?.(index === lastIndex)) {
+      return null;
+    }
     return (
       <li
         id={id}
@@ -432,7 +447,7 @@ export class AtomicSearchBox {
         aria-selected={`${isSelected}`}
         key={suggestion.key}
         data-query={suggestion.query}
-        part={isSelected ? 'active-suggestion suggestion' : 'suggestion'}
+        part={this.makeSuggestionPart(isSelected, suggestion.part)}
         class={`flex px-4 h-10 items-center text-neutral-dark hover:bg-neutral-light cursor-pointer first:rounded-t-md last:rounded-b-md ${
           isSelected ? 'bg-neutral-light' : ''
         }`}
@@ -467,7 +482,11 @@ export class AtomicSearchBox {
         }`}
       >
         {this.suggestionElements.map((suggestion, index) =>
-          this.renderSuggestion(suggestion, index)
+          this.renderSuggestion(
+            suggestion,
+            index,
+            this.suggestionElements.length - 1
+          )
         )}
       </ul>
     );
--- a/packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
+++ b/packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
@@ -122,16 +122,18 @@ export class AtomicSearchBoxRecentQueries {
       key: 'recent-query-clear',
       content: (
         <div class="flex justify-between w-full">
-          <span class="font-bold">
+          <span class="font-bold" part="recent-query-title">
             {this.bindings.i18n.t('recent-searches')}
           </span>
-          <span>{this.bindings.i18n.t('clear')}</span>
+          <span part="recent-query-clear">{this.bindings.i18n.t('clear')}</span>
         </div>
       ),
       onSelect: () => {
         this.recentQueriesList.clear();
         this.bindings.triggerSuggestions();
       },
+      part: 'recent-query-header',
+      hideIf: (isLast: boolean) => isLast,
     };
   }
 
--- a/packages/atomic/src/components/search-box-suggestions/suggestions-common.ts
+++ b/packages/atomic/src/components/search-box-suggestions/suggestions-common.ts
@@ -7,8 +7,10 @@ import {closest} from '../../utils/utils';
 export interface SearchBoxSuggestionElement {
   key: string;
   query?: string;
+  part?: string;
   onSelect(): void;
   content: Element | VNode;
+  hideIf?(isLast: boolean): boolean;
 }
 
 export interface SearchBoxSuggestions {
--- a/packages/atomic/src/pages/examples/suggestions.html
+++ b/packages/atomic/src/pages/examples/suggestions.html
@@ -21,7 +21,7 @@
           suggestions = [];
 
           renderSuggestions(bindings) {
-            return this.suggestions.map((suggestion) => {
+            return this.suggestions.slice(0, 4).map((suggestion) => {
               const content = document.createElement('div');
               content.innerText = suggestion;
               return {
@@ -82,7 +82,7 @@
         <atomic-layout-section section="search">
           <atomic-search-box suggestion-timeout="1000">
             <custom-suggestions></custom-suggestions>
-            <atomic-search-box-query-suggestions></atomic-search-box-query-suggestions>
+            <atomic-search-box-query-suggestions icon="../assets/magnifier.svg"> </atomic-search-box-query-suggestions>
           </atomic-search-box>
         </atomic-layout-section>
         <atomic-layout-section section="main">


packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -122,16 +122,18 @@ export class AtomicSearchBoxRecentQueries {
...
           </span>
-          <span>{this.bindings.i18n.t('clear')}</span>
+          <span part="recent-query-clear">{this.bindings.i18n.t('clear')}</span>
         </div>
       ),
       onSelect: () => {
         this.recentQueriesList.clear();
         this.bindings.triggerSuggestions();
       },
+      part: 'recent-query-header',
```
ThibodeauJF, line 135: Could that part be set on the div at line 124? It would remove the necessity for the prop

packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -122,16 +122,18 @@ export class AtomicSearchBoxRecentQueries {
...
+      hideIf: (isLast: boolean) => isLast,
```
ThibodeauJF, line 136: We should atomically handle that case, it seems like a bug, maybe we can improve the condition on line 113, and remove that prop too.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -419,20 +419,35 @@ export class AtomicSearchBox {
     this.ariaMessage = '';
   }
 
+  private makeSuggestionPart(isSelected: boolean, customPart = '') {
```
ThibodeauJF, line 422:
```suggestion
  private makeSuggestionPart(isSelected: boolean, customPart?: string) {
```

packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
btaillon: If people want to hide the clear button, I think it should be done with a prop on `atomic-search-box-recent-queries`. I wouldn't try to support either method until there's demand for it though.

packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
btaillon: I agree with JF on this, I don't think we should add this prop.

IMO, if people want to hide a suggestion, I think they should just not return it.

liviarett: I thought about it, but I think at this point we have access to this only
Screenshot 2022-05-11 at 19 14 49
and not the actual elements that would be rendered. Do you think it's worth adding it to the state?

liviarett: I guess one thing that we could do is, in `atomic-search-box`, check that the last element doesn't have the `recent-query-clear`. that doesn't expose any of this functionality. what do you think? @ThibodeauJF

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -419,20 +419,32 @@ export class AtomicSearchBox {
...
+
   private renderSuggestion(
     suggestion: SearchBoxSuggestionElement,
-    index: number
+    index: number,
+    lastIndex: number
   ) {
     const id = `${this.id}-suggestion-${index}`;
     const isSelected = id === this.activeDescendant;
+    if (suggestion.key === 'recent-query-clear' && index === lastIndex) {
```
olamothe, line 437: As was discussed in scrum I don't think we should do this.

There should be a mechanism for suggestion provider to mark an element as an "header". 

And then the search box will be able to decide wether to render the header or not when it hit that corner case.

ThibodeauJF:
Proposed solution:

Adding a `part` option to:

```ts
export interface SearchBoxSuggestionElement {
  ...
  part?: string;
}
```

Which is passed down to the li element:
```ts
<li
  ...
  part={`${isSelected ? 'active-suggestion suggestion' : 'suggestion'} ${
    suggestion.part ? suggestion.part : ''
  }`}
```

Setting a part on the header `recent-query-header`

Finally using some CSS to hide it when it's the last child
```ts
ul[part='suggestions'] > li:last-child:is([part~='recent-query-header']) {
  @apply hidden;
}
```

liviarett:
@ThibodeauJF I think this still has the same issues that Olivier mentioned before...
I can still add back the `part` prop if you think it will be useful, but I introduced a new type for the divider (I thought calling it header might make it sound less flexible).
I feel like it might be a breaking change because I made `query` required now... should I leave it as optional?
Let me know what you guys think about this solution!

packages/atomic/src/components/search-box-suggestions/suggestions-common.ts
```diff
@@ -4,17 +4,27 @@ import {buildCustomEvent} from '../../utils/event-utils';
...
+  key: string;
+  content: Element | VNode;
+  onSelect?(): void;
+}
+
 export interface SearchBoxSuggestionElement {
   key: string;
-  query?: string;
-  onSelect(): void;
+  query: string;
```
ThibodeauJF, line 15: We should leave this possibly undefined, for example an instant result element won't necessarily update the query

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -419,35 +419,50 @@ export class AtomicSearchBox {
...
-    suggestion: SearchBoxSuggestionElement,
-    index: number
+    item: SearchBoxSuggestionItem,
+    index: number,
+    lastIndex: number
   ) {
     const id = `${this.id}-suggestion-${index}`;
     const isSelected = id === this.activeDescendant;
+    const isLast = index === lastIndex;
+    const isDivider = isDividerElement(item);
```
ThibodeauJF, line 441: The divider could be hidden whenever there are no associated items, rather than "if is last"

packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -117,15 +119,15 @@ export class AtomicSearchBoxRecentQueries {
...
   }
 
-  private renderClear(): SearchBoxSuggestionElement {
+  private renderClear(): SearchBoxDividerElement {
     return {
       key: 'recent-query-clear',
       content: (
         <div class="flex justify-between w-full">
-          <span class="font-bold">
+          <span class="font-bold" part="recent-query-title">
```
ThibodeauJF, line 127: parts have to be documented at the top of the atomic-search-box component

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -419,35 +419,50 @@ export class AtomicSearchBox {
     this.ariaMessage = '';
   }
 
+  private makeSuggestionPart(isSelected: boolean, isDivider: boolean) {
+    let part = 'suggestion';
+    if (isSelected) {
+      part += ' active-suggestion';
+    }
+    if (isDivider) {
+      part += ' suggestion-divider';
```
ThibodeauJF, line 428: Need to add documentation for this one too

liviarett: true

packages/atomic/src/components/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -7,14 +7,19 @@ import {
...
   SearchBoxSuggestionsBindings,
 } from '../suggestions-common';
 import {once} from '../../../utils/utils';
 import {SafeStorage, StorageItems} from '../../../utils/local-storage-utils';
 
 /**
  * The `atomic-search-box-recent-queries` component can be added as a child of an `atomic-search-box` component, allowing for the configuration of recent query suggestions.
+ *
+ * @part recent-query-title - The 'Recent queries' title.
+ * @part recent-query-clear - The 'Clear' button for clearing recent queries.
```
jpmarceau, line 22:
```suggestion
 * @part recent-query-clear - The 'Clear' button to clear recent queries.
```

jpmarceau (approved):
👍

packages/atomic/src/components/search-box-suggestions/suggestions-common.ts
liviarett: @olamothe @btaillon

olamothe: What's the concern with backward compatibility 🤔 ?

olamothe (approved):

packages/atomic/src/components/search-box-suggestions/suggestions-common.ts
liviarett: I was thinking, if someone already had some suggestions without `query` in their code, it would break.. right?
but I guess since we only exposed this recently, that's OK?

olamothe: Yes, backward compatibility only kicks in when it's publicly documented/official. 

Whatever we create while designing and working on a new feature is fair game to modify and break.

liviarett: great news! I'll merge

PR #1984, opened 2022-05-11
feat(atomic): initial load performance + UX improvements

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1526

Mostly improves how nice the initial load looks

**Before** 
Search call: ~300ms
First Contentful Paint: 261.3 ms
Everything rendered but results: ~750 ms
Largest Contentful Paint: 1074.6 ms

**Now with network & load improvements**
Search call: ~300ms
First Contentful Paint: 255 ms
Everything rendered but results: ~700 ms
Largest Contentful Paint: 922.5ms

Will be even less when it's gzip on the CDN and cache is enabled (for files, search call)

### DIFF
--- a/packages/atomic/src/components/atomic-facet-manager/atomic-facet-manager.tsx
+++ b/packages/atomic/src/components/atomic-facet-manager/atomic-facet-manager.tsx
@@ -64,6 +64,7 @@ export class AtomicFacetManager implements InitializableComponent {
 
   private sortFacets = () => {
     if (!this.searchStatusState.firstSearchExecuted) {
+      this.updateCollapsedState(this.facets);
       return;
     }
     const payload = this.facets.map((f) => ({facetId: f.facetId, payload: f}));
--- a/packages/atomic/src/components/atomic-icon/atomic-icon.tsx
+++ b/packages/atomic/src/components/atomic-icon/atomic-icon.tsx
@@ -43,7 +43,6 @@ export class AtomicIcon implements InitializableComponent {
 
   public error!: Error;
   @State() private svg: string | null = null;
-  private deferRenderingPromise?: Promise<unknown>;
 
   private async fetchIcon(url: string) {
     try {
@@ -79,16 +78,11 @@ export class AtomicIcon implements InitializableComponent {
   @Watch('icon')
   public async updateIcon() {
     const svgPromise = this.getIcon();
-    this.deferRenderingPromise = svgPromise;
     this.svg = await svgPromise;
   }
 
   public initialize() {
     this.updateIcon();
-  }
-
-  public async componentWillRender() {
-    await this.deferRenderingPromise;
   }
 
   public render() {
--- a/packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
+++ b/packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
@@ -109,7 +109,10 @@ export class AtomicLoadMoreResults {
   }
 
   public render() {
-    if (!isAppLoaded(this.bindings.store)) {
+    if (
+      !isAppLoaded(this.bindings.store) ||
+      !this.querySummaryState.hasResults
+    ) {
       return;
     }
 
--- a/packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
+++ b/packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
@@ -12,6 +12,7 @@ import {
   BindStateToController,
   InitializeBindings,
 } from '../../utils/initialization-utils';
+import {isAppLoaded} from '../../utils/store';
 import {Button} from '../common/button';
 
 /**
@@ -108,7 +109,7 @@ export class AtomicLoadMoreResults {
   }
 
   public render() {
-    if (!this.bindings.store.get('firstResultLoaded')) {
+    if (!isAppLoaded(this.bindings.store)) {
       return;
     }
 
--- a/packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
+++ b/packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
@@ -108,7 +108,7 @@ export class AtomicLoadMoreResults {
   }
 
   public render() {
-    if (!this.querySummaryState.hasResults) {
+    if (!this.bindings.store.get('firstResultLoaded')) {
       return;
     }
 
--- a/packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
+++ b/packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
@@ -83,6 +83,7 @@ export class AtomicRefineModal implements InitializableComponent {
     this.querySummary = buildQuerySummary(this.bindings.engine);
     this.facetManager = buildFacetManager(this.bindings.engine);
     this.sort = buildSort(this.bindings.engine);
+    this.watchEnabled(this.isOpen);
   }
 
   private onAnimationEnded() {
@@ -255,5 +256,9 @@ export class AtomicRefineModal implements InitializableComponent {
         {this.renderFooter()}
       </atomic-modal>
     );
+  }
+
+  public componentDidLoad() {
+    this.host.style.display = '';
   }
 }
--- a/packages/atomic/src/components/atomic-refine-toggle/atomic-refine-toggle.tsx
+++ b/packages/atomic/src/components/atomic-refine-toggle/atomic-refine-toggle.tsx
@@ -27,6 +27,7 @@ export class AtomicRefineToggle implements InitializableComponent {
   @Element() public host!: HTMLElement;
   public searchStatus!: SearchStatus;
   private modalRef?: HTMLAtomicRefineModalElement;
+  private buttonRef?: HTMLButtonElement;
 
   @InitializeBindings() public bindings!: Bindings;
   @State() public error!: Error;
@@ -36,11 +37,19 @@ export class AtomicRefineToggle implements InitializableComponent {
 
   public initialize() {
     this.searchStatus = buildSearchStatus(this.bindings.engine);
+  }
+
+  private loadModal() {
     this.modalRef = document.createElement('atomic-refine-modal');
+    this.modalRef.style.display = 'none';
     this.host.insertAdjacentElement('beforebegin', this.modalRef);
+    this.modalRef.openButton = this.buttonRef;
   }
 
   private enableModal() {
+    if (!this.modalRef) {
+      this.loadModal();
+    }
     this.modalRef && (this.modalRef.isOpen = true);
   }
 
@@ -69,7 +78,7 @@ export class AtomicRefineToggle implements InitializableComponent {
         class="p-3 w-full"
         onClick={() => this.enableModal()}
         text={this.bindings.i18n.t('sort-and-filter')}
-        ref={(button) => (this.modalRef!.openButton = button)}
+        ref={(button) => (this.buttonRef = button)}
         part="button"
       ></Button>
     );
--- a/packages/atomic/src/components/atomic-result/atomic-result.tsx
+++ b/packages/atomic/src/components/atomic-result/atomic-result.tsx
@@ -12,7 +12,7 @@ import {
   ResultContextEvent,
 } from '../result-template-components/result-template-decorators';
 import {ObservableMap} from '@stencil/store';
-import {AtomicStore} from '../../utils/store';
+import {AtomicStore, unsetLoadingFlag} from '../../utils/store';
 
 const resultSectionTags = [
   'atomic-result-section-visual',
@@ -82,6 +82,11 @@ export class AtomicResult {
    * Classes that will be added to the result element.
    */
   @Prop() classes = '';
+
+  /**
+   * @internal
+   */
+  @Prop() loadingFlag?: string;
 
   @Listen('atomic/resolveResult')
   public resolveResult(event: ResultContextEvent<FoldedResult | Result>) {
@@ -156,8 +161,8 @@ export class AtomicResult {
   }
 
   public componentDidLoad() {
-    if (!this.store.get('firstResultLoaded')) {
-      this.store.set('firstResultLoaded', true);
+    if (this.loadingFlag) {
+      unsetLoadingFlag(this.store, this.loadingFlag);
     }
     applyFocusVisiblePolyfill(this.host);
   }
--- a/packages/atomic/src/components/atomic-result/atomic-result.tsx
+++ b/packages/atomic/src/components/atomic-result/atomic-result.tsx
@@ -1,12 +1,4 @@
-import {
-  Component,
-  h,
-  Prop,
-  Element,
-  Listen,
-  Event,
-  EventEmitter,
-} from '@stencil/core';
+import {Component, h, Prop, Element, Listen} from '@stencil/core';
 import {FoldedResult, Result, SearchEngine} from '@coveo/headless';
 import {
   ResultDisplayLayout,
@@ -19,6 +11,8 @@ import {
   DisplayConfig,
   ResultContextEvent,
 } from '../result-template-components/result-template-decorators';
+import {ObservableMap} from '@stencil/store';
+import {AtomicStore} from '../../utils/store';
 
 const resultSectionTags = [
   'atomic-result-section-visual',
@@ -53,6 +47,11 @@ export class AtomicResult {
   @Prop() engine!: SearchEngine;
 
   /**
+   * Global state for Atomic.
+   */
+  @Prop() store!: ObservableMap<AtomicStore>;
+
+  /**
    * The result content to display.
    */
   @Prop() content!: ParentNode;
@@ -84,11 +83,6 @@ export class AtomicResult {
    */
   @Prop() classes = '';
 
-  /**
-   * @internal
-   */
-  @Prop() firstResultToRender!: boolean;
-
   @Listen('atomic/resolveResult')
   public resolveResult(event: ResultContextEvent<FoldedResult | Result>) {
     event.preventDefault();
@@ -105,11 +99,6 @@ export class AtomicResult {
       imageSize: this.imageSize!,
     });
   }
-
-  @Event({
-    eventName: 'atomic/removeResultsPlaceholders',
-  })
-  private removeResultsPlaceholders!: EventEmitter;
 
   private containsSections() {
     return Array.from(this.content.children).some((child) =>
@@ -167,8 +156,8 @@ export class AtomicResult {
   }
 
   public componentDidLoad() {
-    if (this.firstResultToRender) {
-      this.removeResultsPlaceholders.emit();
+    if (!this.store.get('firstResultLoaded')) {
+      this.store.set('firstResultLoaded', true);
     }
     applyFocusVisiblePolyfill(this.host);
   }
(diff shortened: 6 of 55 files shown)


packages/atomic/src/components/atomic-facet-manager/atomic-facet-manager.tsx
```diff
@@ -64,6 +64,7 @@ export class AtomicFacetManager implements InitializableComponent {
 
   private sortFacets = () => {
     if (!this.searchStatusState.firstSearchExecuted) {
+      this.updateCollapsedState(this.facets);
```
ThibodeauJF, line 67: collapses the placeholders too so the page height doesn't change that much

packages/atomic/src/components/atomic-refine-toggle/atomic-refine-toggle.tsx
```diff
@@ -36,11 +37,19 @@ export class AtomicRefineToggle implements InitializableComponent {
...
+  private loadModal() {
     this.modalRef = document.createElement('atomic-refine-modal');
+    this.modalRef.style.display = 'none';
     this.host.insertAdjacentElement('beforebegin', this.modalRef);
+    this.modalRef.openButton = this.buttonRef;
   }
 
   private enableModal() {
+    if (!this.modalRef) {
+      this.loadModal();
```
ThibodeauJF, line 51: We won't load the atomic refine modal + atomic modal files until it's needed. I guess most users accessing a search page won't require the modal.

packages/atomic/src/components/atomic-result/atomic-result.tsx
```diff
@@ -146,6 +156,9 @@ export class AtomicResult {
   }
 
   public componentDidLoad() {
+    if (!this.store.get('firstResultLoaded')) {
+      this.store.set('firstResultLoaded', true);
```
ThibodeauJF, line 160: The first result to ever load sets the flag, this and related use of the flag fixes the placeholder

packages/atomic/src/components/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -256,7 +256,7 @@ export class AtomicSearchInterface {
    *
    * Executes the first search and logs the interface load event to analytics, after initializing connection to the headless search engine.
    */
-  @Method() public executeFirstSearch() {
+  @Method() public async executeFirstSearch() {
```
ThibodeauJF, line 259: Screen Shot 2022-05-11 at 11 18 54 AM
my bad!

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -195,26 +195,21 @@ export class ResultListCommon {
...
 
-  public componentDidRender(
-    firstSearchExecuted: boolean,
-    listWrapperRef?: HTMLDivElement
-  ) {
-    if (firstSearchExecuted) {
-      listWrapperRef?.classList.remove('placeholder');
-    }
-  }
-
```
ThibodeauJF, line 206: When the result list loads/render is not a good indicator that the results load, cause the template components files have to load too. This caused a big gap on slow connections

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -235,38 +230,42 @@ export class ResultListCommon {
...
-              getContentOfResultTemplate={getContentOfResultTemplate}
-            />
+            {displayPlaceholders && (
+              <ResultsPlaceholder
+                display={display}
+                density={density}
+                imageSize={imageSize}
+                resultsPerPageState={resultsPerPageState}
+              />
+            )}
```
ThibodeauJF, line 255: Placeholders were kept in the dom, lot's of unused elements in there, should help performance

packages/atomic/src/utils/store.ts
```diff
@@ -32,6 +32,11 @@ export type AtomicStore = {
   facetElements: HTMLElement[];
   sortOptions: SortDropdownOption[];
   iconAssetsPath: string;
+  /**
+   * Flag set to 'true' when the first "atomic-result" & its component have been rendered, replacing the placeholders.
+   * When first loading, Atomic has a rendering delay of its results, hence this is a better indicator than the "firstSearchExecuted" of the Headless state.
+   */
+  firstResultLoaded: boolean;
```
ThibodeauJF, line 39: Instead of using an event, which I first implemented, I decided to use a store flag so it's more easily accessible by any Atomic component or the users themselves

ThibodeauJF: I realise that the "Store" isn't a property on the interface like the engine is, we could add this too.

packages/atomic/src/components/atomic-icon/atomic-icon.tsx
```diff
@@ -79,18 +78,13 @@ export class AtomicIcon implements InitializableComponent {
...
     this.svg = await svgPromise;
   }
 
   public initialize() {
     this.updateIcon();
   }
 
-  public async componentWillRender() {
-    await this.deferRenderingPromise;
-  }
```
ThibodeauJF, line 92: removed because it's blocking rendering. The icon's size doesn't change and the background color is already visible for result icons. It doesn't seem like bad ux

packages/atomic/src/components/atomic-refine-toggle/atomic-refine-toggle.tsx
ThibodeauJF: Same for the explain why modal, also in this PR

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
```diff
@@ -30,6 +40,13 @@ export class AtomicSmartSnippetAnswer {
     this.emitCurrentHeight();
   }
 
+  public componentDidLoad() {
+    // Prevents initial transition
+    setTimeout(() => {
+      this.host.classList.add('animatable');
+    }, 0);
+  }
```
ThibodeauJF, line 48: I was able to see some piece of animation on load, setting the transition after the height is set does it. Open to other working solutions

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-expandable-answer.tsx
```diff
@@ -26,7 +26,7 @@ export class AtomicSmartSnippetExpandableAnswer {
   @Element() public host!: HTMLElement;
 
   @Prop({reflect: true}) expanded!: boolean;
-  @Prop({reflect: true}) htmlContent!: string;
+  @Prop() htmlContent!: string;
```
ThibodeauJF, line 29: Reflecting HTML or CSS is heavy on the DOM for no true reason since this content is inserted in the component.

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -87,14 +88,19 @@ export class AtomicSmartSnippetSuggestions implements InitializableComponent {
...
 
   public initialize() {
     this.smartSnippetQuestionsList = buildSmartSnippetQuestionsList(
       this.bindings.engine
     );
+
+    this.hideDuringRender = true;
+    waitUntilAppLoaded(this.bindings.store, () => {
+      this.hideDuringRender = false;
+    });
```
ThibodeauJF, line 103: this syncs smart snippets initial rendering with the results

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -107,6 +113,11 @@ export class AtomicSmartSnippetSuggestions implements InitializableComponent {
     return styleTag.innerHTML;
   }
 
+  private set hideDuringRender(shouldHide: boolean) {
+    this.host.style.visibility = shouldHide ? 'hidden' : '';
+    this.host.style.position = shouldHide ? 'absolute' : '';
+  }
```
ThibodeauJF, line 119: copied solution from atomic-smart-snippet

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-suggestions.tsx
```diff
@@ -216,14 +227,16 @@ export class AtomicSmartSnippetSuggestions implements InitializableComponent {
...
           {this.renderQuestion(relatedQuestion, index)}
-          <div
-            part={relatedQuestion.expanded ? 'answer-and-source' : ''}
-            class={relatedQuestion.expanded ? 'pl-10 pr-6 pb-6' : 'hidden'}
-            id={this.getRelatedQuestionId(index)}
-          >
-            {this.renderContent(relatedQuestion)}
-            {this.renderSource(relatedQuestion)}
-          </div>
+          {relatedQuestion.expanded && (
```
ThibodeauJF, line 230: extra unnecessary DOM

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -102,9 +104,20 @@ export class AtomicSmartSnippet implements InitializableComponent {
 
   public initialize() {
     this.smartSnippet = buildSmartSnippet(this.bindings.engine);
+    waitUntilAppLoaded(this.bindings.store, () => {
+      this.hideDuringRender = false;
+    });
+  }
+
+  private loadModal() {
```
ThibodeauJF, line 112: like the refine modal, only load it when needed

packages/atomic/src/utils/store.ts
```diff
@@ -32,6 +32,13 @@ export type AtomicStore = {
   facetElements: HTMLElement[];
   sortOptions: SortDropdownOption[];
   iconAssetsPath: string;
+  /**
+   * Dynamic list of flags that allows for syncing various components loading state on initialization.
+   * E.g., waiting for result template component to be rendered.
+   *
+   * This is a better indicator than only the "firstSearchExecuted" of the Headless search state.
+   */
+  loadingFlags: string[];
```
ThibodeauJF, line 41: Key addition of this PR, I started with only a "firstResultLoaded" flag but I realized it wouldn't scale like support smart snippets (also have a delayed rendering) or multiple result lists. Atomic users would potentially be able to add flags of their own, to sync nicely with the rendering of the result list.

I decided not to type for that reason:
- allows users to add new ones
- multiple flags for multiple result lists

packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
```diff
@@ -108,7 +109,10 @@ export class AtomicLoadMoreResults {
   }
 
   public render() {
-    if (!this.querySummaryState.hasResults) {
+    if (
+      !isAppLoaded(this.bindings.store) ||
```
olamothe, line 113: Is there * any * component for which the check on `isAppLoaded` should not be performed when rendering ?

Granted, for many of them it will always be true (eg: results list components). 

Should it be embedded in the `InitializeBindings` function ? Same way we check for generic component errors.

packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -256,4 +257,8 @@ export class AtomicRefineModal implements InitializableComponent {
       </atomic-modal>
     );
   }
+
+  public componentDidLoad() {
+    this.host.style.display = '';
```
olamothe, line 262: What is this needed for ?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
olamothe: That's indeed pretty hackish 🤔 Can we use other lifecyle ? `componentDidRender` ?

packages/atomic/src/components/facets/atomic-facet-placeholder/atomic-facet-placeholder.tsx
```diff
@@ -27,7 +29,7 @@ export const FacetPlaceholder: FunctionalComponent<FacetPlaceholderProps> = ({
         class="bg-neutral rounded h-8"
         style={{width: `${getRandomArbitrary(25, 75)}%`}}
       ></div>
-      <div class="mt-7">{facetValues}</div>
+      {!isCollapsed && <div class="mt-7">{facetValues}</div>}
```
olamothe, line 32: Neat little visual improvement ;)

packages/atomic/src/utils/store.ts
olamothe: > Atomic users would potentially be able to add flags of their own, to sync nicely with the rendering of the result list.

I think that's just "in theory" though.

If we start with typed flags that we only use internally, then nothing stop us from expanding and making it public down the line if it really becomes a need. So I would vote to keep this private/typed/internal only, and then open it up later if needed.

VS starting with it being public, and it's harder to change/modify. What do you think ?

olamothe (approved):

packages/atomic/src/components/atomic-load-more-results/atomic-load-more-results.tsx
ThibodeauJF: From UX perspective, only the pagination should wait here, isAppLoaded will be true when the search is finished and the template components have rendered.

The search box, facet, header components (query summary. sort, etc) all load before the results, and some have their own placeholders. If we wait and display nothing, the page will look much more sluggish as nothing will be displayed for a while

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
ThibodeauJF: `componentDidLoad` is the last of them. I can investigate more on the smart snippet rendering loop though, @btaillon can surely help me with it

packages/atomic/src/utils/store.ts
ThibodeauJF: > multiple flags for multiple result lists
This is used right now, each list has it's unique flag ID. The smart snippet could also be added here...
If we type the flags we might not be able to support this

packages/atomic/src/components/atomic-refine-modal/atomic-refine-modal.tsx
ThibodeauJF: https://github.com/coveo/ui-kit/pull/1984/files#r871352884

packages/atomic/src/components/result-lists/grid-display-results.tsx
```diff
@@ -29,10 +29,12 @@ export const GridDisplayResults: FunctionalComponent<ResultsProps> = (
           )}
           result={result}
           engine={props.bindings.engine}
+          store={props.bindings.store}
```
mbenkirane-coveo, line 32: is the props.bindings.store a type of datastructure or hashmap?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
```diff
@@ -100,9 +102,21 @@ export class AtomicSmartSnippet implements InitializableComponent {
...
+    if (this.modalRef) {
+      return;
+    }
+
     this.modalRef = document.createElement(
       'atomic-smart-snippet-feedback-modal'
     );
+    // console.log('hmm', this.explainWhyBtnRef);
```
liviarett, line 118: some comments forgotten here!

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -235,38 +237,46 @@ export class ResultListCommon {
       return;
     }
 
+    if (resultListState.firstSearchExecuted && !resultListState.hasResults) {
+      return;
+    }
+
+    const displayPlaceholders = !isAppLoaded(this.bindings.store);
```
liviarett, line 244: I don't know if you explained this before, but what is the difference between `isAppLoaded(this.bindings.store);` and `resultListState.firstSearchExecuted`?

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -235,38 +237,46 @@ export class ResultListCommon {
...
-            />
+            {displayPlaceholders && (
+              <ResultsPlaceholder
+                display={display}
+                density={density}
+                imageSize={imageSize}
+                resultsPerPageState={resultsPerPageState}
+              />
+            )}
+            {resultListState.firstSearchExecuted && (
```
liviarett, line 267: should this be 
```js
{displayPlaceholders && <Results...
```
so we guarantee that it's always one or the other?

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet.tsx
ThibodeauJF: 😄

packages/atomic/src/components/result-lists/grid-display-results.tsx
ThibodeauJF: It's an instance of https://stenciljs.com/docs/stencil-store#stencil-store

packages/atomic/src/components/result-lists/result-list-common.tsx
ThibodeauJF: It used to look for these variables but it wasn't enough, the resultListState is the status of headless related to the search call. We need more than that to know that the results in Atomic have truly loaded:

1. The Stencil web component defined in the HTML are mounted, that mean's each file is being loaded separately for each component
2. initialize() method is called on the parent interface and on each of the registered components in the page
3. the result list loads it's result template children result template
4. executeFirstSearch() method is called on the parent interface, `resultListState.isLoading` is now `true`
5. the search fulfills and the results are loaded in the state `resultListState.isLoading` is now `false` & `resultListState.firstSearchExecuted` is true. This is the start of the "gap" where placeholders were hidden but results not yet visible
6. the result list renders the proper results using one of the result templates
7. now that the result template components are in the HTML, the related files are being loaded in order to render each components
8. once the components have loaded & rendered, the `isAppLoaded` should return `true`, this is the end of the visual "gap"

ThibodeauJF: Maybe i can make the following more clear ? https://github.com/coveo/ui-kit/pull/1984/files#diff-6fb7deead3ab9ab5f889dad6759d737983382cc2594ccf00f31a16d99099af82R35-R40

packages/atomic/src/components/result-lists/result-list-common.tsx
ThibodeauJF: Results & placeholders will coexist in the dom for a bit until the results are truly loaded

packages/atomic/src/components/atomic-result/atomic-result.tsx
```diff
@@ -44,6 +46,11 @@ export class AtomicResult {
    */
   @Prop() engine!: SearchEngine;
 
+  /**
+   * Global state for Atomic.
+   */
+  @Prop() store!: ObservableMap<AtomicStore>;
```
btaillon, line 52: I would make this optional and internal

packages/atomic/src/components/atomic-smart-snippet/atomic-smart-snippet-answer.tsx
btaillon: Since from the perspective of this component there's no animation, I would change the class name from `animtable` to `loaded` or `rendered`.

I agree that `componentDidRender` may be preferable. If we use `componentDidRender`, then we should also remove the class on `componentWillRender`.

PR #1985, opened 2022-05-12
feat(headless): add instant results to controller

liviarett:
https://coveord.atlassian.net/browse/KIT-1631?atlOrigin=eyJpIjoiZjZhNWUxOGUzNGI0NGUzMzk0MjVkOWIxYTAyNTUzOWEiLCJwIjoiaiJ9

I already have some barebones code for `atomic` if that makes it easier to understand this change...


packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+import {SearchAPIErrorWithStatusCode} from '../../api/search/search-api-error-response';
+import {SerializedError} from '@reduxjs/toolkit';
+
+export interface InstantResultProps {
+  options: InstantResultOptions;
+}
+
+export interface InstantResults extends Controller {
+  /**
+   * Updates the instant results query and shows instant results for it.
```
olamothe, line 26: I've added our trusted API writers to help us here ;)

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+    engine,
+    instantResultsOptionsSchema,
+    options,
+    'buildInstantResults'
+  );
+
+  const searchBoxId = options.searchBoxId;
+  dispatch(registerInstantResults({id: searchBoxId}));
+
+  const getStateForSearchBox = () => getState().instantResults[searchBoxId];
```
olamothe, line 90: We'll need defensive code around those access (and for the cache too).

Especially since the ID is an option, some someone passing in a wrong option should get a verbose explanation of what's going wrong, instead of a null ref crash.

And an early return from the function.

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const getCached = (q: string) => getStateForSearchBox().cache[q];
+  const getQ = () => getStateForSearchBox().q;
+  const getResults = () => {
+    const cached = getCached(getQ());
```
olamothe, line 95: The cache is a good idea, but it might need to be a bit need to become more sophisticated than this.

* Is there a max size ?
* Is there a max duration ?
* Should an implementer be able to clear the cache if they want to ?
* Should an implementer be able to configure the cache parameters if they want to ?
* What kind of problem/questions can we expect with this cache ? For example, if someone is doing an implementation and they are trying to debug their instant result, and they don't see "new content" from their index because they did not refresh the page, how will they know there's a cache ? Is this overthinking or is it a true problem ?

We can leave it like that for now, but food for thoughts.

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+    return cached.results;
+  };
+
+  return {
+    ...controller,
+
+    updateQuery(q: string) {
+      const cached = getCached(q);
+      if (!q) return;
```
olamothe, line 108: Should this be before `getCached` line ?

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+      if (!cached || cached.error) {
+        dispatch(fetchInstantResults({id: searchBoxId, q}));
+      }
+      dispatch(updateInstantResultsQuery({id: searchBoxId, q}));
+    },
+
+    get state() {
+      return {
+        q: getQ(),
+        isLoading: getCached(getQ())?.isLoading,
```
olamothe, line 118: The `?` operator creates an "annoying" type from the implementer perspective.

Instead of having a simple boolean, they now have to deal with `boolean or undefined or null`, and manage it in their app.

Ideally, this should return just a boolean, with `false` when there's no hit in the cache.

packages/headless/src/features/instant-results/instant-results-actions.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export interface FetchInstantResultsThunkReturn {
+  results: Result[];
+}
+
+export type StateNeededByInstantResults = StateNeededByExecuteSearch &
+  InstantResultSection;
+
+export const fetchInstantResults = createAsyncThunk<
```
olamothe, line 75: At a practical level, the difference between this and the classic "executeSearch" action in `search-actions.ts` is very subtle. Should the automatic did you mean feature kick in here ? 🤔 

Maybe we should move that action there instead ? So they are all co-located, and can share more code with one another.

We most likely also want to log an analytics event associated with it. This can come in a later PR.

packages/headless/src/features/instant-results/instant-results-slice.test.ts
```diff
@@ -0,0 +1,172 @@
...
+const id2 = 'search_box_2';
+
+const getSearchBoxInstantResultsState = (id: string, q = '', cache = {}) => ({
+  [id]: {
+    q,
+    cache,
+  },
+});
+
+describe('history slice', () => {
```
olamothe, line 20:
```suggestion
describe('instant results slice', () => {
```

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/**
+ * Creates a `InstantResults` controller instance.
+ *
+ * @param engine - The headless engine.
+ * @param props - The configurable `InstantResults` properties.
+ * @returns A `InstantResults` controller instance.
+ */
+export function buildInstantResults(
```
olamothe, line 64: There might be a more "natural" way for implementers to create the link between `a search box and it's suggestion` with `the instant results that come associated with it`.

I don't think we have that pattern yet in Headless, but it seems to me the more natural API would be that you must build a instantResult from or with a searchbox controller...

I'd like to discuss that in the scrum with everyone.

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export interface InstantResultProps {
+  options: InstantResultOptions;
+}
+
+export interface InstantResults extends Controller {
+  /**
+   * Updates the instant results query and shows instant results for it.
+   *
+   * @param q - The string value to get instant results for.
```
olamothe, line 28: One thing that seems weird is the `q` keyword used throughout this PR, or `updateQuery`...

Would it make more sense to use the keyword "suggestion" instead ? 

While it's true you could use any query to get instant result from, we might want to encourage/guide implementers to use query suggestions instead (which is a bit more precise/constrained), in order to get good instant results.

packages/headless/src/controllers/instant-results/instant-results.ts
liviarett: my thought with the controller is that you only pass the id once (when creating it), and then next time you call these functions the id will already be stored in this scope and will definitely exist.. am I wrong?

packages/headless/src/controllers/instant-results/instant-results.ts
liviarett: haha I knew there was something weird about this function, but for some reason couldn't register what it was!

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+  InstantResultOptions,
+  instantResultsOptionsSchema,
+} from './instant-results-options';
+import {SearchAPIErrorWithStatusCode} from '../../api/search/search-api-error-response';
+import {SerializedError} from '@reduxjs/toolkit';
+
+export interface InstantResultProps {
+  options: InstantResultOptions;
+}
+
```
npushkarskii, line 23:
```suggestion

/**
 * The `InstantResults` controller allows the end user to manage instant results queries.
 */
```

packages/headless/src/controllers/instant-results/instant-results.ts
npushkarskii, line 26:
```suggestion
   * Updates the specified query and shows instant results for it.
```

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+export interface InstantResults extends Controller {
+  /**
+   * Updates the instant results query and shows instant results for it.
+   *
+   * @param q - The string value to get instant results for.
+   *
```
npushkarskii, line 29:
```suggestion
```

packages/headless/src/controllers/instant-results/instant-results.ts
npushkarskii, line 28:
```suggestion
   * @param q - The query to get instant results for. For more precise instant results, query suggestions are recommended.
```

@olamothe I've added a sentence here based on what you said above as it made total sense to me 🙏 The only thing is a bit unclear at the moment is you said "in order to get good instant results"; I understood it as "more precise instant results", but this may be wrong. Feel free to correct me and/or comment on this subject 😇

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+  state: InstantResultsState;
+}
+
+export interface InstantResultsState {
+  /**
+   * The current query for instant results.
+   */
+  q: string;
+  /**
+   * The instant results for the current query in instant results.
```
npushkarskii, line 44:
```suggestion
   * The instant results for the current query.
```

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+   */
+  results: Result[];
+  /**
+   * Determines if a search is in progress for the current instant results query.
```
npushkarskii, line 48:
```suggestion
   * Determines if a search is in progress for the current query.
```

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+   */
+  isLoading: boolean;
+  /**
+   * Determines if a search is in progress for the current instant results query.
```
npushkarskii, line 52:
```suggestion
   * An error returned when executing an instant results request, if any. This is `null` otherwise.
```

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+   */
+  error: SearchAPIErrorWithStatusCode | SerializedError | null;
+}
+
+/**
+ * Creates a `InstantResults` controller instance.
```
npushkarskii, line 58:
```suggestion
 * Creates an `InstantResults` controller instance.
```

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * @param engine - The headless engine.
+ * @param props - The configurable `InstantResults` properties.
+ * @returns A `InstantResults` controller instance.
```
npushkarskii, line 62:
```suggestion
 * @returns An `InstantResults` controller instance.
```

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Determines if a search is in progress for the current instant results query.
+   */
+  error: SearchAPIErrorWithStatusCode | SerializedError | null;
+}
+
+/**
+ * Creates a `InstantResults` controller instance.
+ *
+ * @param engine - The headless engine.
```
npushkarskii, line 60:
```suggestion
 * @param engine - The Headless engine.
```

packages/headless/src/features/instant-results/instant-results-actions-loader.ts
```diff
@@ -0,0 +1,47 @@
...
+   */
+  updateInstantResultsQuery(
+    payload: UpdateInstantResultQueryActionCreatorPayload
+  ): PayloadAction<UpdateInstantResultQueryActionCreatorPayload>;
+}
+
+/**
+ * Loads the `instantResults` reducer and returns possible action creators.
+ *
+ * @param engine - The headless engine.
```
npushkarskii, line 37:
```suggestion
 * @param engine - The Headless engine.
```

packages/headless/src/features/instant-results/instant-results-actions-loader.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @returns An object holding the action creators.
```
npushkarskii, line 38:
```suggestion
 * @returns An object with the action creators.
```

packages/headless/src/features/instant-results/instant-results-actions.ts
```diff
@@ -0,0 +1,117 @@
...
+export interface RegisterInstantResultActionCreatorPayload {
+  /**
+   * The search box id.
+   */
+  id: string;
+}
+
+export interface UpdateInstantResultQueryActionCreatorPayload {
+  /**
+   * The search box id.
```
npushkarskii, line 27:
```suggestion
   * The search box ID.
```

packages/headless/src/features/instant-results/instant-results-actions.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export const updateInstantResultsQuery = createAction(
+  'instantResults/updateQuery',
+  (payload: UpdateInstantResultQueryActionCreatorPayload) =>
+    validatePayload(payload, instantResultsQueryDefinition)
+);
+
+export interface FetchInstantResultsActionCreatorPayload {
+  /**
+   * The search box id.
```
npushkarskii, line 59:
```suggestion
   * The search box ID.
```

packages/headless/src/features/instant-results/instant-results-actions-loader.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+  registerInstantResults,
+  updateInstantResultsQuery,
+} from './instant-results-actions';
+
+export interface InstantResultsActionCreators {
+  /**
+   * Initializes the `instantResults` state for a search box id.
+   *
+   * @param payload The initial state and options.
```
npushkarskii, line 17:
```suggestion
   * @param payload - The initial state and options.
```

packages/headless/src/features/instant-results/instant-results-actions-loader.ts
```diff
@@ -0,0 +1,47 @@
...
+   */
+  registerInstantResults(
+    payload: RegisterInstantResultActionCreatorPayload
+  ): PayloadAction<RegisterInstantResultActionCreatorPayload>;
+
+  /**
+   * Updates the query expression for instant results for a search box id.
```
npushkarskii, line 24:
```suggestion
   * Updates the query expression to request instant results for the specified search box ID.
```

packages/headless/src/features/instant-results/instant-results-actions-loader.ts
```diff
@@ -0,0 +1,47 @@
...
+  RegisterInstantResultActionCreatorPayload,
+} from './instant-results-actions';
+import {
+  registerInstantResults,
+  updateInstantResultsQuery,
+} from './instant-results-actions';
+
+export interface InstantResultsActionCreators {
+  /**
+   * Initializes the `instantResults` state for a search box id.
```
npushkarskii, line 15:
```suggestion
   * Initializes the `instantResults` state for a search box ID.
```

packages/headless/src/features/instant-results/instant-results-actions.ts
```diff
@@ -0,0 +1,117 @@
...
+} from '../../api/search/search-api-client';
+import {StateNeededByExecuteSearch} from '../search/search-actions';
+import {mapSearchRequest, mapSearchResponse} from '../search/search-mappings';
+import {buildSearchAndFoldingLoadCollectionRequest} from '../search-and-folding/search-and-folding-request';
+import {logQueryError} from '../search/search-analytics-actions';
+import {Result} from '../../case-assist.index';
+
+export interface RegisterInstantResultActionCreatorPayload {
+  /**
+   * The search box id.
```
npushkarskii, line 20:
```suggestion
   * The search box ID.
```

npushkarskii (approved):
Looks great 🎉 Left a couple of comments, though 😋

packages/headless/src/controllers/instant-results/instant-results.ts
liviarett: After our discussion last week, I'll leave the pattern as it currently is

packages/headless/src/controllers/instant-results/instant-results.ts
liviarett: After our conversation last week, it feels to me like we're not coupling instant results to suggestions so tightly, so it feels like it should still be called query, what do you think?

olamothe: Alright !

olamothe (approved):

packages/headless/src/controllers/instant-results/instant-results.ts
olamothe: Enhance later is 👍 .

liviarett: I got excited and added the `expiresAt` key in the latest commits

packages/headless/src/controllers/instant-results/instant-results.ts
liviarett: and I'll add a default id!!

ThibodeauJF (approved):

packages/headless/src/controllers/instant-results/instant-results.ts
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * Updates the specified query and shows instant results for it.
+   *
+   * @param q - The query to get instant results for. For more precise instant results, query suggestions are recommended.
+   */
+  updateQuery(q: string): void;
+  /**
+   * Fetches results from the server for a query.
+   */
+  updateResultsForQuery(q: string): void;
```
btaillon, line 40: Do developers need to call this function? It looks like `updateQuery` is what we expect them to use.

I would not expose this method unless necessary.

packages/headless/src/features/instant-results/instant-results-actions.ts
```diff
@@ -0,0 +1,86 @@
...
+    validatePayload(payload, instantResultsRegisterDefinition)
+);
+
+export const updateInstantResultsQuery = createAction(
+  'instantResults/updateQuery',
+  (payload: UpdateInstantResultQueryActionCreatorPayload) =>
+    validatePayload(payload, instantResultsQueryDefinition)
+);
+
+export const clearExpiredResults = createAction(
```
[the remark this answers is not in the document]
liviarett: definitely! thanks

PR #1987, opened 2022-05-13
fix(atomic): highlight encoding issue w/ result-text & no-results

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1628

### DIFF
--- a/packages/atomic/src/components/atomic-no-results/atomic-no-results.tsx
+++ b/packages/atomic/src/components/atomic-no-results/atomic-no-results.tsx
@@ -67,7 +67,7 @@ export class AtomicNoResults {
   private wrapHighlight(content: string) {
     return `<span class="font-bold" part="highlight">${this.bindings.i18n.t(
       'between-quotations',
-      {text: content}
+      {text: escape(content), interpolation: {escapeValue: false}}
     )}</span>`;
   }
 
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/result-template-components/atomic-result-text/atomic-result-text.tsx
```diff
@@ -53,7 +52,7 @@ export class AtomicResultText implements InitializableComponent {
         closingDelimiter,
         highlights,
       });
-      const innerHTML = escape(highlightedValue)
+      const innerHTML = highlightedValue
```
ThibodeauJF, line 55: The HighlightUtils of headless already escape the value. We were doing it twice

packages/atomic/src/components/atomic-no-results/atomic-no-results.tsx
```diff
@@ -67,7 +67,7 @@ export class AtomicNoResults {
   private wrapHighlight(content: string) {
     return `<span class="font-bold" part="highlight">${this.bindings.i18n.t(
       'between-quotations',
-      {text: content}
+      {text: escape(content), interpolation: {escapeValue: false}}
```
ThibodeauJF, line 70: escaping the value

btaillon (approved):

olamothe (approved):

PR #1988, opened 2022-05-13
style(atomic,headless): enforce curly eslint rule

olamothe:
Fix as well some eslint warning
https://coveord.atlassian.net/browse/KIT-1634

### DIFF
--- a/packages/atomic/.eslintrc.json
+++ b/packages/atomic/.eslintrc.json
@@ -5,6 +5,9 @@
   "env": {
     "browser": true
   },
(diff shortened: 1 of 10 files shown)


liviarett (approved):
🎉

ThibodeauJF (approved):
Could we put it in the "root" eslintrc or will that piss off the Quantic team? 🤔  @nathanlb

nathanlb:
I think we're already compliant but I'd say try it and see if quantic becomes a sea of red.

PR #1991, opened 2022-05-13
fix(headless): add more defensive code for folding + handle parentResult in createCollections

olamothe:
https://coveord.atlassian.net/browse/KIT-1637

### DIFF
--- a/packages/headless/src/features/folding/folding-slice.ts
+++ b/packages/headless/src/features/folding/folding-slice.ts
@@ -9,6 +9,7 @@ import {
   FoldedResult,
   FoldingFields,
   getFoldingInitialState,
+  FoldingState,
 } from './folding-state';
 import {getAllIncludedResultsFrom} from './folding-utils';
 
@@ -119,6 +120,13 @@ function createCollections(
     if (!collectionId) {
       return;
     }
+    if (result.parentResult) {
+      collections[collectionId] = createCollectionFromResult(
+        result,
+        fields,
+        result.parentResult
+      );
+    }
     if (!getChildField(result, fields)) {
       return;
     }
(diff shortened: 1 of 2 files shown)


packages/headless/src/features/folding/folding-slice.ts
```diff
@@ -119,6 +120,13 @@ function createCollections(
     if (!collectionId) {
       return;
     }
+    if (result.parentResult) {
+      collections[collectionId] = createCollectionFromResult(
+        result,
+        fields,
+        result.parentResult
```
btaillon, line 127: It looks like this would break https://coveord.atlassian.net/browse/KIT-1569.

I think what you're looking for is to resolve the parent if it exists instead of the relevant result when the root cannot be resolved using the fields. If so, I think you should replace:
```ts
const resultToUseAsRoot =
  rootResult ??
  resolveRootFromFields(resultsInCollection, fields) ??
  relevantResult;
```
with:
```ts
const resultToUseAsRoot =
  rootResult ??
  resolveRootFromFields(resultsInCollection, fields) ??
  relevantResult.parentResult ??
  relevantResult;
```

liviarett: I should have written some tests for that!

ThibodeauJF (approved):

packages/headless/src/features/folding/folding-slice.ts
```diff
@@ -94,6 +95,7 @@ function createCollectionFromResult(
   const resultToUseAsRoot =
     rootResult ??
     resolveRootFromFields(resultsInCollection, fields) ??
+    relevantResult.parentResult ??
```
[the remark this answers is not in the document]
olamothe: Good idea.

PR #1995, opened 2022-05-18
fix(atomic): clone child nodes for inherited templates in children

liviarett:
https://coveord.atlassian.net/browse/KIT-1645

When using sections, grandchildren weren't actually rendering anything because when I cloned the children for inherited templates, I only got the top level stuff, meaning that we would render the `*-section` component, but nothing that was inside it


olamothe (approved):

ThibodeauJF (approved):
💯

PR #1996, opened 2022-05-18
feat(headless): add Insight Panel API client

lbergeron:
This PR adds an API client allowing to feed an Insight Panel container component. In other words, it provides the calls required to use an existing Insight Panel configuration and interface.

* get the Insight configuration & interface
* perform contextual search queries
* retrieve user actions


packages/headless/src/api/service/insight/user-actions/user-actions-request.ts
```diff
@@ -0,0 +1,55 @@
...
+} from '../insight-params';
+
+export type InsightUserActionsRequest = InsightParam &
+  TicketCreationDateParam &
+  NumberSessionsBeforeParam &
+  NumberSessionsAfterParam &
+  MaximumSessionInactivityMinutesParam &
+  ExcludedCustomActionsParam;
+
+interface TicketCreationDateParam {
```
olamothe, line 14: Feel free to do what you think is best for your use case, but you don't necessarily need to"explode" all parameters like that in various interface.

This is a pattern we use for the `search` params because they are re-used for various use cases (eg: recommendation, product listing). So in that case it's nice to be able to pick and compose different parameters together to build a new type of Request without duplication.

If those parameters are only used for a single route in the Insight Panel API, and not re-used for other services/use cases, then that "explosion" of interface might be overkill.

packages/headless/src/api/service/insight/insight-api-client.ts
```diff
@@ -0,0 +1,106 @@
...
+  async getInterface(
+    req: GetInsightInterfaceRequest
+  ): Promise<InsightAPIResponse<GetInsightInterfaceResponse>> {
+    const response = await PlatformClient.call({
+      ...buildGetInsightInterfaceRequest(req),
+      ...this.options,
+    });
+
+    if (response instanceof Error) {
+      return buildAPIResponseFromErrorOrThrow(response);
```
olamothe, line 62: Since it's a different API (and not a search API error), I'm not certain this will always be the appropriate return type. 

Unless the Insight panel API always copies the Search API error format.

packages/headless/src/api/service/insight/get-interface/get-interface-response.ts
```diff
@@ -0,0 +1,109 @@
...
+  name: string;
+  layout: string;
+  conditions: Condition[];
+  badge: Badge;
+  details: Detail[];
+  resultActions: ResultActions;
+  tags: Tags;
+}
+
+type Condition =
```
olamothe, line 26: I think you could have a generic for this + related interfaces

```typescript

type Condition<ConditionType = 'isDefined' | 'isNotDefined' | 'mustMatch' | 'mustNotMatch'> = {
    field: string;
    conditionType: ConditionType;
} & (ConditionType extends 'mustMatch' | 'mustNotMatch' ? { values: string[] } : {})



// will error if values is not set
const mustMatch: Condition = { field: 'foo', conditionType: 'mustMatch', values: ['a', 'b', 'c'] }

// will error if values is set;
const isNotDefined: Condition = { field: 'foo', conditionType: 'isNotDefined' }
```

packages/headless/src/api/service/insight/insight-api-client.ts
lbergeron: So far the Customer Service API is returning the error body in the same format as the Search API. We're doing the same thing already for the Case Assist API client.

packages/headless/src/api/service/insight/query/query-response.ts
```diff
@@ -0,0 +1,31 @@
...
+  totalCount: number;
+  executionReport?: Record<string, unknown>;
+  facets: Facet[];
+}
+
+interface SearchResult {
+  title: string;
+  clickUri: string;
+  excerpt: string;
+  fields: Record<string, string | number | string[]>; // TODO: make sure those types are correct
```
nathanlb, line 14: Will this TODO be done separately or has this been validated already? This should be the same typing as the SearchAPI Result interface.

nathanlb (approved):
Very nice

packages/headless/src/api/service/insight/query/query-response.ts
lbergeron: I just verified again, and added the missing `boolean` type.

packages/headless/src/api/service/insight/get-interface/get-interface-response.ts
```diff
@@ -0,0 +1,88 @@
+export interface GetInsightInterfaceResponse {
```
btaillon, line 1: If possible, I think using the term `UserInterface` instead of just `Interface` may help distinguish these interfaces from the programmatical concept of interfaces.

nathanlb: or `InsightInterfaceConfig`

lbergeron: I like `InsightInterfaceConfig`, will use that instead.

aelfaleh (approved):

olamothe (approved):

PR #1998, opened 2022-05-18
feat(atomic): made result list outline and divider easily customizable

btaillon:
https://coveord.atlassian.net/browse/KIT-1646

While writing documentation, I realized that customizing the outline and the divider of the result list was complicated and unintuitive, and could easily be broken by our own changes.

I decided to work on this PR before finishing documentation because I wanted to establish how the outline and the divider are going to be customizable.

With this PR, the outline can now be customized like this:
```css
atomic-result-list::part(outline) {
  border-color: red;
}
```
image

And the divider like this:
```css
atomic-result-list::part(outline)::before {
  background-color: blue;
}
```
image

Alternately, I could also add a separate `divider` part just in case we end up with a weird DOM structure in the future.


npushkarskii (approved):
Neat, to say the least! 🤘😎

olamothe (approved):

PR #1999, opened 2022-05-18
feat(atomic): initial implementation of atomic-search-box-internal-results

liviarett:
https://coveord.atlassian.net/browse/KIT-1632
https://user-images.githubusercontent.com/30511433/169099265-edea7132-3e6d-4bd1-908c-860cc08c5a66.mov

An initial implementation  implementation with a little video!

### DIFF
--- a/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
@@ -10,6 +10,8 @@ import {
   StandaloneSearchBox,
   StandaloneSearchBoxState,
   buildStandaloneSearchBox,
+  InstantResults,
+  buildInstantResults,
 } from '@coveo/headless';
 import {
   Bindings,
@@ -56,6 +58,7 @@ export class AtomicSearchBox {
   private inputRef!: HTMLInputElement;
   private listRef!: HTMLElement;
   private querySetActions!: QuerySetActionCreators;
+  private instantResults?: InstantResults;
   private pendingSuggestionEvents: SearchBoxSuggestionsEvent[] = [];
   private suggestions: SearchBoxSuggestions[] = [];
 
@@ -155,6 +158,21 @@ export class AtomicSearchBox {
       return;
     }
     this.pendingSuggestionEvents.push(event.detail);
+  }
+
+  @Listen('atomic/searchBoxInstantResults/register')
+  public registerInstantResults(
+    event: CustomEvent<(instantResults: InstantResults) => void>
+  ) {
+    event.preventDefault();
+    event.stopPropagation();
+    this.instantResults = buildInstantResults(this.bindings.engine, {
+      options: {
+        searchBoxId: this.id,
+        maxResultsPerQuery: 4,
+      },
+    });
+    event.detail(this.instantResults);
   }
 
   @Watch('redirectionUrl')
@@ -265,6 +283,10 @@ export class AtomicSearchBox {
       : this.bindings.i18n.t('query-suggestions-unavailable');
   }
 
+  private updateInstantResultsQuery(q = '') {
+    this.instantResults && this.instantResults.updateQuery(q);
+  }
+
   private async triggerSuggestions() {
     const settled = await Promise.allSettled(
       this.suggestions.map((suggestion) =>
@@ -288,6 +310,14 @@ export class AtomicSearchBox {
       .sort((a, b) => a.position - b.position)
       .map((suggestion) => suggestion.renderItems())
       .flat();
+
+    const defaultSuggestionQ = suggestionElements.find(
+      (suggestion) => suggestion.query
+    )?.query;
+
+    if (defaultSuggestionQ) {
+      this.updateInstantResultsQuery(defaultSuggestionQ);
+    }
 
     const max =
       this.numberOfQueries + suggestionElements.filter(isDividerElement).length;
@@ -457,6 +487,9 @@ export class AtomicSearchBox {
         onClick={() => {
           item.onSelect && item.onSelect();
           this.clearSuggestions();
+        }}
+        onMouseOver={() => {
+          this.updateInstantResultsQuery(suggestion.query);
         }}
         ref={(el) => {
           if (isHTMLElement(item.content)) {
--- a/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
@@ -10,8 +10,7 @@ import {
   StandaloneSearchBox,
   StandaloneSearchBoxState,
   buildStandaloneSearchBox,
-  InstantResults,
-  buildInstantResults,
+  Result,
 } from '@coveo/headless';
 import {
   Bindings,
@@ -32,6 +31,11 @@ import {
 import {AriaLiveRegion} from '../../utils/accessibility-utils';
 import {SafeStorage, StorageItems} from '../../utils/local-storage-utils';
 import {promiseTimeout} from '../../utils/promise-utils';
+import {
+  InstantResultItem,
+  RegisterInstantResultsEvent,
+  UpdateInstantResultsCallback,
+} from '../search-box-suggestions/instant-results-common';
 
 /**
  * The `atomic-search-box` component creates a search box with built-in support for suggestions.
@@ -58,9 +62,10 @@ export class AtomicSearchBox {
   private inputRef!: HTMLInputElement;
   private listRef!: HTMLElement;
   private querySetActions!: QuerySetActionCreators;
-  private instantResults?: InstantResults;
   private pendingSuggestionEvents: SearchBoxSuggestionsEvent[] = [];
+  private pendingInstantResultsEvents: (() => InstantResultItem)[] = [];
   private suggestions: SearchBoxSuggestions[] = [];
+  private instantResultsItems: InstantResultItem[] = [];
 
   @BindStateToController('searchBox')
   @State()
@@ -68,6 +73,8 @@ export class AtomicSearchBox {
   @State() public error!: Error;
   @State() private isExpanded = false;
   @State() private activeDescendant = '';
+
+  @State() private instantResults: Result[][] = [];
   @State() private suggestionElements: SearchBoxSuggestionItem[] = [];
 
   /**
@@ -129,6 +136,7 @@ export class AtomicSearchBox {
         event(this.suggestionBindings)
       )
     );
+    this.pendingInstantResultsEvents.forEach((init) => init());
     this.pendingSuggestionEvents = [];
   }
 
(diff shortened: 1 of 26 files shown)


btaillon:
Was the idea of adding a new event already reviewed? I thought we were going to use the existing suggestion events, and maybe add some more options/methods to it.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -470,28 +516,61 @@ export class AtomicSearchBox {
...
-        {this.suggestionElements.map((suggestion, index) =>
-          this.renderSuggestion(
-            suggestion,
-            index,
-            this.suggestionElements.length - 1
-          )
+        {!!leftElements.length && (
+          <ul
+            id={this.popupId}
+            role="listbox"
```
liviarett, line 548: I'm not quite sure how to go about this aria stuff, never done it before! Could we have a chat tomorrow @btaillon ?

liviarett:
@ThibodeauJF @btaillon what do you guys think now? 🙂

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -230,6 +242,25 @@ export class AtomicSearchBox {
       this.activeDescendantElement?.previousElementSibling || this.lastValue
     );
   }
+  private get showSuggestions() {
+    return this.hasSuggestions && this.isExpanded;
+  }
+
+  private get allSuggestionElements() {
```
btaillon, line 249: I would probably move this into a private variable instead of a getter to avoid calling `renderItems` too many times

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
olamothe: Accessibility if very complex indeed !

For roles, the best resource I know is MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

And then you try to find one that describe best what the element does :D 

I think in this case it would be either... listbox or grid ?

What makes me hesitate for listbox is that it's not "static".

packages/atomic/src/components/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -0,0 +1,97 @@
...
+  private renderItems(): SearchBoxSuggestionItem[] {
+    const results = this.instantResults.state.results.length
+      ? this.instantResults.state.results
+      : this.results;
+    return results.map((result, i) => ({
+      key: `instant-result-${i}`,
+      query: '',
+      content: <div class="flex items-center break-all">{result.title}</div>,
+      onSelect: () => {
+        // TODO: ADD LOGS?
```
olamothe, line 49: I think re-using `LinkWithResultAnalytics` here  (refactored if needed)  should do the trick... 🤔 Totally fine to leave it for another PR !

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -1,6 +1,15 @@
 import SearchIcon from 'coveo-styleguide/resources/icons/svg/search.svg';
```
olamothe, line 1: This whole file is much better ! 💯 

There is nothing specific for instant result, and everything is "generic", so it looks like it's going to be possible for our clients to be able to implement all kinds of advanced suggestion provider on their own.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -286,18 +322,27 @@ export class AtomicSearchBox {
       }
     });
 
-    const suggestionElements = fulfilledSuggestions
-      .sort((a, b) => a.position - b.position)
-      .map((suggestion) => suggestion.renderItems())
-      .flat();
+    this.leftSuggestions = fulfilledSuggestions
```
olamothe, line 325: nitpick: since it's repetitive for both side, just a small inline function should work.

```
const splitSuggestions = (side: 'left' | 'right') => fulfilledSuggestions
      .filter((suggestion) => suggestion.panel === side)
      .sort(this.sortSuggestions);

const left = splitSuggestions('left');
const right = splitSuggestions('right');
```

olamothe (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
btaillon: I think a listbox is fine, so long as we have a single column of results and they aren't displayed as a grid.

My concern right now is the relationship between the two panes. If `rightElements` is strictly always genererated by the focused member of `leftElements`, then perhaps we should establish an [`aria-control`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls) relationship between left pane and the right pane. I have no idea

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -476,28 +527,61 @@ export class AtomicSearchBox {
...
-      <ul
-        id={this.popupId}
-        role="listbox"
-        part="suggestions"
-        aria-label={this.bindings.i18n.t('query-suggestion-list')}
-        ref={(el) => (this.listRef = el!)}
-        class={`w-full z-10 absolute left-0 top-full rounded-md bg-background border border-neutral ${
-          showSuggestions ? '' : 'hidden'
+      <div
+        class={`flex w-full z-10 absolute left-0 top-full rounded-md bg-background border border-neutral ${
```
btaillon, line 552: I would make sure that every element which has styling also has a part

packages/atomic/src/components/atomic-search-interface/analytics-config.ts
```diff
@@ -22,7 +22,6 @@ export function getAnalyticsConfig(
   const defaultConfiguration: AnalyticsConfiguration = {
     analyticsClientMiddleware,
     enabled,
-    documentLocation: document.location.href,
```
btaillon, line 25: Why is this removed?

packages/atomic/src/components/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -0,0 +1,97 @@
...
+      onSuggestedQueryChange: (q) => {
+        this.instantResults.updateQuery(q);
+        return this.onSuggestedQueryChange();
+      },
+      renderItems: () => this.renderItems(),
+    };
+  }
+
+  private onSuggestedQueryChange() {
+    return new Promise((resolve) => {
```
btaillon, line 73: I would make this a void promise so you can call `resolve()` without `null`.
```suggestion
    return new Promise<void>((resolve) => {
```

btaillon (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: yeah, definitely!! I agree 🙂

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: I thought about it but then we don't default to the left panel when not available.. but I guess I could add that into the function too!

packages/atomic/src/components/atomic-search-interface/analytics-config.ts
liviarett: oh, thanks for catching that!! it wasn't supposed to be committed.. I think it was something to do with a bad merge from master and then the linter complaining about this. I'll go through the whole code one more time to make sure there is nothing extra

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: good shout

packages/atomic/tailwind.config.js
```diff
@@ -1,6 +1,7 @@
 const plugin = require('tailwindcss/plugin');
 
 module.exports = {
+  mode: 'jit',
```
ThibodeauJF, line 4: jit is enabled by default in tailwind v3 https://tailwindcss.com/docs/upgrade-guide#migrating-to-the-jit-engine
```suggestion
```

packages/atomic/src/components/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
liviarett: yesss!

packages/atomic/src/components/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -0,0 +1,97 @@
...
+import {
+  dispatchSearchBoxSuggestionsEvent,
+  SearchBoxSuggestionItem,
+  SearchBoxSuggestions,
+  SearchBoxSuggestionsBindings,
+} from '../suggestions-common';
+
+/**
+ * The `atomic-search-box-instant-results` component can be added as a child of an `atomic-search-box` component, allowing for the configuration of instant results behavior.
+ * @internal
```
npushkarskii, line 13: so basically that means that we won't expose this component in the public documentation, right?

It also explains, why this components' parts are not described in this file, but in the [atomic-search-box.tsx](https://github.com/coveo/ui-kit/pull/1999/files#diff-46e3c6b20a52a198d73e328db1b88c9a1b4b44ba84c5e7556011e630fc8f4638)  instead.

If I understand all this correctly, the new component is sort of "service" / "under-the-hood" component.

It all makes sense to me, I just wanted to double-check if it's what we want for our users 😊🙏

@liviarett

npushkarskii (approved):
LGTM! I only left one question, just out of caution 🙏

packages/atomic/src/components/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
liviarett: @npushkarskii it will be exposed actually, but it's still under construction 🚧

npushkarskii: aha, got it, thx!

 If it's going to be exposed, don't we want to specify describe parts of that component in `atomic-search-box-instant-results.tsx`? Pretty much like we did for `atomic-search-box-recent-queries`.

@liviarett 

image

liviarett: ah, so we did this because the recent queries has a header that we wanted to be able to target. but this one doesn't have anything special, at least for now! no parts available there, so it's ok

packages/atomic/src/pages/examples/suggestions.html
```diff
@@ -81,8 +81,9 @@
       <atomic-search-layout>
         <atomic-layout-section section="search">
           <atomic-search-box suggestion-timeout="1000">
-            <custom-suggestions></custom-suggestions>
+            <atomic-did-you-mean></atomic-did-you-mean>
```
ThibodeauJF, line 84: 🤔

ThibodeauJF (approved):

packages/atomic/src/components/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -0,0 +1,97 @@
...
+ */
+@Component({
+  tag: 'atomic-search-box-instant-results',
+  shadow: true,
+})
+export class AtomicSearchBoxRecentQueries {
```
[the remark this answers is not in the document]
liviarett: 🙏  thanks for catching this, I totally missed it! will update in a PR that I have open and should be merged soon!

PR #2011, opened 2022-05-20
fix(atomic): change breakpoints for grid display + nowrap for table

olamothe:
https://coveord.atlassian.net/browse/KIT-1660

### DIFF
--- a/packages/atomic/cypress/integration-screenshots/result-list-screenshots.cypress.ts
+++ b/packages/atomic/cypress/integration-screenshots/result-list-screenshots.cypress.ts
@@ -18,7 +18,7 @@ describe('screenshots for result lists display mode', () => {
   const layouts = ['table', 'list', 'grid'];
   const densities = ['normal'];
   const imageSizes = ['none', 'icon', 'small', 'large'];
-  const viewports = ['iphone-8', 'macbook-16'];
+  const viewports = ['iphone-8', 'ipad-2', 'macbook-16'];
   const combinations: string[][] = cartesianProduct(
     layouts,
     densities,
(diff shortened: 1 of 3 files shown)


packages/atomic/cypress/integration-screenshots/result-list-screenshots.cypress.ts
```diff
@@ -18,7 +18,7 @@ describe('screenshots for result lists display mode', () => {
   const layouts = ['table', 'list', 'grid'];
   const densities = ['normal'];
   const imageSizes = ['none', 'icon', 'small', 'large'];
-  const viewports = ['iphone-8', 'macbook-16'];
+  const viewports = ['iphone-8', 'ipad-2', 'macbook-16'];
```
olamothe, line 21: Adds a bit more screenshots, but we need to test "medium screen size" also...

ThibodeauJF (approved):

packages/atomic/src/components/result-lists/result-list-common.pcss
```diff
@@ -205,10 +205,10 @@
     }
 
     &.image-small {
-      @media not all and (min-width: 1180px) {
+      @media not all and (min-width: 1024px) {
         grid-template-columns: repeat(3, 1fr);
       }
-      @media (min-width: 1180px) {
+      @media (min-width: 1024px) {
```
btaillon, line 211: The default mobile breakpoint in `atomic-search-layout` is 1024px. Setting `1024px` in this file means that it becomes customizable. I don't think that's a problem though? Mentioning it in case anybody else thinks that may be a problem.

btaillon (approved):

PR #2013, opened 2022-05-24
refactor(atomic): add more props validation for facets components

olamothe:
I was getting a bit un-easy for the standardization of all facets props, and I feared we would forget things in the long run. Especially now with the new horizontal facets.

Since it's impossible to have inheritance for Stencil components, and after trying (and failing) to apply a mixin pattern, I've opted to simply add more typings check on `BaseFacet`.

https://coveord.atlassian.net/browse/KIT-1664

### DIFF
--- a/packages/atomic/src/components/facets/facet-common.ts
+++ b/packages/atomic/src/components/facets/facet-common.ts
@@ -32,13 +32,14 @@ export type BaseFacet<FacetType extends AnyFacetType> = {
   NumberOfValuesProp<FacetType> &
   NumberOfIntervalsProp<FacetType> &
   SortCriterionProp<FacetType> &
-  DisplayValuesAsProp;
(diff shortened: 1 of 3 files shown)


packages/atomic/src/components/facets/facet-common.ts
```diff
@@ -5,19 +5,85 @@ import {
...
+  NumberOfIntervalsProp<FacetType> &
+  SortCriterionProp<FacetType> &
+  DisplayValuesAsProp;
+
+type PropsOnAllFacets = {
+  facetId?: string;
   label: string;
   field: string;
-}
+  isCollapsed: boolean;
```
mbenkirane-coveo, line 41: There's no isCollapsed prop on the segmented filter facet

btaillon (approved):

PR #2014, opened 2022-05-24
feat(atomic): add atomic-result-html component

olamothe:
Add two components to output HTML (atomic-html, atomic-result-html), especially useful for results.

Use `DOMPurify` to optionally sanitize the HTML.

Also fixed a bug in Headless with highlight function: When there's no highlights (ie: an empty array is passed to the function), it would return the un-escaped value.

https://coveord.atlassian.net/browse/KIT-1122


btaillon (approved):
looks good! although I'm curious what's the use case for this exactly

olamothe:
> looks good! although I'm curious what's the use case for this exactly

People that want to display field properties from their result which contains HTML.

packages/atomic/src/utils/result-utils.ts
```diff
@@ -67,6 +72,16 @@ export function buildStringTemplateFromResult(
   });
 }
 
+export function getStringValueFromResultORNull(result: Result, field: string) {
```
ThibodeauJF, line 75:
```suggestion
export function getStringValueFromResultOrNull(result: Result, field: string) {
```

ThibodeauJF: +1 add doc

ThibodeauJF (approved):

packages/atomic/src/components.d.ts
```diff
@@ -289,6 +289,16 @@ export namespace Components {
     }
     interface AtomicFrequentlyBoughtTogether {
     }
+    interface AtomicHtml {
+        /**
+          * Specify if the content should be sanitized, using `DOMPurify` (https://www.npmjs.com/package/dompurify).
```
npushkarskii, line 294:
```suggestion
          * Specify if the content should be sanitized using [`DOMPurify`](https://www.npmjs.com/package/dompurify).
```

For this and other occurrences of this phrase 🙏

packages/atomic/src/components.d.ts
```diff
@@ -289,6 +289,16 @@ export namespace Components {
...
+         */
+        "sanitize": boolean;
+        /**
+          * The string value containg HTML to display;
```
npushkarskii, line 298:
```suggestion
          * The string value containing HTML to display;
```

For this and other occurrences of this phrase 🙏

packages/atomic/src/components.d.ts
```diff
@@ -611,6 +621,16 @@ export namespace Components {
     }
     interface AtomicResultFieldsList {
     }
+    interface AtomicResultHtml {
+        /**
+          * The result field which the component should use. This will look in the Result object first, and then in the Result.raw object for the fields. It is important to include the necessary field in the ResultList component.
```
npushkarskii, line 626:
```suggestion
          * The result field which the component should use.
          * If set, Atomic searches for the specified field in the `Result` object first.
          * If there's no such a field, Atomic searches throught the `Result.raw` object.
          * It's important to include the necessary field in the `ResultList` component.
```

For this and other occurrences of this comment🙏

npushkarskii (approved):
Looks just great! 💎 However, I requested a couple of changes 🙏

packages/atomic/src/components/result-template-components/atomic-result-html/atomic-result-html.tsx
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * There is an inherent XSS security concern associated with the usage of this component.
+ *
+ * Use only with fields for which you are certain the data is harmless.
+ */
+@Component({
+  tag: 'atomic-result-html',
+  shadow: false,
+})
+export class AtomicResultHtml implements InitializableComponent {
```
mbenkirane-coveo, line 22: Why would a user use AtomicResultHTML instead of AtomicHTML if it's less safe?

olamothe: Both have the same level of "dangerousness". 

AtomicResultHTML uses AtomicHTML under the hood in its render method.

The difference is that AtomicResultHTML makes it easy to pull values from field results (so, data that is relevance in the context of trying to render result templates), as opposed to a generic "any string" property used by AtomicHTML.

packages/atomic/src/components/result-template-components/atomic-result-html/atomic-result-html.tsx
```diff
@@ -0,0 +1,52 @@
...
+  @InitializeBindings() public bindings!: Bindings;
+  @ResultContext() private result!: Result;
+
+  @Element() private host!: HTMLElement;
+
+  @State() public error!: Error;
+
+  /**
+   * The result field which the component should use.
+   * This will look in the Result object first, and then in the Result.raw object for the fields.
```
liviarett, line 32: maybe add a 'if not found, element won't be included in the dom'?

packages/atomic/src/components/atomic-html/atomic-html.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  public connectedCallback() {
+    if (!this.value) {
+      this.error = new Error('The "value" attribute must be defined.');
+    }
+  }
+
+  public render() {
+    return (
+      <span
```
liviarett, line 41: should we allow to choose the tag?

packages/atomic/src/components/atomic-html/atomic-html.tsx
```diff
@@ -0,0 +1,46 @@
...
+        innerHTML={this.sanitize ? DOMPurify.sanitize(this.value) : this.value}
```
mbenkirane-coveo, line 42: In which cases would a user not want to sanitize their html?

packages/atomic/src/components/result-template-components/atomic-result-html/atomic-result-html.tsx
mbenkirane-coveo: I see !

packages/atomic/src/components/atomic-html/atomic-html.tsx
olamothe: If they want to display a source they really trust "as is". In the scenario that DOMPurify would improperly sanitize something that should not have been.

packages/atomic/cypress/integration/result-list/result-components/result-html.cypress.ts
```diff
@@ -0,0 +1,113 @@
...
+        )
+        .withCustomResponse((response) =>
+          response.results.forEach((result) => {
+            result.raw[field] = rawValue;
+          })
+        )
+        .init();
+    }
+
+    describe('when the "sanitize" prop is true', () => {
```
mbenkirane-coveo, line 85: Would it be possible to implement some sort of unit test to see if the html is getting properly sanitized?

olamothe: Yes, this is what this test does in fact:  It verifies that it sanitize the `onerror` property of the image on line 67

PR #2023, opened 2022-05-25
fix(quantic): formattingFunction passed correctly to duplicated numeric and date facets

mmitiche:
[SFINT-4525](https://coveord.atlassian.net/browse/SFINT-4525)
When the numeric/date facet is in a search interface and the formattingFunction property  is not specified, the QuanticRefineModalContent fails to attribute the default formattingFunction method that should be used to the duplicated numeric/date facet.
QuanticRefineModalContent can't access the default formatting function defined inside the numeric/date facet.

err

The solution is to get the `formattingFunction` from the Quantic Store instead of trying to get it from the facet element directly.

QuanticStore

The formattingFunction is always saved when we register a numeric/date facet to the quantic store.


lbergeron (approved):
It would be nice to have this change covered by tests as well.

nathanlb (approved):

erocheleau (approved):

aelfaleh (approved):

PR #2024, opened 2022-05-25
fix(atomic): sanitize facet value to use them as parts

olamothe:
https://coveord.atlassian.net/browse/KIT-1264

### DIFF
--- a/packages/atomic/src/components/facets/atomic-color-facet/atomic-color-facet.tsx
+++ b/packages/atomic/src/components/facets/atomic-color-facet/atomic-color-facet.tsx
@@ -296,9 +296,7 @@ export class AtomicColorFacet
       this.bindings.i18n
     );
     const isSelected = facetValue.state === 'selected';
-    const partValue = displayValue
-      .match(new RegExp('-?[_a-zA-Z]+[_a-zA-Z0-9-]*'))
-      ?.toString();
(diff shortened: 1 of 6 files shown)


packages/atomic/src/components/facets/atomic-color-facet/atomic-color-facet.tsx
```diff
@@ -296,9 +296,7 @@ export class AtomicColorFacet
       this.bindings.i18n
     );
     const isSelected = facetValue.state === 'selected';
-    const partValue = displayValue
-      .match(new RegExp('-?[_a-zA-Z]+[_a-zA-Z0-9-]*'))
-      ?.toString();
+    const partValue = facetValue.value.replace(/[^a-z0-9]/gi, '');
```
olamothe, line 299: This could technically be considered a breaking change, since it was using `displayValue` before, but now is using the actual value coming from the API.

It's just that it does not seem right to be using the display value here ?

You would need one css file for each language you want to support... 😅

btaillon (approved):

packages/atomic/src/components/facets/atomic-color-facet/atomic-color-facet.tsx
olamothe: Will use 2 parts

PR #2025, opened 2022-05-25
feat(atomic): added heading-level prop to facets

btaillon:
https://coveord.atlassian.net/browse/KIT-1676

The alternative was to recommend implementers to add a heading before each facet, but that would cause screen readers to repeat the same heading twice, so I prefer to do this. This also follows the same pattern as for the smart snippet.

When no heading level is set, the only side-effect of this change is that the `span` in headings is now a `div`.

### DIFF
--- a/packages/atomic/src/components/facets/facet-common.ts
+++ b/packages/atomic/src/components/facets/facet-common.ts
@@ -33,7 +33,8 @@ export type BaseFacet<FacetType extends AnyFacetType> = {
   NumberOfIntervalsProp<FacetType> &
   SortCriterionProp<FacetType> &
   DisplayValuesAsProp &
-  CollapsedProp;
+  CollapsedProp &
+  HeadingLevelProp;
 
 type PropsOnAllFacets = {
   facetId?: string;
@@ -87,6 +88,8 @@ type DisplayValuesAsProp = {
 };
 
 type CollapsedProp = {isCollapsed?: boolean};
+
+type HeadingLevelProp = {headingLevel?: number};
 
 export interface FacetValueProps {
   i18n: i18n;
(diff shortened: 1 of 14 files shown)


packages/atomic/src/components/facets/facet-common.ts
```diff
@@ -88,6 +89,8 @@ type DisplayValuesAsProp = {
 
 type CollapsedProp = {isCollapsed?: boolean};
 
+type HeadingLevelProp = {headingLevel?: number};
```
olamothe, line 92: Should it be mandatory on all types of facets ? Even segmented facets ?

If so, it should probably be non optional, and part of the `props on all facets` type.

The goal of these types is to make it obvious when you create a new type of facet what are all the options you should be implementing, so that there's standardization across very similar components.

We should try to make as much option non optional, when it's possible to do so.

packages/atomic/src/components/facets/facet-header/facet-header.tsx
```diff
@@ -4,12 +4,14 @@ import ArrowBottomIcon from 'coveo-styleguide/resources/icons/svg/arrow-bottom-r
...
 import {i18n} from 'i18next';
 import {Button} from '../../common/button';
+import {Heading} from '../../common/heading';
 
 export interface FacetHeaderProps {
   i18n: i18n;
   label: string;
   numberOfSelectedValues: number;
   isCollapsed: boolean;
+  headingLevel?: number;
```
olamothe, line 14: Should probably be non optional here, and have caller of this component pass in `0` if they don't need heading.

olamothe (approved):

packages/atomic/src/components/facets/facet-common.ts
btaillon: I don't think it should be mandatory for segmented facets

npushkarskii (approved):
Looks good! 🎉

ThibodeauJF (approved):

PR #2029, opened 2022-05-26
fix(atomic): changed category facet lists to nested lists

btaillon:
https://coveord.atlassian.net/browse/KIT-1677

This didn't really change anything with all the screen readers I tested, apart from VoiceOver on Mac OS with Firefox (not Safari), which read out the nesting level every time we focus on a button.

To avoid breaking existing implementations that might've used `display: flex` or `display: grid` on any of the parts, I added `display: contents` on every `ul` and `li` element whose sole purpose is nesting.

This change is slightly breaking, since:
* `[part="parents"]` now wraps `[part="values"]`
* `[part="parent-active"]` used to only wrap `[part="active-parent"]`, but now also wraps `[part="values"]`

I didn't add E2E tests yet, since this is a pretty big change and I don't want to waste time if we choose not to do this. If you guys like those changes, I'll add E2E tests and re-request reviews.

Visually, the component looks the exact same. Nothing has changed.

# Hierarchy comparison
image

## Before
* `ul[part="parents"]`
  * `li > button >` All categories
  * `li > button >` Computers & Tablets
  * `li > button >` Computer Accessories & Peripherals
  * `li > button >` Laptop Accessories
  * `li[part="parent-active"] > button >` Laptop Bags & Cases
* `ul[part="values"]`
  * `li > button >` Laptop Backpacks
  * `li > button >` Laptop Sleeves
  * `li > button >` Laptop Briefcases
  * `li > button >` Laptop Hard-Shell Cases
  * `li > button >` Laptop Messenger Bags

## After
* `ul[part="parents"] > li`
  * `button >` All categories
  * `ul > li`
    * `button >` Computers & Tablets
    * `ul > li`
      * `button >` Computer Accessories & Peripherals
      * `ul > li`
        * `button >` Laptop Accessories
        * `ul > li[part="parent-active"]`
          * `button >` Laptop Bags & Cases
          * `ul[part="values"]`
            * `li > button >` Laptop Backpacks
            * `li > button >` Laptop Sleeves
            * `li > button >` Laptop Briefcases
            * `li > button >` Laptop Hard-Shell Cases
            * `li > button >` Laptop Messenger Bags

### DIFF
--- a/packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
+++ b/packages/atomic/src/components/facets/atomic-category-facet/atomic-category-facet.tsx
@@ -1,4 +1,4 @@
-import {Component, h, State, Prop, Element} from '@stencil/core';
+import {Component, h, State, Prop, Element, Fragment} from '@stencil/core';
 import {
   CategoryFacet,
   buildCategoryFacet,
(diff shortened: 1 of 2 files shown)


btaillon:
> By a breaking change, you mean that the CSS is affected and that the parts now react/appear in a wrong way?

I mean that if, for instance, somebody decided to add a border around the `parents` part, it now appears around both the parents and the children.

ThibodeauJF:
I'd say since it comes with possible breaking changes & we don't gain anything but stylepoints regarding accessibility... let's only do this when we bump to v2

Moreover, none of these "lists" will have more than 1 element, they're not really nested list just a list of parents/grandparents & a list of children:
Screen Shot 2022-05-26 at 3 43 09 PM

PR #2030, opened 2022-05-26
fix(atomic): better logic for display of input on facet-range 

olamothe:
https://coveord.atlassian.net/browse/KIT-1458

### DIFF
--- a/packages/atomic/src/components/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
+++ b/packages/atomic/src/components/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
@@ -3,7 +3,6 @@ import {
   NumericFacet,
   buildNumericFacet,
   NumericFacetState,
-  NumericFacetOptions,
   RangeFacetSortCriterion,
   SearchStatus,
   SearchStatusState,
@@ -102,7 +101,7 @@ export class AtomicNumericFacet
   private manualRanges: NumericRangeWithLabel[] = [];
   private formatter: NumberFormatter = defaultNumberFormatter;
 
-  @BindStateToController('facet')
+  @BindStateToController('facetForRange')
   @State()
   public facetState!: NumericFacetState;
   @BindStateToController('filter')
@@ -112,7 +111,7 @@ export class AtomicNumericFacet
   @State()
   public searchStatusState!: SearchStatusState;
   @State() public error!: Error;
-  @BindStateToController('facetForCountsOnly')
+  @BindStateToController('facetForInput')
   @State()
   public facetForInputState?: NumericFacetState;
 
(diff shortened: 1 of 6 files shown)


packages/atomic/cypress/integration/facets/numeric-facet/numeric-facet.cypress.ts
```diff
@@ -595,10 +595,14 @@ describe('Numeric Facet V1 Test Suites', () => {
         describe('with no facet results', () => {
           const baseSetup = () =>
             visibilitySetup().withCustomResponse((r) => {
-              const idx = r.facets.findIndex(
-                (facet) => facet.field === numericFacetField
-              );
-              r.facets[idx].values = [];
+              r.facets.forEach((facet, i) => {
```
olamothe, line 598: Test change needed since the internal logic is different. We need to "find" the new facet with the generated random ID, and then make only that one return no results.

packages/atomic/src/components/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
```diff
@@ -216,21 +226,55 @@ export class AtomicNumericFacet
     this.dependenciesManager?.stopWatching();
   }
 
-  private initializeFacet() {
+  private initializeFacets() {
```
btaillon, line 229: Unfortunately, I think having two controllers of the same type add a bit too much complexity. Until I actually read what the original issue was about, I had trouble figuring out why there was two controllers.

I think that either we should [extend a custom controller](https://docs.coveo.com/en/headless/latest/usage/extend-headless-controllers/) internal to Atomic, or build that functionality in Headless somehow.

ThibodeauJF: In the end there is a need for 2 facet requests, 2 controllers is the way to go. Having a controller registering 2 facets is also pretty complex

packages/atomic/src/components/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
```diff
@@ -105,6 +112,9 @@ export class AtomicNumericFacet
   @State()
   public searchStatusState!: SearchStatusState;
   @State() public error!: Error;
+  @BindStateToController('facetForCountsOnly')
+  @State()
+  public facetForCountsOnlyState?: NumericFacetState;
```
ThibodeauJF, line 117: recommend: facetForInput

packages/atomic/src/components/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
```diff
@@ -217,21 +227,55 @@ export class AtomicNumericFacet
...
+    // A second facet is initialized only to verify the results count. It is never used to display results to end user.
+    // It serves as a way to determine if the input should be rendered or not, independent of the ranges (manual or automatic) configured in the component
+    if (this.numberOfValues > 0) {
+      const {facet: facetForDisplay, facetId: facetIDForDisplay} =
+        this.initializeFacetForDisplay();
+      this.facet = facetForDisplay;
+      this.facetId = facetIDForDisplay;
+    }
+    if (this.withInput) {
+      const {facet: facetForResultsMatch} = this.initializeFacetForCountsOnly();
```
ThibodeauJF, line 243: would use the same naming to remove confusion

ThibodeauJF (approved):

PR #2033, opened 2022-05-27
feat(atomic): change dayjs language when i18n language changes (DONOTREVIEW)

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1610


btaillon:
In a nutshell, what are the differences between this PR and https://github.com/coveo/ui-kit/pull/2034/files? Are the changes in the sample due to a breaking change? Do either PRs allow custom defining locales?

packages/atomic/scripts/copy-dayjs-locales.js
```diff
@@ -0,0 +1,24 @@
...
+const fs = require('fs');
+const rm = util.promisify(fs.rm);
+const readFile = util.promisify(fs.readFile);
+const writeFile = util.promisify(fs.writeFile);
+
+async function copyDayjsLocales() {
+  const localesJSONData = await readFile('node_modules/dayjs/locale.json');
+  const localesData = JSON.parse(localesJSONData);
+  let fileContent =
+    'export const locales: Record<string, () => Promise<unknown>> = {';
```
olamothe, line 11: What would we gain (if anything) if there was a proper type instead of `unknown` here ?

packages/atomic/src/components/result-template-components/result-template-decorators.tsx
```diff
@@ -55,6 +55,14 @@ export function ResultContext(opts: {folded: boolean} = {folded: false}) {
       return connectedCallback && connectedCallback.call(this);
     };
 
+    component.componentWillRender = function () {
```
olamothe, line 58: Why is this now needed ?

ThibodeauJF: Found a bug (error in console) where something was executed in this method where as the component did not render since there was an error.

Just a slight improvement

PR #2035, opened 2022-05-30
feat(quantic) pre/post install npm scripts added

mmitiche:
[SFINT-4542](https://coveord.atlassian.net/browse/SFINT-4542)
Pre-install and post-install scripts created and added to the npm scripts.

### 1- The pre-install script will do the following task:
- Check if the project where Quantic is being installed in an SFDX project. if it's the case we continue the installation otherwise we cancel it.

### 2- The post-install script will do the following task:
- Read the package directory from the `sfdx-project.json` file.
- Copy Quantic from the node_modules into the package directory.

### DIFF
--- a/packages/quantic/check-sfdx-project.js
+++ b/packages/quantic/check-sfdx-project.js
@@ -1,0 +1,24 @@
+const pathlib = require('path');
+const fs = require('fs');
+
+function getGrandParentPath(path) {
+  const directories = path.split(pathlib.sep);
+  return directories.slice(0, directories.length - 2).join(pathlib.sep);
+}
+
+function main() {
+  try {
+    const projectDirectory = getGrandParentPath(__dirname);
+    fs.accessSync(
+      pathlib.join(projectDirectory, 'sfdx-project.json'),
+      fs.constants.F_OK
+    );
+  } catch (err) {
+    console.error(
+      'Quantic can only be installed in SFDX projects, please refer to this link for more information on how to create an SFDX project: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_create_new.htm'
+    );
+    process.exit(1);
+  }
+}
+
+main();
--- a/packages/quantic/package.json
+++ b/packages/quantic/package.json
@@ -31,7 +31,9 @@
     "deploy:examples": "node_modules/.bin/sfdx force:source:deploy -u LWC -p force-app/examples",
     "doc:generate": "mkdir -p docs/out; jsdoc -c jsdoc-config.json > docs/out/quantic-docs.json",
     "package:create": "node_modules/.bin/ts-node scripts/build/create-package.ts --remove-translations",
-    "package:create:publish": "node_modules/.bin/ts-node scripts/build/create-package.ts --remove-translations --promote"
+    "package:create:publish": "node_modules/.bin/ts-node scripts/build/create-package.ts --remove-translations --promote",
+    "preinstall": "node check-sfdx-project.js",
+    "postinstall": "node setup-quantic.js"
   },
   "dependencies": {
     "@coveo/headless": "1.66.3"
--- a/packages/quantic/package.json
+++ b/packages/quantic/package.json
@@ -63,5 +63,15 @@
     "ts-node": "10.7.0",
     "wait-on": "6.0.1",
     "xml2json": "0.12.0"
-  }
+  },
+  "files": [
+    "force-app/main/default/classes/*",
+    "force-app/main/default/labels/*",
+    "force-app/main/default/lwc/*",
+    "force-app/main/default/cspTrustedSites/*",
+    "force-app/main/default/staticresources/*",
+    "force-app/main/translations/*",
+    "setup-quantic.js",
+    "check-sfdx-project.js"
+  ]
 }
--- a/packages/quantic/setup-quantic.js
+++ b/packages/quantic/setup-quantic.js
@@ -1,0 +1,108 @@
+const fs = require('fs');
+const pathlib = require('path');
+
+function copyFileSync(source, target) {
+  try {
+    let targetFile = target;
+    if (fs.existsSync(target) && fs.lstatSync(target).isDirectory()) {
+      targetFile = pathlib.join(target, pathlib.basename(source));
+    }
+
+    fs.writeFileSync(targetFile, fs.readFileSync(source));
+  } catch (err) {
+    console.error(`Failed to copy file at the following path ${source}`);
+    process.exit(1);
+  }
+}
+
+function copyFolderRecursiveSync(source, target) {
+  try {
+    const targetFolder = pathlib.join(target, pathlib.basename(source));
+    if (!fs.existsSync(targetFolder)) {
+      fs.mkdirSync(targetFolder);
+    }
+
+    if (fs.lstatSync(source).isDirectory()) {
+      const files = fs.readdirSync(source);
+      files.forEach(function (file) {
+        let sourceFilePath = pathlib.join(source, file);
+        if (fs.lstatSync(sourceFilePath).isDirectory()) {
+          copyFolderRecursiveSync(sourceFilePath, targetFolder);
+        } else {
+          copyFileSync(sourceFilePath, targetFolder);
+        }
+      });
+    }
+  } catch (err) {
+    console.error(`Failed to copy folder at the following path ${source}`);
+    process.exit(1);
+  }
+}
+
+function createQuanticDirectory(path) {
+  try {
+    const quanticDirectoryPath = pathlib.join(path, 'quantic');
+    if (fs.existsSync(quanticDirectoryPath)) {
+      fs.rmdirSync(quanticDirectoryPath, {recursive: true});
+    }
+    fs.mkdirSync(quanticDirectoryPath);
+  } catch (err) {
+    console.error('Failed to create Quantic directory');
+    process.exit(1);
+  }
+}
+
+function getPackageDirectory(projectDirectory) {
+  try {
+    const sfdxProjectJson = fs.readFileSync(
+      pathlib.join(projectDirectory, 'sfdx-project.json')
+    );
+    const sfdxProjectObject = JSON.parse(sfdxProjectJson);
+    const defaultPackageDirectory =
+      sfdxProjectObject.packageDirectories.find(
+        (directory) => directory.default
+      ) || sfdxProjectObject.packageDirectories[0];
+    return defaultPackageDirectory.path;
+  } catch (err) {
+    console.error('Failed to get the default package directory.');
+    process.exit(1);
+  }
+}
+
+function getGrandParentPath(path) {
+  const directories = path.split(pathlib.sep);
+  return directories.slice(0, directories.length - 2).join(pathlib.sep);
+}
+
+function main() {
+  try {
+    const sourceDirectory = pathlib.join(
+      __dirname,
+      'force-app',
+      'main',
+      'default'
+    );
+    const translationsDirectory = pathlib.join(
+      __dirname,
+      'force-app',
+      'main',
+      'translations'
+    );
+    const projectDirectory = getGrandParentPath(__dirname);
+    const defaultPackageDirectory = getPackageDirectory(projectDirectory);
+    const defaultPackagePath = pathlib.join(
+      projectDirectory,
+      defaultPackageDirectory
+    );
+    const quanticDirectoryPath = pathlib.join(defaultPackagePath, 'quantic');
+
+    createQuanticDirectory(defaultPackagePath);
+    copyFolderRecursiveSync(sourceDirectory, quanticDirectoryPath);
+    copyFolderRecursiveSync(translationsDirectory, quanticDirectoryPath);
+  } catch (err) {
+    console.error('Failed to setup Quantic.');
+    process.exit(1);
+  }
+}
+
+main();


erocheleau:
Did you forget to include the package-lock?

packages/quantic/check-sfdx-project.js
```diff
@@ -0,0 +1,24 @@
+const pathlib = require('path');
+const fs = require('fs');
+
+function getGrandParentPath(path) {
+  const directories = path.split(pathlib.sep);
+  return directories.slice(0, directories.length - 2).join(pathlib.sep);
```
erocheleau, line 6: So this expect to be in `./node_modules/Quantic` ?

Could be a good idea to add this as a comment for future simplicity if this every breaks?

packages/quantic/setup-quantic.js
```diff
@@ -0,0 +1,108 @@
...
+      defaultPackageDirectory
+    );
+    const quanticDirectoryPath = pathlib.join(defaultPackagePath, 'quantic');
+
+    createQuanticDirectory(defaultPackagePath);
+    copyFolderRecursiveSync(sourceDirectory, quanticDirectoryPath);
+    copyFolderRecursiveSync(translationsDirectory, quanticDirectoryPath);
+  } catch (err) {
+    console.error('Failed to setup Quantic.');
+    process.exit(1);
```
erocheleau, line 104: What happens when a postinstall script fails on npm install?

packages/quantic/setup-quantic.js
```diff
@@ -0,0 +1,108 @@
...
+      'main',
+      'translations'
+    );
+    const projectDirectory = getGrandParentPath(__dirname);
+    const defaultPackageDirectory = getPackageDirectory(projectDirectory);
+    const defaultPackagePath = pathlib.join(
+      projectDirectory,
+      defaultPackageDirectory
+    );
+    const quanticDirectoryPath = pathlib.join(defaultPackagePath, 'quantic');
```
erocheleau, line 97: Should really add the "expected" values of these as comments over each line IMO so if someone has trouble with this, they can see/understand what is going wrong.

mmitiche:
> Did you forget to include the package-lock?

nope, why? are you saying this because of the build failure?  cause this is happening because apparently the post/pre install scripts run even when we do `npm install` to install the local dependencies of Quantic and we dont want that we want the scripts to be executed only when Quantic is used as an external dependency.
I need to update my scripts to make them run only when they detect that Quantic is used as a dependency.

packages/quantic/setup-quantic.js
```diff
@@ -0,0 +1,108 @@
...
+    process.exit(1);
+  }
+}
+
+function createQuanticDirectory(path) {
+  try {
+    const quanticDirectoryPath = pathlib.join(path, 'quantic');
+    if (fs.existsSync(quanticDirectoryPath)) {
+      fs.rmdirSync(quanticDirectoryPath, {recursive: true});
+    }
```
erocheleau, line 47: Should this really delete the repository that already exists?
I don't exactly know what should be done instead but let's assume there was a fuckup or something with the folder this script found/chose, deleting it might cause issues.

Is it possible perhaps to just copy/overwrite what's in the folder instead?

I'd be okay overwriting files we're supposed to control, but let's say someone has made custom quantic components and added them to this folder, this would delete them.

erocheleau (commented):
Some changes for clarity that I think would be nice

packages/quantic/setup-quantic.js
```diff
@@ -0,0 +1,108 @@
...
+      'main',
+      'default'
+    );
+    const translationsDirectory = pathlib.join(
+      __dirname,
+      'force-app',
+      'main',
+      'translations'
+    );
+    const projectDirectory = getGrandParentPath(__dirname);
```
lbergeron, line 91: Can you please share an example of the directory structure of the target working copy?
I'd like to see where each quantic directory goes in the target directory structure.

packages/quantic/setup-quantic.js
mmitiche: When the post-install  script fails all the installation is canceled and the package is automatically deleted.

packages/quantic/check-sfdx-project.js
mmitiche: yeah exactly! I'll add comments to improve this 👌

packages/quantic/setup-quantic.js
mmitiche: Good point👌 I updated this to avoid deleting anything and to just overwrite the files we're supposed to control.

packages/quantic/setup-quantic.js
mmitiche: sure!

packages/quantic/check-sfdx-project.js
```diff
@@ -0,0 +1,24 @@
...
+}
+
+function main() {
+  try {
+    const projectDirectory = getGrandParentPath(__dirname);
+    fs.accessSync(
+      pathlib.join(projectDirectory, 'sfdx-project.json'),
+      fs.constants.F_OK
+    );
+  } catch (err) {
```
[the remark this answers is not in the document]
mmitiche: Logging  error added 👌

packages/quantic/check-sfdx-project.js
```diff
@@ -0,0 +1,24 @@
+const pathlib = require('path');
+const fs = require('fs');
+
+function getGrandParentPath(path) {
```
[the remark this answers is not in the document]
mmitiche: We need to move up exactly 2 directory level because the scripts will be executed: project_name/node_modules/quantic.
So to get the project name path we take two steps up.

packages/quantic/setup-quantic.js
mmitiche: Yes sure! you can find in this repo: https://github.com/mmitiche/example-project-quantic an example sfdx project where I installed my test package containing quantic.
Notice that a quantic folder is created under the force-app directory containing all the Quantic LWCs, classes, labels, static resources, translations and csp trusted sites.

packages/quantic/check-sfdx-project.js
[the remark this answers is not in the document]
mmitiche: Apparently `path.join` doesn’t accept an array as a parameter, only strings are accepted.

nathanlb:
Could the two new scripts be in a scripts/ folder instead of the root of the package?

packages/quantic/check-sfdx-project.js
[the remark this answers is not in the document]
lbergeron: I'm concerned about moving up a fixed number of levels here. There are many tools out there that can install packages from npmjs, and I am not sure if we can assume the directory structure.

Would it be possible to try and detect the `sfdx-project.json` file in a more dynamic way? For instance, maybe we could define the "project directory" as the one that contains the `sfdx-project.json` file AND the `node_modules` directory. So if we find a directory containing the `node_modules` subdirectory, but not the `sfdx-project.json` file, then the project is not compliant.

If we go this route, we'd have to be careful not to introduce a "directory traversal" issue though.

Long story short, we should try with a few package manager tools to see how portable the current implementation is.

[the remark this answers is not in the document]
mmitiche: What do you think of having this function instead of moving up a fixed number of levels:
```
function getProjectPath() {
    return __dirname.split('/node_modules')[0];
}
```
this will satisfy the definition you mentioned of  the "project directory".
and this will address the case where for example quantic would be installed by another package. For example we would still get the correct path to the project when being in the following path:
`/home/project/node_modules/pkg1/node_modules/quantic/pre-install-script`

mmitiche:
> Could the two new scripts be in a scripts/ folder instead of the root of the package?

Sure yeah! I'll update this.

lbergeron:
There is one caveat I just though of regarding the copy of Quantic files into the working directory. This should work well when installing a published Quantic package. However, during development, `npm link` is commonly used when implementing changes in Quantic and a dependent project at the same time (e.g., to validate using a new Quantic feature in a test project). Copying the Quantic files, won't work for this use case.

We might not want to address that right now, but using symbolic links could be helpful.

packages/quantic/setup-quantic.js
```diff
@@ -0,0 +1,132 @@
...
+      // path to source directory: project_name/node_modules/@coveo/quantic/force-app/main/default
+      const sourceDirectory = pathlib.join(
+        __dirname,
+        'force-app',
+        'main',
+        'default'
+      );
+
+      // path to translations directory: project_name/node_modules/@coveo/quantic/force-app/main/translations
+      const translationsDirectory = pathlib.join(
```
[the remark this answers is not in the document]
mmitiche: Okay thanks! so including the translations in the npm package is not needed, I'll remove them then👌

packages/quantic/check-sfdx-project.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function getProjectPath(path) {
+  const directories = path.split(pathlib.sep);
+  return directories.slice(0, directories.length - 3).join(pathlib.sep);
+}
+
+function main() {
+  try {
+    if (quanticIsDependency()) {
```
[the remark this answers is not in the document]
mmitiche: This condition is not to check if the directory structure is as expected but it's just to check if the pre/post install scripts are executed because a user installed quantic as a dependency or because of running `npm install` in the quantic repo. 
In this second case no need to execute the pre/post install scripts so we just skip them. I dont know if an `else block` is needed in this case to log something  to the user.

packages/quantic/scripts/npm/check-sfdx-project.js
```diff
@@ -0,0 +1,37 @@
+const pathlib = require('path');
+const fs = require('fs');
+
+function quanticIsDependency() {
+  const upperDirectories = __dirname.split(pathlib.sep);
+  const firstUpperDirectories = upperDirectories.slice(
+    upperDirectories.length - 5
```
nathanlb, line 7: for future readers it would be nice if you could put `5` in a constant with a descriptive name.

nathanlb (approved):
Looks good! Have you been able to test run these scripts in a mock installation process?

packages/quantic/package.json
```diff
@@ -63,5 +65,13 @@
...
-  }
+  },
+  "files": [
+    "force-app/main/default/classes/*",
+    "force-app/main/default/labels/*",
+    "force-app/main/default/lwc/*",
+    "force-app/main/default/cspTrustedSites/*",
+    "force-app/main/default/staticresources/*",
+    "scripts/npm"
+  ]
```
louis-bompart, line 76: Can you show the result of `npm pack` prior to this change and after this change ;)

packages/quantic/scripts/npm/check-sfdx-project.js
```diff
@@ -0,0 +1,37 @@
...
+  );
+  return (
+    firstUpperDirectories.join(pathlib.sep) ===
+    pathlib.join('node_modules', '@coveo', 'quantic', 'scripts', 'npm')
+  );
```
louis-bompart, line 12: I think this is a bit too complex and error prone, being based on the position of the file.

Instead, use https://nodejs.org/api/modules.html#requireresolverequest-options and trycatch the `MODULE_NOT_FOUND` error.
If the error is thrown, then return false, otherwise, after the require.resolve, return true.
Tada.

packages/quantic/scripts/npm/setup-quantic.js
```diff
@@ -0,0 +1,123 @@
+const fs = require('fs');
+const pathlib = require('path');
+
+function copyFileSync(source, target) {
+  try {
+    let targetFile = target;
+    if (fs.existsSync(target) && fs.lstatSync(target).isDirectory()) {
```
louis-bompart, line 7: Extract this condition in a function and name it, future y'all will be happy.

packages/quantic/scripts/npm/setup-quantic.js
```diff
@@ -0,0 +1,123 @@
...
+      ) || sfdxProjectObject.packageDirectories[0];
+    return defaultPackageDirectory.path;
+  } catch (err) {
+    console.error('Failed to get the default package directory.');
+    console.error(err);
+    process.exit(1);
+  }
+}
+
+function quanticIsDependency() {
```
louis-bompart, line 75: Don't Repeat Yourself. :P

Create a utils file for your scripts.

packages/quantic/scripts/npm/setup-quantic.js
```diff
@@ -0,0 +1,123 @@
...
+
+    fs.writeFileSync(targetFile, fs.readFileSync(source));
+  } catch (err) {
+    console.error(`Failed to copy file at the following path ${source}`);
+    console.error(err);
+    process.exit(1);
+  }
+}
+
+function copyFolderRecursiveSync(source, target) {
```
louis-bompart, line 19: PostInstall has access to npm dependencies. So save yourself some trouble and find a dependency that does that for you, fs-extra prolly would do the trick

packages/quantic/scripts/npm/setup-quantic.js
```diff
@@ -0,0 +1,123 @@
...
+    process.exit(1);
+  }
+}
+
+function getPackageDirectory(projectDirectory) {
+  try {
+    const sfdxProjectJson = fs.readFileSync(
+      pathlib.join(projectDirectory, 'sfdx-project.json')
+    );
+    const sfdxProjectObject = JSON.parse(sfdxProjectJson);
```
louis-bompart, line 62: IIRC you can require JSON files.

packages/quantic/scripts/npm/check-sfdx-project.js
mmitiche: Thanks! I'll try this👌

packages/quantic/scripts/npm/setup-quantic.js
mmitiche: I agree that using a dependency here would make the script much simpler 👌

packages/quantic/package.json
mmitiche: sure!
Before this change it waS basically the whole project:
QuanticPackageBefore
After:
quanticPackage

packages/quantic/scripts/npm/utils.js
```diff
@@ -0,0 +1,36 @@
...
+function quanticIsDependency() {
+  try {
+    // path to npm scripts is expected to be: project_name/node_modules/@coveo/quantic/scripts/npm
+    const upperDirectories = __dirname.split(pathlib.sep);
+    const dirCountAfterNodeModulesParent = 5;
+    const nodeModulesParent = upperDirectories.slice(
+      0,
+      -dirCountAfterNodeModulesParent
+    );
+    require.resolve(
```
lbergeron, line 12: :thinking: Experiencing an error inside the `require.resolve()` invokation would simply return `false` and make other functions just skip over the installation process silently.

We should at least log the error in the `catch` block to help troubleshooting.

lbergeron (approved):
I would add some more error logging, but it looks nice.

packages/quantic/scripts/npm/utils.js
mmitiche: Experiencing an error inside the require.resolve() invocation would mean that we are not in the case where quantic is installed as a dependency but in the case where we just did run the `npm install `command inside the Quantic repo cause doing this also triggers the pre/post install scripts. That's why I didn't log an error cause it's not really an error.
Do you think that we still need to log the error or a message in that case?

lbergeron: Well, you are assuming that the only reason this invocation would fail is if Quantic is not used as a dependency. What if the invocation fails for an unrelated reason like a permission or memory issue?

One way to do it would be to silence the expected error, and log any unexpected error.

packages/quantic/scripts/npm/utils.js
```diff
@@ -0,0 +1,36 @@
+const pathlib = require('path');
+
+function quanticIsDependency() {
```
louis-bompart, line 3: 😕 
Why not
```js
try {
  require.resolve('@coveo/quantic');
  return true;
} catch (e) {
  return false;
}
```

louis-bompart: You can then append the rest of the path to it for example `@coveo/quantic/scripts/utils.js`

This should work.
However I'd say, require the sfdx-project.json. it's location is more guaranteed because it is imposed by salesforce

mmitiche: Yes,  using ` require.resolve('@coveo/quantic/scripts/npm/utils.js')` should indeed achieve what we want 👌
The  sfdx-project.json file is not included in the package so it will not be present under @coveo/quantic thus I don't think it will do the job which is to figure out if quantic is installed as a dependency or not.

PR #2036, opened 2022-05-30
feat(headless): add insight "get interface" action

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2220

This PR adds the insight "get interface" action to Headless. In addition, the action is automatically dispatched when initializing the `InsightEngine`.

There is currently no `InsightInterface` controller but that would be useful in the future so a container component could simply register to updates made to this specific state section. In the meantime, it is still possible for a component to register to engine updates and find out when the interface is available, but having a specific controller would make it easier.


packages/headless/src/features/insight-interface/insight-interface-state.ts
```diff
@@ -0,0 +1,16 @@
...
+import {InsightAPIErrorStatusResponse} from '../../api/service/insight/insight-api-client';
+
+export const getInsightInterfaceInitialState = (): InsightInterfaceState => ({
+  loading: false,
+  config: undefined,
+});
+
+export interface InsightInterfaceState {
+  loading: boolean;
+  config?: {
```
olamothe, line 11: I think a better type would be to have an always defined/available `config`, but with a bunch of empty properties/object in `getInsightInterfaceInitialState().config`.

Its otherwise kind of annoying for a user of Headless to have to add defensive null ref check in a bunch of place in their code.

packages/headless/src/features/insight-interface/insight-interface-slice.ts
```diff
@@ -0,0 +1,25 @@
...
+        state.loading = true;
+      })
+      .addCase(fetchInterface.rejected, (state, action) => {
+        state.loading = false;
+        state.error = action.payload;
+      })
+      .addCase(fetchInterface.fulfilled, (state, action) => {
+        state.loading = false;
+        state.error = undefined;
+        state.config = {
```
olamothe, line 19: I think you could do `state.config = {...action.payload.response}` 🤔

packages/headless/src/features/insight-interface/insight-interface-slice.ts
```diff
@@ -0,0 +1,25 @@
+import {createReducer} from '@reduxjs/toolkit';
+import {fetchInterface} from './insight-interface-actions';
+import {getInsightInterfaceInitialState} from './insight-interface-state';
+
+export const insightInterfaceReducer = createReducer(
+  getInsightInterfaceInitialState(),
+  (builder) => {
+    builder
+      .addCase(fetchInterface.pending, (state) => {
+        state.loading = true;
```
olamothe, line 10: Should probably clear/reset the error on `pending`.

packages/headless/src/app/insight-engine/insight-engine.ts
```diff
@@ -80,6 +83,8 @@ export function buildInsightEngine(
     })
   );
 
+  engine.dispatch(fetchInterface());
```
olamothe, line 86: It might make sense to put this in a function (ie: `fetch()`), instead of when building the controller ?

That way you allow the user to determine when they want to retrieve and execute the network call.

For example, let's say for the preview feature, the implementer might want to fetch the config when the user change a configuration the interface, and not have to recreate the controller to fetch the new configuration.

olamothe (approved):

packages/headless/src/features/insight-interface/insight-interface-state.ts
lbergeron: In this case, knowing whether the interface config is loaded is very important because the container component has to wait for `config` to be filled before it can add the search components to the search interface. So, in my opinion, having a nullable `config` makes it obvious.

packages/headless/src/features/insight-interface/insight-interface-slice.ts
lbergeron: Affecting the state with `action.payload.response` would duplicate the search hub in the state. I can rewrite this to make the intent clearer though.

packages/headless/src/app/insight-engine/insight-engine.ts
lbergeron: Good point, I'll move that into a separate controller.

packages/headless/src/features/insight-interface/insight-interface-slice.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  const errorResponse = {
+    message: 'something bad happened',
+    statusCode: 400,
+    type: 'badluck',
+  };
+
+  let state: InsightInterfaceState;
+
+  beforeEach(() => {
+    state = getInsightInterfaceInitialState();
```
liviarett, line 19: if I understood it correctly, you're never reassigning state, right? I'd suggest getting rid of this and calling `getInsightInterfaceInitialState()` whenever you need the initial state.  Otherwise it makes it a bit confusing to understand that `state` is always the initial state

liviarett (commented):
could we get rid of the `any` [here](https://github.com/coveo/ui-kit/pull/2036/files#file-packages-headless-src-api-service-insight-insight-api-client-test-ts-L16)?
it seems to always be a `string` or an `Error`, so `response: string | Error` works alright if you add a `name` to your errors

packages/headless/src/features/insight-interface/insight-interface-slice.test.ts
lbergeron: Thanks for pointing that out! You are totally right, `state` is never reassigned. I'll rework this part to make it clearer.

lbergeron:
> could we get rid of the `any` [here](https://github.com/coveo/ui-kit/pull/2036/files#file-packages-headless-src-api-service-insight-insight-api-client-test-ts-L16)? it seems to always be a `string` or an `Error`, so `response: string | Error` works alright if you add a `name` to your errors

Of course we can do that. I'll set the type to `string | InsightAPIErrorStatusResponse` to match the tests.

ThibodeauJF (approved):

packages/headless/src/controllers/insight-interface/insight-interface.ts
```diff
@@ -0,0 +1,74 @@
...
+import {buildController, Controller} from '../controller/headless-controller';
+
+/**
+ * The `InsightInterface` controller is responsible for retrieving the Insight interface configuration.
+ */
+export interface InsightInterface extends Controller {
+  /**
+   * Fetches the Insight interface configuration.
+   */
+  fetch(): void;
```
erocheleau, line 26:
```suggestion
  fetchConfiguration(): void;
```
Maybe? Just because it's called an InsightInterface, and only having fetch() could be confusing to think you'd fetch an Interface?

I'm sure you already reflected on this, so put the name you think is good.

erocheleau (approved):
Everything lgtm! Just being picky on the names, you decide what you choose in the end ;)

packages/headless/src/controllers/insight-interface/insight-interface.ts
lbergeron: Well, the function is meant to fetch the interface and, with it, a small part of the Insight configuration. I'll keep `fetch` because it is closer to the "insight interface" concept.

packages/headless/src/controllers/insight-interface/insight-interface.ts
```diff
@@ -0,0 +1,74 @@
...
+    fetch() {
+      dispatch(fetchInterface());
+    },
+  };
+}
+
+function loadInsightInterfaceReducers(
+  engine: InsightEngine
+): engine is InsightEngine<
+  ConfigurationSection &
```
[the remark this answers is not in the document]
lbergeron: No, Prettier is formatting it this way on purpose. I guess `ConfigurationSection` is on its own level to show there is only one generic type. When there are many compound generic types, this way of indenting types makes more sense.

packages/headless/src/features/insight-interface/insight-interface-actions.ts
```diff
@@ -0,0 +1,52 @@
...
+    return {
+      response: fetched.success,
+    };
+  }
+);
+
+const buildGetInsightInterfaceRequest = (
+  state: StateNeededByFetchInterface
+): GetInsightInterfaceConfigRequest => ({
+  accessToken: state.configuration.accessToken,
```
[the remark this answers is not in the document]
lbergeron: Yes, we must keep `accessToken` because it comes from a parameter set that is shared by all headless API clients. Introducing different naming would cause code duplication, and add confusion.

packages/headless/src/features/insight-interface/insight-interface-actions.ts
```diff
@@ -0,0 +1,52 @@
...
+  organizationId: state.configuration.organizationId,
+  url: state.configuration.platformUrl,
+  insightId: state.insightConfiguration.insightId,
```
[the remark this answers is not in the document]
lbergeron: I am basically following the same pattern as Case Assist where the `caseAssistConfiguration` stores the ID in `caseAssistId`.

I do like your suggestion to use `state.insightConfiguration.id` a lot because of its simplicity. I'll try and see how it blends with the rest.

PR #2037, opened 2022-05-30
test(atomic): increase timeout for suggestion promises

liviarett:
https://coveord.atlassian.net/browse/KIT-1691


olamothe (approved):
Small details, but PR title should probably be `test(atomic)`

PR #2040, opened 2022-05-30
fix(atomic): add getAnimations() check

olamothe:
https://coveord.atlassian.net/browse/KIT-1678


ThibodeauJF (approved):

liviarett (approved):

packages/atomic/src/utils/ripple.ts
```diff
@@ -48,9 +48,11 @@ export function createRipple(event: MouseEvent, options: RippleOptions) {
 }
 
 async function cleanupAnimationOnFinish(ripple: HTMLSpanElement) {
-  const animationsFinished = ripple
-    .getAnimations()
-    .map((anim) => new Promise<void>((res) => (anim.onfinish = () => res())));
-  await Promise.all(animationsFinished);
-  ripple && ripple.remove();
+  if ('getAnimations' in ripple) {
```
btaillon, line 51: Maybe we should use [`animationend`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/animationend_event), which has more browser support, instead of `getAnimations`. You could use `listenOnce` from `event-utils`.

PR #2041, opened 2022-05-31
feat(headless):Add insight panel search actions

aelfaleh:
https://coveord.atlassian.net/browse/SVCC-2196

Add insight panel search actions
**Search Actions**
- executeSearch
- fetchMoreResults
- fetchFacetValues


packages/headless/src/features/insight-search/insight-search-state.ts
```diff
@@ -0,0 +1,52 @@
+import {InsightAPIErrorStatusResponse} from '../../api/service/insight/insight-api-client';
+import {
+  InsightQueryResponse,
+  SearchResult,
+} from '../../api/service/insight/query/query-response';
+
+export interface InsightSearchState {
+  isLoading: boolean;
+  error?: InsightAPIErrorStatusResponse | null;
```
lbergeron, line 9: I don't think we should have `| null` since `error` is already marked as optional.

packages/headless/src/test/mock-query-response.ts
```diff
@@ -0,0 +1,47 @@
...
+  searchUid: '',
+  facets: [],
+  results: [],
+  totalCount: 0,
+  executionReport: {},
+  ...response,
+});
+
+export const buildMockSearchResult = (
+  searchReseult: Partial<SearchResult> = {}
```
lbergeron, line 20:
```suggestion
  searchResult: Partial<SearchResult> = {}
```

packages/headless/src/features/insight-search/insight-search-slice.ts
```diff
@@ -0,0 +1,85 @@
...
+) {
+  state.isLoading = true;
+  state.requestId = action.meta.requestId;
+}
+
+export const insightSearchReducer = createReducer(
+  getInsightSearchInitialState(),
+  (builder) => {
+    builder
+      .addCase(insightExecuteSearch.rejected, (state, action) => {
```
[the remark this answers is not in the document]
aelfaleh: yes good catch

packages/headless/src/api/service/insight/query/query-response.ts
```diff
@@ -16,14 +16,14 @@ interface SearchResult {
   percentScore: number;
 }
 
-interface Facet {
+export interface Facet {
```
nathanlb, line 19: This was clearly here before this PR but i'm wondering why we're defining our own facet response interface rather than using the AnyFacetResponse that's already defined and covers several facet types we'll support in the insight panel as well. @lbergeron could you help me understand? We're going to have to end up redefining very similar interfaces so just want to make sure it's worth it.

packages/headless/src/state/state-sections.ts
```diff
@@ -374,3 +378,17 @@ export interface InsightInterfaceSection {
...
+export interface InsightCaseContextSection {
+  /**
+   * The insight case context
+   */
+  insightCaseContext: InsightCaseContextState;
+}
+
+export interface InsightSearchSection {
+  /**
+   * The insight search
```
nathanlb, line 391: I think we can add a bit more detail.
Copying from the SearchState description:
```suggestion
   * The information related to the insight search endpoint.
```

packages/headless/src/state/state-sections.ts
```diff
@@ -374,3 +378,17 @@ export interface InsightInterfaceSection {
    */
   insightInterface: InsightInterfaceState;
 }
+
+export interface InsightCaseContextSection {
+  /**
+   * The insight case context
```
nathanlb, line 384:
```suggestion
   * The insight case context to use with the insight search query.
```

packages/headless/src/api/service/insight/query/query-response.ts
lbergeron: You are bringing a very good point! I think you are right, and I should have used `AnyFacetResponse` right from the start. The good news is that we can still update `InsightQueryResponse` to use `AnyFacetResponse`.

packages/headless/src/test/mock-engine.ts
```diff
@@ -154,17 +158,25 @@ export interface MockInsightEngine
...
+  type: EngineType = 'search'
 ): MockCoreEngine<T> {
   const logger = pino({level: 'silent'});
-  const storeConfiguration = configureMockStore(logger);
+  const storeConfiguration =
+    type === 'search'
+      ? configureMockStore(logger)
+      : configureInsightMockStore(logger);
   const coreState = buildCoreState(config, mockState);
   const store = storeConfiguration(coreState);
```
lbergeron, line 181: Why not expose a hook allowing to build the store instance instead? It could be something like that.

```typescript
function buildMockCoreEngine<T extends AppState>(
  config: Partial<CoreEngine<T>> = {},
  mockState: () => T,
  buildStore: (logger, state) => MockStoreCreator<AppState, DispatchExts> = (logger, state) => configureMockStore(logger)(state)
): MockCoreEngine<T> {
  const logger = pino({level: 'silent'});
  const coreState = buildCoreState(config, mockState);
  const store = buildStore(logger, coreState);
  ...
}
```

The code above is just to share the idea, I did not try it though.

packages/headless/src/features/insight-search/insight-search-state.ts
```diff
@@ -0,0 +1,42 @@
+import {InsightAPIErrorStatusResponse} from '../../api/service/insight/insight-api-client';
+import {
+  InsightQueryResponse,
+  SearchResult,
+} from '../../api/service/insight/query/query-response';
+
+export interface InsightSearchState {
```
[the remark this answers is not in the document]
aelfaleh: what about the other sections?
`InsightConfigurationSection &
  InsightSearchSection &
  InsightInterfaceSection &
  InsightCaseContextSection;`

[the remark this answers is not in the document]
lbergeron: `InsightConfigurationSection`, `InsightInterfaceSection`, and `InsightCaseContextSection` are all storing information that is unique to the Insight use case, so they are not defined elsewhere. On the other hand components like the result list are expecting their data to be stored in `SearchSection`, not `InsightSearchSection`.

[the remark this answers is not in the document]
lbergeron: Thinking about it, having an `InsightSearchState` could still work when a component accesses the state through the controller. At this point, the component does not know whether the data comes from the `search` or the `insightSearch` section, it just reads the data when updated. However, since developers might not always be using controllers, having a search response not in the `search` section might be confusing.

@nathanlb What's your take on that?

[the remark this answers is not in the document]
nathanlb: Given we have insight search actions and reducers at this point I think it makes more sense to also define our own search state. That way we aren't bound by the current SearchState definition. 
As you said the controllers will return their own states so that's abstracted.
And my understanding is an insight panel engine would not have a `search` section in its state object if it has `insightSearch` so I don't see it being confusing.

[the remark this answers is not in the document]
nathanlb: Just updating here that we've decided to go with the existing SearchSection and its reducers.

packages/headless/src/features/insight-search/insight-case-context-state.ts
```diff
@@ -0,0 +1,8 @@
+export interface InsightCaseContextState {
+  caseContext: Record<string, string>;
+}
+
+export const getInsightCaseContextSearchInitialState =
```
nathanlb, line 5: Have you thought about how/where we should update this case context state? It looks like that redux mechanism isn't implemented here so we will need a separate story for that.
My hunch is it'll be preferable to have it in its own features/insight-case-context folder but I'd like to know your idea for this.

nathanlb (approved):
This is looking quite good. I'd like your take on the question I put but GG

packages/headless/src/features/insight-search/insight-search-slice.ts
```diff
@@ -0,0 +1,85 @@
...
+    {
+      requestId: string;
+    }
+  >
+) {
+  state.isLoading = true;
+  state.requestId = action.meta.requestId;
+}
+
+export const insightSearchReducer = createReducer(
```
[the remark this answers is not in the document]
aelfaleh: now we will use `searchReducer`

packages/headless/src/features/insight-search/insight-search-selectors.ts
```diff
@@ -0,0 +1,5 @@
+import {InsightSearchSection} from '../../state/state-sections';
+
+export function firstSearchExecutedSelector(state: InsightSearchSection) {
```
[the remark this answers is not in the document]
aelfaleh: fixed with the new refactor

packages/headless/src/features/insight-search/insight-case-context-state.ts
aelfaleh: I think about this too, I think it would be better another PR to create action that set `CaseContext`

nathanlb: Could you create a jira for that please?

packages/headless/src/features/insight-search/insight-search-slice.test.ts
```diff
@@ -0,0 +1,224 @@
+import {logSearchboxSubmit} from '../query/query-analytics-actions';
```
lbergeron, line 1: We should not need this test file anymore since we're reusing the `searchReducer`.

lbergeron (approved):
Looks good to me!

packages/headless/src/features/insight-search/insight-case-context-state.ts
aelfaleh: https://coveord.atlassian.net/browse/SVCC-2249

packages/headless/src/features/insight-search/insight-search-slice.test.ts
aelfaleh: good catch

packages/headless/src/features/insight-search/insight-search-actions-loader.ts
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Creates an action that executes a search query.
+   *
+   * @example
+   *
+   * ```js
+   * const {logInterfaceLoad} = loadSearchAnalyticsActions(engine);
+   * const {executeSearch} = loadInsightSearchActions(engine);
+   *
+   * engine.dispatch(executeSearch(interfaceLoad()));
```
btaillon, line 24: I think you forgot to end the code block here.
```suggestion
   * engine.dispatch(executeSearch(interfaceLoad()));
   * ```
```

packages/headless/src/api/service/insight/insight-api-client.ts
```diff
@@ -78,7 +78,7 @@ export class InsightAPIClient {
 
   async query(
     req: InsightQueryRequest
-  ): Promise<InsightAPIResponse<InsightQueryResponse>> {
+  ): Promise<InsightAPIResponse<SearchResponseSuccess>> {
```
btaillon, line 81: I see that the endpoint for the insight API and the search API are different, are we certain their response are always gonna be the same?

I just want to make sure that we won't accidentally break something in the insight panel if for some reason we have to adapt the `SearchResponseSuccess` interface.

packages/headless/src/features/insight-search/insight-search-actions-loader.ts
aelfaleh: good catch, thanks 🙏

aelfaleh:
@btaillon  @olamothe @ThibodeauJF , if you have time can you please take a look at this PR , I need another approval from you guys 🙏 
Thank you ❤️

ThibodeauJF (approved):

PR #2044, opened 2022-05-31
feat(atomic): truncate segmented facets values

mbenkirane-coveo:
- Truncates when label is over 15 characters

Screen Shot 2022-05-31 at 2 13 59 PM

Update:

  - Idle state
  Screen Shot 2022-06-06 at 7 45 56 PM
  
  - Hover state
  Screen Shot 2022-06-06 at 7 47 07 PM
  
  - Selected state
  Screen Shot 2022-06-06 at 7 46 26 PM


packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.pcss
```diff
@@ -2,7 +2,7 @@
 @import '../facet-common.pcss';
 
 .box-container {
-  @apply flex items-stretch;
+  @apply flex flex-row;
```
ThibodeauJF, line 5: I think row is the default direction, any difference when you don't specify it?

ThibodeauJF:
Screen Shot 2022-05-31 at 2 43 01 PM
From the updated design, the border 2px is only there when hovering a value, like the box facet

Screen Shot 2022-05-31 at 2 43 36 PM

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.pcss
```diff
@@ -13,21 +13,24 @@
   border-radius: 0px var(--atomic-border-radius) var(--atomic-border-radius) 0px;
 }
 
+.box-container .value-box {
+  @apply hover:border-primary-light rounded-none p-2 border-2 flex truncate;
+  max-width: 20ch;
+}
+
```
ThibodeauJF, line 20:
```suggestion
  @apply hover:border-primary-light rounded-none p-2 border-2 flex;
}

.box-container .value-label {
  max-width: 20ch;
}

```

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.pcss
```diff
@@ -13,21 +13,24 @@
...
 
-.box-container .value-box {
-  @apply hover:border-primary-light rounded-none p-2 border-2;
-}
-
-.value-box-count {
+.value-box .value-box-count {
   @apply pl-1;
+  overflow: visible;
+  width: auto;
```
ThibodeauJF, line 33: in the initial apply at line 31
`@apply pl-1 w-auto mt-0;`

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.pcss
ThibodeauJF: value label already has truncate & it should be the one that has a max num of strings

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.pcss
```diff
@@ -13,21 +13,24 @@
   border-radius: 0px var(--atomic-border-radius) var(--atomic-border-radius) 0px;
 }
 
+.box-container .value-box {
+  @apply hover:border-primary-light rounded-none p-2 border-2 flex truncate;
```
olamothe, line 17: I'm not sure why we are using an `@apply` directive here (and in a lot of places in this file) instead of classes directly on the element ? 🤔 

Is there a reason that the "classic" way does not work ?

btaillon: I've generally seen JF recommend it, so I've been using `@apply` where possible. I think both approaches are fine.

btaillon:
I'm a bit wary of how much CSS is in the `pcss` file here.

I think re-using `FacetValueLabelHighlight` for the segmented facets makes sense, but I'm not entirely sure that re-using `FacetValueBox` is providing value, it may even cause issues when we try to update the box facets.

I would eliminate `FacetValueBox` entirely and move as much of the styling as possible to the `tsx` file as classes. This also makes it easier to verify that every styled element has a part.

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.pcss
olamothe: But why is `@apply` better then ? 🤔

mbenkirane-coveo: Since I'm reusing some components (e.g the facet-value-box component) that are used by other kinds facets, I use the @apply directive to override some CSS properties for the segmented facet only. If I changed the class directly on the component, it would affect the other facets as well.

mbenkirane-coveo:
> 

Yeah, I was thinking about this as well. Don't you think it would be better to generalize the FacetValueBox component? Instead of hardcoding the classes, we could pass them as props.

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.pcss
ThibodeauJF: Apply is better than writing actual css (ex display: none) but the best is using the Tailwind classes directly in the JSX, of course

+1 for not using FacetValueBox, since it looks pretty different

mbenkirane-coveo: I'll create a new functional component !

btaillon:
> Yeah, I was thinking about this as well. Don't you think it would be better to generalize the FacetValueBox component? Instead of hardcoding the classes, we could pass them as props.

As discussed in the scrum after your comment, I also prefer the approach of making a new functional component vs re-using an existing one. I would be okay re-using it if the facets were almost identical and the boxes were almost identical, but I feel like they're all too different and I don't want to worry about breaking some distantly related component when making changes.

packages/atomic/src/components/facets/facet-segmented-value/facet-segmented-value.pcss
```diff
@@ -0,0 +1,24 @@
+.box-container {
```
ThibodeauJF, line 1: we should be able to delete most of what is in that file now that we're using a different functional component

packages/atomic/src/components/facets/facet-segmented-value/facet-segmented-value.pcss
```diff
@@ -0,0 +1,24 @@
...
+  @apply flex;
+}
+
+.box-container li:first-child .value-box {
+  border-radius: var(--atomic-border-radius) 0px 0px var(--atomic-border-radius);
```
btaillon, line 6: This should be do-able with tailwind shorthands. For instance, the `rounded` class applies `border-radius: var(--atomic-border-radius)`. The reason it applies our variable is because it corresponds with "DEFAULT" in `tailwind.config.js`.

packages/atomic/src/components/facets/facet-segmented-value/facet-segmented-value.pcss
```diff
@@ -0,0 +1,20 @@
...
+  @apply rounded-rr;
+}
+
+.value-box .value-label {
+  @apply group-hover:text-primary-light;
+  max-width: 15ch;
+}
+
+.box-container .value-box.selected {
+  @apply border-primary shadow-inner-primary;
```
mbenkirane-coveo, line 15: Wasn't able to apply these classes directly on the component as for some reason the border-primary class would not be applied and the shadow-inner-primary class would not apply the "inset" option inside of "inset 0 0 0 1px var(--atomic-primary)"

packages/atomic/src/components/facets/facet-segmented-value/facet-segmented-value.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const ariaLabel = props.i18n.t('facet-value', {
+    value: props.displayValue,
+    count: props.numberOfResults,
+  });
+
+  return (
+    <li key={props.displayValue}>
+      <Button
+        style="outline-bg-neutral"
```
ThibodeauJF, line 23: You should create a new style for the button that fits the mockups.

see `src/components/common/button-style.tsx` & `src/global/global.pcss`

That should fix the issue you have with the overrides

packages/atomic/tailwind.config.js
```diff
@@ -27,6 +27,13 @@ module.exports = {
         md: 'var(--atomic-border-radius-md)',
         lg: 'var(--atomic-border-radius-lg)',
         xl: 'var(--atomic-border-radius-xl)',
+        //left-rounded
+        'lr': 'var(--atomic-border-radius) 0px 0px var(--atomic-border-radius)',
+        //right-rounded
+        'rr': '0px var(--atomic-border-radius) var(--atomic-border-radius) 0px',
```
ThibodeauJF, line 33: you should be able to use this instead of creating new styles https://tailwindcss.com/docs/border-radius#rounding-sides-separately

eg. `rounded-r`

packages/atomic/src/components/facets/facet-segmented-value/facet-segmented-value.tsx
mbenkirane-coveo: Sounds good

packages/atomic/src/components/facets/facet-segmented-value/facet-segmented-value.pcss
```diff
@@ -0,0 +1,11 @@
+.box-container li:first-child .value-box {
+  @apply rounded-l;
+}
+
+.box-container li:last-child .value-box {
```
ThibodeauJF, line 5:
```suggestion
li:first-child .value-box {
```

packages/atomic/src/components/facets/facet-segmented-value/facet-segmented-value.pcss
```diff
@@ -0,0 +1,11 @@
...
+  @apply rounded-r;
+}
+
+.value-box .value-label {
```
ThibodeauJF, line 9:
```suggestion
.value-label {
```

ThibodeauJF (approved):
Great! Update the title with the correct formatting and you're good to go!

An upcoming pr could be removing this when not defined
Screen Shot 2022-06-07 at 9 24 37 AM

PR #2045, opened 2022-05-31
fix(headless): exposed notify and execute triggers as arrays

btaillon:
https://coveord.atlassian.net/browse/KIT-1696

Tested by replacing the script tag with the following:
<details>
<summary>index.html</summary>

```html
<script type="module">
  import {buildExecuteTrigger, buildNotifyTrigger} from './headless/dist/browser/headless.esm.js'

  function watchExecutions(engine) {
    const executeTrigger = buildExecuteTrigger(engine);
    executeTrigger.subscribe(() => {
      console.log('execute', JSON.parse(JSON.stringify(executeTrigger.state)));
    });
  }

  function watchNotifications(engine) {
    const notifyTrigger = buildNotifyTrigger(engine);
    notifyTrigger.subscribe(() => {
      console.log('notify', JSON.parse(JSON.stringify(notifyTrigger.state)));
    });
  }

  (async () => {
    const searchInterface = document.querySelector('atomic-search-interface');
      accessToken: 'xx564559b1-0045-48e1-953c-3addd1ee4457',
    });

    watchExecutions(searchInterface.engine);
    watchNotifications(searchInterface.engine);

    searchInterface.i18n.addResourceBundle('en', 'caption-filetype', {
      '.html': 'html',
    });
  })();
```
</details>

Then adding a symbolic link from atomic/www like so:
```sh
ln -s ../node_modules/@coveo/headless headless
```


packages/headless/src/features/triggers/triggers-slice.ts
```diff
@@ -1,39 +1,47 @@
...
-            functionName: executeTriggers[0].content.name,
-            params: executeTriggers[0].content.params,
-          }
-        : {functionName: '', params: []};
-      const notifyTriggers =
-        action.payload.response.triggers.filter(isNotifyTrigger);
-      state.notification = notifyTriggers.length
-        ? notifyTriggers[0].content
-        : '';
+      state.redirectTo = redirectTriggers[0] ?? '';
```
olamothe, line 37: What happens at the API level when you configure more than one redirect for a query ?

All of them are sent ? If so, should we do some kind of warning here to at least notify the user that something is probably not configured as it should ?

Might also be a good idea to report this to the search API as something that's kind of weird (that is, being able to have multiple redirection for one query).

packages/headless/src/features/triggers/triggers-slice.ts
```diff
@@ -1,39 +1,47 @@
...
+
+      state.query = queryTriggers[0] ?? '';
```
olamothe, line 39: Same as above.

olamothe (approved):
As discussed on slack yesterday, I think we should handle calling subscriber of a notification change when a new query returns an empty array.

packages/headless/src/features/triggers/triggers-state.ts
```diff
@@ -13,18 +27,34 @@ export interface TriggerState {
...
   /**
    * The JavaScript function to be executed after receiving an execute trigger.
+   *
+   * @deprecated
    */
-  execute: {functionName: string; params: ExecuteTriggerParams};
+  execute: FunctionExecutionTrigger;
+
+  /**
+   * The JavaScript functions to be executed after receiving execute triggers.
```
jpmarceau, line 36:
```suggestion
   * The JavaScript functions to execute after receiving execution triggers.
```

jpmarceau (approved):
👌

packages/headless/src/controllers/triggers/headless-execute-trigger.ts
```diff
@@ -22,13 +24,22 @@ export interface ExecuteTrigger extends Controller {
 export interface ExecuteTriggerState {
   /**
    * The name of the function to be executed.
+   *
+   * @deprecated
```
ThibodeauJF, line 28:
```suggestion
   * @deprecated Use .... instead.
```

would be more helpful to the implementer

packages/headless/src/features/triggers/triggers-slice.ts
ThibodeauJF: open in multiple tabs 😂

ThibodeauJF (approved):

packages/headless/src/index.ts
```diff
@@ -110,6 +110,7 @@ export {
   EcommerceDefaultFieldsToInclude,
 } from './features/fields/fields-state';
 export {buildSearchParameterSerializer} from './features/search-parameters/search-parameter-serializer';
+export type {FunctionExecutionTrigger} from './features/triggers/triggers-state';
```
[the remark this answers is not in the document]
btaillon: Not necessarily, it's a very small interface, but I generally prefer exporting interfaces for every object with more than two properties. This one has only two properties, but it's an array of that object.

packages/headless/src/features/triggers/triggers-slice.ts
btaillon: > What happens at the API level when you configure more than one redirect for a query ?
> 
> All of them are sent ?

The API returns all the redirects.

> If so, should we do some kind of warning here to at least notify the user that something is probably not configured as it should ?
> 
> Might also be a good idea to report this to the search API as something that's kind of weird (that is, being able to have multiple redirection for one query).

Perhaps. I'll ask them about it.

PR #2046, opened 2022-06-01
fix(headless): replace qa by stg

louis-bompart:
Replaced QA by STG.
There's no stg in EU, so I removed this combination from the test, tho I reckon the code doesn't check for bad combinations.


olamothe (approved):
🙇 Thanks !

btaillon (approved):

PR #2048, opened 2022-06-01
feat(atomic): Implement keyboard navigation between panels in search box suggestions

liviarett:
https://coveord.atlassian.net/browse/KIT-1689

Here is a video of the navigation:
https://user-images.githubusercontent.com/30511433/171474146-ae9f7c88-a972-4d83-bdd6-3c57445f0f19.mov

A few screenshots of the aria stuff for ease of review:
`combobox` input
Screenshot 2022-06-01 at 15 17 57
`listbox` div
Screenshot 2022-06-01 at 15 18 44

`option` list item when there is a query, default value (listitem?) when there isn't one, which includes dividers (`query === undefined`) and instant results (`query === ''`)
Screenshot 2022-06-01 at 15 20 02
Screenshot 2022-06-01 at 15 20 41

### DIFF
--- a/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
@@ -18,7 +18,6 @@ import {
 } from '../../utils/initialization-utils';
 import {Button} from '../common/button';
 import {randomID} from '../../utils/utils';
-import {isNullOrUndefined} from '@coveo/bueno';
 import {
   isDividerElement,
   isSuggestionElement,
@@ -48,6 +47,7 @@ import {promiseTimeout} from '../../utils/promise-utils';
  * @part suggestion - A suggested query correction.
  * @part active-suggestion - The currently active suggestion.
  * @part suggestion-divider - An item in the list that separates groups of suggestions.
+ * @part suggestion-with-query - An item in the list that will update the search box query.
  */
 @Component({
   tag: 'atomic-search-box',
@@ -59,7 +59,7 @@ export class AtomicSearchBox {
   private searchBox!: SearchBox | StandaloneSearchBox;
   private id!: string;
   private inputRef!: HTMLInputElement;
-  private listRef!: HTMLElement;
+  private listRefs: HTMLElement[] = [];
   private querySetActions!: QuerySetActionCreators;
   private pendingSuggestionEvents: SearchBoxSuggestionsEvent[] = [];
   private suggestions: SearchBoxSuggestions[] = [];
@@ -70,6 +70,7 @@ export class AtomicSearchBox {
   @State() public error!: Error;
   @State() private isExpanded = false;
   @State() private activeDescendant = '';
+  @State() private previousActiveDescendantElement: HTMLLIElement | null = null;
   @State() private leftSuggestions: SearchBoxSuggestions[] = [];
   @State() private leftSuggestionElements: SearchBoxSuggestionItem[] = [];
   @State() private rightSuggestions: SearchBoxSuggestions[] = [];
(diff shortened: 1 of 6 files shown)


ThibodeauJF:
Can I test this with an existing /examples page?

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -552,7 +596,11 @@ export class AtomicSearchBox {
             el?.replaceChildren(item.content);
           }
         }}
-        {...(isSuggestionElement(item) && {'data-query': item.query})}
+        {...(isSuggestionElement(item) && {
+          role: 'option',
```
btaillon, line 600: I would only add the roles `listbox` and `option` to the left container since it changes the query and corresponds the most with what we think of as suggestions.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -430,6 +458,14 @@ export class AtomicSearchBox {
         e.preventDefault();
         this.focusPreviousValue();
         break;
+      case 'ArrowRight':
+        e.preventDefault();
+        this.focusPanel(this.listRefs[1]);
```
btaillon, line 463: I think we should only call this when `listRefs[1]` exists, like so:
```ts
this.listRefs[1] && this.focusPanel(this.listRefs[1])
```

Otherwise, I think we should change the type of `focusPanel` to something like:
```ts
private focusPanel(panel: HTMLElement | undefined)
```

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -580,41 +629,47 @@ export class AtomicSearchBox {
...
         part="suggestions-wrapper"
         class={`flex w-full z-10 absolute left-0 top-full rounded-md bg-background border border-neutral ${
           this.showSuggestions ? '' : 'hidden'
         }`}
+        role="listbox"
       >
         {!!this.leftSuggestionElements.length && (
           <ul
-            id={this.popupId}
-            role="listbox"
```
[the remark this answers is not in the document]
btaillon: I think we should also give each `ul` element a different label to reflect their purpose.

The right suggestions could have a label that implies they are results. If so, we should probably also change the "left"/"right" side option when building suggestions to something like "suggestions"/"results". I'm very unsure.

btaillon (approved):

olamothe (approved):
One thing that will need to be fixed or prevented is the "flicker" that happens when you quickly switch from one suggestion to another, since each results rendered maybe have different width requirement. And so, things moves a lot.

One approach might be to make the width of each panel configurable ? 
Or to always make them 50/50 by default ?

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: if you plug in the existing components that we have, this is what will happen (actually, I think I did it wrong, it should be 
```
isSuggestionElement(item) && item.query && {
          role: 'option',
```
)
but I am thinking of times when people might attach their own suggestions to one panel or another. do you agree that we should support the `option="role"` in the right panel to account for those cases?

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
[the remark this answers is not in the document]
liviarett: What about the case when we have custom suggestions? We could potentially calculate the role/label depending on its elements, but I'm not sure how it should be done. Can we have two `listbox`s inside the `combobox`? can we unset the role for the `ul` if we keep the wrapping `div` as the `combobox`?

packages/atomic/src/components/search-box-suggestions/suggestions-common.ts
```diff
@@ -71,3 +72,11 @@ export function isDividerElement(
 ): el is SearchBoxDividerElement {
   return !('query' in el);
 }
+
+export function cleanUpString(str: string) {
```
ThibodeauJF, line 76: this could maybe go in `packages/atomic/src/utils/string-utils.ts`

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: that is happening explicitly here
```
 if (this.panelInFocus === panel) {
      return;
    }
```
but weird that it doesn't complain that `listRefs[x]` could be `undefined`! I'll update

liviarett:
@ThibodeauJF this one if you are still interested http://localhost:3333/examples/suggestions.html

PR #2050, opened 2022-06-01
feat(atomic,headless): add support for allowed values for specific facets

olamothe:
Opening PR pre-emptively, mostly for documentation revision.

The changes are not yet in prod at the platform level.

https://coveord.atlassian.net/browse/KIT-1699


jpmarceau (approved):

ThibodeauJF:
I would add it to the `atomic-color-facet` component too & maybe the `atomic-segmented-facet`

packages/atomic/src/components/facets/atomic-facet/atomic-facet.tsx
```diff
@@ -178,6 +178,28 @@ export class AtomicFacet implements InitializableComponent, BaseFacet<Facet> {
...
+   *
+   * ```html
+   * <atomic-facet field="objecttype" allowed-values="Contact,Account,File"></div>
+   * ```
+   *
+   * The maximum amount of allowed values is 25.
+   *
+   * Default value is `undefined`, and the facet uses all available values for its `field` in the current result set.
+   */
+  @Prop() public allowedValues?: string;
```
btaillon, line 201: I feel like the markup in other frameworks like React, Angular and Vue might be a bit weird if we take a comma-separated string instead of an array of string.

Maybe something we could do is something similar to `MapProp`. `Prop()` would generate an array of string in the doc, and `ArrayProp` would split & trim the attribute on commas.

I'm not certain this would work though, since MapProp is basically making other attributes to the one matching the prop, and this would be mapping the same attribute to itself.

btaillon (approved):

PR #2052, opened 2022-06-02
fix(headless): fix invalid number of results for recommendations

olamothe:
The "easy" fix would have been a one liner in the analytics provider class. The bug was due to the fact that the code used search results length, or 0, when determining the `numberOfResults` to send to usage analytics service.  It needed a fallback on recommendations results length.

I took the opportunity to refactor and split the analytics provider between "search" and "recommendation", and using a shared abstract class `BaseAnalyticsProvider`.

https://coveord.atlassian.net/browse/KIT-1701


packages/headless/src/features/product-recommendations/product-recommendations-analytics.actions.ts
```diff
@@ -1,15 +1,11 @@
...
-} from '../../api/analytics/product-recommendations-analytics';
 import {AnalyticsType, makeAnalyticsAction} from '../analytics/analytics-utils';
 
 export const logProductRecommendations = makeAnalyticsAction(
   'analytics/productrecommendations/load',
   AnalyticsType.Search,
-  (client) => client.logRecommendationInterfaceLoad(),
-  (state) =>
+  (client) => client.logRecommendationInterfaceLoad()
+  /*(state) =>
```
btaillon, line 7: Is this intentional?

btaillon (approved):

packages/headless/src/features/product-recommendations/product-recommendations-analytics.actions.ts
olamothe: Nope ! Good catch.

ThibodeauJF (approved):

PR #2053, opened 2022-06-06
feat(atomic): hide facets on "no results" & "error" status (search layout)

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1710

This will hide the facets whenever there are no results or any kind of error.
It would technically be possible at some point that some facet are selected. Currently all facets deselect themselves on a new query in Atomic so this won't be an issue, the selection is also represented by the breadcrumbs which will stay present & allow deselection.

behaviour: https://user-images.githubusercontent.com/4923043/172197431-713449f2-dbe1-4c86-8594-e637a42217bc.mp4


liviarett (approved):
I was looking at this earlier and wondering if there was a way to use `minmax` to hide this when empty and all that, but I dig too deep and it looks convoluted! would probs also have to remove the wrapping div when the facets are not there

btaillon (approved):

mbenkirane-coveo (approved):

olamothe:
Before this change, the magnifier icon would actually be aligned with the search box, but now since it's full width it looks mis-aligned...

What do you think ?

image

PR #2054, opened 2022-06-06
fix(headless): restoreSearchParameters action deletes the unselected values of the standard facet

mmitiche:
The goal of this PR is to figure out if the behaviour I will describe in this description is a bug or not and to answer the following question: **Why when we restore the search parameters we only set the selected facet values and we ignore the unselected values in the facet-set?**

### The problem:
After executing the first search,  building a url manager controller and selecting a value from a standard facet is causing the deletion of all the other values in this standard facet.

https://user-images.githubusercontent.com/86681870/172202709-d1b8849d-2aec-4957-987d-ca30518d61ce.mov

**Why this is happening?**

This is happening because of the following two steps:

1. Building a url manager controller:
When  building a url manager controller by calling `buildUrlManager` we also build a search parameter manager by calling `buildSearchParameterManager`.
When executing `buildSearchParameterManager` we dispatch a `restoreSearchParameters` action.
In [facet-set-slice.ts](https://github.com/coveo/ui-kit/blob/af7d9c59d3d11a682dbac22260c715d9b2667f21/packages/headless/src/features/facets/facet-set/facet-set-slice.ts#L59-L74)  we can see the code executed when this action is dispatched.
We set `state[facetId].currentValues `to the list of selected values. If there is no selected values `state[facetId].currentValues` will be an empty array.

2. Selecting a value from the standard facet:
When selecting a value from a standard value the `toggleSelectFacetValue` action is dispatched. you can find [here](https://github.com/coveo/ui-kit/blob/af7d9c59d3d11a682dbac22260c715d9b2667f21/packages/headless/src/features/facets/facet-set/facet-set-slice.ts#L75-L96) the code executed when this action is executed.
We first check if the facet value selected exists in the `state[facetId].currentValues`.
Since in step 1 we saw that `state[facetId].currentValues`  is an empty array now because of the `restoreSearchParameters action` the selected value will not be found so we enter to this [if block](https://github.com/coveo/ui-kit/blob/af7d9c59d3d11a682dbac22260c715d9b2667f21/packages/headless/src/features/facets/facet-set/facet-set-slice.ts#L88) and we just insert the selected value which will result the behaviour shown in the screen recording above.

### The potential fix if this a bug:
I just updated the piece of code executed when the `restoreSearchParameters` action is dispatched in facet-set-slice.ts. 
The update is to indeed affect the selected values to `state[facetId].currentValues` but also to keep the unselectedValues.
After all thats the goal of the restoreSearchParameters action, just to update the selected facet values, right?

https://user-images.githubusercontent.com/86681870/172207978-3091089d-125a-47c7-832c-71b7e7c146ec.mov

### DIFF
--- a/packages/headless/src/features/facets/facet-set/facet-set-slice.ts
+++ b/packages/headless/src/features/facets/facet-set/facet-set-slice.ts
@@ -62,12 +62,18 @@ export const facetSetReducer = createReducer(
 
         facetIds.forEach((id) => {
           const request = state[id];
-          const values = f[id] || [];
+          const selectedValues = f[id] || [];
+          const unselectedValues = request.currentValues.filter(
+            (facetValue) => !selectedValues.includes(facetValue.value)
+          );
 
-          request.currentValues = values.map(buildSelectedFacetValueRequest);
-          request.preventAutoSelect = values.length > 0;
+          request.currentValues = [
+            ...selectedValues.map(buildSelectedFacetValueRequest),
+            ...unselectedValues.map(buildUnselectedFacetValueRequest),
+          ];
+          request.preventAutoSelect = selectedValues.length > 0;
           request.numberOfValues = Math.max(
-            values.length,
+            selectedValues.length,
             request.numberOfValues
           );
         });
@@ -268,4 +274,10 @@ export function convertFacetValueToRequest(
 
 function buildSelectedFacetValueRequest(value: string): FacetValueRequest {
   return {value, state: 'selected'};
+}
+
+function buildUnselectedFacetValueRequest(
+  facetValue: FacetValueRequest
+): FacetValueRequest {
+  return {...facetValue, state: 'idle'};
 }
--- a/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
+++ b/packages/headless/src/features/facets/facet-set/facet-set-slice.test.ts
@@ -580,13 +580,21 @@ describe('facet-set slice', () => {
     it(`when a facet is found in the #f payload,
     it sets #currentValues to the selected values in the payload`, () => {
       const facetId = 'author';
-      state[facetId] = buildMockFacetRequest();
-
+      const valueA = buildMockFacetValueRequest({value: 'a'});
+      const valueB = buildMockFacetValueRequest({value: 'b'});
+      state[facetId] = buildMockFacetRequest({
+        currentValues: [valueA, valueB],
+      });
       const f = {[facetId]: ['a']};
       const finalState = facetSetReducer(state, restoreSearchParameters({f}));
-
-      const value = buildMockFacetValueRequest({value: 'a', state: 'selected'});
-      expect(finalState[facetId].currentValues).toEqual([value]);
+      const selectedValue = buildMockFacetValueRequest({
+        value: 'a',
+        state: 'selected',
+      });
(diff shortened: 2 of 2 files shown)


mmitiche:
> I think your fix would not cause problems so to speak (ie: no regressions).
> 
> However, the URL manager is designed to be used in a scenario where "the search application is loading and we need to apply an initial state before executing a search" kinda thing.
> 
> Hence the name `restoreSearchParameters` for the action. The scenario is also described here: https://docs.coveo.com/en/headless/latest/usage/synchronize-search-parameters-with-the-url/
> 
> So in that scenario where the app is loading, and no search has been executed yet, there is no unselected values anyway (it's always an empty array). We only know about the selected ones (because they come from the URL).
> 
> And when the first query finishes, only then do we know all the values available for a particular facet.
> 
> So I think your fix is fine to merge, but I'd like to know more about the original bug in SFINT-4552.
> 
> It seems a potential logic problem that you are creating an URL manager when the first search has already been executed anyway ?

Thank you for the clarification @olamothe!
I'll try to describe the real use case that led us to this situation where the URL manager is created when the first search has already been executed.


https://user-images.githubusercontent.com/86681870/172359191-aa48030e-ad3f-4887-a83f-7b0113e2e331.mov

Basically, inside Salesforce we can have different tabs in the same page .
Let's say we have two tabs, the first tab containing a Quantic search interface working with Headless and the second tab can contain anything.
When we first load the page we are in the first tab and everything is working correctly, we create the URL manager we execute the first search... everything as expected. 
The problem appears when we go to the second tab and then we come back to the first tab, when we go to the second tab the Quantic Search Interface is completely unmounted from the page and when we come back its mounted again and initialized again.
The Quantic Search Interface when initialized in this second time will do the following as usual:
- Execute first search.
- Create URL manager.
But this time executing the first search can't be made because a first search has already been executed (we enter to this [if statement block](https://github.com/coveo/ui-kit/blob/82f6bc7400d2d3f0bb0b9e3709fcae832911d768/packages/headless/src/app/search-engine/search-engine.ts#L122-L124))
And when the URL manager  is created it will cause the problem I'm describing.  That's the case where a URL manager is  created when the first search has already been executed.

I thought about trying to solve this from the Quantic side  by adding a condition to check if a first search was already executed so we can create the URL manager only if it's not the case, this would solve the problem of the disappearing facet values but it will give another problem which is that when going to the second tab and coming back to the first tab we will completely lose the features of the URL manager, for example updating the url hash automatically when a search parameter changes.

To narrow this down, this problem can happen anytime the Quantic Search Interface, for some reason, needs to be unmounted from the page and then mounted again.

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -62,12 +62,18 @@ export const facetSetReducer = createReducer(
 
         facetIds.forEach((id) => {
           const request = state[id];
-          const values = f[id] || [];
+          const selectedValues = f[id] || [];
+          const unselectedValues = request.currentValues.filter(
```
ThibodeauJF, line 66: Proposition, rename unselected to idle

ThibodeauJF (approved):

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
mmitiche: updated 👌

PR #2056, opened 2022-06-06
ci: lint only staged files

btaillon:
https://coveord.atlassian.net/browse/KIT-1709

This fix makes commits way faster.

`commit-lint.js` was unused.

`commit-msg.js` is used by Husky.


package.json
```diff
@@ -72,7 +72,7 @@
   },
   "lint-staged": {
     "**/*.ts?(x)": [
-      "npm run lint:fix"
```
btaillon, line 75: the `lint-staged` section defines which scripts to run for each staged files.

In this case, if you had `file-a.tsx`, `file-b.tsx` and `file-c.html` staged, `lint-staged` would run:
`npm run lint:fix file-a.tsx file-b.tsx`

However, the `lint:fix` script executes the following `eslint --fix .`, meaning that everything in the repo was getting fixed every time we ran `git commit`.

In the case of everything getting run with prettier, it already worked properly.

liviarett (approved):
can't wait to pull this!! 👯

olamothe (approved):

PR #2057, opened 2022-06-06
test(atomic): add applitools eyes visual testing framework for atomic result list

wmannard:
https://coveord.atlassian.net/browse/QA-976

This adds:
- A new Cypress test that sends the DOM to be tested to Applitools Eyes (the search call response is mocked)
- A test page (like index.html but the result images are static)
- New commands/config to run this test
- Configurations for Applitools Eyes
- A new Github action

To be done: Add Applitools Eyes integration to github: https://applitools.com/docs/topics/integrations/github-integration.html

I left the existing screenshots tests in. If we want to switch over to this new solution, we'll likely need a paid account.

See README for info about logging in to Applitools console.

Screen Shot 2022-06-06 at 4 18 36 PM


.github/workflows/prbot.yml
```diff
@@ -57,6 +57,16 @@ jobs:
...
+  e2e-atomic-visual-tests:
+    name: 'Run Applitool visual regressions tests on Atomic'
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+      - uses: ./.github/actions/e2e-visual-tests
+    env:
+      APPLITOOLS_API_KEY: ${{ secrets.APPLITOOLS_API_KEY }}
```
wmannard, line 69: This will need to be added into github

wmannard:
I think I messed something up with the lockfiles. I'll put them back the way they were and then re-add the new dependencies.

mbenkirane-coveo:
> I think I messed something up with the lockfiles. I'll put them back the way they were and then re-add the new dependencies.

I don't think you're supposed to push the lock files since they get generated automatically. You can just disregard the lock file changes.

.github/workflows/prbot.yml
olamothe: You can share it with me on lastpass I will add it.

packages/atomic/tsconfig.json
```diff
@@ -24,13 +24,16 @@
...
   },
   "include": ["src", "types/jsx.d.ts"],
   "exclude": [
     "node_modules",
     "src/external-builds",
     "**/*.stories.tsx",
     "headless"
-  ]
+  ],
+  "files": ["./node_modules/@applitools/eyes-cypress/index.d.ts"],
```
olamothe, line 38: Is this really needed ?

packages/atomic/tsconfig.json
```diff
@@ -24,13 +24,16 @@
         }
       }
     ],
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
```
olamothe, line 29: This has an impact on the generated package output.

Would it be possible to have a separate tsconfig just for cypress tests ?

packages/atomic/tsconfig.json
wmannard: Without this line, the linter flags the custom cypress commands like `cy.eyesCheckWindow`

packages/atomic/tsconfig.json
wmannard: Sure. The change was so that I could import the body of the search-api response which I'd saved in a .json file. I'll look into using a separate config. Alternatively, I could use JS instead of TS for the cypress test file.

.github/workflows/prbot.yml
```diff
@@ -57,6 +57,16 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/setup
       - uses: ./.github/actions/e2e-atomic-screenshots
```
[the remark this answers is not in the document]
wmannard: I was thinking we'd leave both for now, then once we decide if we want to adopt this workflow or not, we can delete one of the two.

packages/atomic/tsconfig.json
wmannard: I hadn't noticed that there's a tsconfig file in the cypress folder. Moving this and the other changes there. Would that be OK?

packages/atomic/cypress/tsconfig.json
```diff
@@ -5,7 +5,10 @@
...
     "target": "es6",
     "lib": ["esnext", "dom"],
-    "types": ["cypress", "cypress-axe"]
+    "types": ["cypress", "cypress-axe"],
+    "resolveJsonModule": true,
+    "esModuleInterop": true
   },
-  "include": ["**/*.ts", "**/*.js"]
+  "include": ["**/*.ts", "**/*.js"],
+  "files": ["../node_modules/@applitools/eyes-cypress/index.d.ts"],
```
wmannard, line 13: @olamothe I moved all the tsconfig changes here

btaillon (approved):
looks good!

packages/atomic/README.md
```diff
@@ -63,6 +63,33 @@ npm run cypress-hsp:open
 npm run cypress-hsp:test
 npm run cypress-hsp:test:firefox
 ```
+### Visual regression tests
```
ThibodeauJF, line 66: niceeee

.github/workflows/prbot.yml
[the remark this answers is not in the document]
ThibodeauJF: Good point! I'd like if we compare both for a bit

ThibodeauJF (approved):

lvu285 (approved):

PR #2058, opened 2022-06-06
feat(atomic,headless): added custom default value for search parameters' numberOfResults

btaillon:
https://coveord.atlassian.net/browse/KIT-1703

Prior to this PR, if somebody added a `atomic-results-per-page` component in their page and configured an initial value that isn't 10, the initial value would be added to `numberOfResults` in the URL.

Prior to this PR, a lack of `numberOfResults` in the URL was always interpreted as a value of 10.


packages/headless/src/controllers/search-parameter-manager/headless-search-parameter-manager.ts
```diff
@@ -209,9 +209,8 @@ function getNumberOfResults(state: Partial<SearchParametersState>) {
     return {};
   }
 
-  const numberOfResults = state.pagination.numberOfResults;
-  const shouldInclude =
-    numberOfResults !== getPaginationInitialState().numberOfResults;
+  const {numberOfResults, defaultNumberOfResults} = state.pagination;
```
btaillon, line 212: I decided to name this `defaultNumberOfResults` instead of `initialNumberOfResults` to mimic `advancedSearchQueries`'s `defaultFilters`.

olamothe (approved):
I'm not seeing in the code where we pass the `initial-choice` from `atomic-results-per-page` down to the headless results per page controller `defaultNumberOfResults`.

olamothe:
The fact that the number of results would appear in the URL is not 100% optimal, true, but also not the end of the world ?

Are we certain that this

btaillon:
> I'm not seeing in the code where we pass the `initial-choice` from `atomic-results-per-page` down to the headless results per page controller `defaultNumberOfResults`.

It's passed in `buildResultsPerPage` through the `initialState`.

packages/atomic/cypress/integration/results-per-page.cypress.ts
```diff
@@ -71,9 +71,29 @@ describe('Result Per Page Component', () => {
...
+        cy.document().its('location').its('hash').should('be.empty');
+      });
+
+      describe('then selecting another value, then pressing the back button', () => {
+        before(() => {
+          ResultsPerPageSelectors.buttons().last().click();
+          cy.go('back');
+        });
+
+        it('should not add the initial choice to the URL', () => {
```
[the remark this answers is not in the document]
btaillon: Yes, adding it.

ThibodeauJF (approved):

packages/headless/src/features/search-parameters/search-parameter-selectors.ts
```diff
@@ -19,7 +19,9 @@ export function initialSearchParameterSelector(
       state.advancedSearchQueries?.defaultFilters.cq ??
       getAdvancedSearchQueriesInitialState().defaultFilters.cq,
     firstResult: getPaginationInitialState().firstResult,
-    numberOfResults: getPaginationInitialState().numberOfResults,
+    numberOfResults:
+      state.pagination?.defaultNumberOfResults ??
+      getPaginationInitialState().numberOfResults,
```
liviarett, line 24: how about `getPaginationInitialState().defaultNumberOfResults` to make it a bit clearer?

liviarett: * clearer that you are actually just talking about default values here, and nothing that could change

liviarett (approved):

PR #2059, opened 2022-06-07
fix(atomic): change keys of locales

louis-bompart:
Manual said that. https://www.i18next.com/how-to/faq#how-should-the-language-codes-be-formatted


ThibodeauJF:
No it was buggy!

ThibodeauJF (approved):
thanks!

PR #2060, opened 2022-06-07
fix(atomic): fix issue with changing panels while results load

liviarett:
I finally managed to reproduce a issue I saw every once in a while - when we change the panel in focus before the elements are updated, it breaks like so:
https://user-images.githubusercontent.com/30511433/172345709-eff01237-1ded-4162-ae92-f0c0b0029cb4.mov

Now, keeping track of the element that triggered the change, we can make sure that we don't update the wrong panel:
https://user-images.githubusercontent.com/30511433/172345889-0f89aacc-f8f1-4000-9325-cd83c97d2b9b.mov

I also took the opportunity to fix a little difference in behaviour between mouse and keyboard focus (before, it lost track of the previous element on mouse over)
https://user-images.githubusercontent.com/30511433/172346033-3decc9a0-63a7-4ed1-ad07-c4d4cc856569.mov


packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -414,32 +419,37 @@ export class AtomicSearchBox {
     );
   }
 
-  private isPanelInFocus(panel: 'right' | 'left') {
+  private isPanelInFocus(
+    panel: HTMLElement | undefined,
+    query?: string
```
olamothe, line 424: Why is this param optional ?

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -207,6 +207,13 @@ export class AtomicSearchBox {
     this.activeDescendant = activeDescendant;
   }
 
+  private updateActiveDescendantWithHistory(activeDescendant = '') {
```
olamothe, line 210: What does `withHistory` mean here ?

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -590,7 +600,13 @@ export class AtomicSearchBox {
           this.clearSuggestions();
         }}
         onMouseOver={() => {
-          this.updateActiveDescendant(id);
+          const thisPanel =
```
olamothe, line 603: suggestion: Just for readability, this whole method (`renderSuggestions`) is starting to become very compact/dense, with a lot happening on every line.

Might help to at least extract well named private methods, to reduce the complexity when reading/scanning the file quickly.

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -414,32 +419,37 @@ export class AtomicSearchBox {
...
     if (!this.activeDescendantElement) {
       return false;
     }
-    return this.activeDescendantElement
-      ?.closest('ul')
-      ?.getAttribute('part')
-      ?.includes(`suggestions-${panel}`);
+
+    if (query) {
+      return !!panel?.querySelector(`li[data-query="${query}"]`);
```
btaillon, line 431: you should escape double-quotes in `query`.

olamothe (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: we should do that here too, then?
```jsx
        {...(hasQuery && {
          role: 'option',
          'data-query': item.query,
          'aria-selected': `${isSelected}`,
        })}
```

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
btaillon: What hapepns if we quickly move between 3 different results? Would this still work correctly?

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
btaillon: no, specifically in `querySelectors` since the selector uses a quotation mark.

liviarett: ah yeah, duh

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: `withHistory` means that we are keeping track of the previous element (when we change panels). I can rename to `updateActiveAndPreviousDescendants`, is that better? @olamothe

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: no need I guess!

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: this is the best I could do for you at this time 😂 https://github.com/coveo/ui-kit/pull/2060/commits/536c3ef4b81bb1f81203ce395205cadb3bcf6869 if you have any other suggestions, let me know!

btaillon (approved):

ThibodeauJF (approved):

PR #2065, opened 2022-06-08
chore(headless): extract core controllers for result-list and search-status

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2203

This PR is extracting core controllers out of `result-list` and `search-status` in order to be able to create an `InsightResultList` controller (will be done in a separate PR).

I basically:
* copied `headless-search-status` to `headless-core-search-status` and prefixed the interfaces with `Core`.
* copied `headless-result-list` to `headless-core-result-list` and prefixed the interfaces with `Core`.
* added a hook on `CoreResultListProps` to be able to inject a specific action creator for `fetchMoreResults`. The goal is to keep the API call debouncing logic hidden in the core controller.
* adapted `headless-search-status` and `headless-result-list` to leverage the core controllers.
* copied and adapted the `search-status` and `result-list` tests for the core controllers.

The tests are intentionally duplicated between the original and the core controllers to ensure that the refactoring didn't break the original controllers.

This is my first attempt at extracting core controllers, so please let me know whether I'm doing it right.


ThibodeauJF:
Maybe it should be `headless-core-status` because of the "search" keyword. We could have a `headless-recommendation-status` and `headless-insight-status`

lbergeron:
@ThibodeauJF 

> Maybe it should be `headless-core-status` because of the "search" keyword. We could have a `headless-recommendation-status` and `headless-insight-status`

I guess renaming the controller to `headless-core-status` would make sense.

> Maybe it could even be unique to the search engine. We don't really need to have a common controller for this one.

No, I need a core controller for `status` because it is being referenced by the `result-list` (now core) controller. In addition, the state shape of the `result-list` controller has to be consistent with the `search` use case otherwise the existing UI components won't work.

packages/headless/src/controllers/result-list/headless-result-list.ts
```diff
@@ -101,80 +61,19 @@ export function buildResultList(
...
-    dispatch(fetchMoreResults()).then(() => (lastFetchCompleted = Date.now()));
-  };
+  const coreController = buildCoreResultList(engine, {
+    ...props,
+    fetchMoreResultsActionCreator: fetchMoreResults,
+  });
 
   return {
-    ...controller,
+    ...coreController,
```
olamothe, line 70: Unless I'm mistaken, you could just do : 

```typescript
buildResultList(engine, props) {
 return buildCoreResultList(engine, {
    ...props,
    fetchMoreResultsActionCreator: fetchMoreResults,
  });
}
```

packages/headless/src/controllers/core/result-list/headless-core-result-list.test.ts
```diff
@@ -0,0 +1,155 @@
+import {buildMockSearchAppEngine} from '../../../test/mock-engine';
+import {buildMockResult, MockSearchEngine} from '../../../test';
+import {fetchMoreResults} from '../../../features/search/search-actions';
+import {buildCoreResultList, CoreResultList} from './headless-core-result-list';
+import {registerFieldsToInclude} from '../../../features/fields/fields-actions';
+import {configuration, fields, search} from '../../../app/reducers';
+import {SchemaValidationError} from '@coveo/bueno';
+
+describe('CoreResultList', () => {
```
olamothe, line 9: It's great that you added this, but it's a bit superfluous 😅 

Or we could remove the tests for `headless-result-list.test.ts` instead, they are just overlapping.

If `headless-result-list` starts getting specific functionalities that are not 100% the same as the core result list, then we'll be able to test for those specifically by adding `headless-result-list.test.ts`

olamothe (approved):

packages/headless/src/controllers/core/result-list/headless-core-result-list.test.ts
lbergeron: Thanks for your input! I wasn't sure of the preferred approach for this project. I'll remove the duplicated tests from `headless-result-list.test.ts`.

ThibodeauJF (approved):

aelfaleh (approved):

PR #2066, opened 2022-06-08
feat(atomic): allow to click on item inside suggestion popover

liviarett:
https://coveord.atlassian.net/browse/KIT-1688

I'll change the base branch to master once that one gets merged!

https://user-images.githubusercontent.com/30511433/172644199-77232df5-76cc-4505-b569-419be07fd705.mov

Like I said on standup, I think using templates is enough for the analytics, just needed to make sure that we could actually click on the stuff before they disappeared from the screen (on blur)


btaillon (approved):

mbenkirane-coveo (approved):

packages/atomic/src/components/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -66,8 +66,8 @@ export class AtomicSearchBoxInstantResults {
     const results = this.instantResults.state.results.length
       ? this.instantResults.state.results
       : this.results;
-    return results.map((result: Result) => ({
-      key: `instant-result-${cleanUpString(result.title)}`,
+    return results.map((result: Result, i: number) => ({
+      key: `instant-result-${cleanUpString(result.title)}-${i}`,
```
ThibodeauJF, line 70: This isn't recommended with JSX. Instead of using the result.title, we should use the result `uniqueId` and perhaps remove the need for `cleanUpString`

> Do not use the map-function's index variable as a key. It does not represent a stable identity of an item as it can change if the order of the list changed or if you added an item to the beginning of the list. As such it is not suitable as a key.

https://stenciljs.com/docs/templating-jsx#using-jsx

liviarett: yeah, the problem with the unique ID is that it's got weird characters and it was breaking everything... but I guess I could clean that one up instead! alright, will update!

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -679,43 +708,17 @@ export class AtomicSearchBox {
...
-            {this.leftSuggestionElements.map((suggestion, index) =>
-              this.renderSuggestion(
-                suggestion,
-                index,
-                this.leftSuggestionElements.length - 1,
-                'left'
-              )
-            )}
-          </ul>
+        {this.renderPanel(
```
liviarett, line 711: I added a little refactoring here, as all the code is duplicated for these two panels! let me know what you guys think

liviarett:
new video:

https://user-images.githubusercontent.com/30511433/172899742-3ff00c5c-b514-4a7e-a38e-65564ba7cba8.mov

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -665,6 +665,35 @@ export class AtomicSearchBox {
...
+    }
+
+    return (
+      <ul
+        part={`suggestions suggestions-${side}`}
+        aria-label={this.bindings.i18n.t('query-suggestion-list')}
+        ref={setRef}
+        class="flex-grow"
+        onMouseDown={(e) => {
+          if (e.target === getRef()) {
```
liviarett, line 685: this `onMouseDown` function is the main hero of this PR, making sure that clicking on the panel doesn't blur the input!

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
```diff
@@ -577,7 +577,7 @@ export class AtomicSearchBox {
 
   private onSuggestionClick(item: SearchBoxSuggestionItem) {
     item.onSelect && item.onSelect();
-    this.clearSuggestions();
+    isSuggestionElement(item) && item.query && this.clearSuggestions();
```
liviarett, line 580: this makes sure that we don't blur the input when we click on something that doesn't update the input. I think this makes more sense.. since `clearSuggestions` is accessible from the implementer, elements can deal with it case by case if needed

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
olamothe: Ok ! Hummm I'm wondering if that's going to be "easy" to break down the line.

When we miss clicks analytics recording, it's always causing big problems 😅

What was the problem before this change ? Was it that you could not click on the result ? Like, you would not get redirected to the product itself ? Or just the analytics function that would not properly execute ?

Also, is there the same problem if we use keyboard navigation (ie: tab to the result link on the preview, and press enter) ?

liviarett: https://user-images.githubusercontent.com/30511433/173049524-a6459bc4-c445-4f7c-9977-059c3755f66b.mov

this is a video that shows what happens before this PR. it does log the clicks, but doesn't let the page redirect... 
in general, I think that if an input creates a little popup like that, anywhere you click inside it should leave it open (i.e. clicking on the white space after the queries) unless specifically specified otherwise (like a close button, or, say, updating the query bc we want the results underneath to be immediately visible). 

What this function does, as you might have noticed, is to handle this case in particular, where the click happens on the `ul` where it doesn't have a child. It avoids the input blurring when you click that (and that's why I used mouse down and not mouse over - the blur happens between the mouse down and click events).

Notice the focus outline on this video:
https://user-images.githubusercontent.com/30511433/173050405-941b7be2-619b-4f3a-af14-74f89d08bbc1.mov
for some reason, clicking the `li` elements doesn't remove the focus from the input, but clicking the `ul` does. So to avoid the popup disappearing in both cases, this PR essentially does two things:
- avoids blurring input when clicking the `ul` element
- doesn't clearSuggestions [when a suggestion that has a query is clicked](https://github.com/coveo/ui-kit/pull/2066/files#diff-46e3c6b20a52a198d73e328db1b88c9a1b4b44ba84c5e7556011e630fc8f4638R580) (nothing is prevented in that case)

So, summarising, it don't think it will affect any analytics! 
Thanks for coming to my ted talk! Let me know if you still have questions/concerns!

btaillon (approved):

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
liviarett: it looks like when you press enter on the result, it doesn't navigate to the page, but only handles the input stuff with the query that's already there. what do you think @btaillon ? any ideas?

https://user-images.githubusercontent.com/30511433/173125978-c74f442e-e854-413e-8769-f798593d7a8f.mov

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
ThibodeauJF: clean!

ThibodeauJF:
Careful with these package locks when you merge, I don't think you need any of them in this PR
Screen Shot 2022-06-10 at 3 24 07 PM


e.g. `git checkout origin/master -- packages/headless/package-lock.json`

packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
ThibodeauJF: ideally it would reuse the logic from the `LinkWithResultAnalytics` funcitonal component

```
onClick={(e) => stopPropagationAndProcess(e, onSelect)}
onContextMenu={(e) => stopPropagationAndProcess(e, onSelect)}
onMouseDown={(e) => stopPropagationAndProcess(e, onSelect)}
onMouseUp={(e) => stopPropagationAndProcess(e, onSelect)}
onTouchStart={(e) => stopPropagationAndProcess(e, onBeginDelayedSelect)}
onTouchEnd={(e) => stopPropagationAndProcess(e, onCancelPendingSelect)}
```

ThibodeauJF (approved):

PR #2068, opened 2022-06-08
feat(atomic): single select and no more breadcrumbs for segmented facets

mbenkirane-coveo:
https://coveord.atlassian.net/browse/KIT-1708

### DIFF
--- a/packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
+++ b/packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
@@ -99,7 +99,7 @@ export class AtomicSegmentedFacet
     };
     this.facet = buildFacet(this.bindings.engine, {options});
     this.facetId = this.facet.state.facetId;
-    registerFacetToStore(this.bindings.store, 'facets', {
+    registerFacetToStore(this.bindings.store, 'segmentedFacets', {
       label: this.label,
       facetId: this.facetId!,
       element: this.host,
@@ -140,7 +140,7 @@ export class AtomicSegmentedFacet
   private renderValues() {
     return this.renderValuesContainer(
       this.facetState.values.map((value) =>
-        this.renderValue(value, () => this.facet.toggleSelect(value))
+        this.renderValue(value, () => this.facet.toggleSingleSelect(value))
       )
     );
   }
--- a/packages/atomic/src/utils/store.ts
+++ b/packages/atomic/src/utils/store.ts
@@ -15,7 +15,12 @@ interface FacetValueFormat<ValueType> {
   content?(facetValue: ValueType): VNode;
 }
 
-type FacetType = 'facets' | 'numericFacets' | 'dateFacets' | 'categoryFacets';
+type FacetType =
+  | 'facets'
+  | 'numericFacets'
+  | 'dateFacets'
+  | 'categoryFacets'
+  | 'segmentedFacets';
 type FacetStore<F extends FacetInfo> = Record<string, F>;
 
 export interface SortDropdownOption {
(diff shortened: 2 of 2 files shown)


packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -99,7 +99,7 @@ export class AtomicSegmentedFacet
     };
     this.facet = buildFacet(this.bindings.engine, {options});
     this.facetId = this.facet.state.facetId;
-    registerFacetToStore(this.bindings.store, 'facets', {
+    registerFacetToStore(this.bindings.store, 'segmentedFacets', {
```
btaillon, line 102: Is there any purpose to registering segmented facets in the store?

ThibodeauJF:
I think we might do more to prevent those facets from appearing in the breadbox. Looking at the console, you'll see an error that stems from the breadbox not finding the related facet element.

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
olamothe: Yep it does not seem to serve any purpose as of now. Means all the code changes in `store.ts` for this PR could be removed at the same time.

mbenkirane-coveo: Actually, I did decide to register the facets in the store and added a field in the FacetInfo interface to be able to filter out facets that have breadcrumbs vs that don't

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -16,7 +16,7 @@ import {
   FacetValue,
   SearchStatus,
   SearchStatusState,
-} from '@coveo/headless';
+} from '../../../../headless';
```
btaillon, line 19: Looks like this caused a change in `components.d.ts`. We typically import from `@coveo/headless`, so I think you should revert this line.

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -103,6 +103,7 @@ export class AtomicSegmentedFacet
       label: this.label,
       facetId: this.facetId!,
       element: this.host,
+      hasBreadcrumbs: false,
```
btaillon, line 106: I would still prefer not to register the segmented facets in the store. If we need to register them in the future, I would prefer to register them separately from normal facets, since part of the reason that normal facets are registered is so that they their element can be cloned inside the refine modal. I feel like the behaviour of facets vs segmented facets is too different to register them in the same place.

mbenkirane-coveo: So, do you think it would be better to implement a new buildSegmentedFacet function instead of using the buildFacet function that's currently available

packages/atomic/src/components.d.ts
```diff
@@ -15,6 +15,7 @@ import { AtomicStore } from "./utils/store";
 import { i18n } from "i18next";
 import { InitializationOptions } from "./components/atomic-search-interface/atomic-search-interface";
 import { StandaloneSearchBoxData } from "./utils/local-storage-utils";
+import { FacetSortCriterion as FacetSortCriterion1 } from "../headless";
```
ThibodeauJF, line 18: what ... 🤔 🤔 🤔 🤔

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
ThibodeauJF: Yes!

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
ThibodeauJF: Okay! Let's just go ahead and filter out back in the breadbox & remove the new `hasBreadcrumbs` param, the rest should be good

```
({facetId}) => this.bindings.store.state.facets[facetId]
```

ThibodeauJF (approved):
Looks good!

ThibodeauJF (approved):

PR #2069, opened 2022-06-08
ci: remove cypress reinstalls, commit package-lock files right when bumping

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1719

### Example issue

Created KIT-1718 branch since the latest release

Headless versions are the same in the files (1.70.1)
https://github.com/coveo/ui-kit/blob/KIT-1718/packages/headless/package.json#L18
https://github.com/coveo/ui-kit/blob/KIT-1718/packages/headless/package-lock.json#L3 

Yet we get this error in the e2e test https://github.com/coveo/ui-kit/runs/6798309712?check_suite_focus=true

	/opt/hostedtoolcache/node/16.15.0/x64/bin/npm ci
	npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or 	
	npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing. 	npm ERR! Invalid: lock file's @coveo/headless@1.70.0 does not satisfy @coveo/headless@1.70.1

We somehow have a lockfile with the wrong version with some e2e cypress tests.


.github/actions/setup/action.yml
```diff
@@ -39,11 +39,15 @@ runs:
           !**/node_modules
         key: ${{ github.sha }}
     - uses: actions/cache@v2
+      id: npm-cache
       with:
-        path: |
-          node_modules
-          **/node_modules
-        key: ${{ hashFiles('**/package-lock.json') }}
+        path: ~/.npm
```
ThibodeauJF, line 44: > Do not cache node_modules across builds. This bypasses more intelligent caching packaged with npm or yarn, and can cause issues with Cypress not downloading the Cypress binary on npm install.

https://docs.cypress.io/guides/continuous-integration/introduction#Caching

.github/actions/setup/action.yml
```diff
@@ -39,11 +39,15 @@ runs:
...
+        key: npm-${{ hashFiles('**/package-lock.json') }}
+    - uses: actions/cache@v2
+      id: cypress-cache
+      with:
+        path: ~/.cache/Cypress
```
ThibodeauJF, line 49: > As of [Cypress version 3.0](https://docs.cypress.io/guides/references/changelog#3-0-0), Cypress downloads its binary to the global system cache - on linux that is ~/.cache/Cypress. By ensuring this cache persists across builds you can save minutes off install time by preventing a large binary download.

https://docs.cypress.io/guides/continuous-integration/introduction#Caching

.github/actions/setup/action.yml
btaillon: Looks good, this seems to be the same directory that [`actions/setup-node` uses](https://github.com/actions/setup-node/blob/main/src/cache-utils.ts) when `cache: true` is set on it.

I don't remember the exact reason why I didn't use `setup-node` to cache dependencies.

btaillon (approved):

ThibodeauJF:
@btaillon current solution isn't fully working, I need a bit of help with the angular package, not sure how it works differently than others here...

.github/actions/setup/action.yml
ThibodeauJF: Does that mean we should use this action?

scripts/deploy/bump-version.js
```diff
@@ -18,6 +18,7 @@ async function checkoutLatestMaster() {
 
 async function bumpVersionAndPush() {
   try {
+    // But wait, there's more! https://github.com/lerna/lerna/blob/main/commands/version/README.md#lifecycle-scripts
```
olamothe, line 21: image

olamothe (commented):
I'd be mega tempted to approve, but seems like the e2e are still failing. Do you need help on that front to investigate what's going on ?

ThibodeauJF:
@olamothe there's the angular e2e sample not having the @coveo/atomic-angular package installed that's still a bit problematic

ThibodeauJF:
will reopen later perhaps

PR #2072, opened 2022-06-09
feat(atomic): add show all button to instant results

liviarett:
https://coveord.atlassian.net/browse/KIT-1714

https://user-images.githubusercontent.com/30511433/172877166-089c96dd-1458-47a2-8cb8-9b0b0355eb24.mov

This PR doesn't include the styles I applied in another PR that's still open, so this is focusing only on the button.
I also added the extra border between the two panels that I completely missed at first!

**NOTE: in this PR I reverted what I had done before with the divider stuff. I think we need something a bit more generic that will work with different kinds of suggestions, so I thought it's better to be clearer about the behaviour instead of inferring from hasQuery or not**


olamothe (approved):
There's no reason it shouldn't (famous last word), but the scenario we wanna make sure works perfectly with the "show all results" feature is with the standalone search box.

AKA: When the search box appears in the header of a website (say the homepage), someone start searching, hit "show all results" and get redirected correctly to the full search page.

Because when you are on a full search page, it's not * that * useful, obviously.

liviarett:
@olamothe 
in a standalone search box:
https://user-images.githubusercontent.com/30511433/173059913-899a516a-4efb-4237-8eac-39b6684df8b8.mov

btaillon (approved):

btaillon:
Added doc to review the strings

jpmarceau (approved):
Strings look good 👍

PR #2073, opened 2022-06-09
fix(atomic): focused on first new result when pressing "load more results"

btaillon:
https://coveord.atlassian.net/browse/KIT-1716

This was particularly more complicated than the focus for other components for the following reasons:
1. The focus is triggered by the different component than the one who receives the focus
2. Results triggered their ref twice, once with no children and once with children.
3. Results aren't actually focusable, it's the elements inside them which are focusable. This made testing a bit difficult, and I noticed inconsistencies in whether grid results were focusable.

I came up with the following solutions:
1. The result list and folded result list expose a focus method. The "load more results" button uses a query selector on the search interface (found from its bindings, no changes needed there) to find either result list.
2. If the ref is called and the result is empty, then the ref is ignored. Additionally, `focusOnNextTarget` is called instead of `focusAfterSearch`.
3. The `<a>` element of grid results was given `tabindex="-1"` and `aria-hidden="true"`.


packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -152,6 +162,16 @@ export class ResultListCommon {
     );
   }
 
+  public focusOnNextNewResult(
+    resultListState: FoldedResultListState | ResultListState
+  ) {
+    if (!this.nextNewResultTarget) {
+      throw "Cannot focus on next new result if nextNewResultTarget isn't defined";
```
olamothe, line 169: Is this caught anywhere ? Wouldn't this bubble up to our users even though it's essentially an internal error that they cannot do much about ?

olamothe (approved):

packages/atomic/src/components/result-lists/result-list-common.tsx
btaillon: It isn't caught anywhere, it's meant as an assert. If an error is thrown here, it means there's an error in our code.

ThibodeauJF (changes requested):
change "result list fetching" pattern

ThibodeauJF (approved):

PR #2075, opened 2022-06-10
feat(atomic,headless): add support for allowed values for specific facets

olamothe:
https://coveord.atlassian.net/browse/KIT-1699


packages/headless/src/controllers/core/facets/_common/facet-option-definitions.ts
```diff
@@ -38,3 +38,19 @@ const facetSearchOptionDefinitions: SchemaDefinition<FacetSearchRequestOptions>
 export const facetSearch = new RecordValue({
   values: facetSearchOptionDefinitions,
 });
+
+export const allowedValues = new RecordValue({
+  options: {required: false},
+  values: {
+    type: new StringValue({
```
btaillon, line 45: What does `type: "simple"` mean? Why is it only available on the core facet?

packages/atomic/src/components/facets/atomic-facet/atomic-facet.tsx
```diff
@@ -178,6 +178,28 @@ export class AtomicFacet implements InitializableComponent, BaseFacet<Facet> {
...
+   *
+   * ```html
+   * <atomic-facet field="objecttype" allowed-values="Contact,Account,File"></div>
+   * ```
+   *
+   * The maximum amount of allowed values is 25.
+   *
+   * Default value is `undefined`, and the facet uses all available values for its `field` in the current result set.
+   */
+  @Prop() public allowedValues?: string;
```
btaillon, line 201: I recall commenting on this but I don't remember where, that I wonder if we should support `string[]` props (declared similarly to MapProp) so that in React and Angular we can pass an array directly.

What do you think?

olamothe: We could do that, yes. But it seems out of scope of this PR. There's many props in Atomic like this, so we would need to go at it in one go, not only for this one.

packages/headless/src/controllers/core/facets/_common/facet-option-definitions.ts
olamothe: That's how the API function. It means that for now it only support the `simple` mode, where we match allowed values exactly.

They added that to leave them room to add more complex match mode, like regex or wildcards if we need it eventually.

PR #2076, opened 2022-06-10
feat(headless): add Insight controllers to support the result list

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2203

This PR adds the required controllers to support the result list component for the Insight use case.

Please note that the `executeSearch` and `fetchMoreResults` actions are actually insight-specific as they target the Customer Service API instead of the Search API.


packages/headless/src/controllers/insight/search-status/headless-insight-status.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export interface InsightStatusState extends CoreStatusState {}
+
+/**
+ * Creates a `InsightStatus` controller instance.
+ *
+ * @param engine - The headless engine.
+ * @returns A `InsightStatus` controller instance.
+ */
+export function buildInsightStatus(engine: InsightEngine): InsightStatus {
```
mmitiche, line 26: Not sure about why we need  the `buildInsightStatus` function that simply  calls the `buildCoreStatus` without any additions? will this controller contain additional logic specific to the insight panel in the future?

mmitiche (approved):
Looks good to me!

packages/headless/src/controllers/insight/search-status/headless-insight-status.ts
lbergeron: I'd say the main reason is to enforce the engine type.

Suppose we expose the `buildCoreStatus(engine: CoreEngine)` function directly in the Insight headless bundle. Then, adding any insight-specific logic in the future would cause a breaking change because we'd need to change the type of the `engine` parameter to `InsightEngine` which is more specific than `CoreEngine`. Having the `buildInsightStatus` function give us more freedom to implement insight-specific functionalities in the future without breaking the API.

packages/headless/src/controllers/insight/result-list/headless-insight-result-list.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A scoped and simplified part of the headless state that is relevant to the `InsightResultList` controller.
+ * */
+export interface InsightResultListState extends CoreResultListState {}
+
+export function buildInsightResultList(
+  engine: InsightEngine,
+  props?: InsightResultListProps
+): InsightResultList {
+  return buildCoreResultList(engine, {
```
nathanlb, line 43: IMO in cases like this where there's very little to modify from the core behaviour I think injecting a creator like this is pretty neat!
If we need to do more modifications in different controllers then overriding the exposed functions makes sense.
TLDR I approve of this solution 👍

packages/headless/src/controllers/insight/result-list/headless-insight-result-list.ts
```diff
@@ -0,0 +1,47 @@
...
+  buildCoreResultList,
+  CoreResultListOptions,
+  CoreResultListState,
+} from '../../core/result-list/headless-core-result-list';
+
+export interface InsightResultListOptions extends CoreResultListOptions {}
+
+export interface InsightResultListProps {
+  /**
+   * The options for the `ResultList` controller.
```
nathanlb, line 14:
```suggestion
   * The options for the `InsightResultList` controller.
```

packages/headless/src/controllers/insight/results-per-page/headless-insight-results-per-page.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export type {
+  ResultsPerPage,
+  ResultsPerPageProps,
+  ResultsPerPageInitialState,
+  ResultsPerPageState,
+};
+
+/**
+ * Creates a `ResultsPerPage` controller instance.
```
nathanlb, line 20:
```suggestion
 * Creates a `InsightResultsPerPage` controller instance.
```

packages/headless/src/controllers/insight/result-list/headless-insight-interactive-result.ts
```diff
@@ -0,0 +1,60 @@
...
+export interface InsightInteractiveResultProps
+  extends InteractiveResultCoreProps {
+  /**
+   * The options for the `InteractiveResult` controller.
+   */
+  options: InsightInteractiveResultOptions;
+}
+
+/**
+ * The `InteractiveResult` controller provides an interface for triggering desirable side effects, such as logging UA events to the Coveo Platform, when a user selects a query result.
```
nathanlb, line 29:
```suggestion
 * The `InsightInteractiveResult` controller provides an interface for triggering desirable side effects, such as logging UA events to the Coveo Platform, when a user selects a query result.
```

packages/headless/src/controllers/insight/result-list/headless-insight-interactive-result.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export interface InsightInteractiveResult extends InteractiveResultCore {}
+
+/**
+ * Creates an `InteractiveResult` controller instance.
```
nathanlb, line 34:
```suggestion
 * Creates an `InsightInteractiveResult` controller instance.
```

packages/headless/src/controllers/insight/result-list/headless-insight-interactive-result.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param engine - The headless engine.
+ * @param props - The configurable `InteractiveResult` properties.
```
nathanlb, line 37:
```suggestion
 * @param props - The configurable `InsightInteractiveResult` properties.
```

packages/headless/src/controllers/insight/result-list/headless-insight-interactive-result.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns An `InteractiveResult` controller instance.
```
nathanlb, line 38:
```suggestion
 * @returns An `InsightInteractiveResult` controller instance.
```

packages/headless/src/controllers/insight/quickview/headless-insight-quickview.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface InsightQuickviewProps extends QuickviewCoreProps {}
+export interface InsightQuickviewOptions extends QuickviewCoreOptions {}
+export interface InsightQuickview extends QuickviewCore {}
+export interface InsightQuickviewState extends QuickviewCoreState {}
+
+/**
+ * Creates a `Quickview` controller instance.
+ * @param engine - The headless engine.
+ * @param props - The configurable `Quickview` properties.
+ * @returns A `Quickview` controller instance.
```
nathanlb, line 20:
```suggestion
 * @returns A `InsightQuickview` controller instance.
```

packages/headless/src/controllers/insight/quickview/headless-insight-quickview.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export interface InsightQuickviewProps extends QuickviewCoreProps {}
+export interface InsightQuickviewOptions extends QuickviewCoreOptions {}
+export interface InsightQuickview extends QuickviewCore {}
+export interface InsightQuickviewState extends QuickviewCoreState {}
+
+/**
+ * Creates a `Quickview` controller instance.
+ * @param engine - The headless engine.
+ * @param props - The configurable `Quickview` properties.
```
nathanlb, line 19:
```suggestion
 * @param props - The configurable `InsightQuickview` properties.
```

nathanlb (approved):
Looks good to me just left some suggestions for jsdoc comment updates. Let me know if it was intentional.
A couple requests for clarifications as well.
GG

packages/headless/src/controllers/insight/quickview/headless-insight-quickview.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns A `Quickview` controller instance.
+ */
+export function buildInsightQuickview(
```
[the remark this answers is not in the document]
lbergeron: Since the result list component is relying on the quickview controller, I figured we needed it as well.

However, as you mention, the `/html` call is not present in the back-end so clicking the quickview icon in the result list only open an empty modal box.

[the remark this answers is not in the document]
nathanlb: yeah all good. it does seem to already be necessary despite it's current uselessness 😅

packages/headless/src/controllers/insight/result-list/headless-insight-interactive-result.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export function buildInsightInteractiveResult(
```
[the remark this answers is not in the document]
nathanlb: Just spitballing here in case we can prevent from creating a new one but if we need it then we need it.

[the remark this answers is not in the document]
lbergeron: > and would loosening the argument type to something like CoreEngine which both versions extend work?

That would work now, but it would be a risky move. Suppose we change the argument type to `CoreEngine`. It would not be possible for the `buildInteractiveResult` to use anything specific to the `SearchEngine` in the future. And trying to revert the argument back to `SearchEngine` would be a breaking change. Having the `buildInsightInteractiveResult` function can seem redundant at this moment, but it allows us to use use-case-specific engine features later on.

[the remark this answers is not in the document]
nathanlb: Yeah fair enough. Keeping that flexibility is the spirit of this whole plan so you made the right call. Thanks for explaining the reasoning behind it.

olamothe (approved):

aelfaleh (approved):

PR #2077, opened 2022-06-10
feat(atomic): Hide label when not defined and when the page is in error state

mbenkirane-coveo:
Preview: 
Screen Shot 2022-06-10 at 12 04 41 PM


packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -146,9 +147,19 @@ export class AtomicSegmentedFacet
...
+      !this.facet.state.enabled
+    ) {
+      return <Hidden></Hidden>;
+    }
+
     return (
       <div class="flex">
-        <b class="inline-block my-3 mr-2">{this.label}</b>
+        <b class="inline-block my-3 mr-2">
+          {this.label !== 'no-label' ? this.label.concat(':') : ''}
```
btaillon, line 161: Instead of having a default value of `no-label` at line 61, I think we should make label optional like so:
```ts
  @Prop({reflect: true}) public label?: string;
```

That way, here, you can instead compare `this.label` against `undefined` like so:
```suggestion
          {this.label !== undefined ? this.label.concat(':') : ''}
```

The reason normal facets had a default value of `no-label` was because it's the localization string for the string "No label"/"Sans titre".

mbenkirane-coveo: Sure, will fix !

mbenkirane-coveo: For components that require a label, should I pass 'no-label' or alter those components to make a label optional

btaillon: alter those components

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -109,7 +110,10 @@ export class AtomicSegmentedFacet
   private renderValuesContainer(children: VNode[]) {
     const classes = 'box-container flex';
     return (
-      <FacetValuesGroup i18n={this.bindings.i18n} label={this.label}>
+      <FacetValuesGroup
+        i18n={this.bindings.i18n}
+        label={this.label !== undefined ? this.label : 'no-label'}
```
btaillon, line 115: The shorthand for to `this.label !== undefined ? this.label : 'no-label'` is `this.label ?? 'no-label'`.

However, I would make `label` optional in `FacetValuesGroup`

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -146,9 +150,19 @@ export class AtomicSegmentedFacet
...
+    ) {
+      return <Hidden></Hidden>;
+    }
+
     return (
       <div class="flex">
-        <b class="inline-block my-3 mr-2">{this.label}</b>
+        <b class="inline-block my-3 mr-2">
+          {this.label !== undefined ? this.label.concat(':') : ''}
+        </b>
```
btaillon, line 165: If there's no label, I would remove the `b` element entirely, especially since it has a margin.

```suggestion
        {this.renderLabel()}
```
...
```ts
public renderLabel() {
  if (!this.label) {
    return;
  }
  return (
    <b class="inline-block my-3 mr-2">
      {this.label.concat(':')}
    </b>
  );
}
```

packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -24,7 +24,7 @@ import {
 
 interface Breadcrumb {
   facetId: string;
-  label: string;
+  label?: string;
```
ThibodeauJF, line 27:
```suggestion
  label: string;
```

I don't think that's needed anymore?

mbenkirane-coveo:
> I might not recall correctly, but didn't we remove the `registerFacetToStore` ?

oh yeah, we did.  I'll fix that

packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -283,7 +283,9 @@ export class AtomicBreadbox implements InitializableComponent {
       .flat()
       .map(({value, facetId, field}) => ({
         facetId,
-        label: this.bindings.store.state.facets[facetId].label,
+        label: this.bindings.store.state.facets[facetId]
```
mbenkirane-coveo, line 286: Had to do it this way to avoid getting the error "Cannot read property of undefined" in the console.

packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
mbenkirane-coveo: what should we assign label in facetBreadcrumbs() if we can't assign undefined to it?

packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
liviarett: how about `this.bindings.store.state.facets[facetId]?.label` ?

packages/atomic/src/components/atomic-breadbox/atomic-breadbox.tsx
liviarett: you can give it an empty string

packages/atomic/src/components/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -145,10 +139,25 @@ export class AtomicSegmentedFacet
     );
   }
 
+  private renderLabel() {
+    if (!this.label) {
+      return;
+    }
+    return <b class="inline-block my-3 mr-2">{this.label.concat(':')}</b>;
```
liviarett, line 146:
```suggestion
    return <b class="inline-block my-3 mr-2">{this.label}:</b>;
```

mbenkirane-coveo:
> Hmm careful, seems everything that was done in this PR is now undone? https://github.com/coveo/ui-kit/pull/2068/files When on master it's still there...

Yes ! I was gonna merge master into this branch, when the changes for this branch are approved.

PR #2081, opened 2022-06-10
fix(atomic): focused on correct printable uri part when expanded

btaillon:
https://coveord.atlassian.net/browse/KIT-1728

I also refactored the cypress tests for this component a little bit.


packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
```diff
@@ -72,8 +72,8 @@ export class AtomicResultPrintableUri {
     });
   }
 
-  private getIndexOfEllipsis(parentsCount: number) {
-    return Math.min(parentsCount - 1, this.maxNumberOfParts);
```
btaillon, line 76: I'm not sure in what situation an ellipsis would be displayed and `parentsCount - 1` would be smaller than `this.maxNumberOfParts`...

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -75,8 +75,10 @@ export function FocusTarget() {
           element = el;
           if (focusOnNextTarget) {
             focusOnNextTarget = false;
-            element.focus();
-            onFocusCallback?.();
+            setTimeout(() => {
+              el.focus();
+              onFocusCallback?.();
+            });
```
btaillon, line 81: Same as https://github.com/coveo/ui-kit/pull/2078#discussion_r894688354

I noticed some inconsistencies with the focus on this component too, after all.

packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
ThibodeauJF: we only display an ellipsis when there is more than 1 element that it would unveil, like the breadbox does IIRC

if maxNumberOfParts = 3
`great-grand-parent > ... > parent > child` ❌ 
`great-grand-parent > grand-parent > parent > child` ✔️ 

now it's broken
Screen Shot 2022-06-10 at 3 41 41 PM
Screen Shot 2022-06-10 at 3 41 44 PM

ThibodeauJF (changes requested):

packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
ThibodeauJF: Actually it was already broken 😂 could you fix it in that PR?

btaillon:
> we only display an ellipsis when there is more than 1 element that it would unveil, like the breadbox does IIRC
> 
> if maxNumberOfParts = 3
> `great-grand-parent > ... > parent > child` ❌ 
> `great-grand-parent > grand-parent > parent > child` ✔️ 
> 
> now it's broken
> Screen Shot 2022-06-10 at 3 41 41 PM
> Screen Shot 2022-06-10 at 3 41 44 PM

I'll fix that and add tests to ensure this doesn't get broken again.

packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
```diff
@@ -73,7 +73,9 @@ export class AtomicResultPrintableUri {
   }
 
   private getIndexOfEllipsis(parentsCount: number) {
-    return Math.min(parentsCount - 1, this.maxNumberOfParts);
+    const valuesToHide = Math.max(2, parentsCount - this.maxNumberOfParts);
+    const valuesToShowBeforeEllipsis = parentsCount - valuesToHide - 1;
+    return valuesToShowBeforeEllipsis;
```
btaillon, line 78: These three lines are exactly equivalent to:
```ts
return Math.min(this.maxNumberOfParts - 1, parentsCount - 3);
```

However, I think it's much more readable like this.

btaillon:
I made this behave exactly like `atomic-result-multi-value-text`, where `maxNumberOfValues` is extremely strict.

If `maxNumberOfValues` is 3:
* If there's 4 values available, then 2 are displayed (2 are collapsed)
* If there's 5 values available, then 3 are displayed (2 are collapsed)
* If there's 6 values available, then 3 are displayed (3 are collapsed)
* And so on...

@ThibodeauJF

ThibodeauJF (approved):
🚀

packages/atomic/src/components/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
olamothe: Love this, readability over one-liner any days of the week !

olamothe (approved):

PR #2083, opened 2022-06-13
feat(atomic): handle fields to include in instant results

liviarett:
https://coveord.atlassian.net/browse/KIT-1733

### DIFF
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -841,6 +841,10 @@ export namespace Components {
          */
         "display": ResultDisplayLayout;
         /**
+          * A list of non-default fields to include in the query results, separated by commas.
+         */
+        "fieldsToInclude": string;
+        /**
           * The expected size of the image displayed in the results.
          */
         "imageSize": ResultDisplayImageSize;
@@ -2511,6 +2515,10 @@ declare namespace LocalJSX {
           * The desired layout to use when displaying results. Layouts affect how many results to display per row and how visually distinct they are from each other.
          */
         "display"?: ResultDisplayLayout;
+        /**
+          * A list of non-default fields to include in the query results, separated by commas.
+         */
+        "fieldsToInclude"?: string;
         /**
           * The expected size of the image displayed in the results.
          */
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -248,6 +248,7 @@ export namespace Components {
         "density": ResultDisplayDensity;
         /**
           * A list of non-default fields to include in the query results, separated by commas.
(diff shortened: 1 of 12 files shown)


packages/atomic/src/components.d.ts
```diff
@@ -2511,6 +2515,10 @@ declare namespace LocalJSX {
           * The desired layout to use when displaying results. Layouts affect how many results to display per row and how visually distinct they are from each other.
          */
         "display"?: ResultDisplayLayout;
+        /**
+          * A list of non-default fields to include in the query results, separated by commas.
+         */
+        "fieldsToInclude"?: string;
```
btaillon, line 2521: This change to `components.d.ts` implies that `fieldsToInclude` was also added to the search box component. Is that accidental?

liviarett: Screenshot 2022-06-13 at 13 28 36
Screenshot 2022-06-13 at 13 28 48

both changes on this file are under `interface AtomicSearchBoxInstantResults {`, as far as I can tell. I think you missed lines 834 and 2509. can you confirm for me, please?? I didn't even touch the search box!

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -202,6 +207,14 @@ export class ResultListCommon {
       ) as ResultTemplate<DocumentFragment>[]
     );
 
+    const fieldsToInclude =
```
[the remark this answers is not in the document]
liviarett: option 1: I thought about that, but the result list wouldn't really be needed.. would you still prefer that? happy to do it that way. 
option 2: also thought of it but `registerFieldsToInclude` doesn't seem to be exported, unless I'm mistaken!

liviarett:
I always thought it didn't make much sense to add the fields to include to the result list, because I think it's more of a search thing than a display thing - if we were just filtering from a list of fields, I'd agree that it should stay there, however since it changes the query itself it feels more natural to me to have it as top level as possible.

And also, even though you usually won't have both together, you could do, and the user might not realise, but the fields you add to one will also be queried for the other (not sure if one will cancel the other or not) as we are using the same engine.

So my personal opinion is for it to be on the interface!

liviarett:
I made an attempt, let me know what you guys think @olamothe @btaillon @ThibodeauJF  I can revert

ThibodeauJF (approved):
I like it!

btaillon (approved):

liviarett:
@olamothe since you were part of the discussion, can you confirm you are happy with it so I can merge? 🙂

olamothe:
Yeah, its fine !

PR #2085, opened 2022-06-13
fix: fully update lockfiles before building

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1719
https://threads.com/34427105366

### DIFF
--- a/.github/actions/commit-generated-files/action.yml
+++ b/.github/actions/commit-generated-files/action.yml
@@ -4,7 +4,7 @@ runs:
   using: composite
   steps:
     - run: |
-        if [ ! -z "$(git status --porcelain)" ]; then  
+        if [ ! -z "$(git status --porcelain  -- ':!*package-lock.json')" ]; then  
           git fetch --depth=1
           git checkout ${{ github.head_ref }}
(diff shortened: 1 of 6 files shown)


.github/actions/setup/action.yml
```diff
@@ -1,22 +1,11 @@
 name: 'Setup & build'
 description: 'Setup node, install dependencies & build'
-inputs:
-  with-registry:
-    description: Whether we should use the npmjs.org registry
-    required: false
-    default: 'false'
```
ThibodeauJF, line 7: I think this is dead config https://github.com/coveo/ui-kit/pull/1621

.github/actions/commit-generated-files/action.yml
```diff
@@ -4,7 +4,7 @@ runs:
   using: composite
   steps:
     - run: |
-        if [ ! -z "$(git status --porcelain)" ]; then  
+        if [ ! -z "$(git status --porcelain  -- ':!*package-lock.json')" ]; then
```
ThibodeauJF, line 7: Without this, will always run when we run install with package-lock changes

olamothe (approved):

PR #2087, opened 2022-06-13
feat(headless): add search box controller to the Insight use case

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2202

At this moment the PR diff is not clean because it depends on both #2041 and  #2076.

This PR makes the search box controller available to the Insight use case. To do so,
* the existing search box controller was refactored as a core controller
* an `InsightSearchBox` controller was created

**Limitation**
The query suggest is not yet supported by the `InsightSearchBox` controller because the corresponding back-end call is not available on the Customer Service API.


packages/headless/src/controllers/core/search-box/headless-core-search-box.ts
```diff
@@ -0,0 +1,271 @@
...
+
+export type {SearchBoxOptions, SuggestionHighlightingOptions, Delimiters};
+
+export interface CoreSearchBoxProps {
+  /**
+   * The `CoreSearchBox` controller options.
+   */
+  options?: SearchBoxOptions;
+
+  executeSearchActionCreator?: (
```
olamothe, line 58: Why is this optional ? I think it should be required, and we should remove the check on line 182:

```typescript
if (props.executeSearchActionCreator) {
  await dispatch(props.executeSearchActionCreator(analytics));
}
```

Reasoning is that is should not be possible for a search box extending the core search box to be missing important functionality.

And if there is a good reason for a search box implementation to not expose a functioning `submit()` function that perform a search and retrieve new results, then it means the `submit()` function is probably not "core" for the search box, and that it should probably be removed from the `core-search-box`, and moved to each specific search box implementation.

packages/headless/src/controllers/core/search-box/headless-core-search-box.ts
```diff
@@ -0,0 +1,271 @@
...
+      dispatch(updateQuerySetQuery({id, query: value}));
+      this.showSuggestions();
+    },
+
+    clear() {
+      dispatch(updateQuerySetQuery({id, query: ''}));
+      dispatch(clearQuerySuggest({id}));
+    },
+
+    showSuggestions() {
```
olamothe, line 200: Similar to my above comment, we should either make `props.fetchQuerySuggestionsActionCreator` required, and remove the check on line `202`, or move the `showSuggestions()` function to each specific implementation.

If the insight panel backend service does not yet support query suggestions, we could add a no-op action with a `TODO` to implement it.

packages/headless/src/controllers/core/search-box/headless-core-search-box.ts
lbergeron: You are right, there is no good reason to make this action optional.

packages/headless/src/controllers/core/search-box/headless-core-search-box.ts
```diff
@@ -0,0 +1,271 @@
...
+
+export interface CoreSearchBoxProps {
+  /**
+   * The `CoreSearchBox` controller options.
+   */
+  options?: SearchBoxOptions;
+
+  executeSearchActionCreator?: (
+    arg: SearchAction
+  ) => AsyncThunkAction<any, SearchAction, any>;
```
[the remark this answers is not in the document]
lbergeron: Unfortunately `AnyAction` doesn't work because the `AsyncThunkAction<>` does not have a `type` attribute, which makes sense. Also, I could not find another type matching any async thunk action.

ThibodeauJF (approved):

nathanlb (approved):
Is the insight search box controller exported?

lbergeron:
@nathanlb 
> Is the insight search box controller exported?

Not yet. I figured we could expose them when they are needed in Quantic. It gives a little more room for non-breaking changes to the controllers if needed.

I might be too cautious, but that's the idea behind this choice.

packages/headless/src/controllers/insight/search-box/headless-insight-search-box.ts
```diff
@@ -0,0 +1,87 @@
...
+   * The `InsightSearchBox` controller options.
+   */
+  options?: SearchBoxOptions;
+}
+
+/**
+ * The `InsightSearchBox` headless controller offers a high-level interface for designing a common search box UI controller
+ * with [highlighting for query suggestions](https://docs.coveo.com/en/headless/latest/usage/highlighting/).
+ */
+export interface InsightSearchBox extends Controller {
```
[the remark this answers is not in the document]
lbergeron: Good point! That would make sense indeed.

nathanlb:
> @nathanlb
> 
> > Is the insight search box controller exported?
> 
> Not yet. I figured we could expose them when they are needed in Quantic. It gives a little more room for non-breaking changes to the controllers if needed.
> 
> I might be too cautious, but that's the idea behind this choice.

Not that's fine as long as it's conscious. For testing I assume you export it on your machine then. All good.

aelfaleh (approved):
good stuff 🥂

PR #2079, opened 2022-06-10
feat(atomic): release instant results and add react and angular support

liviarett:
https://coveord.atlassian.net/browse/KIT-1715

PR #2091, opened 2022-06-14
fix(atomic-react): allow atomic-react to use synthetic events in result templates

olamothe:
Original discussion about the problem: https://discuss.coveo.com/t/atomic-react-event-attributes-not-triggered-in-result-templates/7979/3

Since the way result templates were done in AtomicReact was essentially passing a rendered HTML string from React to `atomic-result-list` (which it would then pass down to `atomic-result` to handle as an HTML template content), all events handlers configured on result templates elements by the end user React application in JSX would never be executed.

This changes the logic to so that instead of using `atomic-result` to do the rendering with `innerHTML`, we leave React to do it's job: We give it a root element to attach itself to (ie: `result-root`), and then React is able to call `createRoot(...).render(...)`.

This `createRoot` function is similar to what you normally use in an `App.tsx`, for example: 
```
ReactDOM.render(<App />, document.getElementById('root'));
```

https://coveord.atlassian.net/browse/KIT-1732


packages/atomic/src/components/atomic-result/atomic-result.tsx
```diff
@@ -89,6 +90,15 @@ export class AtomicResult {
    */
   @Prop() loadingFlag?: string;
 
+  /**
+   * Internal function used by atomic-result-list in advanced setup, that allows to bypass the standard HTML template system.
+   * Particularly useful for Atomic React
+   */
```
ThibodeauJF, line 96: `@internal` ?

packages/atomic/src/components/result-lists/list-display-results.tsx
```diff
@@ -12,16 +12,14 @@ export const ListDisplayResults: FunctionalComponent<ResultsProps> = (
         result={result}
         engine={props.bindings.engine}
         store={props.bindings.store}
-        display={props.display}
-        density={props.density}
-        imageSize={props.imageSize}
```
btaillon, line 17: Where does `atomic-result` receive the list's display/density/imageSize after this change?

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -340,8 +339,9 @@ export class ResultListCommon {
 }
 
 export type ResultRenderingFunction = (
-  result: Result | FoldedResult
-) => HTMLElement;
+  result: Result | FoldedResult,
+  root: HTMLElement
+) => string;
```
btaillon, line 344: I find the syntax with both the `root` parameter and the `string` return type a bit confusing. Would it be possible to use a syntax like `(result: Result | FoldedResult) => string | (result: Result | FoldedResult, root: HTMLElement) => void`, and infer the html content from the root's `innerHTML`? The return function could even optionally return a `Promise` that we await before reading the root's content.

olamothe: When `createRoot(root).render(template(rest));` is executed, `root.innerHTML` won't necessarily be rendered, it's really up to React internal, which decides when it schedule the paint.

I can check if there is a way to "be sure" that it's rendered.

packages/atomic/src/components/result-lists/result-list-common.tsx
```diff
@@ -37,22 +37,21 @@ import {
...
 
-export interface BaseResultList<T = FoldedResult | Result>
-  extends InitializableComponent {
+export interface BaseResultList extends InitializableComponent {
   host: HTMLElement;
   templateHasError: boolean;
   resultListCommon: ResultListCommon;
 
-  setRenderFunction?: SetRenderFunction<T>;
+  setRenderFunction?: SetRenderFunction;
```
liviarett, line 45: the idea behind doing this was removing the need to check what kind of result it is in the rendering function, for example `function MyTemplate(result: FoldedResult)` works when passed to `AtomicFoldedResultList` and `function MyTemplate(result: Result) {` works when passed to `AtomicResultList`.
is the user now going to have to do 
```ts
function MyTemplate(result: Result | FoldedResult) {
  if (isFolded(result)) { etc... }
```
or is there some other magic that does this that I didn't see?

liviarett:
we'll need to do the same for the instant results stuff after [this](https://github.com/coveo/ui-kit/pull/2079) is merged

packages/atomic/src/components/result-lists/list-display-results.tsx
```diff
@@ -12,16 +12,14 @@ export const ListDisplayResults: FunctionalComponent<ResultsProps> = (
...
         content={props.getContentOfResultTemplate(result)}
         loadingFlag={props.resultListCommon.loadingFlag}
         ref={(element) =>
           element &&
           props.indexOfResultToFocus === index &&
           props.newResultRef?.(element)
         }
+        {...props}
```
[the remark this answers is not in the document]
olamothe: Yeah that's fine, since the types expected by `atomic-result-result` are strict. 

For example, if we wanted to use `newResultRef` inside `atomic-result`, we'd have to explicitly add it to the props of `atomic-result` (or some weird cast to any), or it's otherwise not accessible.

packages/atomic/src/components/result-lists/result-list-common.tsx
olamothe: The cast happens internally in `ResultListWrapper` and `FoldedResultListWrapper` in Atomic React, which is code we own.

For implementers using Atomic React, they are still receiving the correct type in their templates, so it's transparent and they don't have to do any cast.

By removing the generic here, it's simpler to reuse the `ResultRenderingFunction` type across more files.

PR #2094, opened 2022-06-14
docs(atomic): add E2E tests for instant results

liviarett:


packages/atomic/cypress/integration/search-box/instant-results/search-box-instant-results.cypress.ts
```diff
@@ -0,0 +1,186 @@
...
+        clickUri: `about:blank?${i}`,
+        raw: {
+          urihash: `${i}`,
+        },
+      }));
+
+      response.send(200, newResponse);
+    });
+  }).as(TestFixture.interceptAliases.Search.substring(1));
+};
```
btaillon, line 32: This could be simplified like so:
```suggestion
const setInstantResults = (count: number) =>
  (fixture: TestFixture) => {
    fixture.withCustomResponse(response => {
      response.results = Array.from({length: count}, (_, i) => ({
        title: `Instant Result ${i}`,
        uniqueId: `instant_result_${i}`,
        uri: `about:blank?${i}`,
        clickUri: `about:blank?${i}`,
        raw: {
          urihash: `${i}`,
        },
      }));
    });
  };
```

packages/atomic/cypress/integration/search-box/instant-results/search-box-instant-results.cypress.ts
```diff
@@ -0,0 +1,186 @@
...
+    describe('when hovering over a query', () => {
+      before(() => {
+        SearchBoxSelectors.querySuggestions().eq(0).trigger('mouseover');
+      });
+      SearchBoxAssertions.assertSuggestionIsSelected(0);
+      SearchBoxAssertions.assertHasText('');
+
+      describe('when hovering over an instant result', () => {
+        before(() => {
+          cy.wait(1000);
```
btaillon, line 154: I would avoid `cy.wait` if possible.

Cypress already has [a mechanism](https://docs.cypress.io/guides/core-concepts/retry-ability) whenever we `find` or `get` to wait up to 4 seconds for the element to exist. If `InstantResultsSelectors.results().eq(1)` doesn't exist prior to this `wait`, Cypress should already await it.

btaillon (approved):

packages/atomic/cypress/integration/search-box/instant-results/search-box-instant-results.cypress.ts
liviarett: I think the problem here is that the element exists (but it's a different element, since we're finding it by index). I'll see what I can do!

olamothe (approved):

PR #2104, opened 2022-06-15
feat(atomic): warn user when instant results component is implemented…

liviarett:
… without query suggestions

https://coveord.atlassian.net/browse/KIT-1749


packages/atomic/src/pages/examples/instantResults.html
```diff
@@ -29,7 +29,7 @@
       <atomic-search-layout mobile-breakpoint="700px">
         <atomic-layout-section section="search">
           <atomic-search-box suggestion-timeout="1000">
-            <atomic-search-box-recent-queries icon="assets://xml.svg"> </atomic-search-box-recent-queries>
+            <!-- <atomic-search-box-recent-queries icon="assets://xml.svg"> </atomic-search-box-recent-queries> -->
```
btaillon, line 32: Did you mean to commit this change?
```suggestion
            <atomic-search-box-recent-queries icon="assets://xml.svg"> </atomic-search-box-recent-queries>
```

ThibodeauJF:
Screen Shot 2022-06-16 at 10 18 10 AM
What about this use case? it will be a different component althogether?

liviarett:
> Screen Shot 2022-06-16 at 10 18 10 AM
> 
> What about this use case? it will be a different component althogether?

yeah, I'm thinking so! but in any case, if we decide to reuse we can tweak the logic a little to account for that

ThibodeauJF:
IMO it should adapt to be that behaviour when alone in the <atomic-search-box> children, but we can put a TODO and warn in the meantime!

ThibodeauJF (approved):

liviarett:
> I think I would prefer if there was some configuration we could provide that means "These suggestions are dependent on focusing other suggestions", and the search box could take care of logging the error. Chances are that this may also be useful for accessibility later on.

@btaillon I'm not sure, I think the logic in `atomic-search-box` is already quite convoluted and I'd prefer to leave this responsibility to its own component.. I would agree if we were doing the accessibility stuff that you mentioned, but if that's just something that might come in the future, I think we can reassess then?

PR #2105, opened 2022-06-16
fix(atomic): hide instant results on mobile on page resize

liviarett:
https://coveord.atlassian.net/browse/KIT-1750

This is what happens when you start already on mobile (because I don't update the suggested query when on mobile, no results ever show)
https://user-images.githubusercontent.com/30511433/174060378-fb951f15-910a-427f-be0e-0da3475fe297.mov

However when you resize to mobile, we currently don't update the suggested query, which means that the render function in instant results still shows results for it.
https://user-images.githubusercontent.com/30511433/174060458-05678805-f3cc-495a-862b-2acc92c10762.mov

So I did two things here: update the state of the search box for mobile (clearing the suggested query), but also make sure that the render function doesn't return anything for mobile

### DIFF
--- a/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
@@ -661,20 +661,19 @@ export class AtomicSearchBox {
   }
 
   private async updateSuggestedQuery(suggestedQuery: string) {
-    if (!isMobile(this.bindings.store)) {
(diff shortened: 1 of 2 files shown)


btaillon (approved):

liviarett:
> Unrelated but I think if the video format is mp4, we'll be able to visualize them directly 😄

I was wondering why sometimes it showed and sometimes not! file format never crossed my mind haha very helpful 🙏

PR #2106, opened 2022-06-16
fix(quantic): display clear all filters button only when there is active filters

mmitiche:
[SFINT-4538](https://coveord.atlassian.net/browse/SFINT-4538)
### The problem:
The "Clear all filters" button was always displayed, even if no filter was active.

### The fix:
The clear all filters button is now displayed only when there is at least one active filter.

https://user-images.githubusercontent.com/86681870/174064471-a4719c17-dbcf-48da-ae45-3f8f54728982.mov

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
@@ -92,7 +92,7 @@ export default class QuanticRefineModalContent extends LightningElement {
   }
 
   /**
-   * Updates the hasActiveFilters properety.
+   * Updates the hasActiveFilters property.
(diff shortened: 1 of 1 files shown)


nathanlb (approved):

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -85,6 +91,14 @@ export default class QuanticRefineModalContent extends LightningElement {
     }
   }
 
+  /**
+   * Updates the hasActiveFilters properety.
```
lbergeron, line 95:
```suggestion
   * Updates the hasActiveFilters property.
```

lbergeron (approved):
Nice!

PR #2107, opened 2022-06-16
fix(atomic): only loads focus-visible polyfill on unsupported browsers

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1725

https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#browser_compatibility


packages/atomic/src/global.ts
```diff
@@ -0,0 +1,21 @@
...
+    document.body.querySelector(':focus-visible');
+  } catch (error) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    import('focus-visible/dist/focus-visible.min.js');
+  }
+}
+
+export default function () {
+  setCoveoGlobal();
```
ThibodeauJF, line 19: Should be an improvement (won't have to load in all interfaces)

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -82,10 +82,25 @@ export interface InitializableComponent extends ComponentInterface {
...
+  if (!element.shadowRoot) {
+    return;
+  }
+
+  if (window.applyFocusVisiblePolyfill) {
     window.applyFocusVisiblePolyfill(element.shadowRoot);
+    return;
   }
+
+  window.addEventListener(
```
ThibodeauJF, line 99: when lazy loaded this is necessary https://github.com/WICG/focus-visible#lazy-loading

packages/atomic/src/global/global.ts
```diff
@@ -0,0 +1,51 @@
...
+    import('focus-visible/dist/focus-visible.min.js');
+  }
+}
+
+export default function () {
+  if (!window) {
+    return;
+  }
+
+  setCoveoGlobal();
```
ThibodeauJF, line 49: Should be an improvement (won't have to load in all interfaces)

olamothe (commented):
What would happen if we did * not * load the polyfill at all ?

Crash in old Safari ? Or something else ?
Because if it's the latter and it's not "too bad"... how about we just don't do anything ?

ThibodeauJF:
> Crash in old Safari ? Or something else ?

Old before 15.4 which release 2-3 months ago, but I'm not against removing it in the future

PR #2111, opened 2022-06-16
feat(atomic): update instant results on user input

liviarett:
https://coveord.atlassian.net/browse/KIT-1755

https://coveo.slack.com/files/U02HVCMTRV3/F03KZLNN927/screen_recording_2022-06-16_at_17.29.54.mov

### DIFF
--- a/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/atomic-search-box/atomic-search-box.tsx
@@ -194,9 +194,10 @@ export class AtomicSearchBox {
       searchBoxController: this.searchBox,
       numberOfQueries: this.numberOfQueries,
       suggestedQuery: () => this.suggestedQuery,
-      clearSuggestions: () => this.clearSuggestions(),
-      triggerSuggestions: () => this.triggerSuggestions(),
       getSuggestions: () => this.suggestions,
+      clearSuggestions: () => this.clearSuggestions(),
+      triggerSuggestions: (suggestedQuery?: string) =>
+        this.triggerSuggestions(suggestedQuery),
     };
   }
   private updateBreakpoints = once(() => updateBreakpoints(this.host));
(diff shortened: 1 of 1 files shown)


olamothe (changes requested):
This is not a good idea 😢 

I understand it solves UI problems, and makes our (Atomic dev) lives easier, but that's not a direction we should go.

A good "search as you type" feature is not as simple as "a query on every keystroke", unfortunately. 

You'll see that Google and such do not do something like that, and there's good reasons.

Whenever we show results to end user, it needs to be because we are fairly certain that the results are gonna be good (ie: relevant), not just because we don't have anything better to show.

The query suggestion feature powered by ML is our (semi) assurance that the above will be true: it returns suggestions for which it knows there is a good chance of a good outcome (ie, a click, add to cart, personalized for the end user, etc.)

If we just use partial keywords typed in the search box when we don't have suggestions, it's going to recommend random-ish documents in the index to the end user.

Said differently, if the ML model is not able to return at least one suggestion for the letters currently typed in the search box, there's very high chance that if we query the index with those letters, we'll get weird bad performing documents/products that we will display to the end user.

Also, the QPM consumption will be very high. By just playing with the implementation in this PR, and typing with a normal speed, the amount of rejected query that throws errors in the console should be an indicator. 

It means a standard Coveo for commerce implementation will probably explode their QPM consumption by a factor of... 10x or 100x.

liviarett:
well, @olamothe you cut my wings nicely 😂 
I guess these problems have all been seen before, and I got too excited about this great solution and forgot about them.
But I am a bit confused, since this is part of the designs:
Screenshot 2022-06-16 at 22 33 06

does it mean Johnny got too excited when he designed this and we're not doing it, or that we need to do it but it will require a lottt more work?

olamothe:
Yeah, we won't be implementing this any time soon ;)

PR #2112, opened 2022-06-16
fix(atomic): fixed modal focus with VoiceOver on iOS

btaillon:
https://coveord.atlassian.net/browse/KIT-1726

When opening a modal on iOS using VoiceOver, like the refine modal, the focus wasn't entering the modal and couldn't be fixed with the standard swipe left/right.

I found the following to be the cause:
1. The close couldn't be focused if the modal was being animated.
2. The button responsible for opening the modal can't be re-focused if the modal is being animated on top of it. 
3. The modal couldn't be immediately focused unless it already existed in the DOM prior to the button being clicked.

Fixing issue 3 by appending the modal to the DOM as soon as possible made me find another issue: when the modal was closed, it wasn't removed from the tabindex nor the accessibility DOM.

I made the following changes:
1. Moved the responsibility for the focus from the modal to the focus trap.
  Any component that would implement the focus trap would've needed to manage the focus, so this seems more logical anyway.
2. On iOS exclusively, await the open/close animation before activating/deactivating the focus trap, which triggers the focus.
3. Added `tabindex="-1"` and `aria-hidden="true"` to the modal while the focus trap is deactivated.
4. Removed the logic which appended the modal only once it's initially opened, and instead appended the modal ASAP.
  The old behaviour existed because appending a closed modal would show the close animation as soon as the page loads. I fixed that by adding a new tailwind utility which skips the initial animation.

I wish I could've decided not to try to fix this, but this bug is pretty catastrophic for screen readers.


packages/atomic/src/components/atomic-modal/atomic-modal.tsx
```diff
@@ -46,40 +46,53 @@ export class AtomicModal implements InitializableComponent {
...
     if (isOpen) {
       this.wasEverOpened = true;
       document.body.classList.add(modalOpenedClass);
-      this.focusOnFirstElement();
+      if (isIOS()) {
+        await this.waitForAnimationEnded();
+      }
+      if (watchToggleOpenExecution !== this.currentWatchToggleOpenExecution) {
+        return;
+      }
```
btaillon, line 77: Don't want a race condition

packages/atomic/src/components/atomic-refine-toggle/atomic-refine-toggle.tsx
```diff
@@ -78,7 +74,13 @@ export class AtomicRefineToggle implements InitializableComponent {
         class="p-3 w-full"
         onClick={() => this.enableModal()}
         text={this.bindings.i18n.t('sort-and-filter')}
-        ref={(button) => (this.buttonRef = button)}
+        ref={(button) => {
+          if (!button) {
+            return;
+          }
+          this.buttonRef = button;
+          this.loadModal();
```
btaillon, line 82: On `componentDidLoad`, the `buttonRef` doesn't necessarily exist yet.

packages/atomic/src/components/atomic-modal/atomic-modal.tsx
mbenkirane-coveo: How would this cause a race condition?

packages/atomic/src/utils/device-utils.ts
```diff
@@ -0,0 +1,11 @@
+export function isIOS() {
+  // Source: https://stackoverflow.com/a/58065241
+  const platform = navigator.platform;
```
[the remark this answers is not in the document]
liviarett: what about this answer here? https://stackoverflow.com/a/62094756/8454914
it doesn't include platform and seems quite comprehensive

liviarett:
the navigation looks really good!

https://user-images.githubusercontent.com/30511433/174284930-67267f3c-fc90-42f2-8df3-736af93bbc37.mp4

I noticed that the modal shows up when reloading the page though

packages/atomic/src/utils/device-utils.ts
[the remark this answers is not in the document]
btaillon: Good point, I'll use it.

packages/atomic/src/components/atomic-modal/atomic-modal.tsx
btaillon: If we open/close the modal rapidly.

btaillon:
> I noticed that the modal shows up when reloading the page though

Woops! Will fix.

packages/atomic/src/components/common/atomic-focus-trap.tsx
```diff
@@ -76,9 +84,34 @@ export class AtomicFocusTrap {
...
+
+  hideSelf() {
+    this.parentToHide.setAttribute('aria-hidden', 'true');
+    this.parentToHide.setAttribute('tabindex', '-1');
+  }
+
   @Watch('active')
-  activeChanged(active: boolean) {
-    active ? this.hideSiblingsRecursively(this.host) : this.showAll();
+  async activeChanged(active: boolean, wasActive: boolean) {
```
ThibodeauJF, line 98: this is a bit hard to read, can you move the second block of ifs in well named private methods?

packages/atomic/src/components/common/atomic-focus-trap.tsx
```diff
@@ -96,11 +129,15 @@ export class AtomicFocusTrap {
     getFirstFocusableDescendant(this.host)?.focus();
   }
 
-  connectedCallback() {
-    this.activeChanged(this.active);
+  async connectedCallback() {
```
ThibodeauJF, line 132: this could technically block the connectedCallback of the parent `atomic-search-interface` I think, can you use componentWillLoad?

ThibodeauJF (approved):

liviarett:
just so you don't forget - the modal is still showing on page load

packages/atomic/src/components/common/atomic-focus-trap.tsx
btaillon: I will avoid awaiting anything in `connectedCallback` then.

PR #2114, opened 2022-06-17
fix(quantic): automatically  closing the tab dropdown when it loses focus

mmitiche:
[SFINT-4544](https://coveord.atlassian.net/browse/SFINT-4544)
### The problem

https://user-images.githubusercontent.com/86681870/174283260-c90fc9e0-7287-4c52-b375-055647e49713.mov

### The fix
The tab bar drop-down list now automatically closes when a click is performed outside of it:

https://user-images.githubusercontent.com/86681870/174283099-15821f3b-0319-427f-b964-3d8d9a440d2e.mov


nathanlb (approved):
looks good

packages/quantic/cypress/integration/tab-bar/tab-bar.cypress.ts
```diff
@@ -131,4 +131,32 @@ describe('quantic-tab-bar', () => {
...
+      scope('when loading the page', () => {
+        Expect.displayTabs(true);
+        Expect.displayedTabsEqual([TAB_1, TAB_2]);
+        Expect.tabsInDropdownEqual([TAB_3, TAB_4]);
+        Expect.activeTabContains(TAB_1);
+        Expect.displayMoreButton(true);
+        Expect.displayMoreButtonIcon(true);
+        Expect.displayMoreButtonLabel(moreButtonLabel);
+        Expect.displayDropdown(false);
+      });
```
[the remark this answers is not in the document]
mmitiche: That's true, I got the same feeling when adding these tests x) I'm updating this by removing the  stuff we already validated👌

packages/quantic/cypress/integration/tab-bar/tab-bar.cypress.ts
```diff
@@ -131,4 +131,21 @@ describe('quantic-tab-bar', () => {
...
+    it('should automatically close the dropdown list', () => {
+      cy.viewport(mediumViewportWidth, 900);
+      visitPage();
+
+      scope('when opening the dropdown list', () => {
+        Actions.openDropdown();
+        Expect.displayDropdown(true);
+      });
+
+      scope('when a click is made outside the dropdown list', () => {
```
lbergeron, line 145:
```suggestion
      scope('when clicking outside the dropdown list', () => {
```

packages/quantic/force-app/main/default/lwc/quanticTabBar/quanticTabBar.js
```diff
@@ -296,15 +297,27 @@ export default class QuanticTabBar extends LightningElement {
    * Toggles the combobox.
    * @returns {void}
    */
-  toggleCombobox() {
+  toggleCombobox(event) {
```
lbergeron, line 300: It is out of the scope of this PR, but I wonder why the term "Combobox" is used here. A "Combobox" is a combination of an input box and a dropdown. "Dropdown" should have been used instead.

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticTabBar/quanticTabBar.js
mmitiche: Correct! There is no reason to choose the name "Combobox" over "Dropdown", I will update the naming, it will be more logical.

PR #2115, opened 2022-06-17
fix(headless): prevent facetSearch from being sent in the search request

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1753

I noticed that when investigating a maintenance case. IMO we shouldn't send parameters that aren't supposed to be there... What if one day facetSearch is an actual param? Would all the headless request instantly fail? Anyways that was my reasoning for this change 😄


btaillon (approved):

olamothe (approved):
Yes, we certainly don't want to send useless params.

Even if they don't do anything, it's still un-necessary noise that makes it harder to troubleshoot and cause confusion for people looking at our UI to understand how the API works.

PR #2116, opened 2022-06-17
feat(quantic): add QuanticInsightInterface component

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2245

This PR introduces the `QuanticInsightInterface` component. Its purpose is to bootstrap the Insight engine in order to use the Insight controllers.

An APEX `InsightController` has been added. At this moment, this controller is doing the same thing as `HeadlessController.cls`, but it will eventually call an Insight-specific token generator that would generate a platform token with limited privileges instead of a search token like `CoveoTokenGenerator` does.

The `getHeadlessBundle` function was also added to `QuanticHeadlessLoader` allowing components to retrieve the Headless bundle that matches the loaded use case (e.g., `search`, `case-assist`, `insight`, etc.). It allows a component (e.g., result list) to work with both the `search` and `insight` use cases.

All a component has to do is call `getHeadlessBundle` at initialization, and then instantiate controllers using the retrieved bundle.

```javascript
@api engineId;
resultList;

initialize = (engine) => {
  const headless = getHeadlessBundle(this.engineId);
  this.resultList = headless.buildResultList(engine);
  // continue initialization...
}
```


packages/quantic/force-app/main/default/classes/InsightController.cls
```diff
@@ -0,0 +1,20 @@
...
+     * @return {String} A serialized object containing the following keys:
+     *                      - `platformUrl`: the endpoint to contact Coveo.
+     *                      - `accessToken`: an access token (see https://docs.coveo.com/en/123/apis/get-your-coveo-access-token).
+     *                      - `organizationId`: the ID of the organization to query (see https://docs.coveo.com/en/148/manage-an-organization/retrieve-the-organization-id).
+     */
+    @AuraEnabled(cacheable=true)
+    public static String getHeadlessConfiguration() {
+        /**
+         * e.g. To switch to the Insight Token Provider, you simply need to replace the following line by:
+         * `return InsightTokenProvider.getHeadlessConfiguration();`
```
lbergeron, line 16: @mmitiche `InsightTokenProvider.getHeadlessConfiguration()` is the one that links Quantic to the Coveo for Salesforce package. It is Coveo for Salesforce that will expose a method providing the scoped platform token.

We'll then be able to add `InsightTokenProvider` to Quantic.

mmitiche: Understood 👌 Merging this [Jira](https://coveord.atlassian.net/browse/SFINT-4491) will allow us to expose the method providing the platform token in C4SF. Will try to merge it soon!

mmitiche: The method providing the scoped platform token is now exposed in the C4SF package, we can now proceed and create the InsightTokenProvider.

lbergeron: The `InsightTokenProvider` class has been created.

nathanlb (approved):
Good stuff!

aelfaleh (approved):

packages/quantic/force-app/main/default/lwc/quanticInsightInterface/quanticInsightInterface.js-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<LightningComponentBundle xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>53.0</apiVersion>
```
mmitiche, line 3: Maybe the version should be 54.0 to be consistent with all the other components.

packages/quantic/force-app/main/default/lwc/quanticInsightInterface/quanticInsightInterface.js
```diff
@@ -0,0 +1,61 @@
...
+              },
+            };
+            setEngineOptions(
+              this.engineOptions,
+              CoveoHeadlessInsight.buildInsightEngine,
+              this.engineId,
+              this,
+              CoveoHeadlessInsight
+            );
+            setInitializedCallback(this.initialize, this.engineId);
```
mmitiche, line 49: I don't think we really need to set the initialized callback here since the `initialize` method is basically doing nothing.

lbergeron: Good point, I'll check that.

mmitiche (approved):
Looking good 👌

PR #2119, opened 2022-06-20
fix(headless): stop logging errors when user aborts request in instan…

liviarett:
…t results

https://coveord.atlassian.net/browse/KIT-1743

leaves only warnings:
Screenshot 2022-06-20 at 14 24 33


ThibodeauJF:
For the errors/warnings we still shouldn't log them. We might have to refactor the SearchAPIClient to allow this. We could: 
- add an option "disableAbortWarning" & make the SearchAPIClient options public, disable the warning before an instant result request (and change it back)
- change the [pino logger level](https://github.com/pinojs/pino/blob/master/docs/api.md#logger-level) before the request so the warnings aren't logged (and change it back
- and some other solutions I'm not thinking of...

packages/headless/src/features/instant-results/instant-results-slice.ts
```diff
@@ -49,12 +49,16 @@ export const instantResultsReducer = createReducer(
...
     });
     builder.addCase(fetchInstantResults.fulfilled, (state, action) => {
-      const {results} = action.payload;
       const {cacheTimeout} = action.meta.arg;
       const cached = getCached(state, action.meta);
+
       cached!.isLoading = false;
-      cached!.error = null;
-      cached!.results = results;
+      cached!.error =
```
olamothe, line 56: Is this how we differentiate between an expected error (ie: requests happening too fast, which is "normal" with this controller) VS an actual un-expected error that we want to re-surface to the implementer ?

Why only rely on the `message` property of the error ?

liviarett: from my investigations, whenever I `rejectWithValue` in the thunk, it throws an error and I can't catch it.

however, by catching and returning an error
```
    } catch (e) {
      return e as SerializedError;
    }
```
we dispatch a `fulfilled` action with an error as a payload. 
so my thought here as to check whether it's an error by checking if the payload contains a `message`.

btaillon: By doing this, is the below `builder.addCase(fetchInstantResults.rejected, (state, action) => {` still ever called?

Would it be possible to handle the error there instead?

ThibodeauJF (approved):

packages/headless/src/api/search/search-api-client.ts
```diff
@@ -128,10 +128,12 @@ export class SearchAPIClient implements FacetSearchAPIClient {
   private searchAbortController: AbortController | null = null;
 
   async search(
-    req: SearchRequest
+    req: SearchRequest,
+    disableAbortWarning?: boolean
```
btaillon, line 132: In case we end up adding more options in the future, I would prefer to extract this new option into an interface. E.g.: `SearchOptions`.

packages/headless/src/api/search/search-api-error-response.ts
```diff
@@ -21,9 +22,29 @@ function buildDisconnectedError(
...
+}
+
+export function buildAPIResponseFromErrorOrThrow(
+  error: Error,
+  disableAbortWarning?: boolean
+): {
   error: SearchAPIErrorWithStatusCode;
 } {
+  const asDOMException = error as DOMException;
+  if (disableAbortWarning && asDOMException.code === 20) {
```
btaillon, line 43: Instead of casting `Error` to `DOMException`, would it be possible to change the parameter type from `Error` to `Error | DOMException` and then do this?
```suggestion
  if (disableAbortWarning && asDOMException?.code === 20) {
```

I haven't tested this, I don't actually know if that's something Typescript allows.

btaillon: Also, I just checked, and `code` seems to be deprecated.
This might be better:
```suggestion
  if (disableAbortWarning && error?.name === 'AbortError') {
```

packages/headless/src/api/search/search-api-error-response.ts
```diff
@@ -5,6 +5,7 @@ export interface SearchAPIErrorWithStatusCode {
   statusCode: number;
   message: string;
   type: string;
+  ignore?: boolean;
```
btaillon, line 8: Unless we treat this object as a parameter/request/option rather than a response, I would prefer `ignored` over `ignore`.

btaillon (approved):

packages/headless/src/api/search/search-api-client.ts
liviarett: okie dokie

ThibodeauJF (approved):

olamothe (approved):

PR #2123, opened 2022-06-20
fix(atomic): improved contrast between the stars and their background

btaillon:
https://coveord.atlassian.net/browse/KIT-176

By adding a brown outline, I improved the contrast between the active/inactive part of stars, and between the stars and the background behind them.

I also expanded `star.svg`'s viewbox from 12x13 to 13x13 and added a stroke size of `0.8125`.

# Before
## At normal font size with the accessibility theme
image

## At 4x browser font size with the accessibility theme
image

# After
## At normal font size with the accessibility theme
image

## At 4x browser font size with the accessibility theme
image


packages/atomic/src/components/search/atomic-rating/atomic-rating.pcss
```diff
@@ -1,13 +1,28 @@
 /**
- * @prop --atomic-rating-facet-icon-active-color: Color of the icon when active.
+ * @prop --atomic-rating-icon-active-color: Color of the icon when active.
```
btaillon, line 2: IMO, it makes no sense for `facet` to be part of the variable's name since it's not only used in facets.

ThibodeauJF (approved):

liviarett (approved):
personal opinion: I'm not a huge fan, it makes it look less modern and sleek. couldn't it be part of the accessibility theme only?

olamothe:
> personal opinion: I'm not a huge fan, it makes it look less modern and sleek. couldn't it be part of the accessibility theme only?

That would be nice indeed.

mbenkirane-coveo (approved):

PR #2125, opened 2022-06-20
fix(atomic): encode colon for i18n namespace feature

olamothe:
You could see the problem in `searchuisamples`, with the source facet.

There is a source in that org named `https://www.kcprofessional.com/`.

Atomic would display it as `https.//www.kcprofessional.com/` instead. This is due to the fact that `:` is a reserved character for the i18n namespace feature: https://www.i18next.com/principles/namespaces

~@npushkarskii : We will need to adapt this article: https://docs.coveo.com/en/atomic/latest/usage/atomic-localization/#add-of-modify-captions~

~To be able to localize facet values containing `:`, implementers will need to use the encoded version of the character ( `%3a` ).~

~For example, using the source name above:~

The above is no longer needed with Jean Felix suggestion

https://coveord.atlassian.net/browse/KIT-1772


ThibodeauJF:
What if we changed the separator to something less common? https://www.i18next.com/overview/configuration-options#others

olamothe:
> What if we changed the separator to something less common? https://www.i18next.com/overview/configuration-options#others

Oohhh interesting !

btaillon (approved):

btaillon (approved):
Awesome!!

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -37,6 +37,7 @@ import {loadDayjsLocale} from '../../../utils/dayjs-locales';
 import {loadGlobalScripts} from '../../../global/global';
 import {BaseAtomicInterface} from '../../common/interface/interface-common';
 import {CommonBindings} from '../../common/interface/bindings';
+import {initi18n} from '../../common/interface/i18n';
```
ThibodeauJF, line 40: For this file I implemented the same changes in [my PR](https://github.com/coveo/ui-kit/pull/2145) with a bit more details, I'd appreciate if this file wasn't merged 🙏

ThibodeauJF (approved):

PR #2126, opened 2022-06-20
feat(atomic): disableSearch prop for search box

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1765
We're not changing anything to the OOT behaviour. This is just to allow developers to externally prevent users from triggering searches for X condition

### DIFF
--- a/.github/actions/commit-generated-files/action.yml
+++ b/.github/actions/commit-generated-files/action.yml
@@ -4,6 +4,7 @@ runs:
   using: composite
   steps:
     - run: |
+        git status --porcelain  -- ':!*package-lock.json'
         if [ ! -z "$(git status --porcelain  -- ':!*package-lock.json')" ]; then  
           git stash
           git fetch --depth=1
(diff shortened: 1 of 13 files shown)


btaillon (approved):

olamothe (commented):
Shouldn't this be a global prop, on the search interface ?

I'm thinking the more generic use case is wanting to stop all searches from being executed, not only those originating from the search box.

ThibodeauJF:
Hmm I could make this a prop on the interface & add it to the store so other component can use it.

But it would take a lot of work to adapt all components that trigger searches to be disabled...

Also, would there be an easy way to disable searches in headless at large?

mbenkirane-coveo:
> https://coveord.atlassian.net/browse/KIT-1765

Which other components trigger searches other than the searchbox and facets? If we don't want a facet to trigger a search the user would just hide it, no?

ThibodeauJF:
Yes it's my original though too, mostly only the search box would need to be visible & interactive while preventing searches altogether

We can also keep this as a search box only prop from the time being, and if we make it general to the whole interface, we can deprecate the prop

mbenkirane-coveo:
> Yes it's my original though too, mostly only the search box would need to be visible & interactive while preventing searches altogether
> 
> We can also keep this as a search box only prop from the time being, and if we make it general to the whole interface, we can deprecate the prop

Sounds like a plan

PR #2129, opened 2022-06-21
fix(atomic): improve click interaction in instant result

liviarett:
https://coveord.atlassian.net/browse/KIT-1777

When testing the analytics events for instant results, I realised that I couldn't `cmd + click` a result and open in a new window, which is pretty annoying.
I needed to play a little with the propagation in order for it to keep the pop up open long enough for it to navigate, and decided to trigger a click on the link to keep the `target` behaviour

I also realised that clicking the actual link doesn't currently work on master
Video before the fix:

https://user-images.githubusercontent.com/30511433/174795504-97f8acd2-5782-460c-b076-e439a0dfd229.mp4

Video after the fix:

https://user-images.githubusercontent.com/30511433/174796047-0ca3ad5e-7138-4c53-81ce-5c9d79207718.mp4


packages/atomic/src/components/search/atomic-result/atomic-result.tsx
```diff
@@ -108,6 +113,11 @@ export class AtomicResult {
     event.detail(this.result);
   }
 
+  @Listen('atomic/resolveStopPropagation')
+  public resolveStopPropagation(event: CustomEvent) {
+    event.detail(this.stopPropagation);
+  }
```
btaillon, line 119: I feel like it's odd to have that property on the result instead of the link 🤔 

I'm thinking that maybe we should always stop propagation on links. I don't really see a case where we don't want to?

liviarett: I put it on `atomic-result` to make it more of an internal thing. 
If we don't propagate here, clicking on the result link won't work...

liviarett: this is what happens if I stop propagation in the result link
https://user-images.githubusercontent.com/30511433/174802196-998edb13-bcd6-459e-86d5-053cc5794f29.mp4

liviarett: not just to 'make it an internal thing' actually, but also because it simply doesn't work without it, so I don't want the user to be able to change it themselves

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -98,29 +111,33 @@ export class AtomicSearchBoxInstantResults implements BaseResultList {
...
-        onSelect: () => {
-          buildInteractiveResult(this.bindings.engine, {
-            options: {result},
-          }).select();
-          this.bindings.clearSuggestions();
-          window.location.href = result.clickUri;
-        },
-      })
+      (result: Result) => {
+        return {
```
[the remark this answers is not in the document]
liviarett: absolutely, I had added some more code in this function and forgot to change it back!

packages/atomic/src/components/search/atomic-result/atomic-result.tsx
liviarett: Ok, so to summarise my explanation from stand up:
We currently STOP propagation from the link click.
My suggestion is to allow this event to be propagated so that we can use it in the `li` component and trigger `onSelect`. Otherwise we won't be able to click the link itself when trying to navigate..

And a quick explanation about the methods from [here](https://stackoverflow.com/a/5299841/8454914):
`stopPropagation` will prevent any **parent handlers** from being executed 
`stopImmediatePropagation` will prevent any **parent handlers** and also any **other handlers** from executing

liviarett: terrible naming if you ask me!!!!

olamothe: Well programmers are very good at naming things normally: https://en.wikipedia.org/wiki/HTTP_referer

olamothe (approved):

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -90,6 +89,20 @@ export class AtomicSearchBoxInstantResults implements BaseResultList {
     }
   }
 
+  private getLink(el: HTMLElement) {
+    return el?.querySelector('atomic-result')?.shadowRoot?.querySelector('a');
```
ThibodeauJF, line 93: possible that there are more than 1 links

liviarett: yeah, I thought about that but didn't know how to go about this... maybe look for the atomic-link instead? and then any other links that don't stop propagation should already be good!

PR #2133, opened 2022-06-21
feat(headless): add insight facet controller

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2204

This PR adds the facet controller for the Insight use case. The facet manager controller was also extracted as a core controller at the same time since it is required as well.


packages/headless/src/controllers/insight/facets/facet/headless-insight-facet.test.ts
```diff
@@ -0,0 +1,270 @@
+import {buildInsightFacet, Facet, FacetOptions} from './headless-insight-facet';
```
lbergeron, line 1: These tests are copied from `headless-facet.test.ts` keeping only relevant tests.

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet-manager.ts
```diff
@@ -10,7 +11,12 @@ import {
 import {ProductListingSection} from '../../../state/state-sections';
 import {ProductListingEngine} from '../../../app/product-listing-engine/product-listing-engine';
 
-export type {FacetManager, FacetManagerState, FacetManagerPayload};
+export type {
+  CoreFacetManager,
```
nathanlb, line 15: why does it need to export this type?

packages/headless/src/controllers/insight/facets/facet/headless-insight-facet.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    showLessValues() {
+      coreController.showLessValues();
+      dispatch(fetchFacetValues(logFacetShowLess(getFacetId())));
+    },
+
+    get state() {
+      return {
+        ...coreController.state,
+        facetSearch: facetSearch.state,
```
nathanlb, line 160: The facet search actions are no-op right now correct? So interacting with this instance would simply de nothing?

nathanlb (approved):
Pretty awesome!

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet-manager.ts
lbergeron: It is because I made the `FacetManager` interface extend `CoreFacetManager`, so the `CoreFacetManager` has to be exported to be available in the API documentation.

If you think it is better, `FacetManager` can define the same interface as `CoreFacetManager` without extending it. This way the `CoreFacetManager` interface could remain private.

packages/headless/src/controllers/insight/facets/facet/headless-insight-facet.ts
lbergeron: Good point, I haven't tried enabling facet search through the Insight engine... I'll validate.

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet-manager.ts
olamothe: It's okay that `CoreFacetManager` gets exported. 

We don't really gain much by redefining the same interface again: net net, the users of the library will get an interface/type of a given shape, that they need to deal with and know about, and that we must support in a backward compatible manner.

The shape of the interface is the important thing, not which files it comes from specifically.

olamothe (approved):

packages/headless/src/controllers/insight/facets/facet/headless-insight-facet.ts
nathanlb: Yeah sounds good top me this is a temporary situation anyway.

packages/headless/src/controllers/insight/facets/facet/headless-insight-facet.ts
```diff
@@ -0,0 +1,174 @@
+import {
+  executeSearch,
+  fetchFacetValues,
+} from '../../../../features/insight-search/insight-search-actions';
+import {
+  logFacetClearAll,
+  logFacetUpdateSort,
+  logFacetShowMore,
+  logFacetShowLess,
+} from '../../../../features/facets/facet-set/facet-set-analytics-actions';
```
aelfaleh, line 10: analytics actions need to be insight analytics actions
see https://github.com/coveo/ui-kit/pull/2147

ThibodeauJF (approved):

packages/headless/src/api/analytics/insight-analytics.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      });
+    });
+
+    it('should properly return getSearchUID from searchResponseId if available', () => {
+      const state = getBaseState();
+      state.search = buildMockSearchState({searchResponseId: 'the_id'});
+      state.search.response.searchUid = 'another_id';
+      expect(new InsightAnalyticsProvider(state).getSearchUID()).toEqual(
+        'the_id'
+      );
```
erocheleau, line 108: Can we have a little bit more descriptive name for this test? It's not easily clear that you're expecting the value from `searchResponseId` even if you modify the `searchUid`.

packages/headless/src/api/analytics/insight-analytics.ts
```diff
@@ -0,0 +1,146 @@
...
+  public getSearchUID(): string {
+    return (
+      this.state.search?.searchResponseId ||
+      this.state.search?.response.searchUid ||
+      getSearchInitialState().response.searchUid
+    );
+  }
+  public getPipeline(): string {
+    return (
+      this.state.pipeline || this.state.search?.response.pipeline || 'default'
```
erocheleau, line 48: I don't think it's true that the 'default' pipeline is always the "default" :) someone could make their default pipeline named 'foo' if they wanted.

When would we need those fallbacks where both the state and the last response state are empty exactly?

packages/headless/src/api/analytics/insight-analytics.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  private get responseTime() {
+    return this.state.search?.duration || getSearchInitialState().duration;
+  }
+
+  private mapResultsToAnalyticsDocument() {
+    return this.state.search?.response.results.map((r) => ({
+      documentUri: r.uri,
+      documentUriHash: r.raw.urihash,
+    }));
```
erocheleau, line 82: permanenentId? Not 100% of the reference for this as currently it doesn't look like we send the results on search events. But at least for clicks we definitely need the permanentId

erocheleau: Okay well I validated and this is what atomic does as well I guess that is okay?

I see no mention of the results in this page: https://docs.coveo.com/en/1502/build-a-search-ui/log-search-events
Or even in the swagger doc

erocheleau (approved):
Just check a little bit the comments especially around analytics but everything looks good 👍

packages/headless/src/api/analytics/insight-analytics.test.ts
lbergeron: It was handled in the base PR https://github.com/coveo/ui-kit/pull/2147

packages/headless/src/api/analytics/insight-analytics.ts
lbergeron: It was part of the base PR https://github.com/coveo/ui-kit/pull/2147

PR #2135, opened 2022-06-21
fix(atomic): prevented additional refine modals being created when one already exists

btaillon:
https://coveord.atlassian.net/browse/KIT-1780

### DIFF
--- a/packages/atomic/src/components/search/atomic-refine-toggle/atomic-refine-toggle.tsx
+++ b/packages/atomic/src/components/search/atomic-refine-toggle/atomic-refine-toggle.tsx
@@ -40,6 +40,10 @@ export class AtomicRefineToggle implements InitializableComponent {
   }
 
   private loadModal() {
+    if (this.modalRef) {
+      return;
+    }
+
     this.modalRef = document.createElement('atomic-refine-modal');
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/search/atomic-refine-toggle/atomic-refine-toggle.tsx
```diff
@@ -40,6 +40,10 @@ export class AtomicRefineToggle implements InitializableComponent {
   }
 
   private loadModal() {
+    if (this.modalRef) {
+      return;
```
btaillon, line 44: The smart snippet modal does the same

liviarett (approved):

ThibodeauJF (approved):

packages/atomic/src/components/search/atomic-refine-toggle/atomic-refine-toggle.tsx
btaillon: I think that this logic should be owned by the creator of the modal, since from the modal's point of view, it could've been included in the HTML file (just like the modal example). Additionally, the modal isn't particularly aware of what "kind" of modal it is, it only has a reference to a container which it must hide when inactive.

PR #2136, opened 2022-06-21
feat(atomic): allowed the page to be resized down to 320px

btaillon-coveo:
https://coveord.atlassian.net/browse/KIT-1760

It's not pretty, but it doesn't look downright bugged.

Some of those changes are slightly breaking, so we could wait for v2.

Here are the changes I made:
* Updated the commerce-full page to use the search layout
* Removed the minimum width on the search layout and the query summary
* Made the pager and the results per page wrappable
  * Moved the height from their flex containers to their buttons (❗️possibly breaking)
  * Used gaps instead of margins (❗️possibly breaking)
  * Added `display: contents` to the wrapper around the pager buttons, so they have the same wrapping priority as the previous/next button (❗️possibly breaking)

image
image

### DIFF
--- a/packages/atomic/src/components/search/atomic-results-per-page/atomic-results-per-page.tsx
+++ b/packages/atomic/src/components/search/atomic-results-per-page/atomic-results-per-page.tsx
@@ -122,7 +122,10 @@ export class AtomicResultsPerPage implements InitializableComponent {
 
     return (
       <div class="flex items-center">
-        <span part="label" class="self-start text-on-background text-lg mr-3">
+        <span
+          part="label"
+          class="self-start text-on-background text-lg mr-3 leading-10"
+        >
           {this.bindings.i18n.t('results-per-page')}
         </span>
         <div


packages/atomic/src/components/search/atomic-layout/atomic-search-layout.pcss
```diff
@@ -29,7 +29,7 @@ atomic-search-layout {
   grid-template-areas:
     '. atomic-section-search .'
     '. atomic-section-main .';
-  grid-template-columns: var(--atomic-layout-spacing-x) minmax(18rem, 1fr) var(--atomic-layout-spacing-x);
+  grid-template-columns: var(--atomic-layout-spacing-x) minmax(0, 1fr) var(--atomic-layout-spacing-x);
```
btaillon, line 32: Not sure why there was a minimum width here. Any clue?

olamothe: Last change on this line was about fixing things for table layout 🤔

olamothe (approved):
I think it's fine to merge this now without waiting for v2

packages/atomic/src/components/search/atomic-results-per-page/atomic-results-per-page.tsx
```diff
@@ -122,14 +122,14 @@ export class AtomicResultsPerPage implements InitializableComponent {
 
     return (
       <div class="flex items-center">
-        <span part="label" class="text-on-background text-lg mr-3">
+        <span part="label" class="self-start text-on-background text-lg mr-3">
```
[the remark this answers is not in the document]
btaillon: Woops 🤦  gonna fix that. I think adding a line height should fix it.

PR #2140, opened 2022-06-21
refactor(atomic): begin to extract generic initialization for specific interfaces

olamothe:
I'm once again sorry for the really big diff to review... 
It's hard when touching such core interface/types to try and limit the scope of file changed.

Important changes in this PR include:

* Refactor the stencil store, in order to get different store for each interface. Right now, the common store used by all interface will contain the `loadingFlags` as well as the `assets` path (ie: where to load icons).
* The store is now a wrapper around the stencil one, by tackling on functions instead of standalone function. ie: `setLoadingFlag(store, theflag)` becomes `store.setLoadingFlag(theflag)`. This helps with the `bindings` property of each interface, which contains the `store` object, which now needs to be generic.
* Extract some shared logic for i18n.
* Start to refactor how `bindings` work, and everything that happens in `initialization-utils`. This is still a work in progress: When I do those changes, there will be even more file changes, and it was starting to get really really big.
* Start to implement what it would looks like when we start a new interface (ie: `atomic-recs-interface`, `atomic-svc-insight-interface`). This is still a work in progress, and I will refine in the following days.

https://coveord.atlassian.net/browse/KIT-1742

### DIFF
--- a/packages/atomic/src/utils/result-utils.spec.ts
+++ b/packages/atomic/src/utils/result-utils.spec.ts
@@ -3,7 +3,7 @@ import {
   getSampleSearchEngineConfiguration,
 } from '@coveo/headless';
 import {TestUtils} from '@coveo/headless';
-import {Bindings} from './initialization-utils';
+import {Bindings} from '../components/search/atomic-search-interface/atomic-search-interface';
 import {buildStringTemplateFromResult} from './result-utils';
 
 describe('buildStringTemplateFromResult', () => {


liviarett:
I'm a wee bit confused about the naming
Screenshot 2022-06-22 at 11 10 58
why do we have a `search-interface` folder inside `common`? shouldn't that be called `interface` only? 

also, there is a problem in `/src/utils/result-utils.spec.ts` with the importing of Bindings

packages/atomic/src/components.d.ts
```diff
@@ -1081,6 +1113,42 @@ export namespace Components {
          */
         "label": string;
     }
+    interface AtomicSvcInsightInterface {
```
liviarett, line 1116: Why is it called svc?

packages/atomic/src/components/common/search-interface/bindings.ts
```diff
@@ -0,0 +1,29 @@
+import {HTMLStencilElement} from '@stencil/core/internal';
+import {i18n} from 'i18next';
+import {AnyEngineType} from './search-interface-common';
+
+/**
+ * Bindings passed from the `AtomicSearchInterface` to its children components.
```
liviarett, line 6: should this be 'passed from an interface'?

packages/atomic/src/components/search/atomic-search-interface/store.ts
```diff
@@ -0,0 +1,116 @@
...
+    isMobile() {
+      return !window.matchMedia(
+        makeDesktopQuery(commonStore.state.mobileBreakpoint)
+      ).matches;
+    },
+  };
+}
+
+// https://terodox.tech/how-to-tell-if-an-element-is-in-the-dom-including-the-shadow-dom/
+function isInDocument(element: Node) {
```
liviarett, line 104: I don't think this function makes sense here, does it? I'd suggest leaving it in the common store for now and then moving to utils later?

packages/atomic/src/components/search/atomic-sort-dropdown/atomic-sort-dropdown.tsx
```diff
@@ -9,15 +9,15 @@ import {
   SearchStatusState,
 } from '@coveo/headless';
 import {
-  Bindings,
   BindStateToController,
   InitializableComponent,
   InitializeBindings,
 } from '../../../utils/initialization-utils';
```
liviarett, line 15: how do you guys feel about having aliases for folders like utils and common? so we can to
`@utils/initialization-utils` instead of `../../../utils/initialization-utils`.
I know Sami removed the aliases in the Qubit project and I don't remember what his reasons were, but I can ask if you guys think it could be an improvement!

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -149,3 +149,9 @@ export function getFirstFocusableDescendant(
   }
   return null;
 }
+
+function getSearchUIDFromState(bindings: any): string {
```
liviarett, line 153: I know this `any` (and the other ones that have warnings in this PR) is temporary, but would `unknown` work instead and remove the warnings?

packages/atomic/src/components.d.ts
olamothe: For `service`. I did a poll in a shared channel with other teams, and this was what seemed to reach a consensus.

packages/atomic/src/components/search/atomic-search-interface/store.ts
```diff
@@ -0,0 +1,116 @@
...
+    getAllFacets() {
+      return {
+        ...commonStore.state.facets,
+        ...commonStore.state.dateFacets,
+        ...commonStore.state.categoryFacets,
+        ...commonStore.state.numericFacets,
+      };
+    },
+
+    isMobile() {
```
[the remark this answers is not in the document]
olamothe: I prefer to leave it here since, so far, we (obviously) did not need it for other specific interface. When it's needed, we can move it. It's not complicated.

My logic is that I want to leave common store as barebone as possible, and not tackle things that we might or might not need later. Because it's easier to "move down" a function than it is to remove it.

packages/atomic/src/components/insight/atomic-insight-interface/atomic-insight-interface.tsx
```diff
@@ -1,16 +1,131 @@
...
+const FirstRecommendationExecutedFlag = 'firstInsightRequestExecuted';
+export type Bindings = CommonBindings<
+  InsightEngine,
+  ReturnType<typeof createAtomicSvcInsightStore>,
+  HTMLAtomicSvcInsightInterfaceElement
+>;
 
 /**
- * The `atomic-insight-interface` component is the parent to all other atomic components in a insight panel interface. It handles the headless insight panel engine and localization configurations.
+ * The `atomic-recs-interface` component is the parent to all other atomic components in a recommendation interface. It handles the headless recommendation engine and localization configurations.
```
ThibodeauJF, line 21: filename is insight

packages/atomic/src/components/search/atomic-sort-dropdown/atomic-sort-dropdown.tsx
ThibodeauJF: Worth a try!

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -1,40 +1,23 @@
...
-  /**
-   * Global state for Atomic
-   */
-  store: ObservableMap<AtomicStore>;
-  /**
-   * A reference to the `AtomicSearchInterface` element.
-   */
-  interfaceElement: HTMLAtomicSearchInterfaceElement;
-}
+type AnyBindings = CommonBindings<AnyEngineType, any, HTMLStencilElement>;
```
ThibodeauJF, line 16: We can add an eslint ignore here

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -138,25 +131,22 @@ export function InitializeBindings() {
...
-            this.bindings.i18n.off('languageChanged', updateLanguage);
-
-          try {
-            // When no controller is initialized, updating a property with a State() decorator, there will be no re-render.
-            // In this case, we have to manually trigger it.
-            this.initialize ? this.initialize() : forceUpdate(this);
-          } catch (e) {
-            this.error = e as Error;
-          }
+      const event = buildCustomEvent(initializeEventName, (bindings: any) => {
```
ThibodeauJF, line 134: eslint ignore here too or another type

packages/atomic/src/components.d.ts
ThibodeauJF: Service

packages/atomic/src/components/search/atomic-sort-dropdown/atomic-sort-dropdown.tsx
olamothe: Yeah good idea !

olamothe:
> why do we have a search-interface folder inside common? shouldn't that be called interface only?

Yep ! Will rename.

packages/atomic/src/utils/initialization-utils.tsx
btaillon: Instead of `any`, could we use `AnyBindings`?

btaillon (approved):

packages/atomic/src/components/search/atomic-sort-dropdown/atomic-sort-dropdown.tsx
olamothe: After the internal slack discussion, I would say this this might be worth visiting, but out of scope of this PR.

packages/atomic/src/utils/initialization-utils.tsx
olamothe: Yeah, those any are not good. I want to tackle more refactoring of the types here as part of a separate PR, however.

This PR is starting to become very large as it is. I'll tackle some temp eslint rules.

packages/atomic/src/components/search/atomic-sort-dropdown/atomic-sort-dropdown.tsx
liviarett: for sure!!

liviarett (approved):

PR #2144, opened 2022-06-22
fix(atomic): improved aria labels in facets

btaillon:
https://coveord.atlassian.net/browse/KIT-1782

The "show more", "show less" and "clear filters" buttons lacked an aria label with more context about which facet they belong to.

Those labels already existed, but for the "clear filters" it was only on the "title" and for "show more"/"show less" it used the wrong prop on a functional component.

image


packages/atomic/src/components/search/facets/facet-header/facet-header.tsx
```diff
@@ -54,6 +54,7 @@ export const FacetHeader: FunctionalComponent<FacetHeaderProps> = (props) => {
         part="clear-button"
         class="flex items-baseline max-w-full p-2 text-sm"
         title={clearFiltersForFacet}
+        ariaLabel={clearFiltersForFacet}
```
ThibodeauJF, line 57: aria label + title, won't screen readers repeat both?

olamothe: Yeah I vaguely remember problems about that in JSUI.

btaillon: I do recall that.

In this particular case, I'll remove the `title` since I see no use for it. However, in other cases, I checked and of the 5 screen readers we support, only TalkBack seems to redundantly read both, so I don't think it's too bad to give both of them the same value.

PR #2145, opened 2022-06-22
feat(atomic): additional control over outgoing i18n locales requests

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1773

I'll merge this one after https://github.com/coveo/ui-kit/pull/2140 not to cause you too much issues @olamothe 

Basically this PR allows us to do 2 things:
1. Not request locales files (e.g. en.json) which we know we don't have (new script to pull all available locales in a json at build time)
2. Not request locales files for other namespace than the "translation" one which is the default (we we're doing it for other namespaces such as caption-filetype)

### DIFF
--- a/packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
+++ b/packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
@@ -42,6 +42,7 @@ import {
 } from '../../../utils/local-storage-utils';
 import {loadDayjsLocale} from '../../../utils/dayjs-locales';
 import {loadGlobalScripts} from '../../../global/global';
+import availableLocales from '../../../generated/availableLocales.json';
 
 export type InitializationOptions = SearchEngineConfiguration;
 
@@ -64,6 +65,7 @@ export class AtomicSearchInterface {
   private initialized = false;
   private store = createStore<AtomicStore>(initialStore());
   private i18nPromise!: Promise<TFunction>;
+  private i18nNamespace = 'translation';
 
   @Element() private host!: HTMLAtomicSearchInterfaceElement;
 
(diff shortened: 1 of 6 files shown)


packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -484,9 +486,45 @@ export class AtomicSearchInterface {
...
+    return availableLocales.includes(locale.toLowerCase());
+  }
+
   private get i18nBackendOptions(): BackendOptions {
     return {
-      loadPath: `${getAssetPath(this.languageAssetsPath)}/{{lng}}.json`,
+      loadPath: `${getAssetPath(
+        this.languageAssetsPath
+      )}/{{lng}}.json?lng={{lng}}&ns={{ns}}`,
+      request: async (_options, url, _payload, callback) => {
```
ThibodeauJF, line 498: https://github.com/i18next/i18next-http-backend#backend-options

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -484,9 +486,45 @@ export class AtomicSearchInterface {
...
+        try {
+          const [fetchUrl, searchParams] = url.split('?');
+          const urlParams = new URLSearchParams(searchParams);
+          const lng = urlParams.get('lng')!;
+          const ns = urlParams.get('ns')!;
+
+          if (!this.isI18nLocaleAvailable(lng)) {
+            throw new Error(`Unsupported locale "${lng}"`);
```
btaillon, line 506: I would mention in the message that this isn't a locale we support out of the box, and link instructions on how to add locales.

btaillon (approved):

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
ThibodeauJF: This won't be output to the user, it's caught right away

We shouldn't output a msg to the user here either, if he sets `es-ES` it will fail a first time but then fetch `es` which is fine too

ThibodeauJF: i18n basically fails quietly for unsupported languages and defaults to English

olamothe (approved):

PR #2146, opened 2022-06-22
chore(cypress): fix facet test

nathanlb:
- Modified the test not to expect a specific facet value
- Regenerated package-lock (trying to fix two issues at once 🤞)


btaillon (approved):

ThibodeauJF (approved):
i'm scared of that big package(-lock)

PR #2147, opened 2022-06-22
feat(headless): Add analytics actions to insight-search

aelfaleh:
This Pr complete the second part for: https://coveord.atlassian.net/browse/SVCC-2246

### DIFF
--- a/packages/headless/src/api/analytics/insight-analytics.test.ts
+++ b/packages/headless/src/api/analytics/insight-analytics.test.ts
@@ -1,0 +1,120 @@
+import {CoveoAnalyticsClient} from 'coveo.analytics';
+import pino from 'pino';
+import {getConfigurationInitialState} from '../../features/configuration/configuration-state';
+import {buildMockFacetRequest} from '../../test/mock-facet-request';
+import {buildMockFacetResponse} from '../../test/mock-facet-response';
+import {buildMockFacetValue} from '../../test/mock-facet-value';
+import {buildMockInsightState} from '../../test/mock-insight-state';
+import {buildMockQueryState} from '../../test/mock-query-state';
+import {buildMockResult} from '../../test/mock-result';
+import {buildMockSearchState} from '../../test/mock-search-state';
+import {
+  configureInsightAnalytics,
+  InsightAnalyticsProvider,
+  StateNeededByInsightAnalyticsProvider,
+} from './insight-analytics';
+
+describe('insight analytics', () => {
+  const logger = pino({level: 'silent'});
+
+  it('should be enabled by default', () => {
+    const state = buildMockInsightState();
+    expect(
+      configureInsightAnalytics({state, logger}).coveoAnalyticsClient instanceof
+        CoveoAnalyticsClient
+    ).toBe(true);
+  });
+
+  it('should be enabled if explicitly specified', () => {
+    const state = buildMockInsightState();
+    state.configuration.analytics.enabled = true;
+
+    expect(
+      configureInsightAnalytics({state, logger}).coveoAnalyticsClient instanceof
+        CoveoAnalyticsClient
+    ).toBe(true);
+  });
+
+  it('should be disabled if explicitly specified', () => {
+    const state = buildMockInsightState();
+    state.configuration.analytics.enabled = false;
+
+    expect(
+      configureInsightAnalytics({state, logger}).coveoAnalyticsClient instanceof
+        CoveoAnalyticsClient
+    ).toBe(false);
+  });
+
+  describe('search analytics provider', () => {
+    const getBaseState = (): StateNeededByInsightAnalyticsProvider => ({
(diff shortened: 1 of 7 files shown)


packages/headless/src/features/facets/facet-set/facet-set-insight-analytics-actions.ts
```diff
@@ -0,0 +1,161 @@
...
+  AnalyticsType,
+  makeInsightAnalyticsAction,
+} from '../../analytics/analytics-utils';
+import {
+  buildFacetBaseMetadata,
+  getStateNeededForFacetMetadata,
+  buildFacetSelectionChangeMetadata,
+} from './facet-set-analytics-actions-utils';
+
+export const logFacetShowMore = (facetId: string) =>
```
lbergeron, line 19: @aelfaleh  As we discussed it seems we need to duplicate the analytics actions in order to connect them to the Insight analytics client. However, I see that the product listing facet is using the standard search actions. Now I'm wondering why product listing can reuse the search analytics actions while insight can't.

Can someone confirm if we're doing things the right way?

nathanlb: It's a shame we have to duplicate all this stuff. @aelfaleh @olamothe this is definitely the way to go?

packages/headless/src/features/insight-search/insight-search-analytics-actions.ts
```diff
@@ -0,0 +1,27 @@
...
+export const logQueryError = (error: SearchAPIErrorWithStatusCode) =>
+  makeInsightAnalyticsAction(
+    'search/queryError',
+    AnalyticsType.Search,
+    (client, state) =>
+      client.logQueryError({
+        query: state.query?.q || getQueryInitialState().q,
+        aq: '',
+        cq: '',
+        dq: '',
```
nathanlb, line 23: Are these really required?

packages/headless/src/features/facets/facet-set/facet-set-insight-analytics-actions.ts
aelfaleh: @nathanlb if we choose to create a client analytics per use-case, yes

olamothe: I'm missing some context before I can answer.

Are the analytics events that are to be sent by the insight panel supposed to go through the standard analytics service ? Using the... "old"... search event system ? Or is it supposed to do it differently ?

Commerce events in general (product listing included) are handled in a very different manner, so it might not be a good idea to copy what product listing feature is doing.

If the "old analytics system" is what the insight panel needs to use, then this could probably be done in a more intelligent manner with some refactoring, with much less duplication.

I'm currently pretty swamped but I can try to take a note to look into check-outing this branch and giving it a go...

nathanlb: We want it to track analytics the same way the current insight panel does which is the old search system. injecting a client would be more efficient if it isn't too limiting. @aelfaleh have you seen how that could be done and how it compares to the approach in this PR?

aelfaleh: I'll try it

lbergeron (approved):

olamothe (approved):

erocheleau (approved):
It's looking good with the client injection imo, is that "resolved" or are you still looking for a different ways to do things?

Other than that some small suggestions perhaps on naming things.

packages/headless/src/api/analytics/insight-analytics.ts
```diff
@@ -0,0 +1,146 @@
...
+import {BaseAnalyticsProvider} from './base-analytics';
+
+export type StateNeededByInsightAnalyticsProvider = ConfigurationSection &
+  Partial<
+    SearchHubSection &
+      SearchSection &
+      PipelineSection &
+      QuerySection &
+      SectionNeededForFacetMetadata
+  >;
```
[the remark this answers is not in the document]
aelfaleh: I user prettier to format the document,  it is the same in all project

packages/headless/src/api/analytics/insight-analytics.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+describe('insight analytics', () => {
+  const logger = pino({level: 'silent'});
+
+  it('should be enabled by default', () => {
+    const state = buildMockInsightState();
+    expect(
+      configureInsightAnalytics({state, logger}).coveoAnalyticsClient instanceof
+        CoveoAnalyticsClient
+    ).toBe(true);
```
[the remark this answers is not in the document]
aelfaleh: in the case when the client is disabled it will return `NoopAnalytics` 
in coveo.analytics.js:
image

PR #2152, opened 2022-06-23
fix(atomic): fix arrow navigation on input

liviarett:
https://coveord.atlassian.net/browse/KIT-1785

vid with instant results:

https://user-images.githubusercontent.com/30511433/175350032-4cfee06b-ea85-43c5-9936-a0a4ab8f40e6.mp4

vid without instant results:

https://user-images.githubusercontent.com/30511433/175350118-501b5ace-ae79-4aef-9340-66cd08a46b22.mp4

### DIFF
--- a/packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
+++ b/packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
@@ -11,7 +11,7 @@ import {
   SearchBoxSuggestions,
   SearchBoxSuggestionsBindings,
 } from '../suggestions-common';
-import {once, randomID} from '../../../../utils/utils';
+import {once} from '../../../../utils/utils';
 import {SafeStorage, StorageItems} from '../../../../utils/local-storage-utils';
 
 /**
--- a/packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
+++ b/packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
@@ -11,9 +11,8 @@ import {
   SearchBoxSuggestions,
   SearchBoxSuggestionsBindings,
 } from '../suggestions-common';
-import {once} from '../../../../utils/utils';
+import {once, randomID} from '../../../../utils/utils';
 import {SafeStorage, StorageItems} from '../../../../utils/local-storage-utils';
-import {cleanUpString} from '../../../../utils/string-utils';
 
 /**
  * The `atomic-search-box-recent-queries` component can be added as a child of an `atomic-search-box` component, allowing for the configuration of recent query suggestions.
@@ -156,7 +155,7 @@ export class AtomicSearchBoxRecentQueries {
   private renderItem(value: string): SearchBoxSuggestionElement {
     const query = this.bindings.searchBoxController.state.value;
     return {
-      key: `recent-${cleanUpString(value)}`,
+      key: randomID('recent'),
       query: value,
       content: (
         <div class="flex items-center break-all">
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -156,7 +155,7 @@ export class AtomicSearchBoxRecentQueries {
   private renderItem(value: string): SearchBoxSuggestionElement {
     const query = this.bindings.searchBoxController.state.value;
     return {
-      key: `recent-${cleanUpString(value)}`,
+      key: randomID('recent'),
```
liviarett, line 158: this was causing issues when we had, for instance `dress` and `dress=`. it would select both!

ThibodeauJF: this will always re-render though, what if we were encoding the string instead of cleaning it. It would be a replaceby instead of a remove

ThibodeauJF: https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript

Screen Shot 2022-06-23 at 1 19 07 PM

liviarett: I can definitely try that, but it only rerenders when we hover over a new result, no? like, as you change the results, the dom elements disappear anyway, so is there a problem with the id changing for the next render?

liviarett: I'll remove this from this PR anyway so we can merge the fix quicker if you guys approve

liviarett, line 158:
```suggestion
      key: `recent-${cleanUpString(value)}`,,
```

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -11,9 +11,8 @@ import {
   SearchBoxSuggestions,
   SearchBoxSuggestionsBindings,
 } from '../suggestions-common';
-import {once} from '../../../../utils/utils';
+import {once, randomID} from '../../../../utils/utils';
```
liviarett, line 14:
```suggestion
import {once} from '../../../../utils/utils';
```

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -13,7 +13,6 @@ import {
 } from '../suggestions-common';
 import {once} from '../../../../utils/utils';
 import {SafeStorage, StorageItems} from '../../../../utils/local-storage-utils';
```
liviarett, line 15:
```suggestion
import {SafeStorage, StorageItems} from '../../../../utils/local-storage-utils';
import {cleanUpString} from '../../../../utils/string-utils';
```

ThibodeauJF (approved):

olamothe (approved):

olamothe:
Would be nice to add a cypress tests to verify there is no regression around this in the future.

liviarett:
> Would be nice to add a cypress tests to verify there is no regression around this in the future.
> 
> ie: that the search box is still correctly usable as a "basic: input, even though it's much more than that with the system to provide suggestions.

will do in a separate PR, i'll merge this so it's fixed

PR #2157, opened 2022-06-24
test(atomic): add tests for arrow functionality in instant results/se…

liviarett:
…archbox

https://coveord.atlassian.net/browse/KIT-1787

### DIFF
--- a/packages/atomic/cypress/integration/search-box/instant-results/search-box-instant-results.cypress.ts
+++ b/packages/atomic/cypress/integration/search-box/instant-results/search-box-instant-results.cypress.ts
@@ -10,6 +10,9 @@ import * as SearchBoxAssertions from '../search-box-assertions';
 import * as InstantResultsAssertions from './search-box-instant-results-assertions';
 import {InstantResultsSelectors} from './search-box-instant-results-selectors';
 
+const delay = (force = false) => ({delay: 200, force});
+const downKeys = (count: number) => Array(count).fill('{downarrow}').join('');
+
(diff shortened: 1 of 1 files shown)


packages/atomic/cypress/integration/search-box/instant-results/search-box-instant-results.cypress.ts
```diff
@@ -10,6 +10,9 @@ import * as SearchBoxAssertions from '../search-box-assertions';
 import * as InstantResultsAssertions from './search-box-instant-results-assertions';
 import {InstantResultsSelectors} from './search-box-instant-results-selectors';
 
+const delay = (force = false) => ({delay: 200, force});
```
btaillon, line 13: If the goal is to make sure some results are rendered every time, then I would avoid using a delay and I would instead assert that suggestions have changed after every character. A good example would be `typeFacetSearchQuery` in `facet-common-actions`.

btaillon (approved):

ThibodeauJF (approved):

olamothe (approved):

packages/atomic/cypress/integration/search-box/instant-results/search-box-instant-results.cypress.ts
liviarett: Screenshot 2022-06-27 at 16 54 51
ok, so I could try doing that for example, but the problem is that every time I call `.type`, it blurs the element and I lose the queries panel, so it doesn't work...
I think it will need to be delay for this one

PR #2158, opened 2022-06-24
fix(atomic): dedup recent queries keys

liviarett:
https://coveord.atlassian.net/browse/KIT-1788


packages/atomic/src/utils/string-utils.ts
```diff
@@ -9,3 +9,19 @@ export function cleanUpString(str: string) {
...
+  let hash = 0,
+    i,
+    chr;
+  if (str.length === 0) {
+    return hash;
+  }
+  for (i = 0; i < str.length; i++) {
+    chr = str.charCodeAt(i);
+    hash = (hash << 5) - hash + chr;
+    // hash |= 0; // Convert to 32bit integer
```
liviarett, line 24: this seemed unnecessary to me, but I'm not sure how these bits work. if anyone thinks it should stay I can add it, if not I can remove it

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -156,7 +156,7 @@ export class AtomicSearchBoxRecentQueries {
   private renderItem(value: string): SearchBoxSuggestionElement {
     const query = this.bindings.searchBoxController.state.value;
     return {
-      key: `recent-${cleanUpString(value)}`,
+      key: `recent-${hashString(value)}`,
```
btaillon, line 159: Did you encounter some kind of error when the key wasn't cleaned? As far as I'm aware, this isn't an HTML attribute, so I would expect any string to work.

Could we use `value` directly?

packages/atomic/src/utils/string-utils.ts
btaillon: I would prefer either the native browser [digest method](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) or [this stack overflow answer](https://stackoverflow.com/a/52171480).

ThibodeauJF: TIL

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
liviarett: I had errors when using the instant results `uniqueId`, I can't remember exactly what were the characters that made it break. That's why I thought it safer clean up, but `cleanUpString` also broke when you had recent queries like `['dress', 'dress=']`

packages/atomic/src/utils/string-utils.ts
liviarett: I also didn't know about this! But I think the async method would be a bit annoying... I can change to the other method, but out of curiosity @btaillon, why do you think it's better, is it the performance?
I chose this one because it was a bit easier to read and we don't need anything fancy.. but I just ran some tests and indeed it seems to be much faster
Screenshot 2022-06-27 at 16 02 26
first values are the one imported from java and the second the stack overflow you mentioned, and the values are an average of several tests against several strings

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
olamothe: * The `cleanUpString` function that was used in this file seems to be used in other places in the codebase, for things that will ultimately be output in the DOM as an attribute. So it means it could break in all those other places too ? We should probably fix it everywhere.

* In `cleanUpString`, instead of trying to match on character that can be allowed, would it work to just match on anything not `A-Za-z` and replace with '' ?

```
export function cleanUpString(str: string) {
  return str.replace(/[^a-zA-Z]/g, '').toLocaleLowerCase()
}
```

* This type of function could be easily tested with some basic unit tests (does not need to be full blown Cypress tests). There are already some type of unit tests in the project.

The `hashString` function just seems a bit overkill 😉

liviarett: yeah, I agree that is seems a bit overkill... but your suggestion doesn't really fix it, as it's essentially what it currently does but excluding numbers and spaces.. so you would actually have more clashes than before. for example

```
                                 current func                 your func
['dress', 'dress=']      =>   ['dress', 'dress']          ['dress', 'dress'] 
['dr ess', 'dress=']     =>   ['dr-ess', 'dress']         ['dress', 'dress'] 
['dres s34', 'dre ss']   =>   ['dres-s34', 'dre-ss']      ['dress', 'dress'] 
```

Also, the reason I left instant results and query suggestions the way they were (using `cleanUpString`) is that those values don't come from the user - instant results uses the uniqueId (for which we know what to expect) and query suggestions will hopefully send you valid and 'clean' values that won't clash, so I expect that to be safe. Having said that, bugs lie where you least expect them, so might as well replace it for them all

packages/atomic/src/utils/string-utils.ts
btaillon: yes, I remembered reading this stackoverflow thread a long time ago and reading that this approach was better, but I didn't remember the reason was performance until I read it yesterday to link it here.

IMO both solutions that don't use the digest method are pretty ugly and the chances of us changing parts of them are pretty much null, so I would rather pick the best performance option. I am surprised though that there is such a gap in performance between the two solutions, I thought it would be a very very minor improvement.

I agree that async is a bit annoying, I wish that the digest method was synchronous, I have no idea why it isn't. I trust you to pick whichever solution you see fit.

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
btaillon: uh, I wouldn't expect `key` to be output in the DOM. First and foremost, I would recommend trying it to see if it works, and if it doesn't, this might potentially be a bug with Stencil?

As a last resort, instead of hashing the string, you could take the raw string and replace spaces with something like `%20`.

ThibodeauJF: Wasn't it cleaned up because querySelector was used on it or something like that?

liviarett: ok guys, I think I have a good solution
Screenshot 2022-06-28 at 15 58 37
as you can see, quite a few characters break the selector
below are a few of the encoding functions and their results:
`cyrb53` works fine
`btoa` is deprecated (could be replaced by `buf.toString('base64')`, but as you can see it sometimes outputs `=`s which don't work
`encodeURIComponent` which is meant for URLs doesn't work because it has `%`s.
`hashIt` (copyright pending) seems to work alright - implementation below

Screenshot 2022-06-28 at 16 07 53

```js
function hashIt(str) {
  return str
    .split("")
    .map((ch) => (ch.match(/(\d|\w)+/g) ? ch : ch.charCodeAt(0)))
    .join("");
}
```

what do you guys think? we could call it `encodeForDomAttribute` or something like that, as it's not really hashed

ThibodeauJF: Alright! I don't get all the hashing hate then ! 😄

ThibodeauJF: Solutions look good!

liviarett: udpated!

btaillon (approved):

packages/atomic/src/utils/string-utils.ts
```diff
@@ -2,10 +2,9 @@ export function regexEncode(value: string): string {
   return value.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&');
 }
 
-export function cleanUpString(str: string) {
+export function encodeForDomAttribute(str: string) {
```
olamothe, line 5: This would be a very easy and useful function to unit test !

olamothe (approved):

PR #2159, opened 2022-06-24
test(atomic): improve category facet failing test

liviarett:
https://coveord.atlassian.net/browse/KIT-1789

I can't guarantee 100% passing rate, but it does seem to make it better!

### DIFF
--- a/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
@@ -44,180 +44,180 @@ describe('Category Facet Test Suites', () => {
       new TestFixture().with(addCategoryFacet()).init();
     }
 
-    // describe('verify rendering', () => {
-    //   before(setupWithDefaultSettings);
+    describe('verify rendering', () => {
+      before(setupWithDefaultSettings);
 
-    //   CommonAssertions.assertAccessibility(categoryFacetComponent);
-    //   CommonAssertions.assertContainsComponentError(
-    //     CategoryFacetSelectors,
-    //     false
-    //   );
-    //   CommonFacetAssertions.assertDisplayFacet(CategoryFacetSelectors, true);
-    //   CommonFacetAssertions.assertDisplayPlaceholder(
-    //     CategoryFacetSelectors,
-    //     false
-    //   );
-    //   CategoryFacetAssertions.assertNumberOfChildValues(defaultNumberOfValues);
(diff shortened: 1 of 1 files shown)


packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
```diff
@@ -256,6 +256,8 @@ describe('Category Facet Test Suites', () => {
         function setupSelectFirstParent() {
           setupGoDeeperLastLevel();
           pressParentButton(0);
+          cy.wait(TestFixture.interceptAliases.Search);
+          cy.wait(TestFixture.interceptAliases.UA);
```
btaillon, line 260: If possible, I would instead assert something.

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -60,10 +60,9 @@ export function FocusTarget() {
         ) {
           focusAfterSearch = false;
           if (element) {
-            const el = element;
             // The focus seems to be flaky without deferring, especially on iOS.
             defer().then(() => {
-              el.focus();
```
btaillon, line 66: `element` is declared outside this scope, and could change during this `defer`. I recall that the reason I created this variable was because I once encountered an instance of that bug.

btaillon (changes requested):

packages/atomic/cypress/integration/facets/category-facet/category-facet.cypress.ts
liviarett: I think the assertion would be `CategoryFacetAssertions.assertFocusActiveParent()`, which is already done after this... I couldn't really put my finger on what the problem was here, but a lot of the times the element wouldn't be focused for this assertion.. I'd suggest that we try to add these two lines, but happy to close this PR if you think we shouldn't do it! let me know

packages/atomic/src/utils/accessibility-utils.tsx
btaillon: I guess that would mean the element is occasionally being added late. In this case, perhaps `(element ?? el).focus()` would be best.

ThibodeauJF:
FYI you can use `describe.skip` or `it.skip` to not run some tests

liviarett:
> FYI you can use `describe.skip` or `it.skip` to not run some tests

oh wow, I can't believe I committed this and didn't even notice!!
I think jest doesn't let you do it on `describe`, maybe? so I thought it was the same here and didn't even try. that will save me a lot of headache, thanks!!!

PR #2162, opened 2022-06-27
docs(atomic): move instant results parts to search box

liviarett:
https://coveord.atlassian.net/browse/KIT-1791

since the parts get rendered under `atomic-search-box`'s shadow, I think it makes sense to move them to that component's docs as someone might try to do `atomic-search-box-instant-results::part(instant-results-item)` and it won't work

### DIFF
--- a/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.pcss
+++ b/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.pcss
@@ -11,6 +11,10 @@
   @apply absolute w-5/6 h-5/6 bg-background rounded-full;
 }
 
+[part~='instant-results-item'] {
+  padding: 12px;
+}
+
 [part~='instant-results-show-all'] {
   border-top: 1px solid var(--atomic-neutral);
   justify-content: center;
--- a/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.pcss
+++ b/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.pcss
@@ -16,10 +16,9 @@
 }
 
 [part~='instant-results-show-all'] {
+  @apply py-1 justify-center;
   border-top: 1px solid var(--atomic-neutral);
-  justify-content: center;
   margin-top: auto;
-  padding: 2px 0;
   box-sizing: content-box;
   button {
     background-color: transparent;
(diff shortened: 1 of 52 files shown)


ThibodeauJF (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -59,6 +59,9 @@ import {Bindings} from '../atomic-search-interface/atomic-search-interface';
  * @part active-suggestion - The currently active suggestion.
  * @part suggestion-divider - An item in the list that separates groups of suggestions.
  * @part suggestion-with-query - An item in the list that will update the search box query.
+ * @part instant-results-item - An instant result rendered by `atomic-search-box-instant-results`.
```
jpmarceau, line 62:
```suggestion
 * @part instant-results-item - An instant result rendered by an `atomic-search-box-instant-results` component.
```

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -59,6 +59,9 @@ import {Bindings} from '../atomic-search-interface/atomic-search-interface';
...
+ * @part instant-result-show-all-button - The button to show all items for the current instant results search rendered by `atomic-search-box-instant-results`.
```
jpmarceau, line 63:
```suggestion
 * @part instant-result-show-all-button - The button to show all items for the current instant results search rendered by an `atomic-search-box-instant-results` component.
```

jpmarceau (approved):
👍

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.pcss
```diff
@@ -11,6 +11,10 @@
   @apply absolute w-5/6 h-5/6 bg-background rounded-full;
 }
 
+[part~='instant-results-item'] {
```
olamothe, line 14: Should this instead be using a value from Tailwind ? (ie: `p-6` or similar)

That way if someones wants to theme the "compactness" (I'm not sure how to phrase it better 😄) of their UI, it will also applies to padding for instant result.

olamothe (approved):
Does it mean that any new suggestion provider we add will also become a new `@part` on the search box ?

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.pcss
liviarett: yeah, that's a good shout!

liviarett:
> Does it mean that any new suggestion provider we add will also become a new `@part` on the search box ?

yeah, I'd think so! I believe it's better than misleadingly add it to the component itself but it not working if you tried to use it in that component!

PR #2166, opened 2022-06-27
fix(atomic): improved aria label for rating facet values

btaillon:
https://coveord.atlassian.net/browse/KIT-1793

For every facet value where "Only" is displayed, I added "only" in the aria label.

<table>
<tr>
	<td>

image
	<td>

Rating facet labels:
* Inclusion filter on only 5 stars out of 5; 205 results
* Inclusion filter on 4 stars out of 5; 1163 results
* Inclusion filter on 3 stars out of 5; 146 results
* Inclusion filter on 2 stars out of 5; 25 results
* Inclusion filter on one star out of 5; 16 results
<tr>
	<td>

image
	<td>

Rating range facet labels:
* Inclusion filter on only 5 stars out of 5; 205 results
* Inclusion filter on 4 to 5 stars; 1361 results
* Inclusion filter on 3 to 5 stars; 1503 results
* Inclusion filter on 2 to 5 stars; 1528 results
* Inclusion filter on 1 to 5 stars; 1544 results
</table>

### DIFF
--- a/packages/atomic/src/components/search/facets/atomic-rating-facet/atomic-rating-facet.tsx
+++ b/packages/atomic/src/components/search/facets/atomic-rating-facet/atomic-rating-facet.tsx
@@ -250,6 +250,9 @@ export class AtomicRatingFacet
   }
 
   private formatFacetValue(facetValue: NumericFacetValue) {
+    if (facetValue.start === this.maxValueInIndex) {
+      return this.bindings.i18n.t('stars-only', {count: facetValue.start});
+    }
     return this.bindings.i18n.t('stars', {
       count: facetValue.start,
       max: this.maxValueInIndex,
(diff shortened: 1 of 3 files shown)


ThibodeauJF:
I really don't see why we need the **only** but not the **& up**. IMO only is more implicit for 5 stars than & up is implicit to the 3 to 4, for example

npushkarskii (approved):
LGTM 👍

npushkarskii:
> * Inclusion filter on 2 to 5 stars

we don't need `& up` in aria-labels as we have more explicit text in aria-labels already, e.g., 
`Inclusion filter on 2 to 5 stars`.  You can also refer to my comment in a JIRA ticket where I address how different text and/or visuals clues help different categories of users 🙏

PR #2171, opened 2022-06-28
refactor(atomic): extract generic helper class for interface initialization

olamothe:
Continuation of the work needed for generic atomic interface

* Extract a `CommonAtomicInterfaceHelper` class that works as a sidekick for all the different interfaces. Handles generic stuff needed by all interface, such as coordinating `InitializeEvent`, `i18n + language`, `analytics`, etc.
* Rework some types and interface that had caused `any` to be added in `InitializationUtils` or `AccessibilityUtils` in my last PR.
 
https://coveord.atlassian.net/browse/KIT-1799


packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -330,16 +280,43 @@ export class AtomicSearchInterface
     this.engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
   }
 
-  private engineIsCreated(engine?: SearchEngine): engine is SearchEngine {
-    if (!engine) {
-      console.error(
-        'You have to call "initialize" on the atomic-search-interface component before modifying the props or calling other public methods.',
-        this.host
+  public get bindings(): Bindings {
```
olamothe, line 283: Some not strictly needed reordering of public methods in `atomic-search-interface`.

Pet peeves of mine, but I prefer when public methods are grouped at the top of the file, and private stuff at the bottom.

packages/atomic/src/components/common/interface/interface-common.tsx
```diff
@@ -2,18 +2,158 @@ import type {SearchEngine} from '@coveo/headless';
...
+  constructor(
+    private atomicInterface: BaseAtomicInterface<Engine>,
+    globalVariableName: string
+  ) {
+    setCoveoGlobal(globalVariableName);
+    loadFocusVisiblePolyfill();
+
+    const {
+      connectedCallback: originalConnectedCallback,
+      render: originalRender,
```
ThibodeauJF, line 51: I like the clarity: we should keep this naming scheme for other decorators too

ThibodeauJF (approved):
Great improvement!

packages/atomic/src/components/common/interface/bindings.ts
```diff
@@ -1,13 +1,34 @@
 import {HTMLStencilElement} from '@stencil/core/internal';
 import {i18n} from 'i18next';
 import {AnyEngineType} from './interface-common';
+import {AtomicCommonStoreData} from './store';
+
+export interface ExtensibleStencilStore<
```
liviarett, line 6: how about `CommonStencilStore`, or something with Base or Interface? 
just bc all interfaces are extensible, so it sounds a bit funny

packages/atomic/src/components/common/interface/bindings.ts
```diff
@@ -21,9 +42,15 @@ export interface CommonBindings<
...
    */
-  store: StoreCreation;
+  store: Store;
   /**
    * A reference to the atomic interface element.
    */
   interfaceElement: InterfaceElement;
 }
+
+export type AnyBindings = CommonBindings<
```
liviarett, line 52: typescript still blows my mind

packages/atomic/src/components/common/interface/interface-common.tsx
```diff
@@ -2,18 +2,158 @@ import type {SearchEngine} from '@coveo/headless';
 import {LogLevel} from '@coveo/headless';
 import type {RecommendationEngine} from '@coveo/headless/recommendation';
 import type {InsightEngine} from '@coveo/headless/insight';
-import {i18n} from 'i18next';
+import {i18n, TFunction} from 'i18next';
+import {ComponentInterface, h} from '@stencil/core';
+import {setCoveoGlobal} from '../../../global/environment';
```
liviarett, line 7: I've merged the PR that introduces the aliases, so feel free to update `global`, `utils` and `@components/[common|search|recs]`

packages/atomic/src/components/common/interface/interface-common.tsx
```diff
@@ -2,18 +2,158 @@ import type {SearchEngine} from '@coveo/headless';
...
+import {loadDayjsLocale} from '../../../utils/dayjs-locales';
+import {InitializeEvent} from '../../../utils/initialization-utils';
+import {AnyBindings} from './bindings';
+import {
+  i18nBackendOptions,
+  i18nTranslationNamespace,
+} from '../../common/interface/i18n';
+import Backend from 'i18next-http-backend';
 
 export type AnyEngineType = SearchEngine | RecommendationEngine | InsightEngine;
```
liviarett, line 20: how about moving this to the `bindings` file to avoid the both-ways importing?

packages/atomic/src/components/common/interface/interface-common.tsx
```diff
@@ -2,18 +2,158 @@ import type {SearchEngine} from '@coveo/headless';
...
+
+      return originalRender && originalRender.call(atomicInterface);
+    };
+  }
+
+  public onComponentInitializing(event: InitializeEvent) {
+    event.preventDefault();
+    event.stopPropagation();
+
+    if (this.atomicInterface.engine) {
```
liviarett, line 81: just to confirm, is this right? we're checking that the engine exists and resolving with the bindings?

packages/atomic/src/components/common/interface/bindings.ts
```diff
@@ -1,13 +1,34 @@
 import {HTMLStencilElement} from '@stencil/core/internal';
 import {i18n} from 'i18next';
 import {AnyEngineType} from './interface-common';
+import {AtomicCommonStoreData} from './store';
+
+export interface ExtensibleStencilStore<
+  StoreData extends AtomicCommonStoreData
```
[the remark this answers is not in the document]
olamothe: It would remove the need to explicitly set the type, yes.

However, it's a bit similar to having default argument values for function: They should generally be used when you want to convey to the users of the interface that they should probably not pass in a value for a given function argument the majority of the time in order to "use the function properly".

Here I feel it's better to always require the user of the interface to explicitly specify the type of data the store hold in order to "use the interface correctly", versus having the common one be the default.

So it's more in the spirit of instructing users how the interface should normally be used.

Does it make sense ?

packages/atomic/src/components/common/interface/bindings.ts
olamothe: Yeah, it's powerful.
The type system is actually a complete programming language in itself :D 

https://dev.to/gypsydave5/fizz-buzz-in-typescript-7ip

packages/atomic/src/components/common/interface/interface-common.tsx
olamothe: Yes, great idea !

packages/atomic/src/components/common/interface/interface-common.tsx
olamothe: Yeah. To be frank, I've just copied what was already done in the search interface in `handleInitialization`.

packages/atomic/src/utils/accessibility-utils.tsx
```diff
@@ -56,7 +55,7 @@ export function FocusTarget() {
         }
         if (
           focusAfterSearch &&
-          getSearchUIDFromState(this.bindings) !== lastSearchId
+          this.bindings.store.getUniqueIDFromEngine(this.bindings.engine)
```
liviarett, line 58: here goes grandma Livia again... not sure if my comment disappeared or got resolved or if I never sent it, but better safe than sorry, so here it goes again:
are you missing `!== lastSearchId` here?
and: (for a future PR) would it make sense to save the engine in the store as well to make things simpler to access?

olamothe: Yep, fixed !

PR #2173, opened 2022-06-29
fix(quantic): corner case addressed in the QuanticTabBar component

mmitiche:
https://coveord.atlassian.net/browse/SFINT-4546
When a tab containing the same expression as another tab is selected from the drop-down list:
In the following the Tab2 has the same expression as Tab4

#### Before:

https://user-images.githubusercontent.com/86681870/176433899-6a06798b-2550-4830-98bb-ea693a1a1b7f.mov

#### After:

https://user-images.githubusercontent.com/86681870/176433923-c3767569-c8de-477f-b3d4-2d7717b14fc4.mov


nathanlb (approved):
LGTM

PR #2175, opened 2022-06-29
feat(atomic): hide arrow buttons

mbenkirane-coveo:
Demo:

https://user-images.githubusercontent.com/97299176/177392628-8f60c014-ec11-4836-9caa-ea73764c7863.mov


packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -0,0 +1,189 @@
+import {Component, h, Listen, State} from '@stencil/core';
+import ArrowRightIcon from 'coveo-styleguide/resources/icons/svg/arrow-right-rounded.svg';
+import ArrowLeftIcon from 'coveo-styleguide/resources/icons/svg/arrow-left-rounded.svg';
+import {Button} from '../../common/button';
+import {
+  BindStateToController,
+  InitializableComponent,
+  InitializeBindings,
+} from '../../../utils/initialization-utils';
```
liviarett, line 9: I've merged the aliases now, so you can use
`@components/common....`
`@utils/.....`

package.json
```diff
@@ -80,5 +80,8 @@
   },
   "engines": {
     "node": ">=12.0.0"
+  },
+  "dependencies": {
```
liviarett, line 84: shouldn't this be in the atomic package.json?

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -0,0 +1,189 @@
...
+  @BindStateToController('searchStatus')
+  @State()
+  public searchStatusState!: SearchStatusState;
+  @State()
+  public error!: Error;
+
+  private horizontalScroll?: HTMLDivElement;
+  @State() private hideLeftArrow = true;
+  @State() private hideRightArrow = false;
+  private ro!: ResizeObserver;
```
liviarett, line 49: I'd call it `observer` to be clearer

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    const isOverflowing =
+      this.horizontalScroll.clientWidth < this.horizontalScroll.scrollWidth;
+    const isLeftEdge = this.horizontalScroll?.scrollLeft <= 0;
+    const isRightEdge =
+      this.horizontalScroll.scrollLeft >=
+      this.horizontalScroll.scrollWidth - this.horizontalScroll.clientWidth;
+
+    if (!isOverflowing) {
```
liviarett, line 80: for this block, how about 

```js
this.hideLeftArrow = !isOverflowing || isLeftEdge
this.hideRightArrow = !isOverflowing || isRightEdge
```

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -0,0 +1,189 @@
...
+    this.ro.disconnect();
+  }
+
+  @Listen('mousewheel')
+  handleScroll() {
+    if (!this.horizontalScroll) {
+      return;
+    }
+
+    const isOverflowing =
```
liviarett, line 73: a suggestion for the name: `isScrollable`

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -0,0 +1,189 @@
...
+    }
+  }
+
+  private slideHorizontally(direction: ArrowDirection) {
+    const container = this.horizontalScroll;
+    if (!container) {
+      return;
+    }
+
+    const containerWidth = container.clientWidth;
```
liviarett, line 102: if there is ever a case where container in undefined, like you expect on line 104, this will break but in general, I think you should choose one approach - to save stuff as variables or not, for instance on line 104 you use `container.clientWidth` and the on line 117 you use `containerWidth`. I'd suggest sticking to §`container.clientWidth`, unless you then rename things to `width`, `scrollWidth`, `scrollLeft` (they are a bit shorter and therefore easier to read).

Of course my comment above doesn't apply if you were trying to save the original value before changing it (which would explain the mix of `containerWidth` and `container.clientWidth`) but then I'd rename the variables to 'original.....`

liviarett:
I left a few comments, but it looks really good! there are two points I would suggest:
1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather
2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
mbenkirane-coveo: Nice, I like this !

package.json
mbenkirane-coveo: Yup !

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
mbenkirane-coveo: Sure !

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
mbenkirane-coveo: Sure !

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -0,0 +1,189 @@
...
+    if (direction === 'left') {
+      container.scrollLeft -= pixelsToScroll;
+      if (container.scrollLeft - pixelsToScroll <= 0) {
+        this.hideLeftArrow = true;
+      }
+    } else {
+      container.scrollLeft += pixelsToScroll;
+      if (
+        container.scrollLeft + pixelsToScroll >=
+        containerScrollWidth - containerWidth
```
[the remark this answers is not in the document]
mbenkirane-coveo: > I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears



> I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears

Regarding your first point, I'll make the gradient a bit softer and add feather. For your second point, I'll add an animation when hiding the arrow button which should make it a bit smoother.

mbenkirane-coveo:
> I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears



> I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears



> I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears



> I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears



> I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears



> I left a few comments, but it looks really good! there are two points I would suggest: 1 - the gradient seems a bit too small compared to the designs, I'd go a bit bigger and with a bit more feather 2 - if feels a little bit weird to me that the button disappears right away before the scrolling animation finishes... I'd like to see if it makes more sense to wait for the scrolling to actually get to the edge (or almost) before the button disappears

Regarding your first point, I'll make the gradient a bit softer and add feather. For your second point, I'll add an animation when hiding the arrow button which should make it a bit smoother.

packages/atomic/src/components/search/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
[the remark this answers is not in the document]
mbenkirane-coveo: A bit too overkill imo as well. It's doesn't really reduce the lines of code or really make it more legible.

packages/atomic/package.json
```diff
@@ -67,7 +67,8 @@
     "i18next": "20.6.1",
     "i18next-http-backend": "1.4.1",
     "patch-package": "6.4.7",
-    "stencil-inline-svg": "1.1.0"
+    "stencil-inline-svg": "1.1.0",
+    "resize-observer-polyfill": "^1.5.1"
```
ThibodeauJF, line 71: Doesn't seem to be used

packages/atomic/src/components/facets/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -21,32 +33,107 @@ type ArrowDirection = 'right' | 'left';
...
+      this.handleScroll();
+    });
+    this.observer.observe(this.horizontalScroll as HTMLDivElement);
+  }
+
+  disconnectedCallback() {
+    this.observer.disconnect();
+  }
+
+  @Listen('mousewheel')
```
ThibodeauJF, line 66: Why listen to the mouse wheel event? I don't think this will work on mobile
You can try it out with the dev tools
Screen Shot 2022-07-04 at 11 51 56 AM

ThibodeauJF: Perhaps the `touchmove` event.

mbenkirane-coveo: ok, we actually need both !

lvu285:
Can you check if 
- the arrow should be completed hidden if facet returns no value
- when a facet value is selected, the "facet count' should be blue as well
- Add a space between facet & Sort filter box, right now they are too closed to each other
- When we have Segmented facet & Normal facet with the same field (2 same facets), we will consider as 2 different facets or 1 facet filter ? _ At this moment, we consider them as 2 different facet, 1 is `facetID = A`, the other one is `facetID = A_1` but CLEAR button (breadcrumb) applies for both facets
-

mbenkirane-coveo:
> 

Regarding the last point, you're saying clear should only clear the filters perform by the regular facets ?

packages/atomic/src/components/facets/atomic-segmented-facet-scrollable/atomic-segmented-facet-scrollable.tsx
```diff
@@ -21,32 +33,108 @@ type ArrowDirection = 'right' | 'left';
...
-    const pixelsToScroll = container ? container.clientWidth * 0.75 : 700;
-
-    if (container === null || !container) {
+    if (!container) {
       return;
     }
+
+    const width = container.clientWidth;
+    const scrollWidth = container.scrollWidth;
+    const scrollLeft = container.scrollLeft;
```
ThibodeauJF, line 92: You could get rid of these variables (recommended since they contain no calculations), or use `width` at line 93 for consistency

PR #2176, opened 2022-06-29
fix(headless): properly support authentication query parameter to use with SAML

olamothe:
When using SAML authentication, we need to send the `authentication` query parameter to the API.

This parameter needs to contain the name(s) of the security identity provider to use to resolve the identity of the end user. 

Without this query parameter, it means the user is essentially anonymous/not authenticated as far as the search API is concerned.

I've also added a new `buildSearchEngineWithSamlAuthentication` function to Headless, which takes care of wiring the `accessToken`, `provider`, and `renewAccessToken` function automatically for the implementers.
 

https://coveord.atlassian.net/browse/KIT-1800

### DIFF
--- a/packages/headless/src/api/search/search-api-params.ts
+++ b/packages/headless/src/api/search/search-api-params.ts
@@ -114,17 +114,28 @@ export interface ExcerptLength {
   excerptLength?: number;
 }
 
+export interface AuthenticationParam {
+  authentication?: string;
+}
+
 export const baseSearchRequest = (
-  req: BaseParam,
+  req: BaseParam & AuthenticationParam,
   method: HttpMethods,
   contentType: HTTPContentType,
   path: string
-) => ({
-  accessToken: req.accessToken,
-  method,
-  contentType,
-  url: `${req.url}${path}?${getOrganizationIdQueryParam(req)}`,
-});
+) => {
+  return {
+    accessToken: req.accessToken,
+    method,
+    contentType,
+    url: `${req.url}${path}?${getOrganizationIdQueryParam(req)}${
+      req.authentication ? `&${getAuthenticationQueryParam(req)}` : ''
+    }`,
+  };
+};
 
 export const getOrganizationIdQueryParam = (req: BaseParam) =>
   `organizationId=${req.organizationId}`;
+
+export const getAuthenticationQueryParam = (req: AuthenticationParam) =>
+  `authentication=${encodeURIComponent(req.authentication!)}`;
--- a/packages/headless/src/app/search-engine/search-engine-configuration.ts
+++ b/packages/headless/src/app/search-engine/search-engine-configuration.ts
@@ -1,10 +1,13 @@
-import {RecordValue, Schema} from '@coveo/bueno';
+import {ArrayValue, RecordValue, Schema} from '@coveo/bueno';
 import {
   PostprocessFacetSearchResponseMiddleware,
   PostprocessQuerySuggestResponseMiddleware,
   PostprocessSearchResponseMiddleware,
 } from '../../api/search/search-api-client-middleware';
-import {nonEmptyString} from '../../utils/validate-payload';
+import {
+  nonEmptyString,
+  requiredNonEmptyString,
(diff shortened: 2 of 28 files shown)


packages/headless/src/api/search/search-api-params.ts
```diff
@@ -114,17 +114,28 @@ export interface ExcerptLength {
...
   contentType: HTTPContentType,
   path: string
-) => ({
-  accessToken: req.accessToken,
-  method,
-  contentType,
-  url: `${req.url}${path}?${getOrganizationIdQueryParam(req)}`,
-});
+) => {
+  return {
```
btaillon, line 127: Did you mean to add an explicit return?

btaillon (approved):

packages/headless/src/app/search-engine/search-engine-configuration.ts
```diff
@@ -1,10 +1,13 @@
...
 import {
   PostprocessFacetSearchResponseMiddleware,
   PostprocessQuerySuggestResponseMiddleware,
   PostprocessSearchResponseMiddleware,
 } from '../../api/search/search-api-client-middleware';
-import {nonEmptyString} from '../../utils/validate-payload';
+import {
+  nonEmptyString,
+  requiredNonEmptyString,
+} from '../../utils/validate-payload';
```
liviarett, line 10: same as the other PR, `@utils` now available at shops near you!

olamothe: That's Headless though, it's not configured for this project, is it ? 🤔

packages/headless/src/api/search/search-api-params.ts
```diff
@@ -114,17 +114,28 @@ export interface ExcerptLength {
   excerptLength?: number;
 }
 
+export interface AuthenticationParam {
+  authentication?: string;
```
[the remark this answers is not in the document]
olamothe: The reason is that the actual parameter expected by the API is a string, joined with commas for each provider. 

Since in this file were trying to represent as closely as possible what the API expect, I went with the approach using the "exact type".

packages/headless/src/features/configuration/configuration-slice.test.ts
```diff
@@ -235,6 +238,7 @@ describe('configuration slice', () => {
           apiBaseUrl: 'http://test.com/search',
           locale: 'fr-CA',
           timezone: 'Africa/Johannesburg',
+          authenticationProviders: ['theNewProvider'],
```
mbenkirane-coveo, line 241: What's the difference between theProvider and theNewProvider

packages/headless/src/app/search-engine/search-engine.ts
```diff
@@ -141,6 +143,39 @@ export function buildSearchEngine(options: SearchEngineOptions): SearchEngine {
   };
 }
 
+export async function buildSearchEngineWithSamlAuthentication(
```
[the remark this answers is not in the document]
ThibodeauJF: +1 https://coveo.slack.com/archives/G7HD9P48H/p1656446003618809?thread_ts=1656445662.481239&cid=G7HD9P48H

packages/headless/src/app/search-engine/search-engine-configuration.ts
liviarett: ah no, sorry! didn't think properly.. could do though

packages/headless/src/features/configuration/configuration-slice.test.ts
olamothe: That specific test is about verifying the work the reducer does when changing from an existing state to another.

So going from `theProvider` to `theNewProvider`.

packages/headless/src/app/search-engine/search-engine-configuration.ts
olamothe: You mean adding aliases to Headless ? Yeah it would be possible. But I think it's out of scope for this PR.

PR #2180, opened 2022-07-04
feat: npm workspace migration

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1811 (more details in Epic)

PR build works:
Screen Shot 2022-07-05 at 9 48 06 AM


.github/actions/e2e-atomic-angular/action.yml
```diff
@@ -8,13 +8,17 @@ inputs:
...
+      shell: bash
+      working-directory: ./packages/samples/angular
     - uses: cypress-io/github-action@v4
       name: Run Cypress
       with:
         working-directory: ./packages/samples/angular
         start: npm run dev
         wait-on: 'http://localhost:4200'
-        install-command: npm install
+        install: false
```
ThibodeauJF, line 21: cypress expects a lockfile besides the package.json, so I move the step before

tsconfig.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "./node_modules/gts/tsconfig-google.json"
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "skipLibCheck": true,
```
ThibodeauJF, line 4: was in most packages so I move this config at the top and removed it elsewhere

packages/atomic/package.json
```diff
@@ -86,24 +84,22 @@
     "@storybook/addon-essentials": "6.3.12",
     "@storybook/addon-viewport": "6.3.12",
     "@storybook/api": "6.3.12",
+    "@storybook/builder-webpack5": "^6.3.12",
```
olamothe, line 87: Is this needed to unify react version ?

If so, have you been able to `npm run storybook` and verify that it works okay ?

The storybook integration is quite custom !

packages/atomic/src/components/search/atomic-search-interface/analytics-config.ts
```diff
@@ -2,7 +2,7 @@ import {
   AnalyticsConfiguration,
   SearchEngineConfiguration,
 } from '@coveo/headless';
-import {AnalyticsClientSendEventHook} from '@coveo/headless/node_modules/coveo.analytics';
+import {AnalyticsClientSendEventHook} from '../../../../../../node_modules/coveo.analytics';
```
olamothe, line 5: just `import {AnalyticsClientSendEventHook} from 'coveo.analytics';` should work ? 🤔

olamothe (approved):
Awesome job ! 

It's grunt work, but hopefully it will solve many build problems we've had past few weeks. Lerna is a pain...

packages/atomic/src/components/search/atomic-search-interface/analytics-config.ts
ThibodeauJF: "coveo.analytics" is extraneous, it's not a direct dependency, but through headless

packages/atomic/package.json
ThibodeauJF: To unify the webpack version. Storybook still builds and runs fine 😄

packages/atomic/src/components/search/atomic-search-interface/analytics-config.ts
louis-bompart: > "coveo.analytics" is extraneous,

I think either headless should export AnalyticsClientSendEventHook, or you should install coveo.analytics in this workspace.

olamothe: Yeah should be a direct dependency.

We can `import type` to be explicit that it's just for types, not code.

OR : Re-export the type from headless (since it is a dependency from Headless)
That second option would make sense, since it's actually a type that you need to know about if you implement on top of Headless (as showcased here)

package.json
```diff
@@ -20,39 +20,50 @@
...
     "prettier": "2.6.2",
     "rimraf": "3.0.2",
     "semver": "7.3.7",
+    "ts-jest": "27.1.4",
     "typescript": "4.6.4"
   },
+  "workspaces": [
+    "packages/*",
+    "packages/atomic-angular/*",
+    "packages/atomic-angular/projects/*",
```
louis-bompart, line 64: FYI, nested workspaces are not supported by NPM
https://github.com/npm/cli/issues/4518

louis-bompart (commented):
LGTM, I would be very careful with nested packages. If it works currently nice, but draft a plan of action to change that soon: given that this hierarchy ain't supported by npm, this is likely to cause pain in the future

lerna.json
```diff
@@ -1,9 +1,9 @@
 {
   "packages": [
     "packages/*",
-    "packages/samples/**",
     "packages/atomic-angular/*",
-    "packages/atomic-angular/projects/*"
+    "packages/atomic-angular/projects/*",
```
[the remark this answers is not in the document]
ThibodeauJF: Bumping versions

package.json
olamothe: In this case, we can only put `"packages/atomic-angular/projects/*",` methink, don't need ` "packages/atomic-angular/*",`

packages/samples/atomic-next/package.json
```diff
@@ -3,25 +3,23 @@
...
+    "@types/react-dom": "18.0.5",
     "ncp": "2.0.0",
     "typescript": "4.6.4"
   },
   "scripts": {
-    "build": "next build",
-    "dev": "npm run copy:assets && npm run link:atomic-react && next dev",
-    "link:atomic-react": "cd ../../atomic-react && npm link ../samples/atomic-next/node_modules/react/ && ncp ../atomic/dist ../atomic-react/node_modules/@coveo/atomic/dist",
-    "copy:assets": "ncp node_modules/@coveo/atomic-react/dist/assets public/assets && ncp node_modules/@coveo/atomic-react/dist/lang public/lang public/assets"
+    "build": "tsc --noEmit && tsc --module commonjs --noEmit",
```
liviarett, line 20: out of curiosity, why did you change this?

liviarett (approved):

louis-bompart (approved):
project tree structure seems sounds, lgtm 👍

packages/samples/atomic-next/package.json
ThibodeauJF: I'll improve this one after (had issues with the react version here too), for now I just wanted to make sure it compiles at least, like the atomic-react sample

PR #2183, opened 2022-07-04
feat(atomic): initial work on atomic-insight-interface 

liviarett:
https://coveord.atlassian.net/browse/KIT-1812

https://user-images.githubusercontent.com/30511433/177199156-ce705f52-27f2-440b-ae28-07bf9b680164.mp4

I added a toggle to the sample page so we can easily see how it looks as full page vs as a widget.
If unnecessary, I can remove!

### DIFF
--- a/packages/atomic/src/pages/examples/insights.html
+++ b/packages/atomic/src/pages/examples/insights.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html dir="ltr" lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />
+    <title>Coveo Atomic: Insights Panel example</title>
+
+    <script type="module" src="/build/atomic.esm.js"></script>
+    <script nomodule src="/build/atomic.js"></script>
+    <link rel="stylesheet" href="/themes/coveo.css" />
+    <script type="module">
+      async function main() {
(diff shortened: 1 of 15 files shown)


packages/atomic/src/pages/examples/insights.html
```diff
@@ -0,0 +1,62 @@
...
+
+    <script type="module" src="/build/atomic.esm.js"></script>
+    <script nomodule src="/build/atomic.js"></script>
+    <link rel="stylesheet" href="/themes/coveo.css" />
+    <script type="module">
+      async function main() {
+        await customElements.whenDefined('atomic-svc-insight-interface');
+        const insightInterface = document.querySelector('atomic-svc-insight-interface');
+        await insightInterface.initialize({
+          accessToken: 'xxbb74d0cd-9374-45b9-ae5c-f106e2cc0795',
```
olamothe, line 16: Is this okay to put public ? For `searchuisamples` it's okay since we know that it's all public content, and it's well understood by everyone that adds content in there.

If not, the key needs to be deleted ASAP. Because by now all manner of bots across the internet have stored it ;) 

Also, we'll need something more solid long term, that's available in a production org, as opposed to dev.

It can stay like that for now (as long as the key can be made public).

olamothe (approved):

packages/atomic/src/pages/examples/insights.html
liviarett: I'm not sure, I'll ask Luc later but delete for now!

ThibodeauJF (approved):

PR #2188, opened 2022-07-05
feat(atomic,headless): add search box to insights

liviarett:
https://user-images.githubusercontent.com/30511433/177362508-4a0a8a26-7032-4cff-ae81-1ae728733d5e.mp4


packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
```diff
@@ -0,0 +1,86 @@
+import {Component, h, Prop, State} from '@stencil/core';
+import {buildInsightSearchBox} from '@coveo/headless/insight';
+import {SearchBox, SearchBoxState} from '@coveo/headless';
```
olamothe, line 3: These types should also be coming from `'@coveo/headless/insight'` I think 🤔 

If they are missing it means it's a bug from Headless package.

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
```diff
@@ -0,0 +1,86 @@
...
+import {SearchInput} from '../../common/search-box/SearchInput';
+import {SearchBoxWrapper} from '../../common/search-box/SearchBoxWrapper';
+import {SubmitButton} from '../../common/search-box/SubmitButton';
+
+/**
+ *
+ * @internal
+ */
+@Component({
+  tag: 'atomic-insight-search-box',
```
olamothe, line 19: What's the naming scheme we want to adhere to ?

* `atomic-insight-interface`, `atomic-insight-search-box`, etc.

* `atomic-svc-insight-interface`, `atomic-svc-insight-search-box` etc.

* `atomic-svc-interface`, `atomic-svc-search-box`, etc.

* Other ?

olamothe (approved):
Nice !

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
liviarett: yeah I couldn't find it there.. I did add the missing import for `buildInsightSearchBox` and now see that there is an `InsightSearchBox` too that I can export! 
will tag the headless guys to have a look too!!

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
liviarett: I think that, following headless, `atomic-insight-search-box` and `atomic-insight-interface` make sense and is easier to type compared to needing the extra `svc` there

packages/headless/src/insight.index.ts
```diff
@@ -33,3 +33,8 @@ export {buildController} from './controllers/controller/headless-controller';
 export type {InsightInterfaceState} from './features/insight-interface/insight-interface-state';
 export type {InsightInterface} from './controllers/insight-interface/insight-interface';
 export {buildInsightInterface} from './controllers/insight-interface/insight-interface';
+export type {
```
liviarett, line 36: @lbergeron @aelfaleh hey guys, just tagging you here to confirm that this is the right thing to do

liviarett:
> Nice ! So the API call that you show in your small screen recording, are they hitting `/search/insight/${insight}` in dev ?
> 
> Or `/search` directly ?

this is the URL `https://platformdev.cloud.coveo.com/rest/organizations/orgId_xxxxxx/insight/v1/configs/insightId_xxxxxxxx/search`

packages/headless/src/insight.index.ts
nathanlb: So the plan on our side is to export the insight controllers under the same name as their search page equivalent. The aim is for the Quantic UI components to reference the same build function using the bundle that has been loaded. This is to maximise code reuse and avoid having to create another set of quantic components.
What are your thoughts on that?

nathanlb: So it's confirmed that we'll be exporting all the insight controllers under the same names as their search counterparts. The plan is to rename them to the correct name in their respective controllers/ files. That process is currently underway in a few PRs.

liviarett: no problem, we can have a PR that changes all of this later!

packages/atomic/src/components/common/search-box/ClearButton.tsx
```diff
@@ -0,0 +1,38 @@
+import ClearIcon from 'coveo-styleguide/resources/icons/svg/clear.svg';
```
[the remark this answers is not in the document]
liviarett: yeah, definitely!

liviarett:
@olamothe 
I renamed `InitializationOptions` and `Bindings` to `InsightInitializationOptions` and `InsightBindings` and `SearchInitializationOptions` and `SearchBindings` as I noticed that the components file was appending a `1` to the duplicated name but not importing it correctly in the component
I noticed it when renaming from `atomic-svc-insight-interface` to `atomic-insight-interface`, as it went up the alphabetical list (i < s) which meant `atomic-search-interface` was now importing the insight `InitializationOptions` and requiring an `insightId` together with `accessToken` and `organizationId`
I've split into different commits to make review easier. Please let me know what you think :)

PR #2194, opened 2022-07-06
feat(atomic,headless):  clear only affects regular facets

mbenkirane-coveo:
https://coveord.atlassian.net/browse/KIT-1819


ThibodeauJF:
Good start! Two things I can think of:

1- This option should not be part of the API request nor be sent to the API, you can inspire yourself from https://github.com/coveo/ui-kit/pull/2115/files to omit it, and use the devtools to see what is being sent
2- Breadcrumb values that are returned from Headless should filter out facets with hasBreadcrumbs=false

PR #2195, opened 2022-07-06
fix(atomic,headless): clear standalone search box state when redirecting

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1821

I don't think clearing the search box **query** it's relevant to the SPA use case since you probably keep the same atomic component & the same controller (original reason why **redirectTo** was not reset in the first place).

### DIFF
--- a/packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
+++ b/packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
@@ -13,6 +13,7 @@ import {
   registerStandaloneSearchBox,
   updateAnalyticsToOmniboxFromLink,
   updateAnalyticsToSearchFromLink,
+  resetStandaloneSearchBox,
 } from '../../features/standalone-search-box-set/standalone-search-box-set-actions';
 import {StandaloneSearchBoxAnalytics} from '../../features/standalone-search-box-set/standalone-search-box-set-state';
 import {
@@ -50,6 +51,11 @@ export interface StandaloneSearchBox extends SearchBox {
    * Triggers a redirection.
    */
   submit(): void;
+
+  /**
+   * Reset a standalone search box's state. To be dispatched on SPAs after the redirection has been triggered.
+   */
+  reset(): void;
 
   /**
    * A scoped and simplified part of the headless state that is relevant to the `StandaloneSearchBox` controller.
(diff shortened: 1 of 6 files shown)


mbenkirane-coveo (approved):

packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
```diff
@@ -51,6 +52,11 @@ export interface StandaloneSearchBox extends SearchBox {
    */
   submit(): void;
 
+  /**
+   * Reset a standalone search box's state. To be dispatched on SPAs after the redirection has been triggered.
+   */
+  reset(): void;
```
olamothe, line 58: Suggestion: `afterRedirection` or something similar ? 🤔

olamothe (approved):
One thing that's currently missing in our various test coverage is a setup with a "fake" single page app website...

ie: Homepage, search page, some other page with a recommendation component (when it's going to exists in Atomic), etc.

packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
```diff
@@ -51,6 +52,11 @@ export interface StandaloneSearchBox extends SearchBox {
    */
   submit(): void;
 
+  /**
+   * Reset a standalone search box's state. To be dispatched on SPAs after the redirection has been triggered.
```
jpmarceau, line 56:
```suggestion
   * Resets the standalone search box state. To be dispatched on SPAs after the redirection has been triggered.
```
Curious: what's `SPA` here?

packages/headless/src/features/standalone-search-box-set/standalone-search-box-set-actions-loader.ts
```diff
@@ -35,6 +38,16 @@ export interface StandaloneSearchBoxSetActionCreators {
     payload: RegisterStandaloneSearchBoxActionCreatorPayload
   ): PayloadAction<RegisterStandaloneSearchBoxActionCreatorPayload>;
 
+  /**
+   * Reset a standalone search box's state. To be dispatched on SPAs after the redirection has been triggered.
```
jpmarceau, line 42:
```suggestion
   * Resets the standalone search box state. To be dispatched on SPAs after the redirection has been triggered.
```

jpmarceau (approved):

packages/headless/src/controllers/standalone-search-box/headless-standalone-search-box.ts
ThibodeauJF: single page application, i'll type it out in the docs too

jpmarceau: Thanks, it'll help some users I think 🙂

PR #2199, opened 2022-07-07
docs(atomic): storybook for searchbox suggestions providers

olamothe:
Stories for: 
* atomic-query-suggestions
* atomic-recent-queries
* atomic-instant-results
https://coveord.atlassian.net/browse/KIT-1830


ThibodeauJF (approved):
The custom query suggestion would be super helpful too!

PR #2207, opened 2022-07-11
feat(atomic-react): add support for result list table

olamothe:
https://coveord.atlassian.net/browse/KIT-1842

### DIFF
--- a/packages/samples/atomic-react/src/App.tsx
+++ b/packages/samples/atomic-react/src/App.tsx
@@ -6,11 +6,13 @@ import {FoldedResultListPage} from './pages/FoldedResultListPage';
 import {HeaderLink} from './components/HeaderLink';
 import {FoldedResultListWithCustomChildrenPage} from './pages/FoldedResultListWithCustomChildrenPage';
 import {InstantResultsPage} from './pages/InstantResultsPage';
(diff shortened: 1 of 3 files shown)


packages/samples/atomic-react/src/App.tsx
```diff
@@ -43,6 +50,7 @@ const App: FunctionComponent = () => {
         <FoldedResultListWithCustomChildrenPage />
       )}
       {page === INSTANT_RESULTS_PAGE && <InstantResultsPage />}
+      {page === TABLE_RESULTS_LIST_PAGE && <TableResultListPage />}
```
ThibodeauJF, line 53: It's "result list" and "instant results" but not "results list", that would become confusing fast 😄 

(true for other places in the code as well)

packages/samples/atomic-react/src/pages/TableResultListPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+        imageSize="large"
+        template={MyTemplate}
+      />
+    </AtomicPageWrapper>
+  );
+};
+
+function MyTemplate() {
+  return (
+    <>
```
btaillon, line 39: Is it possible to add a clickable button in the template?

ThibodeauJF (approved):

PR #2208, opened 2022-07-11
feat(atomic): first atomic segmented filter test

mbenkirane-coveo:
- Described a couple tests that I will write in a future PR without implementing them.


packages/atomic/cypress/integration/facets/segmented-facet/segmented-facet-assertions
```diff
@@ -0,0 +1,23 @@
+import {SegmentedFacetSelectors} from './segmented-facet-selectors';
```
ThibodeauJF, line 1: file has no extension

packages/atomic/cypress/integration/facets/segmented-facet/segmented-facet.cypress.ts
```diff
@@ -0,0 +1,50 @@
...
+      false
+    );
+  });
+
+  describe('when select a value', () => {
+    describe('verify rendering', () => {});
+    describe('verify analytic', () => {});
+  });
+
+  describe('with custom #numberOfValues', () => {});
```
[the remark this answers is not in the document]
ThibodeauJF: https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Excluding-and-Including-Tests

packages/atomic/src/components/search/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -145,7 +145,11 @@ export class AtomicSegmentedFacet
     if (!this.label) {
       return;
     }
-    return <b class="inline-block my-3 mx-2">{this.label}:</b>;
+    return (
+      <b class="inline-block my-3 mx-2" part="label">
```
ThibodeauJF, line 149: Btw all parts have to be documented at the top of the components (all parts from the functional components should be documented there as well)

mbenkirane-coveo: Yes !

ThibodeauJF: value-box, value-label, value-count, etc. as well

ThibodeauJF (approved):

packages/atomic/cypress/integration/facets/segmented-facet/segmented-facet-selectors.ts
```diff
@@ -0,0 +1,78 @@
+export const segmentedFacetComponent = 'atomic-segmented-facet';
+export const SegmentedFacetSelectors = {
+  withId(id: string) {
```
btaillon, line 3: This style of selector vs the style of selector that other non-facet components use exists mostly to test facet dependencies. However, I think it makes sense for you to use it here like you just did, since we may want to support that eventually.

btaillon (approved):

PR #2217, opened 2022-07-13
chore(quantic): fixed flaky facet search tests

nathanlb:
A couple facet tests were failing as they expected 'account' facet search to yield only one result.
Instead of relying on a changing index I decided to intercept the facetSearch request and mock the response.

To mock the facet search request response to have only one value:
- Added mock function
- Added mock test setup function
- Moved the concerned tests into their on `it` blocks as, unfortunately cypress doesn't allow you to restore intercepts and overwriting them is messy and doesn't work great either.


aelfaleh (approved):
LGTM

lbergeron (approved):
Looks good to me!

ThibodeauJF (approved):
Thanks!

PR #2222, opened 2022-07-15
fix(quantic): read sfdx error JSON output from stdout instead of stderr

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2349

Apparently a recent version of `sfdx-cli` introduced a change in the way errors are returned. Before, the error JSON response would be returned through `stderr`. It is now returned through `stdout`.

It prevented our `deploy-community` script from deploying on an existing scratch org.


nathanlb:
Is this change compatible with the version of sfdx we're installing in our CI workflows too?

aelfaleh (approved):

erocheleau (approved):
As long as we are sure the output will always be json parsable?

lbergeron:
> As long as we are sure the output will always be json parsable?

You have a good point. So far I've always seen the SFDX CLI return JSON when using the `--json` flag. However, if an issue occur trying to execute the SFDX CLI, then the output won't be JSON. I'll improve that.

lbergeron:
It seems that upgrading the `sfdx-cli` dependency corrupted the package-lock somehow. Opening a [new PR](https://github.com/coveo/ui-kit/pull/2235) that simply fixes the issue and keep dependencies the same.

lbergeron:
It seems that upgrading the `sfdx-cli` dependency corrupted the package-lock somehow. Opening a [new PR](https://github.com/coveo/ui-kit/pull/2235) that simply fixes the issue and keep dependencies the same.

PR #2225, opened 2022-07-19
feat(atomic): scrollable tests

mbenkirane-coveo:
Wrote some basic tests for the scrollable component.
Currently facing an issue with testing the scrolling since the 'wheel' event won't trigger.
Also facing an issue with the keyboard scrolling, since the cypress type function won't work.

To fix: 

- [ ] Fire 'wheel' event in cypress' test environment when scrolling

- [ ] Press right keyboard arrow on cypress 'with right scroll using keyboard' test

*Scroll and Keyboard tests will be skipped until a way to trigger events and to press keyboard buttons on cypress is found.


packages/atomic/cypress/integration/segmented-facet-scrollable-actions.ts
```diff
@@ -0,0 +1,42 @@
...
+  direction === 'left'
+    ? ScrollableSelectors.leftArrow().click()
+    : ScrollableSelectors.rightArrow().click();
+}
+
+export function scroll(direction: direction, msDuration: number) {
+  ScrollableSelectors.horizontalScroll().scrollTo(direction, {
+    duration: msDuration,
+  });
+  cy.get(scrollableComponent).trigger('wheel');
```
mbenkirane-coveo, line 41: Does not trigger the event.

packages/atomic/cypress/integration/segmented-facet-scrollable.cypress.ts
```diff
@@ -0,0 +1,117 @@
...
+      ScrollableAssertions.assertDisplayArrows(true, true);
+    });
+
+    describe('with right scroll using keyboard', () => {
+      function setupKeyboardScrollable() {
+        setupWithOverflowFacets();
+        cy.get(scrollableComponent).click('center');
+        cy.get(segmentedFacetComponent)
+          .shadow()
+          .find('[part="value-box"][aria-pressed="true"]')
```
mbenkirane-coveo, line 58: Unable to press the right arrow key on a value-box

packages/atomic/cypress/fixtures/test-fixture.ts
```diff
@@ -356,6 +356,10 @@ export const addTag = (env: TestFixture, tag: string, props: TagProps) => {
   env.withElement(e);
 };
 
+export const addElement = (env: TestFixture, element: HTMLElement) => {
+  env.withElement(element);
+};
```
btaillon, line 361: I prefer not to add this util. Instead, you could use `addTag` or create case-specific methods like `addSmartSnippet`.

packages/atomic/cypress/integration/segmented-facet-scrollable-actions.ts
```diff
@@ -0,0 +1,42 @@
...
+  );
+}
+
+export const addScrollable =
+  (facetProps: TagProps = {}) =>
+  (env: TestFixture) => {
+    scrollable = generateComponentHTML('atomic-segmented-facet-scrollable');
+    appendSegmentedFacet(facetProps);
+    addElement(env, scrollable);
+  };
```
btaillon, line 29: Instead of appending segmented facets, you could take children as a prop (similar to `addResultList`) and `appendSegmentedFacet` could instead return the generated element (similar to `buildTemplateWithoutSections`). An example would be `result-date.cypress.ts`'s `addResultDateInResultList`.

packages/atomic/cypress/integration/segmented-facet-scrollable-actions.ts
btaillon: Alternatively, you could do `cy.get(scrollableComponent).then(([el]) => el.dispatchEvent(...))`. Although ideally, it would be nicer if the cypress method worked 🤔

btaillon (approved):

packages/atomic/cypress/integration/segmented-facet-scrollable-actions.ts
mbenkirane-coveo: Will try this !

packages/atomic/cypress/fixtures/test-fixture.ts
mbenkirane-coveo: Sure, I can create a case-specific one

packages/atomic/cypress/integration/segmented-facet-scrollable-actions.ts
mbenkirane-coveo: I think that `addResultDateInResultList `'s code is more convoluted and harder to follow.

packages/atomic/cypress/integration/segmented-facet-scrollable-actions.ts
```diff
@@ -0,0 +1,44 @@
...
+} from '../fixtures/test-fixture';
+import {segmentedFacetComponent} from './facets/segmented-facet/segmented-facet-selectors';
+import {
+  scrollableComponent,
+  ScrollableSelectors,
+} from './segmented-facet-scrollable-selectors';
+
+type direction = 'right' | 'left';
+
+let scrollable!: HTMLElement;
```
ThibodeauJF, line 15: We shouldn't save this globally

packages/atomic/cypress/integration/segmented-facet-scrollable-actions.ts
```diff
@@ -0,0 +1,44 @@
...
+
+function appendSegmentedFacet(facetProps: TagProps) {
```
ThibodeauJF, line 17:
```suggestion
function appendSegmentedFacet(scrollable: Element, facetProps: TagProps) {
```

ThibodeauJF: Same for "addScrollable"

packages/atomic/cypress/integration/segmented-facet-scrollable.cypress.ts
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  describe('with invalid segmented facets', () => {
+    function setupWithInvalidFacets() {
+      new TestFixture()
+        .with(addScrollable({field: 'invalidd', 'number-of-values': 4}))
+        .init();
+    }
+    before(setupWithInvalidFacets);
+    ScrollableAssertions.assertDisplayScrollable(false);
```
liviarett, line 105: are there any console errors in this case? if so, it might be nice to check that they are being logged. it looks like console functions are stubbed under `TestFixture.consoleAliases.error/log/warn`:

```js
    cy.window().then((win) => {
      cy.stub(win.console, 'error').as(
        TestFixture.consoleAliases.error.substring(1)
      );
      cy.stub(win.console, 'warn').as(
        TestFixture.consoleAliases.warn.substring(1)
      );
      cy.stub(win.console, 'log').as(
        TestFixture.consoleAliases.log.substring(1)
      );
    });
```

liviarett (approved):

packages/atomic/cypress/integration/segmented-facet-scrollable.cypress.ts
mbenkirane-coveo: There aren't any console errors.

PR #2230, opened 2022-07-20
feat(atomic): add "selected-value" parts for selected facet values

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1823

### DIFF
--- a/packages/atomic/src/components/search/facets/facet-value-box/facet-value-box.tsx
+++ b/packages/atomic/src/components/search/facets/facet-value-box/facet-value-box.tsx
@@ -21,7 +21,7 @@ export const FacetValueBox: FunctionalComponent<FacetValueProps> = (
     <li key={props.displayValue}>
       <Button
         style="outline-bg-neutral"
-        part="value-box"
+        part={`value-box${props.isSelected ? ' selected-value' : ''}`}
         onClick={() => props.onClick()}
         class={`value-box box-border w-full h-full items-center p-2 group ${
           props.isSelected ? 'selected' : ''
--- a/packages/atomic/src/components/search/facets/facet-value-checkbox/facet-value-checkbox.tsx
+++ b/packages/atomic/src/components/search/facets/facet-value-checkbox/facet-value-checkbox.tsx
@@ -22,7 +22,7 @@ export const FacetValueCheckbox: FunctionalComponent<FacetValueProps> = (
       <button
         id={id}
         role="checkbox"
-        part="value-checkbox"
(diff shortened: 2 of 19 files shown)


packages/atomic/src/components/search/facets/facet-value-link/facet-value-link.tsx
```diff
@@ -17,7 +17,9 @@ export const FacetValueLink: FunctionalComponent<FacetValueProps> = (
     <li key={props.displayValue} class={props.class}>
       <Button
         style="text-neutral"
-        part={props.part ?? 'value-link'}
+        part={
+          props.part ?? `value-link${props.isSelected ? ' selected-value' : ''}`
```
mbenkirane-coveo, line 21: Why added selected-value when we can access the aria-pressed value?

ThibodeauJF: Parts are used to access some element from the outside of the web component, it's not doable with aria-pressed AFAIK

olamothe (approved):

packages/atomic/src/components/search/facets/facet-value-checkbox/facet-value-checkbox.tsx
```diff
@@ -22,7 +22,7 @@ export const FacetValueCheckbox: FunctionalComponent<FacetValueProps> = (
       <button
         id={id}
         role="checkbox"
-        part="value-checkbox"
+        part={`value-checkbox${props.isSelected ? ' selected-value' : ''}`}
```
btaillon, line 25: I would skip checkboxes since I've done them in my PR. For things like a checkbox where multiple values can be selected, I prefer to have both a checked and unchecked part. Additionally, I prefer to prefix the part with the generic part (e.g., `value-checkbox-unchecked`).

packages/atomic/src/components/search/facets/facet-value-box/facet-value-box.tsx
```diff
@@ -21,7 +21,7 @@ export const FacetValueBox: FunctionalComponent<FacetValueProps> = (
     <li key={props.displayValue}>
       <Button
         style="outline-bg-neutral"
-        part="value-box"
+        part={`value-box${props.isSelected ? ' selected-value' : ''}`}
```
btaillon, line 24: Since multiple values can be the "selected value" (as far as I'm aware at-least), I would have both a selected and unselected part. I would also prefer to have both parts be prefixed with `value-box` to reflect that they are on the same element (e.g., `value-box-unselected`).

btaillon (approved):

PR #2232, opened 2022-07-21
feat(atomic): added -checked part to checkbox facets

btaillon:
I extracted the facet value's checkbox into an accessible Checkbox functional component. `:checked` doesn't work since this is not a real checkbox.

I also changed the background color of checkboxes when hovered, so that they are accessible even when standalone. It doesn't look any worse than before in a facet value IMO.

When the checkbox is checked, the part `PART-checked` is added.

# Standalone
Checked:
image
Checked & hovered:
image

# Facet value
Checked:
image
Checked & hovered:
image

https://coveord.atlassian.net/browse/KIT-1868

### DIFF
--- a/packages/atomic/cypress/integration/facets/facet/facet-selectors.ts
+++ b/packages/atomic/cypress/integration/facets/facet/facet-selectors.ts
@@ -18,10 +18,14 @@ export const FacetSelectors = {
     return this.shadow().find('[part="placeholder"]');
   },
   selectedCheckboxValue() {
-    return this.shadow().find('[part~="value-checkbox"][aria-checked="true"]');
+    return this.shadow().find(
+      '[part~="value-checkbox"][part~="value-checkbox-checked"][aria-checked="true"]'
+    );
   },
   idleCheckboxValue() {
-    return this.shadow().find('[part~="value-checkbox"][aria-checked="false"]');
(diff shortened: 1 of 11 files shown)


packages/atomic/cypress/integration/facets/facet/facet-selectors.ts
```diff
@@ -18,17 +18,17 @@ export const FacetSelectors = {
...
-    return this.shadow().find('[part="value-checkbox"][aria-checked="false"]');
+    return this.shadow().find('[part~="value-checkbox"][aria-checked="false"]');
   },
   checkboxValueWithText(text: string) {
     return this.shadow()
       .find(`[part="value-label"]:contains("${text}")`)
       .parent()
       .parent()
-      .find('[part="value-checkbox"]');
+      .find('[part~="value-checkbox"]');
```
btaillon, line 31: `~=` will check match:
"value-checkbox value-checkbox-checked"
but will not match:
"value-checkboxx value-checkbox-checked"

ThibodeauJF:
Part of this is a duplicate of a PR I opened yesterday https://github.com/coveo/ui-kit/pull/2230

jpmarceau (approved):
👍

PR #2234, opened 2022-07-22
fix(quantic): changed ... to expand from a A tag to a button

drosenfeld-coveo:
[SVCC-2277](https://coveord.atlassian.net/browse/SVCC-2277)

Changed the clickable object from an A tag to a button as it does not redirect to another page. The mac screenreader now detects it as a expandable object

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticResultPrintableUri/quanticResultPrintableUri.html
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultPrintableUri/quanticResultPrintableUri.html
@@ -7,7 +7,7 @@
       <template for:each={foldedParents} for:item="parent">
         <div class="result-parenturi__container slds-grid slds-grid_vertical-align-center" key={parent.id}>
           <template if:true={parent.isFolded}>
-            <lightning-button variant="base" label="..." title="Expand URI parts" aria-label="Expand URI parts" onclick={expandParents}>...</lightning-button>
+            <button class="slds-button expand-parenturi__container" variant="base" label="..." title="Expand URI parts" aria-label="Expand URI parts" onclick={expandParents}>...</button>
           </template>
           <template if:false={parent.isFolded}>
             <a class={parent.classes} href={parent.uri} target={target}>{parent.name}</a>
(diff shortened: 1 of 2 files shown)


nathanlb:
Could you post of screenshot or little screen capture video of what it looks like after the change?

drosenfeld-coveo:
> Could you post a screenshot or little screen capture video of what it looks like after the change?

Uploading Screen Recording 2022-07-25 at 3.34.17 PM.mov…

aelfaleh (approved):

erocheleau:
If you check my other PRs, I changed most of the <a> for <button> instead of lightning-button and what you need to make it look like before is this:

`class="slds-button"` this will give you the `variant="bare"` look, and you need to add the line-height: 1.5 to override the higher line-height from the salesforce buttons I think.

Don't know if overriding the height is possible if you use a lightning-button.

lvu285:
I really like the part for aria-label, so it will have the consistent text across different screen reader

Probably out-of-scope, but I'm wonder if we can add something, to tell the user that user navigates to the "source-url-part" section? and totally how many "link" part we have right now? ("..." can count as 1 if it's in collapsed state)

nathanlb (approved):

packages/quantic/force-app/main/default/lwc/quanticResultPrintableUri/quanticResultPrintableUri.html
```diff
@@ -7,7 +7,7 @@
       <template for:each={foldedParents} for:item="parent">
         <div class="result-parenturi__container slds-grid slds-grid_vertical-align-center" key={parent.id}>
           <template if:true={parent.isFolded}>
-            <a onclick={expandParents}>...</a>
+            <button class="slds-button expand-parenturi__container" variant="base" label="..." title="Expand URI parts" aria-label="Expand URI parts" onclick={expandParents}>...</button>
```
drosenfeld-coveo, line 10: @acarpentier01 what do you think of the text "Expand URI parts"?

acarpentier01: I think "Expand URI" or "Expand path" would do the trick @drosenfeld-coveo. @jpmarceau Do you agree?

jpmarceau: I think so, unless by `part` you mean something technical, like a CSS part?

drosenfeld-coveo: I dont think the "part" has a specific meaning, I like "Expand URI"

acarpentier01: @erocheleau Just to confirm, is this is a URI or a breadcrumb? If it's a URI, then please go with "Expand URI" @drosenfeld-coveo.

btaillon (approved):

PR #2235, opened 2022-07-22
chore(quantic): read sfdx error JSON output from stdout instead of stderr

lbergeron:
[SVCC-2349](https://coveord.atlassian.net/browse/SVCC-2349)

Apparently a recent version of sfdx-cli introduced a change in the way errors are returned. Before, the error JSON response would be returned through stderr. It is now returned through stdout.

It prevented our deploy-community script from deploying on an existing scratch org.

### DIFF
--- a/packages/quantic/scripts/build/util/sfdx.ts
+++ b/packages/quantic/scripts/build/util/sfdx.ts
@@ -27,7 +27,14 @@ export function sfdx<T = SfdxResponse>(command: string): Promise<T> {
       },
       (error, stdout) => {
         if (error) {
-          reject(error);
+          let jsonOutput;
+          try {
+            jsonOutput = JSON.parse(strip(stdout));
+          } catch (e) {
+            // The output is not JSON. The command most likely failed outside the SFDX CLI.
+          }
+
+          reject(jsonOutput || error);
         }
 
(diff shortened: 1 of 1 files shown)


aelfaleh (approved):

nathanlb (approved):

packages/quantic/scripts/build/util/sfdx.ts
```diff
@@ -27,7 +27,14 @@ export function sfdx<T = SfdxResponse>(command: string): Promise<T> {
       },
       (error, stdout) => {
         if (error) {
-          reject(error);
+          let jsonOutput;
+          try {
+            jsonOutput = JSON.parse(strip(stdout));
+          } catch (e) {
+            // The output is not JSON. The command most likely failed outside the SFDX CLI.
```
mmitiche, line 34: Don't we wanna log something here? just to facilitate debugging by indicating that the command most likely failed outside the SFDX CLI?

mmitiche (approved):

packages/quantic/scripts/build/util/sfdx.ts
lbergeron: No, we don't need to log the error here because the promise is rejected with the error, and it is logged at a higher level.

PR #2237, opened 2022-07-22
feat(quantic): added accessibility functionality to quanticFacetValue

MSLI-Coveo:
[SVCC-2272](https://coveord.atlassian.net/browse/SVCC-2272)
[SVCC-2273](https://coveord.atlassian.net/browse/SVCC-2273)
[SVCC-2274](https://coveord.atlassian.net/browse/SVCC-2274)
[SVCC-2287](https://coveord.atlassian.net/browse/SVCC-2287)
[SVCC-2285](https://coveord.atlassian.net/browse/SVCC-2285)

Screen Shot 2022-07-29 at 10 17 44 AM

Screen Shot 2022-07-29 at 10 17 17 AM

Accessibility
Added dom element "fieldset" and "legend" for voiceover
Edited tabindex to change the accessibility of certain elements.
Function onKeyDown added to track keyboard input

Voiceover
Added attribute "role",   "" for screen reader. (Some element would be included by voice over by default)


packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -31,16 +31,23 @@
...
-                  onselectvalue={onSelectValue} 
-                  key={v.value} 
-                  item={v} 
-                  is-checked={v.checked}
-                  data-cy={v.value}
-                  display-as-link={isDisplayAsLink}>
-                </c-quantic-facet-value>
-              </template>
+              <fieldset> 
+                <legend class="slds-assistive-text">Test</legend>
```
lbergeron, line 35: Why did you put `Test` inside the `legend` element? Shouldn't it contain the facet title or something like that instead?

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -69,5 +73,22 @@ export default class QuanticFacetValue extends LightningElement {
         value: this.formattedFacetValue
       }
     }));
+  }
+    /**
+   * @param {KeyboardEvent} evt
+   */
```
lbergeron, line 79:
```suggestion
    /**
     * @param {KeyboardEvent} evt
     */
```

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -58,6 +60,13 @@ export default class QuanticFacetValue extends LightningElement {
     return new Intl.NumberFormat(LOCALE).format(this.item.numberOfResults);
   }
 
+  get divRole (){
```
erocheleau, line 63:
```suggestion
  get divRole() {
```

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -58,6 +60,13 @@ export default class QuanticFacetValue extends LightningElement {
...
+    return this.displayAsLink ? "button":"checkbox"
```
erocheleau, line 64:
```suggestion
    return this.displayAsLink ? "button" : "checkbox";
```

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -58,6 +60,13 @@ export default class QuanticFacetValue extends LightningElement {
...
+  }
+  get ariaLabelVoiceOver(){
+    return `Filter, ${this.formattedFacetValue}, ${this.numberOfResults}`
```
erocheleau, line 67: @acarpentier01 This is the text that will be read by a screen reader if that makes sense:
@adroletCoveo This is an example of what I was talking about.

Example: 🔊 "Filter, pdf, 1234"

erocheleau:
To be clear about what this proposed solution does @MSLI-Coveo , 
this keeps the original HTML except changing the <li> to be in the Facet component instead of the FacetValue component right? 

Also it adds a FieldSet and Legend + a div role checkbox|button attribute?

erocheleau:
You need to run a linter on your code as well to fix the formatting and to make sure to have `;` everywhere etc.

I am also unsure of having a `aria-checked` attribute on a `<div role="button"` is compliant but maybe we can overlook it. 
We can talk about it during the review.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -31,16 +31,39 @@
...
-                  onselectvalue={onSelectValue} 
-                  key={v.value} 
-                  item={v} 
-                  is-checked={v.checked}
-                  data-cy={v.value}
-                  display-as-link={isDisplayAsLink}>
-                </c-quantic-facet-value>
-              </template>
+              <fieldset> 
+                <legend class="slds-assistive-text">{field}</legend>
```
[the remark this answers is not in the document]
MSLI-Coveo: For this legend element, we might need to consider putting other variables here as the value of variable "field" is not readable by screen readers.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -31,16 +31,39 @@
...
+                      display-new
+                      key={v.value}
+                      onselectvalue={onSelectValue} 
+                      item={v} 
+                      is-checked={v.checked}
+                      data-cy={v.value}
+                      display-as-link={isDisplayAsLink}>
+                    </c-quantic-facet-value>
+                </template>
+              </div>
```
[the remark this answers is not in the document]
MSLI-Coveo: Yes, this is the 2nd potential solution. And it is now moved to [here](https://github.com/coveo/ui-kit/pull/2248)

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.css
```diff
@@ -17,3 +17,56 @@ li {
...
+  display: block;
+  position: absolute;
+  width: 13px;
+  border-top-right-radius: 1px;
+  border-bottom-right-radius: 1px;
+  top: 10px;
+  left: 4px;
+  transform: rotate(-49deg);
+  transform-origin: left center;
+}
```
nathanlb, line 72: Looks like this isn't used anywhere

nathanlb, line 72: Looks like this isn't used anywhere

lvu285:
can you build the live-demo link for me to test with the screen reader ? 
@erocheleau fyi

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
MSLI-Coveo: formatting added

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.html
```diff
@@ -16,17 +17,44 @@
         <span class="slds-truncate slds-var-m-right_xx-small slds-var-m-top_xxx-small nowrap facet__value-text facet__value_selected">{formattedFacetValue}</span>
         <span class="nowrap slds-var-m-top_xxx-small facet__number-of-results facet__value_selected">({numberOfResults})</span>
       </template>
+      <template if:false={isChecked}>
```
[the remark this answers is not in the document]
MSLI-Coveo: formatting added

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.css
MSLI-Coveo: > 

Left over from another implementation. Should be removed/

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
MSLI-Coveo: formatter added

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
MSLI-Coveo: > @acarpentier01 This is the text that will be read by a screen reader if that makes sense: @adroletCoveo This is an example of what I was talking about.
> 
> Example: 🔊 "Filter, pdf, 1234"
not a issue, set to resolved.

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
MSLI-Coveo: formmater added

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.html
```diff
@@ -1,32 +1,54 @@
...
+        name="facet value checkbox"
+        onchange={onSelect}
+        checked={isChecked}
+      ></lightning-input>
+    </template>
+    <template if:true={isChecked}>
+      <span
+        class="slds-truncate slds-var-m-right_xx-small slds-var-m-top_xxx-small nowrap facet__value-text facet__value_selected"
+        >{formattedFacetValue}</span
+      >
```
erocheleau, line 27: You didn't change anything there right? Is it a formatter that changed the look of this file?

erocheleau (commented):
Can you please check the formatting of the file where I commented? I think it would be better if the whole file didn't look like it was midified bcause you formatted it.

Or we take the plunge and accept it as formatted now if that format is the same as everywhere else?

erocheleau:
@MSLI-Coveo I added this jira too, since you are fixing it at the same time in this PR right?

[SVCC-2287](https://coveord.atlassian.net/browse/SVCC-2287)

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.html
MSLI-Coveo: For the HTML formatter, it seems in .prettierignore file, all html files under lwc folder is excluded to use prettier as formatter. Which formatter should be applied to html files under dir lwc?

lbergeron (commented):
It looks pretty good! Well done

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet-actions.ts
```diff
@@ -18,6 +18,9 @@ const numericFacetActions = (selector: AllFacetSelectors) => {
     submitManualRange: () => {
       selector.searchForm().submit();
     },
+    findCollapseButtonPressTabPressSpace: () => {
```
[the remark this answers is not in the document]
MSLI-Coveo: It is quit tricky to find names for these actions. This action intents to

set focus to Collapse button at the top of each facet
press Tab key (expect focus to be set on first selectable row)
Press Enter (expect first selectable row to be selected using Enter key)

packages/quantic/cypress/integration/facets/facet/facet.cypress.ts
```diff
@@ -338,6 +338,21 @@ describe('Facet Test Suite', () => {
...
+    it('should work as expected with accessibility support', () => {
+      setupWithValues();
+      const selectedValue = 'People';
+      scope('should be selected when space is pressed', () => {
+        Actions.findSearchInputPressTabPressSpace();
+        Expect.selectedValuesContain(selectedValue);
+        Expect.numberOfSelectedLinkValues(1);
+      });
+      scope('should be selected when Enter is pressed', () => {
+        Actions.findSearchInputPressTabPressEnter();
```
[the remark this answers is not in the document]
MSLI-Coveo: Its quit tricky to find names for these kinds of actions since testing accessibility feature is normally involved with setting focus to one element and then move focus to other element using key Tab. Under this setup, the element first focused is not the element that ends up being selected.

Any suggestions for the naming convention for these kind of Actions where key Tab is involved as we certainly need to implement similar tests moving forward.

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet-actions.ts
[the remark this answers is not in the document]
lbergeron: So, if I understand correctly, it selects the first value of each facet?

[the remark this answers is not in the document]
MSLI-Coveo: Yes it selects the first value of the numeric-facet

packages/quantic/cypress/integration/facets/facet/facet.cypress.ts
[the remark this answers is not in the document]
lbergeron: > Any suggestions for the naming convention for these kind of Actions where key Tab is involved as we certainly need to implement similar tests moving forward.

Personally, I find it easier to name things by thinking about the outcome. What happens when this function is called (not individual operations, but what it means for the user)?

For example, suppose we have a filter component where each value has a checkbox. The user also has to click an Apply button to make the changes effective.

In order to change the selected value our function would need to:
* uncheck all values
* check the requested value
* click the Apply button

Naming this function `deselectAllValuesThenCheckXThenApply` describes the operations that are performed by the function, but not the outcome.

In this case, the goal of this function is to replace a filter value, so naming the function `replaceFilterValue` is more meaningful. In addition it is loosely coupled to the implementation so the function name remain relevant even if the markup changes.

Regarding accessibility, we could group the accessibility actions together in an object (e.g., `A11yActions`) so the keyboard navigation would become implicit.

What do you think? Does it make sense to you?

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet-actions.ts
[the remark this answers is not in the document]
lbergeron: I guess `selectFirstValue` would better carry the meaning then. Now, to make it clearer that it is meant for accessibility, we have a couple options:

* append `UsingKeyboard` to the function name so it becomes `selectFirstValueUsingKeyboard`
* prefix with `a11y` to make it obvious that it is meant to test accessibility: `a11ySelectFirstValue`
* or use a separate object to store the accessibility actions. The function call would then look like `A11yNumericFacetActions.selectFirstValue()`

Since we are expecting to add many actions to test accessibility, I think the third option would be the cleanest.

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -59,6 +59,14 @@ export default class QuanticFacetValue extends LightningElement {
     return new Intl.NumberFormat(LOCALE).format(this.item.numberOfResults);
   }
 
+  get divRole() {
+    return this.displayAsLink ? 'button' : 'checkbox';
+  }
+
+  get ariaLabelVoiceOver() {
+    return `Filter, ${this.formattedFacetValue}, ${this.numberOfResults}`;
```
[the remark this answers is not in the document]
MSLI-Coveo: No. will check the output of atomic component and update.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +14,139 @@
...
+            <fieldset>
+              <legend class="slds-assistive-text">{field}</legend>
+              <ul>
+                <template if:true={hasParents}>
+                  <li
+                    class="facet__value-option slds-size_1-of-1 slds-p-around_x-small"
+                  >
+                    <div
+                      class="facet__allCategories slds-grid slds-grid_vertical-align-center"
+                      onclick={reset}
```
btaillon, line 45: This element has an onclick but isn't a `button`, therefor unusable by screen readers. I would recommend transforming this into a `button` element.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +14,139 @@
...
+                      <lightning-icon
+                        class="slds-m-right_xxx-small"
+                        icon-name="utility:chevronleft"
+                        size="xx-small"
+                      ></lightning-icon>
+                      <span class="nowrap">{labels.allCategories}</span>
+                    </div>
+                  </li>
+                </template>
+                <template for:each={nonActiveParents} for:item="v">
```
btaillon, line 56: If possible, I would recommend rendering results as a hierarchy, which should improve accessibility. This is what we tried to do in https://github.com/coveo/ui-kit/pull/2029.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +14,139 @@
...
             </div>
           </template>
         </ul>
         <div class="facet__show slds-m-top_x-small slds-p-horizontal_x-small">
           <template if:true={canShowLess}>
-            <button variant="base" label={labels.showLess}
-              class="facet__show-less slds-p-around_none" onclick={showLess}>
+            <button
+              variant="base"
+              label={labels.showLess}
```
btaillon, line 133: did you mean to use `aria-label`? either way, an `aria-label` isn't necessary here since the button already contains the same label as a child.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -16,41 +19,77 @@
...
-            <lightning-icon class="slds-current-color slds-m-right_xx-small" icon-name="utility:close" alternative-text={altText} size="xx-small"></lightning-icon>
-            <span class="nowrap pill__text-container slds-truncate">{clearFilterLabel}</span>
+          <button
+            class="facet__clear-filter slds-button slds-grid slds-grid_vertical-align-center slds-p-top_x-small slds-p-horizontal_x-small"
+            onclick={clearSelections}
+          >
+            <lightning-icon
+              class="slds-current-color slds-m-right_xx-small"
+              icon-name="utility:close"
+              alternative-text={altText}
```
btaillon, line 29: Since the parent is a button, I would instead give thi icon `aria-hidden`.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -16,41 +19,77 @@
...
+              </fieldset>
+              <div
+                class="facet__show slds-m-top_x-small slds-p-horizontal_x-small"
+              >
                 <template if:true={canShowLess}>
-                  <button variant="base" label={labels.showLess}
-                    class="facet__show-less slds-p-around_none" onclick={showLess} data-cy="less">
+                  <button
+                    variant="base"
+                    label={labels.showLess}
```
btaillon, line 77: Same as above, this label is redundant.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
+                class="facet__searchbox-input slds-p-horizontal_x-small slds-var-p-bottom_x-small"
+                name={labels.search}
+                type="search"
+                label="facet-input"
+                variant="label-hidden"
+                placeholder={labels.search}
+              ></lightning-input>
             </div>
           </template>
+
```
erocheleau, line 31: Remove empty line

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
erocheleau: It has the `role="button"` so it works, it will be considered a button by screen readers.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
-              <ul class="slds-size_1-of-1">
+            <fieldset>
+              <legend class="slds-assistive-text">{field}</legend>
+              <ul>
+                <template if:true={hasParents}>
+                  <li class="facet__value-option slds-size_1-of-1 slds-p-around_x-small">
+                    <div
+                      class="facet__allCategories slds-grid slds-grid_vertical-align-center"
+                      role="button"
+                      aria-pressed="false"
```
erocheleau, line 41: The `aria-pressed` attribute should only be used on "toggle" button.
It makes no sense that it's there but has permanently the value "false", if it's a toggle button you need a toggle state, if it's not a toggle button, don't include the `aria-pressed` .

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
+                      onclick={reset}
```
erocheleau, line 42: I believe you may have to also include a keypress listener for it to work with the spacebar?
Does it work without a keypress handler because it has the role="button"?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
erocheleau: Good idea but out of scope for this PR.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
-                </div> 
-              </li> 
-            </template>
-            <template for:each={nonActiveParents} for:item="v">
-              <c-quantic-category-facet-value onselectvalue={onSelectValue} key={v.value} item={v} non-active-parent="true"></c-quantic-category-facet-value>
-            </template>
-            <template if:true={activeParent}>
-              <li class="facet__active-parent slds-grid slds-grid_vertical-align-center slds-p-around_x-small slds-m-left_medium">
-                <span class="nowrap parent-text__bold">{activeParent.value}</span>
-                <span class="facet__number-of-results parent-text__bold slds-var-m-left_xx-small">({activeParent.numberOfResults})</span>
```
erocheleau, line 38: Why did this become a quantic-category-facet-value???

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
-              </li>  
-            </template>
-            <li class="slds-grid">
-              <ul class="slds-size_1-of-1">
```
erocheleau, line 42: Why was that class removed?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
+              <ul>
                 <template for:each={values} for:item="v">
-                  <c-quantic-category-facet-value
-                    onselectvalue={onSelectValue} 
-                    key={v.value} 
-                    item={v}
-                    active-parent={activeParent}>
-                  </c-quantic-category-facet-value>
+                  <li class={categoryFacetLiClass} key={v.value}>
+                    <c-quantic-category-facet-value onselectvalue={onSelectValue} item={v}>
```
erocheleau, line 79: Why did the `active-parent` go?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
-                  item={result} 
-                  is-search-result="true">
-                </c-quantic-category-facet-value>
-              </template>
+              <fieldset>
+                <legend class="slds-assistive-text">{field}</legend>
+                <ul class="slds-size_1-of-1">
+                  <template for:each={facetSearchResults} for:item="v">
+                    <li class={categoryFacetLiClass} key={v.value}>
+                      <c-quantic-category-facet-value onselectvalue={onSelectValue} item={v} is-search-result="true">
```
erocheleau, line 94: Why is this one not prettified like the others?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
+                </ul>
+              </fieldset>
             </template>
             <div class="slds-p-horizontal_x-small">
               <template if:false={hasSearchResults}>
                 <li class="slds-grid slds-var-m-top_x-small">
-                  <span class="slds-truncate slds-var-p-around_small slds-size_1-of-1 facet__search-results_no-match">{noMatchesForLabel}</span> 
+                  <span class="slds-truncate slds-var-p-around_small slds-size_1-of-1 facet__search-results_no-match"
+                    >{noMatchesForLabel}</span
+                  >
```
erocheleau, line 106: This is too weird either:

```
<span class="...">
  {value}
</span>

or

<span class="...">{value}</span>
```
please

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
           </template>
-        </ul>
+        </div>
         <div class="facet__show slds-m-top_x-small slds-p-horizontal_x-small">
           <template if:true={canShowLess}>
-            <button variant="base" label={labels.showLess}
-              class="facet__show-less slds-p-around_none" onclick={showLess}>
+            <button
+              variant="base"
+              aria-label="Show less"
```
erocheleau, line 121: You don't need that `aria-label` since you have a span with a label inside the button's HTML

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -11,84 +11,130 @@
...
+            >
               <span class="facet__show-less_text">{labels.showLess}</span>
             </button>
           </template>
           <template if:true={canShowMore}>
-            <button variant="base" label={labels.showMore}
-              class="facet__show-more slds-p-around_none" onclick={showMore}>
+            <button
+              variant="base"
+              aria-label="Show more"
```
erocheleau, line 131: You don't need that aria-label since you have a span with a label inside the button's HTML

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -332,6 +332,10 @@ export default class QuanticCategoryFacet extends LightningElement {
     return this.withSearch && !!this.input?.value?.length;
   }
 
+  get categoryFacetLiClass() {
+    return this.activeParent ? 'slds-var-m-left_large slds-grid' : 'slds-grid';
```
erocheleau, line 336:
```suggestion
    return `slds-grid ${this.activeParent ? 'slds-var-m-left_large': ''}`;
```

Personal choice if you want to change it but I prefer it like this.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
```diff
@@ -28,7 +53,7 @@
             </div>
           </div>
         </template>
-      </template>
-    </div>
-  </li>
+      </template></template
```
erocheleau, line 56: missing new line between the /template

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
```diff
@@ -1,25 +1,50 @@
...
+    class="slds-size_1-of-1 facet__value-container"
+    onclick={onSelect}
+    tabindex="0"
+    onkeydown={onKeyDown}
+    aria-label={ariaLabelVoiceOver}
+  >
+    <template if:true={isPressed}>
+      <span class="nowrap parent-text__bold">{item.value}</span>
+      <span class="facet__number-of-results parent-text__bold slds-var-m-left_xx-small">({item.numberOfResults})</span>
+    </template>
```
erocheleau, line 14: What is this new "isPressed" ?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
```diff
@@ -1,25 +1,50 @@
...
           <span class="nowrap">{item.value}</span>
           <span class="facet__number-of-results slds-var-m-left_xx-small">({item.numberOfResults})</span>
         </div>
       </template>
       <template if:false={nonActiveParent}>
         <template if:false={isSearchResult}>
-          <div class="nowrap slds-grid slds-grid_vertical-align-center slds-text-color_default facet__child-value-option facet__value-option slds-size_1-of-1 slds-p-around_x-small">
+          <div
+            class="nowrap slds-grid slds-grid_vertical-align-center slds-text-color_default facet__child-value-option facet__value-option slds-size_1-of-1 slds-p-around_x-small"
+          >
```
erocheleau, line 33: On one line please.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
```diff
@@ -1,25 +1,50 @@
...
           </div>
         </template>
         <template if:true={isSearchResult}>
           <div class="facet__value-option slds-size_1-of-1 slds-p-around_x-small slds-text-color_default">
             <div class="slds-grid slds-grid_vertical">
-              <div class="slds-col slds-p-left_none slds-grid_vertical-align-center facet__search-results slds-m-bottom_xxx-small">
-                <lightning-formatted-rich-text class="slds-truncate slds-float_left" value={item.highlightedResult}></lightning-formatted-rich-text>
+              <div
+                class="slds-col slds-p-left_none slds-grid_vertical-align-center facet__search-results slds-m-bottom_xxx-small"
+              >
```
erocheleau, line 43: On one line please.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
```diff
@@ -25,13 +25,7 @@ export default class QuanticCategoryFacetValue extends LightningElement {
    * @defaultValue `false`
    */
   @api isSearchResult = false;
-  /**
-   * Whether the value is an active parent node.
-   * @api
-   * @type {boolean}
-   * @defaultValue `false`
-   */
-  @api activeParent = false;
```
erocheleau, line 34: You cannot remove an attribute that is marked as @api this is a breaking change.

Please explain why this would be necessary?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
```diff
@@ -44,14 +38,18 @@ export default class QuanticCategoryFacetValue extends LightningElement {
...
-
   get facetValue() {
     return this.item.value;
   }
 
+  get isPressed() {
+    return this.item.state === 'selected';
+  }
+
+  get ariaLabelVoiceOver() {
```
erocheleau, line 49:
```suggestion
  get ariaLabel() {
```

packages/quantic/force-app/main/default/lwc/quanticDateFacet/quanticDateFacet.html
```diff
@@ -22,18 +22,22 @@
...
-                onselectvalue={onSelectValue} 
-                key={v.start} 
-                item={v} 
-                is-checked={v.checked}
-                formatting-function={formattingFunction}
-              >
-              </c-quantic-facet-value>
-            </template>
-          </ul>
+          <fieldset>
```
erocheleau, line 25: That's the kind of change I was expecting, short and sweet and it resolves the accessibility issues wtihout touching anything else.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
erocheleau: +1 for this, you already have text on the button's span, no need to have a label on the icon, and aria-hidden will make screen readers completely ignore it.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -16,41 +16,73 @@
...
-                <c-quantic-facet-value 
-                  onselectvalue={onSelectValue} 
-                  key={v.value} 
-                  item={v} 
-                  is-checked={v.checked}
-                  data-cy={v.value}
-                  display-as-link={isDisplayAsLink}>
-                </c-quantic-facet-value>
-              </template>
+              <fieldset>
```
erocheleau, line 48: Perfect

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -16,41 +16,73 @@
...
+                </ul>
+              </fieldset>
               <div class="facet__show slds-m-top_x-small slds-p-horizontal_x-small">
                 <template if:true={canShowLess}>
-                  <button variant="base" label={labels.showLess}
-                    class="facet__show-less slds-p-around_none" onclick={showLess} data-cy="less">
+                  <button
+                    variant="base"
+                    class="facet__show-less slds-p-around_none"
+                    aria-label="Show less"
```
erocheleau, line 71: You don't need that aria-label since you have a span with a label inside the button's HTML

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -16,41 +16,73 @@
...
+                  >
                     <span class="facet__show-less_text">{labels.showLess}</span>
                   </button>
                 </template>
                 <template if:true={canShowMore}>
-                  <button variant="base" label={labels.showMore}
-                    class="facet__show-more slds-p-around_none" onclick={showMore} data-cy="more">
+                  <button
+                    variant="base"
+                    aria-label="Show more"
```
erocheleau, line 81: You don't need that aria-label since you have a span with a label inside the button's HTML

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -60,19 +92,22 @@
...
                   </c-quantic-facet-value>
                 </template>
               </template>
               <div class="slds-p-horizontal_x-small">
                 <template if:false={hasSearchResults}>
                   <li class="slds-grid slds-m-top_x-small">
-                    <span class="slds-truncate slds-p-around_small slds-size_1-of-1 facet__search-results_no-match">{noMatchesForLabel}</span>
+                    <span class="slds-truncate slds-p-around_small slds-size_1-of-1 facet__search-results_no-match"
+                      >{noMatchesForLabel}</span
+                    >
```
erocheleau, line 110: One line please.

packages/quantic/force-app/main/default/lwc/quanticNumericFacet/quanticNumericFacet.html
```diff
@@ -68,15 +68,22 @@
...
-               <template for:each={values} for:item="v">
-                <c-quantic-facet-value 
-                  onselectvalue={onSelectValue} 
-                  key={v.key} 
-                  item={v} 
-                  is-checked={v.checked}
-                  formatting-function={formattingFunction}>
-                </c-quantic-facet-value>
-              </template>
+              <fieldset>
```
erocheleau, line 71: Perfect

erocheleau (changes requested):
Okay while I think this is a good idea to include something like axe-core in the project after we've investigated what this thing actually checks, I would definitely not include this in a PR that is adding a little bit of functionality to FacetValues.

Be careful of scope creep when working on something.

Two scenarios here:

1. this `checkA11y` thing is the solution we've been looking for for all A11y tests of quantic.

In this case, finish this PR with just the scope that was required. And let's look at this lib afterwards and that might save us some time in the future.

2. This is still good a11y tests to have on our components but not the same as the current issues we've been looking at and we'll include this in the future.

In this case, finish this PR and all other a11y issues we have currently listed while keeping tests at a minimum if they are covered by this with the prospect of adding it in the future.

So in both cases, this is out of scope for this PR. Finish this PR with just what is required and then we'll look at this thing together :) So revert all that's related to axe-core and cypress-axe. And we'll take a look before the next issues we tackle.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
MSLI-Coveo: In the original implementation, the selected value is not rendered using `<c-quantic-category-facet-value>`, unlike other values displayed in the filter. Using `<c-quantic-category-facet-value>` to render selected value, isPressed state needed to passed to <c-quantic-category-facet-value>.

Controlling isPressed within `<c-quantic-category-facet-value>` is also coherent with `<c-quantic-facet-value>`

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
MSLI-Coveo: The attribute active-parent was passed down to `<quantic-category-facet-value>` to generate `categoryFacetLiClass` for element `<li>`, and as  `<li>` is moved to <quantic-category-facet>, active-parent is no longer used in  `<quantic-category-facet-value>`, and   `<quantic-category-facet-value>` also has nonActiveParent  to store the same state.

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.html
```diff
@@ -1,32 +1,43 @@
...
-          class="facet__value-checkbox"
-          variant="label-hidden"
-          label="Facet value option"
-          name="facet value checkbox"
-          onchange={onSelect}
-          checked={isChecked}
-        ></lightning-input>
+  <div
+    role={divRole}
+    aria-checked={isChecked}
```
lvu285, line 4: for Facet checkbox, `aria-checked` is correct, but for Facet-display as link, I think `aria-pressed` makes more sense

packages/quantic/force-app/main/default/lwc/quanticNumericFacet/quanticNumericFacet.html
```diff
@@ -16,69 +16,90 @@
...
-          <button class="facet__clear-filter slds-button slds-grid slds-grid_vertical-align-center slds-var-p-top_x-small slds-var-p-horizontal_x-small" onclick={clearSelections}>
-            <lightning-icon class="slds-current-color slds-var-m-right_xx-small" icon-name="utility:close" alternative-text={altText} size="xx-small"></lightning-icon>
+          <button
+            class="facet__clear-filter slds-button slds-grid slds-grid_vertical-align-center slds-var-p-top_x-small slds-var-p-horizontal_x-small"
+            onclick={clearSelections}
+          >
+            <lightning-icon
+              class="slds-current-color slds-var-m-right_xx-small"
+              icon-name="utility:close"
+              alternative-text={altText}
```
btaillon, line 26: Instead of having a button that contains an icon which has alternative test, I think it would be better to have a button with an aria-label (which describes its action) and `aria-hidden="true"` on the icon.

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.html
```diff
@@ -18,8 +18,16 @@
...
-          <button class="facet__clear-filter slds-button slds-grid slds-grid_vertical-align-center slds-p-top_x-small slds-p-horizontal_x-small" onclick={clearSelections}>
-            <lightning-icon class="slds-current-color slds-m-right_xx-small" icon-name="utility:close" alternative-text={clearFilterLabel} size="xx-small"></lightning-icon>
+          <button
+            class="facet__clear-filter slds-button slds-grid slds-grid_vertical-align-center slds-p-top_x-small slds-p-horizontal_x-small"
+            onclick={clearSelections}
+          >
+            <lightning-icon
+              class="slds-current-color slds-m-right_xx-small"
+              icon-name="utility:close"
+              alternative-text={clearFilterLabel}
```
btaillon, line 28: Same as my other comment. I would move the label to the button.

btaillon (approved):

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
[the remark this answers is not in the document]
MSLI-Coveo: Compared the output with Atomic and changed the output.

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet.cypress.ts
```diff
@@ -375,4 +376,17 @@ describe('quantic-numeric-facet', () => {
       });
     });
   });
+
+  describe('with a11y', () => {
```
lvu285, line 380: probably should put more meaningful title, as `a11y` is not a properties of the facet

packages/quantic/cypress/integration/common-assertions.ts
```diff
@@ -0,0 +1,84 @@
...
+    case 'BUTTON':
+      return !element.hasAttribute('disabled');
+    case 'IFRAME':
+      return true;
+    default:
+      return false;
+  }
+}
+
+function* getFocusableDescendants(element: Element): Generator<HTMLElement> {
```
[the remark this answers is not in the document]
MSLI-Coveo: cypress-axe removed

packages/quantic/cypress/integration/common-assertions.ts
```diff
@@ -0,0 +1,84 @@
+function isFocusable(element: Element) {
+  // Source: https://stackoverflow.com/a/30753870
+  if (element.getAttribute('tabindex') === '-1') {
+    return false;
+  }
+  if (element.hasAttribute('tabindex')) {
+    return true;
+  }
```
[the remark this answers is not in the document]
MSLI-Coveo: cypress-axe removed, file no longer exits

packages/quantic/.prettierrc.js
```diff
@@ -6,8 +6,8 @@ module.exports = {
     overrides: [
       {
         files: '**/lwc/**/*.html',
-        options: {parser: 'lwc'}
-      }
-    ]
-  }
+        options: {parser: 'lwc', printWidth: 140},
```
erocheleau, line 9: @lbergeron Are you guys using prettier locally with this file? or was this generated by default from the LWC project?

packages/quantic/cypress/integration/facets/facet-common-actions.ts
```diff
@@ -54,6 +54,9 @@ export const baseFacetActions = (selector: BaseFacetSelector) => {
     clickExpandButton: () => {
       selector.expandButton().click();
     },
+    selectFirstFacetValueWithKeyboardTab: () => {
+      selector.searchInput().tab().type('{Enter}', {force: true});
+    },
```
erocheleau, line 59: Very nice! :)

packages/quantic/cypress/integration/facets/facet/facet.cypress.ts
```diff
@@ -813,4 +813,24 @@ describe('Facet Test Suite', () => {
...
+        true
+      );
+    });
+
+    it('should be accessible through keyboard', () => {
+      Actions.selectFirstFacetValueWithKeyboardTab();
+      Expect.clearFilterContains('Clear filter');
+      Expect.numberOfSelectedCheckboxValues(1);
+      Expect.numberOfIdleCheckboxValues(defaultNumberOfValues - 1);
+    });
```
erocheleau, line 834: Awesome!

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet.cypress.ts
```diff
@@ -375,4 +376,17 @@ describe('quantic-numeric-facet', () => {
...
+    beforeEach('', () => {
+      visitNumericFacetPage(defaultSettings);
+    });
+
+    it('should be accessible through keyboard', () => {
+      Actions.selectFirstNumericFacetValueWithKeyboardTab();
+      Expect.numberOfSelectedCheckboxValues(1);
+      Actions.selectFirstNumericFacetValueWithKeyboardEnter();
+      Expect.numberOfSelectedCheckboxValues(0);
+    });
```
erocheleau, line 390: Ah I like it, you are selecting then de-selecting nice

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
erocheleau: 👍 looks fixed

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
```diff
@@ -1,34 +1,49 @@
...
-          <lightning-icon class="slds-m-right_xxx-small" icon-name="utility:chevronleft" size="xx-small"></lightning-icon>
-          <span class="nowrap">{item.value}</span>
+  <div
+    role="button"
+    class="slds-size_1-of-1 facet__value-container"
+    onclick={onSelect}
+    aria-pressed={isPressed}
+    tabindex="0"
+    onkeydown={onKeyDown}
+    aria-label={ariaLabelVoiceOver}
```
erocheleau, line 9:
```suggestion
    aria-label={ariaLabel}
```

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
```diff
@@ -52,6 +52,14 @@ export default class QuanticCategoryFacetValue extends LightningElement {
     return this.item.value;
   }
 
+  get ariaLabelVoiceOver() {
```
erocheleau, line 55:
```suggestion
  get ariaLabel() {
```

Keep it simple

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
```diff
@@ -52,6 +52,14 @@ export default class QuanticCategoryFacetValue extends LightningElement {
...
+    return `Inclusion filter on ${this.facetValue};`;
+  }
+
+  get isPressed() {
+    return this.item.state === 'selected';
```
erocheleau, line 60: Just to be perfectly safe if something doesn't exist. Otherwise your html will have `aria-pressed="undefined"` or the code could throw an error.
```suggestion
    return this.item?.state === 'selected';
```

packages/quantic/force-app/main/default/lwc/quanticDateFacet/quanticDateFacet.html
```diff
@@ -16,24 +16,36 @@
...
-            <lightning-icon class="slds-current-color slds-m-right_xx-small" icon-name="utility:close" alternative-text={altText} size="xx-small"></lightning-icon>
+          <button
+            class="facet__clear-filter slds-button slds-grid slds-grid_vertical-align-center slds-p-top_x-small slds-p-horizontal_x-small"
+            onclick={clearSelections}
+          >
+            <lightning-icon
+              class="slds-current-color slds-m-right_xx-small"
+              icon-name="utility:close"
+              size="xx-small"
+              aria-hidden="true"
```
erocheleau, line 27: 👍

erocheleau (approved):
Just a few comments to resolve, otherwise this looks fine to me!

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
MSLI-Coveo: :white_check_mark:

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
MSLI-Coveo: Using ariaLabelValue

packages/quantic/.prettierrc.js
MSLI-Coveo: This is printWidth is added by me. Reverted the file. The `.prettierignore` file under root folder disables formatting for html files and css files under dir lwc

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
MSLI-Coveo: reverted

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet-actions.ts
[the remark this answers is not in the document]
MSLI-Coveo: `selectFirstValueUsingKeyboard` used for this action.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
MSLI-Coveo: Could be accessed through spacebar after setting role="button".

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet.cypress.ts
```diff
@@ -375,4 +376,17 @@ describe('quantic-numeric-facet', () => {
       });
     });
   });
+
+  describe('with default numeric facet testing accessibility', () => {
+    beforeEach('', () => {
```
lbergeron, line 381:
```suggestion
    beforeEach(() => {
```

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
```diff
@@ -52,6 +52,15 @@ export default class QuanticCategoryFacetValue extends LightningElement {
     return this.item.value;
   }
 
+  get ariaLabelValue() {
+    return `Inclusion filter on ${this.facetValue};`;
```
lbergeron, line 56: Is the trailing semicolon mandatory?

```suggestion
    return `Inclusion filter on ${this.facetValue}`;
```

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
MSLI-Coveo: No, removing it.

packages/quantic/cypress/integration/facets/numeric-facet/numeric-facet.cypress.ts
MSLI-Coveo: empty string removed

PR #2238, opened 2022-07-22
feat(quantic): added accessibility for quanticTab

MSLI-Coveo:
[SVCC-2279](https://coveord.atlassian.net/browse/SVCC-2279)
Added accessibility for quanticTab using div for the styled border when focused;

Before: (Tab could no be focused using key "Tab" )
Screen Shot 2022-07-25 at 9 31 36 AM

After:
Screen Shot 2022-07-29 at 4 38 16 PM
Screen Shot 2022-07-29 at 4 36 59 PM


packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,9 @@
 <template>
   <template if:true={shouldDisplay}>
-    <li class={tabClass} title={label} role="presentation">
-      <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
-    </li>
+    <div tabindex="0" role="tab" onkeydown={onKeyDown}>
+      <li class={tabClass} title={label} role="presentation">
+        <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
+      </li>
+    </div>
```
erocheleau, line 7:
```suggestion
    <li class={tabClass} title={label} role="presentation">
      <button class="slds-button slds-tabs_default__link" role="tab" onclick={select}>{label}</button>
    </li>
```

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
```diff
@@ -103,4 +103,14 @@ export default class QuanticTab extends LightningElement {
...
+
+  /**
+   * @param {KeyboardEvent} evt
+   */
+  onKeyDown(evt) {
+    if(evt.code==="Enter" || evt.code==="Space"){
+          evt.preventDefault()
+          this.tab.select()
+    }
+  }
```
erocheleau, line 115: we wont need this if we use a button

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
erocheleau: Make sure to have something so the button doesn't have a huge `line-height` look in my last PRs to set it correctly.

erocheleau (changes requested):

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,9 @@
 <template>
   <template if:true={shouldDisplay}>
-    <li class={tabClass} title={label} role="presentation">
-      <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
-    </li>
+    <div tabindex="0" role="tab" onkeydown={onKeyDown}>
```
nathanlb, line 3: I'm not sure `role="tab"`  is appropriate here. We're using the term "Tab" describe this and I think the concept is close enough. But given each tab doesn't actually have an associate tab panel that it displays when selected, I think it might be misleading. They're really buttons that apply a filter. 😅

Curious to know other thoughts on this.

erocheleau: I agree, role="tab" is for a control of type tabset that we don't want to be using.
https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html
It was reflected in the assessment that this "looks" like a tabset but isn't really a tabset, so I would feel better if those were just buttons.

erocheleau (changes requested):
Just needs the changes to use button instead and the tests that we discussed and this will be good to go.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
```diff
@@ -103,4 +103,8 @@ export default class QuanticTab extends LightningElement {
   get tabClass() {
     return `slds-tabs_default__item ${this.isActive ? 'slds-is-active' : ''}`;
   }
+
+  get arialLaelText() {
```
lvu285, line 107: small typo "arialLaelText"

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,15 @@
 <template>
   <template if:true={shouldDisplay}>
     <li class={tabClass} title={label} role="presentation">
-      <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
+      <button
+        class="pill__container slds-grid_vertical-align-center slds-p-horizontal_x-small slds-p-vertical_xx-small slds-button slds-var-m-around_xxx-small"
+        onclick={select}
+        aria-label={arialLaelText}
+      >
+        <a class="slds-tabs_default__link" role="tab" onclick={select}
```
lvu285, line 9: so we still keep the "role= tab" here?

lvu285:
should we add the status of the "tab" to know if it's selected/or not?

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: For <a element, it seems attribute "role" here would not have any effect on accessibility behaviour.Is this attribute needed for other features? @erocheleau If, not We should go head and remove attribute "role" from <a.

MSLI-Coveo:
> Just needs the changes to use button instead and the tests that we discussed and this will be good to go.

Buttons added
Tests added

packages/quantic/cypress/integration/tab/tab-selectors.ts
```diff
@@ -4,13 +4,18 @@ export const tabComponent = 'c-quantic-tab';
...
   active: () => CypressSelector;
 }
 
 export const TabSelectors: TabSelector = {
   get: () => cy.get(tabComponent),
 
   tab: () => TabSelectors.get().find('li.slds-tabs_default__item'),
+  button: () =>
+    TabSelectors.get().find(
+      'button.pill__container.slds-grid_vertical-align-center.slds-p-horizontal_x-small.slds-p-vertical_xx-small'
```
nathanlb, line 17: At first glance this selector seems overly complicated. I feel like `.pill__container` would be enough.
otherwise if we change anything about the classes on this component we'll break the tests.

packages/quantic/cypress/integration/tab/tab.cypress.ts
```diff
@@ -98,4 +100,31 @@ describe('quantic-tab', () => {
...
+      loadFromUrlHash({}, `tab=${tabs.all.label}`);
+    });
+
+    it('when SpaceKey is pressed', () => {
+      TabSelectors.button().first().focus().type(' ');
+      Expect.activeTabContains(tabs.all.label);
+    });
+
+    it('when TabKey && SpaceKey is pressed', () => {
+      TabSelectors.button().first().focus().tab().type(' ');
```
nathanlb, line 115: I think most or all of these `.focus().tab().type(' ')` kind of sequences could be added to `tab-actions.ts`

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
MSLI-Coveo: typo fixed

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: changed to button

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
MSLI-Coveo: function onKeyDown removed

packages/quantic/cypress/integration/tab/tab.cypress.ts
```diff
@@ -98,4 +100,31 @@ describe('quantic-tab', () => {
       Expect.activeTabContains(tabs.knowledge.label);
     });
   });
+
+  describe('when using accessibility', () => {
```
nathanlb, line 104: The cypress tests for Quantic are written to avoid reloading the page as much as possible as it slows down test runs significantly. 
To do so we try to use as few `it()` blocks as possible and instead separate tests with `scope()` blocks within `it()`s.
If you look at other tests like the facet tests it'll be clearer.
So unless you're loading the tabs with different options or from a different url hash, it might be possible to package several tests and assertions like that without reloading.

nathanlb (commented):
Nearly there! The cypress setup isn't super intuitive.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
lvu285: I asked because I saw comment from Etienne & Nathan suggested to remove it, as the tab we use here is not really a 'tab', it's more like a filter, but yes, we have to wait for a double confirmation from @erocheleau ?

packages/quantic/cypress/integration/tab/tab-selectors.ts
lvu285: can we use selector based on the tab title or add another attribute for the "test" itself, I don't think it's a good practice for test to keep the selector this long.

packages/quantic/cypress/integration/tab/tab.cypress.ts
lvu285, line 115: I'm curious why we have to add `type('  ')` here ?

lvu285:
I just remember, we might need to check "MORE" button when Tab is in responsive mode too
in that case, please make sure we can navigate through the Tab list when they are in the dropdown

packages/quantic/cypress/integration/tab/tab.cypress.ts
MSLI-Coveo: type('  ') will trigger a space key press, simulating the situation where users use keyboard to trigger the button.

MSLI-Coveo:
> 

Checked in the local server, it seems users could use Tab key to traverse through the "More" dropdown menu once it is open.

MSLI-Coveo:
> I just remember, we might need to check "MORE" button when Tab is in responsive mode too in that case, please make sure we can navigate through the Tab list when they are in the dropdown

Checked in the local server, it seems users could use Tab key to traverse through the "More" dropdown menu once it is open.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,15 @@
 <template>
   <template if:true={shouldDisplay}>
     <li class={tabClass} title={label} role="presentation">
-      <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
+      <button
+        class="slds-tabs_button slds-button"
+        onclick={select}
+        aria-label={arialLabelText}
+      >
+        <a class="slds-tabs_default__link" role="tab"
```
btaillon, line 9: Since the parent is a button, the `tab` role is unnecessary here. I would simply remove the role for the `a` element.

Additionally, tabs seem to be meant to be part of a `tablist` and associated with a `tabpanel` ([doc](https://www.w3.org/TR/wai-aria-1.2/#tab)), which doesn't make much sense for our search pages, so I think it's simpler to use a toggle button.

Lastly, even when overwriting the `role` of an anchor element, I recall some issues being reported in JSUI about on similar components. I would replace the `a` element with a `span` or remove it.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,15 @@
 <template>
   <template if:true={shouldDisplay}>
     <li class={tabClass} title={label} role="presentation">
```
btaillon, line 3: I would move the title of this `li` to the `button`, in case some screen readers try to read it.

Also, if this `li` has a parent `ul`/`ol`, I would not overwrite the role on either of them, I think it makes sense to have a list of tabs. But that's subjective.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,15 @@
...
-      <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
+      <button
+        class="slds-tabs_button slds-button"
+        onclick={select}
+        aria-label={arialLabelText}
```
btaillon, line 7: to represent which tab is currently selected, I would add `aria-pressed="true"`/`aria-pressed="false"`. Beware that `aria-pressed="false"` is not equivalent to omitting `aria-pressed`. The existence of `aria-pressed` with any value indicates that this is a toggle button, not a regular button.

Alternatively, you could replace the buttons with radio buttons ([doc](https://www.w3.org/TR/wai-aria-1.2/#radio)) to indicate that only one can be selected and you can't press an already selected button. However, that would require more changes since radio buttons are expected to support arrow keys and space bar ([doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role#keyboard_interactions)).

btaillon (changes requested):

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
btaillon: Also, I would avoid in general have an `a` without an `href`.

packages/quantic/cypress/integration/tab/tab.cypress.ts
nathanlb: Great improvement!! Thanks!

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
nathanlb: Yeah good point I thought we'd agreed to remove the `role="tab"` but then again maybe I just lost track.

nathanlb (approved):
Nice work! @btaillon has some good points that should be addressed but thanks for the contribution @MSLI-Coveo!

packages/quantic/cypress/integration/tab/tab-actions.ts
```diff
@@ -5,6 +5,19 @@ function tabActions(selector: TabSelector) {
...
+    tabAndSelectNextButton: (value: string) => {
+      selector.button().contains(value).tab().type(' ');
+    },
+    tabAndSelectPreviousButton: (value: string) => {
+      selector.button().contains(value).tab({shift: true}).type(' ');
+    },
+
+    keyPressButton: (value: string, input = ' ') => {
+      selector.button().contains(value).type(input);
+    },
```
erocheleau, line 20: Why do you need this?

packages/quantic/cypress/integration/tab/tab-selectors.ts
```diff
@@ -4,13 +4,15 @@ export const tabComponent = 'c-quantic-tab';
...
   tab: () => CypressSelector;
+  button: () => CypressSelector;
   active: () => CypressSelector;
 }
 
 export const TabSelectors: TabSelector = {
   get: () => cy.get(tabComponent),
 
   tab: () => TabSelectors.get().find('li.slds-tabs_default__item'),
+  button: () => TabSelectors.get().find('button.slds-tabs_button.slds-button'),
```
erocheleau, line 15: Why is this two different things? Aren't tabs just buttons now?
Shouldn't you just update this to be like this:

```suggestion
  tab: () => TabSelectors.get().find('button.slds-tabs_button.slds-button'),
```

packages/quantic/cypress/integration/tab/tab.cypress.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from '../../page-objects/search';
 import {TabExpectations as Expect} from './tab-expectations';
 import {TabActions as Actions} from './tab-actions';
+import {TabSelector, TabSelectors} from './tab-selectors';
```
erocheleau, line 9: See my comment on tab-selectors.ts don't create a new selector for the button, just modify the tab selector to select the tab after your changes.

packages/quantic/cypress/integration/tab/tab.cypress.ts
```diff
@@ -98,4 +99,19 @@ describe('quantic-tab', () => {
       Expect.activeTabContains(tabs.knowledge.label);
     });
   });
+
+  describe('when using accessibility', () => {
+    it('should be accessible to keyboard', () => {
+      loadFromUrlHash({}, `tab=${tabs.case.label}`);
+
+      Actions.keyPressButton(tabs.all.label);
```
erocheleau, line 107:
```suggestion
      Actions.pressSpaceOnTarget(tabs.all.label);
```

erocheleau:
```suggestion
      Actions.pressSpaceOnElement(tabs.all.label);
```

erocheleau: One of those options would be much clearer to create as a specific function.

A function called `keyPressButton` with a default value that presses space is not clear what it does really with its name (why does it need a target) and also nothing is letting the user know that not passing a 2nd parameter means this will press space by default.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
erocheleau: Very useful, thanks @btaillon

erocheleau (changes requested):

packages/quantic/cypress/integration/tab/tab-actions.ts
```diff
@@ -5,6 +5,19 @@ function tabActions(selector: TabSelector) {
...
     },
+    selectButton: (value: string) => {
+      selector.button().contains(value).click();
+    },
+    tabAndSelectNextButton: (value: string) => {
+      selector.button().contains(value).tab().type(' ');
+    },
+    tabAndSelectPreviousButton: (value: string) => {
+      selector.button().contains(value).tab({shift: true}).type(' ');
+    },
```
erocheleau, line 16:
```suggestion
    tabAndSelectPreviousButton: (value: string) => {
      selector.button().contains(value).tab({shift: true}).type('abc');
    },
```

packages/quantic/cypress/integration/tab/tab-actions.ts
```diff
@@ -5,6 +5,19 @@ function tabActions(selector: TabSelector) {
     selectTab: (value: string) => {
       selector.tab().contains(value).click();
     },
+    selectButton: (value: string) => {
+      selector.button().contains(value).click();
+    },
+    tabAndSelectNextButton: (value: string) => {
```
erocheleau, line 11: findTabPressTabPressSpace: (value: string) => {
    selector.button().contains(value).tab().type(' ');
},
findTabPressShiftTabPressSpace: (value: string) => {
    selector.button().contains(value).tab({shift: true}).type(' ');
}

packages/quantic/cypress/integration/tab/tab-actions.ts
```diff
@@ -5,6 +5,19 @@ function tabActions(selector: TabSelector) {
     selectTab: (value: string) => {
```
erocheleau, line 5:
```
selectTab: (value: string, space: boolean = false) => {
    if(!space) {
        selector.tab().contains(value).click();
    } else (
        selector.tab().contains(value).type(' ');
    )
}
```

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: role=tab removed

packages/quantic/cypress/integration/tab/tab.cypress.ts
MSLI-Coveo: moved to .action.ts file

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: switched `li`,`ui`/`ol` to `div`

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: role="tab" removed

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: `aria-pressed` added. And thanks for the info!!!@btaillon

packages/quantic/cypress/integration/tab/tab-actions.ts
```diff
@@ -5,6 +5,19 @@ function tabActions(selector: TabSelector) {
...
       selector.tab().contains(value).click();
     },
+    selectButton: (value: string) => {
+      selector.button().contains(value).click();
+    },
```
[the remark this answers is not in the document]
MSLI-Coveo: `selectButton` removed. `selectTab` modified.

packages/quantic/cypress/integration/tab/tab.cypress.ts
MSLI-Coveo: selectors modified

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,13 @@
 <template>
   <template if:true={shouldDisplay}>
-    <li class={tabClass} title={label} role="presentation">
-      <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
-    </li>
+    <button
+      class={tabClass}
+      onclick={select}
+      aria-label={arialLabelText}
+      aria-pressed={isActive}
```
btaillon, line 7: Make sure to test that when `isActive` is `true`, `aria-pressed` is set to the string `"true"`, and when `isActive` is `false`, `aria-pressed` is set to the string `"false"`. We've had issues in Atomic with Stencil simply adding/removing an empty attribute based on true/false.

btaillon (approved):

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: variable ariaPressed added to ensure attribute aria-pressed receives a string.

packages/quantic/cypress/integration/tab/tab-actions.ts
```diff
@@ -2,8 +2,16 @@ import {TabSelector, TabSelectors} from './tab-selectors';
...
   return {
-    selectTab: (value: string) => {
-      selector.tab().contains(value).click();
+    selectTab: (value: string, keyPress?: string) => {
+      if (!keyPress) {
+        selector.tab().contains(value).click();
+      } else selector.tab().contains(value).type(keyPress);
+    },
+    findTabPressTabPressSpace: (value: string) => {
+      selector.tab().contains(value).tab().type(' ');
```
erocheleau, line 11:
```suggestion
      // .type(' ') is equivalent to pressing the spacebar.
      selector.tab().contains(value).tab().type(' ');
```

packages/quantic/cypress/integration/tab/tab.cypress.ts
erocheleau, line 9: You are not using these in this file. You need to remove the unused imports.

packages/quantic/cypress/integration/tab/tab.cypress.ts
```diff
@@ -98,4 +99,19 @@ describe('quantic-tab', () => {
       Expect.activeTabContains(tabs.knowledge.label);
     });
   });
+
+  describe('when using accessibility', () => {
+    it('should be accessible to keyboard', () => {
+      loadFromUrlHash({}, `tab=${tabs.case.label}`);
+
+      Actions.selectTab(tabs.all.label, ' ');
```
erocheleau, line 107:
```suggestion
      // Select tab using the spacebar
      Actions.selectTab(tabs.all.label, ' ');
```

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.css
```diff
@@ -0,0 +1,5 @@
+.slds-tabs_button{
```
erocheleau, line 1: Are you overriding a slds default class?
You should not override a slds class,
and if it's not a slds class, you should be using a different nomenclature for your classes
Example: `.tab_button`

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
```diff
@@ -101,6 +101,15 @@ export default class QuanticTab extends LightningElement {
   }
 
   get tabClass() {
-    return `slds-tabs_default__item ${this.isActive ? 'slds-is-active' : ''}`;
+    return `slds-tabs_default__item slds-tabs_button slds-button ${
+      this.isActive ? 'slds-is-active' : ''
+    }`;
+  }
+  get isPressed() {
```
erocheleau, line 108: Line break before a new function please.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
```diff
@@ -101,6 +101,15 @@ export default class QuanticTab extends LightningElement {
...
+    return `${this.isActive}`;
+  }
+
+  get arialLabelText() {
+    return `${this.label}`;
   }
```
erocheleau, line 114: I think I would remove that function for now and just use `this.label` in the HTML, see my other comment in the html file

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,13 @@
 <template>
   <template if:true={shouldDisplay}>
-    <li class={tabClass} title={label} role="presentation">
-      <a class="slds-tabs_default__link" role="tab" onclick={select}>{label}</a>
-    </li>
+    <button
+      class={tabClass}
+      onclick={select}
+      aria-label={arialLabelText}
```
erocheleau, line 6:
```suggestion
      aria-label={label}
```

packages/quantic/cypress/integration/tab-bar/tab-bar-selectors.ts
```diff
@@ -18,9 +18,7 @@ export const TabBarSelectors: TabBarSelector = {
 
   allTabs: () => TabBarSelectors.get().find(tabComponent),
   activeTab: () =>
-    TabBarSelectors.get().find(
-      'li.slds-tabs_default__item.slds-is-active a[role="tab"]'
-    ),
+    TabBarSelectors.get().find('button.slds-tabs_default__item.slds-is-active'),
```
erocheleau, line 21: Good job thinking about modifying that selector too!

erocheleau (changes requested):
Small changes and then good to go!

packages/quantic/cypress/integration/tab/tab.cypress.ts
MSLI-Coveo: comments added

packages/quantic/cypress/integration/tab/tab.cypress.ts
MSLI-Coveo: unused imports removed

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
MSLI-Coveo: arialLabelText removed

packages/quantic/cypress/integration/tab/tab-actions.ts
```diff
@@ -2,8 +2,16 @@ import {TabSelector, TabSelectors} from './tab-selectors';
 
 function tabActions(selector: TabSelector) {
   return {
-    selectTab: (value: string) => {
-      selector.tab().contains(value).click();
+    selectTab: (value: string, keyPress?: string) => {
+      if (!keyPress) {
+        selector.tab().contains(value).click();
+      } else selector.tab().contains(value).type(keyPress);
```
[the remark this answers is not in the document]
MSLI-Coveo: condition changed

packages/quantic/cypress/integration/tab/tab-actions.ts
MSLI-Coveo: comments added

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
MSLI-Coveo: format changed

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.css
```diff
@@ -0,0 +1,5 @@
+.tab_button{
```
erocheleau, line 1: Did this fix the tab bar collapse problem? Showing 1 less tab than what was expected?

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
```diff
@@ -1,7 +1,13 @@
...
+      aria-pressed={isPressed}
```
erocheleau, line 7: Ah well you can just use this then:

```suggestion
      aria-pressed={isActive}
```

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
```diff
@@ -101,6 +101,12 @@ export default class QuanticTab extends LightningElement {
...
 
   get tabClass() {
-    return `slds-tabs_default__item ${this.isActive ? 'slds-is-active' : ''}`;
+    return `slds-button slds-tabs_default__item tab_button ${
+      this.isActive ? 'slds-is-active' : ''
+    }`;
+  }
+
+  get isPressed() {
```
erocheleau, line 109: I don't think you need that function anymore.

erocheleau (commented):
Almost there! 
Make the few final changes and this is good to go!

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.css
MSLI-Coveo: Yes, the tabs now do collapse accordingly when width is decreased.

MSLI-Coveo: Yes, tabs now collapse accordingly when width is decreased

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.js
MSLI-Coveo: Comments added for function isPressed.

packages/quantic/force-app/main/default/lwc/quanticTab/quanticTab.html
MSLI-Coveo: comments added for isPressed

erocheleau (approved):

PR #2239, opened 2022-07-22
feat(quantic): update Quantic components to work with different engines

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2353

This PR updates the Quantic components that are reused in more than one use case to use the Headless bundle that matches the interface. In other words, a `QuanticResultList` declared inside a `QuanticSearchInterface` will use the loaded Search Headless bundle while the same `QuanticResultList` declared inside a `QuanticInsightInterface` will use the Insight Headless bundle.

In most cases, the change is simply to update the component initialization to:
1. store a reference on the headless bundle using the `getHeadlessBundle(engineId)` function
2. update direct references to Headless bundles to use the dynamic reference instead (i.e., `CoveoHeadless.buildSomething(engine)` --> `this.headless.buildSomething(engine)`)

This PR depends on the other Headless controllers PRs as the controller functions have to be exported.
We'll also need to make sure these symbols are exported:

* `buildCriterionExpression`
* `buildRelevanceSortCriterion`
* `buildDateSortCriterion`
* `deserializeRelativeDate`
* `loadNumericFacetSetActions`
* `loadRecentResultsActions`
* `SortOrder`

### DIFF
--- a/packages/headless/src/case-assist.index.ts
+++ b/packages/headless/src/case-assist.index.ts
@@ -78,6 +78,8 @@ export type {
 } from './controllers/document-suggestion-list/headless-document-suggestion-list';
 export {buildDocumentSuggestionList} from './controllers/document-suggestion-list/headless-document-suggestion-list';
 
+export {buildCaseAssistInteractiveResult as buildInteractiveResult} from './controllers/document-suggestion-list/case-assist-headless-interactive-result';
+
 /**
  * @deprecated
  * Types exported for backward compatibility only.
--- a/packages/headless/src/case-assist.index.ts
+++ b/packages/headless/src/case-assist.index.ts
@@ -64,6 +64,7 @@ export type {
   CaseAssistQuickviewOptions,
   CaseAssistQuickviewProps,
 } from './controllers/quickview/case-assist-headless-quickview';
+export {buildCaseAssistQuickview as buildQuickview} from './controllers/quickview/case-assist-headless-quickview';
 export {buildCaseAssistQuickview} from './controllers/quickview/case-assist-headless-quickview';
 
 export type {Result} from './api/search/search/result';
--- a/packages/quantic/force-app/main/default/lwc/quanticBreadcrumbManager/quanticBreadcrumbManager.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticBreadcrumbManager/quanticBreadcrumbManager.js
@@ -1,8 +1,4 @@
-import {
-  LightningElement,
-  track,
-  api
-} from 'lwc';
+import {LightningElement, track, api} from 'lwc';
 import {
   registerComponentForInit,
   initializeWithHeadless,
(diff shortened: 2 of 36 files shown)


packages/headless/src/case-assist.index.ts
```diff
@@ -78,6 +79,8 @@ export type {
 } from './controllers/document-suggestion-list/headless-document-suggestion-list';
 export {buildDocumentSuggestionList} from './controllers/document-suggestion-list/headless-document-suggestion-list';
 
+export {buildCaseAssistInteractiveResult as buildInteractiveResult} from './controllers/document-suggestion-list/case-assist-headless-interactive-result';
```
nathanlb, line 82: This is a breaking change is it not? We would need to also export them under the current names.

ThibodeauJF (approved):

packages/headless/src/case-assist.index.ts
```diff
@@ -84,6 +84,7 @@ export type {
   CaseAssistQuickviewOptions,
   CaseAssistQuickviewProps,
 } from './controllers/quickview/case-assist-headless-quickview';
+export {buildCaseAssistQuickview as buildQuickview} from './controllers/quickview/case-assist-headless-quickview';
```
ThibodeauJF, line 87: why twice? (line 88 too)

nathanlb: For backwards compatibility

packages/quantic/force-app/main/default/lwc/quanticBreadcrumbManager/quanticBreadcrumbManager.js
```diff
@@ -90,18 +87,23 @@ export default class QuanticBreadcrumbManager extends LightningElement {
    */
   initialize = (engine) => {
     this.breadcrumbManager = CoveoHeadless.buildBreadcrumbManager(engine);
```
mmitiche, line 89: We should use `getHeadlessBundle` here as well instead of `CoveoHeadless` directly.

mmitiche (changes requested):
The QuanticBreadcrumbManager is using `CoveoHeadless `directly, it should use `getHeadlessBundle()` to get the right bundle instead.

packages/quantic/force-app/main/default/lwc/quanticBreadcrumbManager/quanticBreadcrumbManager.js
nathanlb: done

mmitiche (commented):
In `quanticTimeframeFacet/quanticTimeframeFacet.js` at line 250 and 258 we should also have `this.headless.buildDateRange()` instead of `CoveoHeadless.buildDateRange()`
Also in lines 616 and 622.

nathanlb:
> In `quanticTimeframeFacet/quanticTimeframeFacet.js` at line 250 and 258 we should also have `this.headless.buildDateRange()` instead of `CoveoHeadless.buildDateRange()` Also in lines 616 and 622.

And L637 & L643 😛.
On in

mmitiche (approved):

PR #2242, opened 2022-07-25
fix(atomic): use custom "closest" method for initializeBindings, resultContext

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1873

### DIFF
--- a/packages/atomic/src/pages/examples/custom.html
+++ b/packages/atomic/src/pages/examples/custom.html
@@ -60,13 +60,20 @@
       import {resultContext} from '/build/index.esm.js';
 
       class CustomTemplateComponent extends HTMLElement {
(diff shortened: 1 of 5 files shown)


packages/atomic/src/pages/examples/custom.html
```diff
@@ -60,13 +60,20 @@
       import {resultContext} from '/build/index.esm.js';
 
       class CustomTemplateComponent extends HTMLElement {
+        initialized = false;
         connectedCallback() {
```
btaillon, line 64: I'm wondering if perhaps we should add a `disconnectedCallback` to clear listeners. This would probably be better demonstrated in the non-template custom component.

btaillon (approved):

PR #2251, opened 2022-07-26
ci: add [skip ci] on bump commits

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1864

### DIFF
--- a/.github/workflows/masterbot.yml
+++ b/.github/workflows/masterbot.yml
@@ -4,12 +4,6 @@ on:
     branches:
       - master
 jobs:
-  allow-workflow:
-    name: 'Allow workflow'
-    if: "!startsWith(github.event.head_commit.message, '[Version Bump]')"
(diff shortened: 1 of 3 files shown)


louis-bompart (approved):

btaillon (commented):
Not sure how I feel about this 🤔  I like when it's explicit at the top of a workflow that it won't run on a version bump. Plus, seeing `[skip ci]` would give me the impression that Jenkins skips it too.

louis-bompart:
> Not sure how I feel about this 🤔  I like when it's explicit at the top of a workflow that it won't run on a version bump. Plus, seeing `[skip ci]` would give me the impression that Jenkins skips it too.

I understand your feeling, tho the issue I see with the "top of the workflow" method, is that you have to start a workflow. This cost resources.
The skip ci tag allow us to not even start spending

olamothe (approved):

PR #2215, opened 2022-07-13
feat(headless): Refactor core and insight controllers

aelfaleh:
The idea is to keep consistency for quantic components when we charge the insight engine.
For exemple:
we have a quantic component that need to be initialized `quanticSearchBox`, first of all we load `CoveoHeadless` to build the component. To do that we change how to load `CoveoHeadless` depend on use case (insight panel for this case)
So instead of creating a new quantic component for Insight Panel we just reuse the existing one.
to keep this import part still valid we need to export them with same name:

```
/** @typedef {import("coveo").SearchBoxState} SearchBoxState */
/** @typedef {import("coveo").SearchBox} SearchBox */
```

initializing function too:
```
/**
   * @param {SearchEngine} engine
   */
  initialize = (engine) => {
    this.searchBox = CoveoHeadless.buildSearchBox(engine, {
      options: {
        numberOfSuggestions: Number(this.numberOfSuggestions),
        highlightOptions: {
          notMatchDelimiters: {
            open: '<b>',
            close: '</b>',
          },
        },
      },
    });
    this.unsubscribe = this.searchBox.subscribe(() => this.updateState());
  }
```

PR #2254, opened 2022-07-26
fix(headless): The state of the category facet's searchbox added to the state returned by the InsightCategoryFacet controller

mmitiche:
### Problem:
- The state returned by the InsightCategoryFacet Controller was missing the state of the category facet's searchbox.

### Solution:
- The piece of state  representing the state of the category facet's searchbox is now added.

### Question:
Before adding the state getter to the HeadlessInsightCategoryFacet controller, a weird behaviour was occurring:

The QuanticCategoryFacet component subscribes to changes occurring on the state of the category facet, when a change occurs a method named `updateState` is called.
The problem is that there is some changes in the state that doesn't trigger the call of `updateState`. For example after executing a query, the QuanticCategoryFacet component didn't get notified about the change in the state which was attributing new values to this facet.

Adding the following to the returned values of the HeadlessInsightCategoryFacet controller magically solved the issue 🧐:
```
    get state() {
      return {
        ...coreController.state,
        facetSearch: facetSearch.state,
      };
    },
```

Please reach out to me if you have any idea about the reasons behind this behaviour, I can reproduce the issue and give more details, Thank you!

### DIFF
--- a/packages/headless/src/controllers/insight/facets/category-facet/headless-insight-category-facet.ts
+++ b/packages/headless/src/controllers/insight/facets/category-facet/headless-insight-category-facet.ts
@@ -2,6 +2,7 @@ import {
   buildCoreCategoryFacet,
   CategoryFacetProps,
   CategoryFacetSearch,
+  CategoryFacetSearchState,
   CategoryFacetValue,
   CoreCategoryFacet,
   CoreCategoryFacetState,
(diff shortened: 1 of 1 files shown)


aelfaleh:
This bug was fixed in this [PR](https://github.com/coveo/ui-kit/pull/2215)

ThibodeauJF (approved):
Your solution is good, we use the same pattern in multiple controllers.

The facet is composed by 2 separate core controllers: the core category facet & the search.

Instead of having to listen to 2 states: facet.state & facet.facetSearch.state, the facet search state is extracted on line 76, but wasn't added back in the "state" property, which you correctly did in this PR.

nathanlb:
So FYI the insight API currently doesn't have a proxy for facet search. So the insight panel facets need to disable the search option. accessing the state should be no-op as i believe is the case already for the regular insight-facet.

PR #2257, opened 2022-07-27
fix(atomic): pass rendering function down to instant result atomic-re…

liviarett:
…sult

https://coveord.atlassian.net/browse/KIT-1885

### DIFF
--- a/packages/atomic-react/README.md
+++ b/packages/atomic-react/README.md
@@ -43,9 +43,9 @@ Coupled with the [`<template>`](https://developer.mozilla.org/en-US/docs/Web/HTM
 
 However, this can be limiting and awkward to use in a React project using JSX.
 
-For every element that use this interface, Atomic React exposes a wrapper - `AtomicResultList`, `AtomicFoldedResultList` and `AtomicSearchBoxInstantResults`. These components require a `template` property that can be used in a more straightforward manner when coupled with JSX.
+For every kind of search interface element, Atomic React exposes a wrapper: `AtomicResultList`, `AtomicFoldedResultList`, or `AtomicSearchBoxInstantResults`. These wrappers require a `template` property that can be used in a more straightforward manner when coupled with JSX.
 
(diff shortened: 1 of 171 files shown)


packages/atomic-react/README.md
```diff
@@ -43,9 +43,9 @@ Coupled with the [`<template>`](https://developer.mozilla.org/en-US/docs/Web/HTM
 
 However, this can be limiting and awkward to use in a React project using JSX.
 
-Atomic React exposes an `AtomicResultList` component with a `template` property that can be used in a more straightforward manner when coupled with JSX.
+For every element that use this interface, Atomic React exposes a wrapper - `AtomicResultList`, `AtomicFoldedResultList` and `AtomicSearchBoxInstantResults`. These components require a `template` property that can be used in a more straightforward manner when coupled with JSX.
 
-The `template` property accepts a function with a `Result` parameter, which can then be used to conditionally render different templates based on properties and fields available in result items.
+The `template` property accepts a function with a `Result` parameter (`FoldedResult` if using `AtomicFoldedResultList`), which can then be used to conditionally render different templates based on properties and fields available in result items.
```
jpmarceau, line 48:
```suggestion
The `template` property accepts a function with either a `Result` parameter in the case of the `AtomicResultList` and `AtomicSearchBoxInstantResults` wrappers, or a `FoldedResult` parameter in the case of the `AtomicFoldedResultList` wrapper. Use those parameters to conditionally render different templates based on properties and fields available in result items.
```
(assuming I understood correctly)

packages/atomic-react/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "build": "npm run clean && npm run compile && npm run copy:assets",
-    "dev": "npm link ../samples/headless-react/node_modules/react",
```
olamothe, line 13: Why change the script name ? 🤔 

Might affect some other scripts in the repo.

olamothe (approved):

packages/atomic-react/package.json
ThibodeauJF: It's cleared as to what it does. I don't think it will affect other scripts, this is called when you run `dev` in the atomic-react samples

ThibodeauJF (approved):

PR #2261, opened 2022-07-27
feat(quantic): Remove usage of header elements

erocheleau:
https://coveord.atlassian.net/browse/SVCC-2286

Using headers is probably a bad idea for two reasons.
1. right now, for a screen reader that reads the ui, it normally ties `<h3>` elements with the previous `<h2>` element. The problem is that in quantic, facet titles are using `<h2>` and the quantic result link of the templates are using `<h3>`. So semantically it looks like the last facet title, is the title of all the result links.
2. It is very possible that customers already use a hierarchy of headers in their page, we can't assume we are the only ones using `<h2>` and `<h3>`.

Before:
image

After:
image

* So I removed all the h2, h3 and replaced by div, and the styling has not changed. 
* I also fixed the cypress selectors.

### DIFF
--- a/packages/quantic/cypress/integration/facets/category-facet/category-facet-selectors.ts
+++ b/packages/quantic/cypress/integration/facets/category-facet/category-facet-selectors.ts
@@ -26,7 +26,7 @@ export type AllFacetSelectors = BaseFacetSelector &
 export const CategoryFacetSelectors: AllFacetSelectors = {
   get: () => cy.get(categoryFacetComponent),
 
-  label: () => CategoryFacetSelectors.get().find('header h2 > span'),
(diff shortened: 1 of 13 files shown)


nathanlb (approved):
I'm surprised that replacing h3 with div results in the same font styling but looks legit to me. The more we learn about a11y the more my palm tends to reach my face. 🤦 
GG

mmitiche (approved):
Looking good 👌

erocheleau:
I think h3 inherently with the lightning design system has been pretty much removed of any default styling. slds has utility classes to use for text sizes and etc. instead of the h elements I think that's why.

PR #2263, opened 2022-07-28
fix(headless): Active tab added to the insight query request

mmitiche:
### Problem:
The active tab information was missing in the insight query request sent to the customer service API.

### Solution: 
The active tab information is now sent in the constant query parameter of the  insight query request.

### DIFF
--- a/packages/headless/src/features/insight-search/insight-search-actions.ts
+++ b/packages/headless/src/features/insight-search/insight-search-actions.ts
@@ -23,6 +23,7 @@ import {
   PaginationSection,
   QuerySection,
   SearchSection,
+  TabSection,
 } from '../../state/state-sections';
 import {requiredNonEmptyString} from '../../utils/validate-payload';
 import {InsightAction} from '../analytics/analytics-utils';
@@ -64,7 +65,8 @@ export type StateNeededByExecuteSearch = ConfigurationSection &
       DateFacetSection &
       CategoryFacetSection &
       PaginationSection &
-      DidYouMeanSection
+      DidYouMeanSection &
+      TabSection
   >;
 
 const fetchFromAPI = async (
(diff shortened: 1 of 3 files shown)


aelfaleh (approved):

nathanlb (approved):
LGTM!

packages/headless/src/features/insight-search/insight-search-actions.ts
```diff
@@ -382,6 +386,15 @@ function getFacetRequests<T extends AnyFacetRequest>(
   return Object.keys(requests).map((id) => requests[id]);
 }
 
+function buildConstantQuery(state: StateNeededByExecuteSearch) {
+  const activeTab = Object.values(state.tabSet || {}).find(
+    (tab) => tab.isActive
+  );
+  const tabExpression = activeTab?.expression.trim() || '';
+
+  return tabExpression;
```
erocheleau, line 395: So activeTab.expression will always be a string and never undefined/null right?

erocheleau (approved):

packages/headless/src/features/insight-search/insight-search-actions.ts
```diff
@@ -303,6 +306,7 @@ const buildInsightSearchRequest = (
       firstResult: state.pagination.firstResult,
       numberOfResults: state.pagination.numberOfResults,
     }),
+    ...(cq && {cq}),
```
erocheleau, line 309: BTW I don't know how necessary it is, but in the current search request to the search api, we do send the 
`tab: tabName` as part of the payload. 
If you are including the `cq` of the tab maybe it makes sense to also include that?

packages/headless/src/features/insight-search/insight-search-actions.ts
mmitiche: Yes absolutely👌

ThibodeauJF (approved):

packages/headless/src/features/insight-search/insight-search-actions.ts
mmitiche: Good point!  I see that the tab parameter is used mainly for analytics purposes since it represents the originLevel2.
This will need to be added as well, I will create a Jira for it and address it in a separate PR just to avoid blocking the UI-KIT release any longer.

erocheleau: Good idea 👍

PR #2265, opened 2022-07-28
feat(headless): do not review

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1883


packages/headless/src/api/preprocess-request.ts
```diff
@@ -1,14 +1,21 @@
...
 
 export interface PlatformRequestOptions extends RequestInit {
   url: string;
 }
 
 export type PlatformClientOrigin = AnalyticsClientOrigin | 'searchApiFetch';
 
+export interface RequestMetadata {
+  clientMethod: SearchApiClientMethod;
+  origin?: SearchOrigin;
```
ThibodeauJF, line 12: A lot more can be added from different clients & origin.
The idea is that `clientMethod` is whatever method is called on the Api client, then `origin` is to differentiate the origin of this method when further differentiation is required

packages/headless/src/api/search/search-metadata.ts
```diff
@@ -0,0 +1,15 @@
+export type SearchApiClientMethod =
```
ThibodeauJF, line 1: We can have the same kind of file for all clients

packages/headless/src/api/preprocess-request.ts
```diff
@@ -1,14 +1,37 @@
...
+import {SearchApiMethod, SearchOrigin} from './search/search-metadata';
 
 export interface PlatformRequestOptions extends RequestInit {
   url: string;
 }
 
+// TODO: V2 add more types for each different client e.g. 'insightApiFetch' or 'commerceApiFetch'
 export type PlatformClientOrigin = AnalyticsClientOrigin | 'searchApiFetch';
 
+export interface RequestMetadata {
```
ThibodeauJF, line 11: A lot more can be added from different clients & origin.
The idea is that clientMethod is whatever method is called on the Api client, then origin is to differentiate the origin of this method when further differentiation is required

packages/headless/src/api/search/search-metadata.ts
```diff
@@ -0,0 +1,15 @@
+export type SearchApiMethod =
```
ThibodeauJF, line 1: We can have the same kind of file for all clients

packages/headless/src/features/fields/fields-actions.ts
```diff
@@ -29,7 +29,7 @@ export const fetchFieldsDescription = createAsyncThunk<
   FieldDescription[],
   void,
   AsyncThunkSearchOptions<ConfigurationSection>
->('fields/description', async (_, {extra, getState, rejectWithValue}) => {
```
btaillon, line 32: I think this is breaking if somebody uses a custom reducer. However,  don't know what the chances are of that, and I do prefer the new names 🤔

btaillon (approved):

packages/headless/src/features/fields/fields-actions.ts
ThibodeauJF: I started by using these but it got confusing. I didn't revert cause in the end I like the new names either

packages/headless/src/api/search/search-api-client.ts
```diff
@@ -83,6 +85,7 @@ export class SearchAPIClient implements FacetSearchAPIClient {
     const response = await PlatformClient.call({
       ...baseSearchRequest(req, 'POST', 'application/json', '/plan'),
       requestParams: pickNonBaseParams(req),
+      requestMetadata: {method: 'plan'},
```
olamothe, line 88: So since those name now become part of the API, and means they should be backward compatible, there should be some kind of test that verifies that the contract is never broken.

packages/headless/src/features/fields/fields-actions.ts
olamothe: It's fine to change this.

olamothe (approved):

ThibodeauJF:
Added some tests! Re-requesting review only for those 😄

btaillon (approved):

packages/headless/src/features/search/search-actions.ts
```diff
@@ -224,7 +230,8 @@ export const fetchPage = createAsyncThunk<
     const fetched = await fetchFromAPI(
       extra.apiClient,
       state,
-      await buildSearchRequest(state)
+      await buildSearchRequest(state),
+      'mainSearch'
```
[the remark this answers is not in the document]
ThibodeauJF: It's also affecting the same piece of state: the result list. It kind of originates from the same components VS the facet values, the instant results or the folding collection which are different result lists in a sense. That's the way I saw "origin" When wanting to preprocess the main search results call customer will surely do the same thing for executeSearch, fetchPage or fetchMore, If we don't bundle them together they surely won't realize executeSearch, fetchPage or fetchMore are linked in a sense.

We could add a "context" meta value in the future for extra-specificity, no need for breaking changes.

PR #2268, opened 2022-07-29
feat(atomic): add insight refine toggle and insight refine modal

liviarett:
https://user-images.githubusercontent.com/30511433/181800739-c13d1a43-1d05-4643-98e7-92e53437909d.mp4

Actual facets will follow after!!


packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
```diff
@@ -0,0 +1,177 @@
+import {Component, h, State, Prop, Element, Watch, Host} from '@stencil/core';
```
ThibodeauJF, line 1: There's a lot of code duplication with the atomic-refine-modal that we could be able to abstract

ThibodeauJF: (In a separate PR as this one is quite big already 😄 )

ThibodeauJF (approved):

packages/atomic/src/utils/utils.ts
```diff
@@ -155,3 +155,18 @@ export function getFocusedElement(
 export async function defer() {
   return new Promise<void>((resolve) => setTimeout(resolve));
 }
+
+// https://terodox.tech/how-to-tell-if-an-element-is-in-the-dom-including-the-shadow-dom/
+export function isInDocument(element: Node) {
```
btaillon, line 160: In my PR, I'm gonna have to extract `getParent` and `contains` (renamed `isAncestorOf`) from `atomic-focus-trap`. Perhaps you could extract the same methods in your PR and re-use `isAncestorOf`/`contains`.

Alternatively, I'm curious if `closest(element, 'document')` (from utils) would be equivalent to what you're doing.

liviarett: I only moved this method to this file, didn't implement anything! could we look into it in a separate PR? it's so big already 😬

btaillon (approved):

PR #2269, opened 2022-07-29
fix(headless): provide runtimeEnvironment for all Analytics Clients

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1891

### DIFF
--- a/packages/headless/src/features/configuration/analytics-params.ts
+++ b/packages/headless/src/features/configuration/analytics-params.ts
@@ -1,0 +1,20 @@
+import {AnalyticsParam} from '../../api/search/search-api-params';
+import {getPageID, getVisitorID} from '../../api/analytics/search-analytics';
+import {AnalyticsState} from './configuration-state';
+
+export const fromAnalyticsStateToAnalyticsParams = async (
+  s: AnalyticsState
+): Promise<AnalyticsParam> => {
+  return {
+    analytics: {
+      clientId: await getVisitorID(s),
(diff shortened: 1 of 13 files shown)


packages/headless/src/features/configuration/analytics-params.ts
```diff
@@ -0,0 +1,20 @@
+import {AnalyticsParam} from '../../api/search/search-api-params';
+import {getPageID, getVisitorID} from '../../api/analytics/search-analytics';
+import {AnalyticsState} from './configuration-state';
+
+export const fromAnalyticsStateToAnalyticsParams = async (
```
ThibodeauJF, line 5: moved to prevent circular imports

olamothe (commented):
How will this work in practice, when someone wants to provide storage for React native, for example ?

ThibodeauJF:
@olamothe 
Runtime is configurable in the engine. It's passed down to Coveo Analytics where users can implement their own way of retrieving the visitor id. For React native some work has been done but it doesn't seem to work as advertised ([in the warning](https://github.com/coveo/coveo.analytics.js/blob/master/src/react-native/react-native-utils.ts)) we might have to improve the solution in that library too.

What was clear for now is that the runtime defined in the config has to be passed to all Analytic clients we create since its the way the visitor id will be properly set & retrieved.

olamothe (approved):

btaillon (approved):

PR #2272, opened 2022-08-01
feat(atomic,headless): add atomic insight facet component

liviarett:
https://user-images.githubusercontent.com/30511433/182188262-21cf69fb-cb29-41be-b51d-be7cbe08c9ea.mp4


packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * The number of values to request for this facet.
+   * Also determines the number of additional values to request each time more values are shown.
+   */
+  @Prop({reflect: true}) public numberOfValues = 8;
+  /**
+   * Whether this facet should contain a search box.
+   * When "true", the search is only enabled when more facet values are available.
+   */
+  @Prop({reflect: true}) public withSearch = true;
```
olamothe, line 113: From what I remember from @lvu285 comment yesterday we could remove the facet search feature (ie: always set it to false in the headless controller)

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -0,0 +1,277 @@
...
+   */
+  @Prop({reflect: true}) public filterFacetCount = true;
+  /**
+   * The maximum number of results to scan in the index to ensure that the facet lists all potential facet values.
+   * Note: A high injectionDepth may negatively impact the facet request performance.
+   * Minimum: `0`
+   * Default: `1000`
+   */
+  @Prop() public injectionDepth = 1000;
+  // @Prop() public customSort?: string; TODO: KIT-753 Add customSort option for facet
```
olamothe, line 145: I would remove the TODO for now.

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -0,0 +1,277 @@
...
+   * ></atomic-insight-facet>
+   *
+   * <!-- To show the facet when value "doc" is selected in the facet with id "abc": -->
+   * <atomic-insight-facet
+   *   depends-on-abc="doc"
+   *   ...
+   * ></atomic-insight-facet>
+   * ```
+   */
+  @MapProp() @Prop() public dependsOn: Record<string, string> = {};
```
olamothe, line 166: I think this prop could also be removed for simplicity's sake.

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -0,0 +1,277 @@
...
+   *
+   * ```html
+   * <atomic-insight-facet field="objecttype" allowed-values="Contact,Account,File"></div>
+   * ```
+   *
+   * The maximum amount of allowed values is 25.
+   *
+   * Default value is `undefined`, and the facet uses all available values for its `field` in the current result set.
+   */
+  @Prop() public allowedValues?: string;
```
olamothe, line 188: Same, could be removed I think.

olamothe (approved):

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
ThibodeauJF: 👍  For the parts in the doc too

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -0,0 +1,277 @@
...
+  public searchStatusState!: InsightSearchStatusState;
+  @State() public error!: Error;
+
+  /**
+   * Specifies a unique identifier for the facet.
+   */
+  @Prop({mutable: true, reflect: true}) public facetId?: string;
+  /**
+   * The non-localized label for the facet.
+   * Used in the `atomic-breadbox` component through the bindings store.
```
ThibodeauJF, line 97:
```suggestion
   * Used in the `atomic-insight-breadbox` component through the bindings store.
```

I think?

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
ThibodeauJF: Why the API doesn't support it?

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -0,0 +1,277 @@
...
+    });
+
+    this.searchStatus = buildInsightSearchStatus(this.bindings.engine);
+  }
+
+  public disconnectedCallback() {
+    this.facetCommon.disconnectedCallback();
+  }
+
+  public componentShouldUpdate(
```
ThibodeauJF, line 241: Can be removed

ThibodeauJF (approved):

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
lvu285: they have not yet supported the search for standard facet for 1st release, maybe the future yes. So if it's too complicated to make it works now, we can remove it, If you can make it works without putting much effort on it, then keep it

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -0,0 +1,277 @@
...
+          isCollapsed={this.isCollapsed}
+        ></FacetPlaceholder>
+      );
+    }
+    return this.facetCommon.render({
+      hasError: this.searchStatusState.hasError,
+      firstSearchExecuted: this.searchStatusState.firstSearchExecuted,
+      isCollapsed: this.isCollapsed,
+      numberOfValues: this.numberOfValues,
+      headerFocus: this.headerFocus,
```
btaillon, line 270: I'd prefer if values that aren't meant to be changed like *Focus and perhaps even headingLevel were passed at initialization instead.

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
liviarett: thank you!

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
liviarett: the problem with the focus is that because of the decorator, it happens after initialization, so it needs to go here!
but the hading level can go to initialization, I'll update

PR #2275, opened 2022-08-01
feat(atomic): keep suggestion active background when navigating in instant result

olamothe:
When navigating in instant result, we received the feedback that it was hard to see for which suggestion the current instant results are shown.

This PR makes it so that when you navigate to the right panel, it keeps the "active" background (`bg-neutral-light`) on the linked suggestion.

https://coveord.atlassian.net/browse/KIT-1850


packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -617,6 +621,30 @@ export class AtomicSearchBox {
...
 
+  private getSuggestionBackgroundClass(
+    isSelected: boolean,
+    side: 'left' | 'right',
+    item: SearchBoxSuggestionElement
+  ) {
+    if (isSelected) {
+      return 'bg-neutral-light';
+    }
+    if (side === 'left' && item.linkedBy) {
```
btaillon, line 632: I haven't tested it, but is it currently possible to show instant results on the left and suggestions on the right?

My impression was that, currently, both sides are treated equally. I wouldn't mind deciding that one side is dependent on the other, since that may allow some accessibility improvements.

olamothe: >  but is it currently possible to show instant results on the left and suggestions on the right?

Not that I'm aware of.

ThibodeauJF (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -585,19 +585,23 @@ export class AtomicSearchBox {
     lastIndex: number,
     side: 'left' | 'right'
   ) {
-    const id = `${this.id}-${side}-suggestion-${item.key}`;
+    const id = this.getSuggestionId(side, item);
+
     const isSelected = id === this.activeDescendant;
```
liviarett, line 590: how about 
```js
    const isSelected =
      id === this.activeDescendant ||
      (this.suggestedQuery === item.query &&
        !this.panelInFocus?.getAttribute('part')?.includes(side));
```
?
removes the need for the extra linking stuff and also supports swapping the sides if we ever want to do that

olamothe: Great suggestion !

btaillon (approved):

PR #2277, opened 2022-08-02
feat(headless): added insight use case to ref doc

nathanlb:
- Added autogen doc definitions to the doc parser
- Followed the trail of errors from missing exports.
  - Unfortunately it looks to me like we have to export the few `Core` interfaces I added here until we remove them in a breaking release. I might've missed something though. Let me know


btaillon (approved):

ThibodeauJF (approved):
Looks ok but I'd request for @samisayegh to review this to be 100% sure, he knows more about the doc parser.

nathanlb:
@samisayegh your expertise has been requested. :) Would you mind weighing in?

samisayegh:
Hey, there is a limitation with the tooling (api-extractor), where an interface needs to be exported for the tool to detect and extract it.

If a high-level interface is extending a core interface, both interfaces need to be made publicly available.
An alternative could be to copy all the properties of the core interface onto the high-level interface and not extend, but it quickly becomes tough to manage the duplication. I don't recommend going down this path ...

In an ideal world, we would be able to keep the core interfaces hidden, but considering the limitation with the tooling, it's ok if certain core interfaces are made public.

packages/headless/doc-parser/use-cases/insight.ts
```diff
@@ -0,0 +1,117 @@
+import {ActionLoaderConfiguration} from '../src/headless-export-resolvers/action-loader-resolver';
+import {ControllerConfiguration} from '../src/headless-export-resolvers/controller-resolver';
+import {EngineConfiguration} from '../src/headless-export-resolvers/engine-resolver';
+
+const controllers: ControllerConfiguration[] = [
+  {
+    initializer: 'buildFacet',
+    samplePaths: {},
```
samisayegh, line 8: Just an fyi, these paths are used to display code samples in the reference doc. They don't need to be available right away but it would be good to add them progressively,

packages/headless/src/insight.index.ts
```diff
@@ -92,6 +85,14 @@ export type {
   CategoryFacetSearch,
   CategoryFacetSearchState,
   CategoryFacetSearchResult,
+  /**
+   * @deprecated This is an internal controller that will be removed in the next version.
+   */
+  CoreCategoryFacet,
+  /**
+   * @deprecated This is an internal controller that will be removed in the next version.
```
samisayegh, line 98: Could it be that we need to keep exporting these core interfaces for documentation purposes?

samisayegh (approved):

packages/headless/src/insight.index.ts
nathanlb: I think that's it. I added them to fix errors the doc parser was throwing.

packages/headless/doc-parser/use-cases/insight.ts
nathanlb: Yeah sounds good. As we flesh out the insight panel use case for with a developer angle we'll improve the documentation.

PR #2279, opened 2022-08-02
fix(headless): respect index limits when reaching max page

olamothe:
https://coveord.atlassian.net/browse/KIT-1859


packages/headless/src/features/pagination/pagination-constants.ts
```diff
@@ -0,0 +1,2 @@
+export const minimumPage = 1;
```
olamothe, line 1: Extracted in a separate file to prevent circular dependencies

packages/headless/src/features/search/search-request.ts
```diff
@@ -15,6 +16,24 @@ export const buildSearchRequest = async (state: StateNeededBySearchRequest) => {
   const sharedWithFoldingRequest =
     await buildSearchAndFoldingLoadCollectionRequest(state);
 
+  // Corner case:
+  // If the number of results requested would go over the index limit (typically 5000 results is the maximum you can reach)
```
ThibodeauJF, line 20:
```suggestion
  // If the number of results requested would go over the index limit (maximumNumberOfResultsFromIndex)
```

packages/headless/src/features/search/search-request.ts
```diff
@@ -15,6 +16,24 @@ export const buildSearchRequest = async (state: StateNeededBySearchRequest) => {
...
+  // we need to request fewer results in order to ensure we do not receive an exception from the index
+  const ensureNumberOfResultsDoesNotGoOverIndexLimit = () => {
+    if (!state.pagination) {
```
ThibodeauJF, line 23: Shouldn't be undefined if the method is called

ThibodeauJF (approved):

packages/headless/src/features/search/search-request.ts
```diff
@@ -15,6 +16,24 @@ export const buildSearchRequest = async (state: StateNeededBySearchRequest) => {
   const sharedWithFoldingRequest =
     await buildSearchAndFoldingLoadCollectionRequest(state);
 
+  // Corner case:
+  // If the number of results requested would go over the index limit (typically 5000 results is the maximum you can reach)
+  // we need to request fewer results in order to ensure we do not receive an exception from the index
+  const ensureNumberOfResultsDoesNotGoOverIndexLimit = () => {
```
btaillon, line 22: Typically, with a name like `ensure`, I would expect something in the state to be changed, not for a clamped value to be returned.

I would rename this to something like `getNumberOfResultsWithinIndexLimit`

btaillon (approved):

packages/headless/src/features/search/search-request.ts
olamothe: TypeScript is not aware of that in the scope of this function.

PR #2280, opened 2022-08-02
test(quantic): Update quantic cypress test to insight use case

aelfaleh:
This PR investigate and propose a new way to do cypress test for quantic component in two use cases: search and insight.

To modify the current community exemple we need to create a new component named `c-example-use-case` : 
```
<template>
  <template if:true={isSearch}>
    <c-quantic-search-interface engine-id={engineId}>
      <slot></slot>
    </c-quantic-search-interface>
  </template>
  <template if:true={isInsight}>
    <c-quantic-insight-interface insight-id={insightId} engine-id={engineId}>
      <slot></slot>
    </c-quantic-insight-interface>
  </template>
</template>
```
Then add evrey component to test inside this new component: 
Exemple:
```
<c-example-use-case slot="preview" use-case={config.useCase} engine-id={engineId}>
      <c-quantic-pager
        engine-id={engineId}
        number-of-pages={config.numberOfPages}>
      </c-quantic-pager>
</c-example-use-case>
```
The attribute `use-case` can accepte string values : search or insight.

in the cypress test:  we need to make some modification to change the use case option:

```
interface PagerOptions {
  useCase: string; // <--- can take search or insight as values
  numberOfPages: number;
}
```
image


nathanlb:
Could you please add a good description of what the changes are? What's the strategy to test the insight use case in quantic? Thanks

packages/quantic/force-app/examples/main/lwc/exampleQuanticBreadcrumbManager/exampleQuanticBreadcrumbManager.js
```diff
@@ -1,27 +1,37 @@
...
+    },
+    {
+      attribute: 'collapseThreshold',
+      label: 'Collapse Threshold',
+      description:
+        'Maximum number of displayed breadcrumb values. When more values are selected, additional values appear under the "More" button.',
+      defaultValue: 5,
+    },
+    {
+      attribute: 'isSearch',
```
nathanlb, line 28: I think using some sort of option like this is a good idea. However I'm concerned that `isSearch` won't scale if/when we end up with more than two use cases.

How about using something like a `useCase` option which accepts a string and can be set to either `"search"`, `"insight"`, `"recommendation"`, etc...

aelfaleh: good catch, I was kind of hesitated between string and boolean for the same, now I can be more sure about string

aelfaleh:
> Could you please add a good description of what the changes are? What's the strategy to test the insight use case in quantic? Thanks

working on that with some screenshots

packages/quantic/force-app/examples/main/lwc/exampleUseCase/exampleUseCase.js
```diff
@@ -0,0 +1,19 @@
...
+    get isSearch() {
+        return this.useCase === 'search';
+    }
+
+    get isInsight() {
+        return this.useCase === 'insight';
+    }
+
+
+
```
lbergeron, line 18: Please remove these blank lines.

lbergeron (commented):
Well done! It looks pretty nice.

packages/quantic/cypress/integration/breadcrumb-manager/breadcrumb-manager.cypress.ts
```diff
@@ -25,8 +32,12 @@ describe('quantic-breadcrumb-manager', () => {
     interceptSearch();
     cy.visit(breadcrumbManagertUrl);
     configure(options);
+    if (options.useCase !== useCaseEnum.search) {
+      cy.wait(1000);
```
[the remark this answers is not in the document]
aelfaleh: I am not waiting for an element, I am waiting for this `window.coveoHeadless[engineId]?.bundle`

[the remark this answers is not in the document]
lbergeron: Well, once the component under test is initialized, I guess we could check for the existence of a specific element or CSS class to know it is initialized, right? If we can find a reliable way to know the component is initialized, the test would be faster (because it should take less than 1sec to initialize) and be more reliable (because Cypress default timeout is 4sec in case it takes longer to initialize).

lvu285:
Nice job Asma!
I have questions: 
1. Does it mean we always run tests for Quantic & Insight at the same time? for example test for add Facet will run for Quantic 
then Insight right away?
2. Can we just run a test for a specific product, for example "CategoryFacet" for Quantic but not for Insight?

packages/quantic/cypress/integration/pager/pager.cypress.ts
```diff
@@ -1,5 +1,6 @@
 import {configure} from '../../page-objects/configurator';
 import {
+  getAlias,
   InterceptAliases,
```
erocheleau, line 4: Just so you see the linter here :)

packages/quantic/cypress/integration/results-per-page/results-per-page.cypress.ts
```diff
@@ -27,8 +29,12 @@ describe('quantic-result-per-page', () => {
       },
     });
     configure(options);
+    if (options.useCase === useCaseEnum.insight) {
```
erocheleau, line 32: So why exactly do we wait only whenever the useCase is insight?

packages/quantic/force-app/examples/main/lwc/exampleQuanticResultQuickview/exampleQuanticResultQuickview.js
```diff
@@ -1,68 +1,76 @@
...
+  isConfigured = false;
 
+  connectedCallback() {
+    registerComponentForInit(this, this.engineId);
+  }
 
-    connectedCallback() {
-        registerComponentForInit(this, this.engineId);
-    }
+  result = {
```
erocheleau, line 13: If you wanted, I'm sure you can create an js file next to this component and define all these object constants in that different file so they get exported and you can import them in this file.
(If you wanted to split it so this component doesn't declare a bunch of constants inline of the code)

packages/quantic/force-app/examples/main/lwc/exampleUseCase/exampleUseCase.js
```diff
@@ -0,0 +1,16 @@
...
+  insightId = '6a333202-b1e0-451e-8664-26a1f93c2faf';
+
+  get isSearch() {
+    return this.useCase === 'search';
+  }
+
+  get isInsight() {
+    return this.useCase === 'insight';
+  }
+}
```
erocheleau, line 16: For future's sake, I think you should use 1 render method here and render different templates instead:

https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_render

```suggestion
import {api, LightningElement} from 'lwc';
import {searchTemplate} from './searchTemplate.html';
import {insightTemplate} from './insightTemplate.html';

export default class ExampleUseCase extends LightningElement {
  @api engineId = '';
  @api useCase = '';

  insightId = '6a333202-b1e0-451e-8664-26a1f93c2faf';

  render() {
    if(this.useCase === 'search') {
      return searchTemplate;
    } else if (this.useCase === 'insight') {
      return insightTemplate;
    }
  }
}
```

packages/quantic/force-app/examples/main/lwc/exampleUseCase/exampleUseCase.html
```diff
@@ -0,0 +1,12 @@
...
+    <c-quantic-search-interface engine-id={engineId}>
+      <slot></slot>
+    </c-quantic-search-interface>
+  </template>
+  <template if:true={isInsight}>
+    <c-quantic-insight-interface insight-id={insightId} engine-id={engineId}>
+      <slot></slot>
+    </c-quantic-insight-interface>
+  </template>
+</template>
```
erocheleau, line 12: For future's sake, I think you should use 1 render method here and render different templates instead:

https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_render

```suggestion
<!-- searchTemplate.html -->
<template>
  <c-quantic-search-interface engine-id={engineId}>
    <slot></slot>
  </c-quantic-search-interface>
</template>

<!-- insightTemplate.html -->
<template>
  <c-quantic-insight-interface insight-id={insightId} engine-id={engineId}>
    <slot></slot>
  </c-quantic-insight-interface>
</template>
```

erocheleau (changes requested):
I think I understand pretty much everything, but this is lots of work, it might be work planning a 10 minutes demo and meeting to discuss the changes with everyone in the team?

packages/quantic/cypress/integration/results-per-page/results-per-page.cypress.ts
lbergeron: It is because the headless bundle has to be downloaded before we can execute the first search query. We don't have this problem with the search use case because the QuanticSearchInterface is triggering the first query automatically. For the insight use case, the first search has to be triggered by the renderer component, not from Quantic.

It would be nice though if the QuanticInsightInterface component could have some sort of marker to inform the test framework that the component is properly initialized.

lbergeron:
> Nice job Asma! I have questions:
> 
> 1. Does it mean we always run tests for Quantic & Insight at the same time? for example test for add Facet will run for Quantic
>    then Insight right away?
> 2. Can we just run a test for a specific product, for example "CategoryFacet" for Quantic but not for Insight?

1. Yes the Search and Insight use cases are tested in the same test suite. For instance, when the facet test suite runs, all facet tests are executed for the Search use case, then all facets tests are executed for the Insight use case. When this test suite is completed, Cypress continues with the next test suite.
2.  No, but that would be possible. We'd have to refactor the test suites to have the Search and Insight tests in different files. Both test suites could reuse a common set of tests though so code duplication wouldn't be an issue. I think that the Cypress tests should always validate all use cases though since they are all shipped as part of Quantic.

packages/quantic/cypress/integration/results-per-page/results-per-page.cypress.ts
aelfaleh: @lbergeron  I create a hidden input with attribute `is-initialized` to replace the `cy.wait(10000`

packages/quantic/cypress/integration/breadcrumb-manager/breadcrumb-manager.cypress.ts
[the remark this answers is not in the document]
aelfaleh: I did the change

packages/quantic/cypress/integration/facets/timeframe-facet/timeframe-facet.cypress.ts
```diff
@@ -11,25 +13,28 @@ import {scope} from '../../../reporters/detailed-collector';
...
+  useCase: string;
 }
 
 describe('quantic-timeframe-facet', () => {
   const pageUrl = 's/quantic-timeframe-facet';
 
   const validRange = {
-    start: '2000-01-01',
-    end: '2000-12-31',
+    start: '01/01/2000',
```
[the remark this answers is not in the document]
aelfaleh: still have problem even in CI

aelfaleh:
@erocheleau  if you have time, it is ready to be reviewed 👌

lvu285 (approved):

erocheleau (approved):
Everything looks good.

Nicely done!

lbergeron (approved):

aelfaleh:
@olamothe  @ThibodeauJF  if you have time guys, I need another approval 🙏

packages/headless/src/controllers/insight/results-per-page/headless-insight-results-per-page.ts
```diff
@@ -33,6 +33,12 @@ export function buildResultsPerPage(
   return {
     ...coreController,
 
+    get state() {
```
olamothe, line 36: Does not seem necessary.

olamothe (approved):

packages/headless/src/controllers/insight/results-per-page/headless-insight-results-per-page.ts
aelfaleh: it doesn't work without it, when we click on result per page

PR #2286, opened 2022-08-03
fix(atomic): improved accessibility for search box suggestions

btaillon:
https://coveord.atlassian.net/browse/KIT-1880

Important to know: suggestions don't use the "real" browser focus, they use a fake/simulated/visual focus.

# The problems

## Incorrect ARIA guideline
The search box was implemented by following ARIA 1.0 guidelines, which uses `aria-owns` to modify the accessibility tree in a way [which causes issues](https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues#aria-10-combobox-problem). The fix for this issue is simple, `aria-owns` should be replaced with `aria-controls`. This is the issue that was reported in the assessment.

## Instant results unreachable
When instant results are present, suggestions were represented as one listbox containing two lists of options. NVDA (and possibly JAWS also) gave no hint that there is a second list of suggestions, and whether we move through suggestions using the Browse mode or the Focus mode, instant results are completely skipped as if they don't exist.

## Instant results not changing
When we focus on a suggestion with any screen reader, instant results are never changed, because the onFocus event is never triggered. I tested this on VoiceOver for macOS and on NVDA for Windows.

## Suggestions have no context
Visually, we see an icon that represents whether a suggestion comes from recent queries or whether a suggestion comes from Coveo. This wasn't represented to screen readers, so I added aria labels.

# The solution
I transformed the suggestions from a listbox into an ["application"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/application_role), which allows custom keyboard controls. By allowing keyboard controls, we don't need the regular focus screen reader focus at all, and setting `aria-activedescendant` to the current suggestion causes NVDA and JAWS to read it out just as if it was focused.

I also adapted aria labels to tell the user when there is a second list of suggestions, and which side is currently selected. I added even more aria labels to represent what type of suggestion is selected.

# Remaining issues
## aria-activedescendant on macOS
Setting/updating `aria-activedescendant` doesn't do anything on VoiceOver for macOS when `role="application"`. `aria-live` is going to be needed in order to notify macOS users of a new selection. Additionally, the way that macOS interacts with the suggestions is different: it needs to move the focus out of the search box and into the suggestions. Instructions need to be adapted.

## Instant results on mobile
Although we don't want to show instant results on mobile, mobile users are still notified that there is two lists of suggestions. I wish to fix this in a separate PR.

## Lack of information for instant results
Instant results contain more visual information than what is represented to screen readers. Using the arrow keys, we cannot select the contents of an instant results, therefor we cannot interact with buttons that would be inside of it and we cannot read individual parts generated by the template such as the title, a rating, fields, etc. This will be a more complex problem to solve, since we don't support templating a string and therefor cannot generate an appropriate aria label.


packages/atomic/src/locales.json
```diff
@@ -53,9 +53,49 @@
...
+  },
   "search-box": {
-    "en": "Input field to perform a search. Insert a query. Press enter to send.",
-    "fr": "Champ de texte pour performer une recherche. Insérer une requête. Appuyez sur Entrée pour envoyer."
+    "en": "Input field to perform a search. Insert a query. To send, press Enter.",
+    "fr": "Champ de texte pour performer une recherche. Insérer une requête. Pour envoyer, appuyez sur Entrée."
+  },
+  "search-box-with-suggestions": {
+    "en": "Search field with suggestions. To begin navigating suggestions, while focused, press Down Arrow. To send, press Enter.",
+    "fr": "Champ de rechercher avec des suggestions. Pour commencer à naviguer à travers les suggestions, avec cible de saisie, appuyez sur Flèche Basse. Pour envoyer, appuyez sur Entrée."
```
jpmarceau, line 70:
```suggestion
    "fr": "Champ de recherche avec des suggestions. Pour commencer à naviguer à travers les suggestions, avec cible de saisie, appuyez sur Flèche Basse. Pour envoyer, appuyez sur Entrée."
```

packages/atomic/src/locales.json
```diff
@@ -577,6 +617,10 @@
     "en": "Recent searches",
     "fr": "Recherches récente"
   },
+  "clear-recent-searches": {
+    "en": "Clear recent searches",
+    "fr": "Effacer recherches récente"
```
jpmarceau, line 622:
```suggestion
    "fr": "Effacer les recherches récentes"
```

jpmarceau (approved):
Very good strings Benjamin 👌

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -590,30 +587,65 @@ export class AtomicSearchBox {
...
+      class: `flex px-4 min-h-[40px] items-center text-neutral-dark hover:bg-neutral-light cursor-pointer ${
+        isSelected ? 'bg-neutral-light' : ''
+      }`,
+      ref: (el: HTMLElement | undefined) => {
+        if (isHTMLElement(item.content)) {
+          el?.replaceChildren(item.content);
+        }
+        el?.setAttribute(
+          'aria-label',
+          this.bindings.i18n.t(
```
ThibodeauJF, line 605: Can we extract the aria label content up top?

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -636,15 +668,15 @@ export class AtomicSearchBox {
...
+    setRef: (el: HTMLElement | undefined) => void,
+    getRef: () => HTMLElement | undefined
   ) {
     if (!elements.length) {
       return null;
     }
 
     return (
-      <ul
+      <div
```
ThibodeauJF, line 679: Thank you shadow DOM! Glad we can do this

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -674,7 +706,13 @@ export class AtomicSearchBox {
         class={`flex w-full z-10 absolute left-0 top-full rounded-md bg-background border border-neutral ${
           this.showSuggestions ? '' : 'hidden'
         }`}
-        role="listbox"
+        role="application"
```
ThibodeauJF, line 709: Could the role stay a simple listbox if there's no double list?

ThibodeauJF (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
ThibodeauJF: Would perhaps add better support for Mac for customers that don't use instant results (probably most customers)

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
btaillon: no because we may need to read the rendered content

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
btaillon: perhaps, but I prefer to keep the behaviour consistent, since both behave pretty differently and it would be like testing two components.

ThibodeauJF: Valid point & the end result will be pretty much the same to the screen reader user

olamothe: Hum... yeah I would tend to agree with Jean-Felix on this one.

I'd say realistically speaking, vast majority of clients will just be using normal query suggestions, and instant results will be the exception.

So it's like we will offer something less robust with role="application" (that's my understanding that it's less robust, correct me if I'm wrong), in order to support something that most client will not use...

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -590,30 +587,65 @@ export class AtomicSearchBox {
     if (index === lastIndex && item.hideIfLast) {
       return null;
     }
+
+    const isButton = item.onSelect || item.query;
+    const commonAttributes = {
```
olamothe, line 592: There's a bit of an overload of ternary operation in there 😂 .

I think it would be more readable if the code was more verbose, with each section extracted into small private function:

```
getClass();
setReference();
setAriaLabel();
```

And just more good old boring `if (x) {return y}`

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
btaillon: Yes, it's a little bit less robust on macOS with VoiceOver. Apart from that, it seems pretty robust.

I would prefer to wait and see if clients report this as an accessibility issue.

ThibodeauJF (approved):

btaillon:
Merged, we can easily revert back to listbox on cases with no instant results if this causes issues, but now it seems robust enough.

PR #2289, opened 2022-08-03
fix(atomic,headless): fix atomic pager size overflow + headless page fraction

olamothe:
Couple different problems fixed in this PR:

* Headless would return fraction of pages when manually changing `firstResult` in the URL
* Atomic would incorrectly display very big pager number in `atomic-pager` (would overflow)
* Fix some alignment issue with next and previous pager button

https://coveord.atlassian.net/browse/KIT-1906

### DIFF
--- a/packages/atomic/src/components/search/atomic-pager/atomic-pager.pcss
+++ b/packages/atomic/src/components/search/atomic-pager/atomic-pager.pcss
@@ -1,7 +1,1 @@
 @import '../../../global/global.pcss';
-
-button,
-.btn-page {
-  @apply min-w-[2.5rem] min-h-[2.5rem];
-  @apply p-1;
-}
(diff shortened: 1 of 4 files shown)


btaillon (approved):

packages/atomic/src/components/search/atomic-pager/atomic-pager.pcss
```diff
@@ -2,5 +2,6 @@
 
 button,
 .btn-page {
-  @apply w-10 h-10;
+  @apply min-w-[2.5rem] min-h-[2.5rem];
+  @apply p-1;
```
ThibodeauJF, line 6: what if we moved these back in the JSX?

ThibodeauJF (approved):

PR #2291, opened 2022-08-04
feat(atomic): add count label on popover

mbenkirane-coveo:
https://user-images.githubusercontent.com/97299176/182834182-7a194e6c-652a-4120-930c-54a79fcfebf5.mov

### DIFF
--- a/packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
+++ b/packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
@@ -52,12 +52,12 @@ export class AtomicPopover implements InitializableComponent {
   @State() public isMenuVisible = false;
   @State() public facetId?: string;
   @State() public facetLabel?: string = 'no-label';
-  @State() public numberOfValues?: number = 0;
-  @State() public numberOfSelectedValues?: string = '0';
+  @State() public hasValues?: boolean = false;
+  @State() public numberOfSelectedValues?: number = 0;
   @Element() host!: HTMLElement;
   private facetElement?: HTMLElement;
-  private getNumberOfValues?: () => number;
-  private getNumberOfSelectedValues?: () => string;
+  private getHasValues?: () => boolean;
+  private getNumberOfSelectedValues?: () => number;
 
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/search/facets/atomic-popover/popover-event.ts
```diff
@@ -3,6 +3,8 @@ import {FacetType} from '../../../common/facets/facet-common';
 export interface InitPopoverEventPayload {
   facetId: string;
   facetType: FacetType;
+  getNumberOfValues: () => number;
```
ThibodeauJF, line 6: can be just `hasValues` since you use it for that purpose.

packages/atomic/src/components/search/facets/atomic-popover/popover-event.ts
```diff
@@ -3,6 +3,8 @@ import {FacetType} from '../../../common/facets/facet-common';
...
+  getNumberOfSelectedValues: () => string;
```
ThibodeauJF, line 7: I'd make this a `number` instead of a string

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
```diff
@@ -117,6 +132,19 @@ export class AtomicPopover implements InitializableComponent {
...
+          title={this.numberOfSelectedValues}
+          part="value-count"
+          class={`value-box-count truncate pl-0.5 w-auto mt-0 text-sm ${
+            this.isMenuVisible
+              ? 'text-primary'
+              : 'text-neutral-dark group-hover:text-primary-light group-focus:text-primary'
+          }`}
+        >
+          {this.bindings.i18n.t('between-parentheses', {
+            text: this.numberOfSelectedValues,
```
ThibodeauJF, line 145: We can use `toLocaleString` on the count number value like we do for other numbers that we display inside atomic

mbenkirane-coveo: sure

PR #2295, opened 2022-08-05
feat(atomic): remove atomic-insight-breadbox from docs, warn when implementing...

liviarett:
breadbox with non-search

I actually didn't find any mention of the breadbox in the Figma file, so removed it at least for now!

https://coveord.atlassian.net/browse/KIT-1907

### DIFF
--- a/packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
+++ b/packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
@@ -21,7 +21,6 @@ import {
   FocusTargetController,
 } from '../../../utils/accessibility-utils';
 import {Bindings} from '../atomic-search-interface/atomic-search-interface';
-import {closest} from '../../../utils/utils';
 
 interface Breadcrumb {
   facetId: string;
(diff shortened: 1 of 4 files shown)


ThibodeauJF (changes requested):
This will break implementations using `atomic-external`

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -84,12 +85,18 @@ export class AtomicBreadbox implements InitializableComponent {
   private breadcrumbShowLessFocus!: FocusTargetController;
 
   public initialize() {
-    this.breadcrumbManager = buildBreadcrumbManager(this.bindings.engine);
-    this.facetManager = buildFacetManager(this.bindings.engine);
-
-    if (window.ResizeObserver) {
-      this.resizeObserver = new ResizeObserver(() => this.adaptBreadcrumbs());
-      this.resizeObserver.observe(this.host.parentElement!);
+    if (!closest(this.host, 'atomic-search-interface')) {
```
ThibodeauJF, line 88: Maybe the bindings could contain what type of interface you're connected to (search/insight)

olamothe (commented):
I understand the JS doc removal, but why explicitly add code for this specific component that verifies that the parent is a search interface, and not for all the other search components ?

The "convention" we went with is that implementers should be using components that match for `insight` (ie: will contain the `insight` prefix in the component tag name).

The changes in atomic-breadbox seems superfluous ?

liviarett:
@olamothe you're right, I think that because it kind of works with the wrong interface already without changing anything, I wanted to make sure it was clear that it wasn't meant to be used like that. but I agree that it should be handled somewhere else and in a more generic way!

btaillon (approved):

ThibodeauJF (approved):

PR #2299, opened 2022-08-05
feat(atomic): implement insight tabs

liviarett:
https://coveord.atlassian.net/browse/KIT-1913

https://user-images.githubusercontent.com/30511433/183128474-1b2fbeac-d00d-4f95-ab05-95cd84735472.mov


btaillon:
I know that for the search case, we wanted to avoid using tabs with expressions since they don't change facets and other parts of the UI. Has there been a discussion already that I'm not aware of on using those tabs with the insight case?

packages/atomic/src/pages/examples/insights.html
```diff
@@ -36,8 +36,14 @@
         width: 100%;
         display: grid;
         grid-template-columns: 1fr auto auto auto;
+        /* grid-template-areas:
```
olamothe, line 39: Remove

packages/atomic/src/pages/examples/insights.html
```diff
@@ -36,8 +36,14 @@
...
+          '1fr auto auto auto'
+          'tags tags tags tags'; */
+
         grid-gap: 0.5rem;
 
+        background: var(--atomic-neutral-light);
```
olamothe, line 45: Probably won't have time to tackle this, but it would have been nice to have `insight-layouts` components for the CSS in this page. The same way we have search layouts components.

packages/atomic/src/components/insight/atomic-insight-tabs/atomic-insight-tabs.tsx
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @internal
+ * @part tab - The button element for each tab.
+ */
+@Component({
+  tag: 'atomic-insight-tabs',
+  styleUrl: './atomic-insight-tabs.pcss',
+  shadow: true,
+})
+export class AtomicInsightTabs {
```
olamothe, line 13: Wonder if we could have piggybacked/re-used what Simo worked on for scrollable segmented facet. It's pretty much the same use case, and looks better than a normal scrollbar (especially on Windows)

olamothe (approved):

olamothe:
> I know that for the search case, we wanted to avoid using tabs with expressions since they don't change facets and other parts of the UI. Has there been a discussion already that I'm not aware of on using those tabs with the insight case?

For the insight panel, what Livia did is perfect.

For the search page you are right, though.

packages/atomic/src/pages/examples/insights.html
liviarett: yeah, I have a task for that in JIRA... the plan is to basically transport these styles over to layout components

packages/atomic/src/components/insight/atomic-insight-tab/atomic-insight-tab.tsx
```diff
@@ -0,0 +1,74 @@
...
+    if (!this.bindings.store.isAppLoaded()) {
+      return;
+    }
+
+    return (
+      <Button
+        style="text-transparent"
+        part="tab"
+        class={this.tabState.isActive ? 'active' : ''}
+        ariaLabel={this.label}
```
btaillon, line 67: I would also set `ariaPressed` to `this.tabState.isActive`

btaillon: I would also make sure that the aria label describes that this is a tab (in addition to containing the visual label).

packages/atomic/src/components/insight/atomic-insight-tab/atomic-insight-tab.pcss
```diff
@@ -0,0 +1,21 @@
...
+  border-radius: none;
+}
+
+.active {
+  &::after {
+    content: '';
+    display: block;
+    width: 100%;
+    height: 4px;
+    border-radius: 4px;
```
btaillon, line 16: I would use `rem` to that they scale with the text.

packages/atomic/src/components/insight/atomic-insight-tabs/atomic-insight-tabs.pcss
```diff
@@ -0,0 +1,7 @@
+@import '../../../global/global.pcss';
+
+:host {
+  grid-column: 1 / 5;
```
btaillon, line 4: `grid-column` is only relevant because the parent component is a grid in `insights.html`. Elsewhere, we don't necessarily know that this may be the case.

I would move `grid-column` to `insights.html`.

btaillon (approved):

packages/atomic/src/components/insight/atomic-insight-tabs/atomic-insight-tabs.pcss
liviarett: yeah you're right. this should be conditional on the layout system that will come later...

packages/atomic/src/locales.json
```diff
@@ -77,6 +77,10 @@
     "en": "Search suggestions. Select one to search.",
     "fr": "Suggestions de recherche. Sélectionnez-en un pour rechercher."
   },
+  "tab-search": {
```
liviarett, line 80: @btaillon would you approve this PR again with this new aria label? I left it as 'facette' in French, let me know if there is a better word!

packages/atomic/src/components/insight/atomic-insight-tab/atomic-insight-tab.pcss
```diff
@@ -0,0 +1,13 @@
+@import '../../../global/global.pcss';
+
+button {
+  @apply relative pb-3 mt-1 mr-6;
+  font-weight: 600;
```
ThibodeauJF, line 5:
```suggestion
  @apply relative pb-3 mt-1 mr-6 font-semibold;
```

ThibodeauJF (approved):

packages/atomic/src/locales.json
btaillon: @liviarett "facette" would be the French equivalent to "Facet". The French equivalent to "tab" would've been "onglet".

This is what I think the string should've been:
```suggestion
    "en": "{{label}} tab.",
    "fr": "Onglet {{label}}"
```

I'm going to add it in an unrelated PR.

PR #2300, opened 2022-08-05
fix(atomic): updated search box instructions when there's no keyboard

btaillon:
https://coveord.atlassian.net/browse/KIT-1910

### DIFF
--- a/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
@@ -37,7 +37,7 @@ import {
   SimpleSearchSuggestion,
 } from './search-suggestion';
 
-import {isMacOS} from '../../../utils/device-utils';
+import {hasKeyboard, isMacOS} from '../../../utils/device-utils';
 
 /**
  * The `atomic-search-box` component creates a search box with built-in support for suggestions.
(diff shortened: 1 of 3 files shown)


packages/atomic/src/locales.json
```diff
@@ -73,6 +73,10 @@
     "en": "Search field with suggestions. Suggestions may be available under this field. To send, press Enter.",
     "fr": "Champ de recherche avec suggestions. Des suggestions pourraient être disponibles sous ce champ. Pour envoyer, appuyez sur Entrée."
   },
+  "search-box-with-suggestions-keyboardless": {
+    "en": "Search field with suggestions. Suggestions may be available under this field. To send, press Enter.",
+    "fr": "Champ de recherche avec suggestions. Des suggestions pourraient être disponibles sous ce champ. Pour envoyer, appuyez sur Entrée."
+  },
```
btaillon, line 79: Same instructions as macOS.

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -238,8 +238,8 @@ export class AtomicSearchBox {
 
   private get isDoubleList() {
     const numberOfSuggestionsLists =
-      (this.leftSuggestions.length ? 1 : 0) +
-      (this.rightSuggestions.length ? 1 : 0);
+      (this.leftSuggestionElements.length ? 1 : 0) +
+      (this.rightSuggestionElements.length ? 1 : 0);
 
     return numberOfSuggestionsLists === 2;
```
liviarett, line 244: seems like a bit os a convoluted way to do
```js
private get isDoubleList() {
    return Boolean(this.leftSuggestionElements.length && this.rightSuggestionElements.length)
```

btaillon: oh right, I converted this code from something like listsCount to isDoubleList and went too quick 🤦

PR #2301, opened 2022-08-05
feat(atomic): add number badge and view results button to insight ref…

liviarett:
…ine toggle

https://coveord.atlassian.net/browse/KIT-1917

Screenshot 2022-08-05 at 19 52 03


packages/atomic/src/components/common/atomic-icon-button/atomic-icon-button.tsx
```diff
@@ -21,24 +21,28 @@ export class AtomicIconButton {
...
+          title={this.tooltip}
+          ref={this.buttonRef}
+        >
+          <atomic-icon
+            icon={this.icon}
+            class="w-4 h-4 shrink-0"
+            aria-hidden="true"
+          ></atomic-icon>
+        </Button>
+        {this.badge && <span part="badge">{this.badge}</span>}
```
btaillon, line 44: Although we don't necessarily need to be accesible for this project, I would add an aria label here just in case we do in the future.

btaillon: Also, I would document this part

packages/atomic/src/components/insight/atomic-insight-refine-toggle/atomic-insight-refine-toggle.pcss
```diff
@@ -0,0 +1,14 @@
+atomic-icon-button::part(badge) {
+  background: var(--atomic-primary);
+  color: var(--atomic-on-primary);
+  width: 1rem;
+  display: block;
+  height: 1rem;
+  border-radius: 1rem;
```
btaillon, line 7: I prefer using a border radius in percentage, so that if we change the dimensions it stays a circle.

btaillon (approved):

packages/atomic/src/components/insight/atomic-insight-refine-toggle/atomic-insight-refine-toggle.pcss
```diff
@@ -0,0 +1,14 @@
+atomic-icon-button::part(badge) {
```
ThibodeauJF, line 1: any reason tailwind wasn't used here?

olamothe: +1

packages/atomic/src/components/common/atomic-icon-button/atomic-icon-button.tsx
olamothe, line 44: A bit nitpicky, but just for the sake of future flexibility, `badge` could be a JSX node instead of a string.

olamothe (approved):
Looks nice !

PR #2302, opened 2022-08-05
feat(atomic): added redirection trigger

btaillon:
https://coveord.atlassian.net/browse/KIT-1921

Opening this partially for discussion (this is the easiest approach).

Do we want developers to be able to disable redirection?
1. On one hand, this can probably be avoided by adding some extra conditions in the admin UI.
2. On the other hand, if somebody has multiple search interfaces that share the same pipeline, it would be easier to disable it at the UI-level. I don't know how often people have multiple search interfaces that use the same pipeline though.

Do we want redirection triggers to be included with the search interface, or a separate component (e.g., `atomic-redirection-trigger`).
1. On one hand, this is an extremely simple "component" that from an admin perspective would probably be expected to come with the search interface.
2. On the other hand, having each trigger use a component may be easier to document, especially for parity with Headless. It also removes some documentation on the search interface itself.

Do we want the redirection trigger to be on by default?
1. On one hand, when somebody configures a redirection in the admin UI, they probably don't expect to need to ask their dev team to support it (unless it's a dev adding the redirection).
2. On the other hand, redirection is something that affects the whole page, and may not be expected if the search interface is just one small part of a larger page.

Personally:
* I think being able to disable redirection is slightly preferable to not providing the option.
* I'm fine with keeping this as a part of the atomic search interface, unless we want more options for some reason.
* I prefer it on by default.

### DIFF
--- a/packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
+++ b/packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
@@ -21,6 +21,8 @@ import {
   loadSearchConfigurationActions,
   loadQueryActions,
   EcommerceDefaultFieldsToInclude,
+  RedirectionTrigger,
+  buildRedirectionTrigger,
 } from '@coveo/headless';
 import i18next, {i18n} from 'i18next';
 import {InitializeEvent} from '../../../utils/initialization-utils';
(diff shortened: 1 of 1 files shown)


ThibodeauJF:
Just leaving this out there that this could be a breaking change in behaviour for a customer that has redirection triggers enabled but uses a separate controller or simply doesn't want redirection with their current implementation.

I'd perhaps add a prop that defaults to false & when we bump to v2, set it to true

olamothe (changes requested):
This is not needed. Redirection trigger are already handled by the standalone search box.

PR #2305, opened 2022-08-08
fix(atomic): remove duplicate query suggestions in recent queries component

olamothe:
https://coveord.atlassian.net/browse/KIT-1916


btaillon (approved):
I like this approach, but I have a very slight preference for removing the suggestion from the Coveo suggestions instead of recent queries, since somebody may find their recent query missing from the recent queries and think it's a bug.

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-recent-queries/atomic-search-box-recent-queries.tsx
```diff
@@ -114,13 +114,18 @@ export class AtomicSearchBoxRecentQueries {
     }
 
     const query = this.bindings.searchBoxController.state.value;
+    const querySuggestions =
```
ThibodeauJF, line 117: A better implementation could be to scope out any other query from the suggestions, whether they are QS or not, at the search box level. We could keep the first occurence of a query & discard of the others.

ThibodeauJF: Currently if you add custom suggestions you'd run in the same duplication problem

ThibodeauJF (approved):
Very nice! I like it

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -510,18 +509,21 @@ export class AtomicSearchBox {
 
   private updateSuggestionElements(query: string) {
     if (!this.isPanelInFocus(this.leftPanelRef, query)) {
-      this.leftSuggestionElements = this.getSuggestionElements(
-        this.leftSuggestions
-      );
-    }
+      this.leftSuggestionElements = this.getAndFilterLeftSuggestionElements();
```
btaillon, line 512: I feel like it would make more sense to filter both sides than not 🤔

packages/atomic/src/utils/utils.ts
```diff
@@ -170,3 +170,24 @@ export function isInDocument(element: Node) {
...
+  items: Item[],
+  propNames: (keyof Item)[]
+) {
+  return items.filter(
+    (item, index, self) =>
+      index ===
+      self.findIndex((foundItem) =>
+        isPropValuesEqual(foundItem, item, propNames)
+      )
+  );
```
btaillon, line 192: Since this is a util, instead of using a *O*², I think it would be worth improving performance by using a `Set`.

```ts
const existingProps = new Set<string>();
return items.filter(
  (item) => {
    const hash = hashProps(item, propNames);
    if (existingProps.has(hash)) {
      return false;
    }
    existingProps.add(hash);
    return true;
  }
);
```

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
olamothe: What would you filter on ?

packages/atomic/src/utils/utils.ts
olamothe: Won't do for now, just seems overkill for the time being.

PR #2306, opened 2022-08-08
feat(atomic): added atomic-notifications

btaillon:
https://coveord.atlassian.net/browse/KIT-1923

image

I made it only display the first notification returned by the API. In your opinions, do you think it should be able to display multiple notifications? If so, I'll rename it to `atomic-notifications`. I'm going to wait for Johnny's response when he's back.

Missing:
* Support for links (suggested by Johnny)
* Allowing atomic-aria-live to read changes that happened at the same time.


packages/atomic/src/components/search/atomic-notification/atomic-notification.tsx
```diff
@@ -0,0 +1,85 @@
...
+  InitializeBindings,
+} from '../../../utils/initialization-utils';
+import {Hidden} from '../../common/hidden';
+import {Bindings} from '../atomic-search-interface/atomic-search-interface';
+import InfoIcon from '../../../images/info.svg';
+import {AriaLiveRegion} from '../../../utils/accessibility-utils';
+import {Heading} from '../../common/heading';
+
+/**
+ * The `atomic-notification` component is responsible for displaying notifications generated by the Coveo Search API (see [Trigger](https://docs.coveo.com/en/1458/tune-relevance/trigger-query-pipeline-feature)).
```
jpmarceau, line 19:
```suggestion
 * The `atomic-notification` component is responsible for displaying notifications generated by the Coveo Search API (see [Trigger](https://docs.coveo.com/en/1458)).
```
(everything after the ID is unnecessary)

jpmarceau (approved):
Excellent strings, as usual 👌

packages/atomic/src/pages/index.html
```diff
@@ -211,6 +211,9 @@
             </atomic-sort-dropdown>
             <atomic-did-you-mean></atomic-did-you-mean>
           </atomic-layout-section>
+          <atomic-layout-section section="notification"
```
ThibodeauJF, line 214: I'm not entirely convinced a new section is needed especially since there's only this component inside (requires a doc update + our customer to implement it), it could easily fit in either results or status.

ThibodeauJF (approved):
Nice!

packages/atomic/src/pages/index.html
btaillon: If it doesn't have its own section, then I'll add it in `status`.

packages/atomic/src/components/search/atomic-notification/atomic-notification.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  public renderNotification(text: string) {
+    return (
+      <div
+        part="notification"
+        class="shadow-lg bg-background border border-neutral-dark rounded-md p-6 flex items-center"
+      >
+        <atomic-icon
+          icon={InfoIcon}
```
olamothe, line 57: Suggestion: Make this icon a prop on the component, with the default value being the `InfoIcon`. It would be nice if there was also a way to say that you want "no icon" at all.

packages/atomic/src/pages/index.html
olamothe: Indeed, adding sections has a "cost" (doc, maintenance, burden on implementers). 

It needs to be really worth it before we add a new one.

olamothe (approved):
Great PR with docs, tests, storybook... 😎 

GG !

btaillon:
Re-requesting reviews because I added support for icons and renamed the component to `atomic-notifications` (despite only showing 1 notification at a time)

ThibodeauJF (approved):

jpmarceau (approved):

olamothe (approved):

PR #2310, opened 2022-08-09
feat(atomic): new popover facet component

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1839


packages/atomic/src/components/common/facets/facet-common-store.ts
```diff
@@ -0,0 +1,20 @@
+import {VNode} from '@stencil/core';
```
ThibodeauJF, line 1: moved out to prevent circular dependencies

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
```diff
@@ -0,0 +1,209 @@
...
+} from '../../../../utils/initialization-utils';
+import {Button} from '../../../common/button';
+import {Hidden} from '../../../common/hidden';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+import ArrowBottomIcon from 'coveo-styleguide/resources/icons/svg/arrow-bottom-rounded.svg';
+import {InitializePopoverEvent, popoverClass} from './popover-type';
+import {FacetInfo} from '../../../common/facets/facet-common-store';
+
+interface ClearPopoverEvent {
+  ownId?: string;
```
btaillon, line 29: I'm not sure I like `ownId`, perhaps `emitterId` or `popoverIdToIgnore`?

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
```diff
@@ -0,0 +1,209 @@
...
+ */
+@Component({
+  tag: 'atomic-popover',
+  styleUrl: 'atomic-popover.pcss',
+  shadow: true,
+})
+export class AtomicPopover implements InitializableComponent {
+  @Element() host!: HTMLElement;
+  private hasValues?: () => boolean;
+  private numberOfSelectedValues?: () => number;
```
btaillon, line 52: Instead of spreading the facet methods & attributes over multiple attributes, I would prefer if they were part of a single object, like `@State() childFacet`.

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
```diff
@@ -0,0 +1,209 @@
...
+        ></div>
+      );
+    }
+
+    if (!this.searchStatus.state.hasResults || !this.hasValues?.()) {
+      return <Hidden></Hidden>;
+    }
+
+    return (
+      <div id={this.popoverId} class="relative">
```
btaillon, line 198: The button has `aria-controls` pointing towards this `id`, however, this element is a parent of the button. I think it would make more sense to move this `id` to the element below which has `part="slot-wrapper"`.

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
```diff
@@ -0,0 +1,209 @@
...
+        {this.renderDropdownButton()}
+        <div
+          part="slot-wrapper"
```
btaillon, line 201: I think it would make more sense to give this element a part name like "popup".

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  private renderDropdownButton() {
+    const label = this.bindings.i18n.t(this.facetInfo!.label);
+    const hasSelectedValues = !!this.numberOfSelectedValues?.();
+    const count = this.numberOfSelectedValues?.().toLocaleString();
+    const ariaLabel = this.bindings.i18n.t(
+      this.isOpen ? 'close-popover' : 'open-popover',
+      {label}
+    );
```
btaillon, line 129: I would avoid toggling between two different labels since `aria-expanded` should already be conveying this (at-least in theory, we should probably test with screen readers). I would use a label like "Popup filter for {{label}}".

btaillon (approved):

olamothe (approved):

PR #2312, opened 2022-08-09
refactor(atomic): extract code for atomic refine modals

liviarett:
https://coveord.atlassian.net/browse/KIT-1893

### DIFF
--- a/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
+++ b/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
@@ -1,0 +1,138 @@
+import {h, VNode} from '@stencil/core';
+import CloseIcon from 'coveo-styleguide/resources/icons/svg/close.svg';
+import {FacetManager, QuerySummary} from '@coveo/headless';
+import {InsightFacetManager, InsightQuerySummary} from '../../insight';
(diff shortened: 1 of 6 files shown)


packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  constructor(props: RefineModalCommonOptions) {
+    this.host = props.host;
+    this.bindings = props.bindings;
+    this.querySummary = props.initializeQuerySummary();
+    this.onClose = props.onClose;
+  }
+
+  public showModal() {
+    this.host.style.display = '';
```
btaillon, line 39: Not sure why the refine modal was setting its `display` attribute. I would expect `atomic-modal` to take care of everything.

btaillon (approved):

ThibodeauJF (approved):

packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
liviarett: good point 🤔  let me have a look

liviarett: actually, let's have a look at that separately, I want to try and fit the full insight feature before I leave!

PR #2314, opened 2022-08-10
feat(atomic): add pager to insight panel

liviarett:
https://coveord.atlassian.net/browse/KIT-1938

turned out to be easier than I thought, so I did it!

https://user-images.githubusercontent.com/30511433/183876043-4d7fd106-1c78-4f0d-9228-1c8f3f349f5c.mov

### DIFF
--- a/packages/atomic/src/components/common/pager/pager-common.tsx
+++ b/packages/atomic/src/components/common/pager/pager-common.tsx
@@ -1,26 +1,24 @@
-import {Pager, SearchStatus} from '@coveo/headless';
 import {EventEmitter, h} from '@stencil/core';
 import ArrowRight from '../../../images/arrow-right.svg';
 import {FocusTargetController} from '../../../utils/accessibility-utils';
 import {randomID} from '../../../utils/utils';
-import {InsightPager, InsightSearchStatus} from '../../insight';
-import {InsightBindings} from '../../insight/atomic-insight-interface/atomic-insight-interface';
-import {Bindings} from '../../search/atomic-search-interface/atomic-search-interface';
 import {Button} from '../button';
+import {AnyBindings} from '../interface/bindings';
 import {RadioButton} from '../radio-button';
+import {Pager, SearchStatus} from '../types';
 
 interface SearchStatusOptions {
-  bindings: Bindings | InsightBindings;
-  initializeSearchStatus(): SearchStatus | InsightSearchStatus;
-  initializePager(): Pager | InsightPager;
+  bindings: AnyBindings;
+  initializeSearchStatus(): SearchStatus;
+  initializePager(): Pager;
   getEventEmitter(): EventEmitter | undefined;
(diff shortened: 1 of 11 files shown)


packages/atomic/src/pages/index.html
```diff
@@ -369,11 +369,10 @@
...
           </atomic-layout-section>
           <atomic-layout-section section="pagination">
-            <atomic-load-more-results></atomic-load-more-results>
-            <!-- Alternative pagination
-              <atomic-pager></atomic-pager>
-              <atomic-results-per-page></atomic-results-per-page>
-              -->
+            <!-- <atomic-load-more-results></atomic-load-more-results>
+            Alternative pagination -->
+            <atomic-pager></atomic-pager>
```
ThibodeauJF, line 374: 🙅 😄  load more result is recommended to be used by UX for search pages

btaillon (approved):

packages/atomic/src/components/common/pager/pager-common.tsx
```diff
@@ -0,0 +1,136 @@
...
+interface SearchStatusOptions {
+  bindings: Bindings | InsightBindings;
+  initializeSearchStatus(): SearchStatus | InsightSearchStatus;
+  initializePager(): Pager | InsightPager;
+  getEventEmitter(): EventEmitter | undefined;
+  getActivePage(): FocusTargetController | undefined;
+}
+
+export class PagerCommon {
+  private bindings: Bindings | InsightBindings;
```
olamothe, line 21: This should be `AnyBindings`

packages/atomic/src/components/common/pager/pager-common.tsx
```diff
@@ -0,0 +1,136 @@
...
+  private searchStatus: SearchStatus | InsightSearchStatus;
```
olamothe, line 22: We should not have any specific types for the `common` components. It will quickly spiral out of control otherwise. 

As a rule of thumb, it should only import types from the `common` folder.  

The way it could be done is that instead of having a `searchStatus` coming from Headless, you declare a type/interface with a prop `hasResult: boolean`.

Since that's the only thing we care about in this file, coming from the `searchStatus`.

And then in search/pager you map the specific `SearchStatus` to the "generic" `hasResults` prop. Same thing for insight panel pager.

Does it make sense ?

packages/atomic/src/pages/index.html
liviarett: ah sorry, I was actually just testing and forgot to change it back 😇

packages/atomic/src/components/common/pager/pager-common.tsx
liviarett: ok, I actually did wonder about this @olamothe.. It does make sense, but I've done this for most components, so I might need to spend my last days redoing these types!

PR #2315, opened 2022-08-10
fix(atomic): initializeBindings fix + small segmented facet improvements

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1934


packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -135,6 +138,8 @@ export function InitializeBindings<SpecificBindings extends AnyBindings>({
 
     component.componentWillLoad = function () {
       const element = getElement(this);
+      element.setAttribute(renderedAttribute, 'false');
```
ThibodeauJF, line 141: I use the element to store information has it appeared doing `component['something']` was only saved once per component type

btaillon (approved):

olamothe (approved):

packages/atomic/src/utils/initialization-utils.tsx
olamothe: Is this something we could have caught with a test ?

PR #2325, opened 2022-08-11
fix(atomic): update common types not to use headless

liviarett:
https://coveord.atlassian.net/browse/KIT-1949


packages/atomic/src/components/common/pager/pager-common.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+interface SearchStatusOptions {
+  bindings: AnyBindings;
+  initializeSearchStatus(): SearchStatus;
+  initializePager(): Pager;
+  getEventEmitter(): EventEmitter | undefined;
+  getActivePage(): FocusTargetController | undefined;
+}
+
+export class PagerCommon {
```
btaillon, line 18: Did you mean to include this in your PR?

packages/atomic/src/components/common/types.ts
```diff
@@ -0,0 +1,238 @@
+export type SearchStatusState = {
```
btaillon, line 1: Rather than copying all these types from Headless, perhaps we could re-export them along with their documentation

packages/atomic/src/components/common/pager/pager-common.tsx
liviarett: no, I don't know what I've done with these branches, complete mess!!

ThibodeauJF (approved):
What Benjamin commented on, otherwise looks good!

packages/atomic/src/components/common/types.ts
olamothe: Let's discuss this in scrum. I'm not convinced this is the right approach.

ThibodeauJF: For now I re-exported the types from the basic @coveo/headless which are still valid so far...

btaillon: I recall that a lot of those types already have a "generic" version of them, called "Core*". Perhaps we could re-export those.

btaillon (approved):

packages/atomic/src/components/common/types.ts
```diff
@@ -1,17 +1,41 @@
...
-  selectPage(page: number): void;
-  previousPage(): void;
-  nextPage(): void;
-  state: {
-    hasNextPage: boolean;
-    hasPreviousPage: boolean;
-    currentPages: number[];
-  };
-};
+export {
```
ThibodeauJF, line 1:
```suggestion
// TODO: revisit using Core Headless Interfaces when using other sub-packages
export {
```

PR #2326, opened 2022-08-11
fix(atomic): improved atomic-aria-live for simultaneous messages

btaillon:
https://coveord.atlassian.net/browse/KIT-1927

With the upcoming `atomic-notification` component, I noticed that screen readers weren't reading all aria-live updates, so I did a bit of investigation.

I made [a test page in codesandbox](https://zpwdgl.csb.app/) with React to test how different screen readers interpret changes to aria-live elements. Here are the results of my tests:
<table>
<tr>
  <th></th>
  <th>JAWS</th>
  <th>NVDA</th>
  <th>macOS VoiceOver</th>
  <th>iOS VoiceOver</th>
  <th>TalkBack</th>
<tr>
  <th>1 aria-live and no delay
  <td>Says the last message.
  <td>Says the last message.
  <td>Says the last message.
  <td>Says the last message.
  <td>Says the last message.
<tr>
  <th>1 aria-live and a delay
  <td>Flaky, needs a delay of at-least 5ms.
  <td>Flaky, needs a delay of at-least 15ms.
  <td>Interrupts previous message.
  <td>Flaky, needs a delay of at-least 30ms.
  <td>Flaky, needs a delay of at-least 25ms.
<tr>
  <th>2 aria-live and no delay
  <td>Read in varying DOM order (randomized per refresh).
  <td>Read in varying DOM order (randomized per refresh).
  <td>Read in varying DOM order (randomized per refresh).
  <td>Read in DOM order.
  <td>Read in change order.
<tr>
  <th>2 aria-live and a delay
  <td>Read in change order (even at 0ms).
  <td>Read in varying DOM order (randomized per refresh). Read in change order when delay >= 15ms.
  <td>Read in varying DOM order (randomized per refresh). Read in change order when delay >= 150ms.
  <td>Read in change order (even at 0ms).
  <td>Read in change order.
</table>

Based on those results, I decided to do the following:
* Instead of setting aria-live directly on the `atomic-aria-live` element, I divided it into separate elements per "region". It's ok for screen readers not to read the previous state of a region, we only care about the current state, just like we generally do for a visual component.
* I introduced a utility to debounce aria message updates. It ensures that we don't update the aria message too fast. It doesn't allow regions to queue up multiple messages, it instead removes the previous message from the same region.
* I improved `atomic-aria-live` by allowing regions to "register" themselves so that an `atomic-aria-live` component doesn't need to be included outside of the `atomic-search-interface` anymore.

### DIFF
--- a/packages/atomic/src/components/search/atomic-search-interface/atomic-aria-live.tsx
+++ b/packages/atomic/src/components/search/atomic-search-interface/atomic-aria-live.tsx
@@ -1,13 +1,20 @@
-import {Component, h, Host, State, Element, Method} from '@stencil/core';
 import {
-  FindAriaLiveEventArgs,
-  findAriaLiveEventName,
-} from '../../../utils/accessibility-utils';
+  Component,
+  h,
+  Host,
+  State,
+  Element,
+  Method,
+  Listen,
+} from '@stencil/core';
+import {FindAriaLiveEventArgs} from '../../../utils/accessibility-utils';
+import {buildDebouncedQueue} from '../../../utils/debounce-utils';
+import {randomID} from '../../../utils/utils';
+
+type Regions = {[regionName: string]: {assertive: boolean; message: string}};
 
 /**
  * The `atomic-aria-live` component notifies screen readers of changes in the search interface.
- *
- * We do not recommend dynamically adding/removing this component from the page. It is better to have this component during the initial render of the page.
  */
 @Component({
   tag: 'atomic-aria-live',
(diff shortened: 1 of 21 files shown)


packages/atomic/src/components/search/atomic-search-interface/atomic-aria-live.tsx
```diff
@@ -1,27 +1,34 @@
...
  */
 @Component({
   tag: 'atomic-aria-live',
   shadow: false,
 })
 export class AtomicAriaLive {
   @Element() private host!: HTMLAtomicAriaLiveElement;
-  @State() private message = '';
-  @State() private assertive = false;
+  @State() private regions: Readonly<Regions> = {};
```
btaillon, line 25: A re-render is only triggered when we replace (`=`) the object, so I made it `Readonly` to ensure we don't just mutate it.

packages/atomic/src/utils/debounce-utils.tsx
```diff
@@ -0,0 +1,52 @@
...
+  let actions: QueuedAction[] = [];
+  let intervalId: ReturnType<typeof setInterval> | null = null;
+
+  function dequeueAction() {
+    const action = actions.shift();
+    if (action) {
+      action.execute();
+    } else {
+      clearInterval(intervalId!);
+      intervalId = null;
```
btaillon, line 28: The interval is only cleared once it's triggered on an empty list. That's on purpose.

ThibodeauJF (approved):

PR #2327, opened 2022-08-11
feat(atomic,headless): add support for custom sort for facets

olamothe:
There's some bug in the API still regarding the feature.

It does not properly handle case sensitivity for facet values, so in some cases the values are not ordered as they should be.

Ticket in search API: https://coveord.atlassian.net/browse/SEARCHAPI-7272

https://coveord.atlassian.net/browse/KIT-753

### DIFF
--- a/packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
+++ b/packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
@@ -113,6 +113,16 @@ export class AtomicFacet implements InitializableComponent, BaseFacet<Facet> {
    */
   @Prop({reflect: true}) public sortCriteria: FacetSortCriterion = 'automatic';
   /**
+   * Specifies a custom order by which to sort the facet values.
+   *
+   * Example:
+   * ```html
+   * <!-- Sort @ticketstatus facet values in a logical order for support tickets -->
+   * <atomic-facet field="@ticketstatus" custom-sort="New,Opened,Feedback,Resolved"></atomic-facet>
+   * ```
+   */
+  @Prop({reflect: true}) public customSort?: string;
+  /**
    * Whether to display the facet values as checkboxes (multiple selection), links (single selection) or boxes (multiple selection).
    * Possible values are 'checkbox', 'link', and 'box'.
    */
@@ -137,8 +147,6 @@ export class AtomicFacet implements InitializableComponent, BaseFacet<Facet> {
    * Default: `1000`
    */
   @Prop() public injectionDepth = 1000;
-  // @Prop() public customSort?: string; TODO: KIT-753 Add customSort option for facet
-
   /**
    * The required facets and values for this facet to be displayed.
    * Examples:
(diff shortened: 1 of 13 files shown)


packages/headless/src/controllers/core/facets/facet/headless-core-facet-options.ts
```diff
@@ -89,14 +91,63 @@ export interface FacetSearchOptions {
...
+  if (!sortCriteria) {
+    return;
+  }
+
+  const schemaForSortCriteria = getSortCriteriaPolymorphicSchema(sortCriteria);
+
+  validateOptions(engine, schemaForSortCriteria!, {sortCriteria}, functionName);
+}
+
+export function getSortCriteriaPolymorphicSchema(
```
ThibodeauJF, line 145: Do we add a v2 todo to remove the string format?

ThibodeauJF (approved):

packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
```diff
@@ -112,6 +112,16 @@ export class AtomicFacet implements InitializableComponent, BaseFacet<Facet> {
...
+  /**
+   * Specifies a custom order by which to sort the facet values.
+   *
+   * Example:
+   * ```html
+   * <!-- Sort @ticketstatus facet values in a logical order for support tickets -->
+   * <atomic-facet field="@ticketstatus" custom-sort="New,Opened,Feedback,Resolved"></atomic-facet>
+   * ```
+   */
+  @Prop({reflect: true}) public customSort?: string;
```
btaillon, line 124: Instinctively, this sounds more like "pinning" values to me. Additionally, I find it unclear what happens to values that aren't included in the custom sort. Are they hidden/removed? Are they sorted alphanumerically?

btaillon: I recall that Stencil supports arrays of strings. Do they split on commas?

packages/headless/src/controllers/core/facets/facet/headless-core-facet-options.ts
```diff
@@ -89,14 +91,63 @@ export interface FacetSearchOptions {
...
+  sortCriteria: FaceSortCriterionStringOrExplicit
+) {
+  if (typeof sortCriteria === 'string') {
+    return facetOptionSortCriteriaStringSchema;
+  }
+
+  return facetOptionSortCriteriaExplicitSchema;
```
btaillon, line 152: Does this mean that we can either custom sort or do a full sort, but not both?

From the JIRA on the search API, they seem to imply that the purpose of the feature is to be able to sort the top values (pin them) and still see more values below those. I would expect that we could (eventually) specify both properties independently.

olamothe: You can see the doc here: 

https://docs.coveo.com/en/13/api-reference/search-api#tag/Search-V2/operation/searchUsingPost

Facets -> Sort criteria -> Click on the `one of` small button.

packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
olamothe: customSort is how it's called in the API. Not pinning.

> Additionally, I find it unclear what happens to values that aren't included in the custom sort. Are they hidden/removed? Are they sorted alphanumerically?

Specified values in customSort will appear first in the facet. Then the API will fill remaining values until `numberOfValues` is reached using either score or alphanumeric, depending on the state of the facet (expanded, collapsed).

> I recall that Stencil supports arrays of strings. Do they split on commas?

This is mimicking the same way we do it for all other such option in Atomic. Either we change it everywhere, or we keep the same pattern.

And the old pattern has served us very well so far, so I see no reason to invest time to change it.

olamothe:
Will re-open after API changes

PR #2332, opened 2022-08-12
feat(headless): added undo method to query trigger

btaillon:
https://coveord.atlassian.net/browse/KIT-1932

What I did:
* Added undo method to controller
  * All it does is update the query and trigger a search
* Prevented queries from being triggered when the original query is queried directly after a correction
* Added trigger analytics to action loaded
* Added query trigger undo analytics

I tested this with Atomic (and will include most extensive cypress tests with Atomic).
The following worked:
* Undoing a query correction by calling the method.
* Undoing a query correction by pressing the browser back button.
* Undoing a query correction by manually searching the original query again.

What doesn't quite work:
* Refreshing the page after pressing the back button. This causes the trigger to change the query again.


jpmarceau (approved):
👌

PR #2333, opened 2022-08-12
feat(quantic): added aria-pressed for pager

MSLI-Coveo:
[2306](https://coveord.atlassian.net/browse/SVCC-2306)

Facet Values’ aria-pressed is implemented along with `SVCC-2272`

Both QuanticPager and QuanticResultsPerPage render quanticNumberButton, so changes are applied to quanticNumberButton. `<lightning-button>` is replaced by `<button>`, since it does not support "aria-pressed".

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.html
+++ b/packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.html
@@ -1,8 +1,3 @@
 <template>
-  <template if:true={selected}>
-    <button class="slds-button slds-button_brand slds-m-left_xx-small" aria-pressed="true" key={pageNumber} onclick={select}>{number}</button>
-  </template>
-  <template if:false={selected}>
-    <button class="slds-button slds-button_outline-brand slds-m-left_xx-small" aria-pressed="false" key={pageNumber} onclick={select}>{number}</button>
-  </template>
+  <button class={buttonClasses} aria-pressed={isPressed} key={pageNumber} onclick={select}>{number}</button>
 </template>
--- a/packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.html
+++ b/packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.html
@@ -1,11 +1,8 @@
 <template>
-  <lightning-button
-    key={pageNumber}
-    variant={variant}
-    label={number}
-    title={number}
-    class="slds-m-left_xx-small"
-    value={number}
-    onclick={select}
-  ></lightning-button>
+  <template if:true={selected}>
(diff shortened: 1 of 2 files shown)


btaillon (approved):
it may be a good idea to add me to PRs that relate to accessibility

ThibodeauJF:
Yes he's the expert 😉

packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.js
```diff
@@ -21,8 +21,8 @@ export default class QuanticNumberButton extends LightningElement {
    */
   @api selected;
 
-  get variant() {
-    return this.selected ? 'brand' : 'brand-outline';
+  get isPressed() {
```
lbergeron, line 24: This property is not used.

packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.html
```diff
@@ -1,11 +1,8 @@
...
-    onclick={select}
-  ></lightning-button>
-</template>
+  <template if:true={selected}>
+    <button class="slds-button slds-button_brand slds-m-left_xx-small" aria-pressed="true" key={pageNumber} onclick={select}>{number}</button>
+  </template>
+  <template if:false={selected}>
+    <button class="slds-button slds-button_outline-brand slds-m-left_xx-small" aria-pressed="false" key={pageNumber} onclick={select}>{number}</button>
+  </template>
+</template>
```
lbergeron, line 8: Instead of using conditional rendering, I'd suggest to add a property to return the CSS classes to apply. I think it would look simpler overall. It is currently hard to spot the differences between what is rendered when the button is pressed or not.

```html
<button class={buttonClasses} aria-pressed={selected} key={pageNumber} onclick={select}>{number}</button>
```

```javascript
get buttonClasses() {
  const classes = ['slds-button', 'slds-m-left__xx-small'];
  classes.push(this.selected ? 'slds-button_brand' : 'slds-button_outline-brand');
  return classes.join(' ');
}
```

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.html
MSLI-Coveo: Removed conditional rendering and added getter for attribute class and aril-pressed.

packages/quantic/force-app/main/default/lwc/quanticNumberButton/quanticNumberButton.js
MSLI-Coveo: Used isPressed in the new implementation as conditional rendering for these btns is removed.

lbergeron (approved):

PR #2336, opened 2022-08-15
feat(atomic): added query trigger to `atomic-did-you-mean` component

btaillon:
https://coveord.atlassian.net/browse/KIT-1932

image

Hovered:
image

I also changed the manual did you mean appearance (based on Johnny's feedback).
<table>
  <tr>
    <th>Before
    <th>After
  <tr>
    <td>

image
    <td>

image

### DIFF
--- a/packages/atomic/src/locales.json
+++ b/packages/atomic/src/locales.json
@@ -612,6 +612,14 @@
     "zh-CN": "没有找到与 {{query}} 相关的结果",
     "zh-TW": "沒有找到與 {{query}} 相關的結果"
   },
+  "showing-results-for": {
+    "en": "Showing results for {{query}}",
+    "fr": "Résultats affichés pour {{query}}"
+  },
+  "search-instead-for": {
+    "en": "Search instead for {{query}}",
+    "fr": "Rechercher à la place {{query}}"
+  },
   "search-tips": {
     "en": "You may want to try using different keywords, deselecting filters, or checking for spelling mistakes.",
     "fr": "Essayez différents mots-clés, déselectionnez des filtres ou vérifiez l'orthographe de vos mots-clés."
--- a/packages/atomic/src/utils/jsx-utils.tsx
+++ b/packages/atomic/src/utils/jsx-utils.tsx
@@ -1,0 +1,44 @@
+import {Fragment, FunctionalComponent, h, VNode} from '@stencil/core';
+import {AnyBindings} from '../components/common/interface/bindings';
+
+export interface LocalizedStringProps {
+  bindings: AnyBindings;
+  key: string;
+  params: Record<string, VNode | string>;
+}
+
+export const LocalizedString: FunctionalComponent<LocalizedStringProps> = ({
+  bindings,
+  key,
+  params,
+}) => {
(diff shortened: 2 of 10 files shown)


packages/atomic/src/utils/jsx-utils.tsx
```diff
@@ -0,0 +1,44 @@
+import {Fragment, FunctionalComponent, h, VNode} from '@stencil/core';
+import {AnyBindings} from '../components/common/interface/bindings';
+
+export interface LocalizedStringProps {
+  bindings: AnyBindings;
+  key: string;
+  params: Record<string, VNode | string>;
+}
+
+export const LocalizedString: FunctionalComponent<LocalizedStringProps> = ({
```
btaillon, line 10: This component inserts JSX into a localized string without the need for `innerHTML` by returning an array of `VNode | string`.

packages/atomic/src/utils/jsx-utils.tsx
```diff
@@ -0,0 +1,44 @@
...
+  bindings,
+  key,
+  params,
+}) => {
+  const delimitingCharacter = '\u001d'; // Unicode group separator
+  const placeholderPrefixCharacter = '\u001a'; // Unicode substitute character
```
btaillon, line 16: I used unicode control characters since I don't expect those to be used in any locales.

packages/atomic/src/locales.json
```diff
@@ -612,6 +612,14 @@
     "zh-CN": "没有找到与 {{query}} 相关的结果",
     "zh-TW": "沒有找到與 {{query}} 相關的結果"
   },
+  "showing-results-for": {
+    "en": "Showing results for {{query}}",
+    "fr": "Résultats affichés pour {{query}}"
+  },
+  "search-instead-for": {
+    "en": "Search instead for {{query}}",
```
jpmarceau, line 620:
```suggestion
    "en": "Search for {{query}} instead",
```

packages/atomic/src/locales.json
```diff
@@ -612,6 +612,14 @@
...
+    "fr": "Rechercher à la place {{query}}"
```
jpmarceau, line 621:
```suggestion
    "fr": "Rechercher {{query}} à la place"
```

jpmarceau (approved):
👍

ThibodeauJF (approved):
Very nice!

packages/atomic/src/locales.json
btaillon: Discussed it with Johnny, and visually it's better to keep the query at the end.

jpmarceau: Oh I see, no worries. This one works decently well both ways.

packages/atomic/src/locales.json
jpmarceau: If {{query}} must stay at the end, do you think it would be possible to go:
```suggestion
    "fr": "Rechercher plutôt {{query}}"
```
Not a dealbreaker, but it's grammattically better 🙂

olamothe (approved):

PR #2337, opened 2022-08-15
chore(atomic): os-agnostic filecheck

louis-bompart:
https://coveord.atlassian.net/browse/KIT-1961

Previous steps (should) already have emitted the types, so we just need to check the files exists and that it's valid.
This tsc command will do it.

@lvu285, if you can confirm that it addresses your problem, it'd be nice ;)


btaillon (dismissed):
Thanks!

ThibodeauJF (approved):
amazing

louis-bompart:
Sorry fellas, outta ask for your review once more, found a simpler way to do it.

btaillon (approved):

PR #2338, opened 2022-08-15
fix(atomic): made search box suggestions usable on iOS VoiceOver

btaillon:
https://coveord.atlassian.net/browse/KIT-1941

iOS VoiceOver was calling the input's `onBlur`, which cleared suggestions.

Instead of detecting `blur` on the search input, I'm detecting when the entire search box component is blurred, and closing suggestions when `tab` is pressed to maintain the normal behaviour on desktop.


ThibodeauJF (approved):
Wow VoiceOver making thing so complicated

PR #2340, opened 2022-08-15
feat(atomic): allowed customization of instant results aria-label

btaillon:
https://coveord.atlassian.net/browse/KIT-1909

Also added examples for react and angular.

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,6 +1,10 @@
+/** @type {import('prettier').Config} */
 module.exports = {
   bracketSpacing: false,
   singleQuote: true,
   trailingComma: 'es5',
-  endOfLine: "auto"
+  endOfLine: 'auto',
+  overrides: [
+    {files: '**/*.{scss,css,pcss,html,mdx}', options: {printWidth: 120}},
+  ],
 };
--- a/packages/atomic-angular/projects/atomic-angular/src/public-api.ts
+++ b/packages/atomic-angular/projects/atomic-angular/src/public-api.ts
@@ -3,4 +3,5 @@
  */
 export * from '@coveo/atomic/headless';
 export * from './lib/stencil-generated/components';
+export {Bindings, i18n} from '@coveo/atomic';
 export {AtomicAngularModule} from './lib/stencil-generated/atomic-angular.module';
--- a/packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
+++ b/packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
@@ -24,6 +24,12 @@ import {
   ResultDisplayLayout,
 } from '../../../common/layout/display-options';
 import {Button} from '../../../common/button';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+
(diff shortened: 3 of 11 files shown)


.prettierrc.js
```diff
@@ -1,6 +1,10 @@
...
   bracketSpacing: false,
   singleQuote: true,
   trailingComma: 'es5',
-  endOfLine: "auto"
-};
\ No newline at end of file
+  endOfLine: 'auto',
+  overrides: [
+    {files: '**/*.{scss,css,pcss,html,mdx}', options: {printWidth: 120}},
+  ],
```
btaillon, line 9: Not extremely related to the PR. I moved this from `package.json` so that we can format any file locally without committing.

packages/atomic-angular/projects/atomic-angular/src/public-api.ts
```diff
@@ -3,4 +3,5 @@
  */
 export * from '@coveo/atomic/headless';
 export * from './lib/stencil-generated/components';
+export {Bindings, i18n} from '@coveo/atomic';
```
btaillon, line 6: These interfaces weren't exported by `@coveo/atomic-angular` and `@coveo/atomic-react`, which means that we could only import from from `@coveo/atomic`, which gives a [`no-extraneous-import`](https://github.com/mysticatea/eslint-plugin-node/blob/v11.1.0/docs/rules/no-extraneous-import.md) error with the recommended ESLint config.

packages/atomic/src/components/search/search-box-suggestions/atomic-search-box-instant-results/atomic-search-box-instant-results.tsx
```diff
@@ -24,6 +24,12 @@ import {
   ResultDisplayLayout,
 } from '../../../common/layout/display-options';
 import {Button} from '../../../common/button';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+
+export type AriaLabelGenerator = (
+  bindings: Bindings,
```
btaillon, line 30: I'm using the **search** bindings specifically, and not just `i18n`, since I don't want to add extra parameters if developers ever need more, and bindings contain the engine, which is useful.

btaillon: Also, the instant results component uses the search bindings currently, it's not compatible with other use cases than search.

packages/atomic/src/pages/accessibility/commerce-full.html
```diff
@@ -69,6 +69,36 @@
...
+          if ('ec_price' in result.raw) {
+            information.push(
+              result.raw.ec_price.toLocaleString(i18n.languages, {
+                style: 'currency',
+                currency: 'USD',
+              })
+            );
+          }
+
+          return information.join(', ');
```
btaillon, line 100: Renders the following labels in the DOM:
* `"Apple Pencil (1st Generation), 4.7 stars out of 5, $99.99, instant result. Button. 1 of 5. In Right list."`
* `"Sport Band for Apple Watch™ 44mm, No ratings available, $49.00, instant result. Button. 3 of 5. In Right list."`

ThibodeauJF (approved):
Clean!

PR #2341, opened 2022-08-15
feat(atomic, headless): clear segmented facet (hasBreadcrumb) values on new query

ThibodeauJF:
After a discussion with UX...
If we want to provide an option we'd do it for all facets as we previously discussed

https://coveord.atlassian.net/browse/KIT-1965


btaillon:
Wouldn't this be breaking?

Perhaps we could keep the option in Headless and just make `hasBreadcrumbs` always `true` in Atomic.

PR #2344, opened 2022-08-16
doc(atomic): missing atomic-icon-button annotations

jpmarceau:
As I generated the 1.95.0 docs, I noticed that this component did not have annotations, except for some parts.
I created this PR to fix it 🙂 
I don't really understand the last two though, i.e., `buttonRef` & `badge` 🤔 Could you please suggest something?


ThibodeauJF:
Hmm perhaps this should be internal we don't need to expose this one to customers

btaillon (approved):

ThibodeauJF (approved):

PR #2347, opened 2022-08-16
feat(atomic): improve atomic-popover interaction

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1937

Now has
- semi-transparent backdrop which closes on click
- close on ESC key
- focus trap for the button/facet

### DIFF
--- a/packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
+++ b/packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
@@ -4,15 +4,7 @@ import {
   SearchStatusState,
   FacetState,
 } from '@coveo/headless';
-import {
-  Component,
-  h,
-  Listen,
-  State,
-  Element,
-  Event,
-  EventEmitter,
-} from '@stencil/core';
+import {Component, h, Listen, State, Element, Host} from '@stencil/core';
 import {
   BindStateToController,
   InitializableComponent,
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
```diff
@@ -211,17 +221,16 @@ export class AtomicPopover implements InitializableComponent {
...
     return (
-      <div class="relative">
-        {this.renderDropdownButton()}
-        <div
-          id={this.popoverId}
-          ref={(el) => (this.popupRef = el!)}
-          part="facet"
-          class={`absolute pt-0.5 z-1 ${this.isOpen ? 'block' : 'hidden'}`}
+      <Host>
+        <atomic-focus-trap
```
ThibodeauJF, line 225: Do I need extra aria attributes here?

btaillon: no need for extra aria attributes. However, when the modal opens, the focus gets completely stuck in macOS Voiceover

btaillon (approved):
make sure this works with all screen readers (as well as the tab key) and make sure to show this to Johnny. Personally, I like the backdrop more than I thought I would.

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
ThibodeauJF: Screen Shot 2022-08-17 at 8 31 20 AM

ThibodeauJF:
I only have voiceover for mac not the other desktop ones & i'll send Johnny the link when it's merged

packages/atomic/src/components/search/facets/atomic-popover/atomic-popover.tsx
ThibodeauJF: Hmm with the VO - up/down key I cannot even access the button in the first place... But with tab it worked i'll look into it

ThibodeauJF: Got it!

PR #2350, opened 2022-08-18
feat(quantic): new QuanticColoredResultBadge component created

mmitiche:
## Colored Result Badge component

- This component will be used inside the result template of the Insight Panel to display the result badges.
ResultTemplate

- In the Insight Panel configuration we receive the following information about the badge to display for each result template: 

```
badge: {
  label: "Document",
  color: "#DABFE9"
}
```
- This component will create a badge with the correct label and the correct color.
- This component contains the logic to create a secondary color from the given primary color.
- The given primary color will be used as the background color of the badge and the secondary color will be used as the text color as well as the border color.
- This component will first convert the given Hex color to RGB, then will convert the RGB color to [HSL](https://www.w3schools.com/colors/colors_hsl.asp)(Hue, Saturation, Lightness) then will update the lightness of this color to generate the secondary color.

#### Examples of the badges created with diffrent colors:
ColoredResultBadge

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticColoredResultBadge/__tests__/quanticColoredResultBadge.test.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticColoredResultBadge/__tests__/quanticColoredResultBadge.test.js
@@ -1,0 +1,180 @@
+import {createElement} from 'lwc';
+import QuanticColoredResultBadge from '../quanticColoredResultBadge';
+
+const defaultPrimaryColor = '#FFF7BA';
+const defaultSecondaryColor = '#E2B104';
+const defaultLabel = 'Case';
+
+const defaultResult = {
+  raw: {
+    fieldOne: 'valueOne',
+  },
+};
+
+const colorsToTest = [
+  {
+    primaryColor: '#DABFE9',
+    expectedSecondaryColor: 'hsl(279, 48.8%, 35.1%)',
+  },
+  {
+    primaryColor: '#EF233C',
+    expectedSecondaryColor: 'hsl(353, 86.4%, 5.7%)',
+  },
+  {
+    primaryColor: '03A06E',
+    expectedSecondaryColor: 'hsl(161, 96.3%, 80.0%)',
+  },
+  {
+    primaryColor: 'FFF',
+    expectedSecondaryColor: 'hsl(0, 0%, 52.0%)',
+  },
+];
+
+const invalidColors = [
+  {
+    primaryColor: '#ZZFFFF',
+    portion: 'red',
+  },
+  {
+    primaryColor: '#FFZZFF',
+    portion: 'green',
+  },
+  {
+    primaryColor: '#FFFFZZ',
+    portion: 'blue',
+  },
+];
+
+const invalidPropertiesError =
+  '"QuanticColoredResultBadge" requires either specified value for label or a result object with a fieldname to display correctly.';
+
+function createTestComponent(color, label = defaultLabel, result, fieldname) {
+  const element = createElement('c-colored-result-badge', {
+    is: QuanticColoredResultBadge,
+  });
+  element.label = label;
+  element.color = color;
+  element.result = result;
+  element.fieldname = fieldname;
+
+  document.body.appendChild(element);
+
+  return element;
+}
+
+// Helper function to wait until the microtask queue is empty.
+function flushPromises() {
+  // eslint-disable-next-line no-undef
+  return new Promise((resolve) => setImmediate(resolve));
+}
+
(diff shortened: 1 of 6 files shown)


packages/quantic/force-app/main/default/lwc/quanticColoredResultBadge/__tests__/quanticColoredResultBadge.test.js
```diff
@@ -0,0 +1,180 @@
...
+    const element = createTestComponent('#FFF7BA', label);
+    await flushPromises();
+
+    const badge = element.shadowRoot.querySelector('.result-badge');
+
+    expect(badge).not.toBeNull();
+    expect(badge.textContent).toBe(label);
+  });
+
+  it('should display the colored badge with the correct label and color the result and fieldname properties are given', async () => {
```
mmitiche, line 95:
```suggestion
  it('should display the colored badge with the correct label and color when the result and fieldname properties are given', async () => {
```

packages/quantic/force-app/main/default/lwc/quanticColoredResultBadge/__tests__/quanticColoredResultBadge.test.js
```diff
@@ -0,0 +1,180 @@
...
+    primaryColor: '#FFFFZZ',
+    portion: 'blue',
+  },
+];
+
+const invalidPropertiesError =
+  '"QuanticColoredResultBadge" requires either specified value for label or a result object with a fieldname to display correctly.';
+
+function createTestComponent(color, label = defaultLabel, result, fieldname) {
+  const element = createElement('c-colored-result-badge', {
```
mmitiche, line 52:
```suggestion
  const element = createElement('c-quantic-colored-result-badge', {
```

packages/quantic/force-app/main/default/lwc/quanticColoredResultBadge/quanticColoredResultBadge.js
```diff
@@ -0,0 +1,113 @@
...
+
+const defaultPrimaryColor = '#FFF7BA';
+const defaultSecondaryColor = '#E2B104';
+const lightnessDegree = 48;
+
+/** @typedef {import("coveo").Result} Result */
+
+/**
+ * The `QuanticColoredResultBadge` component displays a colored badge showing a label.
+ * If the `Label` property is set in this component, its value will be displayed in the badge, Otherwise the `Result` and 'Fieldname' properties are required and the value of the given field will lbe displayed.
```
mmitiche, line 12:
```suggestion
 * If the `Label` property is set in this component, its value will be displayed in the badge, Otherwise the `Result` and `Fieldname` properties are required and the value of the given field will lbe displayed.
```

packages/quantic/force-app/main/default/lwc/quanticColoredResultBadge/__tests__/quanticColoredResultBadge.test.js
```diff
@@ -0,0 +1,180 @@
...
+    portion: 'green',
+  },
+  {
+    primaryColor: '#FFFFZZ',
+    portion: 'blue',
+  },
+];
+
+const invalidPropertiesError =
+  '"QuanticColoredResultBadge" requires either specified value for label or a result object with a fieldname to display correctly.';
```
lbergeron, line 49: So you decided to put the logic into the badge component directly. What made you change your mind?

lbergeron (commented):
I'm not sure about having the "label vs field" logic built into the badge component. I liked the idea of having a simple badge component that just displays the text we're giving it.

packages/quantic/force-app/main/default/lwc/quanticColoredResultBadge/__tests__/quanticColoredResultBadge.test.js
mmitiche: Mainly I wanted to follow the same pattern used in the QuanticLabel component which has this `label vs field` logic, but on a second thought now I think you are right, having a simple component that will simply display the label passed to him is far better. And the logic to choose between displaying a label or field value will only be at a higher level, for example in the insight Panel it will be handled in the `ConfigurableResult` component.
Thank you for your suggestion I'll update this.👌

mmitiche: Component simplified 👌 the badge will now just display  the text we're giving it.

lbergeron (approved):
Looks good to me!

aelfaleh (approved):

PR #2355, opened 2022-08-19
[DO NOT REVIEW] COMSDZ-4: Test GraphQl client for Product listing

thudon-coveo:
[Jira Ticket](https://coveord.atlassian.net/browse/COMSDZ-4)

So this is what it could look like if we'd use GraphQL instead of REST in Headless.

This exemple doesn't use the commerce-service's context so there is a lot of stuff left out.

Also I didn't remove the REST part I just added the GraphQL part and changed the action.

Instead it's using a query to the content-api that returns a _valid error_.
You can try it [here](https://api.qubit.com/placements/)

```
{
    placementContent(
      mode: PREVIEW
      previewOptions: {campaignId: "foo"}
      placementId: "foo"
      attributes: {visitor: {id: ""}}
    ) {
      content
      visitorId
      callbackData
    }
  }
```
you should get: 

```
{
  "errors": [
    {
      "message": "look up property: sql: no rows in result set",
      "path": [
        "placementContent"
      ]
    }
  ],
  "data": {
    "placementContent": null
  }
}
```
I tested it locally on the [SAQdemo](https://github.com/coveo/saqdemo)

image

### DIFF
--- a/packages/headless/src/api/commerce/product-listings/product-listing-api-client.ts
+++ b/packages/headless/src/api/commerce/product-listings/product-listing-api-client.ts
@@ -15,6 +15,28 @@ import {
   ProductListingRequest,
   ProductListingSuccessResponse,
 } from './product-listing-request';
+// eslint-disable-next-line node/no-extraneous-import
+import {ApolloClient, InMemoryCache, gql} from '@Apollo/client';
+
+const QUERY = gql`
+  {
+    placementContent(
+      mode: PREVIEW
+      previewOptions: {campaignId: "foo"}
+      placementId: "foo"
+      attributes: {visitor: {id: ""}}
+    ) {
+      content
+      visitorId
+      callbackData
+    }
+  }
+`;
+
+const apolloClient = new ApolloClient({
+  cache: new InMemoryCache(),
+  uri: 'https://api.qubit.com/placements/query',
+});
 
 export interface AsyncThunkProductListingOptions<
   T extends Partial<ProductListingAppState>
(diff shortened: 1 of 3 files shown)


packages/headless/src/api/commerce/product-listings/product-listing-api-client.ts
```diff
@@ -15,6 +15,28 @@ import {
...
+      previewOptions: {campaignId: "foo"}
+      placementId: "foo"
+      attributes: {visitor: {id: ""}}
+    ) {
+      content
+      visitorId
+      callbackData
+    }
+  }
+`;
```
thudon-coveo, line 34: For this test I used the query Andrey sent in the channel. Obviously in a real implementation this wouldn't here and the values would not be hardcoded.

packages/headless/src/api/commerce/product-listings/product-listing-api-client.ts
```diff
@@ -15,6 +15,28 @@ import {
...
+
+const apolloClient = new ApolloClient({
+  cache: new InMemoryCache(),
+  uri: 'https://api.qubit.com/placements/query',
+});
```
thudon-coveo, line 39: This is the client needed for the GraphQL queries.

packages/headless/src/features/product-listing/product-listing-actions.ts
```diff
@@ -92,12 +92,9 @@ export const fetchProductListing = createAsyncThunk<
...
 >(
   'productlisting/fetch',
-  async (_action, {getState, dispatch, rejectWithValue, extra}) => {
-    const state = getState();
+  async (_action, {dispatch, rejectWithValue, extra}) => {
     const {apiClient} = extra;
-    const fetched = await apiClient.getProducts(
-      await buildProductListingRequest(state)
-    );
+    const fetched = await apiClient.getGraphQlProducts();
```
thudon-coveo, line 97: I had to remove the state for this exemple as it is tied to the commerce-service context.

PR #2356, opened 2022-08-19
fix(atomic): make range inputs (numeric/timeframe) appear on mobile

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1969

I basically simplified the way we determine the facetId for Atomic range/input facets which can contain up to 3 headless facets. The Id was lost or a new was being created when the facets were moved in the refine modal.


packages/atomic/src/components/common/facets/numeric-facet-common.tsx
```diff
@@ -67,70 +68,74 @@ interface NumericFacetCommonRenderProps {
 }
 
 export class NumericFacetCommon {
-  private host: HTMLElement;
```
ThibodeauJF, line 70: These were not really necessary, lots of redefinition & code duplication

packages/atomic/src/components/common/facets/numeric-facet-common.tsx
```diff
@@ -67,70 +68,74 @@ interface NumericFacetCommonRenderProps {
...
-        Array.from(this.host.querySelectorAll('atomic-numeric-range')).map(
-          ({start, end, endInclusive, label}) => ({
-            ...props.buildNumericRange({start, end, endInclusive}),
-            label,
-          })
-        )
-      );
-      this.facetForRange = props.initializeFacetForRange();
-      this.facetId = props.setFacetId(this.facetForRange.state.facetId);
+    const isIdDefinedForInput = !!this.facetId && !this.props.numberOfValues;
```
ThibodeauJF, line 88: Key fix is in that method, preserving the facet id properly

packages/atomic/src/components/search/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
```diff
@@ -206,9 +206,11 @@ export class AtomicNumericFacet
           conditions: parseDependsOn(this.dependsOn),
         }),
       buildNumericRange: buildNumericRange,
-      initializeFacetForInput: () => this.initializeFacetForInput(),
-      initializeFacetForRange: () => this.initializeFacetForRange(),
-      initializeFilter: () => this.initializeFilter(),
+      initializeFacetForInput: (facetId?: string) =>
```
ThibodeauJF, line 209: Common will determine the facet Ids

olamothe (approved):

PR #2360, opened 2022-08-22
fix(atomic,headless): fixed setRange on Firefox and Safari

btaillon:
https://coveord.atlassian.net/browse/KIT-1970

On Chrome, the Search API date format can be parsed implicitly, but on Firefox and Safari, dayjs won't parse the Search API date format. You can try [this demo](https://yi39he.csb.app/?date=2022%2F08%2F08%4019%3A00%3A00&withcustomparseformat=true) on different browsers to reproduce the issue.

This caused setRange to do nothing but clear the facet on Firefox and Safari. I tried to explicitly set the format for setRange, but this caused other errors to appear in the UI (such as in the breadcrumbs). Instead, I decided to try to make the behaviour as consistent between Chrome and other browsers as possible by adding the search API format as a fallback, by adding a `parseDate` util instead of calling dayjs directly. An alternative could be to make a custom plugin to extend dayjs directly, but I personally prefer not to call dayjs directly in the first place.

I also exported the Search API date format as a constant from Headless.

### DIFF
--- a/packages/headless/src/api/search/date/date-format.test.ts
+++ b/packages/headless/src/api/search/date/date-format.test.ts
@@ -1,8 +1,8 @@
-import dayjs from 'dayjs';
 import {
   isSearchApiDate,
   formatDateForSearchApi,
   validateAbsoluteDate,
+  parseDate,
 } from './date-format';
 
 describe('#isSearchApiDate', () => {
@@ -18,8 +18,8 @@ describe('#isSearchApiDate', () => {
 describe('#formatDateForSearchApi', () => {
   it('returns the correct format', () => {
     const date = 818035920000;
-    expect(formatDateForSearchApi(dayjs(date))).toBe(
-      dayjs(date).format('YYYY/MM/DD@HH:mm:ss')
+    expect(formatDateForSearchApi(parseDate(date))).toBe(
+      parseDate(date).format('YYYY/MM/DD@HH:mm:ss')
     );
   });
 });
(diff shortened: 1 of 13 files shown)


packages/headless/src/controllers/core/facets/range-facet/date-facet/date-range.ts
```diff
@@ -16,8 +15,6 @@ import {
 } from '../../../../../api/search/date/relative-date';
 import {isUndefined} from '@coveo/bueno';
 
-dayjs.extend(customParseFormat);
```
btaillon, line 19: Already extended by `date-format.ts`.

olamothe:
> I personally prefer not to call dayjs directly in the first place

Why is that ? 🤔

jpmarceau (approved):
👍

btaillon:
> > I personally prefer not to call dayjs directly in the first place
> 
> Why is that ? 🤔

I have a small preference for not referencing dependencies from too many different places, I think it makes it a bit easier in the event that we have to replace it one day or if we need to wrap it (like in this case).

packages/headless/src/api/search/date/date-format.test.ts
```diff
@@ -18,8 +18,8 @@ describe('#isSearchApiDate', () => {
 describe('#formatDateForSearchApi', () => {
   it('returns the correct format', () => {
     const date = 818035920000;
-    expect(formatDateForSearchApi(dayjs(date))).toBe(
-      dayjs(date).format('YYYY/MM/DD@HH:mm:ss')
+    expect(formatDateForSearchApi(parseDate(date))).toBe(
+      parseDate(date).format('YYYY/MM/DD@HH:mm:ss')
```
olamothe, line 22: There should be the inverse test: What happens when you pass in a date format that is not the API format.

olamothe (approved):

PR #2370, opened 2022-08-24
docs(atomic): fixed folded result list atomic-react sample

btaillon:
https://coveord.atlassian.net/browse/KIT-1973

Unfortunately, that fix makes it almost identical to the "custom children" page, since this is the only way to display folded children right now.

### DIFF
--- a/packages/samples/atomic-react/src/pages/FoldedResultListPage.tsx
+++ b/packages/samples/atomic-react/src/pages/FoldedResultListPage.tsx
@@ -22,8 +22,6 @@ import {
   FoldedResult,
   AtomicResultSectionChildren,
   AtomicFoldedResultList,
-  AtomicResultChildren,
-  AtomicResultChildrenTemplate,
 } from '@coveo/atomic-react';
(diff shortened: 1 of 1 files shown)


packages/samples/atomic-react/src/pages/FoldedResultListPage.tsx
```diff
@@ -63,13 +63,11 @@ function MyTemplate(result: FoldedResult) {
...
       </AtomicResultSectionExcerpt>
       <AtomicResultSectionChildren>
-        <AtomicResultChildren>
-          <AtomicResultChildrenTemplate>
-            <template>
-              <AtomicResultLink />
-            </template>
-          </AtomicResultChildrenTemplate>
-        </AtomicResultChildren>
+        {result.children.map((child) => (
```
olamothe, line 66: Let's remove the un-used imports 😉

olamothe (approved):

packages/samples/atomic-react/src/pages/FoldedResultListPage.tsx
btaillon: woops!

PR #2371, opened 2022-08-24
feat(atomic): add insight panel query summary component

olamothe:
Same old pattern: extract `QuerySummaryCommon` functional component, and re-use for both use cases.

https://coveord.atlassian.net/browse/KIT-1991

### DIFF
--- a/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
+++ b/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
@@ -1,0 +1,124 @@
+import {FunctionalComponent, h} from '@stencil/core';
+import {AnyBindings} from '../interface/bindings';
+import escape from 'escape-html';
+
+interface QuerySummaryCommonProps {
+  bindings: AnyBindings;
+  enableDuration: boolean;
+  querySummaryState: {
+    hasDuration: boolean;
+    durationInSeconds: number;
+    lastResult: number;
+    firstResult: number;
+    total: number;
+    query: string;
+    isLoading: boolean;
+    hasQuery: boolean;
+    hasError: boolean;
+    firstSearchExecuted: boolean;
+    hasResults: boolean;
+  };
+  setAriaLive: (message: string) => void;
+}
+
+export const QuerySummaryCommon: FunctionalComponent<
+  QuerySummaryCommonProps
+> = (props) => {
+  const getDurationInSeconds = (count: number) =>
+    props.bindings.i18n.t('in-seconds', {count});
+
+  const renderDuration = () => {
+    const shouldShow =
+      props.enableDuration && props.querySummaryState.hasDuration;
+    return `<span class="${
+      shouldShow ? '' : 'hidden'
+    }" part="duration"> ${getDurationInSeconds(
+      props.querySummaryState.durationInSeconds
+    )}</span>`;
+  };
+
+  const wrapHighlight = (content: string) => {
+    return `<span class="font-bold" part="highlight">${content}</span>`;
+  };
+
+  const wrapQuery = (content: string) => {
+    return `<span class="font-bold" part="highlight query">${content}</span>`;
+  };
+
+  const getResultOfProps = () => {
+    const locales = props.bindings.i18n.languages as string[];
+    return {
+      count: props.querySummaryState.lastResult,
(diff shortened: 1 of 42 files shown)


packages/atomic/src/components/common/query-summary/query-summary-common.tsx
```diff
@@ -0,0 +1,124 @@
...
+        : 'showing-results-of',
+      getHighlightedResultOfProps()
+    );
+
+    return (
+      // deepcode ignore ReactSetInnerHtml: This is not React code
+      <div
+        part="results"
+        class="overflow-hidden overflow-ellipsis"
+        innerHTML={content + renderDuration()}
```
btaillon, line 96: most of the complexity with this component seems to be related to highlighting several parts of the query summary.

Perhaps we could leverage the new `LocalizedString` JSX util instead to make it much easier to read.

Additionally, there's a lot of code in this file that is unreachable by the insight query component, perhaps we wouldn't even need the `query-summary-common` component anymore if we use the util, since it may become considerably simpler.

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
```diff
@@ -0,0 +1,124 @@
...
+    )}</span>`;
+  };
+
+  const wrapHighlight = (content: string) => {
+    return `<span class="font-bold" part="highlight">${content}</span>`;
+  };
+
+  const wrapQuery = (content: string) => {
+    return `<span class="font-bold" part="highlight query">${content}</span>`;
+  };
```
btaillon, line 46: If you use the LocalizedString util, these could become scoped functional components. E.g.:
```tsx
const Highlighted = ({isQuery, children}: PropsWithChildren<{ isQuery: boolean }>) =>
  <span class="font-bold" part={isQuery ? 'highlight query' : 'highlight'}>{children}</span>

...

return <LocalizedString
  bindings={props.bindings}
  key="..."
  params={{
    someValue: <Highlighted isQuery={false}>{props.querySummaryState.something}</Highlighted>
  }}
/>
```

jpmarceau (approved):
This reminds me, I'd be curious to know if we plan on making the Atomic insight components public at some point? Just so I can plan a bit for when we'll need to adjust our docs generation scripts.

olamothe:
> This reminds me, I'd be curious to know if we plan on making the Atomic insight components public at some point? Just so I can plan a bit for when we'll need to adjust our docs generation scripts.

Not in the short or mid term. Only internal for the foreseeable future.

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
olamothe: That's a good idea, it does simplify the code quite a bit.

As for dropping the QuerySummaryCommon entirely, I think it's still good to keep it

Still useful for common code for params extraction, dealing with plurals, aria-live.

packages/atomic/src/components/insight/atomic-insight-query-summary/atomic-insight-query-summary.tsx
```diff
@@ -0,0 +1,66 @@
...
+        </div>
+      );
+    }
+
+    if (this.querySummaryState.hasError) {
+      return;
+    }
+
+    return (
+      <div class="bg-[#F1F2FF] text-[#54698D] px-6 py-4 italic">
```
btaillon, line 61: Did you mean to use hardcoded colors here?

packages/atomic/src/utils/jsx-utils.tsx
```diff
@@ -5,12 +5,14 @@ export interface LocalizedStringProps {
   bindings: AnyBindings;
   key: string;
   params: Record<string, VNode | string>;
+  i18nCount?: number;
```
btaillon, line 8: IMO every parameter is already related to i18n, so I would replace this with `count`.

btaillon (approved):

packages/atomic/src/components/insight/atomic-insight-query-summary/atomic-insight-query-summary.tsx
olamothe: Yes it's as designed. Not worth it to create variable just for this value in insight panel.

PR #2373, opened 2022-08-24
fix(atomic): fix insight-refine-toggle title and tooltip 

olamothe:
The original story was a relatively small fix: add a tooltip to insight-refine-toggle + change the title of the modal from `Sort and filter` to only `Sort`.

I however took the opportunity to also refactor the code for `RefineModalCommon`. It was kind of a weird class with a bunch of stateful option.

I've changed it to a (hopefully) simpler standard functional component.

https://coveord.atlassian.net/browse/KIT-1975

### DIFF
--- a/packages/atomic/cypress/integration/facets/facet/facet.cypress.ts
+++ b/packages/atomic/cypress/integration/facets/facet/facet.cypress.ts
@@ -275,7 +275,7 @@ describe('Facet v1 Test Suites', () => {
       });
 
       describe('when searching for a value that returns a single result', () => {
-        const query = 'Gregory';
+        const query = 'BillV';
         function setupSearchForSingleValue() {
           setupSelectCheckboxValue();
           typeFacetSearchQuery(FacetSelectors, query, true);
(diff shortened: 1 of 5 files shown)


packages/atomic/cypress/integration/facets/facet/facet.cypress.ts
```diff
@@ -275,7 +275,7 @@ describe('Facet v1 Test Suites', () => {
       });
 
       describe('when searching for a value that returns a single result', () => {
-        const query = 'Gregory';
+        const query = 'BillV';
```
btaillon, line 278: I would mock the response to remove all results besides the first one.

btaillon (approved):

packages/atomic/cypress/integration/facets/facet/facet.cypress.ts
olamothe: I prefer to keep using the actual API (not a mock) unless absolutely necessary.

It does makes our tests a bit more finicky, but we've also caught quite a bit of actual API production issue by not mocking stuff.

PR #2375, opened 2022-08-25
fix(quantic): logic added to disable refine toggle when there is nothing to show in the filters tab

mmitiche:
### The issue:
The refine toggle button was always enabled giving the user the ability to pen the filters modal even if that modal doesn’t contain anything.

### The solution:
The refine toggle button is now disabled in the following cases:
- when there is no results.
- when the sort is hidden and there is no facet rendered.

https://user-images.githubusercontent.com/86681870/186729222-57de1805-372a-45cf-8392-00d9bc2789f3.mov

To achieve I added the following logic:
- To  know if there is results or not, the Quantic Refine Toggle component now subscribes to the search status headless controller.
- To know of the sort is hidden or not, the Quantic Refine Toggle component relies on the property  hideSort passed it.
- To know if no facet is rendered , I implanted a logic that will make all the facets communicate with the Quantic Refine Toggle component through events, in these events each facets tells the the Quantic Refine Toggle component if fits rendered or not.


mmitiche:
I was also thinking if we should completely hide the button instead of disabling it when it's not needed, what do you think from a UX point of view?

packages/quantic/cypress/integration/refine-toggle/refine-toggle.cypress.ts
```diff
@@ -67,6 +89,40 @@ describe('quantic-refine-toggle', () => {
         Expect.displayModal(false);
       });
     });
+
+    describe('when no facet is registered', () => {
```
lbergeron, line 93: Given ln195 defines a `describe` block with the same title, I assume this one should be:

```suggestion
    describe('when facets are registered', () => {
```

packages/quantic/cypress/integration/refine-toggle/refine-toggle.cypress.ts
```diff
@@ -133,6 +191,58 @@ describe('quantic-refine-toggle', () => {
         });
       });
     });
+
+    describe('when no facet is registered', () => {
```
lbergeron, line 195:
```suggestion
    describe('when no facets are registered', () => {
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineToggle/exampleQuanticRefineToggle.js
```diff
@@ -22,6 +22,20 @@ export default class ExampleQuanticRefineToggle extends LightningElement {
         'Indicates whether the refine modal should be opened in full screen.',
       defaultValue: false,
     },
+    {
+      attribute: 'withoutFacets',
```
lbergeron, line 26: :thinking: The options declared here were meant to match the component under test options, not to be used as testing utilities.

I think it would be cleaner to use an action button for the "without facets" test. I could be a "Perform search with no facets" button. This way, it would be clear to a user that the "without facets" is the result of the performed query, and not an option you can set on the component to change its behavior.

The "facets without inputs" case is a little more tricky to handle because the option is meant to configure the example component, not the component under test. I guess you can keep this one like this for now, but we should improve the example component framework to support this. We'd need the same feature to easily switch the use case under test (i.e., search, insight, case assist, etc)

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -225,6 +225,13 @@ export default class QuanticCategoryFacet extends LightningElement {
       this.searchStatus?.state?.isLoading &&
       !this.searchStatus?.state?.hasError &&
       !this.searchStatus?.state?.firstSearchExecuted;
+    
+    const facetRenderingEvent = new CustomEvent('facetRendering', {
```
lbergeron, line 229: If possible, I'd name the event `renderFacet`. In my opinion, it reads better.

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -87,18 +100,37 @@ export default class QuanticRefineToggle extends LightningElement {
...
+
+    const registredFacets = getAllFacetsFromStore(this.engineId);
+    Object.keys(registredFacets).forEach((facetId) => {
+      this.renderedFacets[facetId] = true;
+    });
   };
 
+  get refineButtonDisabled() {
+    const areFacetsRendered = Object.values(this.renderedFacets).reduce(
+      (result, facetRendred) => result || facetRendred,
```
lbergeron, line 123:
```suggestion
      (result, facetRendered) => result || facetRendered,
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineToggle/exampleQuanticRefineToggle.js
mmitiche: `WithoutFacets` and `FacetsWithoutInputs` removed from the options and are implemented as action buttons:
https://user-images.githubusercontent.com/86681870/187200252-bf114320-ad48-4749-a1e5-16bcf43bb747.mov

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineToggle/templateWithFacetsWithoutInputs.html
```diff
@@ -0,0 +1,108 @@
...
+        label="Add Facets Without Inputs"
+        data-id="add-facets-without-inputs"
+        slot="actions"
+        disabled={notConfigured}
+        engine-id={engineId}
+        without-inputs
+      >
+      </c-action-add-facets>
+    </c-configurator>
+    WITHOUT inputs
```
lbergeron, line 32: It this text still useful?

PR #2377, opened 2022-08-25
fix(atomic): fix insight panel widget mode + refactor pager

olamothe:
Original bug was a relatively small fix to the local "widget" mode, which forced the width and height of the panel to look like a small window (in `insights.html`).

While working on that I also found some rendering problems with the pagination section when there are no results or there is a query error.

I moved styling from the layouts to the atomic-insight-pager component itself, so we can better control when it is hidden or shown.

Also refactored PagerCommon from a class to a function component.

https://coveord.atlassian.net/browse/KIT-1957

### DIFF
--- a/packages/atomic/src/components/common/pager/pager-common.tsx
+++ b/packages/atomic/src/components/common/pager/pager-common.tsx
@@ -1,58 +1,56 @@
-import {EventEmitter, h} from '@stencil/core';
+import {EventEmitter, FunctionalComponent, h} from '@stencil/core';
 import ArrowRight from '../../../images/arrow-right.svg';
 import {FocusTargetController} from '../../../utils/accessibility-utils';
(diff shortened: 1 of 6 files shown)


packages/atomic/src/components/common/pager/pager-common.tsx
```diff
@@ -1,58 +1,56 @@
...
-  getActivePage(): FocusTargetController | undefined;
+  searchStatusState: {
+    hasResults: boolean;
+    hasError: boolean;
+  };
+  pagerState: {
+    hasPreviousPage: boolean;
+    currentPages: number[];
+    hasNextPage: boolean;
+  };
```
btaillon, line 19: if only functional components could have a state :sweat_smile:

btaillon (approved):

PR #2380, opened 2022-08-26
test(atomic): add end to end tests for insights panel

olamothe:
Add tests that verifies the basic functionality of the insight panel. It obviously does not go into as detailed as core atomic components.

The reason is that there is a lot of overlap between all the components, so it would be repetitive.
Accessibility for the insight panel builder is also not a requirement.

This * should not * be merged as is: this requires some changes in the insight panel API to support API keys. Right now, it relies on platform token( which expires after 24h). This means this PR will fail builds if merged right away.

https://coveord.atlassian.net/browse/KIT-1952


.github/workflows/prbot.yml
```diff
@@ -47,7 +47,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        spec: ["cypress/integration/*", "cypress/integration/result-list/**/*", "cypress/integration/facets/**/*", "cypress/integration/search-box/**/*"]
+        spec:
+          [
+            'cypress/integration/*',
+            'cypress/integration/result-list/**/*',
+            'cypress/integration/facets/**/*',
+            'cypress/integration/search-box/**/*',
```
btaillon, line 55: I don't think you need brackets here since this is YAML

```suggestion
          - 'cypress/integration/*'
          - 'cypress/integration/result-list/**/*'
          - 'cypress/integration/facets/**/*'
          - 'cypress/integration/search-box/**/*'
```

btaillon (approved):

lvu285 (approved):

nathanlb:
The insight query API now accepts API keys! 🎉

PR #2383, opened 2022-08-30
feat(quantic): Aria-live regions announcements for quantic components

erocheleau:
[SVCC-2310](https://coveord.atlassian.net/browse/SVCC-2310)

First version of an aria-live manager to announce changes in the UI for accessibility purposes.

For this version, I added:
- Announcing the no results from quantic-no-results
- Announcing the query errors from quantic-query-error
- Announcing when a new query is executed through quantic-summary

Each component can import AriaLiveRegion from quanticUtils and declare a new region. 
And then you can use `dispatchMessage` to send an event to update an aria region.
The QuanticSearchInterface if the QuanticAriaLive is present in its content, will listen for these event and trigger the updated message.

The message is invisible, but can be heard with a screen reader.


packages/quantic/force-app/main/default/lwc/quanticAriaLive/__tests__/quanticAriaLiveManager.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+describe('c-quantic-aria-live-manager', () => {
+    afterEach(() => {
+        // The jsdom instance is shared across test cases in a single file so reset the DOM
+        while (document.body.firstChild) {
+            document.body.removeChild(document.body.firstChild);
+        }
+    });
+
+    it('TODO: test case generated by CLI command, please fill in test logic', () => {
```
nathanlb, line 12: test could be improved... or removed

packages/quantic/force-app/main/default/lwc/quanticAriaLive/quanticAriaLive.js
```diff
@@ -0,0 +1,32 @@
+import { LightningElement, track, api } from 'lwc';
+
+export default class QuanticAriaLive extends LightningElement {
```
nathanlb, line 3: If we want this component to be visible in the reference docs its class needs proper comments.

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -168,6 +178,27 @@ export default class QuanticSearchInterface extends LightningElement {
     );
   }
 
+  bindAriaLiveEvents() {
```
nathanlb, line 181: It seems to me like we'll want this capability on other interface types such as Insight and Case Assist and these handlers seem generic enough so perhaps they could be in some sort of Util to be easily reused?

nathanlb (commented):
This looks great!

In the cypress stuff there's [this](https://github.com/coveo/ui-kit/tree/master/packages/quantic/force-app/examples/main/lwc/eventListener) utility wrapper component for testing caught exceptions.
Could be cool to add some assertions for the new sent events.

packages/quantic/force-app/main/default/lwc/quanticSummary/quanticSummary.js
```diff
@@ -76,6 +79,16 @@ export default class QuanticSummary extends LightningElement {
 
   updateState() {
     this.state = this.querySummary.state;
+    if(this.state?.hasResults) {
+      this.updateAriaMessage();
+    }
+  }
+
+  updateAriaMessage() {
+    const docElement = document.createElement('div');
```
mmitiche, line 88: why do we need this manipulation? is it because` this.summaryLabel` contains HTML tags?

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -467,3 +467,38 @@ export class Store {
...
+      }
+    });
+    elem.dispatchEvent(registerRegionEvent);
+  }
+
+  registerRegion();
+
+  return {
+    dispatchMessage,
+    registerRegion,
```
mmitiche, line 502: What's the point of returning `registerRegion` here? since `registerRegion` is already called when executing the `AriaLiveRegion` function and it will not need to be called again from the components.

packages/quantic/force-app/main/default/lwc/quanticAriaLive/quanticAriaLive.js
```diff
@@ -0,0 +1,32 @@
...
+  /** @type {Object.<string, Object>} */
```
btaillon, line 4: You could define the interface above and use it here instead of Object.

Example:
```ts
/**
 * @typedef AriaLiveRegion
 * @property {string} message
 * @property {boolean} assertive
 */
```
```suggestion
  /** @type {Record<string, AriaLiveRegion>} */
```

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -467,3 +467,38 @@ export class Store {
     return store.state[facetType];
   }
 }
+
+export function AriaLiveRegion(regionName, elem, assertive = false) {
+
+  function dispatchMessage(message) {
+    // Creates the event to dispatch an event to an aria-live region
+    const registerRegionEvent = new CustomEvent('arialivemessage', {
```
mmitiche, line 475: I think this event should be named `ariaLiveMessageEvent` instead.
```suggestion
    const ariaLiveMessageEvent = new CustomEvent('arialivemessage', {
```

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -467,3 +467,38 @@ export class Store {
...
+      bubbles: true,
+      detail: {
+        regionName,
+        assertive,
+        message
+      }
+    });
+    elem.dispatchEvent(registerRegionEvent);
```
mmitiche, line 483:
```suggestion
    elem.dispatchEvent(ariaLiveMessageEvent);
```

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -467,3 +467,38 @@ export class Store {
...
+  }
+
+  function registerRegion() {
+    // Creates the event to register an aria-live region
```
mmitiche, line 487: Maybe the comments should be right before the declaration of the function

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -168,6 +178,27 @@ export default class QuanticSearchInterface extends LightningElement {
...
+    this.template.addEventListener('arialivemessage', this.handleAriaLiveMessage.bind(this));
+    this.template.addEventListener('registerregion', this.handleRegisterAriaLiveRegion.bind(this));
+  }
+
+  handleAriaLiveMessage(event) {
+    /** @type {Object} */
```
btaillon, line 187: If possible, I would use the `QuanticAriaLive` type here.

btaillon:
I tested this on the following screen readers and it worked:
* JAWS
* NVDA
* iOS VoiceOver
* macOS VoiceOver (the search box itself is unusable though)
* TalkBack

btaillon (approved):

packages/quantic/force-app/main/default/lwc/quanticAriaLive/__tests__/quanticAriaLiveManager.test.js
erocheleau: Yea it's kind of a WIP 😂

packages/quantic/force-app/main/default/lwc/quanticAriaLive/quanticAriaLive.js
erocheleau: Interesting, do we want to?

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
erocheleau: You mean create like a generic utils method to handle these two events?
Something like this?

`Utils.dispatchMessageHandler(element, event);`

You need to have a reference to the current element since you need to be able to find the aria-live component inside your own DOM

packages/quantic/force-app/main/default/lwc/quanticSummary/quanticSummary.js
erocheleau: Yea the function I use to get what is going to be displayed inside the quanticSummary inserts <b> tags around the query text and things like that, since I didn't want to re-code another function to get the same results, I decided to just sanitize the html into text instead.

And yes the screen reader was actually reading the 🔊`smaller than b greater than query text smaller than forward slash b greater than`

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
erocheleau: You might want to call it to register a new region I guess? Maybe we don't need to.

packages/quantic/force-app/main/default/lwc/quanticNoResults/quanticNoResults.js
```diff
@@ -61,6 +61,8 @@ export default class QuanticNoResults extends LightningElement {
   unsubscribeQuerySummary;
   /** @type {Function} */
   unsubscribeBreadcrumbsManager;
+  /** @type {Object} */
```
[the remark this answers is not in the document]
erocheleau: Yea makes sense

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
erocheleau: The conventions wants event names to be all lowercase. The reason seems to be because attribute names are case insensitive but even names are case sensitive so it can create some conflicts.

But I could use kebab-case if you don't like the name?

erocheleau:
> I tested this on the following screen readers and it worked:
> 
> * JAWS
> * NVDA
> * iOS VoiceOver
> * macOS VoiceOver (the search box itself is unusable though)
> * TalkBack

Wow thanks! I only tested it on NVDA and Windows narrator :P

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
erocheleau: The problem is that selecting that Element with `querySelector` doesn't give me an "instance" of QuanticAriaLive as far as the types are concerned. I've looked for a way to fix this without success :(

I created an interface instead:

```
/**
* @typedef {Object} IQuanticAriaLive
* @method registerRegion
* @method updateMessage
*/
```

That was the only way I found to have it working

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
erocheleau: I misread your message, yes that was a copy-paste error :)

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
erocheleau: I don't know why I even have comments 😂

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
erocheleau: Since I need the element itself I think I would rather keep it directly here? And add new ones for Case Assist and Insight? what do you think?

packages/quantic/force-app/main/default/lwc/quanticAriaLive/quanticAriaLive.js
erocheleau: Anyway I created one

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
nathanlb: When we would add new ones it would be a straight up copy paste though no? 🐟

nathanlb: Not the end of the world I guess just think it would make future development less error-prone.

erocheleau:
Added the tests as promised: 

image

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
erocheleau: Not sure I follow can you flesh out the suggestion?

nathanlb (approved):
Very cool solution!
Given this relies on bubbling events I guess its usage is not as straight forward for interfaces with sibling components that aren't necessarily children of the quantic-search-interface. Correct?
Also do we plan on adding it to the QuanticInsightInterface soon? There shouldn't be anything blocking for that right?

MSLI-Coveo (approved):

erocheleau:
> Very cool solution! Given this relies on bubbling events I guess its usage is not as straight forward for interfaces with sibling components that aren't necessarily children of the quantic-search-interface. Correct? Also do we plan on adding it to the QuanticInsightInterface soon? There shouldn't be anything blocking for that right?

Yea definitely nothing blocking to do that.
Do we HAVE elements that are siblings outside of the quantic-search-interface?

PR #2388, opened 2022-09-01
feat(quantic): change sortCriteria for facetValue

MSLI-Coveo:
[SVCC-2293](https://coveord.atlassian.net/browse/SVCC-2293)
change sortCriteria to "score" for facetValue in order to keep the same order of facetValues after "Show More" or "Show Less" is clicked.

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
@@ -76,7 +76,7 @@ export default class QuanticFacet extends LightningElement {
    *   - `occurrences`
    *   - `automatic`
    * @api
-   * @type  {'score' | 'alphanumeric' | 'occurrences' | 'automatic'}
+   * @type  {'score' | 'alphanumeric' | 'occurrences'}
    * @defaultValue `'automatic'`
    */
   @api sortCriteria = 'alphanumeric';
--- a/packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
@@ -79,7 +79,7 @@ export default class QuanticFacet extends LightningElement {
    * @type  {'score' | 'alphanumeric' | 'occurrences' | 'automatic'}
    * @defaultValue `'automatic'`
    */
-  @api sortCriteria = 'automatic';
+  @api sortCriteria = 'alphanumeric';
   /**
    * Whether this facet should not contain a search box.
    * @api
@@ -221,9 +221,12 @@ export default class QuanticFacet extends LightningElement {
       this.searchStatus?.state?.isLoading &&
       !this.searchStatus?.state?.hasError &&
       !this.searchStatus?.state?.firstSearchExecuted;
-    
+
     const renderFacetEvent = new CustomEvent('renderFacet', {
-      detail: {id: this.facetId ?? this.field, shouldRenderFacet: this.hasValues},
+      detail: {
+        id: this.facetId ?? this.field,
+        shouldRenderFacet: this.hasValues,
+      },
       bubbles: true,
       composed: true,
     });
--- a/packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
@@ -74,12 +74,11 @@ export default class QuanticFacet extends LightningElement {
    *   - `score`
    *   - `alphanumeric`
    *   - `occurrences`
-   *   - `automatic`
    * @api
    * @type  {'score' | 'alphanumeric' | 'occurrences'}
-   * @defaultValue `'alphanumeric'`
+   * @defaultValue `'score'`
    */
-  @api sortCriteria = 'alphanumeric';
+  @api sortCriteria = 'score';
   /**
    * Whether this facet should not contain a search box.
    * @api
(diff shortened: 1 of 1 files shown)


nathanlb:
The other possible values for sortCriteria should work as well though. Changing the default value doesn't feel like a real fix given then users that explicitly set it to something else would still have the same issue.

erocheleau:
> The other possible values for sortCriteria should work as well though. Changing the default value doesn't feel like a real fix given users that explicitly set it to something else would still have the same issue.

The problem is that we were sending: `automatic` as the sortCriteria which is not an actually real value for the sortCriterias.

And the search API has a little bit of a weird default value for facet sort: 

> By default: 
When isFieldExpanded is false in the facet request, and moreValuesAvailable is true in the corresponding facet response, use score
Otherwise, use alphanumeric.

So by sending a value that doesn't exist, effectively when pressing show more, it would actually change our sorting criteria...

nathanlb:
> > The other possible values for sortCriteria should work as well though. Changing the default value doesn't feel like a real fix given users that explicitly set it to something else would still have the same issue.
> 
> The problem is that we were sending: `automatic` as the sortCriteria which is not an actually real value for the sortCriterias.
> 
> And the search API has a little bit of a weird default value for facet sort:
> 
> > By default:
> > When isFieldExpanded is false in the facet request, and moreValuesAvailable is true in the corresponding facet response, use score
> > Otherwise, use alphanumeric.
> 
> So by sending a value that doesn't exist, effectively when pressing show more, it would actually change our sorting criteria from `score` to `alphanumeric`

Ahh je comprend. Yeah I recently realized that it wasn't a supported value and brought it up to the search ui team. They had no idea and had been using automatic for a long time. They decided to keep it as a way to explicitly say "not specified". This change makes sense then but we might as well remove any mention of `'automatic'` from the JSDoc then.

MSLI-Coveo:
> > > The other possible values for sortCriteria should work as well though. Changing the default value doesn't feel like a real fix given users that explicitly set it to something else would still have the same issue.
> > 
> > 
> > The problem is that we were sending: `automatic` as the sortCriteria which is not an actually real value for the sortCriterias.
> > And the search API has a little bit of a weird default value for facet sort:
> > > By default:
> > > When isFieldExpanded is false in the facet request, and moreValuesAvailable is true in the corresponding facet response, use score
> > > Otherwise, use alphanumeric.
> > 
> > 
> > So by sending a value that doesn't exist, effectively when pressing show more, it would actually change our sorting criteria from `score` to `alphanumeric`
> 
> Ahh je comprend. Yeah I recently realized that it wasn't a supported value and brought it up to the search ui team. They had no idea and had been using automatic for a long time. They decided to keep it as a way to explicitly say "not specified". This change makes sense then but we might as well remove any mention of `'automatic'` from the JSDoc while we're at it.

And setting `sortCriteria` to `alphanumeric` would also trigger untracked change of content in some scenarios. It seems we still to use aria-live utility to announce that facetValues are change. Like the example below:

chrome-capture-2022-8-1 (2)

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
```diff
@@ -79,7 +79,7 @@ export default class QuanticFacet extends LightningElement {
    * @type  {'score' | 'alphanumeric' | 'occurrences' | 'automatic'}
    * @defaultValue `'automatic'`
```
nathanlb, line 80:
```suggestion
   * @defaultValue `'alphanumeric'`
```

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
```diff
@@ -79,7 +79,7 @@ export default class QuanticFacet extends LightningElement {
    * @type  {'score' | 'alphanumeric' | 'occurrences' | 'automatic'}
```
nathanlb, line 79:
```suggestion
   * @type  {'score' | 'alphanumeric' | 'occurrences'}
```

MSLI-Coveo: 'automatic' removed from faceValue's sortCriteria type.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
MSLI-Coveo: change default value for sortCriteria to 'alphanumeric'

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
```diff
@@ -76,10 +76,10 @@ export default class QuanticFacet extends LightningElement {
    *   - `occurrences`
    *   - `automatic`
    * @api
-   * @type  {'score' | 'alphanumeric' | 'occurrences' | 'automatic'}
-   * @defaultValue `'automatic'`
+   * @type  {'score' | 'alphanumeric' | 'occurrences'}
+   * @defaultValue `'alphanumeric'`
```
erocheleau, line 80: We should use 'score' as a default instead of alphanumeric though.

erocheleau (changes requested):

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
MSLI-Coveo: Changed @api sortCriteria to score for FacetValue

erocheleau (approved):

PR #2389, opened 2022-09-01
fix(headless): refactor query trigger and search actions to better support undo feature

olamothe:
Refactor of Headless trigger to remove logic embedded in the controller and add it to the state (ignore/correction thingy).

Refactored the query trigger to be processed in a very similar way to standard `did you mean` feature, with a retry directly embedded in the different search actions.

To embed that retry logic, I also did a refactor of the search-actions file: There was quite a bit of repetition.

I created a `AsyncSearchThunkProcessor` class, with two public methods: `fetchFromAPI()`, and `process()`.,

https://coveord.atlassian.net/browse/KIT-1999


packages/headless/src/features/triggers/triggers-state.ts
```diff
@@ -48,6 +48,15 @@ export interface TriggerState {
...
    */
   notifications: string[];
+
+  /**
+   * The modification that should be applied as a result of a query trigger.
+   */
+  queryModification: {
+    originalQuery: string;
+    modification: string;
+    ignore: string;
```
btaillon, line 58: The name `ignore` makes me expect this to be a boolean, but it's a string. I think a name like `queryToIgnore` would be more clear.

Also, I think that `ignore` should be optional, since an empty string is also technically a query.

packages/headless/src/features/triggers/triggers-state.ts
```diff
@@ -48,6 +48,15 @@ export interface TriggerState {
    * The notifications to display to the user after receiving notification triggers.
    */
   notifications: string[];
+
+  /**
+   * The modification that should be applied as a result of a query trigger.
+   */
+  queryModification: {
```
btaillon, line 55: I think it would make more sense for `queryModification` to be `undefined` when no correction occured. Perhaps `ignore` could be extracted out of this object if that's the case.

packages/headless/src/features/triggers/triggers-state.ts
btaillon: I think each three properties here should get a jsdoc description.

packages/headless/src/features/triggers/triggers-state.ts
```diff
@@ -48,6 +48,15 @@ export interface TriggerState {
...
+    originalQuery: string;
+    modification: string;
```
btaillon, line 57: Instead of `modification`, I feel like `newQuery` would make more sense and would be more comparable to `originalQuery`.

btaillon (approved):
looks great!

packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -0,0 +1,346 @@
...
+
+  rejectWithValue: (err: SearchAPIErrorWithStatusCode) => unknown;
+  analyticsAction: SearchAction | null;
+  extra: ClientThunkExtraArguments<SearchAPIClient>;
+}
+
+export class AsyncSearchThunkProcessor<RejectionType> {
+  constructor(
+    private config: AsyncThunkConfig,
+    private onUpdateQueryForCorrection: (modification: string) => void = (
```
ThibodeauJF, line 113: We could define an interface for this param

ThibodeauJF (approved):

packages/headless/src/features/triggers/triggers-state.ts
olamothe: Whenever possible I think it's better to have empty object instead of undefined object to represent an empty state.

Having `undefined` in some places in the state has a much bigger impact on downstream code that needs to check for null ref everywhere, vs having an empty object keeps the complexity local to the state + slice code.

PR #2391, opened 2022-09-03
feat(insight): added field defined condition props to insight result template

nathanlb:
Currently the atomic-insight-result-template only supports matching conditions via markup.
To support the full range of conditions available in the builder via attribute without injecting any javascript the component should have two new props:
- isDefined
- isNotDefined

which each accept a string of values separated by commas.


lbergeron (approved):

packages/atomic/src/components/insight/atomic-insight-result-template/atomic-insight-result-template.tsx
```diff
@@ -24,6 +25,20 @@ export class AtomicInsightResultTemplate {
...
+   * For example, a template with the following attribute only applies to result items whose `filetype` and `sourcetype` fields are defined: `is-defined="filetype,sourcetype"`
+   */
+  @Prop({reflect: true}) isDefined?: string;
+
+  /**
+   * The field that, when defined on a result item, would prevent the template from being applied.
+   *
+   * For example, a template with the following attribute only applies to result items whose `filetype` and `sourcetype` fields are NOT defined: `is-not-defined="filetype,sourcetype"`
+   */
+  @Prop({reflect: true}) isNotDefined?: string;
```
btaillon, line 40: Same as above

```suggestion
  @Prop({reflect: true}) ifNotDefined?: string;
```

btaillon (approved):

ThibodeauJF (approved):

PR #2278, opened 2022-08-02
feat(atomic): improve positioning of modal and wait for app to load before showing it

liviarett:
This solves the issues with loading and scrolling seen in the video 

https://user-images.githubusercontent.com/30511433/182371026-0b03c559-b620-42c7-86ce-897cacf9e0d6.mov

It also works if your interface was to be positioned absolutely

PR #2392, opened 2022-09-06
fix(atomic): fixed atomic-insight-refine-modal open animation

btaillon:
https://coveord.atlassian.net/browse/KIT-2009

Prior to this PR, the animation was fullscreen when the modal was opened for the first time. This was because the modal was initially rendered without knowing the dimensions of the insight interface.

Another change I made in this PR was to remove a redundant event that was used to find out the size of the insight interface. It's not necessary since we already have the bindings.

### DIFF
--- a/packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
+++ b/packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
@@ -1,15 +1,10 @@
 import {Component, h, State, Prop, Element, Watch, Host} from '@stencil/core';
-import {debounce} from 'ts-debounce';
-import {buildCustomEvent} from '../../../utils/event-utils';
 import {
   BindStateToController,
   InitializableComponent,
   InitializeBindings,
 } from '../../../utils/initialization-utils';
-import {
-  InsightBindings,
-  InsightInterfaceDimensions,
-} from '../atomic-insight-interface/atomic-insight-interface';
+import {InsightBindings} from '../atomic-insight-interface/atomic-insight-interface';
 import {
   buildInsightFacetManager,
   InsightFacetManager,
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
```diff
@@ -45,17 +40,14 @@ export class AtomicInsightRefineModal
...
   @Prop({mutable: true}) openButton?: HTMLElement;
 
   @Prop({reflect: true, mutable: true}) isOpen = false;
-
-  private interfaceDimensions?: InsightInterfaceDimensions;
   private facetManager!: InsightFacetManager;
   private resizeObserver?: ResizeObserver;
-  private debouncedUpdateDimensions = debounce(this.updateDimensions, 500);
-  private scrollCallback = () => this.debouncedUpdateDimensions();
+  private scrollCallback = () => this.updateDimensions();
```
ThibodeauJF, line 50: Was the debounce here because it was laggy? How does it look now?

ThibodeauJF (approved):

packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
btaillon: I assumed it was due to waiting for the event handler to be present, but actually it seems to be to prevent refreshing too often. Will fix.

https://github.com/coveo/ui-kit/pull/2278#discussion_r936745666

ThibodeauJF:
@btaillon I *think* this also fixes https://coveord.atlassian.net/browse/KIT-2004

packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
ThibodeauJF: The debounce makes the user experience pretty horrible in full screen mode when scrolling

btaillon: I'll try switching to [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) instead of an event, and I'll remove the debounce.

PR #2393, opened 2022-09-06
feat(atomic): added "filters" header with clear button to insight refine modal

btaillon:
https://coveord.atlassian.net/browse/KIT-2008

Copied from `atomic-refine-modal`'s:
image

### DIFF
--- a/packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
+++ b/packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
@@ -1,14 +1,4 @@
 import {Component, h, State, Prop, Element, Watch, Host} from '@stencil/core';
-import {
-  buildFacetManager,
-  FacetManager,
-  QuerySummary,
-  QuerySummaryState,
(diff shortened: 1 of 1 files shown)


ThibodeauJF (approved):

packages/atomic/src/components/insight/atomic-insight-refine-modal/atomic-insight-refine-modal.tsx
```diff
@@ -1,4 +1,14 @@
 import {Component, h, State, Prop, Element, Watch, Host} from '@stencil/core';
+import {
```
olamothe, line 2: I think it's better to keep the import renamed from the `index.ts` file ( `import {buildInsightFacetManager} from '..'` ) instead of directly from `coveo/headless/insight`.

Couple weeks ago I was going around and doing the modification like you did here, to use the imports directly from headless/insight, but remembered why it causes problems: It will create weirdly named types in `components.d.ts` due to name collision with the `headless` sub package.

So even if it does not change much for this file, I think it's better to keep it uniform for all insight panel components.

olamothe (approved):

PR #2395, opened 2022-09-06
chore(atomic): add comment in facets about isCollapsed & the facet manager

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1958

I'll add the comment to all facets

### DIFF
--- a/packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
+++ b/packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
@@ -120,7 +120,7 @@ export class AtomicFacet implements InitializableComponent, BaseFacet<Facet> {
   @Prop({reflect: true}) public displayValuesAs: 'checkbox' | 'link' | 'box' =
     'checkbox';
   /**
-   * Specifies if the facet is collapsed.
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components.d.ts
```diff
@@ -186,7 +186,7 @@ export namespace Components {
          */
         "injectionDepth": number;
         /**
-          * Specifies if the facet is collapsed.
+          * Specifies if the facet is collapsed. When the facet is the child of an `atomic-facet-manager` component, it will control this property.
```
jpmarceau, line 189: Here, by `it`, do you mean the `atomic-facet-manager`? If so, I suggest:
```suggestion
          * Specifies whether the facet is collapsed. When the facet is the child of an `atomic-facet-manager` component, the facet manager controls this property.
```

spiracha-coveo: This sounds better to me too :P

ThibodeauJF: Thanks!

jpmarceau (approved):
👍

btaillon (approved):

PR #2397, opened 2022-09-07
fix(quantic): issue fixed when QuanticTabBar doesn't contain any tab

mmitiche:
[SVCC-2436](https://coveord.atlassian.net/browse/SVCC-2436)

### The issue:
An error is thrown when trying to call the `getAbsoluteWidth` method with null as the parameter,  this situation happens when there is no Tab passed to the QuanticTabBar: 
 
image-20220907-150608

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticTabBar/quanticTabBar.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticTabBar/quanticTabBar.js
@@ -364,6 +364,9 @@ export default class QuanticTabBar extends LightningElement {
    * @returns {number}
    */
   getAbsoluteWidth(element) {
+    if (!element) {
+      return 0;
+    }
     const paddings = this.getElementPadding(element);
     const padding = paddings.left + paddings.right;
 


aelfaleh (approved):

nathanlb:
If there are no tabs inside of a tab bar we should probably still log an error though. Wouldn't that be a misuse of the component?
IMO In the packaged SFINT HIP component the conditional rendering should cover the entire TabBar based on the presence of tabs in the config. Is that the plan?

mmitiche:
> If there are no tabs inside of a tab bar we should probably still log something though. Wouldn't that be a misuse of the component? IMO In the packaged SFINT HIP component the conditional rendering should cover the entire TabBar based on the presence of tabs in the config. Is that the plan?

Good point👌 I'll also add the logic in the SFINT HIP component to only display the QuanticTabBar only when there is more then one tab.
But this doesn't prevent the fix I introduced here imo. 
I can add  a warning message in the QuanticTabBar component to notify the user that he hasn’t introduced  any tab, that's indeed a misuse of the component.

PR #2400, opened 2022-09-08
fix(quantic): new title for the refine toggle when it's disabled

mmitiche:
[SVCC-2437](https://coveord.atlassian.net/browse/SVCC-2437)
### Issue
Following a discussion with the UX team, we decided to update the title of the refine toggle button when this button is disabled.

### behaviour  we want for the refine toggle button:

When the facets are ON but the query returns no results:
Always keep filter button, we don’t want the layout to change all the time. Two situations here:

- The query is empty but there are filters selected: keep the filter panel accessible. So users can modify the filters as they might be the reason the query is not returning results.

- The query is empty but there are NO filters selected. Disable the button, it doesn’t help much to see an empty filter panel. Add a tooltip to explain the behaviour. example: No filters available for this query

### Demo
https://user-images.githubusercontent.com/86681870/189116268-9b04618b-b757-4e1f-8d83-a6620f79098c.mov


lbergeron (approved):
Looks good!

PR #2402, opened 2022-09-08
fix(quantic): removed unnecessary margins from result template slots

mmitiche:
### The issue
The results templates slots were taking some space even when the slots are completely empty.

### Before:
NoBAdgeBEfore

### After:
NoBAdgesAfter

Note: Changes in this PR do not have any impact when the slots are provided.
Quantic standard result templates still look absolutely  the same:
NoBadgesAfter2Standard


nathanlb (approved):
👍

aelfaleh (approved):
LGTM

PR #2408, opened 2022-09-09
feat(quantic): accessibility of the Quantic facets improved

MSLI-Coveo:
[SVCC-2280](https://coveord.atlassian.net/browse/SVCC-2280)
Accessibility of the Quantic facets improved:

- When selecting a facet value, the focus now stays on that facet value instead of being reset to the top of the page.
- When clearing the filters of a facet. the focus now is set on the header of the facet instead of being reset to the top of the page.
- When clicking `show more`/`show less`, the focus is automatically set on the first facet value of the facet.
- When applying a filter in the the Numeric Facet or the Timeframe facet, the focus stays on the `apply` button instead of being reset to the top of the page.

#### Behaviour of the QuanticFacet:
https://user-images.githubusercontent.com/86681870/191841102-8d2d8240-eaeb-436e-a1fb-1eebc15d0d13.mov

#### Behaviour of the QuanticNumericFacet:
https://user-images.githubusercontent.com/86681870/191841983-699d8ea3-c62c-4de7-b4eb-487c2933b80f.mov

#### Behaviour of the QuanticCategoryFacet:
https://user-images.githubusercontent.com/86681870/191842067-e505a4d7-92c7-41e9-866e-260b7a0235c7.mov

#### Behaviour of the QuanticTimeframeFacet:
https://user-images.githubusercontent.com/86681870/191842098-0d9632c9-167b-4f24-8778-a44776d327e7.mov

#### Behaviour of the QuanticDateFacet:

https://user-images.githubusercontent.com/86681870/191842634-3e310381-9380-4381-b215-67571e4d2a90.mov


mmitiche (changes requested):
Nice solution👌
Just few changes to discard and few questions to answer!
Also I was wondering. what will the screenreader read when the focus is on the header of the `QuanticCardContainer`?

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
```diff
@@ -179,9 +179,18 @@ export default class QuanticFacet extends LightningElement {
...
+      this.nextFocusedIndex =
+        typeof this.nextFocusedIndex === 'number'
+          ? this.nextFocusedIndex
+          : this.values
+              .map((item) => item.value)
+              .indexOf(this.nextFocusedIndex);
+
+      if (this.nextFocusedIndex >= 0) {
+        this.setFocusToFacetValue(this.nextFocusedIndex);
+      }
```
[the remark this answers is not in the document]
MSLI-Coveo: Yeah, this is still 'wip' since I was not sure it would be the implementation that makes most sense.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -226,6 +226,8 @@ export default class QuanticCategoryFacet extends LightningElement {
       facetId: this.facet.state.facetId,
       element: this.template.host,
     });
+
+    this.toggleVisibility = this.toggleFacetVisibility.bind(this);
```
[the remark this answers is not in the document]
MSLI-Coveo: Yes, I was using the function `toggleFacetVisibility` not `toggleFacetVisibility`. The names are quit confusing. I tried to put `this.toggleFacetVisibility = this.toggleFacetVisibility.bind(this)`, but it would break the page and I am not sure why.

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
```diff
@@ -7,10 +7,29 @@ import { LightningElement, api } from 'lwc';
  * <c-quantic-card-container title="Card Example"></c-quantic-card-container>
```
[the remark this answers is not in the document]
MSLI-Coveo: On that part, I am not sure how should we describe this new property to customers. it works with our facet components, but what functions should users put if they only want to use the component alone.

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
```diff
@@ -7,10 +7,29 @@ import { LightningElement, api } from 'lwc';
...
-}
\ No newline at end of file
+  /**
+   * The title label to display in the card header.
+   * @api
+   * @type {string}
+   */
+  @api title;
+  /**
+   * The title label to display in the card header.
```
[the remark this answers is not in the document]
MSLI-Coveo: On that part, I am not sure how should we describe this new property to customers. it works with our facet components, but what functions should users put if they only want to use the component alone.

packages/quantic/force-app/main/default/lwc/quanticNumericFacet/quanticNumericFacet.html
```diff
@@ -4,15 +4,15 @@
...
+      <c-quantic-card-container title={label} toggle-visibility={toggleVisibility}>
         <lightning-button-icon
           class={actionButtonCssClasses}
           slot="actions"
-          onclick={toggleFacetVisibility}
           onmousedown={preventDefault}
           icon-name={actionButtonIcon}
           alternative-text={actionButtonLabel}
           variant="bare"
+          tabindex="-1"
```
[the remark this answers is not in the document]
MSLI-Coveo: For aria-hidden, it seems when `<lightning-button-icon>` is rendered, attribute aria-hidden="false" is added by salesforce. 
Will add `aria-hidden="true"`  to the components.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -234,9 +236,12 @@ export default class QuanticCategoryFacet extends LightningElement {
       this.searchStatus?.state?.isLoading &&
       !this.searchStatus?.state?.hasError &&
       !this.searchStatus?.state?.firstSearchExecuted;
-    
-    const renderFacetEvent = new CustomEvent('renderFacet', {
-      detail: {id: this.facetId ?? this.field, shouldRenderFacet: !!this.hasParentsOrValues},
+
+    const renderFacetEvent = new CustomEvent('renderfacet', {
```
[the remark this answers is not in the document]
MSLI-Coveo: Will discard the change in this PR and open new a new PR to address the issue.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
```diff
@@ -463,8 +477,11 @@ export default class QuanticFacet extends LightningElement {
     );
   }
 
-  setFocusToFirstFacetValue() {
-    const focusTarget = this.template.querySelector('c-quantic-facet-value');
+  setFocusToFacetValue(index) {
+    console.log(this.template.querySelectorAll('c-quantic-facet-value')[index]);
```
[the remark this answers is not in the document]
MSLI-Coveo: console.logs removed

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -3,15 +3,15 @@
...
+    <c-quantic-card-container title={label} toggle-visibility={toggleVisibility}>
       <lightning-button-icon
         class={actionButtonCssClasses}
         slot="actions"
-        onclick={toggleFacetVisibility}
         onmousedown={preventDefault}
         icon-name={actionButtonIcon}
         alternative-text={actionButtonLabel}
         variant="bare"
+        tabindex="-1"
```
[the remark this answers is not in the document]
MSLI-Coveo: aria-hidden added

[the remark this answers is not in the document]
MSLI-Coveo: aria-hidden added

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
```diff
@@ -3,15 +3,16 @@
     <c-quantic-placeholder variant="card" number-of-rows={numberOfValues}></c-quantic-placeholder>
   </template>
   <template if:true={hasParentsOrValues}>
-    <c-quantic-card-container title={label}>
+    <c-quantic-card-container title={label} toggle-visibility={toggleVisibility}>
```
erocheleau, line 6: Hummm I don't think we should be adding the collapse function call to the quantic-card-container, mainly because the `isCollapsed` is the responsibility of the container of the quanticCardContainer and not the container itself.

I think it's doable to add an event `headerkeydown` for example that the card container can emit, and the facets can listen to that event to collapse.

That or we change the structure of the slot inside the card container and instead of having only a slot for actions, we have a slot for the header instead?

I think I prefer the first option.

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.html
```diff
@@ -2,14 +2,12 @@
   <host>
     <article class="card">
       <header class="slds-grid slds-p-horizontal_x-small slds-has-flexi-truncate">
-        <div class="slds-media__body">
-          <div class="slds-grid card__header">
+        <div class="slds-media__body" onclick={toggleVisibility}  onkeydown={onKeyDown} role="button" tabindex="0">
```
erocheleau, line 5:
```suggestion
        <div class="slds-media__body" onclick={handleHeaderClick} onkeydown={handleHeaderKeyDown} role="button" tabindex="0">
```

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
```diff
@@ -7,10 +7,29 @@ import { LightningElement, api } from 'lwc';
...
+   */
+  onKeyDown(evt) {
+    if (
+      (evt.code === 'Enter' || evt.code === 'Space') &&
+      this.toggleVisibility
+    ) {
+      evt.preventDefault();
+      this.toggleVisibility();
+    }
+  }
```
erocheleau, line 34:
```suggestion
  handleHeaderClick(evt) {
     const headerClickEvent = new CustomEvent('headerclick', {...evt});
     this.dispatchEvent(headerClickEvent);
  }
  
  handleHeaderKeyDown(evt) {
     const headerKeyDownEvent = new CustomEvent('headerkeydown', {...evt});
     this.dispatchEvent(headerKeyDownEvent);
  }
```

erocheleau: What do you think of this?

erocheleau (changes requested):

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.html
MSLI-Coveo: Yeah, the first approach is neat!  Will try implement the first solution!

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.html
btaillon: if possible, I recommend replacing this with a `button` element. This way, you wouldn't need `onkeydown`.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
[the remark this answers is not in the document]
btaillon: This feels like a bit much to read, I think it would be better if you copied the FocusTarget util from Atomic (but in a LWC-friendly way, perhaps like a class with a `onRender` method).

btaillon (commented):
accessibility looks good

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js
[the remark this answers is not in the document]
MSLI-Coveo: > This feels like a bit much to read, I think it would be better if you copied the FocusTarget util from Atomic (but in a LWC-friendly way, perhaps like a class with a `onRender` method).

Thanks for the advice, I am not sure we could use the focusTarget implementation from Atomic, the facets values would re-render value with same order but different HTML element, once one value is selected. Is there any other possible to find next focused HTML element without knowing its index within the list?

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
MSLI-Coveo: This implementation is awesome, added Implementation.

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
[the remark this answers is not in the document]
MSLI-Coveo: Property removed under latest implementation.

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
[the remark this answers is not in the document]
MSLI-Coveo: Property removed under latest implementation.

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
```diff
@@ -362,6 +363,8 @@ export default class QuanticTimeframeFacet extends LightningElement {
       format: this.formatFacetValue,
       element: this.template.host,
     });
+
+    this.toggleVisibility = this.toggleFacetVisibility.bind(this);
```
mmitiche, line 367: I don't think this is needed anymore.

mmitiche (approved):
Nice job 👌

mmitiche (approved):

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
```diff
@@ -7,10 +7,37 @@ import { LightningElement, api } from 'lwc';
...
+
+  /**
+   * Sets the focus on the card header.
+   * @api
+   * @type {VoidFunction}
+   */
+  @api setFocus() {
+    const focusTarget = this.template.querySelector('.card_focusable-header');
+    if (focusTarget) {
+      // @ts-ignore
```
erocheleau, line 25: Why the ts-ignore?

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
```diff
@@ -7,10 +7,37 @@ import { LightningElement, api } from 'lwc';
...
+   * @type {string}
+   */
+  @api title;
+
+  /**
+   * Sets the focus on the card header.
+   * @api
+   * @type {VoidFunction}
+   */
+  @api setFocus() {
```
erocheleau, line 22:
```suggestion
  @api setFocusOnHeader() {
```

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -363,6 +373,10 @@ export default class QuanticCategoryFacet extends LightningElement {
       this.facet.toggleSelect(item);
     }
     this.clearInput();
+    this.focusShouldBeInFacet = true;
```
erocheleau, line 376: Maybe 1 function to avoir repeating the same code everywhere?

`this.keepFocusInFacet('header'|'value');` ?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -444,11 +472,62 @@ export default class QuanticCategoryFacet extends LightningElement {
...
+      case 'facetHeader':
+        this.setFocusOnHeader();
+        return;
+      case 'facetValue':
+        if (this.values.length) {
+          this.setFocusOnFirstFacetValue();
+        } else {
+          this.setFocusOnLastNonActiveParent();
+        }
+    }
```
erocheleau, line 494:
```suggestion
    if(this.focusTarget.type === 'facetHeader') {
        this.setFocusOnHeader();
    } else if (this.focusTarget.type === 'facetValue') {
         if (this.values.length) {
          this.setFocusOnFirstFacetValue();
        } else {
          this.setFocusOnLastNonActiveParent();
        }
    }
```

Simplifies it a bit and removes the need for the switch, and the eslint-disable

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -444,11 +472,62 @@ export default class QuanticCategoryFacet extends LightningElement {
...
+  setFocusOnLastNonActiveParent() {
+    const nonActiveParents = this.template.querySelectorAll(
+      '.facet__non-active-parent > c-quantic-category-facet-value'
     );
+    const lastNonActiveParent = nonActiveParents[nonActiveParents.length - 1];
 
+    if (lastNonActiveParent) {
+      // @ts-ignore
+      lastNonActiveParent.setFocus();
+    }
```
erocheleau, line 523:
```suggestion
    if (nonActiveParents.length > 0) {
      // @ts-ignore
      nonActiveParents[nonActiveParents.length - 1].setFocus();
    }
```

I believe this does pretty much the same thing right?

packages/quantic/force-app/main/default/lwc/quanticCategoryFacet/quanticCategoryFacet.js
```diff
@@ -444,11 +472,62 @@ export default class QuanticCategoryFacet extends LightningElement {
...
+   * Sets the focus on the first facet value.
+   */
+  setFocusOnFirstFacetValue() {
     const focusTarget = this.template.querySelector(
-      'c-quantic-category-facet-value'
+      '.facet__value > c-quantic-category-facet-value'
+    );
+
+    if (focusTarget) {
+      // @ts-ignore
```
erocheleau, line 506: I think you should be able to export a type in quantic-facet-value right? and set the type annotation here?

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -86,6 +86,7 @@ export default class QuanticFacetValue extends LightningElement {
    */
   onSelect(evt) {
     evt.preventDefault();
+
```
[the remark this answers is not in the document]
erocheleau: You should apply your own comment :D

packages/quantic/force-app/main/default/lwc/quanticNumericFacet/quanticNumericFacet.js
```diff
@@ -500,4 +532,68 @@ export default class QuanticNumericFacet extends LightningElement {
...
+  /**
+   * Sets the focus on the target element.
+   */
+  setFocusOnTarget() {
+    this.focusShouldBeInFacet = false;
+    if (!this.focusTarget) {
+      return;
+    }
+    // eslint-disable-next-line default-case
+    switch (this.focusTarget.type) {
```
erocheleau, line 545: If you don't have a default, make it work with `if` ;)

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
```diff
@@ -417,12 +420,15 @@ export default class QuanticTimeframeFacet extends LightningElement {
       !this.searchStatus?.state?.firstSearchExecuted;
 
     this.hasResults = this.searchStatus.state.hasResults;
-    
-    const renderFacetEvent = new CustomEvent('renderFacet', {
```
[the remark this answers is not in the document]
erocheleau: Does this comment still apply?

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
```diff
@@ -665,4 +698,66 @@ export default class QuanticTimeframeFacet extends LightningElement {
...
+  /**
+   * Sets the focus on the target element.
+   */
+  setFocusOnTarget() {
+    this.focusShouldBeInFacet = false;
+    if (!this.focusTarget) {
+      return;
+    }
+    // eslint-disable-next-line default-case
+    switch (this.focusTarget.type) {
```
erocheleau, line 711: Same as others

erocheleau (changes requested):
Some changes to simplify a bit the code

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
nathanlb: I'd wager the object returned by lwc querySelector doesn't have `focus` defined.

packages/quantic/force-app/main/default/lwc/quanticDateFacet/quanticDateFacet.js
```diff
@@ -248,4 +276,52 @@ export default class QuanticDateFacet extends LightningElement {
...
+      // @ts-ignore
+      focusTarget.setFocus();
+    }
+  }
+
+  /**
+   * Sets the focus on the facet header.
+   */
+  setFocusOnHeader() {
+    const focusTarget = this.template.querySelector('c-quantic-card-container');
```
nathanlb, line 321: could also be `this.template.querySelector('c-quantic-card-container')?.setFocus();`

packages/quantic/force-app/main/default/lwc/quanticDateFacet/quanticDateFacet.js
```diff
@@ -19,6 +19,13 @@ import expandFacet from '@salesforce/label/c.quantic_ExpandFacet';
 /** @typedef {import("coveo").DateFacetValue} DateFacetValue */
 /** @typedef {import("coveo").SearchStatus} SearchStatus */
 /** @typedef {import("coveo").SearchEngine} SearchEngine */
+/**
```
nathanlb, line 22: this typedef seems to be defined repeatedly. Could it be imported from a common place?

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.html
```diff
@@ -2,14 +2,12 @@
   <host>
     <article class="card">
       <header class="slds-grid slds-p-horizontal_x-small slds-has-flexi-truncate">
-        <div class="slds-media__body">
-          <div class="slds-grid card__header">
+        <div class="slds-media__body card_focusable-header" onclick={handleHeaderClick} onkeydown={handleHeaderKeyDown} role="button" tabindex="0">
```
btaillon, line 5: If `<slot name="actions"></slot>` contains interactive elements, then I recommend making sure there's no parent interactive elements (See [Deque `nested-interactive`](https://dequeuniversity.com/rules/axe/4.2/nested-interactive)).

One solution If you want the header to still be clickable is to make it clickable but not indicate to screen readers that it is a button, and instead make a new button as a child which has the same behavior.

packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.html
```diff
@@ -4,15 +4,16 @@
...
+      <c-quantic-card-container title={label} onheaderclick={toggleFacetVisibility} onheaderkeydown={toggleFacetVisibility}>
         <lightning-button-icon
           class={actionButtonCssClasses}
           slot="actions"
-          onclick={toggleFacetVisibility}
           onmousedown={preventDefault}
           icon-name={actionButtonIcon}
           alternative-text={actionButtonLabel}
           variant="bare"
+          tabindex="-1"
```
btaillon, line 15: `aria-hidden` is specified twice here

packages/quantic/force-app/main/default/lwc/quanticDateFacet/quanticDateFacet.html
```diff
@@ -4,15 +4,16 @@
...
         <lightning-button-icon
           class="actionButtonCssClasses"
           slot="actions"
-          onclick={toggleFacetVisibility}
           onmousedown={preventDefault}
           icon-name={actionButtonIcon}
           alternative-text={actionButtonLabel}
           variant="bare"
+          tabindex="-1"
+          aria-hidden="true"
```
btaillon, line 16: tabIndex is missing here unlike the other files

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.html
```diff
@@ -6,15 +6,16 @@
...
-      <c-quantic-card-container title={label}>
+      <c-quantic-card-container title={label} onheaderclick={toggleFacetVisibility} onheaderkeydown={toggleFacetVisibility}>
         <lightning-button-icon
           slot="actions"
           class={actionButtonCssClasses}
           icon-name={actionButtonIcon}
           alternative-text={actionButtonLabel}
-          onclick={toggleFacetVisibility}
           onmousedown={preventDefault}
           variant="bare"
```
btaillon, line 16: `aria-hidden` is missing here

btaillon (commented):
accessibility look

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.html
```diff
@@ -2,14 +2,12 @@
...
+          <div class="slds-grid card__header" >
             <span class="slds-text-heading_small slds-truncate">{title}</span>
+            <div class="slds-float_right"><slot name="actions"></slot></div>
```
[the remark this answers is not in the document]
mmitiche: This was made by Moshi before, let me check again if it's really necessary.

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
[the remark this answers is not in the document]
mmitiche: not anymore, the change I was talking about was reverted.

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.html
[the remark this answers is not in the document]
mmitiche: This change was made to include the slot action in the `div` element playing the role of a `button` and that will be clicked to open and to close the facets.
QuanticCardContainer

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.css
```diff
@@ -7,4 +7,10 @@
     background-clip: padding-box;
     -webkit-box-shadow: 0 2px 2px 0 rgb(0 0 0 / 10%);
     box-shadow: 0 2px 2px 0 rgb(0 0 0 / 10%);
+}
+
+.card__header{
+    display: flex;
+    cursor: pointer;
+    justify-content: space-between;
```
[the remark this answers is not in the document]
mmitiche: no, this is will give an equivalent display of the QuanticCardContainer, the only change is that the cursor will be a pointer when the card container header is hovered since this element is now clickable and used to open and close the card.

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
mmitiche: exactly 👆

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.html
mmitiche: Thank you for pointing this out! however, we are not passing interactive elements in the `actions` slot. We passing a lightning-button component but we are adding the attribute aria-hidden="true"    to make it non-interactive.

packages/quantic/force-app/main/default/lwc/quanticDateFacet/quanticDateFacet.js
mmitiche: The reason I didn't define this type in only one place and then import it where needed is because this type differs between the different facet types, it's not exactly the same..

packages/quantic/force-app/main/default/lwc/quanticCardContainer/quanticCardContainer.js
erocheleau: Yea but can't we export the type and import here?

mmitiche: We can do something like this to avoid the ts-ignore:
```
/** @type {Element & {focus: function}} */
const focusTarget = this.template.querySelector('.card_focusable-header');
```

packages/quantic/force-app/main/default/lwc/quanticDateFacet/quanticDateFacet.js
nathanlb: Ah my bad then I hadn't noticed the differences

nathanlb (approved):

PR #2413, opened 2022-09-09
chore: enforce engine

louis-bompart:
This cause that:
image

Which is more explicit.

----
https://coveord.atlassian.net/browse/KIT-2024


ThibodeauJF (approved):
We'll need to enforce something more recent than `"npm": "8.x"` as there are issues with 8.5.5

louis-bompart:
> We'll need to enforce something more recent than `"npm": "8.x"` as there are issues with 8.5.5

I don't know which version of `npm` you wish to enforce, so I did not statute on that

PR #2418, opened 2022-09-12
feat(atomic): add warning for possible rendering issues with atomic-icon

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1652

### DIFF
--- a/packages/atomic/cypress/integration/common-assertions.ts
+++ b/packages/atomic/cypress/integration/common-assertions.ts
@@ -1,10 +1,10 @@
 import {getFocusableDescendants} from '../../src/utils/accessibility-utils';
 import {TestFixture} from '../fixtures/test-fixture';
-import {AriaLiveSelectors} from './aria-live-selectors';
 import {ComponentErrorSelectors} from './component-error-selectors';
 
 export interface ComponentSelector {
   // Setting JQuery<HTMLElement> is incompatible with Stencil's HTML elements
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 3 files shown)


packages/atomic/src/components/common/atomic-icon/atomic-icon.tsx
```diff
@@ -61,12 +66,29 @@ export class AtomicIcon implements InitializableComponent<AnyBindings> {
     }
   }
 
+  private isUrlSVG(url: string) {
+    return /.+\.svg$/i.test(url);
+  }
```
btaillon, line 71: Instead of validating the URL, I would prefer to validate the response, since somebody could have an URL that returns an SVG without the extension like `https://mywebsite.invalid/user/benjamin/favorite-emoji`.

ThibodeauJF: Good point!

packages/atomic/cypress/integration/common-assertions.ts
```diff
@@ -81,6 +81,20 @@ export function assertConsoleErrorMessage(msg: string) {
   });
 }
 
+export function assertConsoleWarning(warn = true) {
```
olamothe, line 84: I know you did not cause those linter errors above, but could you fix them at the same time as touching this file ? 😄

olamothe (approved):

packages/atomic/cypress/integration/common-assertions.ts
ThibodeauJF: Sure!

PR #2419, opened 2022-09-12
test(atomic): add small wait after input submit test

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2027

### DIFF
--- a/packages/atomic/cypress/integration-insights-panel/insights-panel.cypress.ts
+++ b/packages/atomic/cypress/integration-insights-panel/insights-panel.cypress.ts
@@ -164,6 +164,7 @@ describe('Insights panel test suites', () => {
         .shadow()
         .find('input')
         .type('test{enter}');
+      cy.wait(200);
     });
     it('display query summary', () => {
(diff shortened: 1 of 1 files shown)


packages/atomic/cypress/integration-insights-panel/insights-panel.cypress.ts
```diff
@@ -164,6 +164,7 @@ describe('Insights panel test suites', () => {
         .shadow()
         .find('input')
         .type('test{enter}');
+      cy.wait(200);
```
btaillon, line 167: I'm a bit confused as to why this is necessary, since in theory the next `it` should be waiting for `querySummary()` to exist.

If possible, I prefer to rely on Cypress's retry mechanism.

ThibodeauJF: It triggers a search which updates the UI, same reason why we have these wait everywhere in the Search cypress tests

olamothe (approved):

PR #2426, opened 2022-09-13
feat(atomic-react): add iife output for atomic react + publish to CDN

olamothe:
https://coveord.atlassian.net/browse/KIT-2025


packages/atomic-react/rollup.config.js
```diff
@@ -0,0 +1,40 @@
...
+  plugins: minify ? [terser()] : [],
+});
+
+export default {
+  input: 'src/index.ts',
+  output: [outputIIFE({minify: true}), outputIIFE({minify: false})],
+  external: ['react', 'react-dom', '@coveo/atomic', '@coveo/atomic/headless'],
+  plugins: [
+    nodePolyfills(),
+    typescript({tsconfig: 'tsconfig.umd.json'}),
```
btaillon, line 27: perhaps this should be named `tsconfig.iife.json`

packages/atomic-react/rollup.config.js
```diff
@@ -0,0 +1,40 @@
+import typescript from '@rollup/plugin-typescript';
+import commonjs from '@rollup/plugin-commonjs';
+import {nodeResolve} from '@rollup/plugin-node-resolve';
+import replace from '@rollup/plugin-replace';
+import nodePolyfills from 'rollup-plugin-polyfill-node';
+import {terser} from 'rollup-plugin-terser';
+
+const outputIIFE = ({minify}) => ({
+  file: `dist/iife/atomic-react${minify ? '.min' : ''}.js`,
+  format: 'iife',
```
btaillon, line 10: why exactly did UMD not work?

olamothe: We already have a a common js output produced by the TypeScript compiler.

I don't think we should remove/rename/modify it to UMD (ie: instead of being produced by TypeScript to common js, now using rollup to produce browser compatible UMD). 

The reason being that it does have an impact on various tools (bundler) down the line which *may* break in unexpected manner. Frontend tools are pretty finicky, after all ;)

So to stay on the safe and coupled with the fact that adding a umd (commonjs + iife) seems to overlap with the current common js output, I opted to ad an iife only.

Thoughts ?

ThibodeauJF (approved):

packages/atomic-react/README.md
```diff
@@ -10,6 +10,13 @@ The integration of React-based projects (using JSX) with Web based components ca
 
 `npm i @coveo/atomic-react`
 
+### Alternate installation -- CDN
+
+The library is also available from Coveo's CDN as an `IIFE` (Immediately Invoked Function Expression). This is an approach that we recommend only for quick prototyping and testing, and that can be used with very minimal front end tooling and build tools.
```
jpmarceau, line 15:
```suggestion
The library is also available in the Coveo CDN as an `IIFE` (Immediately Invoked Function Expression). This is an approach that we recommend only for quick prototyping and testing, and that can be used with very minimal front-end tooling and build tools.
```

packages/atomic-react/README.md
```diff
@@ -10,6 +10,13 @@ The integration of React-based projects (using JSX) with Web based components ca
...
+
+The `IIFE` approach in production is discouraged - the browser has to download the entire library code, regardless of which components are actually used. Using a bundler (for example Webpack) is recommended.
```
jpmarceau, line 17:
```suggestion
We recommend against using the `IIFE` approach in production. The browser has to download the entire library code, regardless of which components are actually used. We rather recommend using a bundler in this situation (for example Webpack).
```

packages/atomic-react/README.md
```diff
@@ -10,6 +10,13 @@ The integration of React-based projects (using JSX) with Web based components ca
...
+
+You can read more on this approach down below.
```
jpmarceau, line 19:
```suggestion
You can read more about this approach [below](#usage-with-cdn-scripts--iife-).
```
(not totally sure about the link though, and it depends on you accepting my suggestion below 😉)

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
   }}
 ></AtomicSearchInterface>;
 ```
+
+## Usage with CDN scripts
```
jpmarceau, line 276:
```suggestion
## Usage with CDN Scripts (IIFE)
```
(I think)

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+
+For this approach to work, different scripts need to be pulled in the page in the correct order, using proper external dependencies with matching versions.
```
jpmarceau, line 278:
```suggestion
For this approach to work, you need to pull different scripts in the page in the correct order, using proper external dependencies with matching versions.
```

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+
+First, identify the version of `@coveo/atomic` that is used by `@coveo-atomic-react`.
+
+This can be done with `npm view @coveo/atomic-react`, and identify the `@coveo/atomic` dependencies:
```
jpmarceau, line 282:
```suggestion
You can do that by running `npm view @coveo/atomic-react`, and then identifying the `@coveo/atomic` dependency:
```

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+.tarball: https://registry.npmjs.org/@coveo/atomic-react/-/atomic-react-1.23.11.tgz
+.shasum: 64ba0a5f686e3638b1180910f6fed5e4980bd9e2
+.integrity: sha512-/qI5O7SWNinBYWAedRW62ko0oF4c/hfapgqKe9xSq/NJfg+KBzcR3SbdyhpaUAWpPX1910GP9cMyr4wf8irJnQ==
+.unpackedSize: 771.1 kB
+
+dependencies:
+@coveo/atomic: 1.108.2
+```
+
+In the above example, we can see that of the time of this writing, the current latest version of `@coveo/atomic-react` is `1.23.11`, and that it is using `@coveo/atomic` at version `1.108.2`.
```
jpmarceau, line 302:
```suggestion
In the above example, you can see that of the time of this writing, the current latest version of `@coveo/atomic-react` is `1.23.11`, and that it is using `@coveo/atomic` at version `1.108.2`.
```

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+
+Then, the same need to be done for the matching version of `@coveo/headless` used by the matching `@coveo/atomic` version.
```
jpmarceau, line 304:
```suggestion
Then, you need to do the same to find the matching version of `@coveo/headless` used by the target`@coveo/atomic` version.
```

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+dependencies:
+@coveo/bueno: 0.42.1                  @stencil/store: 1.5.0                 focus-visible: 5.2.0
+@coveo/headless: 1.103.3              coveo-styleguide: 9.34.4              i18next-http-backend: 1.4.1
+@popperjs/core: ^2.11.6               dayjs: 1.11.5                         i18next: 20.6.1
+@salesforce-ux/design-system: ^2.16.1 dompurify: 2.3.10                     stencil-inline-svg: 1.1.0
+@stencil/core: 2.17.3                 escape-html: 1.0.3                    ts-debounce: ^4.0.0
+
+```
+
+In the above example, we can see that that `@coveo/atomic@1.108.2` has a dependency on `@coveo/headless@1.103.3`.
```
jpmarceau, line 331:
```suggestion
In the above example, you can see that `@coveo/atomic@1.108.2` has the dependency `@coveo/headless@1.103.3`.
```

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+
+In summary, as of the time of this writing, the versions that we should keep in mind are the following:
```
jpmarceau, line 333:
```suggestion
To summarize, as of the time of this writing, the versions that we should keep in mind are the following:
```

packages/atomic-react/README.md
```diff
@@ -10,6 +10,13 @@ The integration of React-based projects (using JSX) with Web based components ca
 
 `npm i @coveo/atomic-react`
 
+### Alternate installation -- CDN
```
jpmarceau, line 13:
```suggestion
### Alternate installation — CDN
```
(you can't tell with the GitHub editor, but the dash I'm using it the length of two small dashes. It's an em dash, to be precise)

packages/atomic-react/README.md
```diff
@@ -265,3 +272,133 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+  <!-- Note the matching major and minor version as explained above -->
+  <script crossorigin type="module" src="https://static.cloud.coveo.com/atomic/v1.108/atomic.esm.js"></script>
+  
+  <!--And then finally @coveo/atomic-react is included -->
+  <!-- Note the matching major and minor version as explained above -->
+  <script crossorigin src="https://static.cloud.coveo.com/atomic-react/v1.23/iife/atomic-react.min.js"></script>
+</head>
+```
+
+Once this is done, you can now start using `CoveoAtomicReact` directly with an inline script tag:
```
jpmarceau, line 363:
```suggestion
Once this is done, you can start using `CoveoAtomicReact` directly with an inline script tag: 
```
(since you already set a time frame with `Once this is done`)

jpmarceau (approved):
Nice, I just have fairly small writing suggestions. 👍

PR #2427, opened 2022-09-14
feat(atomic): add onRedirect event to search box, make Stencil sample a SPA

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1836


packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -144,6 +155,23 @@ export class AtomicSearchBox {
...
+   * Event that is emitted when redirection is triggered. If the default behaviour is not prevented, the search box will directly change the URL.
+   *
+   * @example
+   * <atomic-search-box
+   *   redirection-url="/search"
+   *   onRedirect={(e) => {
+   *     e.preventDefault();
+   *     // handle redirection
+   *   }}
+   * ></atomic-search-box>
```
btaillon, line 168: I think we should give an example that works without JSX.

Also, I would add code highlighting.

```suggestion
   * @example
   * ```html
   * <script>
   *   document.querySelector('atomic-search-box').addEventListener((e) => {
   *     e.preventDefault();
   *     // handle redirection
   *   });
   * </script>
   * ...
   * <atomic-search-box redirection-url="/search"></atomic-search-box>
   * ```
```

btaillon (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
ThibodeauJF: Great!

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -144,6 +155,26 @@ export class AtomicSearchBox {
    */
   @Prop({reflect: true}) public clearFilters = true;
 
+  /**
+   * Event that is emitted when redirection is triggered. If the default behaviour is not prevented, the search box will directly change the URL.
```
olamothe, line 159: I would buff up the explanation (and pull in @jpmarceau at the same time ;) )

We need to explain why you would want the redirection to stop. Need to add an explanation on the relationship with SPA, a possible link to the code sample.

I also think we should broaden a bit the scope of this ticket (ie: open new ticket) to figure out how Atomic standalone search box would work with different frameworks and popular router, and then add new usage article on docs.coveo.com related to those topics.

This does not need to happen right now, but it needs to be put somewhere in the backlog and planified soon-ish IMO.

Standalone search box is extremely important, hard to get right as an implementer (there's a lot of interaction that needs to happen outside the scope of Atomic itself), used in pretty much every implementation, and always causes bugs/issues during implementation. This is coming from experience with JSUI.

The amount of problem the standalone search box caused over the years is... impressive :D So let's try to limit that as much as possible.

olamothe (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
jpmarceau: Should I wait until you adjust the description before reviewing @ThibodeauJF?
I could suggest something, but I'd need an answer to the following question first please: `why you would want the redirection to stop`? I get that if someone doesn't define a `redirection-url` then the search box is part of a SPA, but I don't get how it relates to this description here 🤔 I also understand that someone could use a pipeline redirect, but I don't see how that would stop the redirection...

ThibodeauJF: @jpmarceau This options allows to change the implementation of the redirection inside the browser, instead of doing... 
```js
window.location.href = redirectToUrl;
```
which is not ideal for some frameworks, which have their own routers e.g.,
```jsx
onRedirect={(e) => {
  e.preventDefault();
  const {redirectTo, value} = e.detail;
  const path = `${redirectTo}#q=${value}`;
  this.router.push(path);
}}
```

So it's not "stopping the redirection" it's changing its exact behaviour.

Also created a Jira for the usage docs as Olivier recommended https://coveord.atlassian.net/browse/KIT-2032

jpmarceau, line 159:
```suggestion
   * Event that is emitted when a standalone search box redirection is triggered. By default, the search box will directly change the URL and redirect accordingly, so if you want to handle the redirection differently, use this event.
```

jpmarceau (approved):
👍

PR #2429, opened 2022-09-14
fix(atomic): add rendering condition to insight result list

ThibodeauJF:
https://coveord.atlassian.net/browse/SVCC-2470

### DIFF
--- a/packages/atomic/src/components/insight/atomic-insight-result-list/atomic-insight-result-list.tsx
+++ b/packages/atomic/src/components/insight/atomic-insight-result-list/atomic-insight-result-list.tsx
@@ -156,6 +156,12 @@ export class AtomicInsightResultList {
     if (this.resultListState.hasError) {
       return null;
     }
+    if (
+      this.resultListState.firstSearchExecuted &&
+      !this.resultListState.hasResults
+    ) {
+      return;
+    }
     return (
       <Host>
         {this.templateHasError && <slot></slot>}


nathanlb (approved):
Looks like this will work 👍

packages/atomic/src/components/insight/atomic-insight-result-list/atomic-insight-result-list.tsx
```diff
@@ -156,6 +156,12 @@ export class AtomicInsightResultList {
     if (this.resultListState.hasError) {
       return null;
     }
+    if (
```
olamothe, line 159: Could also be moved/extracted to the piece of code that controls the placeholder display couple lines below.

ie: 
```
{!this.bindings.store.isAppLoaded() && (
  <ListDisplayResultsPlaceholder
    resultsPerPageState={this.resultPerPageState}
    display="list"
    density={this.density}
    imageSize={this.imageSize}
  />
)}
```

could become: 
```
{this.shouldDisplayPlaceholder() && (
  <ListDisplayResultsPlaceholder
    resultsPerPageState={this.resultPerPageState}
    display="list"
    density={this.density}
    imageSize={this.imageSize}
  />
)}
```

olamothe (approved):

packages/atomic/src/components/insight/atomic-insight-result-list/atomic-insight-result-list.tsx
ThibodeauJF: I'd rather not render anything (returning null/undefined) than the list-wrapper, list-root, etc.

In that case you want the result list to be completely hidden.

PR #2430, opened 2022-09-14
feat(atomic): add error message when result-value-text is used with multi value field

olamothe:
https://coveord.atlassian.net/browse/KIT-2031

### DIFF
--- a/packages/atomic/src/components/search/result-template-components/atomic-result-text/atomic-result-text.tsx
+++ b/packages/atomic/src/components/search/result-template-components/atomic-result-text/atomic-result-text.tsx
@@ -64,30 +64,35 @@ export class AtomicResultText implements InitializableComponent {
     }
   }
 
+  private possiblyWarnOnBadFieldType() {
+    const resultValueRaw = ResultTemplatesHelpers.getResultProperty(
+      this.result,
+      this.field
+    );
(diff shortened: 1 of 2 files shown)


ThibodeauJF (approved):

packages/atomic/src/components/search/result-template-components/atomic-result-text/atomic-result-text.tsx
```diff
@@ -63,14 +64,32 @@ export class AtomicResultText implements InitializableComponent {
     }
   }
 
+  private possiblyWarnOnBadFieldType() {
+    const resultValueRaw = ResultTemplatesHelpers.getResultProperty(
+      this.result,
+      this.field
+    );
+    if (isArray(resultValueRaw)) {
+      this.bindings.engine.logger.error(
```
btaillon, line 73: Instead of logging the error in the logger like this, I think we should leverage `this.error` just like `atomic-result-date` does. We could move the validation to `componentWillRender`.

olamothe: In this particular case, I'm not sure that rendering a end user error is a good idea, since this is inside a result template.

PR #2431, opened 2022-09-14
feat(stencil): make SPA with a single interface, switching search box type

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1836

### Some issues that need to be fixed (upcoming)
1. Search box issue of “losing connection” with the suggestions going from `/` (standalone) to `/search` (standard) [JIRA](https://coveord.atlassian.net/browse/KIT-2034)
2. Facets are not unregistering from the Headless state (& the Atomic store), it's crashing warnings in the console [JIRA](https://coveord.atlassian.net/browse/KIT-2033)
3. The `atomic-result` are crashing errors in the console when going to search page more than once (see pic) [JIRA](https://coveord.atlassian.net/browse/KIT-2035)

Screen Shot 2022-09-14 at 4 18 23 PM


btaillon (approved):

packages/samples/stencil/src/components/app-root/app-root.tsx
```diff
@@ -1,30 +1,60 @@
...
-              <Header>
-                <standalone-search-box router={Router}></standalone-search-box>
-              </Header>
-              <h1>Home</h1>
-            </Fragment>
-          )}
-        />
-        <Route path="/search" render={() => <search-page></search-page>} />
-      </Router.Switch>
+      <atomic-search-interface>
```
ThibodeauJF, line 32: Along with the standalone search box (switching the redirection-url) this is the best user experience with the Stencil Router, maybe because the search box won't fully be recreated. I found a few bugs inside Atomic.

In the future we should improve the atomic-external or the search box so it's able to dump its engine and hook to a new one. 🤞

PR #2434, opened 2022-09-15
fix(headless): insightFacet and insightDateFacet controllers set to use the correct analytics actions

mmitiche:
### The issue:
The `HeadlessInsightFacet` and `HeadlessInsightDateFacet` controllers are using the search analytics actions instead of the insight analytics actions.

The two controllers are using the analytics actions `logFacetDeselect` and `logFacetSelect` imported from `facet-set-analytics-actions` instead of the `facet-set-insight-analytics-actions` file.

### The solution:
The two controllers updated to use the analytics actions imported from the `facet-set-insight-analytics-actions` file.


btaillon (approved):

packages/headless/src/features/facets/range-facets/generic/range-facet-utils.ts
```diff
@@ -20,3 +24,15 @@ export const getAnalyticsActionForToggleRangeFacetSelect = (
     ? logFacetDeselect(payload)
     : logFacetSelect(payload);
 };
+
+export const getInsightAnalyticsActionForToggleRangeFacetSelect = (
```
ThibodeauJF, line 28: Would be best if it's in its own file correctly scoped to **insight** so things don't get too convoluted

packages/headless/src/features/facets/range-facets/generic/range-facet-utils.ts
```diff
@@ -3,6 +3,10 @@ import {
   logFacetDeselect,
   logFacetSelect,
 } from '../../facet-set/facet-set-analytics-actions';
+import {
+  logFacetDeselect as logInsightFacetDeselect,
+  logFacetSelect as logInsightFacetSelect,
+} from '../../facet-set/facet-set-analytics-actions';
```
ThibodeauJF, line 9: Description says 

> The two controllers updated to use the analytics actions imported from the facet-set-insight-analytics-actions file.

But here it's importing the same facet-set-analytics-actions file

mmitiche: My bad 😅I just updated the import, thank you!

aelfaleh (approved):
LGTM

nathanlb (approved):
gg

PR #2435, opened 2022-09-15
fix(headless): insightSearchBox controller updated to use the correct analytics action when performing a search

mmitiche:
### The issue:
The `HeadlessInsightSearchBox` controller was using the search analytics actions instead of the insight analytics actions.

The `HeadlessInsightSearchBox` controller was only returning a `HeadlessCoreSearchBox` controller which is using the analytics actions `logSearchboxSubmit` imported from the `query-analytics-actions.ts` in its submit function.
The `HeadlessInsightSearchBox` needs to use the  `logSearchboxSubmit` action imported from the`query-insight-analytics-actions.ts` file.

### The solution:
The controller is updated to use the analytics actions imported from the `query-insight-analytics-actions.ts` file by the redefinition of the submit function.


packages/headless/src/controllers/insight/search-box/headless-insight-search-box.ts
```diff
@@ -40,9 +48,41 @@ export function buildSearchBox(
...
+
+    dispatch(
+      prepareForSearchWithQuery({
+        q: getValue(),
+        enableQuerySyntax,
+        clearFilters,
+      })
+    );
+
+    await dispatch(executeSearch(analytics));
```
lbergeron, line 78: It would be nice to add a test to ensure the analytics action is the one being dispatched.

mmitiche: Sure thing 👌

packages/headless/src/controllers/insight/search-box/headless-insight-search-box.ts
```diff
@@ -40,9 +48,41 @@ export function buildSearchBox(
   engine: InsightEngine,
   props: SearchBoxProps = {}
 ): SearchBox {
-  return buildCoreSearchBox(engine, {
+  const {dispatch} = engine;
+  const searchBox = buildCoreSearchBox(engine, {
```
ThibodeauJF, line 52: Instead of reimplementing part of the search box, could the analytics action be an option on the `headless-core-search-box` ?

btaillon: perhaps the `submit` function on the core searchbox could take the analytics as a parameter, and both the search & insight search boxes would "re-implement" the function without parameters by calling the core search box's with the correct parameter.

mmitiche: Thank you for the suggestions, I went with the second one! let me what you think 👌

packages/headless/src/controllers/insight/search-box/headless-insight-search-box.ts
mmitiche: tests added 👌

ThibodeauJF (approved):
💯

olamothe (approved):

PR #2438, opened 2022-09-15
fix(atomic): render result list content when it's ready

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2035

Easy fix, but hard to find! The result templates (default template included) did not have time to register before the render was trigger. There seem to already have a `ready` variable available.


olamothe:
Good find !

packages/atomic/src/components/search/result-lists/result-list-common.tsx
```diff
@@ -84,6 +84,7 @@ export interface RenderListOptions extends DisplayOptions {
   getContentOfResultTemplate(
     result: Result | FoldedResult
   ): HTMLElement | DocumentFragment;
+  ready: boolean;
```
btaillon, line 87: It looks like `ready` is set  in the `onReady` callback anyway, so perhaps we should handle it within `result-list-common`

btaillon (approved):

packages/atomic/src/components/search/result-lists/result-list-common.tsx
ThibodeauJF: ready is a @State property in the components, state property change trigger re-renders, a private property in the list common wouldnt

packages/atomic/src/components/search/result-lists/result-list-common.tsx
```diff
@@ -252,8 +253,12 @@ export class ResultListCommon {
     resultsPerPageState,
     setListWrapperRef,
     getContentOfResultTemplate,
+    ready,
   }: RenderListOptions) {
     this.updateBreakpoints?.(host);
+    if (!ready) {
+      return;
```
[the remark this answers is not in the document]
ThibodeauJF: Hmm perhaps, but it's not noticeable, simply a race condition for the time it takes for the result template method registration to run. In the use case where its useful, the results are rendered after the search already has been executed.

PR #2439, opened 2022-09-15
fix(atomic): use correct delimiting character for category facet headless controller

olamothe:
https://coveord.atlassian.net/browse/KIT-2036


btaillon (approved):

btaillon (changes requested):
isn't this breaking?

packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -224,7 +224,9 @@ export class AtomicCategoryFacet
       sortCriteria: this.sortCriteria,
       facetSearch: {numberOfValues: this.numberOfValues},
       basePath: this.basePath
-        ? this.basePath.split(',').map((value) => value.trim())
+        ? this.basePath
+            .split(this.delimitingCharacter)
+            .map((value) => value.trim())
```
btaillon, line 229: It's pretty intuitive to me that the prop is split by commas even if that's ont the delimiting character 🤔  I'm curious what Stencil's array props do.

Either way, we should update the prop's description to reflect whichever way we go.

ThibodeauJF:
Unfortunately yes it's breaking, we need to do something "uglier" until the next major. Maybe have a way to escape the commas in the path value

olamothe:
@btaillon @ThibodeauJF : Ready to re-review.

Old format (comma separated) will still work properly, but will not print a warning to the console about deprecation, and the need to move to a JSON string representation.

If you like this approach, after merging this PR I can also do the job of going through other comma separated list properties, and move them to this new format.

packages/atomic/src/utils/props-utils.ts
```diff
@@ -30,6 +38,56 @@ export function MapProp(opts?: MapPropOptions) {
...
+    opts.deprecationWarning
+      ? console.warn(
+          `${msg} This will be enforced in the next major version`,
+          other
+        )
+      : console.error(msg, other);
+
+  return (component: ComponentInterface, variableName: string) => {
+    const {componentWillLoad} = component;
+    const attributeWithBackets = camelToKebab(variableName);
```
btaillon, line 52: Did you mean to rename this? There's no more brackets.

btaillon (approved):

packages/atomic/src/utils/props-utils.ts
olamothe: I removed the bracket annotation, yes

ThibodeauJF (approved):

PR #2443, opened 2022-09-16
fix(quantic): links generated inside facet values

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2453

This PR fixes an issue where an hyperlink could be generated inside a facet value if the value contains an URL, or an email address.

This is caused by the [`lightning-input-rich-text`](https://developer.salesforce.com/docs/component-library/bundle/lightning-formatted-rich-text/documentation) component that is used internally. Out of the box, this component automatically detects URLs and emails in a text and render them as links. Fixing it is very simple, we just have to specify the `disable-linkify` option on the component.


mmitiche (approved):
💯

PR #2445, opened 2022-09-16
chore(atomic): add doc for exported suggestions types, add support for insight search box

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2037

### DIFF
--- a/packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
+++ b/packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
 import {buildCustomEvent} from '../../../utils/event-utils';
 import {closest} from '../../../utils/utils';
 import {Bindings} from '../atomic-search-interface/atomic-search-interface';
-export interface SearchBoxSuggestionElement {
-  key: string;
-  content: Element | VNode;
-  onSelect?(e: Event): void;
 
+/**
+ * Element which will be rendered in the list of suggestions.
+ */
+export interface SearchBoxSuggestionElement {
+  /**
+   * Stable identity which enables Stencil to reuse DOM elements for better performance.
+   * The best way to pick a key is to use a string that uniquely identifies that list item among its siblings (often your data will already have IDs).
+   */
+  key: string;
+  /**
+   * Rendered content of the element.
+   */
+  content: Element | VNode;
+  /**
+   * Hook called when the selection is selected.
+   * @param e DOM event.
+   */
+  onSelect?(e: Event): void;
+  /**
+   * The query associated with the suggestion, which will become the suggested query that would be sent if the search is executed.
+   */
   query?: string;
+  /**
+   * For improved accessibility, provide this property with additional information.
+   * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label
+   */
   ariaLabel?: string;
+  /**
+   * Adds a specific shadow part attribute that can be selected with the CSS ::part pseudo-element.
+   * https://developer.mozilla.org/en-US/docs/Web/CSS/::part
+   */
   part?: string;
+  /**
+   * Hide the suggestion if it's the last in the list.
+   */
   hideIfLast?: boolean;
 }
+
+/**
+ * List of suggestions that will be displayed along other lists when the search box's input is selected.
+ */
 export interface SearchBoxSuggestions {
+  /**
+   * The position of the list of the suggestions relative to the others.
+   */
   position: number;
+  /**
+   * Whether the suggestions should be listed in the right or left panel.
+   */
   panel?: 'left' | 'right';
+  /**
+   * Method that returns the list of elements which will be rendered in the list of suggestions.
+   */
   renderItems(): SearchBoxSuggestionElement[];
+  /**
+   * Hook called when when the user changes the search box's input value.
+   */
   onInput?(): Promise<unknown> | void;
+  /**
+   * Hook called when the suggested query changes.
+   * @param q The new suggested query.
+   */
   onSuggestedQueryChange?(q: string): Promise<unknown> | void;
 }
 
+/**
+ * Event sent from the registered query suggestions to the parent search box.
+ */
 export type SearchBoxSuggestionsEvent = (
+  /**
+   * The bindings passed from the search box to the suggestions.
+   */
   bindings: SearchBoxSuggestionsBindings
 ) => SearchBoxSuggestions;
 
+const searchBoxElements = ['atomic-search-box', 'atomic-insight-search-box'];
+
+/**
+ * The bindings passed from the search box to the suggestions.
+ */
 export interface SearchBoxSuggestionsBindings extends Bindings {
+  /**
+   * The unique id of the search box.
+   */
   id: string;
+  /**
+   * Whether the search box is standalone.
+   */
   isStandalone: boolean;
+  /**
+   * The SearchBox headless controller.
+   */
   searchBoxController: SearchBox;
+  /**
+   * The amount of queries displayed when the user interacts with the search box.
+   * By default, a mix of query suggestions and recent queries will be shown.
+   */
   numberOfQueries: number;
+  /**
+   * Whether to clear all active query filters when the end user submits a new query from the search box.
+   */
   clearFilters: boolean;
+  /**
+   * Retrieves the suggestion query, meaning the query that will be sent if the search is executed.
+   */
   suggestedQuery(): string;
+  /**
+   * Removes the current suggestions.
+   */
   clearSuggestions(): void;
+  /**
+   * Triggers the retrieval of updated suggestions.
+   */
   triggerSuggestions(): void;
+  /**
+   * Retrieves the current suggestions.
+   */
   getSuggestions(): SearchBoxSuggestions[];
+  /**
+   * Retrieves the current suggestions elements.
+   */
   getSuggestionElements(): SearchBoxSuggestionElement[];
 }
 
+/**
+ * Dispatches an event which retrieves the `SearchBoxSuggestionsBindings` on a configured parent search box.
+ * @param event Event sent from the registered query suggestions to the parent search box.
+ * @param element Element on which to dispatch the event, which must be the child of a configured search box.
+ */
 export const dispatchSearchBoxSuggestionsEvent = (
   event: SearchBoxSuggestionsEvent,
   element: HTMLElement
@@ -46,11 +141,11 @@ export const dispatchSearchBoxSuggestionsEvent = (
     buildCustomEvent('atomic/searchBoxSuggestion/register', event)
   );
 
-  const hasParentSearchBox = closest(element, 'atomic-search-box');
-
-  if (!hasParentSearchBox) {
+  if (!closest(element, searchBoxElements.join(', '))) {
     throw new Error(
-      `The "${element.nodeName.toLowerCase()}" component was not handled, as it is not a child of an "atomic-search-box" component`
+      `The "${element.nodeName.toLowerCase()}" component was not handled, as it is not a child of the following elements: ${searchBoxElements.join(
+        ', '
+      )}`
     );
   }
 };
--- a/packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
+++ b/packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
@@ -92,16 +92,15 @@ export interface SearchBoxSuggestionsBindings extends Bindings {
    */
   id: string;
   /**
-   * Whether the search box is standalone.
+   * Whether the search box is [standalone](https://docs.coveo.com/en/atomic/latest/usage/ssb/).
    */
   isStandalone: boolean;
   /**
-   * The SearchBox headless controller.
+   * The [search box headless controller](https://docs.coveo.com/en/headless/latest/reference/search/controllers/search-box/).
    */
   searchBoxController: SearchBox;
   /**
-   * The amount of queries displayed when the user interacts with the search box.
-   * By default, a mix of query suggestions and recent queries will be shown.
+   * The amount of queries displayed when the user interacts with the search box, property set on the [Atomic Search box](https://docs.coveo.com/en/atomic/latest/reference/components/atomic-search-box/#properties).
    */
   numberOfQueries: number;
   /**
--- a/packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
+++ b/packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
@@ -23,7 +23,7 @@ export interface SearchBoxSuggestionElement {
    */
   onSelect?(e: Event): void;
   /**
-   * The query associated with the suggestion, which will become the suggested query that would be sent if the search is executed.
+   * The query associated with the suggestion which will replace the query in the search box if the suggestion is selected.
    */
   query?: string;
   /**
@@ -51,7 +51,7 @@ export interface SearchBoxSuggestions {
    */
   position: number;
   /**
-   * Whether the suggestions should be listed in the right or left panel.
+   * Whether the suggestions should be listed in the right or left panel. By default, the suggestions are listed in the right panel.
    */
   panel?: 'left' | 'right';
   /**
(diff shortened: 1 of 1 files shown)


spiracha-coveo (changes requested):
Great annotations. I left a few edits and a couple of clarifying questions that I had. Thanks!

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
 import {buildCustomEvent} from '../../../utils/event-utils';
 import {closest} from '../../../utils/utils';
 import {Bindings} from '../atomic-search-interface/atomic-search-interface';
+
+/**
+ * Element which will be rendered in the list of suggestions.
+ */
 export interface SearchBoxSuggestionElement {
+  /**
+   * Stable identity which enables Stencil to reuse DOM elements for better performance.
```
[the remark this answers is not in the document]
ThibodeauJF: It's not necessarily unique, but it has to say stable in between renders. I used the Stencil doc https://stenciljs.com/docs/templating-jsx#loops & React doc for inspiration https://reactjs.org/docs/lists-and-keys.html#keys

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   */
   content: Element | VNode;
+  /**
+   * Hook called when the selection is selected.
+   * @param e DOM event.
+   */
   onSelect?(e: Event): void;
-
+  /**
+   * The query associated with the suggestion, which will become the suggested query that would be sent if the search is executed.
```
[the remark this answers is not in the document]
ThibodeauJF: Yes

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   * Method that returns the list of elements which will be rendered in the list of suggestions.
+   */
   renderItems(): SearchBoxSuggestionElement[];
+  /**
+   * Hook called when when the user changes the search box's input value.
+   */
   onInput?(): Promise<unknown> | void;
+  /**
+   * Hook called when the suggested query changes.
+   * @param q The new suggested query.
```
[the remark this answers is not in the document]
ThibodeauJF: On input is when the user types in the search box. Usually all query suggestions will update when this happens.

onSuggestedQueryChange happens more often, whenever the user travels through le list of suggestions. The suggestions themselves won't be updated but the instant results will since they depend on the suggested query.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+ */
 export interface SearchBoxSuggestions {
+  /**
+   * The position of the list of the suggestions relative to the others.
+   */
   position: number;
+  /**
+   * Whether the suggestions should be listed in the right or left panel.
+   */
   panel?: 'left' | 'right';
```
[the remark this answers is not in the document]
ThibodeauJF: The default is the "right"

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   * The best way to pick a key is to use a string that uniquely identifies that list item among its siblings (often your data will already have IDs).
+   */
   key: string;
+  /**
+   * Rendered content of the element.
+   */
   content: Element | VNode;
+  /**
+   * Hook called when the selection is selected.
+   * @param e DOM event.
```
btaillon, line 22: I think we should detail that this is where somebody should execute other actions such as selecting the suggestion. Maybe part of the query suggestion component could be used as an example.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   */
   hideIfLast?: boolean;
 }
+
+/**
+ * List of suggestions that will be displayed along other lists (e.g recent queries) when the search box's input is selected.
+ */
 export interface SearchBoxSuggestions {
+  /**
+   * The position of the list of the suggestions relative to the others.
```
btaillon, line 50: I think we should give some examples, clarifying that other suggestions default to their index in the list, and clarifying that the search box will simply sort suggestions by `position`.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+  /**
+   * Whether the suggestions should be listed in the right or left panel.
+   */
   panel?: 'left' | 'right';
+  /**
+   * Method that returns the list of elements which will be rendered in the list of suggestions.
+   */
   renderItems(): SearchBoxSuggestionElement[];
+  /**
+   * Hook called when the user changes the search box's input value.
```
btaillon, line 62: Similarly to the onSelect comment above, I think we should give an example of what to do here.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
 }
 
+/**
+ * Event sent from the registered query suggestions to the parent search box.
+ */
 export type SearchBoxSuggestionsEvent = (
+  /**
+   * The bindings passed from the search box to the suggestions.
+   */
   bindings: SearchBoxSuggestionsBindings
```
btaillon, line 79: Is this type exported by the `@coveo/atomic` package?

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+/**
+ * The bindings passed from the search box to the suggestions.
+ */
 export interface SearchBoxSuggestionsBindings extends Bindings {
+  /**
+   * The unique id of the search box.
+   */
   id: string;
+  /**
+   * Whether the search box is standalone.
```
btaillon, line 93: I think we should link to some doc about standalone search boxes.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   */
   isStandalone: boolean;
+  /**
+   * The SearchBox headless controller.
```
btaillon, line 97: I think we should include a link to the Headless controller.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   */
   searchBoxController: SearchBox;
+  /**
+   * The amount of queries displayed when the user interacts with the search box.
+   * By default, a mix of query suggestions and recent queries will be shown.
```
btaillon, line 102: I would clarify that this is a limit imposed by the `numberOfQueries` prop on the search box, which will limit the number of suggestions displayed but ignore those that don't have a `query` property.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   */
   numberOfQueries: number;
+  /**
+   * Whether to clear all active query filters when the end user submits a new query from the search box.
+   */
   clearFilters: boolean;
+  /**
+   * Retrieves the suggestion query, meaning the query that will be sent if the search is executed.
```
btaillon, line 110: I think we should describe exactly where this comes from and when it's updated.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+   */
   suggestedQuery(): string;
+  /**
+   * Removes the current suggestions.
+   */
   clearSuggestions(): void;
+  /**
+   * Triggers the retrieval of updated suggestions.
```
btaillon, line 118: Maybe we should document what function/callback this will call, although this may be overkill.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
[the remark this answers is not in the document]
spiracha-coveo:
```suggestion
   * The query associated with the suggestion which will replace the query in the search box if the suggestion is selected.
```

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo:
```suggestion
   * Whether the search box is standalone.
   * https://docs.coveo.com/en/atomic/latest/usage/ssb/
```
Referencing an Atomic article seems to make the most sense to me. Even though this page does not directly address what a standalone search box is, it should. I will add a paragraph to the start of this page.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo:
```suggestion
   * The SearchBox headless controller.
   * https://docs.coveo.com/en/headless/latest/reference/search/controllers/search-box/
```
I agree!

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+/**
+ * List of suggestions that will be displayed along other lists (e.g recent queries) when the search box's input is selected.
+ */
 export interface SearchBoxSuggestions {
+  /**
+   * The position of the list of the suggestions relative to the others.
+   */
   position: number;
+  /**
+   * Whether the suggestions should be listed in the right or left panel.
```
spiracha-coveo, line 54:
```suggestion
   * Whether the suggestions should be listed in the right or left panel. By default, the suggestions are listed in the right panel.
```

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo: The example code I will be using in the usage article utilizes this function. Do you think that should suffice or do we need an example here too?

spiracha-coveo:
```suggestion
   * Hook called when the user changes the search box's input value. This can lead to all the query suggestions being updated.
```

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
```diff
@@ -3,41 +3,136 @@ import {VNode} from '@stencil/core';
...
+  /**
+   * Method that returns the list of elements which will be rendered in the list of suggestions.
+   */
   renderItems(): SearchBoxSuggestionElement[];
+  /**
+   * Hook called when the user changes the search box's input value.
+   */
   onInput?(): Promise<unknown> | void;
+  /**
+   * Hook called when the suggested query changes.
```
spiracha-coveo, line 66:
```suggestion
   * Hook called when the suggested query changes. As a user traverses through the list of suggestions, instant results can be rendered based on the current suggested query.
```

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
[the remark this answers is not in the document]
spiracha-coveo: Makes sense!

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo: Just to clarify, these are the Atomic generated suggestions only?

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo: I think that would be useful!

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo: I am not sure I understand what this means :(

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo: What do you mean by "but ignore those that don't have a `query` property?

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo: The example code I will be using in the usage article utilizes this function. Do you think that should suffice?

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
btaillon: Say you have three elements that each render a bunch of suggestions. Say they are initialized in the following order with the following positions:
* frequentlyAddedToCartSuggestions (position 1337)
* veryIntelligentSuggestions (position 20)
* recentQueriesSuggestions (position 1337)

They will be displayed under the search box in the following order:
* veryIntelligentSuggestions
* frequentlyAddedToCartSuggestions
* recentQueriesSuggestions

By default, when somebody uses our existing suggestion elements (e.g., `atomic-search-box-recent-queries`), the first element will choose `0` as the position for the suggestions it renders, the second will take `1`, the third will take `2`, and so on.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
ThibodeauJF: Yes I think the usage article should be linked in the reference doc, because it won't be enough to get a clear picture of what to do!

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
btaillon: Perhaps the reference doc could have a link to the usage article.

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
ThibodeauJF: Yes

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
ThibodeauJF: All registered suggestions

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
btaillon: A given suggestions "renderer" renders multiple suggestions, and each can have or not have a `query` property independently. A good example is the recent queries suggestions, which render a bunch of suggestions that each have their own `query` property and also renders a header than doesn't have a `query` property. The header doesn't count towards `numberOfQueries`, but the suggestions below it do.

spiracha-coveo: Thanks!

packages/atomic/src/components/search/search-box-suggestions/suggestions-common.ts
spiracha-coveo: Thank you!

spiracha-coveo (approved):
LGTM! Thank you @ThibodeauJF @btaillon

olamothe (approved):

PR #2449, opened 2022-09-16
feat(headless): send the case metadata in the payload of all the insight analytics events

mmitiche:
### Case Metadata(`CaseContext`,`CaseId`,`CaseNumber`) need to be sent in each insight analytics event, example:

SearchBoxSubmitEvent

### What I did in this PR:
- #### I installed the latest version of coveo.analytics, version 2.20.25.
- #### I added the case metadata(caseContext, caseId, CaseNumber) in the payload of every insight analytics call.
- #### I added unit tests to test that the proper payload is sent in every insight analytics call.


packages/headless/src/features/insight-search/insight-search-analytics-actions.ts
```diff
@@ -8,7 +8,12 @@ import {getQueryInitialState} from '../query/query-state';
 export const logFetchMoreResults = makeInsightAnalyticsAction(
   'search/logFetchMoreResults',
   AnalyticsType.Search,
-  (client) => client.logFetchMoreResults()
+  (client, state) =>
+    client.logFetchMoreResults({
+      caseContext: state.insightCaseContext?.caseContext || {},
```
aelfaleh, line 13: May be create an exported  function that state as param and return this object, to be used in all project, may be one day we need to add something , it would be easier to make change
what do you think?

mmitiche: I absolutely agree! Good idea 👌

aelfaleh (approved):
LGTM

olamothe (approved):

PR #2450, opened 2022-09-19
fix(atomic): allow facets to be re-added to the interface

ThibodeauJF:
Solution to problem n. 2 --> https://github.com/coveo/ui-kit/pull/2431

My first solution was to de-register facets from the engine altogether before adding them back. This solution created more issues than it solved on the Atomic side. I went instead with following the warning & adding a facet id + handle reinsertion in the Stencil Store.
https://coveord.atlassian.net/browse/KIT-2033


packages/atomic/src/components/common/interface/store.ts
```diff
@@ -51,11 +51,15 @@ export function createAtomicCommonStore<
       data: StoreData[T][U] & {facetId: U; element: HTMLElement}
     ) {
       if (stencilStore.state[facetType][data.facetId]) {
-        return;
+        stencilStore.state.facetElements =
+          stencilStore.state.facetElements.filter(
+            (facetElement) =>
+              facetElement.getAttribute('facet-id') !== data.facetId
+          );
```
btaillon, line 58: I think we should extract that into a separate function like `removeFacetElements(facetId)`.

btaillon (approved):

PR #2451, opened 2022-09-19
fix(atomic): prevent scrolling behind refine modal on iOS

btaillon:
https://coveord.atlassian.net/browse/KIT-1943

The strategy I went for is to add an event listener which completely disables scrolling and then prevent propagation where we do want scrolling to work.

This should affect all modals, including the feedback modal for smart snippet.

I tested to make sure this doesn't break scrolling on other browsers either. I tested the following:
* iOS Safari
* Android Chrome
* Android Firefox

### DIFF
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -9,6 +9,7 @@ import {
   Event,
   EventEmitter,
   Host,
+  Listen,
 } from '@stencil/core';
 import {isIOS} from '../../../utils/device-utils';
 import {listenOnce} from '../../../utils/event-utils';
@@ -109,6 +110,11 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
     return classes;
   }
 
+  @Listen('touchmove', {passive: false})
+  onWindowTouchMove(e: Event) {
+    this.isOpen && e.preventDefault();
+  }
+
   public componentDidLoad() {
     this.watchToggleOpen(this.isOpen);
   }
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -155,7 +161,18 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
                 part="body-wrapper"
                 class="overflow-auto grow flex flex-col items-center w-full"
               >
-                <div part="body" class="w-full max-w-lg">
+                <div
+                  part="body"
+                  class="w-full max-w-lg"
+                  ref={(element) =>
+                    element &&
+                    element.addEventListener(
```
btaillon, line 169: The listener should be automatically removed when there's no more references to the element.

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -109,6 +110,11 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
     return classes;
   }
 
+  @Listen('touchmove', {passive: false})
```
btaillon, line 113: Only events with `passive: false` can control scrolling.

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
ThibodeauJF, line 169:
```suggestion
                    element?.addEventListener(
```

ThibodeauJF (approved):

PR #2262, opened 2022-07-27
feat(atomic): initial work for recommendations list and interface

olamothe:
https://coveord.atlassian.net/browse/KIT-1861

PR #2454, opened 2022-09-19
feat(atomic): initial work for recommendations list and interface

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1861

Similar to https://github.com/coveo/ui-kit/pull/2262 but adapted tor recent codebase

What it looks like currently:
Screen Shot 2022-09-19 at 4 49 48 PM


packages/atomic/src/components/common/atomic-result/atomic-result.tsx
```diff
@@ -4,22 +4,22 @@ import {applyFocusVisiblePolyfill} from '../../../utils/initialization-utils';
...
+} from '../layout/display-options';
+import {AtomicCommonStore, AtomicCommonStoreData} from '../interface/store';
 
 /**
  * The `atomic-result` component is used internally by the `atomic-result-list` component.
  */
 @Component({
   tag: 'atomic-result',
-  styleUrl: '../../common/result-styles/result.pcss',
+  styleUrl: 'atomic-result.pcss',
```
btaillon, line 22: I'm not sure if the code for the result list is going to stay the same for every implementation, I think it would be safer not to re-use it like this and to instead add an `atomic-recs-result` component just like we did for insight.

ThibodeauJF: Okay I'll move it back to search, but I'll create the new component in an upcoming PR

PR #2461, opened 2022-09-20
fix(headless): fixed didYoumean auto correction originalQuery

btaillon:
https://coveord.atlassian.net/browse/KIT-2042

The new query was present instead of the original query.

### DIFF
--- a/packages/headless/src/features/search/search-actions-thunk-processor.ts
+++ b/packages/headless/src/features/search/search-actions-thunk-processor.ts
@@ -110,6 +110,8 @@ export interface AsyncThunkConfig {
 type QueryCorrectionCallback = (modification: string) => void;
 
 export class AsyncSearchThunkProcessor<RejectionType> {
+  private readonly originalQuery: string;
+
   constructor(
     private config: AsyncThunkConfig,
     private onUpdateQueryForCorrection: QueryCorrectionCallback = (
@@ -117,7 +119,9 @@ export class AsyncSearchThunkProcessor<RejectionType> {
     ) => {
       this.dispatch(updateQuery({q: modification}));
     }
-  ) {}
+  ) {
+    this.originalQuery = this.getOriginalQuery();
+  }
 
   public async fetchFromAPI(
     {mappings, request}: MappedSearchRequest,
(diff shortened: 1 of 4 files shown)


packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -110,14 +110,18 @@ export interface AsyncThunkConfig {
...
   constructor(
     private config: AsyncThunkConfig,
     private onUpdateQueryForCorrection: QueryCorrectionCallback = (
       modification
     ) => {
       this.dispatch(updateQuery({q: modification}));
     }
-  ) {}
+  ) {
+    this.originalQuery = this.getOriginalQuery();
```
olamothe, line 123: Instead of adding a class variable, I would suggest adding a local variable when that value is needed (before doing the state modification with a dispatch)

ie: In `processQueryCorrectionsOrContinue` function, we can keep a local variable for the original query, and then we perform the state modification.

```typescript
  private async processQueryCorrectionsOrContinue(
    fetched: FetchedResponse
  ): Promise<ValidReturnTypeFromProcessingStep<RejectionType> | null> {
    const state = this.getState();
    const successResponse = this.getSuccessResponse(fetched);
    [...]

   // ---> new line <---
    const queryBeforeCorrection = this.getOriginalQuery();

    const {correctedQuery} = successResponse.queryCorrections[0];
    const retried = await this.automaticallyRetryQueryWithCorrection(
      correctedQuery
    );

    [...]


    return {
      ...retried,
      response: {
        ...retried.response.success,
        queryCorrections: successResponse.queryCorrections,
      },
      automaticallyCorrected: true,
      originalQuery: queryBeforeCorrection,
      analyticsAction: logDidYouMeanAutomatic(),
    };
  }
```

Same logic for `processQueryTriggersOrContinue`

In the other places in this class (`processSuccessReponse`) it should not be needed: the query is not modified with a dispatch.

We should also rename `private getOriginalQuery()` to `private getCurrentQuery()`, it's more representative of what it does.

olamothe (approved):

ThibodeauJF (approved):

PR #2463, opened 2022-09-20
test(cypress): upgrade to cypress v10.8.0

olamothe:
I'll use this PR to test the full build + all tests....

Might not be easy to get everything to pass properly after upgrade.

Changes were done with the migration tool provided by Cypress: lots of rename, file move, changes in how configuration files are done.

https://coveord.atlassian.net/browse/KIT-2046


.github/workflows/prbot.yml
```diff
@@ -49,10 +49,10 @@ jobs:
...
         spec:
           [
-            'cypress/integration/*',
-            'cypress/integration/result-list/**/*',
-            'cypress/integration/facets/**/*',
-            'cypress/integration/search-box/**/*',
+            'cypress/e2e/*',
+            'cypress/e2e/result-list/**/*',
+            'cypress/e2e/facets/**/*',
+            'cypress/e2e/search-box/**/*',
```
ThibodeauJF, line 55: It wasn't configurable to keep integration?

packages/atomic/cypress.config.ts
```diff
@@ -0,0 +1,80 @@
+import {defineConfig} from 'cypress';
+
+export default defineConfig({
+  projectId: '5ph2j4',
```
ThibodeauJF, line 4: What is the ID for? I though we didn't use the Cypress platform anymore

.github/workflows/prbot.yml
olamothe: That's what the migration tool recommend, since they now support 2 types of testing: end to end (like we are doing) and component testing (which unfortunately is mostly geared for React and Vue projects)

packages/atomic/cypress.config.ts
olamothe: That's generated by the migration tool also. I must say I'm not certain what it represents.

.github/workflows/prbot.yml
ThibodeauJF: I see!

btaillon (approved):

PR #2467, opened 2022-09-22
fix(atomic): improve styling for the insight panel

lbergeron:
https://coveord.atlassian.net/browse/SVCC-2461

This PR updates the styling of the atomic insight search box to better match the insight panel mock-ups.

* display ellipsis in the searchbox placeholder
* replace the search button by a search icon

searchbox-style


packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
```diff
@@ -77,24 +77,25 @@ export class AtomicInsightSearchBox {
   public render() {
     return (
       <SearchBoxWrapper disabled={this.disableSearch}>
+        <atomic-icon
+          part="submit-icon"
+          icon={SearchIcon}
+          class="w-4 h-4"
+        ></atomic-icon>
```
eledonge-coveo, line 84: I guess no self-closing element in Web Component / Stencil?

eledonge-coveo (approved):

aelfaleh (approved):
LGTM

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
lbergeron: I don't think there are special limitations regarding self-closing elements. Will change that for a self-closing element. Thanks

nathanlb:
Looks good! While we're at it, css on the Tabs gets overridden by the CoveoStyleGuide in the builder and it doesn't display as intended. Do you think adding the right specificity in the atomic styling might do the trick?
Screen Shot 2022-09-29 at 3 19 48 PM

I'm mentioning it here because updating atomic versions in the builder preview is a few extra steps so ideally we avoid doing it more often than necessary. But no obligation to fix it in here.

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.pcss
```diff
@@ -10,6 +10,12 @@
   padding-bottom: 0.675rem;
 }
 
+[part='submit-icon'] {
+  height: 100%;
+  margin: auto 0 auto 1em;
```
olamothe, line 15: I think this can be done with tailwind classes (height/margin).

olamothe (approved):

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
```diff
@@ -77,24 +77,21 @@ export class AtomicInsightSearchBox {
   public render() {
     return (
       <SearchBoxWrapper disabled={this.disableSearch}>
+        <atomic-icon part="submit-icon" icon={SearchIcon} class="w-4 h-4" />
```
olamothe, line 80: For my above comment

```suggestion
        <atomic-icon part="submit-icon" icon={SearchIcon} class="w-4 h-full my-auto mr-0 ml-4" />
```

(I think, please verify :D )

olamothe (approved):

btaillon (approved):

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
lbergeron: Thanks for the suggestion, will try that.

ThibodeauJF (approved):

lbergeron:
> Looks good! While we're at it, css on the Tabs gets overridden by the CoveoStyleGuide in the builder and it doesn't display as intended. Do you think adding the right specificity in the atomic styling might do the trick?
> 
> Screen Shot 2022-09-29 at 3 23 48 PM
> 
> I'm mentioning it here because updating atomic versions in the builder preview is a few extra steps so ideally we avoid doing it more often than necessary. But no obligation to fix it in here.

@nathanlb The conflict happened on the `button` CSS rule. So, changing the rule selector to `button[part="tab"]` could work, but after reading on the philosophy behind Tailwind, specifying the Tailwind classes directly on the button made more sense.

PR #2469, opened 2022-09-22
fix(quantic): logic added to avoid having an empty refine modal

mmitiche:
[SVCC-2454](https://coveord.atlassian.net/browse/SVCC-2454)
#### The issue:
We previously decided that the filters button should be disabled when there is no facets to display in the facets panel, however an issue was found in the following scenario:
- The user is currently in the `all` tab, the filter button is enabled and the filters panel displays the facets.
- The user selects a filter, ex: type: User.
- The user goes to a new tab, ex: the community tab, this tab doesn’t return any results but the filters button is still enabled to allow the user to clear the filter he previously selected.
- The user opens the the filters panel and clears the filter. The filters panel becomes  empty because this tab didn't return any results even without any applied filters.

The issue is that we don’t want the filters panel to empty:

https://user-images.githubusercontent.com/86681870/192801149-377a3ebd-0e27-4831-bcae-24f031370fd8.mov

#### The solution:
After discussion with the UX team, we found that it would be better to display a message explaining the situation when the previously described scenario happens. The message will inform the user that there is no filters available for this specific tab. After closing the filters panel, the button to open it again will be disabled.

https://user-images.githubusercontent.com/86681870/192801186-ff5a6f4e-652c-4ca9-bf59-2b0565d37292.mov


aelfaleh (approved):

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -40,7 +40,8 @@ export default class QuanticRefineToggle extends LightningElement {
   labels = {
     sortAndFilters,
     viewResults,
-    noFiltersAvailableForThisQuery
+    noFiltersAvailableForThisQuery,
```
eledonge-coveo, line 43:
```suggestion
    noFilterAvailableForThisQuery,
```

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -40,7 +40,8 @@ export default class QuanticRefineToggle extends LightningElement {
...
+    noFiltersAvailableForThisQuerInTheCurrentTab,
```
eledonge-coveo, line 44:
```suggestion
    noFilterAvailableForThisQuerInTheCurrentTab,
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineToggle/templateWithFacetsWithoutInputs.html
```diff
@@ -9,6 +9,11 @@
       options={options}
       ontryitnow={handleTryItNow}
     >
+      <c-action-select-tab
```
eledonge-coveo, line 12: Self-closing tag?

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineToggle/templateWithFacets.html
```diff
@@ -9,6 +9,11 @@
       options={options}
       ontryitnow={handleTryItNow}
     >
+      <c-action-select-tab
```
eledonge-coveo, line 12: Self-closing tag?

packages/quantic/force-app/examples/main/lwc/actionSelectTab/actionSelectTab.html
```diff
@@ -0,0 +1,13 @@
+<template>
+  <div class="slds-grid slds-gutters">
+    <div class="slds-col">
+      <lightning-button
```
eledonge-coveo, line 4: Self-closing tag?

eledonge-coveo (approved):
🥇 🚀

packages/quantic/force-app/examples/main/lwc/actionSelectTab/actionSelectTab.html
mmitiche: I think we don’t use self-colosing tags with custom components to respect the HTML standard documented [here](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-non-void-html-element-start-tag-with-trailing-solidus)

packages/quantic/cypress/e2e/refine-toggle/refine-toggle.cypress.ts
```diff
@@ -378,6 +423,49 @@ describe('quantic-refine-toggle', () => {
           });
         });
       });
+
+      describe('when a filter is selected and the search query does not return any results', () => {
```
erocheleau, line 427: Is this the same test twice?
All I see is `addFacets(WithoutInputs?)();`

nathanlb:
I like this solution. It looks to me like the visual distinction between an enabled and disabled filter button is pretty subtle or non-existant. Any way we could be more obvious?

Screen Shot 2022-10-03 at 3 46 18 PM

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -882,4 +882,11 @@
       <protected>false</protected>
       <shortDescription>No filters available for this query</shortDescription>
     </labels>
+    <labels>
+      <fullName>quantic_NoFiltersAvailableForThisQueryInTheCurrentTab</fullName>
```
erocheleau, line 886: Is there a maximum name length? :P

erocheleau:
```suggestion
      <fullName>quantic_NoFilterForCurrentTab</fullName>
```

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -59,9 +59,15 @@ export default class QuanticRefineModalContent extends LightningElement {
   hasActiveFilters = false;
   /** @type {AnyHeadless} */
   headless;
+  /** @type {object} */
+  renderedFacets = {};
+  /** @type {boolean} */
+  areFacetsRendered = false;
```
erocheleau, line 65:
```suggestion
  someFacetsRendered = false;
```
What do you think?

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -128,11 +129,16 @@ export default class QuanticRefineToggle extends LightningElement {
   };
 
   get refineButtonDisabled() {
-    const areFacetsRendered = Object.values(this.renderedFacets).reduce(
+    return (
+      (!this.hasResults && !this.activeFiltersCount) || this.isContentEmpty
```
erocheleau, line 133:
```suggestion
      !(this.hasResults || this.activeFiltersCount) || this.isContentEmpty
```
Do you prefer this?

erocheleau (approved):
Just some suggestions on name changes if you want.

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
nathanlb: Why do you judge this to be a better name?

packages/quantic/cypress/e2e/refine-toggle/refine-toggle.cypress.ts
mmitiche: It's the same test but in two different situations cause the component is supposed to behave in a slightly different way when being displayed with facets with inputs comparing to when being displayed with facets without inputs.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
mmitiche: The sky is the limit x)
joke aside. You absolutely right i'll go with your name 👌

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
mmitiche: yes!

mmitiche:
> I like this solution. It looks to me like the visual distinction between an enabled and disabled filter button is pretty subtle or non-existant. Any way we could be more obvious following slds styles?
> 
> Screen Shot 2022-10-03 at 3 46 18 PM

Totally agree, it's kinda impossible to distinguish between the enabled and disabled state of the refine toggle button in the HIP.
However this problem appears to be in the SFINT side, because the default way to display the Refine Toggle in Quantic is the following:

Enabled state:
Enabled

Disabled state:
Disabed

Only in SFINT and in order to respect the mockups, the refine toggle in the HIP is displayed as follows:
Enabled state:
EnabledHIP

Maybe we'll need to update the way we display the enabled state of the refine toggle in the HIP from the  SFINT side.
I'll check with the UX team to see how we can address this, thank you for pointing this out!

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -882,4 +882,11 @@
...
+      <value>No filters available for this query in the current tab</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>No filters available for this query in the current tab</shortDescription>
```
lbergeron, line 890: You should shorten the short description to fit under 50 characters though.

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -59,9 +59,15 @@ export default class QuanticRefineModalContent extends LightningElement {
...
+
 
   connectedCallback() {
     registerComponentForInit(this, this.engineId);
+    this.addEventListener('renderFacet', this.handleRenderFacetEvent);
```
lbergeron, line 70: It would be nice to remove the event listener in `disconnectedCallback`.

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.html
```diff
@@ -27,6 +27,11 @@
     </div>
     <div class="slds-align_absolute-center" slot="content">
       <div class="refine-modal__content">
+        <template if:true={refineButtonDisabled}>
+          <div class="slds-align_absolute-center slds-text-color_weak refine-modal__empty-message">
+            {labels.noFiltersAvailableForThisQuerInTheCurrentTab}
```
lbergeron: There is a typo in the label key.

```suggestion
            {labels.noFiltersAvailableForThisQueryInTheCurrentTab}
```

nathanlb:
> Maybe we'll need to update the way we display the enabled state of the refine toggle in the HIP from the SFINT side. I'll check with the UX team to see how we can address this, thank you for pointing this out!

Yeah it's nothing critical obviously but could probably be improved, maybe just a noticeably lighter shade of gray on the icon would be enough. Anyway gg

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
lbergeron: Extracting variables makes it easier to read, I think. However, I have trouble finding a good name for the `noResults && noFacetsSelected` part, which is probably why it is hard to read in the first place.

```suggestion
const noResults = !this.hasResults;
const noFacetsSelected = !this.activeFiltersCount;
return (noResults && noFacetsSelected) || this.isContentEmpty;
```

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
mmitiche: absolutely!

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
eledonge-coveo: My bad, I had a discussion with Anny and I learnt that in english we can use 'No' and plural together.

PR #2470, opened 2022-09-23
chore(atomic): refactor result list common

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2050


packages/atomic/src/components/common/result-list/result-list-common-interface.ts
```diff
@@ -0,0 +1,57 @@
...
+  getResultId(result: AnyResult): string;
+  listClasses: string;
+  setNewResultRef(element: HTMLElement, resultIndex: number): void;
+}
+
+export interface ResultListCommonProps {
+  bindings: AnyBindings;
+  host: HTMLElement;
+  loadingFlag: string;
+  getDisplay(): ResultDisplayLayout;
```
olamothe, line 22: ~Why are these all `getX()` function instead of just the value directly ?~

nvm: your explanation is below.

packages/atomic/src/components/search/result-lists/atomic-result-list/atomic-result-list.tsx
```diff
@@ -103,52 +86,38 @@ export class AtomicResultList
...
-  handleInfiniteScroll() {
-    if (
-      this.enableInfiniteScroll &&
-      this.resultListCommon.scrollHasReachedEndOfList
-    ) {
-      this.resultList.fetchMoreResults();
+    if (this.host.hasAttribute('image')) {
+      this.imageSize = this.image;
     }
   }
```
btaillon, line 122: Why are we doing it like that as opposed to having `imageSize` default to `undefined` and use `this.imageSize ?? this.image`?

packages/atomic/src/components/search/result-lists/atomic-result-list/atomic-result-list.tsx
```diff
@@ -158,69 +127,44 @@ export class AtomicResultList
         'Folded results will not render any children for the "atomic-result-list". Please use "atomic-folded-result-list" instead.'
       );
     }
+    this.resultList = buildResultList(this.bindings.engine, {
+      options: {fieldsToInclude: this.mergedFieldsToInclude},
+    });
+    this.resultsPerPage = buildResultsPerPage(this.bindings.engine);
     this.resultListCommon = new ResultListCommon({
+      getNumberOfPlaceholders: () => this.resultsPerPageState.numberOfResults,
```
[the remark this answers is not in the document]
olamothe: What if instead of having a stateful `ResultListCommon` with a `renderList` function, it was rewritten as a functional component ?

Would that even be possible ? 

If so, we'd be able to get rid of these getter functions in favor of the property values directly.

Another thing is that the result list common requires all those properties because a bunch of them are just passed down (plumbing) to other components, and it's opaque.

Would it be possible to achieve something like this (pseudo code) ?

```typescript
//atomic-result-list.tsx:

render() {
return (
   <ResultListCommon 
      placeholder={<ResultsPlaceholder {...this.getPropsForPlaceholderOnly()} />}  
      results={<Results {...this.getPropsForResultsOnly()} />>} 
      {...this.getPropsForResultListCommonOnly())} >
  </ResultListCommon>
}
```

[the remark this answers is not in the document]
ThibodeauJF:
```js
export interface ResultListCommonProps {
  bindings: AnyBindings; // multiple used inside common
  host: HTMLElement; // multiple used inside common
  loadingFlag: string; // used inside common to set and passed down to result to unset
  resultTemplateProvider: ResultTemplateProvider; // extracted all the templating in there, used inside and passed down
  getDisplay(): ResultDisplayLayout; // used for classes & passed down to multiple children
  getDensity(): ResultDisplayDensity; // used for classes & passed down to multiple children
  getImageSize(): ResultDisplayImageSize; // used for classes & passed down to multiple children
  getResultListState(): ResultListCommonState<AnyResult>; // various used inside, passed to multiple children
  getNumberOfPlaceholders(): number; // passed down to children placeholders
  getResultRenderingFunction(): ResultRenderingFunction | null; // passed down to children results (alternative to placeholders
  layoutSelector?: string; // used internally to update breakpoints
  nextNewResultTarget: FocusTargetController; // used internally to handle focusing
}
```

packages/atomic/src/components/common/interface/store.ts
```diff
@@ -1,5 +1,6 @@
 import {createStore} from '@stencil/store';
 import {isInDocument} from '../../../utils/utils';
+import {ResultListInfo} from '../../search/atomic-search-interface/store';
```
btaillon, line 3: `ResultListInfo` is defined in the same file as the Search store, I think it should be moved here.

packages/atomic/src/components/common/interface/result.ts
```diff
@@ -0,0 +1,7 @@
+import {FoldedResult, Result} from '@coveo/headless';
+
+export type AnyResult = Result | FoldedResult;
+
+export function extractFoldedResult(anyResult: AnyResult) {
+  return (anyResult as FoldedResult).result || anyResult;
+}
```
btaillon, line 7: Since we're actually trying to get the result *out* of a folded result, I think this should be named `extractUnfoldedResult`.

packages/atomic/src/components/common/result-list/grid-display-results.tsx
```diff
@@ -1,53 +1,49 @@
...
-export const GridDisplayResults: FunctionalComponent<
-  ResultsProps<AnyBindings>
-> = (props) => {
-  return props.resultListState.results.map((result, index) => {
-    // TODO: support proper engine
+export const GridDisplayResults: FunctionalComponent<ResultListDisplayProps> = (
+  props
+) =>
+  props.getResultListState().results.map((result, index) => {
+    const unFoldedResult = extractFoldedResult(result);
```
btaillon, line 11: Small detail, but since [unfolded](https://www.merriam-webster.com/dictionary/unfolded) is an adjective, I wouldn't capitalize the F here.

```suggestion
    const unfoldedResult = extractFoldedResult(result);
```

btaillon (approved):

packages/atomic/src/components/search/result-lists/atomic-folded-result-list/atomic-folded-result-list.tsx
```diff
@@ -43,41 +40,28 @@ import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
...
 
-  public resultListCommon!: ResultListCommon<Bindings>;
-  private renderingFunction: ResultRenderingFunction | null = null;
-  private loadingFlag = randomID('firstResultLoaded-');
-
-  /**
-   * Whether to automatically retrieve an additional page of results and append it to the
-   * current results when the user scrolls down to the bottom of element
-   */
-  private enableInfiniteScroll = false;
```
ThibodeauJF, line 79: Dead code which we never ended up implementing

PR #2474, opened 2022-09-29
feat(quantic): labels added to numeric facet inputs

mmitiche:
#### Labels added to the min/max input of the numeric facet.

Following Deque's recommendation, labels must always be visible. More details can be found here: [SVCC-2304](https://coveord.atlassian.net/browse/SVCC-2304)

#### Before:
BeforeNumericFacet

#### After: 
AfterNumericFacet


erocheleau (approved):
As simple as that huh? 

Really nice clean fix :)

aelfaleh (approved):
LGTM

PR #2475, opened 2022-09-29
feat(atomic): add calendar time support + result-timespan component

olamothe:
* Add calendar time prop for `atomic-result-date`. Uses https://day.js.org/docs/en/plugin/calendar
* Add `atomic-result-timespan` component which can be used to display duration in a human readable format. Uses https://day.js.org/docs/en/durations/durations

The general idea is to bridge gap for some result template helpers that were available for JSUI, but still missing in Atomic.

https://coveord.atlassian.net/browse/KIT-1940


packages/atomic/src/components/search/result-template-components/atomic-result-date/atomic-result-date.tsx
```diff
@@ -36,6 +41,13 @@ export class AtomicResultDate implements InitializableComponent {
    */
   @Prop({reflect: true}) format = 'D/M/YYYY';
 
+  /**
+   * Specify that date should display as calendar time: https://day.js.org/docs/en/plugin/calendar.
+   *
+   * To modify the calendar time string, use the localization feature: https://docs.coveo.com/en/atomic/latest/usage/atomic-localization/
+   */
+  @Prop({reflect: true}) calendarTime?: boolean;
```
btaillon, line 49: Since this is a boolean, I would make the property name start with a verb, like `useCalendarTime`.

btaillon: Also, instinctively I expect "calendar time" to mean a format like "2020-01-02". Perhaps something like "relative time"/`useRelativeTime` would be more intuitive.

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,71 @@
...
+  @Prop({reflect: true}) field!: string;
+  /**
+   * The unit of the field value.
+   * Available units: https://day.js.org/docs/en/durations/creating
+   */
+  @Prop({reflect: true}) unit = 'ms';
+  /**
+   * Available formats: https://day.js.org/docs/en/durations/format
+   */
+  @Prop({reflect: true}) format = 'HH:mm:ss';
```
btaillon, line 40: I think it would be more useful if we could default to some kind of automatic mode, similar to what you'd find above GitHub comments. As in, it would switch between "seconds", "minutes", "hours", "days", "months", etc based on what's the lowest denominator.

packages/atomic/src/components/search/result-template-components/atomic-result-date/atomic-result-date.tsx
olamothe: Calendar time is the verbiage they use both for dayjs (https://day.js.org/docs/en/plugin/calendar) and momentjs (https://momentjs.com/docs/#/displaying/calendar-time/)

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
olamothe: I'll see what I can come up with

ThibodeauJF (approved):

packages/atomic/src/components/search/result-template-components/atomic-result-date/atomic-result-date.tsx
```diff
@@ -36,6 +41,13 @@ export class AtomicResultDate implements InitializableComponent {
    */
   @Prop({reflect: true}) format = 'D/M/YYYY';
 
+  /**
+   * Specify that date should display as calendar time: https://day.js.org/docs/en/plugin/calendar.
```
jpmarceau, line 45:
```suggestion
   * Whether the date should display in the [calendar time format](https://day.js.org/docs/en/plugin/calendar). 
```

packages/atomic/src/components/search/result-template-components/atomic-result-date/atomic-result-date.tsx
```diff
@@ -36,6 +41,13 @@ export class AtomicResultDate implements InitializableComponent {
...
+   *
+   * To modify the calendar time string, use the localization feature: https://docs.coveo.com/en/atomic/latest/usage/atomic-localization/
```
jpmarceau, line 47:
```suggestion
   * To modify the calendar time string, use the [localization feature](https://docs.coveo.com/en/atomic/latest/usage/atomic-localization/).
```

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+import dayjs from 'dayjs';
+import {DurationUnitType} from 'dayjs/plugin/duration';
+import {InitializeBindings} from '../../../../utils/initialization-utils';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+import {ResultContext} from '../result-template-decorators';
+import duration from 'dayjs/plugin/duration';
+dayjs.extend(duration);
+
+/**
+ * The `atomic-result-timespan` component renders the duration of a number result field.
```
jpmarceau, line 13:
```suggestion
 * The `atomic-result-timespan` component renders a target result number field value as a duration.
```
In the index, does the number have to be the number of milliseconds since unix epoch?

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+export class AtomicResultTimespan {
+  @InitializeBindings()
+  public bindings!: Bindings;
+
+  @ResultContext() private result!: Result;
+
+  @State() public error!: Error;
+
+  /**
+   * The result field which the component should use.
```
jpmarceau, line 27:
```suggestion
   * The target result field.
```

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+   * This will look for the field in the Result object first, and then in the Result.raw object.
```
jpmarceau, line 28:
```suggestion
   * The component first looks for the field in the Result object, and then in the Result.raw object.
```

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+   * It is important to include the necessary field in the ResultList component.
```
jpmarceau, line 29: I'm not sure what you mean here. Would you have an example of how to include the field in the ResultList component?

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+   */
+  @Prop({reflect: true}) field!: string;
+  /**
+   * The unit of the field value.
```
jpmarceau, line 33:
```suggestion
   * The unit of measurement of the field value.
```

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+   * Available units: https://day.js.org/docs/en/durations/creating
+   */
+  @Prop({reflect: true}) unit = 'ms';
+  /**
+   * Specify the format into which the duration should be formatted.
```
jpmarceau, line 38:
```suggestion
   * The format to apply to the result field value.
```

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * If not specified, the component will fallback to an algorithm that approximates the duration if it is very long (years, months or days) or HH:mm:ss if under a day.
```
jpmarceau, line 40:
```suggestion
   * By default, the format is HH:mm:ss when the duration is under a day, and it is an approximation when longer (days, months or years).
```

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * The string displayed when there is an approximation can be modified with localization.
```
jpmarceau, line 42:
```suggestion
   * The string displayed when there is an approximation can be modified with [localization](https://docs.coveo.com/en/atomic/latest/usage/atomic-localization/).
```

packages/atomic/src/components/search/result-template-components/atomic-result-timespan/atomic-result-timespan.tsx
olamothe: I used the same phrasing we have for `atomic-result-date`, `atomic-result-text` etc, regarding their own `field` property.

jpmarceau: Interesting; I see that those descriptions were reviewed over a year ago, and I'm not sure that redoing them would be the highest priority right now (at some point my team should do a full review of the annotations though), but I don't really see what we mean by "include the necessary field". Do you think it means we need to include it in the fields to include?
Not a huge deal, but if we could make an improvement in this PR, it would already be something positive 🙂

olamothe: >  Do you think it means we need to include it in the fields to include?

Yep, that's what it means. We can improve it in this PR, for sure.

jpmarceau: Oh nice thanks!
Yeah I have a suggestion then; in fact the `fieldsToInclude` prop is deprecated in the ResultList.
```suggestion
   * It is important to include the target field in the `fieldsToInclude` list of the parent `search-interface` component.
```

jpmarceau (approved):

btaillon (approved):

packages/atomic/src/components/search/result-template-components/atomic-result-date/atomic-result-date.tsx
louis-bompart: I'm in favor of `relativeTime` because: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

olamothe: Modified to `relativeTime`

PR #2479, opened 2022-10-03
feat(quantic): Quantic Modal made more accessible

mmitiche:
[SVCC-2494](https://coveord.atlassian.net/browse/SVCC-2494)
### Accessibility issues in the Quantic Modal:

- The user can navigate to the content of the Quantic Modal even when the modal is closed.
- The focus is not trapped inside the Quantic Modal.

### Changes made to improve the accessibility of the Quantic modal:

- CSS visibility `hidden` added to the modal when its closed. This value completely removes the component from the accessibility tree.
- Added a Focus Trap mechanism to the Quantic Modal by creating two invisible div elements around the Quantic Modal content. The two div elements will instantly send the focus to the last element and the first element of the modal respectively when they are focused.

### Demo:

https://user-images.githubusercontent.com/86681870/193557287-661499d6-b6cf-4233-9211-f795f073aab9.mov

PS:
The Quantic Modal can set the focus on a custom component using the delegated focus documented [here](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/create_components_focus).


packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -513,4 +512,102 @@ export function AriaLiveRegion(regionName, elem, assertive = false) {
...
+ * Checks whether an element is focusable.
+ * @param {HTMLElement | Element}  element
+ * @returns {boolean}
+ */
+export function isFocusable(element) {
+  // Source: https://stackoverflow.com/a/30753870
+  if (element.getAttribute('tabindex') === '-1') {
+    return false;
+  }
+  if (
```
eledonge-coveo, line 527: Line break between if?

eledonge-coveo (approved):

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -94,4 +103,51 @@ export default class QuanticModal extends LightningElement {
...
+    const contentSlot = this.template.querySelector('slot[name=content]');
+    const footerSlot = this.template.querySelector('slot[name=footer]');
+
+    const allAssignedElements = [
+      // @ts-ignore
+      ...headerSlot.assignedElements(),
+      // @ts-ignore
+      ...contentSlot.assignedElements(),
+      // @ts-ignore
+      ...footerSlot.assignedElements(),
```
erocheleau, line 118: Is there a typedef for those somewhere?

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -94,4 +103,51 @@ export default class QuanticModal extends LightningElement {
...
+    ].map((element) => Array.from([element, ...element.childNodes]));
+    return this.flattenArray(allAssignedElements);
+  }
+
+  focusOnLastElement() {
+    for (const element of this.allAssignedElements.reverse()) {
```
erocheleau, line 124: Is this safe if the allAssignedElements is empty?

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -94,4 +103,51 @@ export default class QuanticModal extends LightningElement {
...
+      const lastFocusableElement = getLastFocusableElement(element);
```
erocheleau, line 125: You're probably missing a type or using the wrong type if `.focus()` is not working no?

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
erocheleau: Can't you do this with a `find` ?

```
const focusableElement = this.allAssignedElements.reverse().find(element => !!getLastFocusableElement(element));
if(focusableElement) {
  focusableElement.focus();
}
```

Not 100% sure this works just out of my head ;)

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -94,4 +103,51 @@ export default class QuanticModal extends LightningElement {
...
+      if (lastFocusableElement) {
+        // @ts-ignore
+        lastFocusableElement.focus();
+        break;
+      }
+    }
+  }
+
+  focusOnFirstElement() {
+    for (const element of this.allAssignedElements) {
```
erocheleau, line 135: Same as above, probably more readable using a `Array.find()` method

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -94,4 +103,51 @@ export default class QuanticModal extends LightningElement {
...
+      }
+    }
+  }
+
+  /**
+   * Flattens an array.
+   * @param {array} array 
+   * @returns array
+   */
+  flattenArray(array){
```
erocheleau, line 150: This exists maybe it does what you are looking for?

[Array.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)

In any case, I don't think a method that flattens an array should live in the `quanticModal` component ;)

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -513,4 +512,102 @@ export function AriaLiveRegion(regionName, elem, assertive = false) {
...
+  }
+}
+
+/**
+ * Returns the last focusable element of for an HTML element.
+ * This function would NOT work with shadow root.
+ * @param {HTMLElement | null} element
+ * @returns {Element | null}
+ */
+export function getLastFocusableElement(element) {
```
erocheleau, line 555: Is it worth making just 1 method here and having a parameter if you want the first or the last?

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
mmitiche: Correct! I just fixed it. getLastFocusableElement was returning Element instead of HTMLElement.

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
mmitiche: I was using the js `flat` method at first, and then found out that it was causing some errors in our unit tests, apparently typescript didn't really like this flat method.
So I had the choice between updating the tsconfig file to make our ts accept the flat method or to just recreate this method my self and avoid the risk of breaking other stuff if I mess up the tsconfig x)

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: sure!

nathanlb: They do seem quite similar

nathanlb (approved):
Still some good comments from others to resolve but i don't have much to add

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: Sure!

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
mmitiche: it would work with a small adjustment:
```
    const focusableElement = this.allAssignedElements
       .reverse()
       .find(element => !!getLastFocusableElement(element));
    if (focusableElement) {
      getLastFocusableElement(focusableElement).focus();
    }
```

packages/quantic/force-app/main/default/lwc/quanticModal/quanticModal.js
```diff
@@ -60,6 +64,11 @@ export default class QuanticModal extends LightningElement {
    */
   @api openModal() {
     this.isVisible = true;
+    // 500ms is the duration time of the animation to open the modal.
+    // eslint-disable-next-line @lwc/lwc/no-async-operation
+    setTimeout(() => {
+      this.focusOnFirstElement();
+    }, 500);
```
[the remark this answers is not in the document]
mmitiche: yeah, actually the focus can be set even before the transition is over.
Basically this animation consists of changing the CSS visibility from `hidden` to `visible` and the CSS opacity from `0` to `1`.
The transition from opacity `0` to `1` will take 500ms while the transition from visibility `hidden` to `visible` will happen instantly because the visibility values are not numeric.
At 0ms the visibility is `hidden`, so we can’t yet set the focus on the modal because a hidden element is not accessible, but when the animation is started this visibility value will change instantly to `visible`  which makes the elements inside the modal accessible and focusable.
What I wanna say is that I could have set the timeout to only 1ms and this will still work properly cause at 1ms the CSS visibility is already set to `visible`.
So yeah, a timeout of 500ms is a full proof that this will execute after the end of the transition that matters for setting the focus.

PR #2483, opened 2022-10-03
feat(atomic): add atomic-result-localized-text component

olamothe:
Component that allows to output localized string more easily.

Can be used that way: 

```javascript
 searchInterface.i18n.addResourceBundle('en', 'translation', {
    foo: 'Hello world {{replace_me_with_field_value}} !',
   foo_plural: 'Hello worlds {{replace_me_with_field_value}}'
 });
``` 
```html
<atomic-result-localized-text key="foo" field-count="the_field_which_will_be_used_to_determine_if_plural_should_be_used" field-value-fieldname="replace_me_with_field_value" ></atomic-result-localized-text>
```

https://coveord.atlassian.net/browse/KIT-2064


ThibodeauJF:
I don't think it's a big deal but the downside here is it's forcing implementer to make a choice:
- use atomic-result-text: gain highlighting but lose the ability to pass a count value
- use atomic-result-localized-text: gain count value, multiple field values, but lose highlighting

Could we add a bit more options to the new component so it can highlight, and perhaps deprecate atomic-result-text & remove it in the next version? Or having those 2 separate component better?

ThibodeauJF (approved):

ThibodeauJF:
👍  If we have an example in our default result template (index.html)

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
```diff
@@ -0,0 +1,81 @@
...
+ *    foo: 'Hello world {{a_field_value}} !',
+ * });
+ * ```
+ *
+ * The component could be configured in such a way to replace `{{a_field_value}}` with a dynamic field value from the result:
+ * ```
+ * <atomic-result-localized-text key="foo" field-value-myfield="a_field_value"></atomic-result-localized-text>
+ * ```
+ *
+ * @MapProp name: fieldValue;attr: field-value;docs: The field and values that define which result items to extract as an i18n parameter. For example, the following configuration extract the value of `filetype` from a result, and assign it to the i18n parameter `the_parameter`: `field-value-filetype="the_parameter"`;type: Record<string, string> ;default: {}
```
jpmarceau, line 27:
```suggestion
 * @MapProp name: fieldValue;attr: field-value;docs: The field and values that define which result items to extract as an i18n parameter. For example, the following configuration extracts the value of the `filetype` field from a result, and assigns it to the i18n parameter `the_parameter`: `field-value-filetype="the_parameter"`;type: Record<string, string> ;default: {}
```

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
```diff
@@ -0,0 +1,81 @@
...
+ * The `atomic-result-localized-text` component renders a target i18n localized string using field values as parameter.
+ *
+ * Given this i18n configuration:
+ * ```
+ * searchInterface.i18n.addResourceBundle('en', 'translation', {
+ *    foo: 'Hello world {{a_field_value}} !',
+ * });
+ * ```
+ *
+ * The component could be configured in such a way to replace `{{a_field_value}}` with a dynamic field value from the result:
```
jpmarceau, line 22:
```suggestion
 * The component could be configured as follows to replace `{{a_field_value}}` with a dynamic field value from the result:
```

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
```diff
@@ -0,0 +1,81 @@
...
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+import {ResultContext} from '../result-template-decorators';
+
+/**
+ * The `atomic-result-localized-text` component renders a target i18n localized string using field values as parameter.
+ *
+ * Given this i18n configuration:
+ * ```
+ * searchInterface.i18n.addResourceBundle('en', 'translation', {
+ *    foo: 'Hello world {{a_field_value}} !',
```
jpmarceau, line 18: Should `a_field_value` be `a_field_name`?
So, for instance, if you want the name of an author to show (e.g., J.K Rowling), then you'd write:
`'Hello world {{author}} !'`, not `'Hello world {{J.K. Rowling}} !'`.
In other words, what users want to specify here is the name of the field (e.g., `author`) rather than a value of the field (`J.K. Rowling`).

olamothe: The example is how it's supposed to work, but probably badly phrased if it's confiusing for you.

The way it works is that you need to define two things:

* What is the placeholder/variable that should be replaced in each i18n string. Here that would be `{{a_field_value}}`. It could have been named anything.

* What is the field that should be used to replace that variable/placeholder. This cannot be named arbitrarily: it has to be a valid field on a result.

So when you configure it like so:

```
<atomic-result-localized-text field-value-author="a_field_value" ></atomic-result-localized-text>
```

It means that the value of the field `author` will be used to replace the placeholder `{{a_field_value}}`. With that explanation, how would you rephrase the documentation/example above.

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  @ResultContext() private result!: Result;
+  /**
+   * The i18n translation key
+   */
+  @Prop() key!: string;
+  /**
+   * The field value to dynamically evaluate.
+   */
+  @MapProp() fieldValue: Record<string, string> = {};
```
btaillon, line 44: When I read "field value", I generally think of just a value, like a string or a number. Maybe this should just be "field"?

```suggestion
  /**
   * The field to dynamically evaluate.
   */
  @MapProp() field: Record<string, string> = {};
```

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
```diff
@@ -0,0 +1,81 @@
...
+          this.result,
+          this.fieldCount!
+        ) as number) ?? 1;
+      return this.bindings.i18n.t(this.key, {...parsedFieldValue, count});
+    }
+
+    return this.bindings.i18n.t(this.key, parsedFieldValue);
+  }
+
+  private parseFieldValue() {
```
btaillon, line 64: I think this should be plural since we parse multiple values.
```suggestion
  private parseFieldValues() {
```

btaillon (approved):

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
```diff
@@ -0,0 +1,81 @@
...
+import {
+  InitializableComponent,
+  InitializeBindings,
+} from '../../../../utils/initialization-utils';
+import {MapProp} from '../../../../utils/props-utils';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+import {ResultContext} from '../result-template-decorators';
+
+/**
+ * The `atomic-result-localized-text` component renders a target i18n localized string using field values as parameter.
```
jpmarceau, line 13:
```suggestion
 * The `atomic-result-localized-text` component renders a target i18n localized string using the values of a target field.
```

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
jpmarceau, line 27: Since `fieldValue` seems to be talking about the value of a field, I think it may be clearer for users if we called this parameter something like: `fieldAndVariable` or `fieldAndTarget`.
```suggestion
 * @MapProp name: fieldAndVariable;attr: field-and-variable;docs: The field from which to extract the target string and the variable used to map it to the target i18n parameter. For example, the following configuration extracts the value of `author` from a result, and assign it to the i18n parameter `name`: `field-value-author="name"`;type: Record<string, string> ;default: {}
```

packages/atomic/src/components/search/result-template-components/atomic-result-localized-text/atomic-result-localized-text.ts
jpmarceau: I agree that by `field value` I also tend to just think of the value of a field. It also took me a while to understand that you use this param to specify both the field and the target i18n parameter. What would you think of changing the param to something like:
`fieldAndVariable`
`fieldAndi18nParam`
`fieldAndTarget`

jpmarceau (approved):
👌

PR #2485, opened 2022-10-04
feat(headless): added actionCause and customData to /v2 analytics object

btaillon:
https://coveord.atlassian.net/browse/KIT-2058

Opened as a draft because some analytics events expect a `searchQueryUid` which comes from the search response, and that isn't fixed yet.

In this PR, I made the following changes:
* Added the `PreparableAnalyticsAction` type, which extends `AsyncThunk` but includes a `prepare` function. `SearchAction`, `InsightAction`, etc. now extend `PreparableAnalyticsAction`.
  * Usable like:
    * `engine.dispatch(myPreparableAnalyticsAction);`
    * 
      ```ts
      const {description, action} = myPreparableAnalyticsAction.prepare({ state, loger, ... });
      ...
      engine.dispatch(action);
      ```
* Changed the signature of **all** `log*` exports to be functions which return specific `PreparableAnalyticsAction` types (e.g., `SearchAction`, `CustomAction`, `InsightAction`).
* Added middleware to make analytics async thunks directly dispatchable.
  * Not necessary, but `engine.dispatch` doesn't show an error when we try to dispatch an asyncThunk, despite it not doing anything. An alternative could be to make `PreparableAnalyticsAction` extend `AsyncThunkAction` instead of `AsyncThunk`, but this would mean that we could no longer access `fulfilled`, `pending`, etc.
  * Enables:
    * `engine.dispatch(logInterfaceLoad());` instead of `engine.dispatch(logInterfaceLoad()());`
    * 
      ```ts
      const {description, action} = myPreparableAnalyticsAction.prepare({ state, loger, ... });
      ...
      engine.dispatch(action); // Instead of engine.dispatch(action());
      ```


ThibodeauJF:
Took a sneak peak, looks good! I like this pattern

PR #2492, opened 2022-10-07
fix(atomic): change facet store label to function that returns i18n translated value

olamothe:
https://coveord.atlassian.net/browse/KIT-2069


ThibodeauJF (approved):
Ah ah!

PR #2494, opened 2022-10-07
feat(atomic,headless): use different InteractiveResult per engine/interface

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2079


btaillon (approved):
Very good!

PR #2495, opened 2022-10-11
feat(quantic): Accessibility of the Quickview and the Quantic Modal improved

mmitiche:
## Issue: 
Previously, we were able to implement a focus trap logic to prevent the user from setting the focus using `tab` and `shift+tab` on elements  that are outside the modals in Quantic (Quantic Quick View and Quantic Refine Modal components). However the screen-readers also offer the ability to navigate through  the elements in the page by clicking the keyboard arrow keys. This is the source of this issue, a user is able to navigate to elements outside the modals in Quantic and the screenreader is in fact reading content that should not be read when the modal is open.

## Solution:
To address this kind of issues the attribute `aria-hidden` is the solution, this attribute needs to be set to `true` on all elements that should no be accessible whenever a modal is open.

### 1. Quantic Quickview:
- When the Quickview modal is opened, the event `quantic__resultpreviewtoggle` is dispatched.
- This event is being listened to by two components: the QuanticA11yHandler and the QuanticResulltList.
- When the QuanticA11yHandler receives the event, it executes a recursive algorithme that sets the `aria-hidden` attribute to true on all the elements inside the search interface except the the QuanticResultList component, because this component is the parent of the QuanticResultQuickview.
- When the QuanticResultList receives the event, it passes the id of the previewed result to each of its QuanticResult components.
- Each QuanticResult will have the following behaviour.
        _ if there is no open Quickview, all the QuanticResult is accessible.
        _ if the Quickview of this QuanticResut is open, `aria-hidden`  is set to true on all the elements of this QuanticResult 
           except the QuanticResulltQuickview component which will remain accessible.
       _  if the Quickview of another QuanticResut is open, `aria-hidden`  is set to true on all the elements of this 
           QuanticResult without exception.

### 1. Quantic RefineModal:
- When the QuanticRefineModal is opened, the event `quantic__refinemodaltoggle` is dispatched by the QuanticRefineToggle.
- This event is being listened to by one component: the QuanticA11yHandler.
- When the QuanticA11yHandler receives the event, it executes a recursive algorithme that sets the `aria-hidden` attribute to true on all the elements inside the search interface except the the QuanticRefineToggle component, because this component is the parent of the QuanticRefineModal.

## Note
This potential solution will make sure that every element in the search interface is not accessible except the open modal whenever a modal is opened. This will make the screen-readers read only what's inside the modal.

### Demo:
In this example, I'm using the arrow keys on the keyboard to try to navigate through the page content after opening the quick view..
Notice that the screen-reader is only reading what's in the modal.

https://user-images.githubusercontent.com/86681870/195598211-698e2fc2-078f-4a8c-9d6c-2876033bdf7d.mov


packages/quantic/force-app/main/default/lwc/quanticResultTemplate/quanticResultTemplate.js
```diff
@@ -19,6 +19,18 @@ import {LightningElement} from 'lwc';
  * </c-quantic-result-template>
  */
 export default class QuanticResultTemplate extends LightningElement {
+  /**
+   * @api
+   * @type {boolean}
+   */
+  @api isAnyPreviewOpen = false;
```
nathanlb, line 26: Have you tested how this is interpreted by the automatic doc generator? I think it takes into account any attribute with the @api tag in the jsdoc. Given this is an internal prop we don't want users to modify, it should not be picked up by the reference doc. I think removing @api on L23 would do it.

packages/quantic/force-app/main/default/lwc/quanticResultList/quanticResultList.js
```diff
@@ -60,15 +60,29 @@ export default class QuanticResultList extends LightningElement {
...
+      this.handleResultPreviewToggle
+    );
   }
 
+  handleResultPreviewToggle = (event) => {
+    if (event.detail.isOpen) {
+      this.openPreviewId = event.detail.resultId;
+    } else {
+      this.openPreviewId = null;
+    }
```
nathanlb, line 83:
```suggestion
    this.openPreviewId = event.detail.isOpen ? event.detail.resultId : null;
```

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -112,6 +118,90 @@ export default class QuanticSearchInterface extends LightningElement {
...
+      const children = Array.from(slot.assignedElements());
+
+      children.forEach((child) => {
+        this.removeElementsFromAccessibilityTree(child, 'C-QUANTIC-RESULT-LIST');
+      });
+    } else {
+      this.nonAccessibleElements.forEach(element => {
+        element.setAttribute('aria-hidden', 'false');
+      })
+      this.nonAccessibleElements = [];
```
nathanlb, line 146: the two functions are nearly identical. Could we refactor?

nathanlb (commented):
Lots of accessibility-related handlers in the quantic-search-interface now... I'm wondering if they could be in another component dedicated to responding to this type of events. The search interface could render it in its .html and place the slot between its tags so the result would be the same but it would clean up the code. Maybe it's just me but it doesn't feel like it fits in the component's job description to handle all of these a11y things. What are your thoughts on that?

mmitiche:
> Lots of accessibility-related handlers in the quantic-search-interface now... I'm wondering if they could be in another component dedicated to responding to this type of events. The search interface could render it in its .html and place the slot between its tags so the result would be the same but it would clean up the code. Maybe it's just me but it doesn't feel like it fits in the component's job description to handle all of these a11y things. What are your thoughts on that?

This makes a lot of sense I totally agree! Additionally, the creation of the component you mentioned will allow us to improve the accessibility in the the QuanticCaseAssistInterface and the QuanticInsightInterface easily and without duplicating any code!

mmitiche:
> Lots of accessibility-related handlers in the quantic-search-interface now... I'm wondering if they could be in another component dedicated to responding to this type of events. The search interface could render it in its .html and place the slot between its tags so the result would be the same but it would clean up the code. Maybe it's just me but it doesn't feel like it fits in the component's job description to handle all of these a11y things. What are your thoughts on that?

The challenge now is to find an appropriate name for this new component x)

nathanlb:
> The challenge now is to find an appropriate name for this new component x)

Haha the toughest part of the job... 🤔 `QuanticA11yHandler`?

packages/quantic/force-app/main/default/lwc/quanticResultTemplate/quanticResultTemplate.js
mmitiche: you are right 👌 the `@api` need to be removed here!

packages/quantic/force-app/main/default/lwc/quanticResult/quanticResult.js
```diff
@@ -71,4 +75,19 @@ export default class QuanticResult extends LightningElement {
     }
     return defaultTemplate;
   }
+
+  get resultPreviewShouldNotBeAccessible() {
+    if (!this.openPreviewId) {
+      return false;
+    }
+    if (this.result.uniqueId !== this.openPreviewId) return true;
+    return false;
```
erocheleau, line 84:
```suggestion
    return this.openPreviewId && (this.result.uniqueId !== this.openPreviewId);
```

Seems like this can be simplified to this, but I don't really understand the condition :)

packages/quantic/force-app/main/default/lwc/quanticResult/quanticResult.js
```diff
@@ -71,4 +75,19 @@ export default class QuanticResult extends LightningElement {
...
+  }
+
+  get isAnyPreviewOpen() {
+    if (!this.openPreviewId) {
+      return false;
+    }
+    return true;
```
erocheleau, line 91:
```suggestion
    return (!!this.openPreviewId); 
```
Not sure...

erocheleau (approved):

packages/quantic/force-app/main/default/lwc/quanticA11yHandler/quanticA11yHandler.js
```diff
@@ -0,0 +1,107 @@
...
+    const childNodes = Array.from(element.childNodes);
+    if (childNodes.length === 0) return false;
+    return childNodes.reduce(
+      (acc, val) => acc || this.isParentOf(val, targetElement),
+      false
+    );
+  }
+
+  /**
+   * Removes all HTML elements from the accessibility tree except for elements whose tag name is the excludedTagName, these elements remain accessible.
```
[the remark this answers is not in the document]
mmitiche: yes 'accessibility tree' is indeed a thing:  https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree#:~:text=The%20accessibility%20tree%20contains%20accessibility,%2C%20attributes%2C%20and%20text%20nodes.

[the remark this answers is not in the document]
mmitiche: In the situation you mentioned, this method will exclude all the elements that have this specific `excludedTagName`, all the elements with that tag name will NOT be removed from the accessibility tree.

packages/quantic/force-app/main/default/lwc/quanticA11yHandler/quanticA11yHandler.js
```diff
@@ -0,0 +1,107 @@
...
+      return false;
+    }
+    /** @type {Array} */
+    const childNodes = Array.from(element.childNodes);
+    if (childNodes.length === 0) return false;
+    return childNodes.reduce(
+      (acc, val) => acc || this.isParentOf(val, targetElement),
+      false
+    );
+  }
```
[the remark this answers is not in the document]
mmitiche: sure yeah 👌

packages/quantic/force-app/main/default/lwc/quanticResult/quanticResult.js
mmitiche: sure yeah! I'll simplify this.

mmitiche: I will try to explain the logic behind this as well:
Here we are dealing with the Quantic Result component, its template looks like this:
```
<c-quantic-result>
   <div>Result title</div>
   <div>Result excerpt</div>
   <c-quantic-result-quickview></c-quantic-quickview>
</c-quantic-result>
```

Another useful to know about the QuanticResult component is that it has a property named `openPreviewId`, this property is set by the QuanticResultList on each QuanticResult.
this property behaves as follows:
- if there is NOT any quickview opened: `openPreviewId===null`
- if there a quickview opened: `openPreviewId==='RESULT_ID_OF_THE_PREVIEWD_RESULT'`


1. First, lets focus on the title and the the excerpt, these elements need to be removed from the accessibility when the quickview is open. To achieve this I created the property `isAnyPreviewOpen` which is basically :

```
get isAnyPreviewOpen(){
   return !!this.openPreviwId
}
```

2.  Secondly, we need to handle the accessibility of the `c-quantic-result-quickview`, this element:
- should be accessible when there is not any opened quickview (`openPreviewId===null`)
- should NOT be accessible when a quickview of another result os opened(`openPreviewId==='RESULT_ID_OF_ANOTHER_RESULT'`)
- should be accessible when the quickview of this particular result is opened(`openPreviewId==='RESULT_ID_OF_THIS_PARTICULAR_RESULT'`)

To translate these three points I we have this property `resultPreviewShouldNotBeAccessible` which is basically:
```
get resultPreviewShouldNotBeAccessible(){
   return this.openPreviwId && (this.result.uniqueId !== this.openPreviewId)
}
```

nathanlb (approved):
Very nice!

packages/quantic/force-app/main/default/lwc/quanticA11yHandler/quanticA11yHandler.js
```diff
@@ -0,0 +1,86 @@
...
+import {isParentOf} from 'c/quanticUtils';
+
+/**
+ * The `QuanticA11yHandler` component handles the accessibility of the Quantic interfaces. It makes the necessary changes to the accessibility tree each time a QuanticResultQuickview or a QuanticRefineModal is opened.
+ * @category Search
+ * @category Insight Panel
+ * @example
+ * <c-quantic-a11y-handler></c-quantic-a11y-handler>
+ */
+export default class QuanticA11yHandler extends LightningElement {
```
btaillon, line 11: This is very similar to `atomic-focus-trap`, and I think the name "A11yHandler" is a bit too generic, perhaps this could also be named "focus trap".

nathanlb: `QuanticFocusAndNavigationEtcTrapHandler`

mmitiche: What about this more specific name:  `QuanticModalA11yHandler`?

nathanlb: I assumed we might have more needs to handle a11y-related events in the future and the fact that this component was generic was a good thing. If we scope it to modals only, and later we need to add another handler that isn't modal-related then we'll need to start wrapping a11y handlers within a11y handlers in a certain order and i can see it getting messy. Am I inventing problems? What do you think?

mmitiche: Well I don't think that every a11y issue will require a solution based on sending and handling events, so far and amongst all the issues highlighted by deque, the modal problem is the only one requiring this approach to be solved. and generally speaking modals are more or less the main accessibility thing that affects all the other elements in the page which is the reason we used this approach of creating a whole new component that will add aria-hidden to multiple elements in the page to make the modals accessible.
All I'm saying is that I don't think we will need to use this approach to solve a11y issues that are not related to modals.

nathanlb: You've been a lot more hands-on than myself for these accessibility fixes so I trust you on that. If `QuanticModalA11yHandler` is what you prefer then go for it 👍

PR #2497, opened 2022-10-11
feat(quantic): Contrast of several elements have been fixed to respect a11y standards

erocheleau:
Before:

contrast problems smaller

After:

image

All these elements have been updated to respect standards of contrast.

**Important Note**

This also introduces usage of overrides of LWC Styling Hooks. Which means by default this code can no longer be run into the LWC Local Development Server without some manipulation first.

These have been documented in the changes in the Readme as well as a comment in the file itself.

(TL;DR you need to remove the styling hooks to run the code in the Local Development Server)


mmitiche (approved):
Looking good👌

packages/quantic/force-app/main/default/lwc/exampleSearch/exampleSearch.css
```diff
@@ -1,3 +1,16 @@
+/* 
+ * Styling hooks are not supported by the LWC Local Development Server.
+ * You can safely comment those out in order to run the code in the Local Dev Server, these are for a11y purpose only.
+ * Issue opened to Salesforce: https://github.com/forcedotcom/salesforcedx-vscode/issues/4493
+ */
+:host {
```
nathanlb, line 6: Given these are in the exampleSearch, a user would need to chose to include them in their page. Maybe we should add a comment saying this is to increase contract for accessibility purposes?

packages/quantic/force-app/main/default/lwc/quanticRecentQueriesList/quanticRecentQueriesList.css
```diff
@@ -25,5 +25,5 @@ a {
 }
 
 .empty-list-message {
-  color: #8e959d;
+  color: var(--lwc-colorTextIconDefault, #747474);
```
nathanlb, line 28: Given we're giving up on ootb localdev support is there a way to define css variables for all of these color values instead of using the hex code every time? Would help avoid mismatches.

nathanlb (approved):
Noice

packages/quantic/force-app/main/default/lwc/exampleSearch/exampleSearch.css
erocheleau: Added a comment, great idea!

packages/quantic/force-app/main/default/lwc/quanticRecentQueriesList/quanticRecentQueriesList.css
erocheleau: I think if we start wanting to do this, we should do this everywhere :D not just for the changes I'm making.
We have colors everywhere, we would need to find the right way to include CSS vars across all components.

PR #2498, opened 2022-10-11
feat(quantic): aria-controls correctly set in the QuanticSearchBox

mmitiche:
### The issue
[SVCC-2283](https://coveord.atlassian.net/browse/SVCC-2283)

A combobox element has aria-controls set to a value that refers to the id of the element that serves as the popup. The problem we have in LWC is that the id values may be transformed when the components are rendered.

That's why I'm proposing this solution based on the section 'Link IDs and ARIA Attributes from Different Templates' of the following documentation : https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_components_accessibility_attributes

### The solution:
- When the QuanticSearchBoxSuggestionsList is rendered, an event is sent to the QuanticSearchBox component containing on its payload the transformed id.
- This transformed id is then set as the `aria-controls` attribute of the search box input.

### Demo:

AriaControls


nathanlb:
I think the standalone searchbox will need the same modifications that are in the searchbox

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -68,15 +68,27 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
 
   /** @type {number} */
   @track selectionIndex = -1;
+  /** @type {boolean} */
+  initialRender = true;
```
erocheleau, line 72: It's a detail but typically the default value of a boolean is false, so to do only 1 render, usually the pattern I've seen is just you don't declare a `hasRendered` variable and then in the renderedCallback

```
if(!this.hasRendered) {
  [...]
  this.hasRendered = true;
}
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -106,4 +118,10 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
   preventDefault(event) {
     event.preventDefault();
   }
+
+  get listboxCssClass() {
+    return `slds-dropdown slds-dropdown_length-5 slds-dropdown_fluid ${
+      this.suggestions?.length ? '' : ' slds-hidden'
```
erocheleau, line 124:
```suggestion
      this.suggestions?.length ? '' : 'slds-hidden'
```
You already have a space

nathanlb (approved):
Nice and readable.
FYI unrelated but it appears we might now be able to extend LWC so we could look at refactoring the SSB to extend the base quantic search box and reduce duplication like we have now.

PR #2504, opened 2022-10-14
chore(atomic): add recslist to storybook

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2077

### DIFF
--- a/packages/atomic/.storybook/recommendation/recs-init.ts
+++ b/packages/atomic/.storybook/recommendation/recs-init.ts
@@ -1,0 +1,44 @@
+import {
+  getSampleRecommendationEngineConfiguration,
+  RecommendationEngineConfiguration,
+} from '@coveo/headless/recommendation';
+import {debounce} from 'lodash';
+import {addons} from '@storybook/addons';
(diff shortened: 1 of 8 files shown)


packages/atomic/.storybook/recommendation/recs-init.ts
```diff
@@ -0,0 +1,44 @@
...
+import CoreEvents from '@storybook/core-events';
+
+interface SearchInterface extends HTMLElement {
+  initialize: (cfg: RecommendationEngineConfiguration) => Promise<void>;
+  getRecommendations: () => Promise<void>;
+}
+
+const orgIdentifier = getSampleRecommendationEngineConfiguration();
+
+export const initializeInterfaceDebounced = (
```
olamothe, line 17: Ooof I remember that hack with the debounce for the search interface...

It worked well so far, so 🤞 ...

One thing though is that we could share all the code between this function and the one in `default-init.ts`. It's all the same, except for document.querySelector.

Or just have one function with a parameter for the query selector.

olamothe (approved):

PR #2509, opened 2022-10-18
test(atomic): fix flaky search box test; remove category facet focus test

olamothe:
https://coveord.atlassian.net/browse/KIT-2095

### DIFF
--- a/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
+++ b/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
@@ -47,26 +47,35 @@ describe('Search Box Test Suites', () => {
     const maxSuggestionsWithoutQuery = numOfSuggestions - 1;
     const maxRecentQueriesWithoutQuery = numOfRecentQueries - 1;
 
+    function setupSearchbox() {
+      return addSearchBox({
+        suggestions: {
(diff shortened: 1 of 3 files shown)


packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
```diff
@@ -47,26 +47,35 @@ describe('Search Box Test Suites', () => {
     const maxSuggestionsWithoutQuery = numOfSuggestions - 1;
     const maxRecentQueriesWithoutQuery = numOfRecentQueries - 1;
 
+    function setupSearchbox() {
```
olamothe, line 50: The problem with this test:

It mocked query suggestion from the API (ie: no matter what's in the search box, it always return a list of 5 ML query suggestion).

Why it failed sometimes:

1- Type {Rec} very quickly
2- Mocked query suggestion API alway returns 5 values (after variable X milliseconds, depending on the API). The search box query suggestion code does not "filter" what's returned by the API (it expect it to return proper values that match the search box content). So the component just displays the 5 values as is, even if they don't match.
3- Mocked recent query always returns 5 values. Those are filtered by Headless/the component only for those that match.
4- Select the first suggestion shown. If at step 2 if resolves quickly it will be a random value coming from the mocked response. If it resolves slowly it will be a mocked recent query returned at step 3. --> Flakiness.

PR #2510, opened 2022-10-18
chore(quantic): fix E2E testing by turning locker off

nathanlb:
### Changes made to the test community:
- Locker turned off to allow the E2E to correctly run in Cypress.
- The experience bundle config option allowing locker to be turned off is new since API v55 so I bumped to v56
  - v56 communities seem to require page and view definitions for `tooManyRequests` and `serviceNotAvailable`

### Changes made to the E2E tests:
1. Accessibility tests skipped:
We previously had few accessibility tests for the `QuanticFacet` and the `QuanticNumericFacet` components. The tests were not testing much and were super flaky, mainly because we were kinda trying to square a circle.
To test A11y with Cypress properly we will need to use some plugins as documented [here](https://www.deque.com/blog/how-to-test-for-accessibility-with-cypress/).
Adequate tests for A11y will come soon: [SVCC-2378](https://coveord.atlassian.net/browse/SVCC-2378)
2. For some mysterious reasons, the `label` property of the `QuanticTimeframe` component wasn't working when the locker was ON.  When turning the locker OFF the property is working correctly soI updated the tests accordingly and I'm currently investigating the reason why this wasn't working as expected when the locker was ON.
3. A change of behaviour was noticed in the `min` and `max` inputs in the `QuanticNumericFacets`, the the two inputs are using the `lightning-input `component.
Previously, when the `lightning-input` was required but its values were empty, an `aria-invalid` attribute was set to true automatically in that component and we were using this attribute to select the inputs in our tests.
Now, `aria-invalid` attribute is not added anymore in the lightning-input, this change broke our tests for the `QuanticNumericFacet`, thus I updated our tests to avoid basing our selection on that attribute anymore.

### DIFF
--- a/packages/quantic/cypress/e2e/facets-2/numeric-facet/numeric-facet.cypress.ts
+++ b/packages/quantic/cypress/e2e/facets-2/numeric-facet/numeric-facet.cypress.ts
@@ -429,4 +429,17 @@ describe('quantic-numeric-facet', () => {
       });
     });
   });
+
+  describe.skip('with default numeric facet testing accessibility', () => {
+    beforeEach(() => {
+      visitNumericFacetPage(defaultSettings);
+    });
+
+    it('should be accessible through keyboard', () => {
+      Actions.selectFirstNumericFacetValueWithKeyboardTab();
+      Expect.numberOfSelectedCheckboxValues(1);
+      Actions.selectFirstNumericFacetValueWithKeyboardEnter();
+      Expect.numberOfSelectedCheckboxValues(0);
+    });
+  });
 });
--- a/packages/quantic/cypress/e2e/facets-2/numeric-facet/numeric-facet.cypress.ts
+++ b/packages/quantic/cypress/e2e/facets-2/numeric-facet/numeric-facet.cypress.ts
@@ -193,13 +193,13 @@ describe('quantic-numeric-facet', () => {
             cy.wait(InterceptAliases.UA.Facet.ClearAll);
 
             Actions.submitManualRange();
-            Expect.displayInputWarning(1);
+            Expect.displayInputMinWarning(true);
             Expect.inputWarningContains();
 
             Actions.inputMinValue(100);
             Actions.submitManualRange();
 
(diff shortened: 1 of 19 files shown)


packages/quantic/cypress/e2e/facets-2/numeric-facet/numeric-facet.cypress.ts
```diff
@@ -429,17 +429,4 @@ describe('quantic-numeric-facet', () => {
       });
     });
   });
-
-  describe('with default numeric facet testing accessibility', () => {
```
nathanlb, line 433: Will we fix these later on? If so it's probably better to .skip them no?

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineModalContent/exampleQuanticRefineModalContent.html
```diff
@@ -70,7 +70,6 @@
             <c-quantic-timeframe
               amount="10"
               unit="year"
-              label="Last decade"
```
nathanlb, line 73: looks like labels aren't being replicated properly in the refine modal. But i'm assuming the tests once passed so is it a regression from a recent change?

packages/quantic/cypress/e2e/facets-2/numeric-facet/numeric-facet.cypress.ts
mmitiche: Yes, skipping them is a better idea, let me fix that👌

lvu285:
I just want to make sure, if locker is ON, nothing is broken except what you mentioned above QuanticTimeframe label and Min&Max input?, and only test is flaky

mmitiche:
> I just want to make sure, if locker is ON, nothing is broken except what you mentioned above QuanticTimeframe label and Min&Max input?, and only test is flaky

yes exactly👌all the other tests are fine.

drosenfeld-coveo (approved):

aelfaleh (approved):

PR #2512, opened 2022-10-20
fix(headless): instant results should disable abort warning

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2084

was removed as part of a [refactor](https://github.com/coveo/ui-kit/commit/8433ad51c4836a5a23630430d72e752eee3dacf7#diff-693a41de1ad1f40792efcdd3272198d1336232eab9da456f2a35bc893cee6038L413)


olamothe (approved):

packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -121,11 +121,12 @@ export class AsyncSearchThunkProcessor<RejectionType> {
 
   public async fetchFromAPI(
     {mappings, request}: MappedSearchRequest,
-    origin: SearchOrigin
+    origin: SearchOrigin,
+    disableAbortWarning?: boolean
```
btaillon, line 125: I'm worried that we may end up having even more optional paramters in the future. Do you think it would make sense to extract `origin` and `disableAbortWarning` into an object?

btaillon (approved):

PR #2515, opened 2022-10-25
fix(quantic):  AriaLiveRegion created before calling its methods in the Quantic summary

mmitiche:
### Bug:
The `AriaLiveRegion` was created after the execution of the the `updateState` method in the Quantic components.
The `updateState` method contains direct calls to the `AriaLiveRegion`'s methods, which  was throwing an error if the AriaLiveRegion wasn't yet created.

### Solution: 
Create the `AriaLiveRegion` before the execution  of the `updateState` method.

### Example of the QuanticSummary:
Before the fix:
example

Basically, if the QuanticSummary is initialized when the interface already has results, we would end up calling `dispatchMessage` of `undefined` throwing an error. This happens in the Coveo Quantic Insight Panel  in C4SF.


lbergeron (approved):
Do you think a test could be added to prevent this bug from coming back in the future?

mmitiche:
> Do you think a test could be added to prevent this bug from coming back in the future?

good point! I don't yet have a clear idea on how such a test can be done but I'll give it a try 👌

nathanlb (approved):

PR #2519, opened 2022-10-28
feat(headless): updated insight interface with facet type

nathanlb:
The insight interface fetch call will now return facet type info if the `detailed=true` query param is set. We want this info returned so the insight panel can render the right type of facet.


ThibodeauJF (approved):

btaillon:
Updated the title to contain the project name

btaillon (approved):

PR #2520, opened 2022-10-28
chore(atomic): removed deprecations

btaillon:
https://coveord.atlassian.net/browse/KIT-2049

# Deprecations (extracted from the Confluence post)
## Renamed
* Renamed `atomic-result-list` and `atomic-result`'s `image` attribute to `image-size`.
* Renamed CSS variables:
  * `--atomic-rating-facet-icon-active-color` to `--atomic-rating-icon-active-color`
  * `--atomic-rating-facet-icon-inactive-color` to `--atomic-rating-icon-inactive-color`
* Renamed `atomic-refine-modal`'s `select-icon` part to `select-icon-wrapper`
* Renamed `atomic-refine-modal`'s `select-icon-size` part to `select-icon`

## Changed
* Removed `atomic-query-summary`'s `enable-duration` attribute.
  * The duration part can be used to style or hide the duration instead.
* Removed `atomic-result`'s engine attribute.
  * This attribute is unused.
* Some attributes were changed to instead take JSON or a JSX string array as a parameter.
  * The following attributes were affected:
    * `atomic-facet`'s `allowed-values`
    * `atomic-category-facet`'s `base-path`
    * `atomic-search-interface`'s `fields-to-include`
    * `atomic-insight-interface`'s `fields-to-include`
    * `atomic-recs-interface`'s `fields-to-include`
  * For instance:
    ```html
    <atomic-facet allowed-values="in progress,completed" />
    ```
    would become
    ```html
    <atomic-facet allowed-values='["in progress", "completed"]' />
    ```
  * With JSX, an array of strings can now be passed directly like so:
    ```jsx
    <AtomicFacet allowed-values={["in progress", "completed"]} />
    ```
  * This change makes it unambiguous what separator is used to divide parts of the path, and allows commas to be escaped.
* Moved `atomic-result-list` and `atomic-folded-result-list`'s `fields-to-include` attribute to `atomic-search-interface`.
* Removed `atomic-result-link`'s `target` attribute.
  * Instead, you can use the attributes slot like so:
    ```html
    <atomic-result-link>
      <a slot="attributes" target="_blank" download></a>
    </atomic-result-link>
    ```
* In atomic-react, removed `AtomicSearchInterface`'s theme attribute.
* Changed `atomic-insight-result`'s `interactive-result` attribute to be required.

### Insight panel
* Removed `atomic-insight-result`'s `engine` attribute.
  * This attribute is unused.
* Changed `atomic-insight-result`'s `interactive-result` attribute to be required.

### Recommendations
* In atomic-react, removed `AtomicSearchInterface`'s `theme` attribute.

### DIFF
--- a/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
+++ b/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
@@ -4,7 +4,7 @@ import {AnyBindings} from '../interface/bindings';
 
 interface QuerySummaryCommonProps {
   bindings: AnyBindings;
-  enableDuration: boolean;
+  withDuration: boolean;
   querySummaryState: {
     hasDuration: boolean;
     durationInSeconds: number;
@@ -88,23 +88,19 @@ export const QuerySummaryCommon: FunctionalComponent<
             params={paramsWithHighlights}
             count={params.count}
           />
-          <span
-            part="duration"
-            class={
-              props.enableDuration && props.querySummaryState.hasDuration
-                ? ''
-                : 'hidden'
-            }
-          >
-            <LocalizedString
-              key="in-seconds"
-              bindings={props.bindings}
-              params={{
-                count:
-                  props.querySummaryState.durationInSeconds.toLocaleString(),
-              }}
-            />
-          </span>
+          {props.withDuration && props.querySummaryState.hasDuration && (
+            <span part="duration">
+              &nbsp;
+              <LocalizedString
+                key="in-seconds"
+                bindings={props.bindings}
+                params={{
+                  count:
+                    props.querySummaryState.durationInSeconds.toLocaleString(),
+                }}
+              />
+            </span>
+          )}
         </Fragment>
       )}
     </div>
--- a/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
+++ b/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
@@ -4,7 +4,6 @@ import {AnyBindings} from '../interface/bindings';
 
 interface QuerySummaryCommonProps {
   bindings: AnyBindings;
-  withDuration: boolean;
   querySummaryState: {
     hasDuration: boolean;
     durationInSeconds: number;
@@ -88,7 +87,7 @@ export const QuerySummaryCommon: FunctionalComponent<
             params={paramsWithHighlights}
             count={params.count}
           />
-          {props.withDuration && props.querySummaryState.hasDuration && (
+          {props.querySummaryState.hasDuration && (
             <span class="hidden" part="duration">
               &nbsp;
               <LocalizedString
--- a/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
+++ b/packages/atomic/src/components/common/query-summary/query-summary-common.tsx
@@ -89,7 +89,7 @@ export const QuerySummaryCommon: FunctionalComponent<
             count={params.count}
           />
           {props.withDuration && props.querySummaryState.hasDuration && (
-            <span part="duration">
+            <span class="hidden" part="duration">
               &nbsp;
               <LocalizedString
                 key="in-seconds"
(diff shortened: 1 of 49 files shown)


packages/atomic/src/utils/props-utils.ts
```diff
@@ -38,31 +32,16 @@ export function MapProp(opts?: MapPropOptions) {
...
-          other
-        )
-      : console.error(msg, other);
-
+export function ArrayProp() {
   return (component: ComponentInterface, variableName: string) => {
     const {componentWillLoad} = component;
     const attributeWithBackets = camelToKebab(variableName);
 
-    if (!componentWillLoad) {
```
btaillon, line 54: Why was `componentWillLoad` required?

ThibodeauJF: i think its something that Stencil fixed at some point

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
```diff
@@ -4,7 +4,7 @@ import {AnyBindings} from '../interface/bindings';
 
 interface QuerySummaryCommonProps {
   bindings: AnyBindings;
-  enableDuration: boolean;
+  withDuration: boolean;
```
ThibodeauJF, line 7: This should be always true too and removed?

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
```diff
@@ -88,23 +88,19 @@ export const QuerySummaryCommon: FunctionalComponent<
...
-              key="in-seconds"
-              bindings={props.bindings}
-              params={{
-                count:
-                  props.querySummaryState.durationInSeconds.toLocaleString(),
-              }}
-            />
-          </span>
+          {props.withDuration && props.querySummaryState.hasDuration && (
+            <span part="duration">
```
ThibodeauJF, line 92:
```suggestion
            <span class="hidden" part="duration">
```

packages/samples/atomic-next/pages/search-page.tsx
```diff
@@ -58,7 +58,17 @@ const SearchPage: NextPage = () => {
...
+      fieldsToInclude={[
+        'ec_price',
+        'ec_rating',
+        'ec_images',
+        'ec_brand',
+        'cat_platform',
+        'cat_condition',
+        'cat_categories',
+        'cat_review_count',
+        'cat_color',
```
ThibodeauJF, line 70: Some of those are already in the default commerce fields, I believe

packages/atomic/src/components/search/result-link/result-link.tsx
```diff
@@ -10,7 +10,7 @@ export interface ResultLinkEventProps {
 
 export interface ResultLinkProps extends ResultLinkEventProps {
   href: string;
-  target: string;
+  target?: string;
```
ThibodeauJF, line 13: Should this be removed too?

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
btaillon: It we want it to be possible to show the duration by using the `duration` part and `display: block`, then we need the duration to be rendered.

In `atomic-insight-query-summary`, we never display the duration, and it's not possible to show it with a part. Since this component is common across both, we need this option.

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
btaillon: I'm unsure whether hiding the duration using a style should be something common shared across all components who may use QuerySummaryCommon, or if it's something that's only specific to `atomic-query-summary` 🤔 

Currently, this resides in `atomic-query-summary.pcss`.

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
ThibodeauJF: Yes it should be always true and always possible to show using the part, I don't get why with insight's it isn't because the API doesn't return it?

packages/atomic/src/components/common/query-summary/query-summary-common.tsx
ThibodeauJF: I think it's safe to say all query summary should have this hidden by default, if you think of Johnny's design as source of truth

olamothe (approved):

PR #2525, opened 2022-10-31
feat(atomic): replaced printable uri target attribute with attributes slot

btaillon:
https://coveord.atlassian.net/browse/KIT-2106

# Deprecations (extracted from the Confluence post)

## Changed
* Removed `atomic-result-printable-uri`'s `target` attribute.
  * Instead, you can use the attributes slot like so:
    ```html
    <atomic-result-printable-uri>
      <a slot="attributes" target="_blank" download></a>
    </atomic-result-printable-uri>
    ```

### DIFF
--- a/packages/atomic/src/components/search/result-link/result-link.tsx
+++ b/packages/atomic/src/components/search/result-link/result-link.tsx
@@ -10,7 +10,6 @@ export interface ResultLinkEventProps {
 
 export interface ResultLinkProps extends ResultLinkEventProps {
   href: string;
-  target?: string;
   className?: string;
   part?: string;
   title?: string;
@@ -55,7 +54,6 @@ export const bindAnalyticsToLink = (
 export const LinkWithResultAnalytics: FunctionalComponent<ResultLinkProps> = (
   {
     href,
-    target,
     className,
     part,
     title,
@@ -76,7 +74,6 @@ export const LinkWithResultAnalytics: FunctionalComponent<ResultLinkProps> = (
       part={part}
       href={filterProtocol(href)}
       title={title}
-      target={target}
       ref={(el) => {
         if (ref) {
           ref(el);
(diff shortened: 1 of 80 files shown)


packages/atomic/src/components/search/result-template-components/atomic-result-printable-uri/atomic-result-printable-uri.tsx
```diff
@@ -15,9 +15,12 @@ import {
   FocusTargetController,
 } from '../../../../utils/accessibility-utils';
 import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+import {getAttributesFromLinkSlot} from '../../result-link/attributes-slot';
 
 /**
  * The `atomic-result-printable-uri` component displays the URI, or path, to access a result.
+ *
+ * @slot attributes - Lets you pass [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) down to link elements, overriding other attributes, to be used exclusively with an "a" tag such as `<a slot="attributes" target="_blank" download></a>`.
```
jpmarceau, line 23:
```suggestion
 * @slot attributes - Lets you pass [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) down to anchor elements, overriding other attributes. 
 * To be used exclusively in anchor elements, such as: `<a slot="attributes" target="_blank" download></a>`.
```

jpmarceau (approved):
👍

packages/atomic/src/components/search/result-link/result-link.tsx
```diff
@@ -10,7 +10,7 @@ export interface ResultLinkEventProps {
 
 export interface ResultLinkProps extends ResultLinkEventProps {
   href: string;
-  target: string;
+  target?: string;
```
ThibodeauJF, line 13: I'd remove like in the other PR

ThibodeauJF (approved):

olamothe (approved):

PR #2527, opened 2022-11-01
fix(headless): folding cq should support spaces

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2111

The collection "Sea Lilies" now loads properly.
Also updated the folding sample as a bonus.
Screen Shot 2022-11-01 at 9 52 58 AM


btaillon (approved):

olamothe (approved):
The fix is fine, but to have a folding field that does not respect this documentation is weird:

https://docs.coveo.com/en/2718/javascript-search-framework/folding-fields

From the doc:

```
Important: 

Moreover, values in the child and parent fields may only contain up to 60 alphanumeric characters
```

ThibodeauJF:
We should read our own docs 🤣

PR #2528, opened 2022-11-01
fix(quantic): fixed facet value overflow

nathanlb:
Fixed the facet value overflow behaviour. It worked correctly before but we seem to have lost it in some more recent modifications. Pretty much every type of facet had the same issue.

Also added some title attributes so a user can hover over a facet value and see the whole string if it's truncated.
Also removed some `nowrap` classes which don't exist. `slds-nowrap` would be valid but clearly we don't need them.

## Before
Screen Shot 2022-11-01 at 1 03 52 PM

#### Regular value
Screen Shot 2022-11-01 at 1 20 44 PM

#### Search result
Screen Shot 2022-11-01 at 4 00 33 PM

### Timeframe Facet
Screen Shot 2022-11-01 at 4 01 03 PM

### Numeric Facet
Screen Shot 2022-11-01 at 4 01 14 PM

#### Regular value
Screen Shot 2022-11-01 at 3 44 27 PM

#### Search result
Screen Shot 2022-11-01 at 3 59 33 PM

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
+++ b/packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
@@ -10,10 +10,10 @@
   >
     <template if:true={nonActiveParent}>
       <div
-        class="facet__parent-value-option facet__value-option slds-grid slds-grid_vertical-align-center slds-size_1-of-1 slds-p-around_x-small"
+        class="facet__parent-value-option facet__value-option slds-grid slds-truncate slds-grid_vertical-align-center slds-size_1-of-1 slds-p-around_x-small"
       >
         <lightning-icon class="slds-m-right_xxx-small" icon-name="utility:chevronleft" size="xx-small"></lightning-icon>
-        <span class="nowrap">{item.value}</span>
+        <span class="slds-truncate" title={item.value}>{item.value}</span>
         <span class="facet__number-of-results slds-var-m-left_xx-small">({item.numberOfResults})</span>
       </div>
     </template>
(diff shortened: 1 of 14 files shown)


packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
```diff
@@ -39,7 +40,7 @@
               <span class="facet__number-of-results slds-var-m-left_xx-small">({item.numberOfResults})</span>
             </div>
             <div class="slds-col slds-p-left_none">
-              <span class="facet__path">{labels.inLabel} {item.localizedPath}</span>
+              <span class="facet__path slds-truncate">{labels.inLabel} {item.localizedPath}</span>
```
mmitiche, line 43: don't we need to specify a `title` attribute here as well?

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.html
```diff
@@ -24,6 +24,7 @@
     <template if:true={isChecked}>
       <span
         class="slds-truncate slds-var-m-right_xx-small slds-var-m-top_xxx-small nowrap facet__value-text facet__value_selected"
+        title={formattedFacetValue}
         >{formattedFacetValue}</span
       >
       <span class="nowrap slds-var-m-top_xxx-small facet__number-of-results facet__value_selected"
```
mmitiche, line 30: can we remove this `nowrap` CSS class? I can't find it defined anywhere.
```suggestion
      <span class="slds-var-m-top_xxx-small facet__number-of-results facet__value_selected"
```

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.html
```diff
@@ -24,6 +24,7 @@
     <template if:true={isChecked}>
       <span
         class="slds-truncate slds-var-m-right_xx-small slds-var-m-top_xxx-small nowrap facet__value-text facet__value_selected"
```
mmitiche, line 26:
```suggestion
        class="slds-truncate slds-var-m-right_xx-small slds-var-m-top_xxx-small facet__value-text facet__value_selected"
```

mmitiche (approved):
LGTM 👌

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.html
nathanlb: yeah i think the correct one would be `slds-nowrap` but clearly we don't need it either. Will do.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.html
nathanlb: Yeah might as well

PR #2530, opened 2022-11-03
feat(quantic): QuanticQueryError updated to support the insight use case

mmitiche:
[SFINT-4674](https://coveord.atlassian.net/browse/SFINT-4674)

### In this PR: 
- `QunticQueryError` component updated to be able to work correctly with both the Headless search bundle and the Headless insight bundle.


nathanlb (approved):
That's what i'm talking about!

lbergeron (approved):

PR #2531, opened 2022-11-03
feat(atomic): add slack custom result templates

jbvouma:
**[SVCINT-1061](https://coveord.atlassian.net/browse/SVCINT-1061)**

I added the templates inside the `folding.html` because Slack uses folded results as well (for threads).

**Results:**
image


packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      justify-content: space-between;
+                    }
+
+                    ::part(button) {
+                      margin: 8px 0 0 50px;
+                    }
+
+                    ::part(show-hide-button) {
+                      margin: 8px 0 0 50px;
+                    }
```
btaillon, line 216: I think it would be easier to understand what these parts refer to if we preceded them with their tag.

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                    ::part(children-root) {
+                      margin-left: 70px;
+                      border-left: thin solid #bcc3ca;
+                    }
+
+                    div.result-root atomic-result-image {
+                      height: 30px;
+                      width: 30px;
+                      margin-right: 16px;
+                      border-radius: 4px;
```
btaillon, line 125: I don't see the border radius when I look at the screenshot you included, you may need to add `overflow: hidden`.

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                    .slack-title-icon {
+                      display: flex;
+                      width: 100%;
+                      margin-top: 8px;
+                    }
+
+                    .slack-badge::part(result-badge-element) {
+                      background-color: #86d8ff;
+                      color: white;
+                      border-radius: 32px;
```
btaillon, line 111: If you want the border radius to be half the badge's height, I would recommend to do this:
```css
height: calc(var(--row-height) / 2);
```

I didn't test this, but it should work in theory, since the badge section exposes a `--row-height` variable which is used by the badge.

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
           <atomic-layout-section section="results">
             <atomic-smart-snippet></atomic-smart-snippet>
             <atomic-folded-result-list image-size="small">
+              <atomic-result-template must-match-sourcetype="Slack">
+                <template>
+                  <style>
+                    div.result-root.with-sections.display-list.density-normal atomic-result-section-visual {
+                      margin-right: 16px;
```
btaillon, line 93: I recommend universally using `rem` instead of `px` for accessibility.

e.g.:
```suggestion
                      margin-right: 1rem;
```

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      background-color: #86d8ff;
+                      color: white;
+                      border-radius: 32px;
+                      padding-top: 2px;
+                      padding-bottom: 2px;
+                    }
+
+                    ::part(children-root) {
+                      margin-left: 70px;
+                      border-left: thin solid #bcc3ca;
```
btaillon, line 118: it's subjective, but we generally use `var(--atomic-neutral)` for borders, so I think it would be a good idea to use it here too.

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      flex-shrink: 0;
+                      height: 32px;
+                      width: 32px;
+                      background-size: cover;
+                      margin-right: 16px;
+                    }
+
+                    .slack-msg-result-section-title {
+                      display: flex;
+                      background-color: #f6f7f9;
```
btaillon, line 138: Instead of `#f6f7f9`, I would recommend using the variable with the same value `var(--atomic-neutral-light)`

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      border: 1px solid #f6f7f9;
```
btaillon, line 139: it looks like the border may be invisible since the background has the same color

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      border-radius: 8px;
+                      flex-grow: 1;
+                      padding: 8px 8px 0px 8px;
+                    }
+
+                    .slack-link {
+                      color: #1372ec;
```
btaillon, line 146:
```suggestion
                      color: var(--atomic-primary);
```

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      font-size: 12px;
+                    }
+
+                    .slack-result-date {
+                      color: #616061;
```
btaillon, line 151: this is fairly close to `var(--atomic-neutral-dark)`.

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      font-size: 16px;
+                      margin-right: 4px;
+                      font-weight: bold;
+                    }
+
+                    .slack-result-thread-bottom-text {
+                      font-style: normal;
+                      font-weight: 700;
+                      font-size: 12px;
+                      color: #626971;
```
btaillon, line 176:
```suggestion
                      color: var(--atomic-neutral-dark);
```

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                    </atomic-result-section-badges>
+                  </div>
+                  <atomic-field-condition must-match-slackmessagetype="ThreadStart,Normal">
+                    <div class="slack-title-icon">
+                      <atomic-result-section-visual class="slack-atomic-result-icon" image-size="icon">
+                        <atomic-result-icon>
+                          <img
+                            loading="lazy"
+                            src="https://static.cloud.coveo.com/searchui/v2.10096/image/slack.svg"
+                            class="thumbnail"
```
btaillon, line 259: The thumbnail class seems unused

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                      background-color: #f6f7f9;
+                      border: 1px solid #f6f7f9;
+                      border-radius: 8px;
+                      flex-grow: 1;
+                      padding: 8px 8px 0px 8px;
+                    }
+
+                    .slack-link {
+                      color: #1372ec;
+                      font-size: 12px;
```
btaillon, line 147: I would recommend that you use our font size variables here.

For instance: https://github.com/coveo/ui-kit/blob/fd0382e36f60be331f309feaf79a8380da41921b/packages/atomic/src/themes/coveo.css#L36

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                  <atomic-field-condition if-not-defined="slackmessagetype">
+                    <div class="slack-title-icon">
+                      <atomic-result-section-visual class="slack-atomic-result-icon" image-size="icon">
+                        <atomic-result-icon>
+                          <img
+                            loading="lazy"
+                            src="https://static.cloud.coveo.com/searchui/v2.10096/image/slack.svg"
+                            class="thumbnail"
+                          />
+                        </atomic-result-icon>
```
btaillon, line 325: Instead of an `img` for the default icon, I recommend that you use `atomic-icon` for SVG icons.

```suggestion
                        <atomic-result-icon>
                          <atomic-icon
                            icon="https://static.cloud.coveo.com/searchui/v2.10096/image/slack.svg"
                          />
                        </atomic-result-icon>
```

packages/atomic/src/pages/examples/folding.html
```diff
@@ -76,6 +86,365 @@
...
+                                <div>
+                                  <div>
+                                    <span class="slack-result-author-name">
+                                      <atomic-result-text field="slackmessageauthor"></atomic-result-text>
+                                    </span>
+                                    <span>
+                                      <atomic-result-date
+                                        class="slack-result-date"
+                                        field="date"
+                                        format="h:mm A - MM/DD/YYYY"
```
btaillon, line 422: I guess that's subjective, but I think the MM/DD/YYYY and DD/MM/YYYY formats are confusing, I would prefer the ISO format

btaillon (approved):
very good example, looks great!

packages/atomic/src/pages/examples/folding.html
btaillon: beware: `icon="assets://slack"` isn't equivalent, since the "slack" icon in our assets is colorless.

mikegron (approved):
Awesome!

Pour confirmer @ThibodeauJF ont met ca en exemple parce qu'il y a pas vraiment de plan d'avoir des "default templates" right?

btaillon:
> Awesome!
> 
> Pour confirmer @ThibodeauJF ont met ca en exemple parce qu'il y a pas vraiment de plan d'avoir des "default templates" right?

There's no plan to have default templates.

packages/atomic/src/pages/examples/folding.html
jbvouma: Yes it woul be invisible (kind of) otherwise it is black.

packages/atomic/src/pages/examples/folding.html
jbvouma: Yes this radius isn't needed here

packages/atomic/src/pages/examples/folding.html
jbvouma: Didn't know about the `var(--...`

packages/atomic/src/pages/examples/folding.html
jbvouma: not needed. will be removed.

dguerinCoveo (approved):

PR #2532, opened 2022-11-03
fix(quantic): display of QuerySummary adjusted in the example search page

mmitiche:
[SFINT-4565](https://coveord.atlassian.net/browse/SFINT-4565)

 -  display of QuerySummary adjusted in the example search page.

### Before the fix:

image-20220630-182746

### After the fix:
ExampleSearch

Additionally I fixed a margin issue in the QuanticNumberButton:
#### Before: 
Before

#### After: 
MarginFixed

PS: The same exact fix will be applied to the insightFullSearch in SFIINT


nathanlb (approved):
gg

lbergeron (approved):

PR #2537, opened 2022-11-04
fix(quantic): padding of the QuanticSort component fixed

mmitiche:
[SFINT-4537](https://coveord.atlassian.net/browse/SFINT-4537)

- #### padding of the QuanticSort component fixed

#### Before:
image-20220525-203544

#### After:
sort
Sort2


erocheleau (approved):
much better

PR #2538, opened 2022-11-04
feat(quantic): added example insight panel

nathanlb:
Added example insight panel component in the `force-app/examples` folder. 
This component will not be packaged with quantic.
It is to enable an easier development flow for insight panel components.

Screen Shot 2022-11-06 at 9 38 38 AM


mmitiche (approved):

packages/quantic/force-app/examples/main/lwc/exampleInsightPanel/exampleInsightPanel.js
```diff
@@ -0,0 +1,70 @@
...
+} from 'c/quanticHeadlessLoader';
+
+export default class ExampleInsightPanel extends LightningElement {
+  /** @type {string} */
+  @api engineId = 'example-insight-panel';
+  /** @type {string} */
+  @api insightId = '6a333202-b1e0-451e-8664-26a1f93c2faf';
+
+  connectedCallback() {
+    this.template.addEventListener(
```
mmitiche, line 20: It would be nice to remove the event listener when the component is disconnected.

packages/atomic-angular/projects/atomic-angular/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
...
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.33.6](https://github.com/coveo/ui-kit/compare/@coveo/atomic-angular@1.33.5...@coveo/atomic-angular@1.33.6) (2022-11-04)
+
+**Note:** Version bump only for package @coveo/atomic-angular
+
+
+
+
+
```
btaillon, line 13: there's a bunch of unrelated changes in the PR 🤔

nathanlb: Indeed. I think I ran `npm run build` and then committed the whole change set without combing through it. I'll revert those. But It's weird that this would even happen tbh.

nathanlb: nvm it was just that I hadn't pushed the updated parent branch.

packages/quantic/force-app/examples/main/lwc/exampleInsightPanel/exampleInsightPanel.js-meta.xml
```diff
@@ -0,0 +1,22 @@
...
+    </targets>
+    <targetConfigs>
+        <targetConfig targets="lightningCommunity__Default">
+            <property name="engineId" type="String" default="example-search" label="Enter Coveo Headless engine ID"/>
+            <property name="insightId" type="String" default="default" label="The insight panel config ID"/>
+        </targetConfig>
+        <targetConfig targets="lightning__AppPage, lightning__RecordPage, lightning__HomePage">
+            <property name="engineId" type="String" default="example-search" label="Enter Coveo Headless engine ID"/>
+            <property name="insightId" type="String" default="default" label="The insight panel config ID"/>
+        </targetConfig>
```
lbergeron, line 20: Why are you using two identical target configs? Couldn't you use the same config for all targets?

```suggestion
        <targetConfig targets="lightning__AppPage, lightning__RecordPage, lightning__HomePage, lightningCommunity__Default">
            <property name="engineId" type="String" default="example-search" label="Enter Coveo Headless engine ID"/>
            <property name="insightId" type="String" default="default" label="The insight panel config ID"/>
        </targetConfig>
```

lbergeron (approved):

packages/quantic/force-app/examples/main/lwc/exampleInsightPanel/exampleInsightPanel.js-meta.xml
nathanlb: It's been a while since I made the one for the exampleSearch but i'm about 90% sure it used to be that you needed two separate ones for Community builder and Console app builder. I just tried with your suggestion and it looks like it's working now. 👍 Nice!

PR #2545, opened 2022-11-09
fix(headless): modify delimiting character for facet in url manager

olamothe:
As discussed in Slack.

New format for facets in `search-parameter-manager` will be `f-facetid=facetvalue1,facetvalue2` as opposed to `f[facetid]=facetvalue1,facetvalue2` before.

https://coveord.atlassian.net/browse/KIT-2119

# Deprecations
## Headless
* The [UrlManager](https://docs.coveo.com/en/headless/latest/reference/search/controllers/url-manager/) now serializes facets with a dash instead of brackets.
  * For instance, `f[my-facet]=valueA,valueB` would become `f-my-facet=valueA,valueB`
  * This is because brackets are [reserved characters in RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2), which get serialized by certain applications resulting in broken search page URLs.

## Atomic
* The hash in the URL representing the state of the search page now serializes facets with a dash instead of brackets.
  * For instance, `f[my-facet]=valueA,valueB` would become `f-my-facet=valueA,valueB`
  * This is because brackets are [reserved characters in RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2), which get serialized by certain applications resulting in broken search page URLs.

### DIFF
--- a/packages/atomic/cypress/e2e/breadbox.cypress.ts
+++ b/packages/atomic/cypress/e2e/breadbox.cypress.ts
@@ -188,7 +188,7 @@ describe('Breadbox Test Suites', () => {
       new TestFixture()
         .with(addBreadbox())
         .with(addFacet({field, label}))
-        .withHash(`f[${field}]=${activeValues.join(',')}`)
+        .withHash(`f-${field}=${activeValues.join(',')}`)
         .init();
       BreadboxSelectors.breadcrumbButton().then((buttons) =>
         cy.wrap(buttons.filter(':visible').length).as('numberOfVisibleButtons')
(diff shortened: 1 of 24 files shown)


olamothe:
@btaillon I'm not sure where you are keeping tracks of the changes to be documented, but let's not forget that one ;)

ThibodeauJF (approved):

nathanlb (approved):
Si ca passe ca passe ✅

olamothe:
> Si ca passe ca passe ✅

It's for v2 release, as part of breaking change for Headless.

btaillon (approved):
I think we should add a e2e test where we serialize all reserved characters, and then test a URL that has multiple facets with weird names.

btaillon:
> @btaillon I'm not sure where you are keeping tracks of the changes to be documented, but let's not forget that one ;)

I keep track of changes by adding them in the description of the PR and updating the v2 confluence document

PR #2547, opened 2022-11-10
feat(quantic): 

mmitiche:
[SFINT-4687](https://coveord.atlassian.net/browse/SFINT-4687)

#### Why do we need a white background in the `QuanticTabBar` component?
Because currently in the InsightFull search, the Quantic Tab Bar component is only component with a grey background, it was pervasively discussed with our UX team that a white background would be better.

InsightFullSearch

#### After this PR:
InsightFulSearchAfter


lbergeron (approved):

nathanlb (approved):
Was the grey added for accessibility? IMO light theme should be the default because every other component has a white background. But not a big deal feel free to merge as is.

mmitiche:
> Was the grey added for accessibility? IMO light theme should be the default because every other component has a white background. But not a big deal feel free to merge as is.

nope, this wasn't added with the accessibility stuff, it was made like that in the first time we created this component.

PR #2548, opened 2022-11-10
ci: publish prerelease versions

btaillon:

In theory, this PR should cause the following to happen:
* Any commit on a branch titled `prerelease/**` should execute the same CI as the `master` branch
  * Commit bumps on this branch will instead bump packages with Lerna's default `-pre.#` suffix (see [conventional-prerelease](https://github.com/lerna/lerna/tree/main/commands/version#--conventional-prerelease)).
  * Jenkins will skip updating the `alpha` NPM tag for any commit that isn't on the `master` branch.
  * The deployment pipeline will disable `package_rollout` for any commit that isn't on the `master` branch, which will prevent it from ever reaching `stg` and `prd` (which are responsible for publishing to the staging CDN, publishing to the prod CDN, updating the NPM `beta` tag and updating the NPM `latest` tag).

### DIFF
--- a/.deployment.config.json
+++ b/.deployment.config.json
@@ -190,7 +190,8 @@
     "no_endpoint": true
   },
   "package_rollout": {
-    "only_consider_changesets_after": "b244fe702d8e96d016a52715e92c8131acfde3ba"
+    "only_consider_changesets_after": "b244fe702d8e96d016a52715e92c8131acfde3ba",
+    "disabled": $[STOP_AT_DEV]
   },
   "deployment_config_version": 7
 }
--- a/.github/workflows/masterbot.yml
+++ b/.github/workflows/masterbot.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
+      - 'prerelease/**'
 jobs:
   build:
     name: 'Build'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,10 +1,17 @@
+def parseSemanticVersion(String version) {
+  def semanticVersionRegex = /^(((([0-9]+)\.[0-9]+)\.[0-9]+)(?:\-.+)?)$/
+  def (_, prerelease, patch, minor, major) = (version =~ semanticVersionRegex)[0]
+  return [major, minor, patch, prerelease]
+}
+
 node('heavy && linux && docker') {
   checkout scm
   def tag = sh(script: "git tag --contains", returnStdout: true).trim()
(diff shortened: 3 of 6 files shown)


.deployment.config.json
```diff
@@ -190,7 +190,8 @@
     "no_endpoint": true
   },
   "package_rollout": {
-    "only_consider_changesets_after": "b244fe702d8e96d016a52715e92c8131acfde3ba"
+    "only_consider_changesets_after": "b244fe702d8e96d016a52715e92c8131acfde3ba",
+    "disabled": $[STOP_AT_DEV]
```
btaillon, line 194: The quotation marks are missing here, and it's on purpose. This will be replaced by the variable directly.

.github/workflows/masterbot.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
+      - 'prerelease/**'
```
btaillon, line 6: Inspired by https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs

Jenkinsfile
```diff
@@ -1,10 +1,17 @@
+def parseSemanticVersion(String version) {
+  def semanticVersionRegex = /^(((([0-9]+)\.[0-9]+)\.[0-9]+)(?:\-.+)?)$/
+  def (_, prerelease, patch, minor, major) = (version =~ semanticVersionRegex)[0]
+  return [major, minor, patch, prerelease]
+}
```
btaillon, line 5: Tested this in a repl.

Try executing the following in https://www.jdoodle.com/execute-groovy-online/:
```groovy
def parseSemanticVersion(String version) {
  def semanticVersionRegex = /^(((([0-9]+)\.[0-9]+)\.[0-9]+)(?:\-.+)?)$/
  def (_, prerelease, patch, minor, major) = (version =~ semanticVersionRegex)[0]
  return [major, minor, patch, prerelease]
}

def printVersion(String version) {
  def (major, minor, patch, prerelease) = parseSemanticVersion(version)
  println "== ${version} =="
  println "Major: ${major}"
  println "Minor: ${minor}"
  println "Patch: ${patch}"
  println "Prerelease: ${prerelease}"
}

printVersion("123.456.789-pre.1")
print "\n"
printVersion("123.456.789")
```

ThibodeauJF (approved):

PR #2549, opened 2022-11-10
feat(atomic): added query suggestions to the insight search box

nathanlb:
Like the title says I added query suggestions to the `<atomic-insight-search-box>`.
I'll add a couple cypress tests but figured the review process could get a little head start.
I haven't contributed much to Atomic so please let me know if things are "up to code". (haha pun)

Demo

https://user-images.githubusercontent.com/16785453/201170036-e7c0ffff-5ff5-44fe-a4cb-7440a05ae361.mov


nathanlb:
> It's a bit painful, but we should extract the common code (in between this new search box and the other) into a new common class... Similar to the result list common or the facet common 😬

my middle name is "pain"

btaillon (commented):
I'm not sure if search box common is doing enough work here.

I wonder if it would be possible to create a generic `@internal` component similar to `atomic-modal` which contains almost all the logic.

What the atomic-search-box is doing currently is similar to the WCAG [Editable Combobox With List Autocomplete](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-autocomplete-list.html), so our component could be something like `atomic-autocomplete-combobox`. 

Ideally, I think the only responsibility for the insight and search components would be to listen for events bubbling up to them and forward them to the generic component.

If all of this is too complicated, don't worry about it, I can review this and we can merge this.

btaillon:
One thing I should probably mention, is that a significant part of the complexity we have in this component is due to instant results, since it needs to support a left and right pane. We basically can't cleanly represent this to screen readers, so I changed the role of the suggestions to "application", which allows us to manually handle accessibility. This may make it impossible to extract more code...

nathanlb:
Yeah the insight panel is unlikely to have two panes given the width of the component.

Moving some common helpers to searchboxcommon makes sense to me but further than that i'm concerned the two searchboxes would become unnecessarily linked together despite working quite differently when it comes to query suggestions.

ThibodeauJF (approved):
Good enough for now, we can extract more in the future as needed

PR #2550, opened 2022-11-15
feat(quantic): new QuanticResultAction component created

mmitiche:
[SFINT-4476](https://coveord.atlassian.net/browse/SFINT-4476)
[SFINT-4477](https://coveord.atlassian.net/browse/SFINT-4477)
Design doc: https://docs.google.com/document/d/1P3BAsHCgS4QpLxCfT1XJya4f1fvdo2FkKJ5DmzhLaHI/edit

### Demo:

https://user-images.githubusercontent.com/86681870/201935990-ab89ee5a-f4d9-4239-ba8e-1056ab69d372.mov

### Usage:
```
    <c-quantic-result-action-bar>
      <c-quantic-result-action toggle-mode selected={attachSelected} onclick={attach} icon-name="utility:attach" label-when-off="Attach to case" label-when-on="Detach from case"></c-quantic-result-action>
      <c-quantic-result-action onclick={send} icon-name="utility:email" label-when-hover="Send as email"></c-quantic-result-action>
      <c-quantic-result-action onclick={post} icon-name="utility:chat" label-when-hover="Post to feed"></c-quantic-result-action>
      <c-quantic-result-action onclick={preview} icon-name="utility:preview" label-when-hover="Quick view"></c-quantic-result-action>
      <c-quantic-result-action onclick={copy} icon-name="utility:copy" label-when-hover={copyLabel}></c-quantic-result-action>
    </c-quantic-result-action-bar>
```

### Role of the Quantic Result Action Component:
Displays a result action button with multiple options, it supports the toggle mode which allows the button to behave like a toggle button, an example of this is the Attach to case button in the demo video linked.

QuanticResultAction

### Role of the Quantic Result Action Bar component:
Informs the Quantic Result Action components about their order so the proper CSS classes will be applied to each Result Action:

- When a Quantic Result Action is in the first position, the left corners of the button are rounded.

- When a Quantic Result Action is in the last position, the right corners of the button are rounded.

- When a Quantic Result Action is in a middle position, all the corners of the button are sharp.

[SFINT-4477]: https://coveord.atlassian.net/browse/SFINT-4477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathanlb:
Could you possibly add it to the exampleInsightPanel so we can see how it's intended to be fully integrated into an interface?

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+ * @category Insight Panel
+ * @example
+ * <c-quantic-result-action-bar>
+ *   <c-quantic-result-action onclick={actionOne} icon-name="utility:attach" label-when-hover="Action One"></c-quantic-result-action>
+ *   <c-quantic-result-action onclick={actionTwo} icon-name="utility:email" label-when-hover="Action Two"></c-quantic-result-action>
+ *   <c-quantic-result-action onclick={actionThree} icon-name="utility:chat" label-when-hover="Action Three"></c-quantic-result-action>
+ * </c-quantic-result-action-bar>
+ */
+export default class QuanticResultActionBar extends LightningElement {
+  resultActionButtons = [];
```
lbergeron, line 15: It would be nice to type this variable.

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+   *  Returns the sorted result action buttons based on their order in the DOM.
+   */
+  getSortedButtons(buttons) {
+    const sortedButtons = Object.values(buttons);
+
+    sortedButtons.sort((a, b) => {
+      const position = a.ref.compareDocumentPosition(b.ref);
+
+      if (
+        position & Node.DOCUMENT_POSITION_FOLLOWING ||
```
erocheleau, line 66: Bitwise `&` ?

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+        position & Node.DOCUMENT_POSITION_CONTAINED_BY
```
erocheleau, line 67:
```suggestion
        if(position && (Node.DOCUMENT_POSITION_FOLLOWING || Node.DOCUMENT_POSITION_CONTAINED_BY))
```

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+   */
+  handleButtonRegister = (event) => {
+    event.stopPropagation();
+    const callbacks = event.detail;
+    const ref = event.target;
+    this.resultActionButtons.push({callbacks, ref});
+  };
+
+  /**
+   * Handles the slot change event.
```
erocheleau, line 36: But what does it actually do? ;)

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+  connectedCallback() {
+    this.addEventListener('resultactionregister', this.handleButtonRegister);
+  }
+
+  disconnectedCallback() {
+    this.removeEventListener('resultactionregister', this.handleButtonRegister);
+  }
+
+  /**
+   * Handles the result action button registration event.
```
erocheleau, line 26: What does it do?

lbergeron (commented):
A few things are still missing, but it looks very good so far!

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
```diff
@@ -0,0 +1,40 @@
...
+  left: 50%;
+}
+
+.slds-popover_tooltip {
+  width: max-content;
+}
+
+.result-action_tooltip {
+  position: absolute;
+  top: -3rem;
```
erocheleau, line 12: That is dependent on the font-size, have you tried changing font-size to see if that still remains aligned?

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
```diff
@@ -0,0 +1,40 @@
...
+  --sds-c-button-radius-border: 4px 0px 0px 4px;
+}
+
+.result-action_middle {
+  --sds-c-button-radius-border: 0px;
+}
+
+.result-action_last {
+  --sds-c-button-radius-border: 0px 4px 4px 0px;
+}
```
erocheleau, line 40: Is there a way to work with `:first-child` and `:last-child` instead here?

Something like this:

```
.result-actionList:first-child .result-action_button {
  --sds-c-button-radius-border: 4px 0px 0px 4px;
}

.result-actionList:last-child .result-action_button {
  --sds-c-button-radius-border: 0px 4px 4px 0px;
}
```

You'd have to be careful if there's only 1 button for example by instead of setting the class `result-action_button` you could give it the class `result-action_only-button` so you can give it both rounded corners (I don't know how first-child and last-child will behave if there's only 1 child element)

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+ *  The `QuanticResultActionBar` component displays the Quantic Result Actions in a proper way with the right styles.
+ * @category Search
+ * @category Insight Panel
+ * @example
+ * <c-quantic-result-action-bar>
+ *   <c-quantic-result-action onclick={actionOne} icon-name="utility:attach" label-when-hover="Action One"></c-quantic-result-action>
+ *   <c-quantic-result-action onclick={actionTwo} icon-name="utility:email" label-when-hover="Action Two"></c-quantic-result-action>
+ *   <c-quantic-result-action onclick={actionThree} icon-name="utility:chat" label-when-hover="Action Three"></c-quantic-result-action>
+ * </c-quantic-result-action-bar>
+ */
```
erocheleau, line 13: The other idea I had in mind for the design of this component was the following, did you consider it working like this instead and would it make it more simple?

```
//Template.html

<c-quantic-result-action-bar actions={actions}></c-quantic-result-action-bar>
```

```
//Template.js

this.actions = [
{ actionTrigger: (details) => {...}, icon: "utility:attach", labelWhenHover: "Action One"},
{ actionTrigger: (details) => {...}, icon: "utility:email", labelWhenHover: "Action Two"},
{ actionTrigger: (details) => {...}, icon: "utility:chat", labelWhenHover: "Action Three"},
];
```

```
//QuanticResultActionBar.html

<template>
  <template for:each={actions} for:item="act">
    <c-quantic-result-action onclick={act.actionTrigger} icon-name={act.icon} label-when-hover={act.labelWhenHover}></c-quantic-result-action>
  </template>
</template>
```

erocheleau: What do you think of handling that via just a code array versus adding multiple HTML, where you need to write code for the handler function anyway?

Also the handler function of each of these actions need to know the current result they are rendering so they are able to get some result fields for example for attach, etc.

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+        const order =
+          index === 0
+            ? 'first'
+            : index === array.length - 1
+            ? 'last'
+            : 'middle';
+        resultActionButton.callbacks.setOrder(order);
+      });
+    }
+  }
```
erocheleau, line 54: If possible, we should try to remove the need for this. I think each button should know their function to call on click, and the order shouldn't matter.

packages/quantic/force-app/main/default/lwc/quanticResultAction/__tests__/quanticResultAction.test.js
```diff
@@ -0,0 +1,339 @@
...
+
+      const statefulIconButton = element.shadowRoot.querySelector(
+        'lightning-button-icon-stateful'
+      );
+
+      expect(statefulIconButton).not.toBeNull();
+      expect(statefulIconButton.iconName).toBe(exampleIconName);
+    });
+
+    it('should dispatch a click event when the lightning button is clicked', async () => {
```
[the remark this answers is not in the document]
mmitiche: Absolutely, the loading state is much needed, let me see how we can support it here.

mmitiche:
> Could you possibly add it to the exampleInsightPanel so we can see how it's intended to be fully integrated into an interface? With the visibility toggle on hover and the click handler setup.

sure yeah, I'll add that!

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/__tests__/quanticResultActionBar.test.js
```diff
@@ -0,0 +1,138 @@
+// @ts-ignore
```
erocheleau, line 1: jest tests in quantic? :P

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,92 @@
...
+      orderClass = 'result-action_first';
+    } else if (order === 'middle') {
+      orderClass = 'result-action_middle';
+    } else if (order === 'last') {
+      orderClass = 'result-action_last';
+    }
+    this.resultActionOrderClasses = orderClass
+      ? `${commonButtonClass} ${orderClass}`
+      : commonButtonClass;
+  };
```
erocheleau, line 80: If possible we should get rid of this complexity

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Returns the label to be dusplayed in the tooltip.
+   */
+  get displayedLabel() {
+    return !this.toggleMode
+      ? this.labelWhenHover
+      : this.selected
+      ? this.labelWhenOn
+      : this.labelWhenOff;
```
erocheleau, line 90: Not really readable when on multiple lines:

```suggestion
    return (!this.toggleMode) ? (this.labelWhenHover) : ((this.selected) ? this.labelWhenOn : this.labelWhenOff);
```

For me, this is clearer.

But don't hesitate to break the code into parts to make it more readable.

```
let tooltipText = this.labelWhenHover;

if(this.toggleMode) {
  tooltipText = this.selected ? this.labelWhenOn : this.labelWhenOff;
}

return tooltipText;
```

packages/quantic/force-app/main/default/lwc/quanticResultAction/__tests__/quanticResultAction.test.js
```diff
@@ -0,0 +1,339 @@
...
+      const element = createTestComponent();
+      await flushPromises();
+
+      const iconButton = element.shadowRoot.querySelector(
+        'lightning-button-icon'
+      );
+
+      expect(iconButton).not.toBeNull();
+      expect(iconButton.iconName).toBe(exampleIconName);
+    });
```
erocheleau, line 116: I wouldn't be against merging all this in a single test:

```
it('should display correctly', async () => {
  const element = createTestComponent();
  await flushPromises();

  const iconButton = element.shadowRoot.querySelector(
    'lightning-button-icon'
  );
  const tooltip = element.shadowRoot.querySelector('.slds-popover_tooltip');
  
  expect(iconButton).not.toBeNull();
  expect(tooltip).not.toBeNull();
  expect(tooltip.textContent).toBe(exampleLabelWhenHover);
  expect(iconButton.iconName).toBe(exampleIconName);
});
```

packages/quantic/force-app/main/default/lwc/quanticResultAction/__tests__/quanticResultAction.test.js
```diff
@@ -0,0 +1,339 @@
...
+      const iconButton = element.shadowRoot.querySelector(
+        'lightning-button-icon'
+      );
+
+      expect(functionsMocks.listener).toHaveBeenCalledTimes(1);
+      expect(iconButton.className).toBe('result-action_button');
+    });
+  });
+
+  describe('when the toggle mode is enabled', () => {
```
erocheleau, line 191: Same as before, I would merge some of the tests, probably all tests where the `selected` is false.

erocheleau (changes requested):
If possible we should try to get rid of the first, middle, last position state. This should be managed by the DOM imo.

Did you mean to write jest tests for Quantic? :P 

Also did you consider handling the buttons via just an array and a `for:each` template instead of using slots?
The only thing I had not found a solution for that one was the quickview to be able to re-use the quickview component, but I was thinking the click handler could trigger a click on an "invisible" quantic-quickview button to simplify the whole thing.

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   *  Returns the sorted result action buttons based on their order in the DOM.
+   */
+  getSortedButtons(buttons) {
```
nathanlb, line 59: Why not define this as a proper getter?
```suggestion
  get sortedButtons(buttons) {
```

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
mmitiche: yes, that's how the `compareDocumentPosition` method is documented: https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
mmitiche: I'll clarify that in the comment, it basically sends the order information to the QuanticResultAction components inside the slots.

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
mmitiche: will clarify the comment as well, it registers the result action.

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
mmitiche: good point! let me try that and see.

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
nathanlb: agreed

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/__tests__/quanticResultActionBar.test.js
nathanlb: puzzling

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
mmitiche: The special thing about the `result-action_first` `result-action_middle` and `result-action_last` classes is that these classes need to be set directly on the `lightning-button-icon` and the `lightning-button-icon-stateful` components. these components are custom components from the Salesforce library and that's the only way to customize them following the documentation., that's why the css classes are created and used at the QuanticResultAction level.
 
Your suggestion means that these CSS rules and the `result-action_button` css class will be defined at the QuanticResultActionBar level, which is kinda problematic because the styles  can't cross the shadow DOM boundaries  and they will not be visible inside the `QuanticResultAction` component

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
mmitiche: Yes, I did consider using an array of object instead but I ended up with the slot solution because of the following reasons:
- First because the components I created are inspired by the [lightning-button-group](https://developer.salesforce.com/docs/component-library/bundle/lightning-button-group/example) component which accomplishes a very similar job as our QuanticResultActionBar component and  uses this Slots method.
- I also wanted to stay consistent with the other Quantic components that try to achieve a similar job. For example: the QuanticTabBar component, this component receives multiple QuanticTab components in its slot instead of an array of tab labels.
- Maybe more of a personal feeling, but I feel like using slots to address this kind of situation more natural than passing a whole array object property. 

What do you think? what are the pros of using an array of object property instead of using the HTML slots?
 
Any thoughts on that @nathanlb and @lbergeron ? I would like to hear your opinion as well if possible!

nathanlb: I have a slight preference for the slot method for the same reasons you outlined. However the slot handlers in the actionBar component seem a little messy.

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
mmitiche: to be fair, this doesn’t have anything to do with the function that need to be called on click and the order doesn't matter when it comes to that.
The only thing this order logic is doing, is to make the QuanticResultAction components display in the proper way.
 by proper wayI mean the following:

- when order === 'first' display the button with rounded left corners.
- when order === 'middle' display the button with sharp corners.
- when order === 'last' display the button with rounded right corners.
- when order === null display the button with ALL  rounded corners. this happens when there is only one result action button.

That's what's this logic is doing, the Quantic Result Action  component needs this information about the order, I could've used a property to feed this information to the QuanticResultAction but since the QuanticResultAction already has this information, why not just passing it to the QuanticResultAction without the need to add any new property.

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * The result to perform the action on.
+   * @api
+   * @type {Result}
+   */
+  @api result;
+
+  /** @type {string} */
+  resultActionOrderClasses;
+
```
mmitiche, line 58:
```suggestion
```

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,149 @@
...
+   * @api
+   * @type {string}
+   */
+  @api label;
+  /**
+   * The label displayed in the tooltip when the result action button is selected.
+   * @api
+   * @type {string}
+   */
+  @api labelWhenOn;
```
nathanlb, line 24:
```suggestion
  @api selectedLabel;
```
To tie it to the `selected` prop.

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,149 @@
...
+  _selected;
+  /** @type {boolean} */
+  _loading;
+
+  connectedCallback() {
+    this.setSelected(this.selected);
+    this.setLoading(this.loading);
+
+    const resultActionRegister = new CustomEvent(
+      'qunatic__resultactionregister',
```
nathanlb, line 68:
```suggestion
      'quantic__resultactionregister',
```

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Applies the proper CSS order class.
+   * @param {'first' | 'middle' | 'last'} order
+   */
+  applyCssOrderClass = (order) => {
+    const commonButtonClass = 'result-action_button';
+    let orderClass = '';
+
+    if (order === 'first') {
+      orderClass = 'result-action_first';
```
nathanlb, line 89: is it possible to do this with css directly using selectors like `:last-child` and `:first-child`? maybe that would be easier.

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,149 @@
+import {LightningElement, api} from 'lwc';
+
+/** @typedef {import("coveo").Result} Result */
+
+/**
+ * The `QuanticResultAction` component allows the end user to perform a specific action on a result.
+ * @category Search
+ * @category Insight Panel
+ * @example
+ * <c-quantic-result-action result={result} event-name={attachtocase} icon-name="utility:chat" label="Attach result" selected label-when-on="Detach result"></c-quantic-result-action>
```
nathanlb, line 10:
```suggestion
 * <c-quantic-result-action result={result} event-name="attachtocase" icon-name="utility:chat" label="Attach result" selected label-when-on="Detach result"></c-quantic-result-action>
```

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
```diff
@@ -0,0 +1,83 @@
+import {LightningElement} from 'lwc';
+
+/**
+ *  The `QuanticResultActionBar` component displays multiple result actions in a proper way and with the right styles.
+ * @category Search
+ * @category Insight Panel
+ * @example
+ * <c-quantic-result-action-bar>
+ *   <c-quantic-result-action onclick={actionOne} icon-name="utility:attach" label-when-hover="Action One"></c-quantic-result-action>
```
nathanlb, line 9: the prop names for `c-quantic-result-action` might still change but they're also a bit out of sync with what's currently in the PR.

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
nathanlb: you're so right 😅

nathanlb (commented):
I appreciate the effort you put in to write some jest tests but, as you know, we usually don't do that for quantic components so could you explain why you think we should do it here?

mmitiche:
> I appreciate the effort you put in to write some jest tests but, as you know, we usually don't do that for quantic components so could you explain why you think we should do it here?

Well, the main reason  is because unit tests felt more adequate to test the QuanticResultAction component since its building block that will be used to build other components like the QuanticCopyToClipboard for instance.
I thought about creating unit tests for the QuanticResultAction component and then add FTs when we create the QuanticCopyToClipboard component. This way we would really have a functionality to test in the FTs.
Same thing applies for the QuanticResultActionBar, unit tests for now and when we will have more than one result action we can setup new FTs to test this component as well. For example we can setup a test with the QuanticResultActionBar containing both the QuanticClipboard and the QuanticResultAction.
That's my thoughts about that, this way we would have Unit tests + FTs pretty much to cover these new components.
What do you think?

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
mmitiche: The reason I couldn't use CSS directly to address this is the following:
https://github.com/coveo/ui-kit/pull/2550#discussion_r1023173920

nathanlb: Ah okok my bad thanks

nathanlb (approved):

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
erocheleau: Okay well nevermind then if this requires a bitwise `&` haha

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
erocheleau: wow TIL

erocheleau (approved):
I think it looks good after you've made those changes if they are possible

erocheleau:
Also, are we planning to add the jest tests as part of the build process? Are they already included somewhere or are these the first Jest tests in Quantic?

nathanlb:
> Also, are we planning to add the jest tests as part of the build process? Are they already included somewhere or are these the first Jest tests in Quantic?

we already have a few jest tests for the QuanticHeadlessLoader that run on PRs so the new ones run too. See [here](https://github.com/coveo/ui-kit/actions/runs/3533539048/jobs/5929416626#step:5:340)

packages/quantic/force-app/main/default/lwc/quanticResultActionBar/quanticResultActionBar.js
erocheleau: I'm convinced there's a better CSS only way to handle that 😂

mmitiche: haha x) any idea in mind ?

erocheleau:
> > Also, are we planning to add the jest tests as part of the build process? Are they already included somewhere or are these the first Jest tests in Quantic?
> 
> we already have a few jest tests for the QuanticHeadlessLoader that run on PRs so the new ones run too. See [here](https://github.com/coveo/ui-kit/actions/runs/3533539048/jobs/5929416626#step:5:340)

Perfect, I had asked without checking :)

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -0,0 +1,149 @@
...
+   * @type {Result}
+   */
+  @api result;
+
+  /** @type {string} */
+  resultActionOrderClasses;
+  /** @type {boolean} */
+  _selected;
+  /** @type {boolean} */
+  _loading;
```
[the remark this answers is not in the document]
mmitiche: Mainly because this pattern  would break the automatic doc generation, the loading property will no longer be considered ass public property from the POV of the doc generator algorithm.

PR #2551, opened 2022-11-15
docs(headless): updated FetchInstantResultsActionCreatorPayload['cacheTimeout'] doc

btaillon:
https://coveord.atlassian.net/browse/KIT-2126

Copied from `InstantResultOptions['cacheTimeout']`

### DIFF
--- a/packages/headless/src/features/instant-results/instant-results-actions.ts
+++ b/packages/headless/src/features/instant-results/instant-results-actions.ts
@@ -1,4 +1,4 @@
-import {createAction} from '@reduxjs/toolkit';
+import {Result} from '../../case-assist.index';
 import {InstantResultSection} from '../../state/state-sections';
 import {
   validatePayload,
@@ -6,7 +6,8 @@ import {
   requiredNonEmptyString,
 } from '../../utils/validate-payload';
 import {StateNeededByExecuteSearch} from '../search/search-actions';
-import {Result} from '../../case-assist.index';
+import {createAction} from '@reduxjs/toolkit';
+
 export interface RegisterInstantResultActionCreatorPayload {
   /**
    * The search box ID.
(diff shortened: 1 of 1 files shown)


jpmarceau (approved):
👍

ThibodeauJF (approved):

packages/headless/src/features/instant-results/instant-results-actions.ts
```diff
@@ -73,7 +74,7 @@ export interface FetchInstantResultsActionCreatorPayload {
    */
   maxResultsPerQuery: number;
   /**
-   * The maximum items to be stored in the instant result list for each query.
+   * Number in milliseconds that cached results will be valid for. Defaults to 1 minute. Set to 0 so that results never expire.
```
acote-coveo, line 77: Headless has its own default of 1 minute or it assumes SearchAPI's default is of 1 minute?

packages/headless/src/features/instant-results/instant-results-actions.ts
```diff
@@ -73,7 +74,7 @@ export interface FetchInstantResultsActionCreatorPayload {
    */
   maxResultsPerQuery: number;
   /**
-   * The maximum items to be stored in the instant result list for each query.
+   * Number in milliseconds that cached results will be valid for. Set to 0 so that results never expire.
```
acote-coveo, line 77: > Set to 0 so that results never expire.

Shouldn't it be something like: `Set to 0 so that results are never cached.` ?

btaillon: This comment correctly reflects the current behavior though: a value of 0 is interpreted as "never expire".

See: https://github.com/coveo/ui-kit/blob/3519baf5857089d7c91f4ff32002cb0885382d69/packages/headless/src/features/instant-results/instant-results-state.ts#L30-L35

I believe the reason for this behavior is because we don't want to encourage customers to have "no cache", since we expect `updateQuery` to be called often (e.g., whenever somebody hovers a search suggestion). Whenever `updateQuery` is called, a whole new `/v2` search query is triggered if the cache is expired.

acote-coveo: Alright then omit my comment, this seems completely unrelated to our SearchAPI cache.Thanks for the explanation!

acote-coveo (approved):

PR #2553, opened 2022-11-16
ci: add prerelease

btaillon:
I'm merging this in `master` since in theory it shouldn't affect it. Changes from https://github.com/coveo/ui-kit/pull/2548 were copied here.

# New prerelease process
## To start a major prerelease (e.g., `1.2.3` → `2.0.0-pre.0`):
1. Make sure that HEAD refers to a version bump on `master`.
2. Create a new branch prefixed with `prerelease/` (e.g., `prerelease/headless_atomic_v2`)
3. Run `npm run bump:version:major-prerelease -- @coveo/package1 @coveo/package2 @coveo/package3`.
4. Push the new commit and its tags to the branch.

## Updating a prerelease branch:
1. Wait for `master` to reference a version bump.
2. Pull changes from `master` into the prerelease branch.

## Adding new changes to the prerelease branch:
1. Open pull requests towards the release branch
2. Squash as usual.

## Officially releasing (e.g., `2.0.0-pre.15` → `2.0.0`):
1. Create a pull request from the prerelease branch to “master”.
    * Associate with a JIRA which QA will use to validate.
    * Update the changelogs manually
    * Update all dependents to use the prerelease version (include the `-pre.` suffix)
3. Squash after a version bump on `master`.

# Changes in this PR
1. Added a `bump:version:major-prerelease` script to the root `package.json`
    * This script will create a major prerelease version bump on the current branch, and will give it all the appropriate tags, but won't push it.
    * The commit will contain the same message that `lerna` gives version bumps, including `[skip-ci]`.
    * Changelogs aren't affected.
2. Non-bump commits on a `prerelease/**` branch now trigger the same CI as `master`, except:
    * Only packages matching `/pre\.[0-9]+$/` are upgraded.
      * If a dependent is out of scope (doesn't match the above version pattern), its dependencies won't be upgraded.
    * Changelogs aren't updated.
3. Bump commits on a `prerelease/**` branch now trigger the same CI as `master`, except:
    * Only packages matching `/pre\.[0-9]+$/` are published to NPM.
    * The `alpha` tag on NPM isn't updated.
    * The package rollout stops at the `dev` environment.
4. Some scripts were converted from CommonJS (*.js) to ESM (*.cjs)
5. Various calls to `execSync` were replaced with a utility `execute` function
      * `execute`, in addition to returning its output, copies stdout & stderr to the console to make debugging easier. I caught some issues early thanks to this.

# Other details
* Packages are going to be published to the CDN based on their `major`/`minor`/`patch` version, the prerelease version will be ignored. Therefor, `2.0.0` will already be available on the dev CDN, and will be overwritten every time we update the prerelease. Out of scope packages will also update their CDN, but in theory the files should be the same anyway.


scripts/deploy/manual-major-prerelease.mjs
```diff
@@ -0,0 +1,73 @@
...
+import {commitVersionBump, stageAll, tagExists, tagPackages} from '../git.mjs';
+import {
+  getPackageDefinitionFromPackageName,
+  updatePackageVersion,
+} from '../packages.mjs';
+import {prereleaseSuffix} from '../prerelease.mjs';
+
+/**
+ * @typedef {import('../packages.mjs').PackageDefinition} PackageDefinition
+ */
```
btaillon, line 13: This is the JSDoc equivalent to `import type { PackageDefinition } from '../package';`.

scripts/git.mjs
```diff
@@ -0,0 +1,65 @@
...
+    readFileSync(resolve('.', 'lerna.json')).toString()
+  );
+  const lernaMessageSection = lernaConfig.command.version.message;
+  const updatedPackagesMessageSection = updatedPackages
+    .map(
+      (packageDef) => `\u0020-\u0020${packageDef.name}@${packageDef.version}`
+    )
+    .join('\n');
+  const message = `${lernaMessageSection}\n\n${updatedPackagesMessageSection}`;
+  await execute('git', ['commit', '--no-verify', '-m', message]);
```
btaillon, line 54: I tested this, and multi-line commits work.

scripts/packages.mjs
```diff
@@ -0,0 +1,82 @@
...
+    );
+    if (packageName === JSON.parse(originalContentAsText).name) {
+      newContent = newContent.replace(
+        /("version"\s*:\s*\")([^"]*)/,
+        '$1' + newVersion
+      );
+    }
+    writeFileSync(fullPath, newContent);
+  });
+}
```
louis-bompart, line 82: Why not parse, edit then, write? It'll be more maintainable than a RegExp nah?

louis-bompart (commented):
Nothing seems particularly wrong, tho there's too many things for me to judge/give an approve given my limited knowledge of your CI/CD

scripts/deploy/manual-major-prerelease.mjs
ThibodeauJF: nice touch

scripts/exec.mjs
```diff
@@ -0,0 +1,47 @@
...
+function trimNewline(str) {
+  return str.endsWith('\n') ? str.slice(0, -1) : str;
+}
+
+/**
+ * @param {string} command
+ * @param {readonly string[]} [args]
+ * @returns {Promise<string>}
+ */
+export function execute(command, args = []) {
```
ThibodeauJF, line 15: Could we add a package to do this type of thing? I feel like it's a bit out of our scope

scripts/packages.mjs
louis-bompart: https://github.com/sindresorhus/detect-indent

ThibodeauJF:
Nice work!

It would be nice to update the readme for upcoming pre-releases.

Some nitpick: doing the refactor to modules in the same PR as adding the pre-release specific code will make it hard to discern changes when looking back. These kind of things were a bit annoying when fixing something in the dep. process in the past.

I'd wait for Olivier's feedback too before merging

btaillon:
> Nice work!
> 
> It would be nice to update the readme for upcoming pre-releases.
> 
> Some nitpick: doing the refactor to modules in the same PR as adding the pre-release specific code & the migration from Lerna... it will make it hard to discern changes when looking back. These kind of things were a bit annoying when fixing something in the dep. process in the past.
> 
> I'd wait for Olivier's feedback too before merging

Agreed, I ideally would've liked to do it separately, but unfortunately I wasn't able to use the semantic monorepo tools with commonJS.

ThibodeauJF:
What if it was
1. PR1: switch to monorepo tools for our current release process + mjs conversion
2. Test that the main release process still works properly & iterate
3. PR2: add prerelease
4. Test, iterate...

btaillon:
> What if it was
> 
> 1. PR1: switch to monorepo tools for our current release process + mjs conversion
> 2. Test that the main release process still works properly & iterate
> 3. PR2: add prerelease
> 4. Test, iterate...

We can do that, but in this case PR1 may be as big as this PR (possibly bigger since we'd need to add changelogs and conventional graduating). It would be more consistent across both releases though. Is this what you prefer?

louis-bompart:
PR1: cjs -> mje
PR2: monorepo tools
PR3: prerelease
Etc?

PR #2557, opened 2022-11-18
feat(quantic): added link value for numeric facet

nathanlb:
Added `displayValuesAs` prop to allow both `'link'` and `'checkbox'` options on quantic numeric facets

I renamed a testing util to be more precise, hence the many changed files.


mmitiche (approved):
LGTM 👌

packages/quantic/cypress/page-objects/case-assist.ts
```diff
@@ -62,10 +63,8 @@ export function mockCaseClassification(field: string, value: Array<object>) {
 
 export function mockSfPicklistValues(field: string, values: Array<object>) {
   cy.intercept(
-    {
-      url: '**/aura?*',
-      query: {'aura.RecordUi.getPicklistValuesByRecordType': '1'},
-    },
+    routeMatchers.auraPicklistValues,
```
nathanlb, line 66: The recent change that removed the optional chaining on [this line](https://github.com/coveo/ui-kit/blob/c090c1bd676792027815c52507e2aef8ccfaab91/packages/quantic/cypress/page-objects/case-assist.ts#L71) has exposed the fact that the `query` parameter of the matcher never worked so I used a full path with wildcards instead to fix it.

PR #2558, opened 2022-11-21
feat(quantic): QuanticResultCopyToClipboard component created

mmitiche:
[SFINT-4481](https://coveord.atlassian.net/browse/SFINT-4481)
### QuanticResultCopyToClipboard component:

- The `QuanticResultCopyToClipboard` component allows the end user to copy a result to clipboard.

- The default template used to generate the text to copy to clipboard is the following: `${title}\n${clickUri}`, this default template has been chosen based on the template currently used to copy a result to clipboard in the standard agent panel.

### Demo:

https://user-images.githubusercontent.com/86681870/205173401-252b96e3-db76-48c7-8e59-d72d7e71093e.mov

### Example Quantic Result Copy To Clipboard page:

https://user-images.githubusercontent.com/86681870/205173483-1b3784b2-ca09-4557-8aa3-f11cf86bc15f.mov


packages/quantic/force-app/main/default/lwc/quanticQueryError/quanticQueryError.js
```diff
@@ -111,27 +112,10 @@ export default class QuanticQueryError extends LightningElement {
...
-  copyToClipboardFallback(text) {
-    const el = document.createElement('textarea');
-    el.value = text;
-    document.body.appendChild(el);
-    el.select();
-    document.execCommand('copy');
-    document.body.removeChild(el);
+
+    copyToClipboard(text).catch((err) => {
+      console.error('Copy to clipboard action failed.', err);
```
mmitiche, line 117:
```suggestion
      console.error('Copy to clipboard failed.', text, err);
```

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -688,3 +688,48 @@ export function isParentOf(element, targetElement) {
...
+    } catch (err) {
+      copyToClipboardFallback(text);
+    }
+  } else {
+    copyToClipboardFallback(text);
+  }
+}
+
+/**
+ * @param {string} text
```
nathanlb, line 705: Why is this the only one with a typedef? Not that I think they should all have one either it just seems odd

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  /**
+   * Resets the original label after 500ms.
+   */
+  resetOriginalLabel() {
+    // eslint-disable-next-line @lwc/lwc/no-async-operation
+    setTimeout(() => {
+      this.displayedLabel = this.label;
+    }, 500);
```
nathanlb, line 123: 500 seems a bit short. maybe 1 second? I'm thinking of an old grandma user who likely wouldn't have time to notice a difference.

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
```diff
@@ -0,0 +1,125 @@
...
+   */
+  @api label = this.labels.copy;
+  /**
+   * The label to be displayed in the tooltip of the button when the action is successful.
+   * @api
+   * @type {string}
+   */
+  @api successLabel = this.labels.copied;
+  /**
+   * The template used to generate the text to copy to clipboard.
```
nathanlb, line 51:
```suggestion
   * The template used to generate the text to copy to clipboard.
   * Use references to result [fields](https://docs.coveo.com/en/2036/index-content/about-fields) to get their value.
```

nathanlb (approved):
Looks good!

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: I'll fix that yeah all the functions need some comments and some typedefs👌

packages/quantic/force-app/main/default/lwc/quanticQueryError/quanticQueryError.js
```diff
@@ -111,27 +112,10 @@ export default class QuanticQueryError extends LightningElement {
 
   async handleCopyToClipboard() {
```
erocheleau, line 113: We already had a CopyToClipboad inside the QuanticError? 🤯

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -688,3 +690,65 @@ export function isParentOf(element, targetElement) {
     false
   );
 }
+
+/**
+ * Copies text to clipboard using the Clipboard API.
+ * https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
```
erocheleau, line 696: Maybe just a tad bit of information as to why this navigator API would fail?

Also does that work on mobile?

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -688,3 +690,65 @@ export function isParentOf(element, targetElement) {
...
+
+/**
+ * Copies text to clipboard using the DOM. 
+ * @param {string} text
+ */
+export function copyToClipboardFallback(text) {
+  const el = document.createElement('textarea');
+  el.value = text;
+  document.body.appendChild(el);
+  el.select();
```
erocheleau, line 719: Would this take away the current focus of the user?

let's say I have my cursor in the searchbox for example, would calling this remove the current place where the focus is?

If that is the case, I would argue against using that fallback for accessibility reasons to not mess up with the user's focus.

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -688,3 +690,65 @@ export function isParentOf(element, targetElement) {
...
+  return object ? object[key] : undefined;
+}
+
+/**
+ * Generates a text from a result based on a given template.
+ * @param {string} template 
+ * @param {Result} result 
+ * @returns {string}
+ */
+export function buildTemplateTextFromResult(template, result) {
```
erocheleau, line 745: Can you please let us know why this is there and not in the copy to clipboard file?
Is there another place where you would be using that?

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
```diff
@@ -0,0 +1,126 @@
...
+  getHeadlessBundle,
+} from 'c/quanticHeadlessLoader';
+import {copyToClipboard, buildTemplateTextFromResult} from 'c/quanticUtils';
+import {LightningElement, api} from 'lwc';
+
+/** @typedef {import("coveo").Result} Result */
+/** @typedef {import("coveo").InsightEngine} InsightEngine */
+
+/**
+ * The `QuanticResultCopyToClipboard` component allows the end user to copy a result to clipboard.
```
erocheleau, line 15: `to copy a result's details to clipboard.`

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
```diff
@@ -0,0 +1,126 @@
...
+ * @category Insight Panel
+ * @example
+ * <c-quantic-result-copy-to-clipboard engine-id={engineId} result={result} label="Copy" success-label="Copied!" text-template="${title}\n${clickUri}"></c-quantic-result-copy-to-clipboard>
```
erocheleau, line 18: `<c-quantic-result-copy-to-clipboard engine-id={engineId} result={result} label="Copy" success-label="Copied!" text-template="${title}\n${clickUri}"></c-quantic-result-copy-to-clipboard>`

`<c-quantic-result-copy-to-clipboard engine-id={engineId} result={result} text-template="${title}\n${clickUri}"></c-quantic-result-copy-to-clipboard>`

You don't need the values for the labels if you have a default copy/copied no?

erocheleau (commented):
look really good, just a couple of questions and that's beautiful!

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: Sure yeah, I will try investigate why it can fail and explain this in a comment. cause for now I just copied the logic that we had in the QuanticError component directly. 
Yes this works on mobile👌

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: Very nice point! thank you.
I'll test that to exactly know that behaviour occurring, but at this point I'm really considering the removal of the fallback method since it's also marked as deprecated.

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: Yes sure, actually I found this function very useful and it can help with multiple result actions. For example the Post To Feed and the Send As Email  actions will use it as well to generate a text from the result to be set in the feed or in the Email.
That's why I think it would be better in the QuanticUtils so it can be easily re-used by us in SFINT and by any customer who wants to create his own result actions.

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
mmitiche: yeah indeed, I added the values for the label just to showcase how these properties can be used. I will change the values to custom values to make more sense.

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
erocheleau: Gotcha makes sense!

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
erocheleau: Yea I just feel like if already have values for these labels and they're also translated, we shouldn't "encourage" the user to change them.

PR #2559, opened 2022-11-21
fix(quantic): fixed tab dropdown visibility

nathanlb:
Overflow is now shown so we can see the dropdown.
Also added an arbitrary max width of 150px so long tab names get truncated.

Screen Shot 2022-11-21 at 3 16 54 PM
Screen Shot 2022-11-21 at 3 17 11 PM


erocheleau (approved):
👍

lbergeron (approved):
I guess we could also update [this expectation](https://github.com/coveo/ui-kit/blob/master/packages/quantic/cypress/e2e/default-1/tab-bar/tab-bar-expectations.ts#L60) to ensure the more drop-down is visible.

mmitiche (approved):

PR #2561, opened 2022-11-22
ci: add prerelease process

btaillon:
https://coveord.atlassian.net/browse/KIT-2150

Divided & adapted from: https://github.com/coveo/ui-kit/pull/2553

We're not migrating the normal release process yet since I'm running out of time.

# New prerelease process
Look at the `README.md` to know the new release process: https://github.com/coveo/ui-kit/blob/49021bbc970304df66cdf66423bce052f7c43101/README.md#L57-L81

# Changes in this PR
1. Added a `bump:version:major-prerelease` script to the root `package.json`
    * This script will create a major prerelease version bump on the current branch, and will give it all the appropriate tags, but won't push it.
    * The commit will contain the same message that `lerna` gives version bumps, including `[skip-ci]`.
    * Changelogs aren't updated.
2. Non-bump commits on a `prerelease/**` branch now trigger the same CI as `master`, except:
    * Only packages matching `/pre\.[0-9]+$/` are upgraded.
      * If a dependent is out of scope (doesn't match the above version pattern), its dependencies won't be upgraded.
    * Changelogs aren't updated.
3. Bump commits on a `prerelease/**` branch now trigger the same CI as `master`, except:
    * Only packages matching `/pre\.[0-9]+$/` are published to NPM.
    * The `alpha` tag on NPM isn't updated.
    * The package rollout stops at the `dev` environment.

# Other details
* Packages are going to be published to the CDN based on their `major`/`minor`/`patch` version, the prerelease version will be ignored. Therefor, `2.0.0` will already be available on the dev CDN, and will be overwritten every time we update the prerelease. Out of scope packages will also update their CDN, but in theory the files should be the same anyway.
  * For example, `@coveo/headless@2.0.0-pre.2` would exclusively be uploaded to the following in the CDN:
    * https://staticdev.cloud.coveo.com/headless/v2.0/headless.esm.js
    * https://staticdev.cloud.coveo.com/headless/v2.0.0/headless.esm.js


scripts/git.mjs
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * @param {string} tag
+ */
+export async function tagExists(tag) {
+  return !!(await execute('git', ['tag', '-l', tag]));
+}
+
+export async function stageAll() {
+  await execute('git', ['add', '.']);
+}
```
louis-bompart, line 37: I reckon there's some stuff that may be available or could be available in semantic-monorepo-tools

scripts/exec.mjs
```diff
@@ -28,20 +31,25 @@ export function execute(command, args = []) {
 
     proc.stdout.on('data', (chunk) => {
       console.log(trimNewline(chunk.toString()));
-      data += chunk.toString();
+      dataChunks.push(Buffer.from(chunk));
```
louis-bompart, line 34: Why not just use buffer from the get-go?
Solutions B, exposes or copy the spawnUtils from semantic monorepo tools, you'll save some yourself some headaches

scripts/packages.mjs
```diff
@@ -1,4 +1,7 @@
+import detectIndent from 'detect-indent';
+import {existsSync, writeFileSync} from 'fs';
 import {readFileSync} from 'fs';
+import {resolve} from 'path';
```
louis-bompart, line 4:
```suggestion
import {existsSync, readFileSync, writeFileSync} from 'node:fs';
import {resolve} from 'node:path';
```

scripts/packages.mjs
```diff
@@ -12,11 +15,76 @@ export const packageDirsNpmTag = [
...
-  return {name, version};
+  return JSON.parse(readFileSync(fullPath).toString());
+}
+
+/**
+ * @param {string} packageDir E.g.: `samples/some-package`
+ * @returns {PackageDefinition}
+ */
+export function getPackageDefinitionFromPackageDir(packageDir) {
+  const fullPath = resolve('.', 'packages', packageDir, 'package.json');
```
louis-bompart, line 38: DRY 1.a

scripts/packages.mjs
```diff
@@ -12,11 +15,76 @@ export const packageDirsNpmTag = [
...
+ * @param {string} newVersion
+ * @param {string[]} depdendenciesPackageDirs E.g.: [`samples/some-package`]
+ */
+export function updatePackageVersion(
+  packageName,
+  newVersion,
+  depdendenciesPackageDirs
+) {
+  depdendenciesPackageDirs.forEach((packageDir) => {
+    const fullPath = resolve('.', 'packages', packageDir, 'package.json');
```
louis-bompart, line 66: DRY 1.b

scripts/packages.mjs
```diff
@@ -12,11 +15,76 @@ export const packageDirsNpmTag = [
...
+      }
+      if (packageName in (manifest.peerDependencies || {})) {
+        manifest.peerDependencies[packageName] = newVersion;
+      }
+    }
+    writeFileSync(
+      fullPath,
+      JSON.stringify(manifest, undefined, indent || '  ')
+    );
+  });
```
louis-bompart, line 89: - [ ] try refactoring CDX method + this one into semantic-monorepo-tools.

scripts/prerelease.mjs
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+export const prereleaseSuffix = 'pre';
+
+/**
+ * @param {string} version
+ */
+export function isPrereleaseVersion(version) {
+  return !!version.match(new RegExp(`-${prereleaseSuffix}\\.[0-9]+\$`));
+}
```
louis-bompart, line 27: - [ ] Use `semver.prerelease`

louis-bompart (changes requested):

scripts/exec.mjs
btaillon: I never really familiarized myself with buffers, I didn't think they were as simple as they actually are, and I was copying part of a recipe from StackOverflow. I'm gonna use buffers directly instead of having an array of buffers.

scripts/prerelease.mjs
btaillon: I assumed I couldn't import `semver` because it's a CommonJS module, but I didn't know that this only prevents deconstructed imports. TIL.

scripts/packages.mjs
btaillon: Going to skip this for now, since semantic monorepo tools use the topology

scripts/git.mjs
btaillon: replacing `stageAll` with `gitAdd`

scripts/packages.mjs
```diff
@@ -1,17 +1,23 @@
-import {readFileSync} from 'fs';
-import {resolve} from 'path';
-import {fileURLToPath} from 'url';
+import detectIndent from 'detect-indent';
+import {existsSync, writeFileSync, readFileSync} from 'node:fs';
+import {resolve} from 'node:path';
+import {fileURLToPath} from 'node:url';
 
-export const packageDirsNpmTag = [
+export const packageDirsNpmTag = /** @type {const} */ ([
```
btaillon, line 6: Leverages https://github.com/microsoft/TypeScript/pull/45464

btaillon:
TODO: also bump versions in private packages

scripts/packages.mjs
louis-bompart: u mean CLI I guess but Ok!

olamothe (approved):

ThibodeauJF (approved):

louis-bompart (approved):
lgtm overall, tho I don't know/understand all the process because well I'm not familiar with it

PR #2562, opened 2022-11-22
fix(quantic): changed assertion to hash

nathanlb:
clicking the suggestion actually redirects so we can't assert on the searchbox anymore.

### DIFF
--- a/packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box.cypress.ts
+++ b/packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box.cypress.ts
@@ -77,7 +77,7 @@ describe('quantic-standalone-search-box', () => {
 
       scope('when selecting from suggestions', () => {
         Actions.clickFirstSuggestion();
-        Expect.inputContains('test');
+        Expect.urlHashContains('/global-search/%40uri#q=test');
       });
 
       scope('when submitting a search', () => {


packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box.cypress.ts
```diff
@@ -77,7 +77,7 @@ describe('quantic-standalone-search-box', () => {
 
       scope('when selecting from suggestions', () => {
         Actions.clickFirstSuggestion();
-        Expect.inputContains('test');
+        Expect.urlHashContains('/global-search/%40uri#q=test');
```
lbergeron, line 80: The path that is passed to `urlHashContains` is kind of weird. Shouldn't it validate this instead?

```suggestion
        Expect.urlHashContains('q=test');
```

erocheleau: I mean it's more specific which is good no?

erocheleau (approved):

packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box.cypress.ts
nathanlb: keeping the whole /global-search/%40uri# part also validates that the default option is being set properly so IMO it's better too.

nathanlb: Default option:
https://github.com/coveo/ui-kit/blob/c5ed50049c3e9298a4bb11984a08199630bcff28/packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js#L74

nathanlb: ah then it's just a naming issue. urlHashContains checks that `cy.url()` includes the specified substring. I would just rename the expectation to `urlContains`. What do you think?

lbergeron: @erocheleau 

> I mean it's more specific which is good no?

Generally speaking, I agree with you. In this case validating the URL path using `Expect.urlHashContains` feels wrong to me. Having an expectation like `Expect.urlEndsWith` would make more sense, in my opinion.

lbergeron: I just checked [`urlHashContains` implementation](https://github.com/coveo/ui-kit/blob/fix/SVCC-2604/packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box-expectations.ts#L61) and it does not validate the hash at all, just the URL. Renaming the expectation to `urlContains` would do just fine.

lbergeron (approved):

PR #2564, opened 2022-11-23
feat(atomic): add relevance inspector modal

olamothe:
https://coveord.atlassian.net/browse/KIT-2139

[KIT-2139]: https://coveord.atlassian.net/browse/KIT-2139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -68,6 +68,7 @@ export class AtomicSearchInterface
   @Element() public host!: HTMLAtomicSearchInterfaceElement;
 
   @State() public error?: Error;
+  @State() openRelevanceInspector = false;
```
btaillon, line 71: the verb `open` makes it sound to me like this is a method as opposed to a state. I think we should rename this to something like `relevanceInspectorIsOpen`.

btaillon (approved):

packages/atomic/src/components/search/atomic-relevance-inspector/atomic-relevance-inspector.tsx
```diff
@@ -1,57 +1,59 @@
...
 
   public render() {
-    if (!this.relevanceInspectorState.isEnabled) {
-      return;
-    }
-
-    // TODO: Display data in a cleaner manner
+    const {platformUrl, organizationId} =
+      this.bindings.engine.state.configuration;
+    const {searchUid} = this.bindings.engine.state.search.response;
```
btaillon, line 27: If possible, I think we should try to get the searchUid associated with the particular search that returned the result which was clicked. I say this because we may see results from a previous query when somebody presses the "Show more" button.

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -446,6 +472,7 @@ export class AtomicSearchInterface
     return [
       this.engine && (
         <atomic-relevance-inspector
+          open={this.openRelevanceInspector}
```
ThibodeauJF, line 475: Should we conditionally render the atomic-relevance-inspector component depending on the `enableRelevanceInspector` value?

ThibodeauJF (approved):

packages/atomic/src/components/search/atomic-relevance-inspector/atomic-relevance-inspector.tsx
olamothe: It's a fair point 🤔 

However, I do not see an easy way to achieve that except to add a new (cryptic) property on each results in Headless, that associates them with their originating search ID.

We do some shenanigans to make sure the search ID does not change on load more/facet search etc, so that analytics stay consistent.

Whereas here, we'd have to add a property that is only useful for this feature, and that "most people" should not bother with. Which I don't really like.

I'm tempted to leave it as is, and we'll see if it causes problems.

Also, in a somewhat near future, we'll add capabilities to modify the search parameters with the replay API.

Let's say that we show you the wrong page of results in the admin console, it should be possible to workaround that problem in the relevance inspector UI in the console.

PR #2566, opened 2022-11-25
ci: fix package.json paths

btaillon:

Paths are no longer relative to the JS file.

### DIFF
--- a/packages/samples/angular/.eslintrc.json
+++ b/packages/samples/angular/.eslintrc.json
@@ -3,5 +3,8 @@
   "parser": "@typescript-eslint/parser",
   "plugins": ["@typescript-eslint"],
   "extends": ["./node_modules/gts"],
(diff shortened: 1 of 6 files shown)


packages/samples/angular/.eslintrc.json
```diff
@@ -3,5 +3,8 @@
   "parser": "@typescript-eslint/parser",
   "plugins": ["@typescript-eslint"],
   "extends": ["./node_modules/gts"],
-  "ignorePatterns": ["**/*.js", "dist"]
+  "ignorePatterns": ["**/*.js", "dist"],
+  "rules": {
+    "node/no-unpublished-import": "off"
```
btaillon, line 8: The root `package.json` has `private: true`, which triggers this error when importing `import {workspacesRoot} from '../../../../scripts/packages.mjs';`.

ThibodeauJF (approved):

PR #2571, opened 2022-11-28
ci: fix invalidate cloudfront getMajorVersion

btaillon:
https://coveord.atlassian.net/browse/KIT-2150

Not sure how I missed this. Merging ASAP because it could break other things.


scripts/packages.mjs
```diff
@@ -45,7 +45,7 @@ export function getPackagePathFromPackageDir(packageDir) {
  * @returns {import('@lerna/package').RawManifest}
  */
 function getPackageManifestFromPackagePath(fullPath) {
-  return resolve(JSON.parse(readFileSync(fullPath).toString()), 'package.json');
```
btaillon, line 48: Not sure why anything at all was able to pass despite this error

PR #2573, opened 2022-11-29
feat(headless): New action creator added to logCopyToClipboard analytics

mmitiche:
### In this PR:

1. I created two new dispatch-able actions for both the search and insight use cases to log the Copy To Clipboard analytics event.
The Copy To Clipboard analytics event is sent when the user clicks on the Copy To Clipboard result action button  on a result.
2. I've also updated coveo.analytics to the latest version.

https://user-images.githubusercontent.com/86681870/204535701-65d2717f-0709-4fa8-8c43-a519fe49ded2.mov

copyToClipboarEvent


ThibodeauJF (approved):
Cool! Maybe you'd like to add a new controller similar to "interactive result" to leverage the new action easily in Atomic/Quantic

packages/headless/src/features/analytics/search-analytics-actions-loader.ts
```diff
@@ -725,6 +727,19 @@ export interface SearchAnalyticsActionCreators {
     void,
     AsyncThunkInsightAnalyticsOptions<StateNeededBySearchAnalyticsProvider>
   >;
+  /**
+   * The event to log when the Copy To Clipboard result action is clicked.
+   *
+   * @param result - The result.
+   * @returns A dispatchable action.
+   */
+  logCopyToClipboard(result: Result): AsyncThunkAction<
```
olamothe, line 736: Do we really need/want to also re-export it here ? The concept of "result action" seems only applicable for the insight panel use case.

olamothe (approved):

btaillon:
The events you added in coveo.analytics are missing their `make*` counterparts, which are needed for Headless v2.

packages/headless/src/features/analytics/search-analytics-actions-loader.ts
mmitiche: indeed! I've just re-checked with the Quantic team, probably we will need this feature for the search use case in the future but for now we only need it for the insight use case, so it would make sense to only add this logic for the insight use case especially that it's easy to add the logic for search when needed.
I'll update my PR accordingly.

mmitiche:
> 



> The events you added in coveo.analytics are missing their `make*` counterparts, which are needed for Headless v2.

Thank you for pointing this out @btaillon ! I would like to understand more what's needed to make this events compatible with Headless v2, is it something that need to be updated in this PR or it's an update that needs to be done in coveo.analytics?
Also what do you mean by "are missing their `make` counterparts"? do you have an example of an event  defined in the proper way needed by Headless v2?

btaillon:
Nevermind, I forgot that this change doesn't affect insight.

PR #2574, opened 2022-11-29
fix(headless): disabled analytics when doNotTrack is on

btaillon:
https://coveord.atlassian.net/browse/KIT-2151

In v1, this isn't strictly needed, since `coveo.analytics` already disables analytics requests on its own, and Headless doesn't happen to include any identifiable information in the `analytics` object of search requests (it includes a clientId, but it's reset after each request).

In v2, this is now needed because search requests include the `actionCause` and `customData` of analytics events in their `analytics` object.

This PR removes the `analytics` object from search requests when doNotTrack is on, which is the same behaviour as when analytics are disabled.

An alternative could be to try to detect which client `coveo.analytics` initialized and disable analytics based on that.

I'm merging this in `v1` since this is non-breaking.

### DIFF
--- a/packages/headless/src/app/engine.ts
+++ b/packages/headless/src/app/engine.ts
@@ -8,15 +8,16 @@ import {
   Middleware,
   Reducer,
 } from '@reduxjs/toolkit';
-import coveoAnalytics from 'coveo.analytics';
 import {Logger} from 'pino';
 import {
   disableAnalytics,
   enableAnalytics,
   updateAnalyticsConfiguration,
+  UpdateAnalyticsConfigurationActionCreatorPayload,
   updateBasicConfiguration,
 } from '../features/configuration/configuration-actions';
 import {SearchParametersState} from '../state/search-app-state';
+import {doNotTrack} from '../utils/utils';
 import {analyticsMiddleware} from './analytics-middleware';
 import {EngineConfiguration} from './engine-configuration';
 import {LoggerOptions} from './logger';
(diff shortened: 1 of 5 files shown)


packages/headless/src/app/engine.ts
```diff
@@ -157,6 +158,9 @@ export function buildEngine<
 
   if (analytics) {
     const {analyticsClientMiddleware, ...rest} = analytics;
+    if (analytics.enabled && coveoAnalytics.donottrack.doNotTrack()) {
```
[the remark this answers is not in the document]
btaillon: If you're asking about the function itself, it should, here's its module:
```ts
import {hasNavigator} from './detector';

const doNotTrackValues = ['1', 1, 'yes', true];

export function doNotTrack(): boolean {
    return (
        hasNavigator() &&
        [
            (<any>navigator).globalPrivacyControl,
            (<any>navigator).doNotTrack,
            (<any>navigator).msDoNotTrack,
            (<any>window).doNotTrack,
        ].some((value) => doNotTrackValues.indexOf(value) !== -1)
    );
}

export default doNotTrack;
```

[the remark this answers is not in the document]
olamothe: What if instead of adding this `if` check here, we modified the initial configuration done at the beginning of this function to force the options `analytics.enabled = false` on the engine when the do not track header is set ?

With an info message in the console that explains that analytics were disabled because of the header.

olamothe (approved):

packages/headless/src/features/configuration/configuration-actions.ts
```diff
@@ -143,8 +144,11 @@ export type AnalyticsRuntimeEnvironment = IRuntimeEnvironment;
 
 export const updateAnalyticsConfiguration = createAction(
   'configuration/updateAnalyticsConfiguration',
-  (payload: UpdateAnalyticsConfigurationActionCreatorPayload) =>
-    validatePayload(payload, {
+  (payload: UpdateAnalyticsConfigurationActionCreatorPayload) => {
+    if (payload.enabled && doNotTrack()) {
```
olamothe, line 148:
```suggestion
    if (doNotTrack()) {
```

olamothe (approved):

packages/headless/src/utils/utils.ts
```diff
@@ -43,3 +43,20 @@ export function omit<T>(key: keyof T, obj: T) {
...
+
+/**
+ * Logic copied from coveo.analytics.
+ */
+export function doNotTrack() {
+  if (typeof navigator === 'undefined') {
+    return false;
+  }
+  return [
+    (<any>navigator).globalPrivacyControl,
```
ThibodeauJF, line 57: we could disable this warning for this file

ThibodeauJF (approved):

packages/headless/src/utils/utils.ts
btaillon: will do that and remove repetition.

PR #2578, opened 2022-11-30
fix(headless): fixed issues when publishing package and updating tags

btaillon:
https://coveord.atlassian.net/browse/KIT-2155

Will merge with the following name to force a version bump (although it will generate a changelog entry):
> fix(headless): fixed issues when publishing package and updating tags

This time, I tested `publish.mjs` locally, which I should've done the first time.

### DIFF
--- a/JenkinsfileStagingRelease
+++ b/JenkinsfileStagingRelease
@@ -15,7 +15,10 @@ node('linux && docker') {
   ])
 
   dockerUtils.withDocker(image: 'node:16', args: '-e HOME=/tmp -e NPM_CONFIG_PREFIX=/tmp/.npm') {
-    
+    stage('Setup') {
+      sh 'npm ci'
+    }
+
     stage('Npm publish') {
       withCredentials([
       string(credentialsId: 'NPM_TOKEN', variable: 'NPM_TOKEN')]) {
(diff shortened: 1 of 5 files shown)


JenkinsfileStagingRelease
```diff
@@ -15,7 +15,10 @@ node('linux && docker') {
   ])
 
   dockerUtils.withDocker(image: 'node:16', args: '-e HOME=/tmp -e NPM_CONFIG_PREFIX=/tmp/.npm') {
-    
+    stage('Setup') {
+      sh 'npm ci'
+    }
```
btaillon, line 20: Copied from the production Jenkinsfile

scripts/exec.mjs
```diff
@@ -44,12 +44,10 @@ export function execute(command, args = []) {
     proc.on('exit', (code) =>
       code === 0
         ? resolve(trimNewline(dataBuffer.toString('utf8')))
-        : reject(
-            JSON.stringify({
-              code,
-              error: trimNewline(errorBuffer.toString('utf8')),
-            })
-          )
```
btaillon, line 52: I recall adding the stringify to ensure that the error would show up in the console when I first made this function. However, since then, I made this function log errors, so there's no purpose to using `JSON.stringify` here.

ThibodeauJF (approved):

louis-bompart (approved):

PR #2581, opened 2022-12-01
feat(headless): refactored field suggestion options

btaillon:
https://coveord.atlassian.net/browse/KIT-2101

The purpose of this Jira was to remove deprecated options, however, I noticed there was also issues with the documentation for field suggestions; field suggestions are tightly tied with facets, and their options are used to register a facet, which is important to know for users if they want to understand the behaviour of field suggestions (especially when field suggestions and a facet share the same facet id).

In this PR, I tested one possible solution to this problem in my opinion:
1. Adding documentation to field suggestions themselves that clarify that they actually initialize a facet.
2. Moving all field suggestions options under "facet" sub-options, since they are all used to register a facet. I think this really clarifies the purpose of all options nested under it and emphasizes the fact that field suggestions rely on facets. If we have to add options to field suggestions, this will also clarify that which options aren't used to register a facet.

Other alternatives would be:
* Alternative A: Not changing the options hierarchy at all, and only doing documentation.
* Alternative B: Moving `field` (and maybe `facetSearch`) out of the `facet` sub-options to make the hierarchy less deep and the options prettier. This would however force us to re-declare all facet options in a new interface that is specific to field suggestions, just like it was prior to this change, since we (presumably) can't use `Omit` with the doc parser.

# Deprecations
## Changed
* Removed the `FieldSuggestionsFacetSearchOptions` interface.
  * `FacetSearchOptions` is used instead.
* Removed the `CategoryFieldSuggestionsFacetSearchOptions` interface.
  * `CategoryFacetSearchOptions` is used instead.
* Refactored `FieldSuggestionsOptions`, nesting all its options under `facet: FacetOptions`.
  * `buildFieldSuggestions(engine, { options: { field: 'filetype' } });` would become `buildFieldSuggestions(engine, { options: { facet: { field: 'filetype' } } });`
* Refactored `CategoryFieldSuggestionsOptions`, nesting all its options under `facet: FacetOptions`.
  * `buildCategoryFieldSuggestions(engine, { options: { field: 'category' } });` would become `buildCategoryFieldSuggestions(engine, { options: { facet: { field: 'category' } } });`
* Removed the `delimitingCharacter` option from `FieldSuggestions`
  * It didn't do anything.
* Removed the `delimitingCharacter` option from `CategoryFieldSuggestions`
  * It didn't do anything.

### DIFF
--- a/packages/headless/src/controllers/field-suggestions/category-facet/headless-category-field-suggestions.ts
+++ b/packages/headless/src/controllers/field-suggestions/category-facet/headless-category-field-suggestions.ts
@@ -1,4 +1,4 @@
-import {CoreEngine} from '../../..';
+import {CoreEngine} from '../../../app/engine';
 import {
   categoryFacetSet,
   configuration,
@@ -9,7 +9,6 @@ import {SearchEngine} from '../../../app/search-engine/search-engine';
 import {SearchThunkExtraArguments} from '../../../app/search-thunk-extra-arguments';
 import {registerCategoryFacet} from '../../../features/facets/category-facet-set/category-facet-set-actions';
 import {defaultCategoryFacetOptions} from '../../../features/facets/category-facet-set/category-facet-set-slice';
-import {CategoryFacetSortCriterion} from '../../../features/facets/category-facet-set/interfaces/request';
 import {
   CategoryFacetSearchSection,
   CategoryFacetSection,
@@ -17,12 +16,12 @@ import {
   SearchSection,
 } from '../../../state/state-sections';
 import {loadReducerError} from '../../../utils/errors';
-import {omit} from '../../../utils/utils';
 import {
   buildController,
   Subscribable,
 } from '../../controller/headless-controller';
 import {determineFacetId} from '../../core/facets/_common/facet-id-determinor';
+import {CategoryFacetOptions} from '../../facets/category-facet/headless-category-facet';
 import {buildCategoryFacetSearch} from '../../facets/category-facet/headless-category-facet-search';
 
 export interface CategoryFieldSuggestionsValue {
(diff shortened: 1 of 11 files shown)


packages/headless/src/controllers/field-suggestions/category-facet/headless-category-field-suggestions.ts
```diff
@@ -119,89 +118,11 @@ export interface CategoryFieldSuggestions extends Subscribable {
...
-   *
-   * @defaultValue `true`
-   */
-  filterByBasePath?: boolean;
-
-  /**
-   * Whether to exclude the parents of folded results when estimating the result count for each field suggestion.
-   *
-   * @defaultValue `true`
+   * The options used to register the category facet used by the field suggestions controller.
```
jpmarceau, line 123:
```suggestion
   * The options used to register the category facet used under the hood by the field suggestions controller.
```

packages/headless/src/controllers/field-suggestions/category-facet/headless-category-field-suggestions.ts
```diff
@@ -213,6 +134,10 @@ export interface CategoryFieldSuggestionsProps {
 
 /**
  * Creates a `CategoryFieldSuggestions` controller instance.
+ *
+ * This controller initializes a category facet, but exposes state and methods that are relevant for suggesting fields values based on a query.
+ * It's important not to initialize a category facet with the same `facetId` but different options, because only the options of the controller which is built first will be taken.
```
jpmarceau, line 139:
```suggestion
 * This controller initializes a category facet under the hood, but exposes state and methods that are relevant for suggesting field values based on a query.
 * It's important not to initialize a category facet with the same `facetId` but different options, because only the options of the controller which is built first will be taken into account.
```

packages/headless/src/controllers/field-suggestions/facet/headless-field-suggestions.ts
```diff
@@ -212,6 +139,10 @@ export interface FieldSuggestionsProps {
 
 /**
  * Creates a `FieldSuggestions` controller instance.
+ *
+ * This controller initializes a facet, but exposes state and methods that are relevant for suggesting fields values based on a query.
+ * It's important not to initialize a facet with the same `facetId` but different options, because only the options of the controller which is built first will be taken.
```
jpmarceau, line 144:
```suggestion
 * This controller initializes a facet under the hood, but exposes state and methods that are relevant for suggesting field values based on a query.
 * It's important not to initialize a facet with the same `facetId` but different options, because only the options of the controller which is built first will be taken into account.
```

jpmarceau (approved):
👍

jpmarceau:
Just to make sure I understand correctly:
You were initially planning on removing `delimitingCharacter`, `injectionDepth`, `hasBreadcrumbs` and `facetSearch`, but it turns out the only one you'll remove is `delimitingCharacter`. The others are kept, but moved, like all other facet options, under `facet: FacetOptions`?

Am I understanding this correctly?

btaillon:
> Just to make sure I understand correctly: You were initially planning on removing `delimitingCharacter`, `injectionDepth`, `hasBreadcrumbs` and `facetSearch`, but it turns out the only one you'll remove is `delimitingCharacter`. The others are kept, but moved, like all other facet options, in an actual `facet: FacetOptions`?
> 
> Am I understanding this correctly? Snapshot from my headless v2 PR:
> 
> Screen Shot 2022-12-01 at 8 03 03 PM

Yes exactly. The reason I was gonna remove those options initially was because I expected at some point that we'd separate facets from field suggestions, those options would become useless. However, we had discussions and field suggestions are gonna keep depending on facets.

packages/samples/headless-react/package.json
```diff
@@ -30,7 +30,7 @@
   },
   "scripts": {
     "dev": "react-scripts start",
-    "build": "tsc --noEmit && tsc --module commonjs --noEmit",
+    "build": "echo 'no'",
```
ThibodeauJF, line 33: hmm

ThibodeauJF (approved):

packages/headless/src/controllers/field-suggestions/facet/headless-field-suggestions.ts
```diff
@@ -20,13 +19,13 @@ import {
...
 } from '../../../state/state-sections';
 import {loadReducerError} from '../../../utils/errors';
-import {omit} from '../../../utils/utils';
 import {
   buildController,
   Subscribable,
 } from '../../controller/headless-controller';
 import {determineFacetId} from '../../core/facets/_common/facet-id-determinor';
 import {buildFacetSearch} from '../../core/facets/facet-search/specific/headless-facet-search';
+import {FacetOptions} from '../../facets/facet/headless-facet-options';
```
jpmarceau, line 28: As I was working on the migration docs for the fieldSuggestionOptions removal, I noticed that the `FacetOptions` annotations don't appear in the parsed_doc.json that we use to generate docs.
So we don't have anything about that currently [in the reference docs](https://docs.coveo.com/en/headless/latest/reference/search/controllers/facet/#facetpropsfacetoptions).
That's a bit annoying, because it means that by removing the FieldsSuggestionFacetSearchOptions, we also lose their docs.

It seems fine [for category facets](https://docs.coveo.com/en/headless/latest/reference/search/controllers/category-facet/#categoryfacetoptions), somehow. 

I admit this is a bit mysterious to me, but if there's a simple fix I'm not seeing it would be cool.
Otherwise I can create a separate Jira for later?

jpmarceau: 👍 
Created https://coveord.atlassian.net/browse/KIT-2174

PR #2583, opened 2022-12-05
fix(atomic): fixed insight search section grid behaviour

nathanlb:
The grid behaviour was yielding some weird looks with certain combinations of tabs, toggles, searchbox.
- The toggles would wrap to the next row when no tabs were present to take an entire row.
- The number of columns would not adjust to the number of elements present and empty rows still take up space so the right spacing wouldn't be correct.

FYI I can't say i'm a `display: grid` expert so let me know if you see a better way that doesn't involve as much javascript

Before
Screen Shot 2022-12-04 at 6 19 23 PM
Screen Shot 2022-12-05 at 1 15 53 PM

After
Screen Shot 2022-12-05 at 12 13 35 PM
Screen Shot 2022-12-05 at 12 15 02 PM
Screen Shot 2022-12-05 at 12 14 49 PM
Screen Shot 2022-12-05 at 12 13 24 PM
Screen Shot 2022-12-05 at 12 14 38 PM

### DIFF
--- a/packages/atomic/src/components/insight/atomic-insight-layout/insight-layout.ts
+++ b/packages/atomic/src/components/insight/atomic-insight-layout/insight-layout.ts
@@ -43,6 +43,7 @@ export function buildInsightLayout(element: HTMLElement, widget: boolean) {
       padding-left: 1.5rem;
       padding-right: 1.5rem;
       box-sizing: border-box;
+      overflow-x: hidden;
       ${!hasTabs ? 'padding-bottom: 1.5rem;' : ''}
     }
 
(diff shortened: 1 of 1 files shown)


ThibodeauJF (approved):
Implementation with the grid looks good, perhaps a flex implementation would be a tad simpler for the search section (leveraging flex-grow for the search bar)

nathanlb:
> Implementation with the grid looks good, perhaps a flex implementation would be a tad simpler for the search section (leveraging flex-grow for the search bar)

Yeah i prefer flex but only because i know it better. If you won't think less of me for going off the grid then we can do that. I pushed an update.

btaillon (approved):

PR #2585, opened 2022-12-06
fix(atomic): adapt ArrayProp for v2 changes (& other small deprecations)

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2161


packages/atomic/src/components.d.ts
```diff
@@ -40,7 +40,7 @@ export namespace Components {
         /**
           * The base path shared by all values for the facet.  Specify the property as an array using a JSON string representation: ```html  <atomic-category-facet base-path='["first value", "second value"]' ></atomic-category-facet> ```
          */
-        "basePath"?: string[];
+        "basePath": string[] | string;
```
btaillon, line 43: Is there some way we could only allow javascript/JSX frameworks to pass an array? I prefer not to expose a type to a framework for which it's useless just because we need it internally or for some unrelated framework.

olamothe (approved):

PR #2597, opened 2022-12-08
feat(headless): add src to use quickview in iframe

nathanlb:
Added `srcPath` to the quickview state to be used optionally if the implementation requires usage of the GET endpoint rather than the POST.

Quantic was also updated to leverage the change.

Misc changes:
- Fixed a typo in the readme. It formatted the whole file on save... is my linting setup wrong? I'm using `Prettier ESLint`
- Added the `engines` config to headless because, while attempting to use `URL` I got an error saying we needed to be compatible with Node v8. Looks like the root config doesn't get inherited by the packages.
- Speaking of `URL` apparently it looks like React Native has [some differences](https://www.npmjs.com/package/react-native-url-polyfill#user-content-why-do-we-need-this) with its URL implementation to I created a util to replace it for my purposes.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,9 +1,5 @@
 {
-  "ignorePatterns": [
-    "node_modules",
-    "stencil-generated",
-    "dist"
-  ],
+  "ignorePatterns": ["node_modules", "stencil-generated", "dist"],
   "env": {
     "jest": true,
     "es6": true
@@ -32,7 +28,9 @@
         "@typescript-eslint/no-empty-interface": [
           "error",
           {"allowSingleExtends": true}
-        ]
+        ],
+        "node/no-unsupported-features/node-builtins": "off",
+        "node/no-unsupported-features/es-builtins": "error"
       }
     },
     {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,9 +28,7 @@
         "@typescript-eslint/no-empty-interface": [
           "error",
           {"allowSingleExtends": true}
-        ],
-        "node/no-unsupported-features/node-builtins": "off",
-        "node/no-unsupported-features/es-builtins": "error"
+        ]
       }
     },
     {
--- a/packages/headless/package.json
+++ b/packages/headless/package.json
@@ -68,5 +68,9 @@
     "@microsoft/tsdoc": "0.14.1",
     "jest-fetch-mock": "3.0.3",
     "ts-node": "10.9.1"
+  },
+  "engines": {
+    "node": ">=16.0.0",
+    "npm": ">=8.6.0"
   }
 }
--- a/packages/headless/src/api/search/html/html-api-client.ts
+++ b/packages/headless/src/api/search/html/html-api-client.ts
@@ -44,6 +44,9 @@ export const buildSrcPath = (req: HtmlRequest, path: string) => {
       `${req.requestedOutputSize}`
     );
   }
+  if (req.visitorId !== undefined) {
+    url.searchParams.append('visitorId', `${req.visitorId}`);
+  }
   return url.href;
 };
 
--- a/packages/headless/src/api/search/html/html-api-client.ts
+++ b/packages/headless/src/api/search/html/html-api-client.ts
@@ -33,8 +33,17 @@ export const buildSrcPath = (req: HtmlRequest, path: string) => {
   url.searchParams.append('access_token', req.accessToken);
   url.searchParams.append('organizationId', req.organizationId);
   url.searchParams.append('uniqueId', req.uniqueId);
-  if (req.q) {
+  if (req.q !== undefined) {
     url.searchParams.append('q', req.q);
+  }
+  if (req.enableNavigation !== undefined) {
+    url.searchParams.append('enableNavigation', `${req.enableNavigation}`);
+  }
+  if (req.requestedOutputSize !== undefined) {
+    url.searchParams.append(
+      'requestedOutputSize',
+      `${req.requestedOutputSize}`
+    );
   }
   return url.href;
 };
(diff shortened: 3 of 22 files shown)


.eslintrc.json
```diff
@@ -32,7 +28,9 @@
         "@typescript-eslint/no-empty-interface": [
           "error",
           {"allowSingleExtends": true}
-        ]
+        ],
+        "node/no-unsupported-features/node-builtins": "off",
```
nathanlb, line 32: apparently this is deprecated and it wasn't properly using our defined minimum node version. eslint was mad at me becuase `URL` is node >= 10 which should be ok.

btaillon: instead I think we should make sure that it ESLint the correct version

packages/headless/src/api/search/html/html-api-client.ts
```diff
@@ -27,6 +27,29 @@ export interface HtmlAPIClientOptions {
...
+    url.searchParams.append(
+      'requestedOutputSize',
+      `${req.requestedOutputSize}`
+    );
+  }
+  if (req.visitorId !== undefined) {
+    url.searchParams.append('visitorId', `${req.visitorId}`);
+  }
+  return url.href;
+};
```
btaillon, line 51: I'm not sure if we should use the `URL` object, since we aim to make Headless as platform-agnostic as possible and it looks like React Native has [some differences](https://www.npmjs.com/package/react-native-url-polyfill#user-content-why-do-we-need-this) with its `URL` implementation.

packages/headless/src/controllers/core/quickview/headless-core-quickview.ts
```diff
@@ -31,11 +39,18 @@ export interface QuickviewOptions {
    * The maximum preview size to retrieve, in bytes. By default, the full preview is retrieved.
    */
   maximumPreviewSize?: number;
+
+  /**
+   * Whether to only update the `srcPath` attribute when using `fetchResultContent` rather than updating `content`.
+   * Use this if you are using an iframe with `state.srcPath` as the source url.
+   */
+  onlySrcPath?: boolean;
```
btaillon, line 47: Since the purpose of the `srcPath` in the first place is to be used inside of an iframe, I think we should include that in the name. Otherwise, I think we should remove `src` from the name somehow.

For instance:
* `iFrameUrl`/`onlyIFrameUrl`
* `iFramePath`/`onlyIFramePath`
* `iFrameSrc`/`onlyIFrameSrc`
* `downloadUrl`/`onlyDownloadUrl`
* `contentUrl`/`onlyContentUrl`
* etc

btaillon (approved):

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -280,6 +289,10 @@ export default class QuanticResultQuickview extends LightningElement {
     }
   }
 
+  onIframeLoaded() {
```
btaillon, line 292: nitpick: most examples I found online capitalize `iframe` like two words (Inline Frame -> `IFrame` in pascal case / `iFrame` in camel case).

packages/headless/src/api/search/html/html-api-client.ts
nathanlb: Hmmm would you prefer straight up string interpolation? That felt sketchy.

ThibodeauJF (approved):

packages/headless/src/api/search/html/html-api-client.ts
nathanlb: I added a utility class to avoid using URL. Let me know what you think.

packages/headless/package.json
```diff
@@ -68,5 +68,9 @@
     "@microsoft/tsdoc": "0.14.1",
     "jest-fetch-mock": "3.0.3",
     "ts-node": "10.9.1"
+  },
+  "engines": {
+    "node": ">=16.0.0",
+    "npm": ">=8.6.0"
```
louis-bompart, line 74: @btaillon , are we on npm 9 yet or not?
(lockfile shenanigans)

packages/headless/src/utils/url-utils.ts
```diff
@@ -0,0 +1,35 @@
+export class URLPath {
```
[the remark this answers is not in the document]
nathanlb: Lol I was 100% with you but @btaillon doesn't want me to use URL because maybe react-native has some different behaviour. That makes sense too but I was asked to come up with a util so here it is. We're not going back now ;)

packages/headless/package.json
nathanlb: What I was using was `URL` though so i technically don't need it anymore. But it felt weird to leave it as-is.

erocheleau (approved):
Looks good to me! Nice one

For fun you tested with local dev server and also community?
Did you test with Locker/LWS?

packages/headless/src/api/search/html/html-api-client.ts
```diff
@@ -27,6 +28,26 @@ export interface HtmlAPIClientOptions {
...
+  if (req.enableNavigation !== undefined) {
+    url.addParam('enableNavigation', `${req.enableNavigation}`);
+  }
+  if (req.requestedOutputSize !== undefined) {
+    url.addParam('requestedOutputSize', `${req.requestedOutputSize}`);
+  }
+  if (req.visitorId !== undefined) {
+    url.addParam('visitorId', `${req.visitorId}`);
+  }
+  return url.href;
```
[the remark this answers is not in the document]
nathanlb: Ah hmm interesting... I guess the best we can do is log it.

nathanlb:
I tested in community with locker and it works fine. LWS quantic as a whole doesn't work ATM so didn't test with that.

PR #2598, opened 2022-12-08
fix(atomic): added warning when declaring a template with section and non-section elements

btaillon:
https://coveord.atlassian.net/browse/KIT-2081

**Caveat**: the error is only logged when the template is declared using HTML. Results rendered using Javascript or using React/Angular do not log an error.

This took me much longer than expected because I came up with a way to refactor the rendering of result template, which I wanted to leverage to validate templates when rendered regardless of the framework. I stashed that solution and will re-explore it when we refactor the result list.

<details>
<summary>Details on the refactor I stashed</summary>

My refactor would've eliminated lots of repetition in `atomic-result`/`atomic-insight-result`/`atomic-recs-result` and would've completely removed the `string` type for templates. Rendering and validation of result templates would've been handled by a functional component which takes `{ content?: ParentNode } | { renderingFunction?: ResultRenderingFunction }` as a prop. However, this solution broke in React since `createRoot(...).render()` returns a different output than `renderToString`.
</details>

### DIFF
--- a/packages/atomic/cypress/e2e/result-list/result-template.cypress.ts
+++ b/packages/atomic/cypress/e2e/result-list/result-template.cypress.ts
@@ -1,5 +1,9 @@
+import {tableElementTagName} from '../../../src/components/search/atomic-table-result/table-element-utils';
 import {generateComponentHTML, TestFixture} from '../../fixtures/test-fixture';
-import {assertContainsComponentError} from '../common-assertions';
+import {
+  assertConsoleWarning,
+  assertContainsComponentError,
+} from '../common-assertions';
 import {
   addFieldValueInResponse,
   addResultList,
@@ -9,6 +13,7 @@ import {
 import {
   resultListComponent,
   ResultListSelectors,
+  resultSectionTags,
 } from './result-list-selectors';
 import {addResultTable} from './result-table-actions';
 import {ResultTableSelectors} from './result-table-selectors';
@@ -59,6 +64,42 @@ describe('Result Template Component', () => {
     });
 
     assertContainsComponentError(ResultTemplateSelectors, true);
+  });
+
+  describe('when it has both section and non-section elements', () => {
+    beforeEach(() => {
+      new TestFixture()
+        .with(
+          addResultList(
+            buildTemplateWithoutSections([
+              new Text('hello'),
+              generateComponentHTML(resultSectionTags.actions),
+            ])
+          )
+        )
+        .init();
+    });
+
+    assertConsoleWarning(true);
+  });
+
+  describe('when it has strictly section and table-element elements', () => {
+    beforeEach(() => {
+      new TestFixture()
+        .with(
+          addResultList(
+            buildTemplateWithoutSections([
+              generateComponentHTML(tableElementTagName),
+              new Text('    \n    '),
+              generateComponentHTML(resultSectionTags.actions),
+            ])
+          )
+        )
+        .init();
+    });
+
+    assertContainsComponentError(ResultTemplateSelectors, false);
+    assertConsoleWarning(false);
   });
 
   describe('with a visual section', () => {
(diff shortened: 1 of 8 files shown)


packages/atomic/cypress/e2e/result-list/result-template.cypress.ts
```diff
@@ -61,6 +66,42 @@ describe('Result Template Component', () => {
...
+  });
+
+  describe('when it has strictly section and table-element elements', () => {
+    beforeEach(() => {
+      new TestFixture()
+        .with(
+          addResultList(
+            buildTemplateWithoutSections([
+              generateComponentHTML(tableElementTagName),
+              new Text('    \n    '),
```
btaillon, line 93: HTML templates tend to contain indentation and newline characters as text nodes. We don't want to log an error for those.

packages/atomic/src/components/common/layout/sections.ts
```diff
@@ -7,20 +9,26 @@ export const resultSectionTags = [
   'atomic-result-section-emphasized',
   'atomic-result-section-excerpt',
   'atomic-result-section-bottom-metadata',
-] as const;
+] as const);
+
+type SetValueType<T> = T extends Set<infer U> ? U : never;
```
btaillon, line 14: this is just a quick helper I made, it's only needed here. I can extract it to a util file if you prefer.

packages/atomic/src/components/common/layout/sections.ts
```diff
@@ -1,4 +1,6 @@
-export const resultSectionTags = [
+import {isElementNode} from '../../../utils/utils';
+
+export const resultSectionTags = new Set([
```
btaillon, line 3: I prefer making this a `set` since its purpose is to validate whether a tag name is a section.

jpmarceau (approved):
👍 

I'll specify this warning in the v2 migration docs.

ThibodeauJF (approved):
Any warnings appearing with our current examples? I know it was the case with "multi value  text"

btaillon:
> Any warnings appearing with our current examples? I know it was the case with "multi value text"

Just tested, and there was a warning concerning the folding children section since it isn't part of the section tags list. I added it to the list.

packages/atomic/src/utils/utils.ts
```diff
@@ -56,13 +56,28 @@ export function parseHTML(string: string) {
   return new window.DOMParser().parseFromString(string, 'text/html');
 }
 
+export function isElementNode(node: Node): node is Element {
```
olamothe, line 59: I think quite a bit of code in this PR could benefit from good ol' unit test. Since a lot is pure function !

packages/atomic/src/components/common/result-templates/result-template-common.tsx
```diff
@@ -14,6 +17,36 @@ interface ResultTemplateCommonProps {
...
+  if (
+    isElementNode(node) &&
+    node.tagName.toLowerCase() === tableElementTagName
+  ) {
+    return 'table-column-definition';
+  }
+  return 'other';
+}
+
+function separateSectionAndNonSectionNodes(nodes: NodeList) {
```
olamothe, line 36: Suggestion:
```suggestion
function groupNodeByType(nodes: NodeList) {
```

Which would return an object where the keys are the `nodeType`.
```
{
'section': [...nodes],
'other': [...nodes],
'metadata': [...nodes],
'table-column-definition': [...nodes]
}
```

And then you could just deconstruct at the caller level.
Would that work ?

olamothe (approved):

packages/atomic/src/components/common/result-templates/result-template-common.tsx
btaillon: Yes! Will do that.

PR #2600, opened 2022-12-08
fix(atomic): add density and image size to key prop for atomic-result

olamothe:
https://coveord.atlassian.net/browse/KIT-2093


ThibodeauJF (approved):

packages/atomic/src/components/common/result-list/list-display-results.tsx
```diff
@@ -7,7 +7,9 @@ export const ListDisplayResults: FunctionalComponent<ResultListDisplayProps> = (
 ) =>
   props.getResultListState().results.map((result, index) =>
     props.renderResult({
-      key: props.getResultId(result),
+      key: `${props.getResultId(
```
ThibodeauJF, line 10: If im not mistaken getResultId is used just for this purpose, this could be an upgrade to that prop instead of a copypaste

PR #2601, opened 2022-12-09
doc(atomic): added note about delimiting-character usage

srosenberg-coveo:
https://coveord.atlassian.net/browse/DOC-11297

### DIFF
--- a/packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
+++ b/packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
@@ -147,6 +147,8 @@ export class AtomicCategoryFacet
     'occurrences';
   /**
    * The character that separates values of a multi-value field.
+   *
+   * Note: If you use the [recommended formatting](https://docs.coveo.com/en/atomic/latest/reference/components/atomic-category-facet/#usage-notes) for the associated multi-value field, you must set this value to `|` or the facet won't display properly.
    */
   @Prop({reflect: true}) public delimitingCharacter = ';';
   /**


ThibodeauJF (approved):
Thanks!

packages/atomic/src/components.d.ts
```diff
@@ -42,7 +42,7 @@ export namespace Components {
          */
         "basePath"?: string | string[];
         /**
-          * The character that separates values of a multi-value field.
+          * The character that separates values of a multi-value field.  Note: If you use the [recommended formatting](https://docs.coveo.com/en/atomic/latest/reference/components/atomic-category-facet/#usage-notes) for the associated multi-value field, you must set this value to `|` or the facet won't display properly.
```
btaillon, line 45: Instead of "recommended" I think we should say "example".

btaillon:
In the example itself where `|` is used as the delimiting character, I would imply that `|` is being used as the delimiting character.

E.g.:
```
A Category Facet requires a formatted multi-value field that’s formed in a hierarchy. For instance, if the `delimiting-character` is `|`, it may look like this: `clothing; clothing|sale; clothing|sale|shoes; clothing|sale|shoes|kids; clothing|sale|shoes|kids|sneakers;` [...]
```

srosenberg-coveo:
> In the example itself where `|` is used as the delimiting character, I think we should imply that `|` is being used as the delimiting character.

I'll have to do this in a separate PR because this part of the article lives in the Docs site repo.

jpmarceau (approved):
👍 

I second Benjamin's suggestion regarding the usage note update 🙂

PR #2603, opened 2022-12-09
feat(quantic): added quantic-result-text component

nathanlb:
Added component for displaying a field value in a result template.

Screen Shot 2022-12-09 at 5 30 16 PM

We're planning on adding components for numeric, date and multivalue fields too so ideally those could wrap this component and leverage the `formattingFunction` prop.

Also added our [Bueno](https://github.com/coveo/ui-kit/tree/master/packages/bueno) validation library as a dependency. It's the recommended way to validate inputs and especially relevant in a javascript environment like LWC. Given I wanted to have some validation on the component I added bueno which we can use going forward.

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticResultText/__tests__/quanticResultText.test.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultText/__tests__/quanticResultText.test.js
@@ -34,7 +34,16 @@ function createTestComponent(options = defaultOptions) {
   return element;
 }
 
+// Helper function to wait until the microtask queue is empty.
+function flushPromises() {
+  // eslint-disable-next-line @lwc/lwc/no-async-operation
+  return new Promise((resolve) => setTimeout(resolve, 0));
+}
+
 describe('c-quantic-result-text', () => {
+  let mockedConsoleError;
+  let isStringMock;
+
   function cleanup() {
     // The jsdom instance is shared across test cases in a single file so reset the DOM
     while (document.body.firstChild) {
@@ -43,14 +52,26 @@ describe('c-quantic-result-text', () => {
     jest.clearAllMocks();
   }
 
+  beforeEach(() => {
+    mockedConsoleError = jest.fn();
+    isStringMock = jest.fn().mockReturnValue(true);
+    console.error = mockedConsoleError;
+    // @ts-ignore
+    global.Bueno = {
+      isString: isStringMock,
+    };
+  });
+
   afterEach(() => {
     cleanup();
   });
 
   describe('when no result is given', () => {
-    it('should show an error', () => {
+    it('should show an error', async () => {
       // @ts-ignore
       const element = createTestComponent({field: exampleField});
+      await flushPromises();
+
       const errorMessage = element.shadowRoot.querySelector(errorSelector);
 
       expect(errorMessage).not.toBeNull();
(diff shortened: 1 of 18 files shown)


packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.js
```diff
@@ -0,0 +1,78 @@
...
+    // @ts-ignore
+    return this.result.raw[this.field];
+  }
+
+  /**
+   * Whether the given result has a value for the given field
+   * @returns {boolean}
+   */
+  get hasFieldValue() {
+    return !!this.fieldValue.length;
```
lbergeron, line 65: This statement will fail if the value is not a string. If, for some reason, the field value is not of the expected type, the component should fail gracefully. It could raise a warning and make sure `hasFieldValue` returns `false` in such case.

nathanlb: You're right. I'm in the process of adding the bueno library as a quantic dependency to do better input validation. that commit is coming soon.

nathanlb: and some optional chaining couldn't hurt

packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.js
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * The function used to format the displayed value.
+   * @api
+   * @type {Function}
+   * @param {string} value
+   * @returns {string}
+   */
+  @api formattingFunction;
+
+  error;
```
mmitiche, line 41: Adding a type for this would be nice.

packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.js
```diff
@@ -0,0 +1,85 @@
...
+   * The value of the given result field.
+   * @returns {string | undefined}
+   */
+  get fieldValue() {
+    // @ts-ignore
+    return this.field ? this.result?.raw[this.field] : undefined;
+  }
+
+  /**
+   * Whether the given result has a value for the given field
```
mmitiche, line 68:
```suggestion
   * Whether the given result has a value for the given field.
```

packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.html
```diff
@@ -0,0 +1,13 @@
+<template>
+    <div class="result-text slds-grid slds-grid_vertical-align-center">
```
mmitiche, line 2: I'm not sure if the `result-text` CSS class is really needed here, I don't see it in the CSS file or in the tests.

packages/quantic/force-app/main/default/lwc/quanticResultText/__tests__/quanticResultText.test.js
```diff
@@ -0,0 +1,131 @@
...
+
+  describe('when no result is given', () => {
+    it('should show an error', async () => {
+      // @ts-ignore
+      const element = createTestComponent({field: exampleField});
+      await flushPromises();
+
+      const errorMessage = element.shadowRoot.querySelector(errorSelector);
+
+      expect(errorMessage).not.toBeNull();
```
mmitiche, line 77: Can we also check if the console.error method was called here? Something like:
```
expect(console.error).toHaveBeenCalledTimes(1);
```

mmitiche (approved):
Looks good to me, well done!

packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.html
nathanlb: no honestly i thought i would use it then left it in as a potential hook point for a selector or something. I'll remove it though.

lbergeron (approved):
It looks good to me. Were you able to validate whether the package size is still ok?

ThibodeauJF (approved):

PR #2605, opened 2022-12-12
feat(headless): add support for attached results feature

erocheleau:
[SFINT-4478]: 
Added a state in headless for the attached results.
Added a new controlled for the Insight Engine to handle an AttachToCase component.

The TL;DR design of the feature,
1. the original state is empty
2. something will setAttachedResults to be the list of currently attached results coming from CRM for example,
3. after that the AttachResultsController will let you check if a result is attached, attach new results, or detach results.

Attaching results and Detaching results only update the current state, the actual handling or storing attachedResults somewhere in a CRM need to be implemented on the integration side.

In this video you can see the feature in action.

https://user-images.githubusercontent.com/2488155/207098517-2d856582-fb46-46d7-aa0a-1bc511018623.mp4

[SFINT-4478]: https://coveord.atlassian.net/browse/SFINT-4478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/attached-results/attached-results-slice.ts
```diff
@@ -0,0 +1,40 @@
...
+      .addCase(detachResult, (state, action) => {
+        state.results = state.results.filter((result) => {
+          const isPermanentIdEqual =
+            result?.permanentId !== undefined &&
+            result?.permanentId === action.payload?.result?.permanentId;
+          const isUriHashEqual =
+            result?.uriHash !== undefined &&
+            result?.uriHash === action.payload?.result?.uriHash;
+          const isCaseIdEqual = result.caseId === action.payload.result.caseId;
+          return !isCaseIdEqual || (!isPermanentIdEqual && !isUriHashEqual);
```
btaillon, line 36: I'm not sure I understand where the attached results are originally gonna be coming from. Could we assume that, like the Search API, it has a `uniqueId`?

Also, I think this logic should be extracted into a `attachedResultsEqual` function.

erocheleau: The attached results are coming from a record in a CRM (for example Salesforce)
And the goal is to compare them sometimes with results coming from the SearchAPI.
It's confusing but these are two different structures :(

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      uriHash: undefined,
+    });
+    engine.state.attachedResults.results.push(mockAttachedResult);
+
+    const mockSearchResult = buildMockResult({
+      raw: buildMockRaw({
+        permanentid: testPermanentId,
+      }),
+    });
+    expect(attachToCase.isAttached(mockSearchResult)).toBeTruthy();
```
nathanlb, line 68: toBe(true) or toBeTruthy(), that is the question...

more seriously though given it's a boolean I would recommend `toBe(true)` and `toBe(false)` as it's more specific than truthy/falsy.

packages/headless/src/features/attached-results/attached-results-state.ts
```diff
@@ -0,0 +1,24 @@
+export interface AttachedResult {
+  articleLanguage?: string;
```
nathanlb, line 2: Are these values all stored in the salesforce CoveoAttachedResult record?

packages/headless/src/features/attached-results/attached-results-slice.ts
```diff
@@ -0,0 +1,54 @@
...
+      .addCase(setAttachedResults, (state, action) => {
+        const {results: attachedResults, message} = action.payload;
+
+        if ('attachedResults' in state && state['results']?.length > 0) {
+          return;
+        }
+
+        state['results'] = attachedResults;
+        state['message'] = message;
+      })
```
nathanlb, line 26: why not use isNullOrUndefined since it's imported?

packages/headless/src/features/attached-results/attached-results-slice.ts
```diff
@@ -0,0 +1,54 @@
...
+  getAttachedResultsInitialState(),
+  (builder) => {
+    builder
+      .addCase(setAttachedResults, (state, action) => {
+        const {results: attachedResults, message} = action.payload;
+
+        if ('attachedResults' in state && state['results']?.length > 0) {
+          return;
+        }
+
```
nathanlb, line 23: no need for optional chaining safety on `result`?

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.ts
```diff
@@ -0,0 +1,129 @@
...
+  props: AttachToCaseProps = {}
+): AttachToCase {
+  if (!loadAttachedResultsReducers(engine)) {
+    throw loadReducerError;
+  }
+
+  const controller = buildController(engine);
+  const {dispatch} = engine;
+
+  validateInitialState(
```
nathanlb, line 81: Are you setting the initial state anywhere? Looks like your just validating it.
A common pattern in headless controllers with an initial state is to "register" them whent he controller is built.
You can see [headless-recent-result-list](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/controllers/recent-results-list/headless-recent-results-list.ts) as an example. 
I see you already have an action called setAttachedResults. My suggestion would be to rename it to `registerAttachedResults` to be a bit more consistend and dispatch the action right after validating the initial state. On the UI side there won't be any need to load and dispatch the action to register the exisiting results.

nathanlb (commented):
Looks real nice! Just a few things I'd like your view on.

packages/headless/src/features/attached-results/attached-results-actions.ts
```diff
@@ -0,0 +1,91 @@
...
+    name: nonEmptyString,
+    permanentId: nonEmptyString,
+    resultUrl: nonEmptyString,
+    source: nonEmptyString,
+    title: requiredNonEmptyString,
+    uriHash: nonEmptyString,
+  },
+});
+
+export const setAttachedResults = createAction(
```
mmitiche, line 50: Don't we need to validate that at least one of the two fields, `permanentId` and `uriHash `, is present in the results  we try to attach with this action?

packages/headless/src/features/attached-results/attached-results-slice.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    const action = setAttachedResults({
+      results: [],
+      message: 'foo',
+    });
+    const finalState = attachedResultsReducer(state, action);
+    expect(finalState).toStrictEqual(action.payload);
+    expect(finalState.results.length).toEqual(0);
+  });
+
+  it('#setAttachedResults correctly adds the result as attach so we can find it later', () => {
```
mmitiche, line 47: I can't see the difference between this test and the `#setAttachedResults sets the state #results correctly` test, can you please elaborate?

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const mockSearchResult = buildMockResult({
+      raw: buildMockRaw({
+        permanentid: otherPermanentId,
+      }),
+    });
+    expect(attachToCase.isAttached(mockSearchResult)).toBeFalsy();
+  });
+
+  it('calling #attach should trigger the attachResult action with the correct payload', () => {
```
mmitiche, line 104: Adding a test to check that the analytics action was triggered as well would be nice.

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -65,11 +53,35 @@ export default class QuanticResultAction extends LightningElement {
...
+  }
 
-  connectedCallback() {
-    this.setSelected(this.selected);
-    this.setLoading(this.loading);
+  set selected(isSelected) {
+    this._selected = isSelected;
+  }
 
+  /**
```
mmitiche, line 71: We'll need to update the unit tests of this component to reflect these changes, I can help with this!

mmitiche (commented):
Looks good so far 👌

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
mmitiche: And the E2E tests of the QuanticResultCopyToClipboard component x)

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.test.ts
erocheleau: Sure I was not aware of any preferences and vscode suggested `toBeTruthy` first haha

packages/headless/src/features/attached-results/attached-results-state.ts
erocheleau: Yup,
That's the question I asked at a scrum, either we keep the bare minimum in headless state or we just add those values here but make them optional pretty much.

packages/headless/src/features/attached-results/attached-results-slice.ts
erocheleau: Sorry I don't get where you'd want me to use the isNullOrUndefined ?

packages/headless/src/features/attached-results/attached-results-state.ts
nathanlb: If the values would be used in a future quantic-based attachedResults component then It's probably good to keep them here so the component doesn't need to fetch the data anywhere else.

packages/headless/src/features/attached-results/attached-results-slice.ts
```diff
@@ -0,0 +1,54 @@
...
+      });
+  }
+);
+
+const attachedResultsMatchIds = (
+  result1: AttachedResult,
+  result2: AttachedResult
+) => {
+  const isPermanentIdEqual =
+    result1?.permanentId !== undefined &&
```
nathanlb, line 48: here is where i was asking about `nullOrUndefined`. Looks like my comments got out of sync.

packages/headless/src/features/attached-results/attached-results-state.ts
```diff
@@ -0,0 +1,24 @@
...
+  permanentId?: string;
+  resultUrl: string;
+  source: string;
+  title: string;
+  uriHash?: string;
+}
+
+export interface AttachedResultsState {
+  results: AttachedResult[];
+  message?: string;
```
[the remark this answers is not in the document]
erocheleau: I removed that part of the state in the end it was not useful

packages/headless/src/features/attached-results/attached-results-slice.test.ts
erocheleau: In the second test I'm testing that after setting the results, we can actually find the result we've just set.
In the first test, I test the action payload and the number of results that were set.

I guess they are very similar but slightly different. 🤷

erocheleau: I merged both tests

packages/headless/src/features/attached-results/attached-results-slice.ts
erocheleau: ahhh yes makes sense now!

packages/headless/src/features/attached-results/attached-results-actions.ts
erocheleau: Because of "reasons" I don't want to "crash" if you load attached results that have bad data.

I think in the past there was even a third use-case where you could only load CaseAttach results (Native Salesforce attached knowledge articles, non-coveo results) those would have neither field.

But I am not certain of the backwards-compatibility we want to keep with that feature...

So let's say you'd load results that contain neither, it'd be okay, they just wouldn't show up as attached ever.
But IMO that shouldn't "crash" the attached results feature just because you have 1 bad result with missing field values I think anyway.

mmitiche: Makes a lot of sense! thank you for the clarification👌

packages/headless/src/features/attached-results/attached-results-actions-loader.ts
```diff
@@ -0,0 +1,44 @@
...
+   * @param loading - Optional state of loading.
+   * @returns A dispatchable action.
+   */
+  setAttachedResults(
+    payload: SetAttachedResultsActionCreatorPayload
+  ): PayloadAction<SetAttachedResultsActionCreatorPayload>;
+}
+
+/**
+ * Loads the `debug` reducer and returns possible action creators.
```
nathanlb, line 31:
```suggestion
 * Loads the `attachedResults` reducer and returns possible action creators.
```

packages/headless/src/features/attached-results/attached-results-state.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  permanentId?: string;
+  /**
+   * The url of the attached result.
+   */
+  resultUrl: string;
+  /**
+   * The source name where the attached result comes from.
+   */
+  source: string;
```
nathanlb, line 33: Is this for the Coveo source? I'm wondering why it's required while most of the others aren't.

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -65,11 +53,35 @@ export default class QuanticResultAction extends LightningElement {
...
+   * Specifies whether the result action button is in loading state or not.
+   * @api
+   * @type {boolean}
+   */
+  @api 
+  get loading() {
+    return this._loading;
+  }
+  set loading(isLoading) {
```
nathanlb, line 80: minor nitpick but here we have the get and set with no new line in between and those for `selected` have a new line. either is fine but ideally not both. 😅

nathanlb (approved):
Looking good 🏅

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.ts
```diff
@@ -0,0 +1,113 @@
...
+        const permanentIdMatches =
+          attached.permanentId &&
+          attached.permanentId === result.raw.permanentid;
+        const uriHashMatches =
+          attached.uriHash && attached.uriHash === result.raw.urihash;
+        return permanentIdMatches || uriHashMatches;
+      });
+    }
+    return attached;
+  };
```
olamothe, line 85: Suggestion: 
```suggestion
  const isResultAttached = (result: SearchResult) => {
   const {permanentid, urihash} = result.raw;
      if (!permanentid && !urihash) {
        return false;
      }

      return engine.state.attachedResults.results.some((attached) => {
        return (
          attached.permanentId === permanentid || attached.uriHash === urihash
        );
      });
  };
```

packages/headless/src/features/attached-results/attached-results-state.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  caseId: string;
+  /**
+   * The Id of an attached Salesforce Knowledge Article.
+   */
+  knowledgeArticleId?: string;
+  /**
+   * The permanentId, the identifier of an attached result.
+   */
+  permanentId?: string;
```
olamothe, line 25: The casing of this seems to differ slightly (but annoyingly ;) ) from what the search API returns. Same comment for `urihash` below.

And it leaks in other places in this PR (for example when comparing `result.permanentid === attachedResult.permanentId` )

Similar to my above comment, I think in general this interface could benefit from `Pick`-ing properties from `Result`, and extending with the new properties that are specifically needed for attach result feature.

packages/headless/src/features/attached-results/attached-results-state.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export interface AttachedResultsState {
+  /**
+   * The array of results that are attached.
+   */
+  results: AttachedResult[];
+  /**
+   * A loading state, used to sync loading the attached results between multiple components.
+   */
+  loading?: boolean;
```
olamothe, line 52: I don't think this should be optional 🤔 ?

Just `false` by default.

Otherwise downstream, you have to deal with `true | false | undefined` as opposed to just `true | false`

olamothe (approved):

packages/headless/src/features/attached-results/attached-results-state.ts
erocheleau: yea I just removed it actually it was breaking some other thing to be required :D

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.ts
```diff
@@ -0,0 +1,113 @@
...
+   * The permanentid of the result, a unique identifier of a result.
+   */
+  permanentid?: string;
+  /**
+   * The urihash of the result, a unique identifier of a result. Used only for backwards compatibility.
+   */
+  urihash?: string;
+}
+
+export interface SearchResult {
```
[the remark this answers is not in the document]
erocheleau: Yea removed that and I'm using the Result type as-is since that's actually what is expected.

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.ts
erocheleau: I'll check first indeed that seems much better
I'll do a variation of that I think because in a case where both `attached.permanentId` and `result.permanentId` are undefined, but urihash are different but not undefined, this would mark the results as attached.

`undefined === undefined || false` still returns true :)

packages/headless/src/features/attached-results/attached-results-state.ts
erocheleau: very true!

packages/headless/src/features/attached-results/attached-results-state.ts
erocheleau: Yea, the interface is different on purpose, the camelCase stuff is the data coming from the CRM AttachedResults, that's why this thing is so complicated... Because we need to convert the data from an external source to match the searchapi format.

I refactored it a little bit so it uses a lot more of the searchapi result format of data. It should be clearer I think?

mmitiche (commented):
LGTM 👌
The only missing part I can find is the the update that need to be done in the QuanticResultCopyToClipboard component since we changed how we set the loading in the QuanticResultAction component.

erocheleau:
> LGTM 👌 The only missing part I can find is the the update that need to be done in the QuanticResultCopyToClipboard component since we changed how we set the loading in the QuanticResultAction component.

Oh good point I'll include that 👍

mmitiche (approved):
Perfecto 👌

packages/headless/src/controllers/insight/attach-to-case/headless-attach-to-case.ts
```diff
@@ -0,0 +1,133 @@
...
+   */
+  options: AttachToCaseOptions;
+}
+
+export interface AttachToCaseOptions {
+  /**
+   * The result to attach, detach.
+   */
+  result: Result;
+  caseId: string;
```
nathanlb, line 36: no description for caseId?

nathanlb (approved):
chick'n swell 🐔

PR #2606, opened 2022-12-12
fix(atomic): replace i18n default escape function with dom purify

olamothe:
https://coveord.atlassian.net/browse/KIT-2175

### DIFF
--- a/packages/atomic/src/components/common/interface/i18n.ts
+++ b/packages/atomic/src/components/common/interface/i18n.ts
@@ -1,4 +1,5 @@
 import {getAssetPath} from '@stencil/core';
+import DOMPurify from 'dompurify';
 import Backend, {BackendOptions} from 'i18next-http-backend';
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/common/interface/i18n.ts
```diff
@@ -57,5 +58,8 @@ export function initi18n(atomicInterface: BaseAtomicInterface<AnyEngineType>) {
     nsSeparator: '___',
     fallbackLng: 'en',
     backend: i18nBackendOptions(atomicInterface),
+    interpolation: {
+      escape: (str) => DOMPurify.sanitize(str),
```
btaillon, line 62: Technically, with DOMPurify, we are now allowing HTML elements. I assume they weren't allowed before.

I don't really have an issue with that though.

btaillon (approved):

ThibodeauJF (approved):

PR #2614, opened 2022-12-14
fix: fixed v2 dependency linking

btaillon:
https://coveord.atlassian.net/browse/KIT-2165

* Some dependencies weren't automatically bumped, so I made them use a caret version selector
* I manually removed the duplicate v2 changelog entries
* It looks like the package-lock was broken, I fixed it by running `lerna bootstrap` then `npm install`.


olamothe (approved):

louis-bompart (approved):
lgtm, but a fix-it-twice subsequent PR will be needed I reckon

PR #2615, opened 2022-12-14
feat(Quantic): fix triggering query suggestions for any key press

jelmedini:
Fixing unnecessary calls when typing, calling the api only when the value is changed


packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -210,6 +210,13 @@ export default class QuanticSearchBox extends LightningElement {
     this.input.blur();
   }
 
+  handleKeyValues() {
+    if (this.searchBox.state.value !== this.input.value) {
+      this.suggestionList?.resetSelection();
+      this.searchBox.updateText(this.input.value);
+    }
+  }
```
erocheleau, line 218: Should we actually have a debounce as well?
Like a small delay before calling the suggestions?

erocheleau (approved):
Just added a question, but your changes are good

mmitiche (approved):
LGTM👌

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
jelmedini: I don't know if it s relevant here regarding how the component should work, but if so we need to do this improvement in several locations on this file

PR #2617, opened 2022-12-14
feat(quantic): new internal property isHovered added to the QuanticResult component

mmitiche:
[SFINT-4745](https://coveord.atlassian.net/browse/SFINT-4745)

#### This new property will then be used from the result templates, to allow us implement the following behaviour:

https://user-images.githubusercontent.com/86681870/207876955-2259ce7f-3da9-49c2-9f13-7a3913272a4b.mov

#### Usage in a result template:

ysage


nathanlb (approved):
Simple and effective 👍 
out of scope of this PR but in the SFINT quantic insight panel I think it would look nicer if the background change goes edge-to-edge.
So perhaps the left/right padding could be on the result template rather than the outside container. Something to consider.

mmitiche:
> Simple and effective 👍 out of scope of this PR but in the SFINT quantic insight panel I think it would look nicer if the background change goes edge-to-edge. So perhaps the left/right padding could be on the result template rather than the outside container. Something to consider.

Sure yeah! I'm fixing this in the InsightRenderer component 👌

packages/quantic/force-app/main/default/lwc/quanticResult/quanticResult.js
```diff
@@ -40,8 +40,13 @@ export default class QuanticResult extends LightningElement {
 
   @track resultHasPreview = true;
 
+  /** @type {boolean} */
+  isHovered = false;
+
   connectedCallback() {
     this.template.addEventListener('haspreview', this.onHasPreview);
+    this.template.host.addEventListener('mouseenter', this.setHoverState);
```
lbergeron, line 48: I guess it is out of scope, but what about keyboard navigation?

packages/quantic/force-app/main/default/lwc/quanticResult/quanticResult.js
```diff
@@ -40,8 +40,13 @@ export default class QuanticResult extends LightningElement {
 
   @track resultHasPreview = true;
 
+  /** @type {boolean} */
+  isHovered = false;
```
lbergeron, line 44: What would you think of naming this attribute `actionsDisplayed`? It would reduce the coupling with the mouse input.

nathanlb: The way I see it this property allows developer users to hook onto the "hovered" state when making custom result templates and this need might be for something else than displayed result actions. Also the result actions are tied to the insight panel use case right now but the quanticResult component is used in search as well. So keeping the property name more generic is more appropriate in my opinion.

lbergeron: I agree with you that `isHovered` is not a bad name. The thing that bothers me a little is that it works only with the mouse input.

I just realized that the attribute is private. There is no problem using `isHovered` then. If we want to improve accessibility down the road, we can rename the attribute without breaking anything.

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticResult/quanticResult.js
erocheleau: Yea I think we could switch to something like `isFocused` instead and support maybe the keyboard navigation?

Don't have to support keyboard for this PR, but we could already use the name `focused` instead of `hovered` for clarity as well I think I prefer that name, what about you all?

erocheleau (approved):

packages/quantic/force-app/main/default/lwc/quanticResult/quanticResult.js
nathanlb: Hover and Focus are two different things though. But I agree we should ultimately offer both. For example, the result actions should display both on hover and on focus.

mmitiche: I agree with Nathan, sometimes we want two different behaviours on hover and on focus since they are two separate things. We now have the `isHovered` property and I'll create a new ticket to add the `isFocused` property, the property that will allow us to support the keyboard navigation.

lbergeron: I can't find a single term that conveys the idea of `focused or hovered`. Perhaps `focused` is the best one.

PR #2622, opened 2022-12-16
fix(quantic): fallback copyToClipboard method added to support this operation when Locker is ON

mmitiche:
[SFINT-4736](https://coveord.atlassian.net/browse/SFINT-4736)

## The issue: 
The Clipboard API is not available in a Salesforce community where locker is activated. 

## The solution:
Fallback method added to perform the copy to clipboard operation even when Locker is activated.

#### Note:
The fallback method causes a loss of focus, that's why I added to some logic to properly reset the focus after performing the copy to clipboard operation.

## Demo:

https://user-images.githubusercontent.com/86681870/208134619-dcc77984-6111-465b-b1eb-c2def401b270.mov


packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -697,13 +697,34 @@ export function isParentOf(element, targetElement) {
  * @param {string} text
  */
 export async function copyToClipboard(text) {
```
nathanlb, line 699: Do you think these are jest-testable?

nathanlb (approved):
Looking good but maybe we can add a test or two?

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: Sure yeah! let me try to add that!

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticQueryError/quanticQueryError.js
```diff
@@ -25,6 +25,8 @@ import { errorMap, genericError } from './errorLabels.js';
  * <c-quantic-query-error engine-id={engineId}></c-quantic-query-error>
  */
 export default class QuanticQueryError extends LightningElement {
+  static delegatesFocus = true;
```
erocheleau, line 28: What is that supposed to do?

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
```diff
@@ -107,6 +108,7 @@ export default class QuanticResultCopyToClipboard extends LightningElement {
         this.engine.dispatch(this.actions.logCopyToClipboard(this.result));
         this.displayedLabel = this.successLabel;
         this.resetOriginalLabel();
+        this.template.host.focus();
```
erocheleau, line 111: This sets focus after the fallback? Because the fallback uses the create element thing to focus, then copy?

Might want to add a comment specifically explaining that right there so we understand in 2 weeks what this does ;)

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -697,13 +697,34 @@ export function isParentOf(element, targetElement) {
...
-  await navigator.clipboard.writeText(text);
+  if (navigator?.clipboard?.writeText) {
+    try {
+      await navigator.clipboard.writeText(text);
+    } catch (err) {
+      copyToClipboardFallback(text);
+    }
+  } else {
+    copyToClipboardFallback(text);
+  }
```
erocheleau, line 708: Can't you just do? Would the catch, catch the not-available navigator api if it's not present?

```suggestion
try {
    await navigator.clipboard.writeText(text);
} catch (err) {
    copyToClipboardFallback(text);
}
```

erocheleau (approved):
Seems good, just a question

packages/quantic/force-app/main/default/lwc/quanticQueryError/quanticQueryError.js
mmitiche: This is related to the point you highlighted in a previous PR: https://github.com/coveo/ui-kit/pull/2558#discussion_r1038511134
Basically, the fallback copy to clipboard method will reset the focus to the beginning of the page. That's why I'm adding this logic in order to reset the focus to the copyToClipboard button using the [delegatesFocus](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_components_focus).

packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard/quanticResultCopyToClipboard.js
mmitiche: sure yeah👌

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: yes better!

PR #2623, opened 2022-12-16
feat(atomic): implement quickview pagination/navigation/keyword hit

olamothe:
Add buttons in the footer of each individual Quickview in order to navigate through the result set.

* Add `quickviews` to `ResultListInfo` which contains the total number of results currently displayed, and the position of each result that possess a Quickview.  As a reminder`ResultListInfo` is stored in the atomic store, when a result list is registered.

* Not all index results have a Quickview, so the pagination need to handle "skipping" results in the result list. Trying to fetch a Quickview that does not have the `hasHtmlVersion` flag will end up erroring at the API/index level.

* Pagination is wrapped: When you reach the last result, you end up back to the first one, and vice versa.

* Handle resizing the iframe when new content load. This is in order to have a scroll bar adapted to the content of the quickview (or no scrollbar if it's not needed).

* Added a sidebar to perform keyword navigation in quickviews.

* The logic for keyword navigation has been heavily influenced/copy pasted from JSUI. It contains a lot of fixes that had been done over the year. There's quite a bit of gotcha about stemming and embedded keywords that needs to be accounted for.

https://coveord.atlassian.net/browse/KIT-2184

[KIT-2184]: https://coveord.atlassian.net/browse/KIT-2184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -46,6 +59,25 @@ export class AtomicQuickviewModal {
...
+    }
+    // This "reset" of the iframe height allows the recalculation of the proper height needed for new content being added to the iframe
+    // when the end user navigates through the available quickview
+    // Since setting a new height on the iframe will cause the resize observer to essentially "call itself", we also add a flag to stop double resizing
+    this.iframeRef.style.height = '0';
+    let shouldRecalculateResize = true;
+    new ResizeObserver(([documentElementObserved]) => {
+      if (this.iframeRef && shouldRecalculateResize) {
+        this.iframeRef.style.height = `${
+          documentElementObserved.contentRect.height + 20
```
btaillon, line 74: I prefer to avoid unnamed constants like `20` here.

Perhaps there's another property or some CSS we could use to avoid needing it?

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -46,6 +59,25 @@ export class AtomicQuickviewModal {
...
+        }px`;
+      }
+      shouldRecalculateResize = !shouldRecalculateResize;
+    }).observe(documentWriter.documentElement);
```
btaillon, line 78: I think we should unobserve this somewhere.

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -0,0 +1,140 @@
...
+  <atomic-icon-button
+    buttonStyle="text-transparent"
+    class={`w-fit ${minimize ? '' : 'ml-auto'}`}
+    icon={MinimizeIcon}
+    labelI18nKey="minimize"
+    tooltip={i18n.t('minimize')}
+    clickCallback={() => onMinimize(!minimize)}
+    badge={
+      highlightKeywords && minimize ? <slot>{wordsLength}</slot> : undefined
+    }
```
btaillon, line 68: We should also add `aria-expanded="true" / aria-expanded="false"` and `aria-controls="unique-id-of-keywords-navigation` here.

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -0,0 +1,140 @@
...
+const KeywordsNavigation: FunctionalComponent<
+  Pick<QuickviewSidebarProps, 'i18n'> & {
+    words: Record<string, QuickviewWordHighlight>;
+  }
+> = ({words, i18n}) => {
+  return (
+    <Fragment>
+      {Object.values(words).map((keyword) => {
+        return (
+          <div class="flex items-center w-100 my-4 bg-background border border-neutral rounded-lg overflow-x-auto">
```
btaillon, line 106: This is missing a key:
```suggestion
          <div key={keyword.text} class="flex items-center w-100 my-4 bg-background border border-neutral rounded-lg overflow-x-auto">
```

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -0,0 +1,140 @@
...
+            <div class="flex items-center grow p-4 border-r">
+              <div
+                class="w-5 h-5 flex-none mr-2"
+                style={{backgroundColor: keyword.color}}
+              ></div>
+              <div class="grow mr-2">{keyword.text}</div>
+              <div class="flex-none">{keyword.occurrences}</div>
+            </div>
```
btaillon, line 114: I think this section should get an aria-label to describe the relationship between the text and the number of occurences, something like "navigate between {occurences} occurences of {text}".

Potentially better: you could set `aria-hidden="true"` to this component, and wrap the navigation buttons in a `<FieldsetGroup>` with a label.

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -0,0 +1,140 @@
...
+    QuickviewSidebarProps,
+    'i18n' | 'minimize' | 'onMinimize' | 'highlightKeywords'
+  > & {wordsLength: number}
+> = ({i18n, minimize, onMinimize, highlightKeywords, wordsLength}) => (
+  <atomic-icon-button
+    buttonStyle="text-transparent"
+    class={`w-fit ${minimize ? '' : 'ml-auto'}`}
+    icon={MinimizeIcon}
+    labelI18nKey="minimize"
+    tooltip={i18n.t('minimize')}
```
btaillon, line 64: Instead of just `minimize` as an aria-label, I think we should describe what component we're minimizing.

Also, I think we should avoid the wording of "minimize", since screen readers would read it a bit weirdly (e.g., "Minimize keywords navigation, expanded").

One solution which would also be relevant for the tooltip would be to use "toggle" instead:
> Toggle keywords navigation

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -142,6 +169,27 @@ export class AtomicQuickviewModal implements InitializableComponent {
...
+    }
+    if (this.highlightKeywords) {
+      doc.getElementById('CoveoDisableHighlightStyle')?.remove();
+    } else {
+      const head = doc.head;
+      const style = doc.createElement('style');
+      style.setAttribute('id', 'CoveoDisableHighlightStyle');
+      head.appendChild(style);
+      style.appendChild(
+        doc.createTextNode(`[id^=CoveoHighlight] {
```
btaillon, line 185: It looks like this should be:
```suggestion
        doc.createTextNode(`[id^="${HIGHLIGHT_PREFIX}"] {
```

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -142,6 +169,27 @@ export class AtomicQuickviewModal implements InitializableComponent {
     return !!this.content && !!this.result;
   }
 
+  private handleKeywordHighlight() {
```
btaillon, line 172: This method name seems redundant with the above `watchHighlightKeywords`.

I think this method should be split in two: `enableHighlights` / `disableHighlights`.

btaillon (commented):
I left some comments concerning accessibility, but accessibility should probably be fixed in a separate PR.

Even though this component is very visual, I think screen reader accessibility for this component is still pretty important, since it could also be useful to screen reader users.

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
olamothe: That just some arbitrary "breathing room", as it's uncomfortable to read text if the last line of the quickview is exactly aligned with the bottom of the iframe.

I can name it.

packages/atomic/src/components/common/result-list/result-list-common.tsx
```diff
@@ -67,6 +67,30 @@ export class ResultListCommon<SpecificResult extends AnyResult = AnyResult>
     this.props.nextNewResultTarget.focusOnNextTarget();
   }
 
+  public get quickviews() {
```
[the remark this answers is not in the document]
olamothe: I'm interested in hearing what Jean Felix's opinion first about this first.
If that's the consensus I will change it in a separate PR and implement Quickview pagination directly in Headless.

The main reason I did it here is that it seems to be really a UI choice/concern to manage navigating between preview of different results in a result list.

For this PR though, I would leave it as is.

packages/headless/src/api/search/search/stemming.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Terms to highlight (with stemming) in the results
```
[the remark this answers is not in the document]
olamothe: @jpmarceau : What do you think we should add as a description for "stemming" ?

Benjamin's point is right I think, but stemming can be something complicated to explain.

Maybe this article ? https://docs.coveo.com/en/1576/searching-with-coveo/about-stemming

packages/atomic/src/components/search/result-template-components/quickview-word-highlight/quickview-word-highlight.tsx
```diff
@@ -0,0 +1,201 @@
...
+  existingHighlight: QuickviewWordHighlight,
+  newWordHighlight: QuickviewWordHighlight,
+  newWordElement: HTMLElement
+) => {
+  existingHighlight.addElement(newWordElement as HTMLElement);
+
+  // Special code needed to workaround invalid HTML returned by the index with embedded keyword
+  if (existingHighlight.occurrences === newWordHighlight.occurrences) {
+    existingHighlight.text += newWordHighlight.text;
+  }
```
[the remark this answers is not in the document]
olamothe: Indeed, this is a bad copy paste from JSUI.

I don't think we need this anymore.

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
olamothe: I will try to address accessibility issues in this ticket: https://coveord.atlassian.net/browse/KIT-2199

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
olamothe: Similar as above, I will address accessibility issues in a different PR.

packages/headless/src/api/search/search/stemming.ts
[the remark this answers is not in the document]
jpmarceau: I like Benjamin's idea of an example. It should answer most people's questions. I'd just tweak it a little because I think it'd be clearer if we distinguished stemming from query corrections.
I also like the idea of linking to that document for people who'd like to know more, since explaining what stemming is takes quite a bit of space. I'm not sure I'd do that in code annotations.

```suggestion
 * Terms to highlight (with stemming) in the results.
 * E.g., for the query "indexing consultant", where a result contains "Techzample is a consulting firm specializing in search indexes. It employs consultants that can analyze and improve your search index.", the `TermsToHighlight` might be `{ "indexing": ["indexes", "index"], "consultant": ["consulting", "consultants"] }`.
 * To learn more about stemming, see [About Stemming](https://docs.coveo.com/en/1576).
```

packages/atomic/src/components/common/result-list/result-list-common.tsx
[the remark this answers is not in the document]
ThibodeauJF: I agree with Benjamin, it seems a better fit for Headless, even if it stems from a UI requirement, in the end it handles state, and it could be used by many UIs

ThibodeauJF (approved):

packages/atomic/src/components/common/result-list/result-list-common.tsx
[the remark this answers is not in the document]
olamothe: Ok !
https://coveord.atlassian.net/browse/KIT-2205

PR #2484, opened 2022-10-04
feat(atomic,headless): add atomic-recs-result component, add missing rec state

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2051

PR #2625, opened 2022-12-19
fix(atomic): fixed if-defined and if-not-defined template conditions

nathanlb:
In [this PR](https://github.com/coveo/ui-kit/pull/2484/files#diff-a262166f61e001e18087844ddae38b46adc5a836fe92fc51a875f8ab28d12d78) which refactored the result template components to use a common ResultTemplateCommon component it appears a mistake was made and the `ifDefined` and `ifNotDefined` props for creating conditions were no longer used which essentially breaks the feature.

Proposed changes:
- Add a utility function to the common component for crafting the `ifDefined` and `ifNotDefined` conditions
- Use the utility in the `atomic-insight-result-template` and `atomic-recs-result-templates` where i found those no-op/broken props.
- Add service-core as codeowner for the insight atomic components so we can help catch this kind of breakage in the future.
- Added a salesforce template to the example atomic insight panel using this type of result template condition to be able to validate.


packages/atomic/src/pages/examples/insights.html
```diff
@@ -79,6 +79,50 @@
...
+                </style>
+                <atomic-result-section-badges>
+                  <atomic-field-condition must-match-sourcetype="Salesforce">
+                    <atomic-result-badge label="Salesforce" class="salesforce-badge"></atomic-result-badge>
+                  </atomic-field-condition>
+                </atomic-result-section-badges>
+                <atomic-result-section-title><atomic-result-link></atomic-result-link></atomic-result-section-title>
+                <atomic-result-section-excerpt
+                  ><atomic-result-text field="excerpt"></atomic-result-text
+                ></atomic-result-section-excerpt>
```
lbergeron, line 109:
```suggestion
                <atomic-result-section-excerpt>
                  <atomic-result-text field="excerpt"></atomic-result-text>
                </atomic-result-section-excerpt>
```

lbergeron (approved):
Can you please add a few tests [here](https://github.com/coveo/ui-kit/blob/master/packages/atomic/cypress/e2e/result-list/result-template.cypress.ts) to cover the change?

nathanlb:
> Can you please add a few tests [here](https://github.com/coveo/ui-kit/blob/master/packages/atomic/cypress/e2e/result-list/result-template.cypress.ts) to cover the change?

The test suite you linked is for the `atomic-result-template` component which is unaffected by my change. But I could indeed add some tests for the `atomic-insight-result-template`. I looked the other way because we currently only have rudimentary smoke tests for the insight use case but I was probably wrong to do so.

lbergeron:
> > Can you please add a few tests [here](https://github.com/coveo/ui-kit/blob/master/packages/atomic/cypress/e2e/result-list/result-template.cypress.ts) to cover the change?
> 
> The test suite you linked is for the `atomic-result-template` component which is unaffected by my change. But I could indeed add some tests for the `atomic-insight-result-template`. I looked the other way because we currently only have rudimentary smoke tests for the insight use case but I was probably wrong to do so.

Sorry for pointing the wrong file. However, having a test would be nice to avoid potential regressions on this fix.

packages/atomic/src/components/common/result-templates/result-template-common.tsx
```diff
@@ -164,3 +164,19 @@ export function makeMatchConditions(
   }
   return conditions;
 }
+
+export function makeDefinedConditions(
```
olamothe, line 168: As Luc pointed out, I think this function here could be easily unit tested

olamothe (approved):
It seems there's a ... divergence(?) between `atomic-recs-result-template`, `atomic-insight-result-template` and `atomic-result-template` in how they are instantiated, which is really not ideal.

I think this is out of scope of your PR, but eventually we'll need to refactor.

packages/atomic/src/components/common/result-templates/result-template-common.tsx
olamothe: Yes for straightforward `expected in -> expected out` kinda function we have UT in the project.

Does not need to be in the utils folder.

nathanlb: I added some tests 👍

PR #2629, opened 2022-12-21
test(atomic): improve breadcrumb rendering tests

olamothe:
https://coveord.atlassian.net/browse/KIT-2129


packages/atomic/cypress/e2e/facets/category-facet/category-facet.cypress.ts
```diff
@@ -784,7 +784,10 @@ describe('Category Facet Test Suites', () => {
       }
 
       describe('verify rendering', () => {
-        before(setupSelectedCategoryFacet);
+        before(() => {
+          setupSelectedCategoryFacet();
+          cy.wait(TestFixture.interceptAliases.Search);
```
samisayegh, line 789: Could it be better to place the wait inside the function triggering the search? i.e. inside `setupSelectedCategoryFacet`

Said differently, do we have scenarios where we want to setup a facet and not wait for the search to finish before continuing? For example, this function is called on line 802 without waiting. Intentional? Or a potential flaky test?

Same comment for the other setup functions,

samisayegh (approved):

packages/atomic/cypress/e2e/facets/category-facet/category-facet.cypress.ts
olamothe: > Said differently, do we have scenarios where we want to setup a facet and not wait for the search to finish before continuing?

Yes ! For analytics verification.

PR #2630, opened 2022-12-21
fix(headless): improve escape highlight function

olamothe:
The escape function, like the comment says, is inspired by underscore.

Problem was that the implementation was no where near what underscore does, and was buggy.

Re-implemented it properly.

https://coveord.atlassian.net/browse/KIT-2189

[KIT-2189]: https://coveord.atlassian.net/browse/KIT-2189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/utils/highlight.ts
+++ b/packages/headless/src/utils/highlight.ts
@@ -150,11 +150,20 @@ function suggestionWithDelimiters(
  */
 
 export function escape(str: string) {
-  return str
-    .replace(/&/g, '&amp')
-    .replace(/</g, '&lt')
-    .replace(/>/g, '&gt')
-    .replace(/"/g, '&quot')
-    .replace(/`/g, '&#96')
-    .replace(/'/g, '&#x27');
+  const mapOfCharToEscape: Record<string, string> = {
+    '&': '&amp;',
+    '<': '&lt;',
+    '>': '&gt;',
+    '"': '&quot;',
+    "'": '&#x27;',
+    '`': '&#x60;',
+  };
+
+  const source = '(?:' + Object.keys(mapOfCharToEscape).join('|') + ')';
+  const testRegexp = RegExp(source);
+  const replaceRegexp = RegExp(source, 'g');
+
+  return testRegexp.test(str)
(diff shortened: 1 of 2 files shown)


packages/headless/src/utils/highlight.ts
```diff
@@ -150,11 +150,20 @@ function suggestionWithDelimiters(
...
 
 export function escape(str: string) {
-  return str
-    .replace(/&/g, '&amp')
-    .replace(/</g, '&lt')
-    .replace(/>/g, '&gt')
-    .replace(/"/g, '&quot')
-    .replace(/`/g, '&#96')
-    .replace(/'/g, '&#x27');
+  const mapOfCharToEscape: Record<string, string> = {
```
btaillon, line 153: If possible, I would prefer to avoid having to maintain our own map of characters. If this map of characters, as used by Lodash, is later determined to be a vulnerability, we won't be notified since we aren't using Lodash's own function.

It looks like [`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) encodes all those characters except the apostrophe, which shouldn't be a problem for HTML, but I don't see any guarantee that the output of `encodeURIComponent` is innerHTML-safe.

I think we should use a library for this.

olamothe: > It looks like [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) encodes all those characters except the apostrophe, which shouldn't be a problem for HTML, but I don't see any guarantee that the output of encodeURIComponent is innerHTML-safe.

encodeURIComponent is to encode characters to be used in URL. It's not really related to what we are really interested here.

> I think we should use a library for this.

I disagree. 

I get your first point about security, but I think it's a bit over-engineered. 

The chances that a new HTML characters will magically appear one day and introduce a problem is... low 😄 .

To illustrate differently, Lodash implementation has not changed in 6 years (which is pretty much the age of the library).

btaillon (approved):

PR #2631, opened 2022-12-23
feat(headless): insight quickview using proper endpoint

nathanlb:
*WIP as the proxy endpoint is not yet in production*
Modified the insight quickview controller to use the quickview endpoint on the customer-service-api.

### DIFF
--- a/packages/headless/src/controllers/insight/quickview/headless-insight-quickview.test.ts
+++ b/packages/headless/src/controllers/insight/quickview/headless-insight-quickview.test.ts
@@ -1,0 +1,132 @@
+import {configuration, resultPreview} from '../../../app/reducers';
+import {updateContentURL} from '../../../features/result-preview/result-preview-actions';
+import {logDocumentQuickview} from '../../../features/result-preview/result-preview-analytics-actions';
+import {buildMockResult} from '../../../test';
+import {
+  buildMockInsightEngine,
+  MockInsightEngine,
+} from '../../../test/mock-engine';
+import {buildMockResultPreviewState} from '../../../test/mock-result-preview-state';
+import {
+  buildQuickview,
+  QuickviewOptions,
+  Quickview,
+} from './headless-insight-quickview';
+
+describe('Insight Quickview', () => {
+  let engine: MockInsightEngine;
+  let options: QuickviewOptions;
+  let quickview: Quickview;
+
+  function initQuickview() {
+    quickview = buildQuickview(engine, {options});
+  }
+
+  beforeEach(() => {
+    engine = buildMockInsightEngine();
+    options = {
+      result: buildMockResult(),
+      maximumPreviewSize: 0,
+    };
+
+    initQuickview();
+  });
+
+  it('initializes', () => {
+    expect(quickview).toBeTruthy();
+  });
+
+  it('it adds the correct reducers to engine', () => {
+    expect(engine.addReducers).toHaveBeenCalledWith({
+      configuration,
+      resultPreview,
+    });
+  });
+
+  it('exposes a subscribe method', () => {
+    expect(quickview.subscribe).toBeTruthy();
+  });
+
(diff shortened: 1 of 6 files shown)


packages/headless/src/controllers/insight/quickview/headless-insight-quickview.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    };
+
+    initQuickview();
+  });
+
+  it('initializes', () => {
+    expect(quickview).toBeTruthy();
+  });
+
+  it('it adds the correct reducers to engine', () => {
```
jelmedini, line 39: double **it**, `it('adds the correct reducers to engine', () ...`

jelmedini (approved):

packages/headless/src/features/insight-search/insight-result-preview-request-builder.ts
```diff
@@ -0,0 +1,28 @@
...
+import {
+  HtmlRequest,
+  HtmlRequestOptions,
+} from '../../api/search/html/html-request';
+import {StateNeededByHtmlEndpoint} from '../result-preview/result-preview-request-builder';
+
+export async function buildInsightResultPreviewRequest(
+  state: StateNeededByHtmlEndpoint,
+  options: HtmlRequestOptions,
+  url: string
```
btaillon, line 11: Could the `url` be built inside this function instead of the controller? It looks like everything needed is already in the `state`

nathanlb: Yeah it got more complicated because to build the url i need the insight configuration ID which is not part of `StateNeededByHtmlEndpoint. I can cast it but it feels a bit sketchy. I'll try a few things.

nathanlb: Do you see a better way than:

```typescript
/**
 * Creates an insight `Quickview` controller instance.
 * @param engine - The insight engine.
 * @param props - The configurable `Quickview` properties.
 * @returns A `Quickview` controller instance.
 */
export function buildQuickview(
  engine: InsightEngine,
  props: QuickviewProps
): Quickview {
  const fetchResultContentCallback = () => {
    engine.dispatch(logDocumentQuickview(props.options.result));
  };

  const path = '/quickview';
  const coreProps = {
    options: {
      ...props.options,
      onlyContentURL: true,
    },
  };

  return buildCoreQuickview(
    engine,
    coreProps,
    (state, options) =>
      buildInsightResultPreviewRequest(
        state as StateNeededByInsightHtmlEndpoint,
        options
      ),
    path,
    fetchResultContentCallback
  );
}
```
where
```typescript
export type StateNeededByInsightHtmlEndpoint = StateNeededByHtmlEndpoint &
  InsightConfigurationSection;
```
?

btaillon: yeah, in that case I think you'd change the state needed for the function to include the configuration section.

For the controller, to ensure that you have all the needed reducers (and engine.state has the correct type), I recommend copying the pattern we have in files like `headless-facet.ts` where we call `load*Reducers`

packages/headless/src/controllers/insight/quickview/headless-insight-quickview.ts
```diff
@@ -26,11 +29,21 @@ export function buildQuickview(
   };
 
   const path = '/quickview';
+  const coreProps = {
+    options: {
+      ...props.options,
+      onlyContentURL: true,
```
btaillon, line 35: Since `onlyContentURL` is always `true`, I recommend removing the irrelevant properties from the Insight buildQuickview (both the `props` and the return type) to reflect that.

Since that change would be breaking, you may have to simply mark some properties as `@deprecated`, and we can remove them in v3.

btaillon (approved):

PR #2632, opened 2023-01-02
fix(atomic): fix modal refine not respecting DOM order

olamothe:
Original problem discussed here: https://discuss.coveo.com/t/support-00093151-facet-order-is-not-respected-in-a-modal-on-mobile-view/8923/3

This PR reworks how we order facet in the refine toggle (the toggle that appears for facets and tabs when we open in mobile, or for the insight panel).

Previously, the code would use the facet manager internally to reorder all facets in the interface and then place them in  whatever order the manager dictates. Problem is that it would also reorder facets that are actually outside the `atomic-facet-manager` component.

Now instead the code relies on this:
1- If there is a facet manager component in the page, it will reorder it's children in the DOM
2- If there are facets component outside the facet manager, those will just appear in the order of the page that the implementer decided.

So the code queries for all facets registered in the interface using their tag as a CSS selector, which will return them in an array that respect their order in the DOM.

https://coveord.atlassian.net/browse/KIT-2187

[KIT-2187]: https://coveord.atlassian.net/browse/KIT-2187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathanlb (approved):
LGTM

packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
```diff
@@ -112,15 +111,15 @@ export function getClonedFacetElements(
...
-  const sortedFacetsElements = facetManager
-    .sort(facetElementsPayload)
-    .map((f) => f.payload);
+  const allFacetTags = Object.keys(
+    aggregate(facetElements, (el) => el.tagName.toLowerCase())
+  );
+
+  const allFacetsInOrderInDOM = document.querySelectorAll(
+    allFacetTags.join(',')
+  );
```
btaillon, line 120: I think we should avoid using `querySelector` from `document`, in case we eventually have cases where there's multiple search interfaces in the page.

I think we should use `querySelectorAll` relative to the search interface instead.

olamothe: Yeah good point !

PR #2633, opened 2023-01-02
fix(atomic): insight tab section full width

nathanlb:
Set the tab section to full width in the insight layout.

Before
Screen Shot 2023-01-02 at 9 47 04 AM

After
Screen Shot 2023-01-02 at 9 46 50 AM


jelmedini (approved):
Sounds good!

PR #2636, opened 2023-01-05
feat(headless): updated insight result template detail type

nathanlb:
The customer-service-api now returns the field type for result template details to allow the ui to render the appropriate result-template-[text|number|date|multi-value] component.

### DIFF
--- a/packages/headless/src/api/service/insight/get-interface/get-interface-config-response.ts
+++ b/packages/headless/src/api/service/insight/get-interface/get-interface-config-response.ts
@@ -38,9 +38,12 @@ interface Badge {
   color: string;
 }
 
+export type InsightFieldType = 'string' | 'date' | 'number' | 'multi';
+
 interface Detail {
   field: string;
   label?: string;
+  fieldType?: InsightFieldType;
 }
 
 type ResultAction =


packages/headless/src/api/service/insight/get-interface/get-interface-config-response.ts
```diff
@@ -38,9 +38,12 @@ interface Badge {
   color: string;
 }
 
+export type InsightFieldType = 'string' | 'date' | 'number' | 'multi';
+
 interface Detail {
   field: string;
```
btaillon, line 44: Does the type of `field` vary based on the field type, or is it always a string?

nathanlb: The type is always a string, it's the field name. Given that, i'm not sure examples make sense.

mmitiche (approved):

lbergeron (approved):

ThibodeauJF (approved):

packages/headless/src/api/service/insight/get-interface/get-interface-config-response.ts
nathanlb: Does this change things for you?

PR #2637, opened 2023-01-05
feat(headless): added insight user actions state and controller

nathanlb:
Added a controller for fetching and managing user actions from the customer-service-api
The API is still WIP so this can't be merged yet but it's a big PR so the more time we have to review the better.

There's a year old design doc [here](https://docs.google.com/document/d/1CzOVLZlQGAPvBDIsYdokDmN0-dFmTOAMhojGPpai3V4/edit#heading=h.iliynul9hvrs) that simply lists the attributes of the controller and state added.

Example Quantic Component:
Screen Shot 2023-01-05 at 8 57 46 AM


nathanlb:
Still missing analytics

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
```diff
@@ -0,0 +1,145 @@
...
+   * @defaultValue `[]`
+   */
+  excludedCustomActions?: string[];
+}
+
+export interface UserActionsState {
+  /**
+   * The timeline of user actions.
+   */
+  timeline: UserActionTimeline;
```
mmitiche, line 54: I have nothing against the current implementation , I'm just wondering it wouldn’t be more organized to have the timeline in the state as follows:
```
timeline: {
  precedingSessions: UserSession[],
  session: UserSession,
  followingSessions: UserSession
}
```

jelmedini (approved):
maybe i am missing something as I dont know  deeply the project yet, but the code seems good for me! clean and clear 👍 (just didn t see cypress tests, is there a need of any ? )

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
mmitiche, line 54: Follow up question, what will the state contain when for example we try to fetch **2 following sessions** but in reality there is only **one session** that has occurred after the creation of the ticket?

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
```diff
@@ -0,0 +1,145 @@
...
+ */
+export interface UserActions extends Controller {
+  /**
+   * Fetch the list of user actions surrounding the ticket creation.
+   */
+  fetchUserActions(): void;
+  /**
+   * Fetch the list of user actions including one more session preceding the ticket creation.
+   */
+  fetchWithPrecedingSession(): void;
```
mmitiche, line 92: Out of curiosity, do we currently plan to have a button in the front-end that allows to fetch one more preceding/following session?
I dont see any in the mockups so I'm assuming this action is made for potential future use right?

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  return {
+    ...controller,
+
+    get state() {
+      return getState();
+    },
+
+    fetchUserActions() {
+      dispatch(fetchUserActions());
```
mmitiche, line 125: Will the list of user actions fetched by this action contain the `case creation` event? or we need to add this event manually based on the `ticketCreationDate` option?

mmitiche (commented):
Looking good so far 👌

nathanlb:
> maybe i am missing something as I dont know deeply the project yet, but the code seems good for me! clean and clear 👍 (just didn t see cypress tests, is there a need of any ? )

Thanks for reviewing! No Headless has no cypress tests as there is no visual component to it. It's purely business logic and we only do unit tests.

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: This is precisely the kind of discussion i was hoping for! I did it this way based on the API design we have had for a while. If we want a structure like this then I would do it at the API level especially considering it isn't made yet and have the state reflect it.
If this makes the API and the state much easier to use then maybe we should do that. I'm interested in hearing others' opinions as well so let's have a longer discussion about it next Monday when Luc is back.

nathanlb: The API (which is still WIP) will return the requested number of sessions up to the maximum number it finds. So if we request 2 preceding sessions but it only finds 1 or 0, then it will return 1 or 0.
I guess there could be a kind of paging system where the response contains the total number of sessions before and after to guide the client... 🤔. It would be more complex to implement but the ui could disable the buttons if it already knows there aren't any more sessions.

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: It's right there in the mockup. 😅 "Show preceding sessions", "Show following sessions". Hmmm I think i see a pattern here with your previous question. Maybe we interpreted it differently.
Did you see us requesting all sessions at once and putting them in the state, then showing/hiding them in the UI?
I was treating it as multiple requests.
Both would work.

If we fetch them all at once then we would set a default number of sessions before/after on load and that's that. But I'm not sure how we determine the best setting.

If we fetch new sessions based on user clicks then it's more requests but potentially lighter payloads and we don't have to guess.

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: Good question. TBH I think the easiest is to omit the case creation event from the timeline given that a timestamp is passed to the API as the ticketCreationDate. Using a timestamp to fetch events from the database is pretty easy but trying to match the salesforce timestamp with a specific event in the database sounds like a headache as I'm not sure they would necessarily match. So the ticket creation event would be omitted and on the UI side we trust that the timestamp supplied is the correct one and use that to mark the event in the timeline.

But we wouldn't have the searchhub like we see in the mockups.

What do you think?

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
drosenfeld-coveo: I like this idea but it brings up the questions when we pass in a date, do we expect to find a session that matches that date, what ive been doing is assuming its just a date and finding the closest user actions. Also how do you guys find the date of the UA you are looking for? because in snowflake they are pretty precise (to the milliseconds)

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: @drosenfeld-coveo here's a discussion that's in line with [your question](https://github.com/coveo/ui-kit/pull/2637#discussion_r1062941455)

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: Ok so I seem to have I misunderstood the UX design. What is prescribed by UX is to fetch 3 following sessions and 2 preceding sessions on load. That would be "All the timeline sessions". Then the UI can show/hide the sessions but it will already have all of them.
This sounds good to me so I would remove these two actions. Sorry for the confusion that's my bad.

mmitiche: Yes that's exactly what I thought, I thought that we are requesting all the sessions all at once.
Like you mentioned both would work, however I have some considerations I want to mention:
- The "Show preceding sessions" button is supposed to directly show **2** preceding  sessions and not display the preceding sessions one by one.
- The "Show following sessions" button is supposed to directly show **3** following  sessions and not display the following sessions one by one. (That's why I understood that we will fetch all the sessions at once, cause we know exactly how many sessions we wanna display). 
- If we go with second approach of fetching new sessions based on user clicks, wouldn't be better to consider having a loading state for every type of user sessions,  preceding, following and case creation session? Because as a user of the QuanticUserActionTimeline, when I click for example on  the "Show preceding sessions" button, it would feel a bit weird to see all the component loading again, I would expect to only the preceding sessions to be loading.

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
mmitiche: This sounds good to me, we can handle the case creation event from the UI, we would still have the headache of trying to display the case creation event in the right order among all the user actions, but I think it is possible.

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: Yeah your points are all super valid which is why i'm happy to do exactly what's in the mockups. We do one fetch of (main session + 2 before + 3 after). If it has fewer sessions it just has fewer sessions. Then in the quantic component one click shows all the preceding or all the following sessions with no load time. Easy, simple.

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: we can also just completely omit any case creation event from the timeline. Thinking about it i don't see it being relevant to show creation of other cases and the main case creation event in the UI will just be created from the timestamp.

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
lbergeron: @mmitiche I like your suggestion to organize the sessions this way. The API will be easier to consume indeed.

packages/headless/src/api/service/insight/user-actions/user-actions-request.ts
```diff
@@ -45,10 +40,8 @@ export const buildInsightUserActionsRequest = (
     ...baseInsightRequest(req, 'POST', 'application/json', '/useractions'),
     requestParams: {
       ticketCreationDate: params.ticketCreationDate,
-      numberSessionsBefore: params.numberSessionsBefore ?? 50,
-      numberSessionsAfter: params.numberSessionsAfter ?? 50,
-      maximumSessionInactivityMinutes:
```
lbergeron, line 50: Why did you remove this parameter?

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
```diff
@@ -0,0 +1,125 @@
...
+ * The UserActions controller is responsible for fetching user actions surrounding a given case creation event.
+ */
+export interface UserActions extends Controller {
+  /**
+   * Fetch the list of user actions surrounding the ticket creation.
+   */
+  fetchUserActions(): void;
+  /**
+   * The state of the UserActions controller.
+   * */
```
lbergeron, line 89:
```suggestion
  /**
   * The state of the UserActions controller.
   */
```

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
lbergeron: Considering the API parameter is named `ticketCreationDate`, it would actually make sense to me to have the API return a case creation event matching the provided date. If the provided reference point was arbitrary (case creation date or something else), then inserting the case creation event would become problematic.

Does it makes sense to you?

packages/headless/src/features/insight-user-actions/insight-user-actions-actions.ts
```diff
@@ -0,0 +1,69 @@
...
+
+interface RegisterUserActionsPayload {
+  ticketCreationDate: string;
+  numberSessionsBefore?: number;
+  numberSessionsAfter?: number;
+  excludedCustomActions?: string[];
+}
+
+const registerUserActionsPayloadSchema = {
+  ticketCreationDate: nonEmptyString,
```
lbergeron, line 24: We should validate the date format as well.

lbergeron (commented):
Looks very good so far!

packages/headless/src/api/service/insight/user-actions/user-actions-request.ts
nathanlb: We found out more about snowflake and how visits are determined. The data is already segmented by visitID which helps us a lot to determine sessions but "visits" are defined as 30 minutes of inactivity by the data-platform and we can't change that. So trying to work around an arbitrary value for `maximumSessionInactivityMinutes` seems like a total nightmare and I suggested we remove it.

lbergeron: Thanks for the explanation. I totally agree with you :+1:

packages/headless/src/features/insight-user-actions/insight-user-actions-actions.ts
nathanlb: I added an option to validate ISO datetime strings in Bueno. [See here](https://github.com/coveo/ui-kit/pull/2637/commits/0fa788fdb05e8c22d2199c5eadfd71473efb6bcf#diff-77f7c6dbefaa1941ae6e15438b7aa4c0dc17d9da6e387f5b39f0aabc6b99d36c). This way we could use it in quantic as well to validate the component prop. What do you think?

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
nathanlb: As discussed on slack we'll go with the suggested format. The PR has already been updated.

packages/headless/src/features/insight-user-actions/insight-user-actions-actions.ts
lbergeron: I didn't review all the code, but the validator part looks good.

packages/headless/src/api/service/insight/user-actions/user-actions-response.ts
```diff
@@ -1,23 +1,41 @@
+export enum UserActionType {
```
lbergeron, line 1: I guess we'll need to add another type for "Ticket created".

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
```diff
@@ -0,0 +1,125 @@
...
+  options: UserActionsOptions;
+}
+
+export interface UserActionsOptions {
+  /**
+   * The ticket creation date in the [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html) format.
+   */
+  ticketCreationDate: string;
+  /**
+   * The number of sessions to include occuring before the ticket creation session.
```
lbergeron, line 32:
```suggestion
   * The number of sessions to include occurring before the ticket creation session.
```

packages/headless/src/controllers/insight/user-actions/headless-user-actions.ts
```diff
@@ -0,0 +1,125 @@
...
+   * @defaultValue `0`
+   */
+  numberSessionsBefore?: number;
+  /**
+   * The number of sessions to include occuring after the ticket creation session.
```
lbergeron, line 37:
```suggestion
   * The number of sessions to include occurring after the ticket creation session.
```

packages/headless/src/features/insight-user-actions/insight-user-actions-slice.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    ],
+  };
+
+  it('should have an initial state', () => {
+    expect(insightUserActionsReducer(undefined, {type: 'foo'})).toEqual(
+      getInsightUserActionsInitialState()
+    );
+  });
+
+  describe('#fetcUserActions', () => {
```
lbergeron, line 43:
```suggestion
  describe('#fetchUserActions', () => {
```

packages/headless/src/features/insight-user-actions/insight-user-actions-slice.test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+        expect(modifiedState.timeline).toBeUndefined();
+      });
+    });
+
+    describe('when fetching the user actions succeeds', () => {
+      const response: InsightUserActionsResponse = {
+        timeline: {
+          precedingSessions: [session],
+          session: session,
```
lbergeron, line 107: Is that really expected that the API returns the same session twice?

packages/headless/src/features/insight-user-actions/insight-user-actions-slice.test.ts
nathanlb: your ability to spot these is unmatched! sorry for the typos 😅

packages/headless/src/features/insight-user-actions/insight-user-actions-slice.test.ts
nathanlb: No not at all I just reused the fixture.

packages/headless/src/features/insight-user-actions/insight-user-actions-slice.test.ts
nathanlb: the original function actually had that typo but the search&replace doesn't go through 'strings' unfortunately.

packages/headless/src/api/service/insight/user-actions/user-actions-response.ts
nathanlb: Yes good point

PR #2642, opened 2023-01-06
feat(headless): made search box only update enableQuerySyntax when explicitly set

btaillon:
The issue: in Atomic, it's currently impossible for a user to submit a query using the query syntax with `atomic-search-box`.

The cause of this issue is that Headless's search box controller takes `enableQuerySyntax` as an option (optional) and **always** updates `engine.state.query.enableQuerySyntax` when calling `submit`.

There's two obvious solutions to the problem:
1. We add an `enableQuerySyntax` option to the `atomic-search-box`.
2. We make Headless search boxes only update `enableQuerySyntax` when it was explicitly set as an option.

In this PR, I want with solution 2, because:
1. It's possible to globally update `enableQuerySyntax` by either dispatching `updateQuery` or through the url/searchparams managers. I don't know if we want to encourage either of those, but it seems about as intuitive to assume that it's a global property than to assume it's not, and we don't have any warning not to (that I'm aware of).
2. The alternative solution would probably require adding some more documentation, so it's quicker for me to make this change.

I didn't add tests yet in case you prefer solution 1.

https://coveord.atlassian.net/browse/KIT-2201

### DIFF
--- a/packages/headless/src/controllers/core/search-box/headless-core-search-box-options.ts
+++ b/packages/headless/src/controllers/core/search-box/headless-core-search-box-options.ts
@@ -45,11 +45,10 @@ export interface SearchBoxOptions {
 
 type DefaultSearchBoxOptions = Pick<
   SearchBoxOptions,
-  'enableQuerySyntax' | 'numberOfSuggestions' | 'clearFilters'
+  'numberOfSuggestions' | 'clearFilters'
 >;
 
 export const defaultSearchBoxOptions: Required<DefaultSearchBoxOptions> = {
-  enableQuerySyntax: false,
   numberOfSuggestions: 5,
   clearFilters: true,
 };
--- a/packages/headless/src/controllers/core/search-box/headless-core-search-box.ts
+++ b/packages/headless/src/controllers/core/search-box/headless-core-search-box.ts
@@ -173,7 +173,7 @@ export function buildCoreSearchBox(
   const getState = () => engine.state;
 
   const id = props.options?.id || randomID('search_box');
-  const options: Required<SearchBoxOptions> = {
+  const options = {
     id,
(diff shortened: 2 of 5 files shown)


packages/headless/src/controllers/core/search-box/headless-core-search-box-options.ts
```diff
@@ -45,11 +45,10 @@ export interface SearchBoxOptions {
 
 type DefaultSearchBoxOptions = Pick<
   SearchBoxOptions,
-  'enableQuerySyntax' | 'numberOfSuggestions' | 'clearFilters'
+  'numberOfSuggestions' | 'clearFilters'
 >;
 
 export const defaultSearchBoxOptions: Required<DefaultSearchBoxOptions> = {
-  enableQuerySyntax: false,
```
olamothe, line 52: One thing that's important to keep is that query syntax should be disabled by default for 99% of use cases.

We need to make it so that implementers have to go "out of their way" to have a search box where query syntax is enabled. We don't want to end up with implementers having to always specify that they want query syntax disabled.

Leaving the property `undefined` when it's sent to the API (when the option is not set explicitly on the controller) means that query syntax will be enabled at the API level. We do not want that.

I think it would be better to add an option on atomic search box.

As for why we want query syntax disabled in the vast majority of case, it's because query syntax is "all or nothing". And when you submit some special characters, the query syntax will kick in at the API and index level, and it's easy to get 4XX errors due to invalid query syntax.

For most of our search interface that will be used by "normal non Coveo knowledgeable" end user, it's easy to see that someone could add `[` or `(`, `@` etc. characters in the search box without knowing that they are query syntax sensitive character, and they would get back a cryptic error with no way of knowing how to fix it.

Said differently, query syntax in the search box is a power user feature.

PR #2643, opened 2023-01-09
feat(atomic): add quickview hit highlight preview bar

olamothe:
Add support for hit highlight preview in quickview.

image

It's the feature on the right hand side that adds markers in the scroll bar that identifies where in the documents the keywords are located. This kinda mimics what happens when you CTRL+F in Chrome.

This rework a bit what was done for the `quickview-iframe` and `quickview-sidebar`.

* Remove the code to auto resize the iframe, and instead force the modal row to always be 100% with grid template. This means that the scrollbar + hit preview will be much better "in sync" when you navigate the keywords with the sidebar on the left. The scrollbar in the quickview will exactly match the markers.

* Lift the `words` calculation from the `quickview-sidebar` component and instead move it to be a stateful property in the modal itself. The calculation of the map of all words to highlights + and all the associated HTMLElements is then passed as a prop to `quickview-sidebar` and `preview bar`.

* To ensure that the above point works well, we need to make sure that once a word map has been calculated, we do not needlessly overwrite the iframe content. Before this PR, I realized that the iframe content was being re-written when we were using the "enable/disable" highlight feature, for example. Previously the only thing that would happen is a small performance hit. With these new changes however, we cannot do this, as the HTMLElements reference in the word map will change on rewrite, and break the keyword navigation feature.

* To make sure we do not rewrite the iframe needlessly, I added an hidden div containing the result unique ID inside the iframe. When it is already present in the iframe, we can skip rewriting content.

* The preview bar is embedded/written to the iframe. This ensure that the calculation for the height of each preview unit is easy/straightforward. Unfortunately on the flip side, this means that we cannot really use JSX. So it's essentially just a function that receives the iframe as a reference and writes directly to it.

https://coveord.atlassian.net/browse/KIT-2200


packages/atomic/src/utils/color-utils.ts
```diff
@@ -0,0 +1,68 @@
+export const rgbToHsv = (r: number, g: number, b: number) => {
```
olamothe, line 1: This is imported from JSUI, which itself had imported this from some stack overflow post.

It's some generic color utility function. 
We use it for the preview bar to add a nice contrast border around each hit marker (the `saturatedColor`)

packages/atomic/src/components/search/result-template-components/quickview-word-highlight/quickview-word-highlight.tsx
```diff
@@ -7,9 +8,10 @@ export class QuickviewWordHighlight {
   public occurrences = 0;
   public color: string;
   public invertColor: string;
+  public saturateColor: string;
```
btaillon, line 11: I think this should be named `saturatedColor`

btaillon: what I would like even more would be if we could rename these variables after their function rather than their color (e.g., `color`, `focusedColor`, etc)

packages/atomic/src/components/search/result-template-components/quickview-preview-bar/quickview-preview-bar.ts
```diff
@@ -0,0 +1,57 @@
...
+  previewUnit.style.position = 'absolute';
+  previewUnit.style.top = `${(elementPosition / docHeight) * 100}%`;
+  previewUnit.style.width = '100%';
+  previewUnit.style.height = '1px';
+  previewUnit.style.border = `1px solid ${word.saturateColor}`;
+  previewUnit.style.backgroundColor = word.color;
+  return previewUnit;
+};
+
+export const buildQuickviewPreviewBar = (
```
btaillon, line 30: IMO I would prefer a name like `buildQuickviewScrollbarHighlights` or `buildQuickviewHighlightedScrollbar`, but I don't feel strongly about it.

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -75,17 +57,34 @@ export class AtomicQuickviewModal implements InitializableComponent {
...
+          />
+        </div>
+        <div class="overflow-auto relative">
           <QuickviewIframe
+            result={this.result}
             content={this.content}
-            onSetIframeRef={(ref) => (this.iframeRef = ref)}
+            onSetIframeRef={(ref) => {
+              this.iframeRef = ref;
+              setTimeout(() => {
```
btaillon, line 79: Is this because the height of the elements/iframe aren't correct until we wait a bit?

If so, I would prefer if we somehow named this delay. E.g.:
```js
this.iframeRef = ref;
await this.waitForIFrameHeightCalculations(); // Just returns a `new Promise<void>(resolve => setTimeout(resolve));`
this.words = getWordsHighlights(
  this.termsToHighlight,
  this.iframeRef
);
```

btaillon: Alternatively, we could abstract away the iframe when it exits `QuickviewIframe`. `QuickviewIframe` could return the bare minimum (e.g., `document` and/or `documentWriter`) instead of the iframe, and it could also take care of waiting for height calculations to be done before calling this callback prop.

E.g.:
```tsx
<QuickviewIframe
  result={this.result}
  content={this.content}
  onContentRendered={({ renderedContent, writer }) => {
    this.renderedContent = renderedContent;
    this.writer = writer;
    this.words = getWordsHighlights(
      this.termsToHighlight,
      this.writer
    );
  }}
/>
```

packages/atomic/src/components/search/result-template-components/quickview-preview-bar/quickview-preview-bar.ts
```diff
@@ -0,0 +1,57 @@
+import {QuickviewWordHighlight} from '../quickview-word-highlight/quickview-word-highlight';
+
+const buildPreviewBar = (documentWriter: Document) => {
+  const bar = documentWriter.createElement('div');
+  bar.style.position = 'fixed';
+  bar.style.top = '0';
+  bar.style.right = '0';
+  bar.style.width = '15px';
+  bar.style.height = '100%';
```
btaillon, line 9: for accessibility, we can probably just add `aria-hidden="true"` to this, since screen reader users are already going to be able to navigate using the next/previous buttons.

btaillon (approved):
awesome! I made some suggestions, but this is pretty easy code to read.

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
olamothe: It's because when you `document.write` and `document.close` on a document (iframe included), it's not synchronous. 

It's written at the end of the current call stack.

So before you can use `querySelector` on it and have it contains the right content to calculate the word map, we need to push it back.

I can add a name for the wait 👍

packages/atomic/src/components/search/result-template-components/quickview-preview-bar/quickview-preview-bar.ts
olamothe: Ok, I like the shorter name, and would be tempted to leave it like this.

But if other people have an opinion on it I will change it.

PR #2648, opened 2023-01-11
feat(atomic): add feature to disable specific keywords highlights 

olamothe:
image

Add a feature to disable specific keywords highlighting. In the above screenshot,  the keyword "for" is disabled, and so will not appear highlighted in the iframe, nor will it appear in the hit preview sidebar.

This modifies the stateful `highlightKeywords` property from a simple boolean to an object which contains the map of all current keyword, their index identifier and wether they are enabled or not.

One particular request from Johnny is that the following should be respected.

1- Search for `Coveo for salesforce`
2- Remove highlighting on the keyword `for`
3- Navigate through quickviews using the pagination at the bottom OR close the Quickview and open another one for the same query: the keyword `for` should still be disabled.
4- Change the query to `Coveo for commerce`, obtain a new set of result, and open a new quickview
5- The keyword `for` should now be re-enabled.

In order for # 5 to work, I added a `reset()` method on the quickview modal, which gets called when the quickview result component (the icon) gets unloaded from the DOM.

https://coveord.atlassian.net/browse/KIT-2206


ThibodeauJF:
Screen Shot 2023-01-11 at 11 31 42 AM
The button's height and icon alignment is a bit weird on my end

ThibodeauJF:
Once I click "-" I can't instantly reclick "+" the button is disabled (while the hover state is active)
Screen Shot 2023-01-11 at 11 33 31 AM

ThibodeauJF:
For the "How to" article, I get up to the page 2 on 10, but no more
Screen Shot 2023-01-11 at 11 37 14 AM

louis-bompart (approved):

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -86,47 +92,86 @@ const HighlightKeywordsCheckbox: FunctionalComponent<
...
+  > & {
     words: Record<string, QuickviewWordHighlight>;
   }
-> = ({words, i18n}) => {
+> = ({words, i18n, highlightKeywords, onHighlightKeywords}) => {
   return (
     <Fragment>
       {Object.values(words).map((keyword) => {
+        const wordIsEnabled =
+          highlightKeywords.highlightAll &&
```
btaillon, line 107: `highlightAll` implies to me that a value of `true` would mean that all keywords are highlighted.

In this case, I think the name `enabled` or flipping it to something like `highlightNone` would be more intuitive.

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -86,47 +92,86 @@ const HighlightKeywordsCheckbox: FunctionalComponent<
...
-                buttonStyle="text-transparent"
-                class="border-0"
-                labelI18nKey="previous"
-                tooltip={i18n.t('previous')}
-                icon={ArrowUp}
-                clickCallback={() => keyword.navigateBackward()}
-              ></atomic-icon-button>
+            <div
+              class={`flex grow items-center bg-background border border-neutral rounded-lg overflow-x-auto ${
+                !wordIsEnabled ? 'pointer-events-none opacity-50' : ''
```
btaillon, line 118: What is `pointer-events-none` for in this case?

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -86,47 +92,86 @@ const HighlightKeywordsCheckbox: FunctionalComponent<
...
+            <atomic-icon-button
+              class={`${
+                highlightKeywords.highlightAll
+                  ? ''
+                  : 'pointer-events-none opacity-50'
+              }`}
+              buttonStyle="text-transparent"
+              icon={wordIsEnabled ? Remove : Add}
+              labelI18nKey={
+                wordIsEnabled ? 'quickview-remove-word' : 'quickview-add-word'
```
btaillon, line 160: I prefer to avoid changing the `aria-label` of something that toggles, I recall getting accessibility issues about this a long time ago. You set the key to always be `quickview-add-word` and add `aria-pressed="true/false"` (not `undefined`, since a value of `false` will communicate to screen readers that this is an unpressed toggle button).

olamothe:
> Kapture.2023-01-11.at.11.35.38.mp4


Known issue that needs to be fixed in Headless: https://coveord.atlassian.net/browse/KIT-2198

olamothe:
> For the "How to" article, I get up to the page 2 on 10, but no more Screen Shot 2023-01-11 at 11 37 14 AM

What do you mean by "How to article" ?

In any case, not all documents in the index possess a Quickview. You cannot access all documents.

It's possible that there's only one available in a given result set, and in this case the pager indeed becomes useless. That's a good point.

olamothe:
> Once I click "-" I can't instantly reclick "+" the button is disabled (while the hover state is active) Screen Shot 2023-01-11 at 11 33 31 AM
> 
> I guess it's out of scope (a todo)?

You sure about that ? Working well here.

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
olamothe: So that there is no mouse interaction when a word is disabled. For example, this disables all visual hover effect on buttons.

btaillon: does this prevent the user from selecting the text and copying it?

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -169,6 +221,21 @@ export class AtomicQuickviewModal implements InitializableComponent {
     return this.bindings.engine.state.search.response.termsToHighlight;
   }
 
+  private handleHighlightsScripts() {
+    if (this.highlightKeywords.highlightNone) {
+      this.enableHighlights();
+    } else {
+      this.disableHighlights();
+    }
```
btaillon, line 229: I'm kind of confused by this condition 🤔 is it inverted?

btaillon (approved):

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
olamothe: forgot it when I inverted the meaning of the flag per your suggestion.

Fixing

PR #2650, opened 2023-01-11
fix(headless,atomic,atomic-react,atomic-angular,quantic): made query pipeline update with jwt tokens

btaillon:
https://coveord.atlassian.net/browse/KIT-2207

# The problem
Access tokens can be [JWTs](https://en.wikipedia.org/wiki/JSON_Web_Token), which amongst other things, can (optionally) contain:
* A query pipeline
* A search hub
* A user display name

Prior to this PR, we were parsing the access token whenever somebody tries to update either 3 of these values, and replacing them with the value from the search token if it exists.

Unfortunately, this meant that only actions which are meant to update these some of these three values would update them to match the token's values.

When initializing the engine, if no `search` object in the configuration was included, then the pipeline and search hub would not be updated to match the token. If no `analytics` object in the configuration was included, then the user display name was not updated to match the token.

A temporary fix for this issue is to add an empty `search` and `analytics` object to the configuration, as you can see here: https://p8968n.csb.app/

# The fix
I made the action responsible for setting the access token (and also responsible for renewing the access token) also update the three properties.

See the diff from the Redux dev tools, where the action only has an access token and the org as its payload:
Screen Shot 2023-01-11 at 2 47 19 PM

An alternative fix would've been to always dispatch the actions responsible for updating the query pipeline, search hub and user display name, but I felt like it made more logical sense to always update them with the token.


packages/headless/src/app/search-engine/jwt-reducer.ts
```diff
@@ -32,10 +33,6 @@ const possiblyWarnOnMismatch = (
     return;
   }
 
-  if (stateProp === tokenValue) {
-    return;
-  }
```
btaillon, line 37: Since the state prop will usually already be updated with the token's value, this would prevent us from warning the user that they are trying to set a different pipeline/searchhub than the one configured in the token.

packages/headless/src/app/search-engine/jwt-reducer.ts
```diff
@@ -109,10 +116,16 @@ const handleMismatchOnSearchHub = (
     payload,
     logger
   );
-  if (shouldReconcileValues(jwt.searchHub, state.searchHub)) {
-    state.searchHub = jwt.searchHub!;
-  }
+  return updateSearchHub(jwt, state);
```
btaillon, line 119: I didn't remove the logic that updated the pipeline/etc from the actions responsible for updating them directly, since their own reducers aren't aware that the access token contains the "correct" value, and would update them.

ThibodeauJF (approved):

packages/headless/src/app/search-engine/jwt-reducer.test.ts
```diff
@@ -12,10 +13,17 @@ import {getSearchHubInitialState} from '../../features/search-hub/search-hub-sta
 import {createMockState} from '../../test';
 import {jwtReducer} from './jwt-reducer';
 
+function cloneDeep<T>(value: T): T {
+  return JSON.parse(JSON.stringify(value));
+}
+
```
louis-bompart, line 19: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone ?

louis-bompart (approved):

packages/headless/src/app/search-engine/jwt-reducer.test.ts
btaillon: > packages/headless/src/app/search-engine/jwt-reducer.test.ts:17:10 - error TS2304: Cannot find name 'structuredClone'.

:(

olamothe (approved):

PR #2651, opened 2023-01-11
feat(atomic): add ipx footer items + small changes

jbvouma:
[SVCINT-1950](https://coveord.atlassian.net/browse/SVCINT-1950)

Added some footer items similar to the ones in [figma](https://www.figma.com/file/gvajjJbCnGu7L8ZeeWYCgf/IPX-Builder?node-id=185%3A59049&t=56DGKuVIlzxndenR-0) + some small changes in the header and the pagination.

**Before**
image

**After**
image

[SVCINT-1950]: https://coveord.atlassian.net/browse/SVCINT-1950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/pages/examples/ipx.html
```diff
@@ -189,12 +189,32 @@
...
+          />
```
jbvouma, line 216: It's the coveo logo. I'm not to sure whether to keep it here.

louis-bompart: 🤔 should we add it to our known icons maybe?

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -189,12 +189,32 @@
...
+              color: var(--atomic-primary);
+              text-decoration: none;
+            }
+
+            .ipx-footer-icon {
+              aspect-ratio: auto;
+              width: 5rem;
+              height: 1.25rem;
+            }
+          </style>
```
louis-bompart, line 208: notsure. this is a publi repo, is it smart to mention coveo@coveo there?

packages/atomic/src/pages/examples/ipx.html
ThibodeauJF: We could add it to the assets perhaps yes, but optional, shipping it to users is not really important

packages/atomic/src/pages/examples/ipx.html
ThibodeauJF: **Ideally** it would work anywhere yes, not just for internal folks, but it can always be done later... I don't think the presence of Coveo@Coveo is necessarily a mistake, depends on your requirements for IPX

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
```diff
@@ -19,6 +19,11 @@ export class AtomicIPXButton {
    */
   @Prop({reflect: true}) public label = 'Workplace Search';
 
+  /**
+   * The button icon.
+   */
+  @Prop({reflect: true}) public icon = '';
```
ThibodeauJF, line 25:
```suggestion
  @Prop({reflect: true}) public icon?: string;
```
Then the check on line 46 can just be `this.icon ?`

ThibodeauJF (approved):

packages/atomic/src/pages/examples/ipx.html
jbvouma: It's just an example of what we want for clients. It can be replaced by something else.

packages/atomic/src/pages/examples/ipx.html
jbvouma: I'll keep it here since it's just a simple example.

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -38,7 +38,7 @@
             "
             section="search"
           >
-            <atomic-search-box></atomic-search-box>
+            <atomic-search-box style="padding-bottom: 0.875rem"></atomic-search-box>
```
btaillon, line 41: not super important since this is just an example, but I think we should generally avoid using `style=""` and instead use classes and style tags.

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
```diff
@@ -38,14 +43,23 @@ export class AtomicIPXButton {
...
         <span part="button-icon">
-          <atomic-icon
-            part="ipx-close-icon"
-            icon={this.getIcon(CloseIcon)}
-          ></atomic-icon>
-          <atomic-icon
-            part="ipx-search-icon"
-            icon={this.getIcon(SearchIcon)}
-          ></atomic-icon>
+          {this.icon ? (
```
btaillon, line 46: Instead of having multiple conditions here and rendering multiple buttons, I think it would be easier to read if `CloseIcon` and `SearchIcon` were passed down from the parent component to this component's `icon` property, and if the `part` was a prop.

dguerinCoveo (approved):

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
```diff
@@ -38,14 +43,23 @@ export class AtomicIPXButton {
...
+            [
+              <atomic-icon
+                part="ipx-close-icon"
+                icon={this.getIcon(CloseIcon)}
+              ></atomic-icon>,
+              <atomic-icon
+                part="ipx-search-icon"
+                icon={this.getIcon(SearchIcon)}
+              ></atomic-icon>,
+            ]
```
btaillon, line 61: We generally need to add keys to children of an array when we render an array, I'm not sure what are the exceptions.

For this reason, when I want to render multiple items like this, I prefer to wrap them in a `<Fragment>`.

olamothe (approved):

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
jbvouma: I like the idea to have these two as prop of the parent. I'll try it

pathiery (commented):
same comments as @btaillon

packages/atomic/src/pages/examples/ipx.html
jbvouma: Done ✅ !

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
jbvouma: I removed the condition.

pathiery (approved):

louis-bompart (approved):

PR #2652, opened 2023-01-11
feat(atomic): add new prop to allow to control the iframe sandbox attribute

olamothe:
https://coveord.atlassian.net/browse/KIT-2203

### DIFF
--- a/packages/atomic/src/components/search/result-template-components/atomic-quickview/atomic-quickview.tsx
+++ b/packages/atomic/src/components/search/result-template-components/atomic-quickview/atomic-quickview.tsx
@@ -38,15 +38,15 @@ export class AtomicQuickview implements InitializableComponent {
   public quickviewState!: QuickviewState;
 
   /**
-   * The sandbox attribute to apply to the quickview iframe.
+   * The `sandbox` attribute to apply to the quickview iframe.
    *
-   * The quickview is loaded inside an iframe with a [sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) attribute for security reasons.
+   * The quickview is loaded inside an iframe with a [`sandbox`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) attribute for security reasons.
    *
-   * This attribute primarily exists in order to protect against potential XSS attacks that could originate from the document being displayed.
+   * This attribute exists primarily to protect against potential XSS attacks that could originate from the document being displayed.
    *
    * By default, the sandbox attributes are: `allow-popups allow-top-navigation allow-same-origin`.
    *
-   * `allow-same-origin` is not optional, and must always be included in the list of allowed capabilities in order for the component to function properly.
+   * `allow-same-origin` is not optional, and must always be included in the list of allowed capabilities for the component to function properly.
    */
   @Prop() public sandbox =
     'allow-popups allow-top-navigation allow-same-origin';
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/search/result-template-components/atomic-quickview/atomic-quickview.tsx
```diff
@@ -37,6 +37,20 @@ export class AtomicQuickview implements InitializableComponent {
   @State()
   public quickviewState!: QuickviewState;
 
+  /**
```
olamothe, line 40: @jpmarceau 🙇 Thanks !

louis-bompart (commented):
> `allow-same-origin` is not optional, and must always be included in the list of allowed capabilities in order for the component to function properly.

Should we dummy proof?
  a. Add it if the user forgot it
  b. Throw a warning/error to the user
  c. All the above
  d. Obi-Wan Kenobi

ThibodeauJF (approved):

packages/atomic/src/components/search/result-template-components/atomic-quickview/atomic-quickview.tsx
```diff
@@ -37,6 +37,20 @@ export class AtomicQuickview implements InitializableComponent {
...
+   * The sandbox attribute to apply to the quickview iframe.
+   *
+   * The quickview is loaded inside an iframe with a [sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) attribute for security reasons.
+   *
+   * This attribute primarily exists in order to protect against potential XSS attacks that could originate from the document being displayed.
+   *
+   * By default, the sandbox attributes are: `allow-popups allow-top-navigation allow-same-origin`.
+   *
+   * `allow-same-origin` is not optional, and must always be included in the list of allowed capabilities in order for the component to function properly.
```
jpmarceau, line 49: Clear and well-written description Oli 👌 
I only have minor suggestions:

```suggestion
   * The `sandbox` attribute to apply to the quickview iframe.
   *
   * The quickview is loaded inside an iframe with a [`sandbox`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) attribute for security reasons.
   *
   * This attribute exists primarily to protect against potential XSS attacks that could originate from the document being displayed.
   *
   * By default, the sandbox attributes are: `allow-popups allow-top-navigation allow-same-origin`.
   *
   * `allow-same-origin` is not optional, and must always be included in the list of allowed capabilities for the component to function properly.
```

jpmarceau (approved):

packages/atomic/src/components/search/result-template-components/atomic-quickview/atomic-quickview.tsx
```diff
@@ -37,6 +37,20 @@ export class AtomicQuickview implements InitializableComponent {
...
+  /**
+   * The `sandbox` attribute to apply to the quickview iframe.
+   *
+   * The quickview is loaded inside an iframe with a [`sandbox`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) attribute for security reasons.
+   *
+   * This attribute exists primarily to protect against potential XSS attacks that could originate from the document being displayed.
+   *
+   * By default, the sandbox attributes are: `allow-popups allow-top-navigation allow-same-origin`.
+   *
+   * `allow-same-origin` is not optional, and must always be included in the list of allowed capabilities for the component to function properly.
```
btaillon, line 49: I think we should also add validation for that.

btaillon (approved):

PR #2655, opened 2023-01-12
test(atomic): fixed flaky breadcrumb assertion

btaillon:
https://coveord.atlassian.net/browse/KIT-2220

There might be more flaky tests, but overwhelmingly I saw this assertion fail a lot.

The definition of `breadcrumbClearFacetValueButton` is the following:
```js
{
  breadcrumbClearFacetValueButton: () =>
    BreadboxSelectors.breadcrumbButton().find('atomic-icon')
}
```

The test was asserting that `breadcrumbClearFacetValueButton` did not exist, which means that Cypress first asserted that `breadcrumbButton` exists, which it does temporarily after deselecting a facet but only for a short moment, which I believe caused tests to *sometimes* work.

This assertion is done in all facet test files.


packages/atomic/cypress/e2e/breadbox.cypress.ts
```diff
@@ -208,16 +208,20 @@ describe('Breadbox Test Suites', () => {
...
 
       describe('verify accessibility', () => {
+        beforeEach(setupFacetWithMultipleSelectedValuesAndShowMore);
+
         BreadboxAssertions.assertFocusBreadcrumb('@numberOfVisibleButtons');
       });
 
       describe('verify rendering', () => {
+        before(setupFacetWithMultipleSelectedValuesAndShowMore);
+
```
btaillon, line 224: This fix is unrelated to the main issue, but it also failed in this PR. A test failed because it said there was no registered alias for `@numberOfVisibleButtons`.

I couldn't quite figure out why the alias is sometimes registered and sometimes not, but I found [a recommendation from Cypress to use the `beforeEach` hook instead](https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Aliases-are-reset-before-each-test), so I'm trying that.

louis-bompart (approved):

packages/atomic/cypress/e2e/search-box/search-box-selectors.ts
```diff
@@ -8,6 +8,8 @@ export const SearchBoxSelectors = {
   submitButton: () =>
     SearchBoxSelectors.shadow().find('[part="submit-button"]'),
   querySuggestions: () => SearchBoxSelectors.shadow().find('[data-query]'),
+  querySuggestion: (query: string) =>
+    SearchBoxSelectors.shadow().find(`[data-query="${query}"]`),
```
btaillon, line 12: Another test failed, this time it complained that the suggestion at index 1 was not "Recent query 1", but instead "Recent query 0".

I'm guessing this is just something weird going on with the order in which elements are appended in the DOM and I'm eliminating the index from the test.

This is, as far as I'm aware, the only test dependent on the order of suggestions, so if you prefer, I can undo those changes.

olamothe (approved):

PR #2656, opened 2023-01-12
feat(quantic): the quantic result field component created

mmitiche:
[SFINT-4767](https://coveord.atlassian.net/browse/SFINT-4767)

#### - According to the type property, this component renders one of the following: `QuanticResultText`, `QuanticResultNumber`, `QuanticResultDate`, `QuanticResultMultiValueText`

[SFINT-4767]: https://coveord.atlassian.net/browse/SFINT-4767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultField/quanticResultField.js
```diff
@@ -0,0 +1,57 @@
...
+  render() {
+    if (this.type === 'multi') {
+      return multiValueTemplate;
+    } else if (this.type === 'number') {
+      return numberTemplate;
+    } else if (this.type === 'date') {
+      return dateTemplate;
+    }
+    return stringTemplate;
+  }
```
erocheleau, line 56:
```suggestion
const templateTypeMap = {
  'multi' : multiValueTemplate,
  'number': numberTemplate,
  'date': dateTemplate,
  'string': stringTemplate,
  'default': stringTemplate,
};

render() {
  return templateTypeMap[this.type] || templateTypeMap['default'];
}
```

What do you think?

packages/quantic/force-app/main/default/lwc/quanticResultField/__tests__/quanticResultField.test.js
```diff
@@ -0,0 +1,96 @@
...
+  },
+  {
+    name: 'number',
+    element: 'c-quantic-result-number',
+  },
+  {
+    name: 'multi',
+    element: 'c-quantic-result-multi-value-text',
+  },
+];
```
erocheleau, line 22: I would add a random type to test the default mapping.

erocheleau (commented):
Proposed improvement

packages/quantic/force-app/main/default/lwc/quanticResultField/quanticResultField.js
nathanlb: With that object you could also add a prop validation check like
```javascript
const templateTypes = Object.keys(this.templateTypeMap)
if (!templateTypes.includes(this.type)) {
  this.error = `The provided type ${this.type} is invalid. The type must be one of ${templateTypes.join(' | '}`;
}
```

lbergeron (commented):
Nice!

packages/quantic/force-app/main/default/lwc/quanticResultField/quanticResultField.js
mmitiche: Sure yeah! thank you 👌

packages/quantic/force-app/main/default/lwc/quanticResultField/quanticResultField.js
```diff
@@ -0,0 +1,69 @@
...
+    number: numberTemplate,
+    date: dateTemplate,
+    string: stringTemplate,
+    default: stringTemplate,
+  };
+
+  connectedCallback() {
+    const templateTypes = Object.keys(this.templateTypeMap).slice(0, 4);
+    if (!templateTypes.includes(this.type)) {
+      console.error(
```
nathanlb, line 58: Just saying but if instead of the multiple template files we had one quanticResultField.html with some conditional rendering we could display the error in red like we do for many other components. 👀

nathanlb (approved):

packages/quantic/force-app/main/default/lwc/quanticResultField/quanticResultField.js
lbergeron: ...or we could show an error template in such case. I like how the component uses specific templates, but I don't have strong opinions on that matter.

lbergeron (approved):

packages/quantic/force-app/main/default/lwc/quanticResultField/quanticResultField.js
mmitiche: I will do that! a new template for the error to get the best of both words!

nathanlb: Just for my own knowledge, what's the advantages you like of the templates?

lbergeron: In this situation it looks clean to me to have separate templates. I'd say it is more readable. Each template is super simple and its scope is well defined. It is very easy to know what is going to be rendered.

When everything is done in a single template, it is often more difficult to get the same understanding of the component.

For instance, if you have a single template, it is difficult for a developer to know whether an error message can be displayed at the same time as the value. With separate templates, it is obvious.

lbergeron: I could easily imagine situations where using separate templates would not make sense. It is a matter of using the right tool for the job.

nathanlb: That's valid. I think i have an aversion to files with 1 or 2 lines in them

nathanlb: I'm fine with the templates so just to try to illustrate my personal feelings about it I'd compare it to putting each pair of my socks in a separate drawer. I can label the drawers and it's easy to find the socks I want but it's cumbersome. 😛

lbergeron: I like the analogy :smile:. Honestly, I don't mind whether the component uses a single template or many.

nathanlb: Neither do I really we can keep it as it is now. 👍

PR #2657, opened 2023-01-12
feat(headless): reset preview content on new query

olamothe:
My initial thought was to add a new action to clear the preview content from the state.

This would be used by UI's when they want to "clear" the Quickview from the page, and don't need it loaded in memory anymore.

However I realized we could also just clear it automatically for our users whenever there is a new query executed, since... that's pretty much when you would want to unload any displayed Quickview from a page anyway...

Said differently, you load a quickview from a given list of result currently displayed to users.

When a new query is performed (and a new set of results is retrieved), then keeping the Quickview content in memory seems superfluous and we can just clear it automatically.

https://coveord.atlassian.net/browse/KIT-2198

### DIFF
--- a/packages/headless/src/features/result-preview/result-preview-slice.ts
+++ b/packages/headless/src/features/result-preview/result-preview-slice.ts
@@ -18,17 +18,10 @@ export const resultPreviewReducer = createReducer(
         state.isLoading = false;
       })
       .addCase(updateContentURL.fulfilled, (state, action) => {
-        const {contentURL} = action.payload;
-
-        state.contentURL = contentURL;
+        state.contentURL = action.payload.contentURL;
       })
       .addCase(executeSearch.fulfilled, (state) => {
-        const {content, isLoading, uniqueId, contentURL} =
-          getResultPreviewInitialState();
-        state.content = content;
-        state.isLoading = isLoading;
-        state.uniqueId = uniqueId;
-        state.contentURL = contentURL;
+        Object.assign(state, getResultPreviewInitialState());
       });
   }
 );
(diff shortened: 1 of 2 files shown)


packages/headless/src/features/result-preview/result-preview-slice.ts
```diff
@@ -19,6 +20,14 @@ export const resultPreviewReducer = createReducer(
...
 
+        state.contentURL = contentURL;
+      })
+      .addCase(executeSearch.fulfilled, (state) => {
+        const {content, isLoading, uniqueId, contentURL} =
+          getResultPreviewInitialState();
+        state.content = content;
+        state.isLoading = isLoading;
+        state.uniqueId = uniqueId;
         state.contentURL = contentURL;
```
louis-bompart, line 31: I think destructuring would be more readable here (suggestion WIP, prettier incoming)

```suggestion
      .addCase(executeSearch.fulfilled, (state) => {
        ({content: state.content, isLoading: state.isLoading, uniqueId: state.uniqueId, contentURL: state.contentUrl} =
          getResultPreviewInitialState());
```

packages/headless/src/features/result-preview/result-preview-slice.ts
```diff
@@ -19,6 +20,14 @@ export const resultPreviewReducer = createReducer(
       .addCase(updateContentURL.fulfilled, (state, action) => {
         const {contentURL} = action.payload;
 
+        state.contentURL = contentURL;
```
louis-bompart, line 23:
```suggestion
        ({contentURL:state.contentURL} = action.payload);
```

louis-bompart (approved):
LGTM, nitpik codequal

btaillon (approved):

packages/headless/src/features/result-preview/result-preview-slice.ts
olamothe: With hindsight: Does not work due to immutability enforced by redux tool kit.

PR #2658, opened 2023-01-13
fix(atomic): improved atomic-icon error handling

btaillon:
https://coveord.atlassian.net/browse/KIT-2196

Also added a util so we can retrieve the exact error thrown.

### DIFF
--- a/packages/atomic/cypress/e2e/common-assertions.ts
+++ b/packages/atomic/cypress/e2e/common-assertions.ts
@@ -85,9 +85,12 @@ export function assertConsoleErrorMessage(msg: string) {
     cy.get(TestFixture.consoleAliases.error)
       .getCalls()
       .should((calls) => {
-        function isError(obj: any): obj is Error {
+        function isError(obj: unknown): obj is Error {
           return (
+            obj instanceof Object &&
+            'name' in obj &&
             typeof obj['name'] === 'string' &&
+            'message' in obj &&
             typeof obj['message'] === 'string'
           );
         }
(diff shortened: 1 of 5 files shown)


packages/atomic/cypress/support/commands.ts
```diff
@@ -27,6 +28,9 @@ declare global {
       distanceTo(
         getSubjectB: () => Chainable<JQuery<HTMLElement>>
       ): Chainable<{horizontal: number; vertical: number}>;
+      getCalls<TArgs extends any[], TReturnValue>(): Chainable<
```
ThibodeauJF, line 31: Should shut down those warnings ;)

ThibodeauJF (approved):

packages/atomic/cypress/e2e/common-assertions.ts
```diff
@@ -82,7 +82,20 @@ export function assertConsoleError(error = true) {
 
 export function assertConsoleErrorMessage(msg: string) {
   it('should log an error containing the appropriate message to the console', () => {
-    cy.get(TestFixture.consoleAliases.error).should('be.calledWithMatch', msg);
+    cy.get(TestFixture.consoleAliases.error)
+      .getCalls()
+      .should((calls) => {
+        function isError(obj: any): obj is Error {
```
louis-bompart, line 88:
```suggestion
        function isError(obj: unknown): obj is Error {
```
?

packages/atomic/cypress/support/commands.ts
louis-bompart: I think it's important to still bring a lot of scrutiny on the usage of `any`, 'cause otherwise, well you end up with JavaScript.

There could be case-by-case ignore tho. (although preferably another type such as unknown should be used when possible)

olamothe (approved):

louis-bompart (approved):

PR #2663, opened 2023-01-16
fix(quantic): fixed alignment of input when error state is on for numericFacet

SimonMilord:
[SFINT-4712](https://coveord.atlassian.net/browse/SFINT-4712)

Fixed alignment for Numeric Min/Max facet when error message pop-up

Before: 

image

After:

Screen Shot 2023-01-16 at 12 27 39 PM

[SFINT-4712]: https://coveord.atlassian.net/browse/SFINT-4712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathanlb:
There are a couple administrative things wrong with your PR. Totally normal for your first ones BTW.
1. The jira ID in the branch name needs to be formatted the same way it shows in Jira
  - SFINT-4712 ✅ 
  - sfint4712 🚫  (AFAIK any automated process searching for the Jira ID won't be able to recognize this unfortunately.)
2. The same thing is true for the link you added to the PR description:
  - [SFINT-4712](https://coveord.atlassian.net/browse/SFINT-4712) ✅
  - [SVCC-SFINT-4712](https://coveord.atlassian.net/browse/SFINT-4712) 🚫 (Not a valid jira ID)

FYI In this repo the process that's likely to use these is the deployment pipeline. PRs that don't have a findable Jira ID will be flagged and will block the deployment.
Adding the Jira ID in both the PR description AND the branch name is the best IMO because why not. But I think technically just one would work too FYI.

PR #2664, opened 2023-01-16
feat(atomic): add support to close modal on escape

olamothe:
https://coveord.atlassian.net/browse/KIT-2218

### DIFF
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -131,7 +131,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
 
   public disconnectedCallback(): void {
     this.closeOnEscape &&
-      document.removeEventListener('keyup', this.closeOnEscape);
+      document.body.removeEventListener('keyup', this.closeOnEscape);
   }
 
   private updateBreakpoints = once(() => updateBreakpoints(this.host));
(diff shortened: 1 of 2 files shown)


ThibodeauJF (approved):

SimonMilord (approved):
Lgtm!

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -109,13 +110,28 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
     return classes;
   }
 
+  private handleCloseOnEscape() {
+    this.closeOnEscape = (e: KeyboardEvent) => {
+      if (e.key.toLowerCase() === 'escape') {
+        this.close();
+      }
+    };
+    document.body.addEventListener('keyup', this.closeOnEscape);
```
btaillon, line 119: Can this be replaced with a `@Listen` decorator with the `document` target? https://stenciljs.com/docs/events#target

olamothe: 🤔 Yeah probably !

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -89,6 +89,13 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
     }
   }
 
+  @Listen('keyup', {target: 'body'})
+  handleCloseOnEscape(e: KeyboardEvent) {
+    if (e.key.toLowerCase() === 'escape') {
```
louis-bompart, line 94: 🤔 why lowerCase and not just "Escape" ?
https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

olamothe: Out of habit after all these years doing web dev, I always lower case compare anything that's coming from DOM API due to not knowing by heart where are all the landmines.

PR #2666, opened 2023-01-17
fix(quantic): fixed alignment of input when error state is on for numericFacet

SimonMilord:
[SFINT-4712](https://coveord.atlassian.net/browse/SFINT-4712)

This is a new PR since I changed the name of the branch following the heads up by nathan in the previous PR.

Fixed alignment for Numeric Min/Max facet when error message pop-up

Before: 

image

After:

Screen Shot 2023-01-16 at 12 27 39 PM

[SFINT-4712]: https://coveord.atlassian.net/browse/SFINT-4712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mmitiche (commented):
Nice job! I see that your changes fixed issue, however I think it also introduced another small issue:
NumericFacet
The inputs of the Numeric Facets are no longer taking the full width of the facet, do you think we can fix that?

SimonMilord:
> Nice job! I see that your changes fixed issue, however I think it also introduced another small issue: NumericFacet The inputs of the Numeric Facets are no longer taking the full width of the facet, do you think we can fix that?

HERE AFTER FIX:

Screen Shot 2023-01-19 at 11 47 54 AM

mmitiche (approved):
LGTM!

PR #2675, opened 2023-01-19
Pager doesn't show the correct max number of pages

sherakama:
The pager is limited to showing a number of pages divisible by the number of results per page and the 5000 item limit per query. Unfortunately, 5000 is the results per query limit and an index can have many more items. 

When you limit the max count to 5000 (number of items max returned) the pager fails to calculate the correct number of pages available. For example, my index has 310,000 items and if I am displaying 50 at a time, I should have a pager that goes to 6200 pages. However, I cannot get it to render more than a number divisible by 5000 and this creates a bad UI when in combination with the facet counts. 

Expected behaviour:

The pager will render the correct total max page and clicking on the pager page will create a fetch query starting at the correct record number 

For reference of where the 5000 limit came from.
https://docs.coveo.com/en/1461/build-a-search-ui/query-parameters#RestQueryParameters-numberOfResults

Full disclosure the code in the PR renders the correct amount of pager items but breaks when clicked on.

### DIFF
--- a/packages/headless/src/features/pagination/pagination-slice.ts
+++ b/packages/headless/src/features/pagination/pagination-slice.ts
@@ -31,7 +31,6 @@ import {
   nextPage,
 } from './pagination-actions';
 import {
-  maximumNumberOfResultsFromIndex,
   minimumPage,
 } from './pagination-constants';
 import {getPaginationInitialState, PaginationState} from './pagination-state';
@@ -159,9 +158,5 @@ export function calculateMaxPage(
   totalCountFiltered: number,
   numberOfResults: number
 ) {
-  const totalCount = Math.min(
-    totalCountFiltered,
-    maximumNumberOfResultsFromIndex
-  );
-  return Math.ceil(totalCount / numberOfResults);
+  return Math.ceil(totalCountFiltered / numberOfResults);
 }


sherakama:
Clicking on a pager item results in a 400 bad request error. A snippet with a log id for you.

```
{"statusCode":400,"message":"Bad Request","type":"ServerReturnedSoapFaultException","executionReport":[{"error":"Bad Request","children":[{"name":"RequestId","description":"The id correlating logs for this request","duration":0,"X-Request-ID":"31b977fb-fee7-4b47-9805-902e098a0d14"},{"name":"Organization ID","description":"Organization ID","duration":0,"coveo_organization":"stanfordgraduateschoolofbusinessnonproduction2sg4tcu4c"},
```

ThibodeauJF:
Hi @sherakama we appreciate you taking the initiative & opening a pull request, many thanks!

Unfortunately, we have to keep this limit hardcoded in Headless, we can't allow the pagination to return the following error as you pointed out:

> Full disclosure the code in the PR renders the correct amount of pager items but breaks when clicked on.

The index won't return results past 5000, it will spit out an error. It's a property of search engines, similar to Google where despite it saying `About 10,290,000,000 results (0.48 seconds)` it won't allow you to reach results over around 1000. For us, allowing the user to navigate over the allowed maximum number of results and getting the error is not proper UX. 

All this aside, since you use Headless you have full control over the UI of your interface, you could perhaps adapt the UI of your pager to display more pages, but we are reluctant to remove the limit in our base product, which affects all customers.

Hope you understand, thank you!

sherakama:
Oh, this was unclear in the documentation.

My assumption was that the results returned through the API would be limited to 5000 items and could start anywhere within the total amount of items available. I am used to APIs with very large data sets using these types of controls to paginate the API responses and expected to be able to set the `firstResult` parameter to something greater than 5000.

IMHO this is a gap in your service. We are not using the Coveo service to build a google like search engine, but rather, a directory where the end user could easily want to browse through several thousand items. 

I think the first step would be to document on the Query Parameter page that end users are not able to set the value of `firstResult` to something greater than 5000

https://docs.coveo.com/en/1461/build-a-search-ui/query-parameters#RestQueryParameters-firstResult

sherakama:
I re-read my last comment and realized I forgot to say thank you and sounded snarky. Thank you for the quick reply on the PR. I'll bring this up with Alan our account manager to see what he says others are doing and then to the UX team.

Thanks.

ThibodeauJF:
I'll make a note to the doc team about clarifying the first result parameter, good point!

PR #2676, opened 2023-01-19
SVCINT-1999 Feat: example of ipx with target selector

pathiery:
[SVCINT-1999](https://coveord.atlassian.net/browse/SVCINT-1999)

- Split the ipx components into 3: `atomic-ipx-modal`, `atomic-ipx-embedded` and `atomic-ipx-body`
- Create a new `scaleUpModalIPX` animation with a lower `translateY` value to avoid bugs
- Update a whole bunch of components to make sure `height` and `width` is inherited throughout the search interface into the `part=container` elements
- Update the `ipx.html` file to be able to toggle between the button and embedded ipx
- Probably other changes I forgot about

[SVCINT-1999]: https://coveord.atlassian.net/browse/SVCINT-1999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

https://user-images.githubusercontent.com/71142397/214706917-836e8a5c-ddad-40a0-8636-c6401d253fab.mov

### DIFF
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -55,7 +55,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   @Prop({mutable: true}) close: () => void = () => (this.isOpen = false);
   @Prop({reflect: true}) noFocusTrap = false;
   /**
-   * Wether to display the open and close animations over the entire page or the atomic-modal only.
+   * Whether to display the open and close animations over the entire page or the atomic-modal only.
    */
   @Prop({reflect: true}) animateOverEntirePage = true;
   @Prop({reflect: true}) isIPX = false;
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -57,7 +57,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   /**
    * Whether to display the open and close animations over the entire page or the atomic-modal only.
    */
-  @Prop({reflect: true}) boundaries: 'page' | 'element' = 'page';
+  @Prop({reflect: true}) boundary: 'page' | 'element' = 'page';
 
   @Event() animationEnded!: EventEmitter<never>;
 
@@ -188,7 +188,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
         <div
           part="backdrop"
           class={`
-            ${this.boundaries === 'page' ? 'fixed' : 'absolute'}
+            ${this.boundary === 'page' ? 'fixed' : 'absolute'}
             left-0 top-0 right-0 bottom-0 z-[9999]
           `}
           onClick={(e) => e.target === e.currentTarget && this.close()}
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -58,7 +58,6 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
    * Whether to display the open and close animations over the entire page or the atomic-modal only.
    */
   @Prop({reflect: true}) boundaries: 'page' | 'element' = 'page';
-  @Prop({reflect: true}) isIPX = false;
 
   @Event() animationEnded!: EventEmitter<never>;
 
@@ -140,13 +139,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
       <article
         part="container"
         class={`flex flex-col justify-between bg-background text-on-background
-          ${
-            this.isOpen
-              ? this.isIPX
-                ? 'animate-scaleUpModalIPX'
-                : 'animate-scaleUpModal'
-              : 'animate-slideDownModal'
-          }
+          ${this.isOpen ? 'animate-scaleUpModal' : 'animate-slideDownModal'}
           ${this.wasEverOpened ? '' : 'invisible'}`}
         onAnimationEnd={() => this.animationEnded.emit()}
         ref={(ref) => (this.animatableContainer = ref)}
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -57,7 +57,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   /**
    * Whether to display the open and close animations over the entire page or the atomic-modal only.
    */
-  @Prop({reflect: true}) animateOverEntirePage = true;
+  @Prop({reflect: true}) boundaries: 'page' | 'element' = 'page';
   @Prop({reflect: true}) isIPX = false;
 
   @Event() animationEnded!: EventEmitter<never>;
@@ -195,7 +195,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
         <div
           part="backdrop"
           class={`
-            ${this.animateOverEntirePage ? 'fixed' : 'absolute'}
+            ${this.boundaries === 'page' ? 'fixed' : 'absolute'}
             left-0 top-0 right-0 bottom-0 z-[9999]
           `}
           onClick={(e) => e.target === e.currentTarget && this.close()}
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -54,6 +54,11 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   @Prop({reflect: true, mutable: true}) isOpen = false;
   @Prop({mutable: true}) close: () => void = () => (this.isOpen = false);
   @Prop({reflect: true}) noFocusTrap = false;
+  /**
+   * Wether to display the open and close animations over the entire page or the atomic-modal only.
+   */
+  @Prop({reflect: true}) animateOverEntirePage = true;
+  @Prop({reflect: true}) isIPX = false;
 
   @Event() animationEnded!: EventEmitter<never>;
 
@@ -134,9 +139,15 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
     const Content = () => (
       <article
         part="container"
-        class={`flex flex-col justify-between bg-background text-on-background ${
-          this.isOpen ? 'animate-scaleUpModal' : 'animate-slideDownModal'
-        } ${this.wasEverOpened ? '' : 'invisible'}`}
+        class={`flex flex-col justify-between bg-background text-on-background
+          ${
+            this.isOpen
+              ? this.isIPX
+                ? 'animate-scaleUpModalIPX'
+                : 'animate-scaleUpModal'
+              : 'animate-slideDownModal'
+          }
+          ${this.wasEverOpened ? '' : 'invisible'}`}
         onAnimationEnd={() => this.animationEnded.emit()}
         ref={(ref) => (this.animatableContainer = ref)}
       >
@@ -183,7 +194,10 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
       <Host class={this.getClasses().join(' ')}>
         <div
           part="backdrop"
-          class="fixed left-0 top-0 right-0 bottom-0 z-[9999]"
+          class={`
+            ${this.animateOverEntirePage ? 'fixed' : 'absolute'}
+            left-0 top-0 right-0 bottom-0 z-[9999]
+          `}
           onClick={(e) => e.target === e.currentTarget && this.close()}
         >
           {this.noFocusTrap ? (
--- a/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
+++ b/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
@@ -20,7 +20,7 @@ interface RefineModalCommonProps {
   isOpen: boolean;
   openButton?: HTMLElement;
   noFocusTrap?: boolean;
-  animateOverEntirePage?: boolean;
+  boundaries?: 'page' | 'element';
   isIPX?: boolean;
 }
 
@@ -96,7 +96,7 @@ export const RefineModalCommon: FunctionalComponent<RefineModalCommonProps> = (
       }}
       exportparts={exportparts}
       noFocusTrap={props.noFocusTrap}
-      animateOverEntirePage={props.animateOverEntirePage}
+      boundaries={props.boundaries}
       isIPX={props.isIPX}
     >
       {renderHeader()}
--- a/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
+++ b/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
@@ -21,7 +21,6 @@ interface RefineModalCommonProps {
   openButton?: HTMLElement;
   noFocusTrap?: boolean;
   boundaries?: 'page' | 'element';
-  isIPX?: boolean;
 }
 
 export const RefineModalCommon: FunctionalComponent<RefineModalCommonProps> = (
@@ -97,7 +96,6 @@ export const RefineModalCommon: FunctionalComponent<RefineModalCommonProps> = (
       exportparts={exportparts}
       noFocusTrap={props.noFocusTrap}
       boundaries={props.boundaries}
-      isIPX={props.isIPX}
     >
       {renderHeader()}
       {...children}
--- a/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
+++ b/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
@@ -20,6 +20,8 @@ interface RefineModalCommonProps {
   isOpen: boolean;
   openButton?: HTMLElement;
   noFocusTrap?: boolean;
+  animateOverEntirePage?: boolean;
+  isIPX?: boolean;
 }
 
 export const RefineModalCommon: FunctionalComponent<RefineModalCommonProps> = (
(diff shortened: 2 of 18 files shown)


packages/atomic/src/pages/examples/ipx.html
```diff
@@ -9,9 +9,17 @@
...
-      (async () => {
-        await customElements.whenDefined('atomic-search-interface');
-        const searchInterface = document.querySelector('atomic-search-interface');
+      const ipxSelectorToButtonMode = (searchInterface, hostDiv) => {
+        searchInterface.style.display = 'inline';
+        hostDiv.style.display = 'none';
+      };
+
+      const ipxButtonToSelectorMode = (searchInterface, hostDiv) => {
+        searchInterface.style.display = 'none';
```
dguerinCoveo, line 18: I'm wondering if that should be in the component itself 🤔 If we scale the IPX, we might start having lots of these small changes that when added up won't make a lot of sense and will be difficult for user to understand and use

pathiery: This exists solely to toggle between the 2 ipx, but yes I agree that if the examples get more complex, we should split them up into separate pages !

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -20,6 +28,35 @@
...
+        const ipxInSelector = searchInterface.cloneNode(true);
+
+        await initializeSearchInterface(searchInterface);
+
+        const hostDiv = document.querySelector('div#host-div');
+        hostDiv.style.display = 'none';
+        hostDiv.appendChild(ipxInSelector);
+        ipxInSelector.querySelector('atomic-ipx-modal').setAttribute('is-embedded', 'true');
+        ipxInSelector.removeChild(ipxInSelector.querySelector('atomic-ipx-button'));
+        await initializeSearchInterface(ipxInSelector);
```
btaillon, line 46: I think this (and the `cloneNode` call above) should be extracted into a separate function, like `createEmbeddedInterfaceClone`.

packages/atomic/src/pages/examples/ipx.html
dguerinCoveo: But if you create an IPX that uses a target selector, you have to always set the style of the `atomic-search-interface` display to `inline` no?

packages/atomic/src/components/ipx/atomic-ipx-modal/atomic-ipx-modal.tsx
```diff
@@ -80,6 +83,10 @@ export class AtomicIPXModal implements InitializableComponent<AnyBindings> {
     this.isOpen && e.preventDefault();
   }
 
+  public componentWillLoad(): void {
+    this.isOpen = this.isOpen || this.isEmbedded;
```
btaillon, line 87: Is the embedded version still visually a modal? If not, I recommend removing the `atomic-focus-trap` from the embedded version. This is important because:
* The atomic focus trap prevents keyboard navigation outside of its children.
* The atomic focus trap hides anything outside of it from assistive technology (like screen readers).
* The atomic focus trap has `role="dialog"`, which conveys to assistive technology that there's a dialog currently opened.

dguerinCoveo (approved):

packages/atomic/src/components/ipx/atomic-ipx-modal/atomic-ipx-modal.tsx
```diff
@@ -33,6 +33,8 @@ export class AtomicIPXModal implements InitializableComponent<AnyBindings> {
   @State() public error!: Error;
 
   @Prop({mutable: true}) source?: HTMLElement;
+
+  @Prop() isEmbedded = false;
```
btaillon, line 37: I'm not sure how I feel about having the embedded option in a component titled modal, unless it looks like a modal.

Could you give me some examples of cases where this would be used?

mikegron: Maybe rename the modal for "window" 🤔

packages/atomic/src/components/ipx/atomic-ipx-modal/atomic-ipx-modal.tsx
```diff
@@ -46,20 +48,21 @@ export class AtomicIPXModal implements InitializableComponent<AnyBindings> {
 
   @Watch('isOpen')
   async watchToggleOpen(isOpen: boolean) {
+    if (this.isEmbedded) {
+      return;
```
mikegron, line 52: Tu serais mieux de mettre ca dans `componentDidLoad` a la place pour pas caller la méthode pentoute imo

packages/atomic/src/components/ipx/atomic-ipx-modal/atomic-ipx-modal.tsx
```diff
@@ -95,7 +102,11 @@ export class AtomicIPXModal implements InitializableComponent<AnyBindings> {
       <Host class={this.getClasses().join(' ')}>
         <div
           part="backdrop"
-          class="fixed left-0 top-0 right-0 bottom-0 z-[9999]"
+          class={
+            this.isEmbedded
+              ? ''
+              : 'fixed left-0 top-0 right-0 bottom-0 z-[9999]'
```
mikegron, line 108: lol classique z index, je metterais 99999 pour être safe :trollface: (not)

packages/atomic/src/components/ipx/atomic-ipx-modal/atomic-ipx-modal.tsx
```diff
@@ -46,20 +48,21 @@ export class AtomicIPXModal implements InitializableComponent<AnyBindings> {
...
-    } else {
-      document.body.classList.remove(modalOpenedClass);
-      if (watchToggleOpenExecution !== this.currentWatchToggleOpenExecution) {
-        return;
+      if (watchToggleOpenExecution === this.currentWatchToggleOpenExecution) {
+        this.focusTrap!.active = true;
       }
+      return;
+    }
+    document.body.classList.remove(modalOpenedClass);
```
mikegron, line 64: Juste pour mon info, `document.body` ici réfère bien au body du Shadow Root right?

pathiery: nenon, c'est le `body` tout en haut du html. faut espérer que ça brise rien dans le host website

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -53,6 +53,11 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   @Prop({mutable: true}) container?: HTMLElement;
   @Prop({reflect: true, mutable: true}) isOpen = false;
   @Prop({mutable: true}) close: () => void = () => (this.isOpen = false);
+  /**
+   * Wether to display the open and close animations over the entire page or the atomic-modal only.
```
dguerinCoveo, line 57:
```suggestion
   * Whether to display the open and close animations over the entire page or the atomic-modal only.
```

dguerinCoveo: > The short answer is that wether is a male goat or sheep, whether means “if,” and weather means the state of the atmosphere.

😆

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -53,6 +53,11 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
...
+   */
+  @Prop({reflect: true}) animateOverEntirePage = true;
+  @Prop({reflect: true}) isIPX = false;
```
dguerinCoveo, line 60: Could atomic-ipx-modal be its own component instead? I know the DX team is not super kean with forks in base components 🤔

packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
```diff
@@ -19,6 +19,8 @@ interface RefineModalCommonProps {
   };
   isOpen: boolean;
   openButton?: HTMLElement;
+  animateOverEntirePage?: boolean;
+  isIPX?: boolean;
```
dguerinCoveo, line 23: same question

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -340,8 +408,13 @@
           </a>
         </div>
       </atomic-ipx-modal>
-      <atomic-ipx-button is-modal-open="true"></atomic-ipx-button>
+      <atomic-ipx-button is-modal-open="true" label="Workplace Search"></atomic-ipx-button>
```
dguerinCoveo, line 411: Could you just use `Help` for now since Workplace is kind of dead

dguerinCoveo (commented):
Looks really nice GG! My main comment is about modifying the base components, but I think its a DX decision

packages/atomic/src/components/ipx/atomic-ipx-embedded/atomic-ipx-embedded.tsx
```diff
@@ -0,0 +1,62 @@
...
+export class AtomicIPXEmbedded implements InitializableComponent<AnyBindings> {
+  @InitializeBindings() public bindings!: AnyBindings;
+  @Element() public host!: HTMLElement;
+
+  @State() public error!: Error;
+
+  /**
+   * The container to hide from the tabindex and accessibility DOM when the modal is closed.
+   */
+  @Prop({mutable: true}) container?: HTMLElement;
```
btaillon, line 36: You should sync with https://github.com/coveo/ui-kit/pull/2697

packages/atomic/src/components/ipx/atomic-ipx-refine-modal/atomic-ipx-refine-modal.tsx
```diff
@@ -148,10 +156,9 @@ export class AtomicIPXRefineModal implements InitializableComponent {
         {this.interfaceDimensions && (
           <style>
             {`atomic-modal::part(backdrop) {
-            top: ${this.interfaceDimensions.top}px;
-            left: ${this.interfaceDimensions.left}px;
             width: ${this.interfaceDimensions.width}px;
             height: ${this.interfaceDimensions.height}px;
+            ${this.isOpen ? '' : 'display: none;'}
```
btaillon, line 161: This will cause the animation to instantly stop when closing the modal

packages/atomic/src/components/ipx/atomic-ipx-refine-modal/atomic-ipx-refine-modal.tsx
```diff
@@ -78,32 +78,40 @@ export class AtomicIPXRefineModal implements InitializableComponent {
...
+      !!atomicModalDimensions &&
+      this.dimensionChanged(atomicModalDimensions)
+    ) {
+      this.updateDimensions(atomicModalDimensions);
     }
     window.requestAnimationFrame(() => this.onAnimationFrame());
   }
 
-  private dimensionChanged() {
+  private dimensionChanged(atomicModalDimensions: DOMRect) {
```
btaillon, line 105: I didn't remember that we had all that dimension logic in the refine modal... copying it over to IPX may be the reason you had issues with the animation and `overflow: hidden` not working. I wonder what would happen if we removed all of the dimension code from IPX (including the `top`, `left`, `width` and `height` properties below).

packages/atomic/src/components/ipx/atomic-ipx-embedded/atomic-ipx-embedded.tsx
pathiery: Looks like it's a different issue, there's actually more to making `atomic-ipx-embedded` than removing the focus trap, especially looking at the css of both variants.
We're gonna need their feature for the refine modal though, since we don't want there to be a focus trap, as the page is embedded, I guess we'll wait for that to be merged and sync !

packages/atomic/src/components/ipx/atomic-ipx-refine-modal/atomic-ipx-refine-modal.tsx
pathiery: @btaillon not sure what you mean, where do you want me to try to remove those values from ?

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
pathiery: @btaillon

packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
pathiery: @btaillon

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
btaillon: I thought it already was? I guess you mean for the refine modal.

I don't think there's a significant enough difference between the two to need a new refine modal component. The biggest difference here is some styling/animation and the space occupied by the component. I am actually tempted to investigate this myself and see if I'm luckier in finding a solution that removes the need for the `isIPX` parameter. Since this is internal, it's not a problem for now.

packages/atomic/src/components/ipx/atomic-ipx-embedded/atomic-ipx-embedded.tsx
btaillon: > We're gonna need their feature for the refine modal though, since we don't want there to be a focus trap, as the page is embedded, I guess we'll wait for that to be merged and sync !

Concerning the option to remove the focus trap, this is actually very similar to the context for the refine case. They don't want elements behind the modal to be focusable, but they also don't want to prevent users from tabbing out of the search page itself.

Either way, their fix should work for IPX too.

packages/atomic/src/components/ipx/atomic-ipx-refine-modal/atomic-ipx-refine-modal.tsx
btaillon: Basically, there's a bunch methods related to dimensions on the `atomic-ipx-refine-modal` and `atomic-insight-refine-modal` which aren't needed for `atomic-refine-modal`. I believe those were needed to ensure that the refine modal doesn't spill out of the boundaries in the refine page, I completely forgot about it until now. With the new option to make `atomic-modal` not expand beyond its own size, these are likely no longer necessary, and may even have caused the issue you saw before.

> where do you want me to try to remove those values from ?

https://github.com/coveo/ui-kit/blob/51b378972e506ecd67b791f2b3e8dd6ca6f974eb/packages/atomic/src/components/ipx/atomic-ipx-refine-modal/atomic-ipx-refine-modal.tsx#L159-L160

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
mikegron: Ish ouin un bool comme ca c'est semi 😐

packages/atomic/src/components/ipx/atomic-ipx-refine-modal/atomic-ipx-refine-modal.tsx
pathiery: huh, looks like removing it from `atomic-ipx-refine-modal` maintains all functionality !

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
dguerinCoveo: Hmm yeah I'm a bit confused. Why do we need to change some props in `atomic-modal` if we don't directly use it in an IPX? 🤔

pathiery: we need to pass it to `refine-modal-common` so that it gets passed to `atomic-modal` so that we can override the animation settings :/

olamothe: Yeah that's the exact kind of thing that should not happening the project, this is a slippery slope. This needs to stay generic.

Just a very slight change that will make all the difference would be to have a prop on the animation name itself.

```
@Prop() animationOnOpen = 'animateScaleUpModal'
@Prop() animationOnClose = 'animateSlideDownModal'
```

And you can then leverage tailwind to pass arbitrary animation.
eg: https://tailwindcss.com/docs/animation#arbitrary-values

packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
olamothe: Same comment as above for me ;)

olamothe (changes requested):

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -54,6 +54,10 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   @Prop({reflect: true, mutable: true}) isOpen = false;
   @Prop({mutable: true}) close: () => void = () => (this.isOpen = false);
   @Prop({reflect: true}) noFocusTrap = false;
+  /**
+   * Whether to display the open and close animations over the entire page or the atomic-modal only.
+   */
+  @Prop({reflect: true}) boundaries: 'page' | 'element' = 'page';
```
olamothe, line 60: Wonder if we should not use singular form here ? `page boundary` or `element boundary` VS `page boundaries` or `element boundaries` ... 🤔

olamothe (approved):
Great job ! 👍

btaillon (approved):

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
btaillon: Looks like the correct word is "boundary", TIL

PR #2679, opened 2023-01-20
docs(atomic-react): various atomic-react samples improvements

btaillon:
https://coveord.atlassian.net/browse/KIT-2195

I made the following changes:
* Renamed package `atomic-react` to `@coveo/atomic-react-samples`
* Moved `search-page.tsx` from the `atomic-next` samples into `index.tsx`.
  * `index.tsx` was re-exporting the contents of `search-page.tsx` using `dynamic` to disable server-side rendering. This was done because Atomic doesn't support server-side rendering. However, in NextJS, every file in the `pages` directory represents a page, so this solution caused two pages to exist: `/` and `/search-page`, where the former worked as expected and the latter would show an error.
* Wrapped every React function component engine or controller initialization in `useMemo`/`useEffect` with dependencies
  * If this sounds unfamiliar, see [Tip: Optimizing Performance by Skipping Effects](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects) and [useMemo](https://beta.reactjs.org/reference/react/useMemo).
  * Without this, any change in state or props (including `children`, I tested it) of the component would cause a whole new engine to be initialized. This was fine in most cases for us since:
    * We often didn't have any props or state, or simply didn't change them. 
    * Changing the `engine` of `AtomicSearchInterface` does nothing, the `AtomicSearchInterface` component will continue to use the old `engine`.
* Made `server.ts` unsubscribe from its listener when it's no longer needed.
* Improved the logic of `SamlPage` to avoid re-creating the SAML client
  * In theory, the old logic should be re-creating a new SAML client on every re-render, which should be at-least twice since the SAML page updates a state. The new logic wraps everything in `useMemo`/`useEffect` and uses the proper dependencies to only re-run each initialization when their parameters change.


packages/samples/headless-react/src/pages/SamlPage.tsx
```diff
@@ -1,38 +1,41 @@
...
+  const samlClient = useMemo(
+    () => buildSamlClient(samlClientOptions),
+    [samlClientOptions]
+  );
 
   useEffect(() => {
-    getToken();
-  }, []);
+    samlClient.authenticate().then(setInitialAccessToken);
+  }, [samlClient]);
```
btaillon, line 18: In isolation, we would want to call `authenticate` whenever a new SAML client is built, since we wouldn't know why it was rebuilt. This is the paradigm I tried to apply for every `useMemo`/`useEffect`.

ThibodeauJF (approved):

olamothe (approved):
Really nice ! Always important to keep the samples up to date/correctly built.

As you said, quite a bit of people will use those to start their project, and this will help introduce any weird problems in their deployments.

Great idea to fix all of those !

packages/samples/headless-react/src/pages/SamlPage.tsx
```diff
@@ -1,41 +1,49 @@
...
-}: PropsWithChildren<SamlClientOptions>) {
-  const [initialAccessToken, setInitialAccessToken] = useState('');
-  const samlClient = useMemo(
-    () => buildSamlClient(samlClientOptions),
-    [samlClientOptions]
-  );
+const samlClientOptions: SamlClientOptions = {
+  organizationId: '',
+  provider: '',
+};
```
btaillon, line 16: Moved the options here, since the SAML client's `authenticate` method is deterministic and we don't want to suggest the possibility of changing the options of the SAML client.

ThibodeauJF (approved):

packages/samples/headless-react/src/pages/SamlPage.tsx
btaillon: I could try to add a util for authentication that isn't deterministic, or I could update `@coveo/auth` (not sure what I'd do yet)

PR #2680, opened 2023-01-20
fix(quantic): issue fixed with the QuanticResultText

mmitiche:
[SFINT-4756](https://coveord.atlassian.net/browse/SFINT-4756)

### Issue:

Missing validation for empty-string in the `field` property of QuanticResultText:
using the component like so wouldn’t trigger any error:
```
<c-quantic-result-text result={result} label="SourceMising" field=""></c-quantic-result-text> 
```

Currently  it displays the Label without the field

### Solution:
Added validation for empty string, now the components displays an error when the `field` property is empty.

[SFINT-4756]: https://coveord.atlassian.net/browse/SFINT-4756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
Looks good!

PR #2683, opened 2023-01-20
docs(headless): moved headless-react controller initialization to inside of component

btaillon:
https://coveord.atlassian.net/browse/KIT-2195

This PR won't be merged, it's purely for exploration.

# Context
For most Headless controllers, the `@coveo/headless-react-samples` package has two corresponding components, which are both published to the reference doc in https://docs.coveo.com:
* A functional component, which takes a controller as a prop.
* A class component, which takes the controller's options as props and initializes the controller itself.

# Problem
The class components and functional components take different props, which kind of implies that we recommend both approaches to implements. I think we should have one recommended approach.

Also, the functional components are currently subscribing to their controller once. If the controller given to them as a parameter changes, they won't unsubscribe from the old controller and re-subscribe to the new one. There's a decent bit of logic involved in initializing and subscribing to changes from a controller in a React functional component.

# Solution
There are two things to decide:
* The props taken by the component
* How the controller is initialized inside (or outside) the component

### Solution 1 (the one implemented in this PR)
Controller options are nested in the component's props.

For example
```tsx
<FacetExample controllerOptions={{ facetId: 'abc', field: 'def' }} />
```

Pros:
* Inside the component, `controllerOptions` can be passed directly to the `deps` array of `useEffect`/`useMemo` without any destructuring.

### Solution 2
Controller options are used directly as the component's props.

For example:
```tsx
<FacetExample facetId="abc" field="def" />
```

Pros:
* Much prettier (IMO)
* No need to define an interface inside the component (unless it takes other props as well)

Cons:
* Inside the component, it may be tempting to add all `props` to the `deps` array of a `useEffect`/`useMemo`. This would cause the whole controller to be rebuilt whenever any options change, even if implementers add some purely cosmetic props and those are the ones changing. The alternative for the user is to destructure all controller options and pass that as the `deps` array, which is inconvenient.

### Solution 3
Components each take only the necessary options as props, not all of the controller's options.

For example:
```tsx
<FacetExample facetId="abc" field="def" />
```

Pros:
* It's intuitive which props to add to the `deps` of `useEffect`/`useMemo`, ESLint may even recommend them automatically.
* From the outside of the component, this is as pretty as solution B.

Cons:
* A lot of code repetition since implementers need to write a whole interface for the props they receive.
* If we deprecate an option in Headless, code editors will only underline the prop in the controller initialization, which is less likely to be seen.
* Implementers may choose weird names for props, which may be harder to read when they open support cases.

### Solution 4
We take the controller as a prop. This is the same as on `master` currently (but with some bug fixes).

For example:
```tsx
<FacetExample controller={facetController} />
```

Pros:
* No need for a `useMemo` to initialize the controller, and therefore no need for a deps array that takes only the correct options. The component will be re-rendered whenever the controller changes, and only the `subscribe` call needs to be wrapped in a `useEffect`.

Cons:
* Implementers need to initialize controllers in the context of where they are rendered, which pollutes their code very badly if they render a lot of different components.

### Solution A (the one implemented in this PR)
We wrap the controller initialization in a `useMemo`, we call `useState` to be able to re-render the component, and we wrap the `subscribe` call in a `useEffect` to update the state.

For example:
```ts
const controller = useMemo(
  () => buildFacet(engine!, {options: controllerOptions}),
  [engine, controllerOptions]
);
const [state, setState] = useState(controller.state);
useEffect(
  () => controller.subscribe(() => setState(controller.state)),
  [controller]
);
```

Cons:
* Quite a few lines just to initialize a controller, and a bit of thought needs to be put into what we pass to the `deps` arrays, which leaves lots of room for error.
* Some `useEffect` and `useMemo` hook may be called with the old state (and never the new one) when the controller is replaced. This happened when I updated `DependentFacetPage`, and I wasted over an hour debugging it.

### Solution B
We create a helper which takes the build function as a first parameter, the engine as a second parameter, and the options as a third parameter.

This helper could be either hosted in the samples, or actually exported by Headless (or exported from a new package).

For example:
```ts
import { buildFacet } from '@coveo/headless';
import { useController } from '../utils/coveo-headless';
...
const { controller, state } = useController(buildFacet, engine, { field, facetId });
```

Pros:
*  We can ensure there's one universal way to initialize Headless controllers in a functional component, which we can patch if we find errors.
* More intuitive for the user. We can manage the deps array for them and they don't need to understand React as much.

Cons:
* Useless for class components (although React recommends that new components be functional components).
* Implementers either need to copy it from our doc or we need to somehow export it from one of our packages, which is more work for us.

### Solution C (I really don't like this one)
Same as solution B, but we take the whole controller as a parameter.

For example:
```ts
import { buildFacet } from '@coveo/headless';
import { useHeadlessSubscribe } from '../utils/coveo-headless';
...
const { controller } = useMemo(
  buildFacet(engine, { field, facetId }),
  [engine, field, facetId]
)
const { state } = useHeadlessSubscribe(controller);
```

Pros:
* Smaller utility

Cons:
* Useless for class components (although React recommends that new components be functional components).
* Implementers either need to copy it from our doc or we need to somehow export it from one of our packages, which is more work for us.
* The controller needs to be built in a `useMemo` and take a dependencies array.

# Discussion

Which solution do you guys prefer?

My preferred solution right now would be 3B.

Also, although React plans to keep maintaining class components, they [are writing new components as functional components](https://reactjs.org/docs/hooks-intro.html#gradual-adoption-strategy), I don't see much of a reason to write any new component as a class component. Do you think we should remove class components? I don't see much value in them.


packages/samples/headless-react/src/components/facet-manager/facet-manager.class.tsx
```diff
@@ -8,7 +8,13 @@ import {
 import {Children, Component, ContextType, ReactElement} from 'react';
 import {AppContext} from '../../context/engine';
 
-type FacetManagerChild = ReactElement<{facetId: string}>;
+export interface FacetManagerClassChildProps {
+  controllerOptions: {facetId: string};
+  /** @deprecated */
+  facetId?: string;
```
btaillon, line 14: Would be removed when all components are converted.

packages/samples/headless-react/src/pages/DependentFacetPage.tsx
```diff
@@ -1,15 +1,14 @@
 import {
```
btaillon, line 1: I pretty much completely changed this doc page since we can no longer subscribe to the controller from outside the component.

IMO, it's easier to read as documentation now, since everything is in one single file.

packages/samples/headless-react/src/pages/DependentFacetPage.tsx
```diff
@@ -36,116 +35,241 @@ const Examples = {
...
+}
 
-    this.engine = buildSearchEngine({
-      configuration: getSampleSearchEngineConfiguration(),
-    });
+const FacetWithDependencies: FunctionComponent<FacetWithDependenciesProps> = ({
+  engine,
+  controllerOptions,
+  conditions,
+}) => {
```
btaillon, line 51: The old doc approach was subscribing to the controller from outside the component and hiding it. The new approach is more composed, it initializes the controller along with its dependencies at the same place where it gets rendered. We don't need multiple subscribers. I prefer this, personally.

packages/samples/headless-react/src/pages/DependentFacetPage.tsx
```diff
@@ -36,116 +35,241 @@ const Examples = {
...
-    });
-    this.fileTypeFacet = buildFacet(this.engine, {
-      options: {field: 'filetype'},
+  // Initialize conditions.
+  useEffect(() => {
+    if (!conditions?.length) {
+      return;
+    }
+    const conditionsManager = buildFacetConditionsManager(engine, {
+      facetId: controllerOptions.facetId,
```
btaillon, line 69: When switching tabs, the `state` in this `useEffect` call always contains an old facet, therefore `state.facetId` cannot be used, which isn't very intuitive.

A solution for this may be to replace `useState`/`useMemo` of solution A with something else so that the `state` always contains the correct value, I haven't had time to investigate this yet. Solution B or C would inherently solve this problem.

packages/samples/headless-react/src/components/facet-manager/facet-manager.class.tsx
olamothe: Yeah, for the sample, I'm not sure we should bother with any deprecation notice...

I think it's fair to just update them/change them completely in a minor version, unless there is something I'm not seeing.

olamothe (dismissed):
> Solution 1 (the one implemented in this PR)

I like this one personally. It's straightforward enough. I don't we need to over engineer this.


> How the controller is initialized inside (or outside) the component
> [...]
> Solution B
We create a helper which takes the build function as a first parameter, the engine as a second parameter, and the options as a third parameter.

I think it would make sense to have that helper indeed, since React is obviously:

1- Highly popular and used by many clients.
2- Tricky to "get everything right".

So I think it makes a lot of sense that we create something for that. 
It's more work for us indeed, but I don't think it should be that much ?

In general, great PR 👍 !!

packages/samples/headless-react/src/components/facet-manager/facet-manager.class.tsx
btaillon: yeah, what I meant is that the final PR which would be merged wouldn't have that. The `@deprecated` here is basically a TODO for myself.

ThibodeauJF:
👍  This needed improvement, I like 1 or 3 & B

ThibodeauJF (approved):

PR #2686, opened 2023-01-23
feat(atomic): add recommendations in ipx

jbvouma:
[SVCINT-2004](https://coveord.atlassian.net/browse/SVCINT-2004)

We added recommendations when an IPX is opened prior a first search. We don't have a UX design yet.

I also made some changes to have the modal opened by default in the ipx.html example.

Coveo-Atomic-IPX-example-Google (4)

### DIFF
--- a/packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
+++ b/packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
@@ -32,7 +32,7 @@ export class AtomicIPXButton {
   /**
    * Whether the IPX layout is open.
    */
-  private isOpen = false;
+  private isOpen = this.host.classList.contains('btn-open');
 
   private async onClick() {
     this.isOpen ? this.close() : this.open();
--- a/packages/atomic/src/components/recommendations/atomic-recs-result/atomic-recs-result.tsx
+++ b/packages/atomic/src/components/recommendations/atomic-recs-result/atomic-recs-result.tsx
@@ -161,8 +161,9 @@ export class AtomicRecsResult {
     }
     return (
       // deepcode ignore ReactSetInnerHtml: This is not React code
-      <Host class={resultComponentClass}>
+      <Host exportparts="result-list-grid-content" class={resultComponentClass}>
         <div
+          part="result-list-grid-content"
           class={`result-root ${this.layout
             .getClasses()
             .concat(this.classes)
--- a/packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
+++ b/packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
@@ -469,6 +469,11 @@ export class AtomicSearchInterface
         this.searchStatus.state.hasError
       );
 
+      this.host.classList.toggle(
+        'atomic-search-interface-search-executed',
+        this.searchStatus.state.firstSearchExecuted
+      );
+
       if (
         this.searchStatus.state.firstSearchExecuted &&
         this.store.hasLoadingFlag(FirstSearchExecutedFlag)
--- a/packages/atomic/src/pages/examples/ipx.html
+++ b/packages/atomic/src/pages/examples/ipx.html
@@ -16,7 +16,16 @@
           accessToken: 'xx564559b1-0045-48e1-953c-3addd1ee4457',
           organizationId: 'searchuisamples',
         });
-        searchInterface.executeFirstSearch();
+        //searchInterface.executeFirstSearch();
+
+        await customElements.whenDefined('atomic-recs-interface');
+        const recsInterface = document.querySelector('atomic-recs-interface');
+        await recsInterface.initialize({
+          accessToken: 'xx564559b1-0045-48e1-953c-3addd1ee4457',
+          organizationId: 'searchuisamples',
+        });
+        recsInterface.getRecommendations();
+
         searchInterface.i18n.addResourceBundle('en', 'caption-filetype', {
           '.html': 'html',
         });
@@ -63,12 +72,57 @@
         width: 5rem;
         height: 1.25rem;
       }
+
+      atomic-recs-list::part(label) {
+        font-size: var(--atomic-text-xl);
+        padding-top: 0.75rem;
+        padding-bottom: 0.25rem;
+      }
+
+      atomic-recs-list::part(outline) {
+        padding: 0.5rem;
+        border: none;
+      }
+
+      atomic-recs-list::part(result-list-grid-clickable) {
+        inset: unset;
+      }
+
+      atomic-recs-list::part(result-list-grid-content) {
+        width: 27rem;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+
+      atomic-search-interface:not(.atomic-search-interface-search-executed)
+        > atomic-ipx-modal
+        > div[slot='body']
+        > atomic-layout-section {
+        display: none;
+      }
+
+      atomic-search-interface.atomic-search-interface-search-executed
+        > atomic-ipx-modal
+        > div[slot='body']
+        > atomic-recs-interface
+        > atomic-recs-list {
+        display: none;
+      }
+
+      body:not(.atomic-ipx-modal-opened)
+        > atomic-search-interface
+        > atomic-ipx-modal
+        > div[slot='body']
+        > atomic-recs-interface {
+        display: none;
+      }
     </style>
   </head>
 
   <body>
     <atomic-search-interface fields-to-include='["snrating","sncost"]'>
-      <atomic-ipx-modal>
+      <atomic-ipx-modal is-open="true">
         <div slot="header">
           <atomic-layout-section class="search-section" section="search">
             <atomic-search-box class="search-box"></atomic-search-box>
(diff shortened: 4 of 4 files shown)


packages/atomic/src/components/recommendations/atomic-recs-result/atomic-recs-result.tsx
```diff
@@ -161,8 +161,9 @@ export class AtomicRecsResult {
     }
     return (
       // deepcode ignore ReactSetInnerHtml: This is not React code
-      <Host class={resultComponentClass}>
+      <Host exportparts="result-list-grid-content" class={resultComponentClass}>
```
jbvouma, line 164: I needed this exports for more styling of the content inside the grid.

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -469,6 +469,11 @@ export class AtomicSearchInterface
         this.searchStatus.state.hasError
       );
 
+      this.host.classList.toggle(
+        'atomic-search-interface-search-executed',
+        this.searchStatus.state.firstSearchExecuted
+      );
```
jbvouma, line 475: Thisone is to help displaying/hiding the recommendations when needed.

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -63,12 +72,57 @@
...
+        > atomic-recs-interface {
+        display: none;
+      }
     </style>
   </head>
 
   <body>
     <atomic-search-interface fields-to-include='["snrating","sncost"]'>
-      <atomic-ipx-modal>
+      <atomic-ipx-modal is-open="true">
```
jbvouma, line 125: Let the modal opened by default

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -225,19 +279,24 @@
...
-          />
+          <a href="https://coveo.com/" target="_blank">
```
jbvouma, line 291: Add a redirect link on the Coveo icon @lvu285 .

pathiery (approved):

packages/atomic/src/components/recommendations/atomic-recs-result/atomic-recs-result.tsx
```diff
@@ -161,8 +161,9 @@ export class AtomicRecsResult {
     }
     return (
       // deepcode ignore ReactSetInnerHtml: This is not React code
```
louis-bompart, line 163: I reckon you don't need deepcode ignore comment, just address the notice in GH
(@btaillon amarite?)

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -225,19 +279,24 @@
           <atomic-layout-section section="pagination">
             <atomic-load-more-results class="pagination-more-result"></atomic-load-more-results>
           </atomic-layout-section>
+          <atomic-recs-interface fields-to-include='["cat_rating_count"]'>
```
y-lakhdar, line 282: Is there a reason to why `cat_rating_count` should be included?

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -63,12 +72,57 @@
...
+      }
+
+      atomic-search-interface:not(.atomic-search-interface-search-executed)
+        > atomic-ipx-modal
+        > div[slot='body']
+        > atomic-layout-section {
+        display: none;
+      }
+
+      atomic-search-interface.atomic-search-interface-search-executed
```
y-lakhdar, line 105: Should the `div[slot='body'] > atomic-layout-section` be unhidden at this stage?

packages/atomic/src/components/recommendations/atomic-recs-result/atomic-recs-result.tsx
btaillon: I didn't know you could address it in GH. Either it wasn't there last time I checked, or I never thought about using this feature. I have a small preference for keeping it in the code though, since that's searchable.

btaillon: I don't know why this particular ignore is on that line though, that looks wrong.

packages/atomic/src/components/recommendations/atomic-recs-result/atomic-recs-result.tsx
btaillon: I didn't know you could add `exportparts` on the root of a component, TIL.

If you want to re-export parts though, I would prefer if it was explicitly from each parent component, so that we know we need to document them with `@part`.

packages/atomic/src/pages/examples/ipx.html
jbvouma: Not really! I just let it here as an example for our setup in the Search Interface Service. I'll remove the field and let the array empty.

packages/atomic/src/pages/examples/ipx.html
jbvouma: Yes we want to show the search results when the class `atomic-search-interface-search-executed` exists
and hide the list of recommendations.

packages/atomic/src/components/recommendations/atomic-recs-result/atomic-recs-result.tsx
```diff
@@ -161,8 +161,9 @@ export class AtomicRecsResult {
...
-      <Host class={resultComponentClass}>
+      <Host exportparts="result-list-grid-content" class={resultComponentClass}>
         <div
+          part="result-list-grid-content"
```
btaillon, line 166: The result root is technically already style-able by adding a `style` tag to its template and using class selectors. It's even possible to target specific display modes. See https://github.com/coveo/ui-kit/blob/d1f650d2097b8873c69c15910aff2547ff5d4aee/packages/atomic/src/pages/index.html#L256-L259

I don't think we should add this unless we decide that we also need it for `atomic-result`/`atomic-result-list`, which I don't think we do.

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
btaillon: What's the particular problem you're trying to solve with this class? Perhaps it should be something we fix within Atomic instead.

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -63,12 +72,57 @@
...
+      atomic-recs-list::part(result-list-grid-clickable) {
+        inset: unset;
+      }
+
+      atomic-recs-list::part(result-list-grid-content) {
+        width: 27rem;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
```
btaillon, line 96: I prefer to keep result styling scoped to specific templates. See [displaying results (styling)](https://docs.coveo.com/en/atomic/latest/usage/displaying-results/#styling).

packages/atomic/src/pages/examples/ipx.html
btaillon: At-least while the initial search is being performed, we shouldn't hide the results, since we have placeholders.

Prior to the first search being executed at all, I think it should be the responsibility of the javascript that initializes the interface to hide it if they don't want it yet.

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -225,19 +279,24 @@
           <atomic-layout-section section="pagination">
             <atomic-load-more-results class="pagination-more-result"></atomic-load-more-results>
           </atomic-layout-section>
+          <atomic-recs-interface fields-to-include='["cat_rating_count"]'>
+            <atomic-recs-list label="Recommended articles" number-of-recommendations="10"> </atomic-recs-list>
```
btaillon, line 283: Regardless of whether or not you apply my other comments, I definitely think we should have a result template here. The default is more of a dev placeholder IMO.

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -225,19 +279,24 @@
...
-          />
+          <a href="https://coveo.com/" target="_blank">
+            <atomic-icon
+              class="footer-icon"
```
btaillon, line 293: It might be better to use the SVG directly here, especially since `atomic-icon` tries to have a 1:1 aspect ratio by default and this icon isn't a square.

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
```diff
@@ -32,7 +32,7 @@ export class AtomicIPXButton {
   /**
    * Whether the IPX layout is open.
    */
-  private isOpen = false;
+  private isOpen = this.host.classList.contains('btn-open');
```
btaillon, line 35: ❗️we don't want to expose classes outside the shadowDOM, and it looks like this would be exposed. Additionally, it's not prefixed with `atomic-` and uses a fairly generic name, therefore may be affected by other styling on the implementer's page.

I would recommend replacing this with a prop that has `mutable: true` and `reflect: true`, which will have the same effect without the consequences, and will also remove the need for `classList.add`/`classList.remove` in the code.

btaillon (changes requested):

packages/atomic/src/pages/examples/ipx.html
jbvouma: Removed ✅!

packages/atomic/src/pages/examples/ipx.html
jbvouma: At first when a user open a modal, we want to have recommandations (so no initial search).
When the user executes manually the initial search the placeholders are there (tested).

Is there a difference to have it in the JS or CSS if we want (in our case) to always hide the placeholders prior the inital search?

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
jbvouma: The main issue was to show the recommendations and search results in the same modal (recs first then hide it when an initial search is executed) but not at the same time.

I was looking for the simplest way to do it without having to create an `atomic-ipx-interface` which would have been an entire copy of the search one.

After discussion with JF, he advised me to use a similar class.

btaillon (approved):

dguerinCoveo (approved):
awesome!

PR #2691, opened 2023-01-24
feat(atomic): add result link for quickview header

olamothe:
So that's a big "scope down" of my initial idea, which was to make everything in the header configurable in result templates. 

So that someone could embed custom components in the header, for example, as we could see in the mockups.

I spent some time figuring out how to make it all work in the first place. The main problem is that we wanted to have arbitrary result templates components (ie: atomic-result-link, atomic-result-field, custom result template components, etc) working *outside* of their normal location (outside of the result list).

The modal is not located inside the result list/inside a result template, and we would need to do quite a bit of gymnastic to make it all work.

However, even with this working, I still had the problem of having unknown components included in the header and making sure things are displaying correctly. Which would require even more code, more documentation, more maintenance.

And at the end of they day, I'm uncertain of the level of usefulness of it all. So I prefer if we wait and see if there's any demand for it.

I've changed it so it's just a simple link with proper analytics tracking. I'm thinking this should be more than good enough for the time being.

https://coveord.atlassian.net/browse/KIT-2204

### DIFF
--- a/packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
+++ b/packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
@@ -9,7 +9,9 @@ import {
   Watch,
   Method,
   VNode,
+  Fragment,
 } from '@stencil/core';
+import CloseIcon from '../../../../images/close.svg';
 import {
   InitializableComponent,
   InitializeBindings,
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -83,11 +84,25 @@ export class AtomicQuickviewModal implements InitializableComponent {
   }
 
   private renderHeader() {
-    // TODO: Header should be slottable from result template definition
+    let headerContent: VNode | null = null;
```
louis-bompart, line 87: Is there a specific reason to use null v.s undefined? 🤔

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -83,11 +84,25 @@ export class AtomicQuickviewModal implements InitializableComponent {
...
+    if (this.result) {
+      const interactiveResult = buildInteractiveResult(this.bindings.engine, {
+        options: {result: this.result!},
```
louis-bompart, line 90: Do you really need `!` ?

louis-bompart (commented):
🤔 (will clone later)

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
olamothe: Not particularly, just that null is typically used in JSX when you want to render "nothing". 

We need to always render a header for the modal itself, just that it should be empty there is no result to use for the result link.

y-lakhdar (approved):

PR #2693, opened 2023-01-25
fix(atomic): fixed insight numeric facet not showing up

nathanlb:
The cause of the issue is from the facetId not being passed to the NumericFacetCommon object. As a result the `facetId` is recomputed and given 2 instances of the facet exists (because of the refine modal), the `facetId` is appended with some random string.
The facet doesn't show up because it's looking for values in the state using the incorrect `facetId`.

I also renamed `insights-panel` to `insight-panel` in some testing files.


louis-bompart (commented):
fix seems legit.
suggestion for fix-it-twice:
https://github.com/coveo/ui-kit/blob/05ad471d7f7c5d1f063d177fdf1af9a2a50893b7/packages/atomic/src/components/common/facets/numeric-facet-common.tsx#L180
add a check there.

nathanlb:
> fix seems legit. suggestion for fix-it-twice:
> 
> https://github.com/coveo/ui-kit/blob/05ad471d7f7c5d1f063d177fdf1af9a2a50893b7/packages/atomic/src/components/common/facets/numeric-facet-common.tsx#L180
> 
> 
> add a check there.

Thanks for the review!
Am I understanding correctly that you're suggesting I add validation that the `facetId` argument is defined? From what I could tell it makes sense for it to be marked as optional though given it isn't required by the numeric facet component and the numericFacetCommon object contains the logic to handle that.

lbergeron (approved):
Looks good to me!

olamothe:
> > fix seems legit. suggestion for fix-it-twice:
> > https://github.com/coveo/ui-kit/blob/05ad471d7f7c5d1f063d177fdf1af9a2a50893b7/packages/atomic/src/components/common/facets/numeric-facet-common.tsx#L180
> > 
> > add a check there.
> 
> Thanks for the review! Am I understanding correctly that you're suggesting I add validation that the `facetId` argument is defined? From what I could tell it makes sense for it to be marked as optional though given it isn't required by the numeric facet component and the numericFacetCommon object contains the logic to handle that.

I think @louis-bompart suggestion's is correct, however it's a bit more complex that what he suggest. It's true that setting the ID is optional for end users (the prop on the web component itself), but internally its not optional and so we generate one.

Reading the code and it seems the `NumericFacetCommon` is just badly isolated for the parent component on the ID generation stuff. There's some weird back and forth between the two classes.
And so you end up with problems like Nathan got here.

The fix it twice would be to refactor that. And the problem is also present for other types of facets.

For the scope of this PR I think your fix is fine, and it would be on us to revisit that.

PR #2696, opened 2023-01-25
refactor(atomic): rework atomic-icon-button to a functional component

olamothe:
The goal of this PR is to refactor the web component that we had for icon-button into a purely functional component.

The main reason is that you cannot easily pass down props with a web component (as opposed to a functional component, where it's a simple spread of props down stream): you have to recopy everything "manually".

This is causing issue when/if we need to send props down to the icon or button component, for example to pass accessibility properties on the button.

I've removed the `atomic-icon-button` component, which was marked as `@internal`. In theory this should be legit, as it means it was not appearing in any documentation.

However, in practice, it's still possible that an implementation out there is using it. 

Indeed, when a web component is registered in the DOM, anyone can in theory use it in a given page: it's not restricted to us only. 

If you prefer, I could put it back, but mark it as deprecated and not to be used. My gut feeling is telling me that it's overkill, however...

https://coveord.atlassian.net/browse/KIT-2222


nathanlb:
The change makes sense to me and we have no hard references to the atomic-icon-button component that I could find.
Only think is the icons don't look properly vertically centered in the buttons like they were before.

Screen Shot 2023-01-25 at 5 27 23 PM

packages/atomic/src/components/common/iconButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  icon: string;
+}
+
+export const IconButton: FunctionalComponent<IconButtonProps> = (props) => {
+  return (
+    <div class="relative" part="container">
+      <Button
+        {...props}
+        class="p-3 relative"
+        part="button"
```
btaillon, line 16: I think all parts in here should become required props, like either:
```ts
export interface IconButtonProps extends ButtonProps {
  badge?: VNode;
  buttonRef?: (el?: HTMLButtonElement) => void;
  icon: string;
  buttonPart: string;
  iconPart: string;
  badgePart: string;
}
```
or
```ts
export interface IconButtonProps extends ButtonProps {
  badge?: VNode;
  buttonRef?: (el?: HTMLButtonElement) => void;
  icon: string;
  partPrefix: string;
}
```

The reason is that all these components have styling, and therefore require parts to be customized. But also, a name like "button" or "icon" may not be relevant to a particular component, and a component may render multiple icon buttons.

packages/atomic/src/components/common/iconButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+      >
+        <atomic-icon
+          icon={props.icon}
+          class="w-4 h-4 shrink-0"
+          aria-hidden="true"
+          part="icon"
+        ></atomic-icon>
+      </Button>
+      {props.badge && (
+        <span class="absolute block h-4 w-4 text-center bg-primary text-on-primary rounded-[100%] text-xs leading-4 -top-2 -right-2">
```
btaillon, line 27: I think we should add a part for this

packages/atomic/src/components/insight/atomic-insight-edit-toggle/atomic-insight-edit-toggle.tsx
```diff
@@ -16,11 +17,12 @@ export class AtomicInsightEditToggle {
...
   public render() {
     return (
-      <atomic-icon-button
+      <IconButton
+        style="outline-neutral"
         icon={EditIcon}
-        labelI18nKey="edit-insight"
-        click-callback={this.clickCallback}
-        tooltip={this.tooltip}
+        ariaLabel="Edit"
```
btaillon, line 23: should this be localized?

packages/atomic/src/components/insight/atomic-insight-history-toggle/atomic-insight-history-toggle.tsx
```diff
@@ -15,11 +17,12 @@ export class AtomicInsightHistoryToggle {
...
   public render() {
     return (
-      <atomic-icon-button
+      <IconButton
+        style="outline-neutral"
         icon={Clockicon}
-        labelI18nKey="insight-history"
-        click-callback={this.clickCallback}
-        tooltip={this.tooltip}
+        ariaLabel="history"
```
btaillon, line 23: should this be localized?

btaillon (approved):

louis-bompart (approved):
lgtm

packages/atomic/src/components/insight/atomic-insight-edit-toggle/atomic-insight-edit-toggle.tsx
olamothe: Not necessary here. I

Insight panel in atomic is an "internal user only" component, not meant to be used directly by clients as of today, and so localization is not important at the moment.

packages/atomic/src/components/insight/atomic-insight-history-toggle/atomic-insight-history-toggle.tsx
olamothe: Idem.

olamothe:
> The change makes sense to me and we have no hard references to the atomic-icon-button component that I could find. Only thing is the icons don't look properly vertically centered in the buttons like they were before.
> 
> Before Screen Shot 2023-01-25 at 5 32 35 PM
> 
> After Screen Shot 2023-01-25 at 5 27 23 PM

Good catch ! Fixed.

nathanlb (approved):

PR #2697, opened 2023-01-25
feat(atomic): added prop to omit the modal focus trap

nathanlb:
[SVCC-2776](https://coveord.atlassian.net/browse/SVCC-2776)

The Atomic insight panel is currently exclusively used in the hosted insight panel builder and the focus trap is disabling the builder form elements from being interacted with which is pretty annoying.

This PR adds an optional prop to be able to render the atomic-modal without a focus-trap component.

A future improvement could be to still allow the modal to hide its sibling elements (under the modal) which would make lots of sense in the context of an insight panel or IPX which is only a widget on the screen.

[SVCC-2776]: https://coveord.atlassian.net/browse/SVCC-2776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -77,9 +77,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
       if (watchToggleOpenExecution !== this.currentWatchToggleOpenExecution) {
         return;
       }
-      if (!this.noFocusTrap) {
-        this.focusTrap!.active = true;
-      }
+      !this.noFocusTrap && (this.focusTrap!.active = true);
     } else {
       document.body.classList.remove(modalOpenedClass);
       if (isIOS()) {
@@ -88,9 +86,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
       if (watchToggleOpenExecution !== this.currentWatchToggleOpenExecution) {
         return;
       }
-      if (!this.noFocusTrap) {
-        this.focusTrap!.active = false;
-      }
+      !this.noFocusTrap && (this.focusTrap!.active = false);
     }
   }
 
@@ -135,7 +131,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   public render() {
     this.updateBreakpoints();
 
-    const content = (
+    const Content = () => (
       <article
         part="container"
         class={`flex flex-col justify-between bg-background text-on-background ${
@@ -191,7 +187,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
           onClick={(e) => e.target === e.currentTarget && this.close()}
         >
           {this.noFocusTrap ? (
-            content
+            <Content />
           ) : (
             <atomic-focus-trap
               role="dialog"
@@ -201,7 +197,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
               container={this.container ?? this.host}
               ref={(ref) => (this.focusTrap = ref)}
             >
-              {content}
+              <Content />
             </atomic-focus-trap>
           )}
         </div>
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -53,6 +53,7 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   @Prop({mutable: true}) container?: HTMLElement;
   @Prop({reflect: true, mutable: true}) isOpen = false;
   @Prop({mutable: true}) close: () => void = () => (this.isOpen = false);
+  @Prop({reflect: true}) noFocusTrap = false;
 
   @Event() animationEnded!: EventEmitter<never>;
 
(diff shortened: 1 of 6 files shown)


btaillon:
> A future improvement could be to still allow the modal to hide its sibling elements (under the modal) which would make lots of sense in the context of an insight panel or IPX which is only a widget on the screen.

Actually, doing this right away sounds like it would be a better idea, since disabling the focus trap allows all elements under the modal to be interacted with using the keyboard, and as you said this could be useful for IPX. You could add a `scope` property to `atomic-focus-trap`, whose default value would be `document.body`.

Then, you could replace `document.body` in the following with `this.scope`: https://github.com/coveo/ui-kit/blob/65867cb458ea556efdb4c6f347fb5a9dd4df5887/packages/atomic/src/components/common/atomic-focus-trap.tsx#L68-L70

And you could add a new condition after the following that returns/exits when the `focusedElement` isn't a descendant of `this.scope`: https://github.com/coveo/ui-kit/blob/65867cb458ea556efdb4c6f347fb5a9dd4df5887/packages/atomic/src/components/common/atomic-focus-trap.tsx#L121-L123

In the insight refine modal, you could pass the insight search interface as the scope.

nathanlb:
Wow @btaillon almost looks like you've thought about this before! Thanks for the solution I'll try it out.

nathanlb:
It works great for hiding but I still need to figure out the focus. When reaching the last focusable element the next one is still within the scope it's just hidden. Trying to find the right exit condition

nathanlb:
The full solution mentioned by @btaillon is what we should implement eventually but it'll take more time to do correctly than I currently have to get this fix out. The temporary solution to just remove the focus trap for the insight panel is suitable for me at the moment.

louis-bompart:
> The full solution mentioned by @btaillon is what we should implement eventually but it'll take more time to do correctly than I currently have to get this fix out. The temporary solution to just remove the focus trap for the insight panel is suitable for me at the moment.

I leave approval discretion to @btaillon , but temp solution seems 'alright' to me if the change from the 'tempfix' to the 'truefix' can be done without any breaking changes, and that the truefix is tracked on a Jira that is included in the post-release improvements of your project (and not the ongoing improvement oubliette epic ;) )

If either conditions aren't fullfilled, you may want to:
- either patch the fix in your project 💩 
- find the time to do the truefix rightaway ;)

nathanlb:
> I leave approval discretion to @btaillon , but temp solution seems 'alright' to me if the change from the 'tempfix' to the 'truefix' can be done without any breaking changes, and that the truefix is tracked on a Jira that is included in the post-release improvements of your project (and not the ongoing improvement oubliette epic ;) )
> 
> If either conditions aren't fullfilled, you may want to:
> 
> * either patch the fix in your project 💩
> * find the time to do the truefix rightaway ;)

I'll definitely create another jira which I can link here. SVCC or SVCINT will likely work on it very soon I just need this fix quicker than we're able to implement the absolute best version.

All the components concerned are internal so there should be no breakage.

pathiery:
@nathanlb turns out we need this for the IPX as well (since we can't have the focus trap on an embedded element).... Interested in knowing if I can leverage your work for that 🤔

nathanlb:
@pathiery yes you totally need this 👍 
I already have a branch with my early attempts at properly implementing "focus trapping for widgets". My plan is to create a jira in SVCC for this and do it separately. We can transfer the jira depending on who ends up doing it. Your need for the full solution will come sooner than ours.

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -76,7 +77,9 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
       if (watchToggleOpenExecution !== this.currentWatchToggleOpenExecution) {
         return;
       }
-      this.focusTrap!.active = true;
+      if (!this.noFocusTrap) {
+        this.focusTrap!.active = true;
+      }
```
btaillon, line 82: You could check for the absence of the ref instead.
```suggestion
      this.focusTrap && (this.focusTrap.active = true);
```

If you're in too much of a hurry, you can merge anyway

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -85,7 +88,9 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
       if (watchToggleOpenExecution !== this.currentWatchToggleOpenExecution) {
         return;
       }
-      this.focusTrap!.active = false;
+      if (!this.noFocusTrap) {
+        this.focusTrap!.active = false;
+      }
```
btaillon, line 93: Same as above
```suggestion
      this.focusTrap && (this.focusTrap.active = false);
```

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
```diff
@@ -130,67 +135,75 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
   public render() {
     this.updateBreakpoints();
 
+    const content = (
```
btaillon, line 138: Nitpick, but I generally prefer callbacks when separating parts of the state, even when we're rendering them either way.

Something like:
```tsx
const Content = () => ...;
...
return <something><Content /></something>
```

Since you're in a hurry to merge, it's not really a big deal.

btaillon (approved):

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
nathanlb: Ah I didn't know you guys accepted this kind of shorthand. I personally like it but sometimes people find it harder to read.

drosenfeld-coveo (approved):

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
nathanlb: Did it!

PR #2698, opened 2023-01-26
feat(headless-react): added useController hook

btaillon:
https://coveord.atlassian.net/browse/KIT-2236

This is a follow-up to https://github.com/coveo/ui-kit/pull/2683

# The first problem
Initializing Headless controllers in React isn't the most intuitive thing.
* We need to pass the correct dependency arrays to `useEffect` and `useMemo`.
  * E.g.: If we add the props as a dependency and they aren't memoized somewhere, we can end up in an infinite re-render loop.
* Because Headless calls the listener as soon as we call `subscribe`, if we're not careful, there will be a re-render right after the original render.
* If we need to rebuild the controller (e.g., when the engine or props are changed), the old state will still be present until the next call to the `subscribe` function's listener.

# The first solution
We can export a `useController` utility hook which handles all the above complexity for React developers, and allows us to patch any issues we may find and even improve performance.

I decided to create the hook so that it can be used like so:
```tsx
const { controller, state } = useController(buildFacet, engine, {
  options: { field: "objecttype" },
});
```

I decided to take three parameters as opposed to a callback which returns a controller because we need to know when the rebuild the controller. If we decide not to support rebuilding the controller and take the props as a one-time-only thing, then we are going against [React's principle of not omitting dependencies](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528).

I decided to return the controller and the state as two different objects, so that components are able to subscribe to changes in either like so:
```tsx
const { controller, state } = useController(buildFacet, engine, {
  options: { field: "objecttype" },
});

useEffect(() => {
  // Initialize something whenever the controller is initialized or replaced.
}, [controller]);

const valuesCount = useMemo(() => state.values.length, [state.values]);
```

I decided not to exclude the `subscribe` function from the returned `controller` because:
* The above approach is more akin to how React works, and easier for us to patch.
* The `state` as returned by `useController` is a snapshot of the state, not a getter, therefore the listener would theoretically only be able to access a stale state and a stale controller.

# The second problem
We need a way to somehow publish this new hook.

# The second solution
I decided to add a new package, `@coveo/headless-react`, which re-exports everything from `@coveo/headless` but requires React as a dependency and exports hooks.

The pros of this approach:
* If we find that some controllers, utilities, etc don't work that well in React, we are able to remove or substitute them for something more React-friendly.
* React applications only need one dependency: `@coveo/headless-react`.

The cons of this approach:
* We end up with another package to release, which almost completely duplicates `@coveo/headless` and takes up space in the CDN.
* Implementers may follow documentation that's specific to React, and also follow generic documentation, and they may think they need both `@coveo/headless` and `@coveo/headless-react`.
* With the current CI, we can't synchronize the version of two packages to always match, therefore `@coveo/headless-react` would have a completely different version than `@coveo/headless`, which would be very confusing, especially when reading `engine.state.version`.
* ❗️ Some weird workarounds would be necessary in order to export `@coveo/headless-react/insight` and so on, which I haven't included in this PR.
* ❗️ `@coveo/atomic` uses `@coveo/headless`, but `@coveo/atomic-react` would probably need `@coveo/headless-react` as a peer dependency instead. I haven't looked into it yet, but I'm pretty sure this would result in some atomic-react implementations having both bundles (unless bundlers are smart enough to detect that `@coveo/headless-react` is simply re-exporting the same dependency).

Alternative approaches:
* Making `@coveo/headless-react` standalone, which would have `@coveo/headless` as a peer dependency.
* Only including the utility in our doc and recommending clients copy and paste it.

# Possible improvement
* Instead of exporting `useController`, we could have a script that auto-generates hooks for every controller and copies their documentation from Headless (e.g., `useFacet(engine, props)`).
  * Might be a bit overkill, but the current approach gives the following error if we forget to pass the props, which isn't very friendly:
image

# What's in this PR
* Added the `@coveo/headless-react@1.0.0` package.
* Updated the `facet.fn.tsx` sample in the Headless-react samples.

# What's not in this PR
* The sub-packages of `@coveo/headless-react` (e.g., `@coveo/headless-react/insight`).
* Documentation for the new hook and package.
* Tests (this will require **a lot** of tests)
  * Server-side rendering tests


.deployment.config.json
```diff
@@ -48,6 +48,39 @@
...
+      "id": "deploy-headless-react-patch-to-s3-version",
+      "s3": {
+        "bucket": "{terraform.infra.infra.bucket_binaries}",
+        "directory": "proda/StaticCDN/headless-react/v$[HEADLESS_REACT_PATCH_VERSION]",
+        "source": "packages/headless-react/dist",
+        "parameters": {
+          "acl": "public-read"
+        }
+      }
+    },
```
btaillon, line 61: I copied almost everything from `@coveo/atomic-react`

packages/headless-react/package.json
```diff
@@ -0,0 +1,47 @@
+{
```
btaillon, line 1: Almost everything here is copied from `@coveo/atomic-react`, I basically just removed some unneeded dependencies.

packages/headless-react/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "rollup": "^2.79.0",
+    "rollup-plugin-polyfill-node": "^0.10.2",
+    "rollup-plugin-terser": "^7.0.2",
+    "@rollup/plugin-commonjs": "^22.0.2",
+    "@rollup/plugin-node-resolve": "^14.1.0",
+    "@rollup/plugin-typescript": "^8.5.0"
+  },
+  "peerDependencies": {
+    "react": ">=18.0.0",
+    "react-dom": ">=18.0.0"
```
btaillon, line 45: I'm hesitant on whether we should only support newer React versions for this package, just like `atomic-react` or support more versions.

I copied this from `atomic-react` for now.

packages/headless-react/src/hooks/controller-hook.ts
```diff
@@ -0,0 +1,139 @@
+import type {Controller, CoreEngine, Unsubscribe} from '@coveo/headless';
+import {deepEqual} from 'fast-equals';
+import {useEffect, useReducer, DependencyList, useRef} from 'react';
+
+function useDeepCompareMemo<T>(
```
btaillon, line 5: Will explain the need for each of these custom hooks below.

packages/headless-react/src/hooks/controller-hook.ts
```diff
@@ -0,0 +1,139 @@
...
+export function useController(
+  builder: (engine: CoreEngine, props?: object) => Controller,
+  engine: CoreEngine,
+  props?: object
+): UseControllerReturnType<Controller> {
+  useOnChange(builder, () =>
+    console.error(
+      'A Coveo Headless controller builder was changed in between renders. This can happen when a custom callback is passed as a builder instead of an build function imported directly from Headless.'
+    )
+  );
```
btaillon, line 125: If the user initializes a controller like so, React will detect `builder` as being changed on every single re-render:
```
const { controller, state } = useController((e, p) => buildFacet(e, p), engine, props);
```

This is because React uses `Object.is(a, b)` to compare whether a value in a dependency array was changed, which essentially compares pointers as opposed to a deep comparison, and callbacks are passed as-is to functions.

packages/headless-react/src/hooks/controller-hook.ts
```diff
@@ -0,0 +1,139 @@
...
+  const {controller, getStateSnapshot, onStateChanged} = useDeepCompareMemo(
+    () => buildController(builder, engine, props),
+    props ?? {},
+    [builder, engine]
+  );
```
btaillon, line 130: If we passed props as a normal dependency in a dependency array, then the following would be detected as new props on every single new re-render:
```tsx
const { controller, state } = useController(buildFacet, engine, { options: { field: 'filetype' } });
```

This is because, just like I mentioned above, React compares pointers, not values.

Unfortunately, since the above example is intuitive and maybe even something we would want to recommend, we need to do a deep comparison.

packages/headless-react/src/hooks/controller-hook.ts
```diff
@@ -0,0 +1,159 @@
...
+export function useController(
+  builder: (engine: CoreEngine, props?: object) => Controller,
+  engine: CoreEngine,
+  props?: object
+): UseControllerReturnType<Controller> {
+  if (useHasChanged([builder])) {
+    console.error(
+      'A Coveo Headless controller builder was changed in between renders. This can happen when a custom callback is passed as a builder instead of an build function imported directly from Headless.'
+    );
+  }
```
btaillon, line 151: If the user initializes a controller like so, React will detect `builder` as being changed on every single re-render:
```
const { controller, state } = useController((e, p) => buildFacet(e, p), engine, props);
```

This is because React uses `Object.is(a, b)` to compare whether a value in a dependency array was changed, which essentially compares pointers as opposed to a deep comparison, and callbacks are passed as-is to functions.

I made the `useOnChange` hook so that we could log an error when the reference is changed. It works almost exactly like `useEffect`, except:
* It's called synchronously on render as opposed to after the render.
* It's only called when the value is changed, never on the initial value.

packages/headless-react/src/hooks/controller-hook.ts
```diff
@@ -0,0 +1,159 @@
...
+  const {controller, getStateSnapshot, onStateChanged} = useDeepCompareMemo(
+    () => buildController(builder, engine, props),
+    props ?? {},
+    [builder, engine]
+  );
+  const state = useSyncMemoizedStore(onStateChanged, getStateSnapshot);
```
btaillon, line 157: React made the [`useSyncExternalStore`](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) hook, which does pretty much exactly what we need:
* Has an initial state that updates when the store is changed (in this case, `onStateChanged`/`getStateSnapshot` are the store).
* Re-renders the component with the new state whenever the listener is called.

I even tested `useSyncExternalStore` with those two exact parameters.

However, `useSyncExternalStore` causes an infinite re-rendering loop, like it's detecting `onStateChanged` and/or `getStateSnapshot` to be different on every single render (even when other hooks don't detect a change for the same dependencies). `useSyncExternalStore` is meant to subscribe to a store that was statically imported from another module, not a memoized value, so I believe that's why I'm not finding much about the issue online. I tried messing with `useSyncExternalStore` for over an hour.

For now, this will do. I'll add a lot of tests, and we can update this in the future.

packages/samples/headless-react/src/components/facet-manager/facet-manager.class.tsx
```diff
@@ -8,7 +8,13 @@ import {
 import {Children, Component, ContextType, ReactElement} from 'react';
 import {AppContext} from '../../context/engine';
 
-type FacetManagerChild = ReactElement<{facetId: string}>;
+export interface FacetManagerClassChildProps {
+  controllerOptions: {facetId: string};
+  /** @deprecated */
+  facetId?: string;
```
btaillon, line 14: Will remove when I make the final PR replacing all samples.

packages/samples/headless-react/package.json
```diff
@@ -1,11 +1,12 @@
 {
   "name": "@coveo/headless-react-samples",
-  "description": "Samples with @coveo/headless & React",
+  "description": "Samples with @coveo/headless-react & React",
   "version": "0.0.0",
   "private": true,
   "dependencies": {
     "@coveo/auth": "^1.9.1",
     "@coveo/headless": "^2.0.0",
```
btaillon, line 8: Will remove once I replaced it in all samples. This PR is long enough

packages/samples/headless-react/src/components/facet/facet.fn.tsx
```diff
@@ -1,16 +1,19 @@
...
-export const Facet: FunctionComponent<FacetProps> = (props) => {
-  const {controller} = props;
-  const [state, setState] = useState(controller.state);
-
-  useEffect(() => controller.subscribe(() => setState(controller.state)), []);
+export const Facet: FunctionComponent<FacetProps> = ({controllerOptions}) => {
+  const {engine} = useContext(AppContext);
+  const {controller, state} = useController(buildFacet, engine!, {
+    options: controllerOptions,
+  });
```
btaillon, line 16: Here's what this look like in the dev tools (notice the `hooks` section): 
image

packages/headless-react/src/hooks/controller-hook.ts
olamothe: edit: Ok, I saw in your description that this is already something you've thought of as a possible improvement.

Great !

btaillon:
TODO: test this with the url manager, since it has a stricter listener

olamothe (commented):
That's a great PR with a thorough explanation !

The code is however not trivial to read/understand. But this can be expected I guess, with the pretty complex problem it's trying to solve. 

What kind of testing do you have in mind when you say this will need a lot of tests ? 

Because obviously what we had before in our samples "worked", but had a lot of limitation (ie: we do not change a controller once it's created, or we don't support the global engine being modified), or some components were rendering needlessly, so it's more of a subtle performance and usability problem.

Which is not really easy to catch with automated test.

> Alternative approaches:
> Making @coveo/headless-react standalone, which would have @coveo/headless as a peer dependency.

So I think that should be the way to go ? 🤔 

Naturally, I tend to see these React hook as a "sister" library of Headless, and it's not standalone as much as an additional tool in our toolbelt.

As for the name, should we emphasize the fact that it's focused on hook (`@coveo/headless-react-hooks`) or do you think it's possible we would start adding more React utility in there that are not hooks related ?

louis-bompart:
Eh boi that's a nice one. In this first comment, I'll be focusing on the cons of the second solution.
> We end up with another package to release, which almost completely duplicates @coveo/headless and takes up space in the CDN.

Could we not tinker the bundler so that @coveo/headles-react@x.y.z CDN file exports its own sauce and when it need to refer to @coveo/headless, point towards the @coveo/headless@x.y.z CDN files

> With the current CI, we can't synchronize the version of two packages to always match, therefore @coveo/headless-react would have a completely different version than @coveo/headless, which would be very confusing, especially when reading engine.state.version.

If @coveo/headless-react specify properly its @coveo/headless version and enforce a match (i.e should be a prod and peerDependencies with exact version), should be fine.


>❗️ @coveo/atomic uses @coveo/headless, but @coveo/atomic-react would probably need @coveo/headless-react as a peer dependency instead.

Why not direct prod dependency?

> but I'm pretty sure this would result in some atomic-react implementations having both bundles (unless bundlers are smart enough to detect that @coveo/headless-react is simply re-exporting the same dependency).

I reckon that if the two version of @coveo/headless perfectly match, they should be deduped by npm and thus by the bundler too. :famouslastwords:

louis-bompart:
To rebounce on naming the package, I like @olamothe suggestions, especially to avoid confusion with
https://www.npmjs.com/package/@coveo/create-headless-vue
and future: https://www.npmjs.com/package/@coveo/create-headless-react (which will replace cra-template)

packages/headless-react/src/hooks/controller-hook.ts
btaillon: Yes! I agree that this would be more idiomatic. I also just thought of another advantage: we would know at build-time whether one of our controllers doesn't match the expected parameters.

I will go with this approach then.

packages/headless-react/src/hooks/controller-hook.ts
```diff
@@ -0,0 +1,159 @@
...
+
+function buildController(
+  builder: (engine: CoreEngine, props?: object) => Controller,
+  engine: CoreEngine,
+  props?: object
+) {
+  const headlessController = builder(engine, props);
+  const {subscribe: _, state: __, ...methods} = headlessController;
+  return {
+    controller: bindMethods(methods, headlessController),
```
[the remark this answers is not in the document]
btaillon: Because some of our controllers reference `this`, and by default, `this` is whichever object is present before the period (`.`), which in this case would only be methods.

packages/headless-react/src/hooks/controller-hook.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function bindMethods<T extends object>(methods: T, obj: object): T {
+  Object.entries(methods).forEach(([key, value]) => {
+    if (typeof value === 'function') {
+      (methods as Record<string, unknown>)[key] = value.bind(obj);
+    } else if (typeof value === 'object') {
+      bindMethods(value, (obj as Record<string, object>)[key]);
+    }
+  });
```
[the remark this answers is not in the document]
btaillon: Agreed

btaillon:
What I gather from both of you guys is that we should have the package instead of `@coveo/headless-react-utils`, which would only contain utilities. It would also ideally include one `use` method per controller. I feel good about this approach, and will work on this now.

btaillon:
> What kind of testing do you have in mind when you say this will need a lot of tests ?
> 
> Because obviously what we had before in our samples "worked", but had a lot of limitation (ie: we do not change a controller once it's created, or we don't support the global engine being modified), or some components were rendering needlessly, so it's more of a subtle performance and usability problem.

I was thinking of making either unit tests or functional tests (somewhere in-between), where I would use a real engine but fake controllers.

PR #2700, opened 2023-01-30
feat(atomic): quickview better accessibility support

olamothe:
Changes include:
* Aria live support for quickview pagination in modal. Reads out loud when a quickview is loading/ when a new one is loaded, as well as the pager (Quickview X out of Y).
* Added better support for minimize keywords highlight button, with `aria-controls`.
* Added better support for keywords navigation, with a `FieldsetGroup`, which announce how you can navigate between hits in a quickview (and the number of occurrences)

https://coveord.atlassian.net/browse/KIT-2199


packages/atomic/src/components/search/result-template-components/quickview-iframe/quickview-iframe.tsx
```diff
@@ -7,7 +7,9 @@ const writeDocument = (documentWriter: Document, content: string) => {
   documentWriter.open();
   documentWriter.write(content);
   documentWriter.close();
-  documentWriter.body.scrollTop = 0;
+  if (documentWriter.scrollingElement) {
```
olamothe, line 10: Kind of unrelated to this PR, just a bug fix for the scrollbar when a quickview load.

Not all documents will use to body to scroll (this depends on the type of document in the index), so using `scrollingElement` is a better catch all mechanism.

packages/atomic/src/components/common/atomic-focus-trap.tsx
```diff
@@ -34,7 +34,13 @@ export class AtomicFocusTrap {
   private readonly hiddenElements: Element[] = [];
 
   hide(element: Element) {
-    if (element.hasAttribute('aria-hidden')) {
+    // aria-hidden -> already hidden
+    // aria-live or atomic-aria-live -> must not be hidden otherwise it won't announce dynamic changes in the live region
+    if (
```
olamothe, line 39: This was discussed on slack: Essentially, the focus trap (when the modal is opened) would add `aria-hidden=true` on the aria live region, preventing it from being announced by the screen reader.

packages/atomic/src/components/search/result-template-components/atomic-quickview/atomic-quickview.tsx
```diff
@@ -105,6 +109,17 @@ export class AtomicQuickview implements InitializableComponent {
...
+        this.quickviewAriaMessage = this.bindings.i18n.t('quickview-loading');
+      } else {
+        if (!this.quickviewState.isLoading) {
+          this.quickviewAriaMessage = this.bindings.i18n.t('quickview-loaded', {
+            first: this.quickviewState.currentResult,
+            last: this.quickviewState.totalResults,
+            title: this.result.title,
+          });
+        }
+      }
```
louis-bompart, line 122: `if(a) else if (!a)`

am I misreading?

I would do
```suggestion
      this.quickviewAriaMessage = this.quickviewState.isLoading
        ? this.bindings.i18n.t('quickview-loading')
        : this.bindings.i18n.t('quickview-loaded', {
            first: this.quickviewState.currentResult,
            last: this.quickviewState.totalResults,
            title: this.result.title,
          });
```

louis-bompart (changes requested):
confused by the if-else-if

louis-bompart (approved):

packages/atomic/src/components/search/result-template-components/atomic-quickview-sidebar/atomic-quickview-sidebar.tsx
```diff
@@ -134,28 +143,35 @@ const Keywords: FunctionalComponent<
...
+                <div class="flex px-2">
+                  <IconButton
+                    partPrefix="sidebar-next"
+                    icon={ArrowDown}
+                    disabled={!wordIsEnabled}
+                    style="text-transparent"
+                    class="border-0"
+                    ariaLabel={i18n.t('next')}
+                    title={i18n.t('next')}
+                    onClick={() => keyword.navigateForward()}
```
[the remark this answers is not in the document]
olamothe: It seems overkill, since the "selection" is not even something you can act on ? 
ie: You can do nothing with the selection, it's purely a visual trick that changes the color of the background of individual keyword.

Tempted to just leave it as is for now ?

PR #2702, opened 2023-01-30
feat(atomic): add container part to breadbox

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2252
https://discuss.coveo.com/t/missing-shadow-parts-custom-properties/7355/35

### DIFF
--- a/packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
+++ b/packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
@@ -37,7 +37,7 @@ const ELLIPSIS = '...';
  * The `atomic-breadbox` component creates breadcrumbs that display a summary of the currently active facet values.
  *
  * @part container - The container of the whole component, list & label.
- * @part breadcrumb-list-container - The container for the list of breadcrumb buttons.
+ * @part breadcrumb-list-container - The container of the list of breadcrumb buttons.
  * @part breadcrumb-list - The list of breadcrumb buttons.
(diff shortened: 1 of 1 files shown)


louis-bompart (approved):

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -36,6 +36,7 @@ const ELLIPSIS = '...';
 /**
  * The `atomic-breadbox` component creates breadcrumbs that display a summary of the currently active facet values.
  *
+ * @part container - The container of the component.
```
btaillon, line 39: I think we should mention here that this container surrounds the label and breadcrumb list.

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -405,7 +406,7 @@ export class AtomicBreadbox implements InitializableComponent {
     }
 
     return (
-      <div class="text-on-background text-sm flex">
+      <div part="container" class="text-on-background text-sm flex">
```
btaillon, line 409: It looks like a part is also missing for the element with `relative grow` below, which might get annoying if somebody tries to change the layout of the component (e.g., changing it to use two lines).

We could, instead of adding a part to the element below, move its styling into the `ul` element and remove the `div`.

btaillon (approved):

PR #2703, opened 2023-01-30
feat(headless): add instant results analytics

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2247

### DIFF
--- a/packages/headless/src/features/search/search-slice.ts
+++ b/packages/headless/src/features/search/search-slice.ts
@@ -2,6 +2,7 @@ import {createReducer, PayloadAction} from '@reduxjs/toolkit';
 import {
   executeSearch,
   fetchFacetValues,
+  fetchInstantResults,
   fetchMoreResults,
   fetchPage,
 } from './search-actions';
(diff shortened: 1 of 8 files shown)


y-lakhdar (approved):

louis-bompart (approved):

packages/headless/src/features/search/search-slice.ts
```diff
@@ -90,6 +91,9 @@ export const searchReducer = createReducer(
       state.response.facets = action.payload.response.facets;
       state.response.searchUid = action.payload.response.searchUid;
     });
+    builder.addCase(fetchInstantResults.fulfilled, (state, action) => {
+      state.response.searchUid = action.payload.searchUid;
+    });
```
btaillon, line 96: If we use instant results and then right click a result's title, is it associated with the current search or the instant results search?

ThibodeauJF: Hmm it always seems to get the `search.searchResponseId` value actually, which stems from the "true search".
This PR fixes the search event ID but not the click event associated ID, damn

ThibodeauJF: We might need to store a separate id for instant results and leverage it in the analytics, I don't know how complex this can get.

ThibodeauJF: https://coveord.atlassian.net/browse/KIT-2254

Let's get this merged, then I'll attack this one

PR #2706, opened 2023-02-01
test(atomic): add cypress test for quickview 

olamothe:
Pretty straightforward hopefully, just adds cypress tests for Quickview.

One thing I had to fight with before finding the problem was the config we had for cypress (not sure why/where it comes from exactly..) which set `experimentalFetchPolyfill` to true.

This caused all kind of weird problem with the low level fetch function that retrieves an ArrayBuffer for the HTML content of the preview in Headless. Also caused some problem with the `query-error` component.

https://coveord.atlassian.net/browse/KIT-2250


packages/atomic/cypress/e2e/result-list/result-components/quickview.cypress.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const addQuickviewInResultList = () =>
+  addResultList(
+    buildTemplateWithSections({
+      actions: generateComponentHTML(quickviewComponent),
+    })
+  );
+
+const openModal = () => {
+  cy.intercept('**/build/iconButton*').as('quickviewResources');
```
olamothe, line 21: This intercept/wait logic is needed to workaround a bug currently in atomic-modal, where if it gets opened "too fast" before it has a chance to load internal resources, it fails to perform it's job properly.

It's a bug that's currently only apparent for Cypress, since here we are interested in opening the Quickview ASAP, which causes flakyness in the test execution.

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -84,22 +90,25 @@ export class AtomicQuickviewModal implements InitializableComponent {
...
     this.content = undefined;
     this.result = undefined;
+    this.interactiveResult = undefined;
   }
 
   private renderHeader() {
     let headerContent: VNode | null = null;
     if (this.result) {
-      const interactiveResult = buildInteractiveResult(this.bindings.engine, {
+      this.interactiveResult = buildInteractiveResult(this.bindings.engine, {
```
olamothe, line 99: Issue found while writing the tests: analytics on the header link needs to be a component instance, or you end up with improper reference to the current result.

olamothe:
Whelp looks like this will need more refinement for the whole test suite, some tests started to fail with the fetch polyfill changes.

btaillon (approved):

ThibodeauJF (approved):

PR #2709, opened 2023-02-01
fix(atomic,headless): log proper search ID with instant results

ThibodeauJF:
Undoes some of https://github.com/coveo/ui-kit/pull/2703 for a better solution

### DIFF
--- a/packages/headless/src/api/analytics/instant-result-analytics.ts
+++ b/packages/headless/src/api/analytics/instant-result-analytics.ts
@@ -1,0 +1,23 @@
+import {SearchAnalyticsProvider} from './search-analytics';
+
+export class InstantResultsAnalyticsProvider extends SearchAnalyticsProvider {
+  private get instantResultsSearchUid() {
+    const state = this.getState().instantResults;
+    if (!state) {
+      return null;
+    }
+
+    for (const id in state) {
+      for (const query in state[id].cache) {
+        if (state[id].cache[query].isActive) {
+          return state[id].cache[query].searchUid;
+        }
+      }
+    }
+
+    return null;
+  }
+  public getSearchUID(): string {
+    return this.instantResultsSearchUid || super.getSearchUID();
+  }
+}
--- a/packages/headless/src/features/instant-results/instant-result-analytics-actions.ts
+++ b/packages/headless/src/features/instant-results/instant-result-analytics-actions.ts
@@ -1,0 +1,33 @@
+import {InstantResultsAnalyticsProvider} from '../../api/analytics/instant-result-analytics';
+import {Result} from '../../api/search/search/result';
+import {
+  partialDocumentInformation,
+  documentIdentifier,
+  validateResultPayload,
+  makeAnalyticsAction,
+  AnalyticsType,
+  ClickAction,
+  SearchAction,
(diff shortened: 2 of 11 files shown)


packages/headless/src/features/instant-results/instant-result-analytics-actions.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+    (getState) => new InstantResultsAnalyticsProvider(getState)
+  );
+
+export const logInstantResultsSearch = (): SearchAction =>
+  makeAnalyticsAction(
+    'analytics/instantResult/searchboxAsYouType',
+    AnalyticsType.Search,
+    (client) => client.makeSearchboxAsYouType(),
+    (getState) => new InstantResultsAnalyticsProvider(getState)
```
ThibodeauJF, line 32: Now that both events have this provider, no more issues, it will get the Search ID from the "active" instant result list.

packages/headless/src/features/query/query-analytics-actions.ts
```diff
@@ -10,10 +10,3 @@ export const logSearchboxSubmit = (): SearchAction =>
     AnalyticsType.Search,
     (client) => client.makeSearchboxSubmit()
   );
-
-export const logSearchboxAsYouType = (): SearchAction =>
```
olamothe, line 14: Isn't this a breaking change ? Was it exported globally ?

And someone could want to code an actual "search as you type" kind of experience (not instant result per say), and so that action would still be useful in this case I'd say.

packages/headless/src/features/instant-results/instant-results-slice.ts
```diff
@@ -74,6 +84,8 @@ const makeEmptyCache = (
     error: null,
     results: [],
     expiresAt: 0,
+    isActive: false,
```
olamothe, line 87: This flag is meant to represent when there is a focus on the search box VS blur ?

packages/headless/src/api/analytics/instant-result-analytics.ts
```diff
@@ -0,0 +1,23 @@
+import {SearchAnalyticsProvider} from './search-analytics';
+
+export class InstantResultsAnalyticsProvider extends SearchAnalyticsProvider {
+  private get instantResultsSearchUid() {
+    const state = this.getState().instantResults;
```
olamothe, line 5: Is there a reasonable scenario where that property would be undefined ?

When we instantiate that class, in `instant-result-analytics-action`, is the `getState()` return typed properly such that we know it's always defined ?

If so, we could have a constructor that enforce the proper type.

olamothe (approved):
Neat !

packages/headless/src/features/query/query-analytics-actions.ts
ThibodeauJF: No it's past of the past PR I made yesterday and its not exported

packages/headless/src/api/analytics/instant-result-analytics.ts
ThibodeauJF: Done ✔️

louis-bompart (approved):

PR #2712, opened 2023-02-02
fix(atomic): make suggestion (button) text left align by default

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2257


btaillon (approved):

btaillon:
This should probably be a `feat` instead of `fix` since it has *some* chances of breaking an existing implementation's styling.

PR #2720, opened 2023-02-08
fix(headless): remove facetSearch analytics

louis-bompart:
KIT-2261

-----

Given that:
- the facetSearch call does not return a searchUUID,
- a search analytics log requires a searchUUID
- how the feature work (see jira)

I reckon that we shouldn't log analytics on facet search. This is further confirmed imo by the fact that the old JSUI doesn't log analytics on facet search either.

draft to run the tests

### DIFF
--- a/packages/headless/src/features/facets/facet-set/facet-set-analytics-actions.ts
+++ b/packages/headless/src/features/facets/facet-set/facet-set-analytics-actions.ts
@@ -46,19 +46,6 @@ export const logFacetShowLess = (facetId: string): SearchAction =>
     }
   );
 
-export const logFacetSearch = (facetId: string): SearchAction =>
-  makeAnalyticsAction(
-    'analytics/facet/search',
-    AnalyticsType.Search,
-    (client, state) => {
-      validatePayload(facetId, facetIdDefinition);
-      const stateForAnalytics = getStateNeededForFacetMetadata(state);
-      const metadata = buildFacetBaseMetadata(facetId, stateForAnalytics);
-
-      return client.makeFacetSearch(metadata);
-    }
-  );
-
 export interface LogFacetUpdateSortActionCreatorPayload {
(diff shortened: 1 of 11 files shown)


louis-bompart:
@coveo/search should I squash and merge this PR or rebase merge?
I splitted my changes, one commit per package, atomic and quantic only have test changes so 🤷 
The fix for atomic & quantic is technically the bump of the headless version that'll occur on publish

packages/headless/src/features/facets/facet-set/facet-set-analytics-actions.ts
```diff
@@ -46,19 +46,6 @@ export const logFacetShowLess = (facetId: string): SearchAction =>
     }
   );
 
-export const logFacetSearch = (facetId: string): SearchAction =>
```
ThibodeauJF, line 49: That's good that we didn't export it

louis-bompart: Well if it was exported, I think I would've no-op it, deprecate it, and a console.warn.

olamothe: So, if we compare with what the JSUI is doing, it's indeed not logging a search event.

It is logging a custom event though... 🤔

So I think the actual proper fix is to keep everything else the same, except changing the event type in `coveo.analytics`, and the type on line 52 here to `AnalyticsType.Custom`


Those custom event should then not be causing any incoherent events in UA.

olamothe: edit: After discussion in scrum removing this event is the right thing, so disregard the above.

olamothe (approved):

nathanlb (approved):

PR #2722, opened 2023-02-10
feat(quantic): quantic feedback modal component created

mmitiche:
[SFINT-4801](https://coveord.atlassian.net/browse/SFINT-4801)

- New Quantic Feedback Modal component created.
- This component uses the new[ Lightning Modal](https://developer.salesforce.com/docs/component-library/bundle/lightning-modal/documentation) component.
- This component overlays a message modal on top of the current app window, the modal contains a form that allows the user to give feedback.

### Usage
To use the component, follow these steps:

1. Import the component:
```
import FeedbackModal from 'c/quanticFeedbackModal';
```
2. Open the modal:
```
    await FeedbackModal.open({
      size: 'small',
      label: 'Example feedback modal',
      description: 'Example feedback modal',
      options: this.prepareOptions(),
      submitFeedback: () => {},
    });
```

### Demo:

https://user-images.githubusercontent.com/86681870/218473097-85458a9c-9d40-49e8-95a5-e0dbe56d7b46.mov

#### When invalid properties are provided:
QFME

### Tests:
This component is now covered by unit tests, because a lot of flakiness has been noticed when testing the component with e2e Cypress tests, a random occasional error occurs with the Lightning Modal component when testing with cypress.

Since the Lightning Modal is relatively a new Lightning component, the stubs (programs that simulate components that a module undergoing tests depends on) of the components used by the Lightning Modal component are not yet included in the `sfdx-lwc-jest`. That's why I had to create these stubs under `force-app/test/jest-mocks/lightning`.

More info about creating stubs of the lightning component can be found [here](https://trailhead.salesforce.com/content/learn/modules/test-lightning-web-components/mock-other-components)

The stub of the Lightning Modal component was copied from the lwc-recipes folder: https://github.com/trailheadapps/lwc-recipes/blob/main/force-app/test/jest-mocks/lightning/modal.js.

[SFINT-4801]: https://coveord.atlassian.net/browse/SFINT-4801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/cypress/e2e/default-1/feedback-modal/feedback-modal.cypress.ts
```diff
@@ -0,0 +1,245 @@
+import {openModal} from '../../../page-objects/actions/action-open-modal';
+import {configure} from '../../../page-objects/configurator';
+import {scope} from '../../../reporters/detailed-collector';
+import {stubConsoleError} from '../../console-selectors';
+import {FeedbackModalActions as Actions} from './feedback-modal-actions';
+import {FeedbackModalExpectations as Expect} from './feedback-modal-expectations';
+
+const invalidArrayError = 'The options provided are not in a valid table.';
```
mmitiche, line 8:
```suggestion
const invalidArrayError = 'The options provided are not in a valid array.';
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,215 @@
...
+      if (!this.error) {
+        this.validateOptions();
+      }
+      this.validated = true;
+    });
+  }
+
+  validateOptionList() {
+    if (!Bueno.isArray(this.options)) {
+      console.error('The options provided are not in a valid table.');
```
mmitiche, line 84:
```suggestion
      console.error('The options provided are not in a valid array.');
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,215 @@
...
+   * Handles the details input change.
+   * @param {CustomEvent} event
+   */
+  handleDetailsChange(event) {
+    this.detailsInputError = false;
+    this.detailsValue = event.target.value;
+  }
+
+  /**
+   * Indicates whether the reason input should be displayed.
```
mmitiche, line 165:
```suggestion
   * Indicates whether the details input should be displayed.
```

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -484,4 +484,44 @@
...
+    <customLabels>
+      <label>Expliquer pourquoi</label>
+      <name>quantic_ExplainWhy</name>
+    </customLabels>
+    <customLabels>
+      <label>Sélectionnez la raison</label>
+      <name>quantic_SelectTheReason</name>
+    </customLabels>
+    <customLabels>
+      <label>Quelle était la raison ?</label>
```
mmitiche, line 496:
```suggestion
      <label>Quelle était la raison?</label>
```

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -484,4 +484,44 @@
       <label>Voir moins</label>
       <name>quantic_SmartSnippetShowLess</name>
     </customLabels>
+    <customLabels>
+      <label>Expliquer pourquoi</label>
```
SimonMilord, line 488: I don't want to nitpick but I was wondering for consistency do we want "Expliquez" instead? I see you are already using 'Sélectionnez' or maybe the latter should end with 'er'?

SimonMilord (approved):
GG 🔥🔥🔥

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,215 @@
...
+    this.detailsInputError = false;
+    this.detailsValue = event.target.value;
+  }
+
+  /**
+   * Indicates whether the details input should be displayed.
+   * @returns {boolean}
+   */
+  get displayDetailsInput() {
+    return this.optionsValue === 'other';
```
lbergeron, line 169: Since the `other` option  has a special behavior, it should appear in the `options` documentation.

However, considering the options can be customized, what would you think of accepting these optional attributes when configuring the options: `withDetails`, `optionalDetails` instead of relying on a special option value?

It would allow an implementer to include details for more than one option and/or make the details optional if needed.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,215 @@
...
+ * @category Search
+ * @category Insight Panel
+ */
+export default class QuanticFeedbackModal extends LightningModal {
+  /**
+   * @api
+   * The list of options to be displayed in the modal.
+   * @type {Array<{label: string, value: string}>}
+   */
+  @api options;
```
lbergeron, line 31: I'm thinking this component might be a good candidate to configure options using a slot. It could be a nice way to support different option types, like regular options and comment options.

```html
<c-quantic-feedback-modal>
  <c-quantic-feedback-option value="did-not-answer" label="This didn't answer my question" />
  <c-quantic-feedback-comment-option value="partially-answer" label="This only partially answered my question" optional-comment />
  <c-quantic-feedback-comment-option value="other" label="Other" />
</c-quantic-feedback-modal>
```

Maybe it is too much. Let me know what you think of the idea.

lbergeron (commented):
I have a doubt about configuring the options using JSON, but it looks very good!

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
```diff
@@ -0,0 +1,341 @@
...
+          selectors.successMesage
+        );
+        expect(successMesage).not.toBeNull();
+        expect(functionsMocks.submitFeedback).toHaveBeenCalledWith({
+          details: undefined,
+          value: exampleValue,
+        });
+      });
+
+      describe('when the option "others" is selected', () => {
```
mmitiche, line 312:
```suggestion
      describe('when the option "other" is selected', () => {
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/feedbackForm.css
```diff
@@ -0,0 +1,11 @@
+textarea {
+  resize: none;
```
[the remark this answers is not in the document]
mmitiche: There would be no issue if we allow the vertical resize of the textarea, the only reason I'm preventing the resize is to fully respect the mockups  of this component we there is no resize button. displayed in the textarea.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
mmitiche: Very neat idea! this would add a very useful freedom when using the component, will implement that 👌

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
mmitiche: The only issue I found with this is that this component is not supposed to be include in the HTML template like all the the other LWC components.
This component is a bit special since it extends `LightningModal` instead of `LightningElement` and the we can use it only by importing it in a JS file and opening it like so: 
```   
await FeedbackModal.open({
      size: 'small',
      label: 'Example feedback modal',
      description: 'Example feedback modal',
      options: this.prepareOptions(),
      submitFeedback: () => {},
    });
```
We can pass props with this static method `open` but not sure we can pass slots.

packages/quantic/force-app/main/translations/fr.translation-meta.xml
mmitiche: nitpicking is what PR reviews are for 👌 thanks!
will ask @acarpentier01  for help to make sure the labels are correct!

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -917,4 +917,74 @@
       <protected>false</protected>
       <shortDescription>Show less</shortDescription>
     </labels>
+    <labels>
```
mmitiche, line 920: @acarpentier01 your help will be much appreciated here, let me know if there are any improvements you think I can add to these labels.

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -484,4 +484,44 @@
       <label>Voir moins</label>
       <name>quantic_SmartSnippetShowLess</name>
     </customLabels>
+    <customLabels>
```
mmitiche, line 487: @acarpentier01, here are the french translations.
any feedback would be welcome, thank you!

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/feedbackForm.html
```diff
@@ -0,0 +1,36 @@
+<template>
+  <template if:true={validated}>
+    <lightning-modal-header label={labels.explainWhy}></lightning-modal-header>
+    <lightning-modal-body>
+      <div class="slds-var-p-around_x-large">
+        <lightning-radio-group class="slds-var-m-bottom_large" name="feedback" label={labels.selectTheReason}
+          options={options} value={feedbackValue} type="radio" onchange={handleOptionChange}></lightning-radio-group>
```
erocheleau, line 7:
```suggestion
          options={options} value={feedbackValue} type="radio" onchange={handleFeedbackReasonChange}></lightning-radio-group>
```

Maybe?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/feedbackForm.html
```diff
@@ -0,0 +1,36 @@
...
+        <div class="details-input__container">
+          <template if:true={displayDetailsInput}>
+            <div class={detailsFormClass}>
+              <label class="slds-form-element__label details-input__label" for="details-input">
+                {labels.whatWasTheReason}</label>
+              <div class="slds-form-element__control">
+                <textarea rows="4" id="details-input" aria-describedby="details-error" class="slds-textarea"
```
erocheleau, line 14: Are you sure about the `aria-describedby` `details-error` here?
Seems like a described by should be used to describe the purpose of the field/input, not described by the error message?

erocheleau: I think it might make sense following this page:
https://www.w3.org/WAI/tutorials/forms/notifications/

But also there is the pattern under : `Binary messages`
If the error message "appears" while you're typing or after an action and is not already visible in the page, you will need to use a `aria-live` to tell screen readers that a new message has appeared in the page.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
erocheleau: Good question Mehdi :) 
I would've also advocated for slots but this is a good point if you can't figure out where the slots go in this describe object, it's definitely okay to do it like you did.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -917,4 +917,74 @@
...
+      <fullName>quantic_WhatWasTheReason</fullName>
+      <value>What was the reason?</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>What was the reason?</shortDescription>
+    </labels>
+    <labels>
+      <fullName>quantic_FillOutField</fullName>
+      <value>Please fill out this field</value>
```
acarpentier01, line 943: If you apply my previous comment, then consider updating this string to "You must fill out this field"

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+    selectOneOfOptions,
+    selectTheReason,
+    whatWasTheReason,
+    fillOutField,
+    yourResponsesHelp,
+    thankYouForYourFeedback,
+  };
+
+  /** @type {string} */
+  feedbackValue;
```
erocheleau, line 53: Why always `*Value` ?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+  /** @type {string} */
+  detailsValue;
+  /** @type {boolean} */
+  feedbackSubmitted = false;
```
erocheleau, line 57:
```suggestion
  isFeedbackSubmitted = false;
```
Just a suggestion but it makes it clearer that this is a boolean. Also I think the `= false;` is superfluous but I have no problem keeping it since it's clearer as well.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+  /** @type {boolean} */
+  detailsInputError = false;
```
erocheleau, line 59:
```suggestion
  hasInputDetailsError = false;
```
Same thing, just a suggestion.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+  /** @type {string} */
+  error;
+  /** @type {boolean} */
+  validated = false;
```
erocheleau, line 63:
```suggestion
  isValidated = false;
```

or `isValid` ?

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -484,4 +484,44 @@
...
+      <label>Expliquer pourquoi</label>
+      <name>quantic_ExplainWhy</name>
+    </customLabels>
+    <customLabels>
+      <label>Sélectionnez la raison</label>
+      <name>quantic_SelectTheReason</name>
+    </customLabels>
+    <customLabels>
+      <label>Quelle était la raison?</label>
```
acarpentier01, line 496: If you applied my earlier comment, then consider updating this string to "Veuillez fournir des détails" or "Veuillez fournir la raison"

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+    getBueno(this).then(() => {
+      this.validateTheOptionsProperty();
+      if (!this.error) {
+        this.validateOptions();
+      }
+      if (!this.error) {
+        this.validateTheSubmitFeedbackProperty();
+      }
+      this.validated = true;
+    });
```
erocheleau, line 82: Do we really need to use that? It is mostly used to validate that the options received are strings, etc.?
I don't think we really need to enforce this, do you think it will cause an issue later down the road?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  validateTheOptionsProperty() {
+    if (!Bueno.isArray(this.options)) {
+      console.error('The options provided are not in a valid array.');
+      this.setError();
+    } else if (!this.options.length) {
+      console.error('At least one option must be specified.');
+      this.setError();
+    }
```
erocheleau, line 92:
```suggestion
    if (Array.isArray(this.options) && this.options.length > 0) {
      return true;
    }
    console.error('The options provided must be a non-empty array.');
    this.setError();
```

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -484,4 +484,44 @@
...
+      <name>quantic_WhatWasTheReason</name>
+    </customLabels>
+    <customLabels>
+      <label>Veuillez remplir ce champ</label>
+      <name>quantic_FillOutField</name>
+    </customLabels>
+    <customLabels>
+      <label>Veuillez choisir l'une des options suivantes</label>
```
acarpentier01, line 504:
```suggestion
      <label>Veuillez choisir l'une des options précédentes </label>
```
Assuming the message is displayed under the options.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+        if (!Bueno.isString(value)) {
+          console.error(`The "${value}" value is not a valid string.`);
+          this.setError();
+        }
+        if (!Bueno.isString(label)) {
+          console.error(`The "${label}" label is not a valid string.`);
+          this.setError();
+        }
+      });
+    }
```
erocheleau, line 115: Yea I don't think it's necessary to validate all of that in this component.
Look at the lightning `datatable` component, it also receives options as part of a structured object.
I think we should just make sure that we are as forgiving as possible when stuff is missing. For example in the case of the data table, if you omit the label of a field, it just doesn't displays a label (but also doesn't just displays "undefined")

image

I feel that otherwise all this validation is adding a lot of code that will mostly just trip the user when they are making their component/code, but provide little to no value to the user.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+
+  validateTheSubmitFeedbackProperty() {
+    if (typeof this.submitFeedback !== 'function') {
+      console.error('The submitFeedback property is not a valid function.');
+      this.setError();
+    }
+  }
+
+  setError() {
+    this.error = 'c-quantic-feedback-modal Error';
```
erocheleau, line 126: What is this exactly?

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -917,4 +917,74 @@
...
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Please fill out this field</shortDescription>
+    </labels>
+    <labels>
+      <fullName>quantic_SelectOneOfOptions</fullName>
+      <value>Please select one of these options</value>
```
acarpentier01, line 950: Since all the options are highlighted in red, you could just say "Please select an option"

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+    }`;
+  }
+
+  /**
+   * Handles the option change.
+   * @param {CustomEvent} event
+   */
+  handleOptionChange(event) {
+    this.detailsInputError = false;
+    this.optionsInput.setCustomValidity();
```
erocheleau, line 160: Why do you need the `setCustomValidity` ?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+  handleDetailsChange(event) {
+    this.detailsInputError = false;
+    this.detailsValue = event.target.value;
+  }
+
+  /**
+   * Indicates whether the details input should be displayed.
+   * @returns {boolean}
+   */
+  get displayDetailsInput() {
```
erocheleau, line 178:
```suggestion
  get shouldDisplayDetailsInput() {
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Handles the option change.
+   * @param {CustomEvent} event
+   */
+  handleOptionChange(event) {
+    this.detailsInputError = false;
+    this.optionsInput.setCustomValidity();
+    this.optionsInput.reportValidity();
+    this.feedbackValue = event.detail.value;
```
erocheleau, line 162: Why don't you store here the `this.options` that match the `event.detail.value` ?

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -484,4 +484,44 @@
...
+    <customLabels>
+      <label>envoyer le commentaire</label>
+      <name>quantic_SendFeedback</name>
+    </customLabels>
+    <customLabels>
+      <label>Terminé</label>
+      <name>quantic_Done</name>
+    </customLabels>
+    <customLabels>
+      <label>Vos réponses permettent d'améliorer l'expérience de recherche.</label>
```
acarpentier01, line 520:
```suggestion
      <label>Vos commentaires permettent d'améliorer l'expérience de recherche.</label>
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  /**
+   * Indicates whether the details input should be displayed.
+   * @returns {boolean}
+   */
+  get displayDetailsInput() {
+    const selectedOption = this.options.find(
+      (option) => option.value === this.feedbackValue
+    );
```
erocheleau, line 181: Or I would make this a function

```
getSelectedOption() {
  return this.options.find(option => option.value === this.optionsInput.value);
}
```

Calling `.value` on the `lightning-radio-group` should give you the "value" of the selected radio button as per documentation, worth checking out if it's true :P

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+   * Indicates whether the feedback form is valid.
+   * @returns {boolean}
+   */
+  get feedbackFormIsValid() {
+    if(!this.feedbackValue){
+      return false;
+    }
+    const selectedOption = this.options.find(
+      (option) => option.value === this.feedbackValue
+    );
```
erocheleau, line 195: See you're calling the `get selected option` once more here, make it a function IMO

packages/quantic/jest.config.js
```diff
@@ -12,8 +12,12 @@ module.exports = {
...
   moduleNameMapper: {
-    "^@salesforce/i18n/": "<rootDir>/typings/lwc/customlabels.d.ts"
+    '^@salesforce/i18n/': '<rootDir>/typings/lwc/customlabels.d.ts',
+    '^lightning/modal$': '<rootDir>/force-app/test/jest-mocks/lightning/modal/modal',
+    '^lightning/modalBody$': '<rootDir>/force-app/test/jest-mocks/lightning/modalBody/modalBody',
+    '^lightning/modalFooter$': '<rootDir>/force-app/test/jest-mocks/lightning/modalFooter/modalFooter',
+    '^lightning/modalHeader$': '<rootDir>/force-app/test/jest-mocks/lightning/modalHeader/modalHeader',
   },
-  modulePathIgnorePatterns: ['.cache']
+  modulePathIgnorePatterns: ['.cache'],
```
erocheleau, line 21: Seems like the joint fixing the build issues has been useful ;)

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
```diff
@@ -0,0 +1,436 @@
...
+    element[key] = value;
+  }
+
+  document.body.appendChild(element);
+  return element;
+}
+
+// Helper function to wait until the microtask queue is empty.
+function flushPromises() {
+  // eslint-disable-next-line @lwc/lwc/no-async-operation
```
erocheleau, line 56: It would be nice to find the config line to put in the eslintrc to ignore that error in any `__tests__/*` file instead of having to have that line everywhere

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -917,4 +917,74 @@
...
+    <labels>
+      <fullName>quantic_SelectTheReason</fullName>
+      <value>Select the reason</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Select the reason</shortDescription>
+    </labels>
+    <labels>
+      <fullName>quantic_WhatWasTheReason</fullName>
+      <value>What was the reason?</value>
```
acarpentier01, line 936: Consider updating to "Please provide details." or "Please provide a reason."

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -917,4 +917,74 @@
...
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>What was the reason?</shortDescription>
```
acarpentier01, line 939: ditto

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
```diff
@@ -0,0 +1,436 @@
...
+        .fn()
+        .mockImplementation(
+          (value) => typeof value === 'number' && !isNaN(value)
+        ),
+      isArray: jest
+        .fn()
+        .mockImplementation(
+          (value) => Object.prototype.toString.call(value) === '[object Array]'
+        ),
+    };
```
erocheleau, line 104: You could remove all that if you remove the validation of the properties.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
```diff
@@ -0,0 +1,436 @@
...
+          selectors.errorMessage
+        );
+
+        expect(errorMessage).not.toBeNull();
+        expect(console.error).toHaveBeenCalledWith(
+          errors.invalidValueTypeError(invalidValue)
+        );
+      });
+    });
+  });
```
erocheleau, line 236: Does it really makes sense to validate all of that AND have to write tests for it?

erocheleau (changes requested):
I don't think we should spend so much on validating all the options format and everything like that... 

Should be a team decision however, what does everyone think?

acarpentier01 (approved):
Looks great! 👍

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/feedbackForm.css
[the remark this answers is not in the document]
nathanlb: agreed!

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
lbergeron: @mmitiche You are right, `LightningModal` does not seem to support slots. I didn't realize that, sorry.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
nathanlb: On one hand salesforce error messages often leave more to be desired, on the other I think you're right that it does add a considerable amount of code and I think any code that can be validated with tests should.
I'm all for improving the developer experience by adding informative error messages but only to prevent truly breaking options so there's a balance to strike. My understanding is that Atomic does a lot of prop validation and it would be unfortunate if Quantic didn't try to deliver a similar quality of dev experience.

Anyway in this particular case, the code and the tests are already written so if the validations are limited to ensuring the component doesn't crash I'm for it.

nathanlb (commented):
Cool component! Shame we had to copy over all of those mocks. I would recommend leaving a TODO somewhere if you think these are only temporary and linking the repo where they were taken from.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
mmitiche: sure yeah 👌

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
mmitiche: As you may have already noticed in my PR, I'm all for improving the developer experience by adding this type of validation.
The main reason is that this will help avoid uncertain/unknown behaviour with Headless down the road.
For example in this component, the value of a given option is passed  down the road to a function of one of the Headless controllers, if this function accepts only a parameter of type string and throws an error otherwise, it wouldn't be easy to recognize the source of the issue just by looking at the error logged by Headless. However if we handle this prop validation in Quantic the user can solve the problem in no time while developing his project.
That's the value I see in making this kind of validations and frankly, this seems useful to me.
What do you think @erocheleau ?

mmitiche:
> Looks great! 👍

@acarpentier01  Thank you so much for your input 👌

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
```diff
@@ -0,0 +1,234 @@
...
+   * The list of options to be displayed in the modal.
+   * @type {Array<{label: string, value: string, withDetails: boolean, detailsRequired: boolean}>}
+   */
+  @api options;
+  /**
+   * @api
+   * The function that will be executed when the feedback is submitted.
+   * @type {function}
+   */
+  @api submitFeedback;
```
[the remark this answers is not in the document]
mmitiche: I like the suggestion but unfortunately properties starting with "on" are reserved for event handlers so this name is considered as an invalid property name.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
mmitiche: just to distinguish between the values of the two inputs in the modal, the feedback value and the details value.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/feedbackForm.css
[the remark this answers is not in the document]
mmitiche: resize functionality re-added after confirming with UX.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/feedbackForm.html
mmitiche: I updated this to directly use the `lightning-textarea` component, which handles this a11y concerns for us.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
mmitiche: it's sets the error message to be displayed when one of the props is invalid.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
mmitiche: removed 👌

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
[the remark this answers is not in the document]
nathanlb: Ah i see. thanks. `handleSubmit` would be good too as submitFeedback sounds like a boolean to me but it's a nitpick so go with your gut.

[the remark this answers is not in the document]
mmitiche: sure yeah `handleSubmit` feels better👌

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/__tests__/quanticFeedbackModal.test.js
```diff
@@ -0,0 +1,436 @@
...
+};
+
+function createTestComponent(options = defaultOptions) {
+  const element = createElement('c-quantic-feedback-modal', {
+    is: QuanticFeedbackModal,
+  });
+
+  for (const [key, value] of Object.entries(options)) {
+    element[key] = value;
+  }
```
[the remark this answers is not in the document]
mmitiche: it would do the same thing, but personally I feel like the way we currently using looks like a more natural and readable way to affect props to the component.

packages/quantic/force-app/main/default/.eslintrc.json
```diff
@@ -7,5 +7,13 @@
...
     "CoveoHeadlessInsight": "readonly",
     "Bueno": "readonly"
-  }
+  },
+  "overrides": [
+    {
+      "files": ["**/__tests__/**"],
+      "rules": {
+        "@lwc/lwc/no-async-operation": "off"
+      }
```
erocheleau, line 16: nice!

packages/quantic/force-app/main/default/lwc/quanticFeedbackModal/quanticFeedbackModal.js
erocheleau: After offline discussion, later down the road we might move all this validation to a single centralized files, but for now it'll remain in this file.

erocheleau (approved):

lvu285:
In the design we have this comment from UX, so we won't make this, because it's complicated?
```Once the feedback is given. It shouldn’t be possible to change it or give feedback again. 
It would imply removing «Thank your for your feedback» 

*Unless it’s a lot more work.  I don’t have a strong opinion on that matter. We can discuss it. ```

mmitiche:
> In the design we have this comment from UX, so we won't make this, because it's complicated?
> 
> ```
> Once the feedback is given. It shouldn’t be possible to change it or give feedback again. 
> It would imply removing «Thank your for your feedback» 
> 
> *Unless it’s a lot more work.  I don’t have a strong opinion on that matter. We can discuss it. ```
> ```

@lvu285 I'm still confirming with UX how exactly we wanna hide the `Thank your for your feedback` message, once this is done I'll create a new ticket for supporting this feature in our component.

PR #2723, opened 2023-02-10
fix(atomic): ignore empty query when detecting duplicates in search box suggestions

olamothe:
You can check the JIRA for a link to the discuss that sparked this change.

https://coveord.atlassian.net/browse/KIT-2275

### DIFF
--- a/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
+++ b/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
@@ -1,9 +1,7 @@
-import {VNode} from '@stencil/core';
 import {
   dispatchSearchBoxSuggestionsEvent,
   SearchBoxSuggestionElement,
 } from '../../../src/components/search/search-box-suggestions/suggestions-common';
-import {buildCustomEvent} from '../../../src/utils/event-utils';
 import {
   SafeStorage,
   StorageItems,
--- a/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
+++ b/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
@@ -1,3 +1,9 @@
+import {VNode} from '@stencil/core';
+import {
+  dispatchSearchBoxSuggestionsEvent,
+  SearchBoxSuggestionElement,
+} from '../../../src/components/search/search-box-suggestions/suggestions-common';
+import {buildCustomEvent} from '../../../src/utils/event-utils';
 import {
   SafeStorage,
   StorageItems,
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -33,7 +34,7 @@ import {
 } from '../../../utils/local-storage-utils';
 import {promiseTimeout} from '../../../utils/promise-utils';
 import {updateBreakpoints} from '../../../utils/replace-breakpoint';
-import {getUniqueItemsByProperties, once, randomID} from '../../../utils/utils';
```
ThibodeauJF, line 36: Might as well remove or update that method, it doesn't seem used anywhere else

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -516,7 +517,15 @@ export class AtomicSearchBox {
 
   private getAndFilterLeftSuggestionElements() {
     const suggestionElements = this.getSuggestionElements(this.leftSuggestions);
-    return getUniqueItemsByProperties(suggestionElements, ['query']);
+    // We want to remove duplicates elements for the same `query` property, but only when it's not empty.
+    // Since the custom search box suggestion system does not enforce the query property (it's optional),
+    // it's a system that custom search box suggestions component will use to provide a "title" or separator section.
+    // We don't want to detect multiple separator section with no query as duplicate
```
ThibodeauJF, line 523: Maybe an e2e test that covers the case would be better than the comment (which can be detailed in the test description)

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -516,7 +517,15 @@ export class AtomicSearchBox {
...
+    return suggestionElements.filter(
+      (suggestionElement, i, self) =>
+        isNullOrUndefined(suggestionElement.query) ||
+        i === self.findIndex((other) => other.query === suggestionElement.query)
+    );
```
louis-bompart, line 528: nipic, but your function is O(n^2): you're going thru the array twice.
But you could go thru it only once.
I think also that this implementation, albeit more verbose is also clearer and would avoid the comment.


```suggestion
    return suggestionElements.filter((suggestionElement) => {
      if (isNullOrUndefined(suggestionElement.query)) {
        return true;
      }
      if (queryAlreadyFound.has(suggestionElement.query)) {
        return false;
      } else {
        queryAlreadyFound.add(suggestionElement.query);
        return true;
      }
    });
```

louis-bompart (commented):
Might be good indeed to have an E2E, the filter function was pretty unclear too IMO

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
olamothe: Yeah, I thought about adding a test. The problem here though is that it requires creating a "custom component" if you wish, or at least mimic the custom search box provider system. 

We cannot test this with standard atomic component.

I can try to add something that would not needlessly complexify the test setup...

packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
```diff
@@ -146,6 +171,110 @@ describe('Search Box Test Suites', () => {
         );
       });
 
+      describe('with custom suggestions provider', () => {
```
olamothe, line 174: Most of the test setup complexity is because here we are not testing an atomic component per say, but some arbitrary code that register/hooks into the custom search box suggestion system.

louis-bompart (approved):

Issue #189, opened 2020-10-16
[BB pr#189] ci(headless): fix build issues


coveobot:
>  **Pull request** :twisted_rightwards_arrows: created by **@samisayegh** on 2020-10-16 18:10
> Last updated on 2020-10-19 13:17
> Original Bitbucket pull request id: 189
>
> Participants:
>
> * **@acote-coveo** (reviewer)
> * **@btaillon** (reviewer)
> * **@ThibodeauJF** (reviewer)
> * bitbucket user **coveo-anti-throttling_02**
> * **@jlblais** (reviewer)
> * **@olamothe** (reviewer) :heavy_check_mark:
> * **@samisayegh**
> * **@nrawji** (reviewer) :heavy_check_mark:
>
> Source: Commit `bd8898784520` on branch `KIT-192`> Destination: https://bitbucket.org/coveord/ui-kit/commits/a8bce3f4e0d8 on branch master
> Merge commit: https://github.com/None/commit/c8e02fb313b0
>
> State: **`MERGED`**

https://coveord.atlassian.net/browse/KIT-192

PR #2724, opened 2023-02-13
ci: added nx and refactored scripts

btaillon:
https://coveord.atlassian.net/browse/KIT-2277

image

# Changes in this PR
## Add Nx
In this PR, my primary goal is to introduce [Nx](https://nx.dev/) to ensure that tasks like "build", when run on the whole repo (e.g., by the CI), will build every package in the correct order.

Prior to this PR, we were using [Lerna](https://lerna.js.org/) for this purpose, but we'd like to phase out Lerna from the repo since it modifies the `package-lock` and we'd like to eventually introduce features like caching. You can find discussions  about task execution contained in [our monorepo tooling discussion](https://docs.google.com/document/d/1agjDUxyUlZYuWN-SMIuhuPohxeT9p1SUMwkYjn0udvk). I think it's safe to move forward with Nx.

## Add default target dependencies
Nx allows us to define what task (Nx calls them targets) has to be run before one another. We can define default target dependencies (e.g., on each project, the "build" target depends on the "build" targets of their dependencies). We can also define specific target dependencies for a specific target.

For instance, in this PR, I defined that the `build` target, by default, depends on parent `build` targets. If we ask Nx to build Quantic, this will cause it to build Bueno then Headless then Quantic.

## Refactor scripts
Scripts are inconsistent across our different packages. For instance, we interchangeably use the names `start`/`dev`/`serve` to mean the same thing, and it's unclear whether these tasks require us to build first.

In this PR, I added clear definitions of what tasks should do in each projects and I adapted each project to meet those definitions. This way, I'll be able to make CI changes (like adding default target dependencies) without worrying as much about the specifics of each package.

Almost the entire PR is compromised of this change. I could've divided the PR to isolate this change, but it wouldn't have been very useful since the other two changes are so small and help visualize this one.

**First thing when reviewing this PR**: I recommend looking at `README.md`'s changes, then `nx.json`, then the `package.json` files. Everything else is a consequence of those.


README.md
```diff
@@ -54,6 +54,113 @@ The following Visual Studio Code extensions are recommended:
 - [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)
 - [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)
 
+## Scripts and tasks
+
+Scripts for each `package.json` should follow these rules:
+
+### Nesting
+
+Some scripts may be nested under other scripts using the colon (`:`) symbol. This can interchangeably be used to mean either:
```
btaillon, line 63: A flaw with this approach is that Nx's CLI uses the colon symbol to divide the project name, target name and configuration name (see [Use Executor Configurations](https://nx.dev/recipes/executors/use-executor-configurations), which uses `nx run mylib:build:production` as an example).

The only impact I noticed from this is that, when running a script whose name contains a colon using Nx, we can't use the short syntax, we have to use `--target=`.

btaillon: I don't think the above problem is necessarily bad enough on its own to justify changing the symbol. However, if we were to replace scripts with [`nx:run-command` executors](https://nx.dev/packages/nx/executors/run-commands#nx:run-commands) and [executor configurations](https://nx.dev/recipes/executors/use-executor-configurations), this problem would be inherently solved and we could keep only the relevant scripts in `package.json` files.

README.md
```diff
@@ -54,6 +54,113 @@ The following Visual Studio Code extensions are recommended:
...
+  - E.g.:
+    ```json
+    {
+      "e2e": "cypress run --browser chrome",
+      "e2e:firefox": "cypress run --browser firefox",
+      "e2e:watch": "cypress open"
+    }
+    ```
+
+### `build` script
```
btaillon, line 84: In this PR, I also prefixed all scripts which generate artifacts with `build:`.

README.md
```diff
@@ -54,6 +54,113 @@ The following Visual Studio Code extensions are recommended:
...
+
+- Generates all the files needed for a dependent project to run.
+- Doesn't expect another script from the same project to be run before this script.
+- If the build fails, exits with an error.
+
+### `dev` script
```
btaillon, line 90: The choice to use `dev` instead of `start` or `serve` is very debatable, I'd be curious what your opinions are.

I chose `dev` because it's unambiguous.
  * `start` could mean "start the current (production) build".
  * `serve` could mean "serve the current (production) build".

I think `start` would also be a great choice since it's one of NPM's defaults (we can run it with `npm start` instead of `npm run start`). However, it's a little bit more ambiguous.

README.md
```diff
@@ -54,6 +54,113 @@ The following Visual Studio Code extensions are recommended:
...
+
+- Generates all the files needed for a dependent project to run.
+- Doesn't expect another script from the same project to be run before this script.
+- Automatically re-runs when there's a code change in the project or its dependencies.
+- Doesn't exit until the user chooses to.
+- May serve the project with a URL.
+
+### `prod` script
```
btaillon, line 98: I'm not sure what name to use for this script 🤔 

I used `prod` because it contrasts with `dev`, but `prod` sounds like it could be a script meant to deploy to prod.

README.md
```diff
@@ -54,6 +54,113 @@ The following Visual Studio Code extensions are recommended:
...
+- If any test fails, exits with an error.
+
+### `test:watch` script
+
+Does the same thing as the `test` script, except:
+
+- Re-runs tests when there's a code change in the project.
+- Doesn't exit until the user chooses to.
+
+### `e2e` script
```
btaillon, line 119: We use Cypress for all e2e tests, but I prefer using a tooling-agnostic name.

nx.json
```diff
@@ -0,0 +1,55 @@
...
+        "cacheableOperations": []
+      }
+    }
+  },
+  "targetDefaults": {
+    "build": {
+      "dependsOn": [
+        "^build"
+      ]
+    },
```
btaillon, line 15: Dependencies are only enforced when running scripts using Nx. Doing `npm run build -w @coveo/quantic` behaves the same now as it did before.

If you want to build all projects, you would use `npx nx run-many --target=build`. If you want to build one project and all its parents, you would use `npx nx run --project=quantic --target=build`.

package.json
```diff
@@ -2,27 +2,24 @@
...
-    "postinstall": "husky install & patch-package && npm run link:angular -w @coveo/atomic-angular-samples",
-    "start": "lerna run --parallel start",
-    "dev": "npm run start -- --scope @coveo/headless --scope @coveo/atomic",
-    "build": "lerna run clean && lerna run build",
-    "test": "lerna run --parallel test",
-    "test:watch": "lerna run --parallel test:watch --no-prefix",
-    "lint:check": "eslint .",
-    "lint:fix": "eslint --fix .",
+    "postinstall": "husky install && patch-package && npm run link:angular -w @coveo/atomic-angular-samples",
+    "dev:atomic": "concurrently \"npm run -w @coveo/headless dev\" \"npm run -w @coveo/atomic dev\"",
```
btaillon, line 6: Defining a dependency in Nx between dev scripts and parent dev scripts doesn't work currently, since it runs the parent's dev script (Headless in this case) and then hangs instead of starting the second one. This will have to be handled in a case-by-case basis.

Until then, I'm using `concurrently` to run both of these scripts in parallel.

package.json
```diff
@@ -2,27 +2,24 @@
...
+    "build": "nx run-many --target=build",
+    "test": "nx run-many --target=test",
+    "e2e": "nx run-many --target=e2e",
+    "lint:check": "npx eslint .",
+    "lint:fix": "npx eslint --fix .",
```
btaillon, line 11: Ideally, those would be run at a package-level, but this is too big of a change for this PR, I'll do it separately.

packages/atomic/README.md
```diff
@@ -40,57 +40,29 @@ Ref: https://docs.cypress.io/
...
 
 To test the current Atomic build against the hosted search pages for Trials, use the following commands:
 
 ```sh
-npm run cypress-hsp:open
-npm run cypress-hsp:test
-npm run cypress-hsp:test:firefox
-```
-
-### Visual regression tests
```
btaillon, line 68: We removed applitools

packages/atomic/package.json
```diff
@@ -25,29 +25,31 @@
   ],
   "scripts": {
     "clean": "rimraf -rf dist/*",
-    "build": "npm run build:locales && npm run build:stencil && npm run doc:list-assets && npm run build:storybook && npm run validate:definitions",
```
btaillon, line 28: `doc:list-assets` is only used by Storybook, so I moved it into `build:storybook`.

packages/quantic/package.json
```diff
@@ -11,28 +11,32 @@
...
+    "dev:sfdx": "sfdx force:source:deploy -p force-app/main && sfdx force:lightning:lwc:start --port 3334",
     "test:unit": "../../node_modules/.bin/lwc-jest",
     "test:unit:debug": "lwc-jest --debug",
     "test:unit:watch": "lwc-jest --watch",
     "test:unit:coverage": "lwc-jest --coverage",
-    "test:typings": "tsc",
-    "cypress:open": "npx cypress open",
-    "cypress:test": "npx cypress run --browser chrome",
-    "cypress:test:detailed": "npx cypress run --browser chrome --headless --reporter cypress/reporters/detailed-reporter.js",
+    "validate:types": "tsc",
```
btaillon, line 26: We also have a `validate` task for Atomic. I could add `validate` to the list of scripts in the README if you prefer.

packages/quantic/package.json
```diff
@@ -11,28 +11,32 @@
     "url": "git+https://github.com/coveo/ui-kit"
   },
   "scripts": {
-    "test": "npm run test:lint && npm run test:typings && npm run test:unit",
-    "test:lint": "eslint force-app/main/default/lwc/ --format junit -o reports/eslint.xml",
+    "test": "npm run build:staticresources && npm run lint:check:tests && npm run validate:types && npm run test:unit",
```
btaillon, line 14: `build:staticresources` is necessary for running tests and it runs fairly quickly, so I added it here and removed it from the CI. This makes it consistent with the rule I set, which is that `test` should work without doing anything first.

Eventually, with Nx, we can make `build:staticresources` a dependency of `test`, and it will only re-run it when necessary.

packages/quantic/package.json
```diff
@@ -11,28 +11,32 @@
...
     "test:unit:watch": "lwc-jest --watch",
     "test:unit:coverage": "lwc-jest --coverage",
-    "test:typings": "tsc",
-    "cypress:open": "npx cypress open",
-    "cypress:test": "npx cypress run --browser chrome",
-    "cypress:test:detailed": "npx cypress run --browser chrome --headless --reporter cypress/reporters/detailed-reporter.js",
+    "validate:types": "tsc",
+    "e2e:watch": "npx cypress open",
+    "e2e": "npx cypress run --browser chrome",
+    "e2e:detailed": "npx cypress run --browser chrome --headless --reporter cypress/reporters/detailed-reporter.js",
```
btaillon, line 29: maybe we should always run in detailed mode?

packages/quantic/package.json
```diff
@@ -11,28 +11,32 @@
...
     "setup:examples": "../../node_modules/.bin/ts-node scripts/build/deploy-community.ts",
     "deploy:main": "../../node_modules/.bin/sfdx force:source:deploy -u LWC -p force-app/main",
     "deploy:examples": "../../node_modules/.bin/sfdx force:source:deploy -u LWC -p force-app/examples",
-    "doc:generate": "mkdir -p docs/out; jsdoc -c jsdoc-config.json > docs/out/quantic-docs.json",
-    "package:create": "../../node_modules/.bin/ts-node scripts/build/create-package.ts --remove-translations",
-    "package:create:publish": "../../node_modules/.bin/ts-node scripts/build/create-package.ts --remove-translations --promote",
-    "npm:publish": "node ../../scripts/deploy/publish.mjs",
+    "build:doc": "mkdir -p docs/out; jsdoc -c jsdoc-config.json > docs/out/quantic-docs.json",
+    "publish:sfdx": "../../node_modules/.bin/ts-node scripts/build/create-package.ts --remove-translations",
+    "promote:sfdx": "npm run publish:sfdx -- --promote",
```
btaillon, line 35: I wasn't exactly sure what `package:create` and `package:create:publish` did. Do these names make sense?

packages/samples/angular/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "dev": "npm run copy:assets && ng serve",
+    "dev": "npm run build:assets && dev:angular",
```
btaillon, line 6: The angular sample doesn't have a `build` script, therefore it doesn't run when the CI does `npm run build`. Should I add it?

packages/samples/angular/package.json
```diff
@@ -3,10 +3,10 @@
...
+    "dev:angular": "ng serve",
     "link:angular": "node scripts/link-hoisted-dependencies.mjs",
-    "watch": "ng build --watch --configuration development",
```
btaillon, line 8: `watch` seems redundant since its purpose is pretty much covered by `dev`, so I removed it.

scripts/deploy/update-npm-tag.mjs
```diff
@@ -3,16 +3,10 @@ import {promisify} from 'node:util';
...
 } from '../packages.mjs';
 
-async function main() {
-  const requests = packageDirsNpmTag
-    .map((dir) => getPackageDefinitionFromPackageDir(dir))
-    .map(({name, version}) => updateNpmTag(name, version));
-
-  await Promise.all(requests);
-}
+const pkg = getPackageManifestFromPackagePath(process.cwd());
```
btaillon, line 9: Prior to this PR, `update-npm-tag.mjs` was executed at the root of the repo. Now, it gets executed once for every package.

.github/actions/setup/action.yml
```diff
@@ -40,6 +40,3 @@ runs:
     - if: steps.build-cache.outputs.cache-hit != 'true'
       run: npm run build
       shell: bash
-    - if: steps.build-cache.outputs.cache-hit != 'true'
-      run: npm run doc:generate
```
btaillon, line 44: Moved into `build`

.github/workflows/manual-package.yml
```diff
@@ -59,7 +59,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
-      - run: npm run copy:staticresources -w @coveo/quantic
```
btaillon, line 62: Moved into `test`

Jenkinsfile
```diff
@@ -33,25 +33,14 @@ node('heavy && linux && docker') {
...
 
-      stage('Generate docs') {
-        sh 'npm run doc:generate'
-      }
-    }
-
-    stage('Clean working directory') {
-      sh 'git checkout -- .'
-      sh 'git clean -f'
-    }
```
btaillon, line 44: This was originally added because `npm run build` would occasionally update the README (see https://github.com/coveo/ui-kit/issues/189#issuecomment-733037682). It shouldn't be necessary anymore.

If it turns out that this is necessary, then a cleanup step should be added to `npm:publish` IMO.

README.md
btaillon: I found another problem with using the `:` symbol. When I run targets from vscode using the extension, I get errors.

package.json
louis-bompart: You cant use --target=targetA,targetB ?

louis-bompart (approved):
LGTM overall, great job!

package.json
btaillon: Maybe, but I know nx has a limit of parallel targets because they're meant to be processes that end eventually. We can change that limit (as far as I'm aware), but I feel like I'd be working against Nx's principles.

Until I improve the `dev` scripts so that they are "resolved" once the page is served, I prefer to keep it like this. It'll be replaced eventually anyway.

louis-bompart: Oh yeah, it's watch scripts, nevermind carry on.

ThibodeauJF:
👍  Did you pull in master with my new lil package, `atomic-hosted-page`? 🥺

.github/actions/e2e-stencil/action.yml
```diff
@@ -12,7 +12,7 @@ runs:
       name: Run Cypress
       with:
         working-directory: ./packages/samples/stencil
-        start: npm start
+        start: npm run dev
```
ThibodeauJF, line 15: Run dev is typically at the top to run only atomic along w/ headless, perhaps you're not using it, but we should keep it consistent with before?

README.md
ThibodeauJF: https://github.com/coveo/ui-kit/pull/2724/files#r1106304160 related

README.md
ThibodeauJF: Like dev, switching the script names might confuse the repo users, some may have local setups running commands for tests or that kind of thing.

ThibodeauJF: If I'm alone thinking this, ignore these comments 😄

btaillon:
> 👍 Did you pull in master with my new lil package, `atomic-hosted-page`? 🥺

Haven't yet, will do before merging.

.github/actions/e2e-stencil/action.yml
btaillon: I removed `dev` at the root `package.json`. Trying to run `npm run dev` should give an error, which should prompt people to look at the `package.json` and find `dev:atomic`. I think it's fine.

ThibodeauJF: Alright! I do like the coherence

olamothe (approved):

README.md
```diff
@@ -54,6 +54,113 @@ The following Visual Studio Code extensions are recommended:
 - [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint)
 - [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)
 
+## Scripts and tasks
```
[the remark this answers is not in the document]
btaillon: I agree! I think we should move any "internal" instructions to a `CONTRIBUTING.md` file. I'll do that.

packages/quantic/package.json
btaillon: For some reason, I thought the CI was using `e2e:detailed` but I was wrong.

Good point about re-running it locally for details, I'll keep `e2e:detailed` and `e2e` separated then.

packages/quantic/package.json
nathanlb: meh that's ok

packages/quantic/package.json
nathanlb: yep that's pretty good

nathanlb (approved):
TBH I didn't review the entire thing but the bit related to Quantic look good.
Overall this seems like a smart way of doing things. 🧠 👍 
Looks like there was lots of work involved GG and thanks

ThibodeauJF (approved):

PR #2727, opened 2023-02-14
feat(quantic): add the feedback feature to the quantic smart snippet

mmitiche:
[SFINT-4810](https://coveord.atlassian.net/browse/SFINT-4810)

- Added the feedback feature to the Quantic Smart Snippet component.
- To achieve this, I used  two quantic components: the Quantic Feedback component and the Quantic Feedback Modal component.
- The user can now like or dislike a smart snippet.
- If the user dislikes the smart snippet, the `explain why` button appears.
- if the user clicks the `explain why` button, the Quantic Feedback Modal is opened displaying a form where the user can explain why they didn't like the smart snippet.
- we log UA events to track all these feedback actions.

#### Demo:

https://user-images.githubusercontent.com/86681870/219030783-bd47a8d2-84c1-4e37-a9a4-baed6c8fc89a.mov

[SFINT-4810]: https://coveord.atlassian.net/browse/SFINT-4810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
LGTM!

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet-expectations.ts
```diff
@@ -5,13 +5,22 @@ import {
...
+}
+
 function smartSnippetExpectations(selector: SmartSnippetSelector) {
   return {
     displaySmartSnippetCard: (display: boolean) => {
       selector
         .smartSnippetCard()
         .should(display ? 'exist' : 'not.exist')
-        .log('should display the smart snippet');
+        .log(`${should} display the smart snippet`);
```
lbergeron, line 23: `should` is a function, and `logDetail` should be used instead of `log`. The other expectations should be fixed the same way.

```suggestion
        .logDetail(`${should(display)} display the smart snippet`);
```

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet.cypress.ts
```diff
@@ -141,5 +148,56 @@ describe('quantic-smart-snippet', () => {
...
+        scope('when closing the feedback modal', () => {
+          Actions.clickFeedbackCancelButton();
+          Expect.logCloseSmartSnippetFeedbackModal();
+        });
+
+        scope('when selecting a feedback option', () => {
+          const exampleDetails = 'example details';
+          Actions.clickSmartSnippetExplainWhyButton();
+          Expect.logOpenSmartSnippetFeedbackModal();
+          Actions.clickFeedbackOption(feedbackOptions.length - 1);
```
lbergeron, line 190: Since the `other` option is special, we could define a constant to make it clearer in the test.

```
const otherOption = 'other';
const feedbackOptions = [..., otherOption];
```

```suggestion
          Actions.clickFeedbackOption(otherOption);
```

packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js
```diff
@@ -160,6 +176,89 @@ export default class QuanticSmartSnippet extends LightningElement {
...
+    if (feedback?.details) {
+      this.smartSnippet.sendDetailedFeedback(feedback.details);
+    } else if (feedback?.value) {
+      this.smartSnippet.sendFeedback(feedback.value);
+    }
+    this.feedbackSubmitted = true;
+    this.hideExplainWhyFeedbackButton = true;
+  }
+
+  get options() {
```
lbergeron, line 244: I'm wondering whether setting options this way is too rigid for customers.

Suppose a customer wants different options, then he has to duplicate the `quanticSmartSnippet` component just to override the options.

Suppose a customer wants the detail textarea for more than one option, then he has to duplicate and modify the `quanticFeebackModal`, `quanticFeedback`, and `quanticSmartSnippet` components.

I think making this component more modular could go a long way for customers implementing Smart Snippets.

lbergeron (commented):
It looks very nice! However, I'm worried about the lack of component modularity/customization options. If a customer needs a slightly different behavior, the only option is to duplicate component(s) to make custom one(s). Following this path makes it very difficult for a customer to keep up with Quantic changes over time.

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet-expectations.ts
mmitiche: Thanks I'll fix them all!

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet.cypress.ts
mmitiche: Sure thing I'll simplify this👌

packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js
mmitiche: The reason why I'm setting the options this way is because in Headless only these predefined options are allowed in the analytics sent when submitting a feedback.
For instance the `sendFeedback` method of the Headless Smart Snippet controller can only be called with a parameter of type `SmartSnippetFeedback ` which is a type defined in Headless as: 
```type SmartSnippetFeedback = "does_not_answer" | "partially_answers" | "was_not_a_question"```

For this reason I didn't consider adding the ability to accept `options` as a public property for example.
If a customer wants different options, then he has to duplicate the quanticSmartSnippet component NOT just to override the options but also to handle how the feedback will be submitted since in that particular case the methods that come with the Headless Smart Snippet controller will be no help for this customer.

Basically  we can't offer a lot of freedom here just because we currently don't have the same freedom in the Headless Smart Snippet controller.

Comment (mmitiche) [reply packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js:244]: I hear you and this makes perfect sense to me as well!
I think this would need to be in a separate PR just because this addition would require a change in Headless as well: 
we will need to change the signature of the function `sendFeedback` from:
```
function sendFeedback(feedback: SmartSnippetFeedback)
```
to: 
```
function sendFeedback(feedback: SmartSnippetFeedback, details?: string)
```
After doing this change in Headless, we will be able to offer more freedom to our customers in Quantic when it comes to giving smart snippet feedbacks.

Does this sound right to you?

Comment (lbergeron) [reply packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js:244]: It seems like a good plan to me

**Review by lbergeron — APPROVED** [review]

File: packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml [hunk 4]
```diff
@@ -945,6 +945,41 @@
       <protected>false</protected>
       <shortDescription>Show less</shortDescription>
     </labels>
+    <labels>
```
mmitiche, line 948: Hey @acarpentier01!
Can you help me review these labels before merging this PR?

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -500,6 +500,26 @@
       <label>Voir moins</label>
       <name>quantic_SmartSnippetShowLess</name>
     </customLabels>
+    <customLabels>
```
mmitiche, line 503: @acarpentier01, here are the french translations as well.
Thank you!

acarpentier01 (approved):
Looks great! 👍

mmitiche:
> Looks great! 👍

Thank you so much @acarpentier01 👌

PR #2732, opened 2023-02-16
chore: make `npm run build` works on windows

louis-bompart:
Headless: relied a bit too much on how linux run stuff. Replaced by `execa` which is more reliable (avg 68M/downloads)
Bonus, added some prefixing on the logs for it to not be a blorb of incomprenhesible stuff

Quantic:
- Relied on mkdir from unix. `-p` aint a thing on windows (without powershell). Replaced by mkdirp for os-agnostic build.
- Piping is likely to break on Windows, so I used native JSDOC options (and fixed what I'd describe as a dumb limitation of the template). I did compare the output of the console and in files, it match.


louis-bompart:
For Headless, essentially here's what happen when you install a package with `bin` such as `mkdirp` or the api-extractor of MSFT:
 - On Linux, it become available on the path if the user is in the project
 - On Windows it is not available unless you run it through a npm script (npm does special sauce), or if it is globally installed.

packages/headless/scripts/extract-documentation.mjs
```diff
@@ -10,9 +10,13 @@ async function main() {
     .map((file) => `${dir}/${file}`);
   console.log('extracting documentation using', paths);
 
-  const jobs = paths.map((path) =>
-    execute('api-extractor', ['run', '-l', '-c', path])
-  );
+  const jobs = paths.map((path) => {
+    const apiProcess = execa('api-extractor', ['run', '-l', '-c', path], {});
```
btaillon, line 14: `execa` seems to work a lot like `process.spawn`, where it returns you the streams instead of a combined/final output.

I added `exec.mjs` to make it easy to execute a command, pipe the output to the console for debugging, and return the combined output (or throw). If this isn't too time-sensitive, I think we should try to replace `exec.mjs`'s `process.spawn` with execa.

btaillon (approved):

packages/headless/scripts/extract-documentation.mjs
louis-bompart: Agree that'd be nice eventually. For now, because there's a lot of change happening, I wanted to have the smallest footprint possible.

ThibodeauJF (approved):

packages/quantic/package.json
```diff
@@ -30,7 +30,8 @@
     "setup:examples": "../../node_modules/.bin/ts-node scripts/build/deploy-community.ts",
     "deploy:main": "../../node_modules/.bin/sfdx force:source:deploy -u LWC -p force-app/main",
     "deploy:examples": "../../node_modules/.bin/sfdx force:source:deploy -u LWC -p force-app/examples",
-    "build:doc": "mkdir -p docs/out; jsdoc -c jsdoc-config.json > docs/out/quantic-docs.json",
+    "prebuild:doc": "mkdirp docs/out",
```
nathanlb, line 33: is this script a monorepo thing that will be used automatically or it's just a utility for local setup?

nathanlb (approved):
🪟

packages/quantic/package.json
louis-bompart: https://www.npmjs.com/package/mkdirp

PR #2734, opened 2023-02-16
ci(bueno,headless,atomic): add caching

btaillon:
https://coveord.atlassian.net/browse/KIT-2282

In this PR, I added Nx caching to `bueno`, `headless`, and `atomic` by defining their tasks in their respective `project.json` files.

Here are the outputs of `npm run build` at the root of a clean repo.
<table>
<tr>
	<th>First run (6 mins)
	<th>Second run (4 mins)
	<th>Third run (1 min)
<tr>
	<td>

image
	<td>

image
	<td>

image
</table>

As you can see in the above table, some tasks are failing to cache properly on the first run, and only use the cache on the third run. Specifically:
* `headless:cached:build:doc`
* `atomic:cached:build:stencil`
* `atomic:cached:build:storybook`

I've been investigating these for hours and wasn't able to make their cache work. Using more specific executors may fix this, so I'll try it out locally. If you see anything that you suspect could cause this, please do comment about it, even if it's a far stretch.

Even with the above errors, a significant amount of time is saved.

# How targets work
In order to understand this PR, you need to understand how Nx's targets work. I will write it here since Nx's documentation is a bit complicated.

Targets are like `package.json` scripts, except they are defined in `project.json` and are more powerful. Here's what a target may look like:
```json
"cached:build:doc:extract": {
  "dependsOn": ["cached:build:definitions"],
  "inputs": [
    "{projectRoot}/scripts/extract-documentation.mjs",
    "{projectRoot}/config/api-extractor/**",
    "{projectRoot}/dist/definitions/**"
  ],
  "outputs": ["{projectRoot}/temp"],
  "executor": "nx:run-commands",
  "options": {
    "command": "node ./scripts/extract-documentation.mjs",
    "cwd": "packages/headless"
  }
}
```

Targets can depend on other targets using the `dependsOn` property. Targets can be cached by using the `inputs` and `outputs` property. Targets will run an `executor` with the `options` we give it (e.g., `nx:runs-commands` simply runs command(s)).

My understanding is that Nx runs targets like so:
Non-cached targets:
1. If the target has dependencies (`dependsOn`), **always** run them first.
2. Run the target's executor.

Cached targets:
1. If the target has dependencies (`dependsOn`), **always** run them first, regardless of whether `inputs` changed or not.
2. Calculate a single hash for all the files matched by the `inputs` globs.
  * If there's no `inputs` glob, inherit the inputs defines in `{workspaceRoot}/nx.json`, or inherit the `default` named input (which defaults to `{projectRoot}/**/*`).
3. If there's an existing cache matching the calculated hash, apply it.
  * Warning: applying the cache will not only add new directories or files, it may delete some.
4. If there's no existing cache matching the calculated hash, run the target then add all files or directories from `outputs` to the cache.

# Approaches
In this PR, I made various small choices, which I will explain here.

## Small targets
I decided to make every target as small as possible. Instead of having one big `build` target, I created several smaller targets. 

To make refactoring easier, I also tried to make every target as self-contained as possible, strictly defining their own dependencies, inputs and outputs. Consequently, some target may redundantly have the same dependency as one of their deeper dependencies, but I think it's easier to read this way, and Nx is smart enough not to run the dependency multiple times.

However, I don't see examples of this in Nx's documentation; they seem to generally make a single big `build` and `test` target.

Here are the pros of having smaller tasks:
* Easier cache debugging
  * Instead of the cache failing for the entire `build` target, it fails for a specific sub-target, which narrows down where the issue can be.
* Clear ownership of files
  * We can easily identify what command is responsible for generating a file, and what command requires what file.
* More parallelism
  * For instance, when running `npm run build` at the root of the repo, `atomic:cached:build:locales` is run in parallel with `bueno:cached:build:esbuild`, which is a more efficient use of computing resources.
* Easily convertible from/to executors
  * What Nx's executors do is generally equivalent to a single of our sub-targets. With this approach, we can easily convert from/to executors if we need/want to.

## Independent dev targets
For now, I made `dev` targets (and `:watch` targets) not have any dependencies.

If dev targets depend on parent dev targets, then only their parent will be run because the parent `dev` target technically never "finishes". I want to fix this in a separate PR by using executors.

If dev targets depend on parent build targets, then we can't run multiple of them in parallel, which is the workflow defined in the root `package.json`'s `dev:atomic` script.

## Tests caching
I chose to cache unit tests since their results should always be the same for a given input. This makes it easy to, for instance, lazily run `nx run-many --target=test` and have it test only affected projects. This will be mostly useful if/when we make the CI share its cache.

I chose not to cache e2e and integration tests, since both can be flaky and give different outcomes when re-run.

## Targets-only
I decided to extract all scripts out of `package.json` into targets (most often using the [`nx:run-commands`](https://nx.dev/packages/nx/executors/run-commands#nx:run-commands) executor) so that:
* only scripts which are relevant for us to run during development are available in `package.json` files.
* all of a project's tasks logic resides in one place (the `project.json` file).

If we *really* need to run sub-tasks, we can still do that by using `nx run` (e.g., `nx run atomic:\"some:task\"`).

## Executor choices
There are several executors available we can use. For instance:
* [`@nrwl/jest:jest`](https://nx.dev/packages/jest/executors/jest)
  * Runs a project's tests using Jest.
* [`@nrwl/esbuild:esbuild`](https://nx.dev/packages/esbuild/executors/esbuild)
  * Builds a project using ESBuild.
* [`@nrwl/js:tsc`](https://nx.dev/packages/js/executors/tsc)
  * Builds a project using Typescript's TSC.
* [`@nrwl/linter:eslint`](https://nx.dev/packages/linter/executors/eslint)
  * Lints a project (optionally fixing).
* [`@nrwl/storybook:build`](https://nx.dev/packages/storybook/executors/build)
  * Builds storybook.
* [`@nrwl/storybook:storybook`](https://nx.dev/packages/storybook/executors/build)
  * Serves storybook in dev mode.
* [`@nrwl/web:dev-server`](https://nx.dev/packages/web/executors/dev-server)
  * Builds then serves a generic web project.
* [`@nxext/stencil:*`](https://nxext.dev/docs/stencil/executors.html) (unofficial)
  * Builds, serves, tests Stencil

For this PR, I chose to stick to `run-commands` because:
1. It's easier to review.
  * Less changes in the PR, and the commands are pretty much the same as they were in `package.json`.
6. It's easier to revert out of `Nx` if we need to.
  * JSUI uses Gulp, which is harder to maintain, and reverting out of it would be very difficult.
7. It's more predictable.
  * We already know what the commands do. Some executors take complex configurations, and could have unexpected side-effects.

I made an exception for the `jest` executor because it's configured identically to the command line and at some point I wanted to divide it into [Executor configurations](https://nx.dev/recipes/executors/use-executor-configurations). However, this isn't very important, and I can change it to a `nx:run-commands` if you want.

## Moved `external-builds`
Prior to this PR, Headless was building a dev/browser version of itself directly in the Atomic `src` directory. I noticed that Nx was inconsistently choosing to cache or not to cache those files, which I thought was the source of some caching issues I had, so I replaced `packages/atomic/src/external-builds` with `packages/headless/www`.

This change did not fix the issues I had, but in the spirit of making projects as self-contained as possible, I decided to keep this change.


.github/actions/setup/action.yml
```diff
@@ -38,6 +39,10 @@ runs:
         key: cy-${{ hashFiles('**/package-lock.json') }}
     - run: npm ci
       shell: bash
+    - uses: actions/cache@v3
+      with:
+        path: node_modules/.cache/nx
+        key: nx-${{ github.sha }}
```
btaillon, line 45: I'm not gonna remove the cache action above yet until I've finished adding caching to all necessary tasks.

This new action is helpful because, now, when we call a command line `npm run prod`, it also builds the parent tasks. By adding the cache, it _should_ know not to rebuild everything.

nx.json
```diff
@@ -3,52 +3,48 @@
...
+          "cached:build:doc:extract",
+          "cached:build:doc",
+          "cached:build:locales",
+          "cached:build:stencil",
+          "cached:build:storybook:assets-list",
+          "cached:build:storybook",
+          "cached:validate:storybook",
+          "cached:validate:definitions",
+          "cached:test:unit"
+        ]
```
btaillon, line 18: Unfortunately, based on Nx's documentation, it looks like all cacheable operations need to be defined at the root `nx.json`.

I added the `cached:` prefix to cachable tasks for the following reasons:
* Viewable in projects
  * While modifying/refactoring task definitions, it's very important to know which tasks are cached and which aren't.
* Searchability
  * It's not particularly intuitive when reading `project.json` files that some tasks are cached at all, let alone how to make a task cacheable. With this change, the `cached:` prefix is a bit of a red flag. When somebody searches `cached:`, they'll find `cacheableOperations`.
* Separation
  * Both the `atomic` and `atomic-hosted-page` projects have a `build:stencil` task. However, we don't want to cache `atomic-hosted-page`'s (yet).

package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "postinstall": "husky install && patch-package && npm run link:angular -w @coveo/atomic-angular-samples",
     "dev:atomic": "concurrently \"npm run -w @coveo/headless dev\" \"npm run -w @coveo/atomic dev\"",
+    "clean": "nx run-many --target=clean",
```
btaillon, line 7: While testing, I frequently needed to clean the packages, so I added a clean task here.

packages/atomic/package.json
```diff
@@ -24,33 +24,27 @@
     "collection-manifest.json"
   ],
   "scripts": {
```
btaillon, line 26: I decided to convert all tasks to `nx:run-commands` tasks instead of `nx:run-scripts` tasks so we can have a clean `package.json` file with only the tasks that are relevant to us while developing.

packages/atomic/project.json
```diff
@@ -1 +1,264 @@
...
+    "e2e": ["{projectRoot}/cypress/**"],
+    "stories": ["{projectRoot}/src/**/*.{?(*([^.]).)stories.{tsx,css},mdx}"],
+    "source": [
+      "{projectRoot}/stencil.config.ts",
+      "{projectRoot}/collection-manifest.json",
+      "{projectRoot}/tailwind.config.js",
+      "{projectRoot}/svg.transform.js",
+      "{workspaceRoot}/tsconfig.json",
+      "{projectRoot}/tsconfig.json",
+      "{projectRoot}/src/**/*([^.]).!({{spec.{ts,tsx},spec.{ts,tsx}.snap},{?(*([^.]).)stories.{tsx,css},mdx}})",
```
btaillon, line 13: This line here is a huge mess for two reasons.

1. A named input cannot reference another named input.
    * For instance, I couldn't do `"source": ["!tests", "!stories"]`
2. Negative matchers always have priority
    * For instance, if you have the following named inputs:
      ```json
      {
        "tests": ["{projectRoot}/**/*.spec.ts"],
        "source": [
          "{projectRoot}/**/*.ts",
          "!{projectRoot}/**/*.spec.ts"
        ]
      }
      ```
      and you have a task with the following inputs:
      ```json
      {
        "inputs": ["source", "tests"]
      }
      ```
      Then the `spec.ts` files won't be detected, just as-if you only had `source` as an input.
  * I could be wrong about this, but so far my tests lead me to believe this.

`nx` uses [`fast-glob`](https://github.com/mrmlnc/fast-glob) for matching inputs, so I tested all complex globs like this one to make sure that all correct files are detected.

.github/actions/setup/action.yml
```diff
@@ -41,3 +42,7 @@ runs:
     - if: steps.build-cache.outputs.cache-hit != 'true'
       run: npm run build
       shell: bash
+    - uses: actions/cache@v3
+      with:
+        path: node_modules/.cache/nx
+        key: nx-${{ github.sha }}
```
btaillon, line 48: I'm not gonna remove the `build-cache` action above yet until I've finished adding caching to all necessary tasks.

This new action is helpful because, now, when we call a command line `npm run prod`, it also builds the parent tasks. By adding the cache, it _should_ know not to rebuild everything.

packages/atomic/project.json
```diff
@@ -1 +1,264 @@
...
+  },
+  "targets": {
+    "clean": {
+      "dependsOn": [],
+      "inputs": [],
+      "outputs": [],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "rimraf dist docs loader www storybook-static src/components/search/atomic-search-interface/lang src/generated",
+        "cwd": "packages/atomic"
```
btaillon, line 25: A big flaw with `nx:run-commands` is that the default `cwd` is the workspace root (`ui-kit`). Another flaw with the `cwd` is that it doesn't support variables nor `{projectRoot}`. We have to explicitly repeat `packages/atomic`.

It's ugly, but honestly not that big of a deal to copy and paste.

packages/atomic/project.json
```diff
@@ -1 +1,264 @@
...
+      "outputs": [
+        "{projectRoot}/dist",
+        "{projectRoot}/www",
+        "{projectRoot}/docs",
+        "{projectRoot}/loader",
+        "{projectRoot}/src/components.d.ts"
+      ],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "node --max_old_space_size=6144 ../../node_modules/@stencil/core/bin/stencil build",
```
btaillon, line 64: There's an unofficial Stencil executor we could use (see [`@nxext/stencil:build`](https://nxext.dev/docs/stencil/executors.html)), and this executor seems to just use the same config we already have, which would remove the need for referencing `../../node_modules` directly and probably ensure that `dev` works properly without messing with it so much.

The package itself doesn't seem very popular, but their monorepo which contains more executors has 350 stars: https://github.com/nxext/nx-extensions

I haven't tested it yet, since my concern for now wasn't to make `dev` scripts work and I didn't wanna add too much more to this PR.

packages/atomic/project.json
```diff
@@ -1 +1,264 @@
...
+        "cwd": "packages/atomic"
+      }
+    },
+    "cached:validate:definitions": {
+      "dependsOn": ["^build", "cached:build:stencil"],
+      "inputs": ["{projectRoot}/dist/types/components.d.ts"],
+      "outputs": [],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "tsc --noEmit --esModuleInterop --skipLibCheck ./dist/types/components.d.ts",
```
btaillon, line 74: There's an official `tsc` executor (see [`@nrwl/js:tsc`](https://nx.dev/packages/js/executors/tsc#@nrwl/js:tsc)), but its configuration is quite different from what we're using at the moment, so I preferred to stick to the command line.

packages/atomic/project.json
```diff
@@ -1 +1,264 @@
...
+      ],
+      "outputs": ["{projectRoot}/storybook-static"],
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": ["build-storybook", "ncp dist/ storybook-static/"],
+        "parallel": false,
+        "cwd": "packages/atomic"
+      }
+    },
+    "build": {
```
btaillon, line 121: for "top-level" tasks like "build" or "test", I decided to use the `nx:noop` executor with dependencies so that we can simply add as dependencies the list of tasks we need to be run, and Nx will figure out in what order they need to be run.

packages/bueno/project.json
```diff
@@ -1 +1,109 @@
...
+      "inputs": ["^source", "source"],
+      "outputs": ["{projectRoot}/dist"],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "node esbuild.mjs",
+        "cwd": "packages/bueno"
+      }
+    },
+    "cached:build:definitions": {
+      "dependsOn": ["cached:build:esbuild"],
```
btaillon, line 33: Technically, `tsc -d --emitDeclarationOnly --declarationDir dist/definitions` is able to run successfully without running `esbuild` first.

However, this task emits its declarations inside the `dist/definitions/` folder and `esbuild` emits its build artifacts in the `dist/` folder. If we run both tasks in parallel or in a varying order, the `esbuild` task's cache will randomly contain or not contain the definitions, and may occasionally delete the `dist/definitions/` directory (which did happen while I was testing, until I added this dependency).

If we really want those tasks to run in parallel, I could emit the definitions in a separate folder and then add another task which copies them into the `dist` folder.

**Question**: do you think I should do it or not?

packages/bueno/project.json
```diff
@@ -1 +1,109 @@
...
+    "dev": {
+      "dependsOn": [],
+      "inputs": [],
+      "outputs": [],
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": [
+          "nx run bueno:\"dev:esbuild\"",
+          "nx run bueno:\"dev:definitions\""
+        ],
```
btaillon, line 77: Normally, we should **never** run multiple `nx run` in parallel, since they both might automatically run the same dependent tasks at the same time. I made an exception here for `dev` since both tasks have no dependencies and since I just want it to work the same as in `master`.

packages/bueno/project.json
```diff
@@ -1 +1,109 @@
...
+        "parallel": true
```
btaillon, line 78: `true` is the default value, but I prefer to specify it explicitly since it's not very intuitive what the default value is.

packages/headless/project.json
```diff
@@ -1 +1,166 @@
...
+      "inputs": [],
+      "outputs": [],
+      "executor": "@nrwl/jest:jest",
+      "options": {
+        "jestConfig": "packages/headless/jest.config.js",
+        "watch": true,
+        "color": true,
+        "silent": false
+      },
+      "configurations": {
```
btaillon, line 154: At some point, I had all `test` tasks combined into one, which is why there's these configurations here. I had to divide them since they're not all cached and they have different dependencies.

I kept the configurations here since they could be useful.

You can use a specific configuration like so:
```sh
~/repos/ui-kit $ npx nx run headless:"test:watch":unit
```
OR
```sh
~/repos/ui-kit $ npm run -w @coveo/headless test:watch -- --configuration=unit
```

packages/atomic/project.json
olamothe: So what happens exactly here if we mess up, forget to update this ?

1- The caching stops being run optimally, and so our build take longer than needed when working locally (not the end of the world) 

2- It will somehow not take new code change into account when it should. So when you work locally, you might get something "corrupted" (for a lack of a better word), that does not actually represent changes you've made locally in a source file ?

If it's 2... that's a big blocker.

olamothe (commented):
I'll be honest, I have a fear that these changes will work for a couple weeks/months, and for some time we'll have caching that works well.

But something will sooner or later break this wall of configuration, and we'll have to spend some brain CPU cycle to maintain this.

If the configuration breaking is just caching that is sub-optimal, then it's more of a "meh" 🤷, we can live with it, nothing is blocking us to work and we can always "push it back".

If it does break the build, or completely break the local development flow, then it's a huge problem.

I hope it's the former ?

packages/atomic/project.json
```diff
@@ -1 +1,264 @@
-{}
+{
```
ThibodeauJF, line 1: This whole file might be a pain to maintain or reverse engineer if something unexpected happens. 

Why not incrementally add these settings with the low hanging fruits like the files we copy (80/20 rule)?

packages/atomic/project.json
ThibodeauJF: I'd favor using the core executor

packages/bueno/project.json
```diff
@@ -1 +1,109 @@
-{}
+{
```
ThibodeauJF, line 1: I don't think any optimization is necessary for bueno, its lightweight and quick. Perhaps we'll replace it with zod as well

packages/atomic/project.json
btaillon: For this particular line, it would be the first problem: changing some test or story files would cause an unnecessary rebuild of Atomic and its dependants.

However, for other lines than this one, the second problem is a real concern.

If we forget to update inputs to include new file types, then these changes won't be detected, and they might not be rebuilt locally (or even in the CI, if we decide to enable Nx Cloud). This wouldn't necessary be breaking though, since it shouldn't affect the CI, and fixing it should be pretty intuitive.

If we forget to update outputs, then we have a bigger problem. Locally, when loading from the cache, we would randomly get corrupted builds as you said. In the CI, even without Nx Cloud, it would run the first setup job correctly, but then the next jobs would use a corrupted build. This would be a particularly big deal for the publish job(s).

One way to mitigate the problem would be to automatically infer inputs/outputs of each task/target. For example: inferring inputs and outputs from `stencil.ts`. I imagine that's part of why Nx made executors. For example: a Stencil executor should be able to automatically have the correct outputs (AFAIK) by interpreting them from the Stencil configuration file.

Another solution would be to give up on caching. I think caching is going to get increasingly important as we scale the repo by adding more and more projects, but we could skip caching on specific critical jobs (e.g., publishing) or events (e.g., bump commits).

packages/bueno/project.json
btaillon: Good point, I should probably introduce some "good enough" defaults for smaller packages like this.

packages/atomic/project.json
btaillon: Most of the complexity comes from targets like "build:stencil" and "build:storybook", so you're right that going for low-hanging fruits like copying would make it simpler. However, these two steps take a long time to run, and therefore save a much more significant amount of time when cached.

I can incrementally add these changes, just like I didn't put as much effort in `dev` and `watch` scripts yet, but if we end up eventually adding the bigger/longer steps, we'll still end up with a config that's just as complex.

Using proper executors should significantly simplify the config since we wouldn't need to define their inputs/outputs (AFAIK).

btaillon: Update: I just found out that Nx supports "plugins", which allow us to dynamically generate `project.json` file contents (see [Project Inference](https://nx.dev/recipes/advanced-plugins/project-inference-plugins)). This could help us reduce repetition and write our configuration as code.

packages/atomic/project.json
btaillon: Relevant update: https://github.com/coveo/ui-kit/pull/2734#discussion_r1111984473

btaillon:
Closing this. I will instead make multiple PRs where I do individual tasks instead of whole projects.

PR #2738, opened 2023-02-21
fix(atomic,headless): fix analytics for number of results and results array

olamothe:
Two fix here: 

* For the `numberOfResults` we send to UA, we were sending the length of the results received client side. This is in contrast to the total number of results found in the index for a particular query that the JSUI sends. This was reported by a customer through support on discuss. As a rule of thumb, the way JSUI report events is... "law", for historical reasons. All in all this is small detail, and should not have much impact on the reporting or ML side of things. It's mostly only relevant when it's 0.

* There was a bigger bug though with the `numberOfResults` property and `results` array send to UA, as we were always sending the last previous state, as opposed to the current one for the current query. 

I spent some time trying to see if we could simplify all the work that's done in `analytics-utils`, because the code there is *quite* complex, for no really good reason.

However, any simplification I could think doing there would be a breaking change on the public API. So I think we'll have to live with it for some time.

So instead I focused on adding some tests in Atomic that would start to catch those kind of issues.

https://coveord.atlassian.net/browse/KIT-2301


packages/headless/src/features/search/search-actions.ts
```diff
@@ -98,19 +98,19 @@ export const executeSearch = createAsyncThunk<
...
-      analyticsClientMiddleware,
-      preprocessRequest,
-      logger,
-    });
+  const {description: eventDescription} = await searchAction.prepare({
+    getState: () => config.getState(),
+    analyticsClientMiddleware,
+    preprocessRequest,
+    logger,
+  });
```
btaillon, line 106: Ok, now I remember the reason why I added the whole "prepare" mechanism. The description here is meant to contain the same custom data as the analytics event we send, so I wanted to make sure they're consistent with eachother.

In `coveo.analytics`, I manually ensured that *some* parameter outside of customData was using the latest state, but it was the exception.

Instead, we should probably make it so that *everything* outside of `customData` uses the latest state.

With the current changes, the custom data may be different in the `/v2` call's `analytics.customData` object, and in the `/search` call's `customData` object.

btaillon: Alternatively, it may be worth revisiting whether we really want to include analytics in `/v2` calls

louis-bompart (approved):
Given the existing test coverage, the issue at stakes, and the test added, LGTM.

> However, any simplification I could think doing there would be a breaking change on the public API. So I think we'll have to live with it for some time.

Analytics is a critical function of the library, it's important that we take care of it.
I would slate an issue for V3 (We have an Epic for that).

packages/headless/src/features/search/search-actions.ts
olamothe: Discussed in scrum: What's important is the action cause. Custom data that changes as a result of the query are obviously not something that can be reported to the API that we use to perform a request.

PR #2744, opened 2023-02-24
fix(quantic): removed the result from the event sent by the result action component

mmitiche:
[SFINT-4794](https://coveord.atlassian.net/browse/SFINT-4794)

- Sending the result in the custom event with the result was causing a big delay when the result was too large.
- Turns out that the result is not really needed in the event sent by the Quantic Result Action component.
- The Quantic Result Action is meant to be wrapped by other components used in the result templates which give them access to the result object, thus there is no need to send the result in the custom event. Example:  [Quantic Result Copy To Clipboard component](https://github.com/coveo/ui-kit/tree/master/packages/quantic/force-app/main/default/lwc/quanticResultCopyToClipboard).

[SFINT-4794]: https://coveord.atlassian.net/browse/SFINT-4794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
@@ -38,6 +38,7 @@ export default class QuanticResultAction extends LightningElement {
    * The result to perform the action on.
    * @api
    * @type {Result}
+   * @deprecated
    */
   @api result;
   /**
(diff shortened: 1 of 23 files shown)


packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
```diff
@@ -38,6 +38,7 @@ export default class QuanticResultAction extends LightningElement {
    * The result to perform the action on.
    * @api
    * @type {Result}
+   * @deprecated
```
erocheleau, line 41: You can add some details after `@deprecated` maybe to explain the deprecation or to say "deprecated since version X"

SimonMilord (approved):

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.js
mmitiche: sure yeah 👌

PR #2748, opened 2023-02-27
fix(atomic): use closest for internal initialization utils

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2313

Some context, we used the [custom event return value](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent#return_value) as a basis for when a component has been initialized properly or not (the event’s payload is an “initialize” callback. The return value is basically:

- false if event is cancelable, and at least one of the event handlers which received event called [Event.preventDefault()](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault). We call prevent default from atomic-search-interface, for example/
- Otherwise true.

This is synchronous, if dealing with a simple DOM, like our basic search interfaces, this will always work. When we started to expose these custom events for our custom elements, we started to have issues with Stencil. When those events where dispatched from inside another Stencil component, they would reach the “atomic-search-interface” but with some delay, causing the dispatchEvent return value to be true. A typical race condition problem. We could argue that Stencil’s rendering inner workings is at fault, or that the method we used to make sure the component is initialized properly was not the best solution (flaky). https://github.com/coveo/ui-kit/pull/2242or we could have some sort of asynchronous polling method to check if init is called. 

TLDR: there is a false negative init error when using imbricated Stencil components.

### DIFF
--- a/packages/atomic/src/utils/initialization-utils.tsx
+++ b/packages/atomic/src/utils/initialization-utils.tsx
@@ -23,6 +23,7 @@ export const initializeEventName = 'atomic/initializeComponent';
 const initializableElements = [
   'atomic-recs-interface',
   'atomic-search-interface',
+  'atomic-relevance-inspector',
   'atomic-insight-interface',
   'atomic-external',
 ];
--- a/packages/atomic/src/utils/initialization-utils.tsx
+++ b/packages/atomic/src/utils/initialization-utils.tsx
@@ -144,29 +144,34 @@ export function InitializeBindings<SpecificBindings extends AnyBindings>({
       const element = getElement(this);
       element.setAttribute(renderedAttribute, 'false');
       element.setAttribute(loadedAttribute, 'false');
-      buildCustomEvent(initializeEventName, (bindings: SpecificBindings) => {
-        this.bindings = bindings;
+      const event = buildCustomEvent(
+        initializeEventName,
+        (bindings: SpecificBindings) => {
+          this.bindings = bindings;
 
-        const updateLanguage = () => forceUpdateComponent(this);
-        this.bindings.i18n.on('languageChanged', updateLanguage);
-        unsubscribeLanguage = () =>
-          this.bindings.i18n.off('languageChanged', updateLanguage);
+          const updateLanguage = () => forceUpdateComponent(this);
+          this.bindings.i18n.on('languageChanged', updateLanguage);
+          unsubscribeLanguage = () =>
+            this.bindings.i18n.off('languageChanged', updateLanguage);
 
-        try {
-          // When no controller is initialized, updating a property with a State() decorator, there will be no re-render.
-          // In this case, we have to manually trigger it.
-          if (this.initialize) {
-            this.initialize();
-            if (forceUpdate) {
+          try {
+            // When no controller is initialized, updating a property with a State() decorator, there will be no re-render.
+            // In this case, we have to manually trigger it.
+            if (this.initialize) {
+              this.initialize();
+              if (forceUpdate) {
+                forceUpdateComponent(this);
+              }
+            } else {
               forceUpdateComponent(this);
             }
-          } else {
-            forceUpdateComponent(this);
+          } catch (e) {
+            this.error = e as Error;
           }
-        } catch (e) {
-          this.error = e as Error;
         }
-      });
+      );
+
+      element.dispatchEvent(event);
 
       if (!closest(element, initializableElements.join(', '))) {
         this.error = new MissingInterfaceParentError(
--- a/packages/atomic/src/utils/initialization-utils.tsx
+++ b/packages/atomic/src/utils/initialization-utils.tsx
@@ -21,6 +21,7 @@ export type InitializeEventHandler = (bindings: AnyBindings) => void;
 export type InitializeEvent = CustomEvent<InitializeEventHandler>;
 export const initializeEventName = 'atomic/initializeComponent';
 const initializableElements = [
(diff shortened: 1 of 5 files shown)


packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -144,35 +144,31 @@ export function InitializeBindings<SpecificBindings extends AnyBindings>({
...
           }
+        } catch (e) {
+          this.error = e as Error;
         }
-      );
+      });
 
-      const canceled = element.dispatchEvent(event);
-      if (canceled) {
+      if (!closest(element, initializableElements.join(', '))) {
```
ThibodeauJF, line 171: Diff is wack, only changed this line & removed the event const.

"Closest" is used at line 44 already

louis-bompart (commented):
I have a hard time understanding what we're doing here.
Could you put some flesh on the issue or PR to describe the issue and/or an example of the issue we want to solve?

Tests would be nice too :D

btaillon:
Our events are normally supposed to bubble up through the ShadowDOM (see [`Event.composed`](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed)). I'm concerned about why the event isn't getting canceled, and whether it could cause other issues.

Is this issue happening for a specific environment?

packages/atomic/src/utils/initialization-utils.tsx
ThibodeauJF: https://github.com/coveo/ui-kit/pull/2242

ThibodeauJF:
@btaillon @louis-bompart 
Stencil screws that up, they'll intercept the event but it will bubble properly. This is why "closest" is used for the initializeBindings method already.

It's more apparent when we create some type of race condition. In the Stencil sample:

1. The results-manager component initializes as a stencil component
2. Atomic is inserted in the head and loads
3. atomic-result-list initializes and dispatches the event from inside the results-manager.
4. Stencil intercepts the event, and re-dispatches it, screwing up atomic-result-list's initialization 

If Atomic loads first, the atomic-result-list component initializes **before** results-manager is recognize as a Stencil component.

I'll add a new sanity test in the stencil sample package.

packages/samples/stencil/src/pages/hsp.html
```diff
@@ -0,0 +1,68 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Atomic in Stencil</title>
+    <link rel="stylesheet" href="/build/atomic/themes/coveo.css" />
+    <script type="module" src="/build/app.esm.js"></script>
+
+    <script>
+      const script = document.createElement('script');
+      script.type = 'module';
+      script.src = 'https://static.cloud.coveo.com/atomic/v2/atomic.esm.js'; // fix: src="/build/atomic/atomic.esm.js"
```
ThibodeauJF, line 13: switch it to see the fix in action

btaillon:
@ThibodeauJF

> Stencil intercepts the event, and re-dispatches it, screwing up atomic-result-list's initialization

Ah, I didn't know Stencil intercepts events like this. What are the conditions for an event to get canceled? Is it canceled by the parent Stencil component when it renders custom elements it doesn't own?

> If Atomic loads first, the atomic-result-list component initializes before results-manager is recognize as a Stencil component

I'm a bit confused about how this is possible. If `results-manager` is responsible for appending `atomic-result-list` to itself (like in your PR), then I would expect that initializing Atomic first makes it see `results-manager` as being an empty element.

I get your point though.

packages/samples/stencil/src/pages/hsp.html
```diff
@@ -0,0 +1,68 @@
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Atomic in Stencil</title>
+    <link rel="stylesheet" href="/build/atomic/themes/coveo.css" />
+    <script type="module" src="/build/app.esm.js"></script>
```
ThibodeauJF, line 8: Switch for `https://static.cloud.coveo.com/atomic/v2/atomic.esm.js` to see it fail

louis-bompart:
> Intercepted was treated as cancelled
by who/what code?

ThibodeauJF:
Deleted my older comments to avoir confusion if anyone is looking at this later one.
Added more context in the description.

btaillon (approved):

louis-bompart (approved):

packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -171,8 +173,9 @@ export function InitializeBindings<SpecificBindings extends AnyBindings>({
         }
       );
 
-      const canceled = element.dispatchEvent(event);
-      if (canceled) {
+      element.dispatchEvent(event);
+
+      if (!closest(element, initializableElements.join(', '))) {
```
olamothe, line 178: Do we have any kind of test for the "unhappy" path ?

ie: If I put an `atomic-facet` randomly outside of an interface, it should throw an error.

olamothe (approved):

packages/atomic/src/utils/initialization-utils.tsx
ThibodeauJF: Yes it's covered

PR #2751, opened 2023-02-27
fix(quantic): improved aria labels of the facet values

mmitiche:
[SVCC-2382](https://coveord.atlassian.net/browse/SVCC-2382)

- Added the facet count ti the aria label of the category facet values.
- Localized the labels used as the aria labels of the facet values.

#### Before: 
we were not using localized labels so we didn't support plural and singular labels.

https://user-images.githubusercontent.com/86681870/221673259-519d2719-6618-4291-8cfe-7aef9a7f31f7.mov

#### After: 
Now we are using localized labels so we do support plural and singular labels.

https://user-images.githubusercontent.com/86681870/221673295-d58b5b91-b15b-4595-b2d0-6729d34b0a11.mov

[SVCC-2382]: https://coveord.atlassian.net/browse/SVCC-2382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -1050,4 +1050,18 @@
       <protected>false</protected>
       <shortDescription>People also ask</shortDescription>
     </labels>
+    <labels>
+      <fullName>quantic_InclusionFilter</fullName>
+      <value>Inclusion filter on {{0}}; {{1}} result</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Inclusion filter on field; 1 result</shortDescription>
+    </labels>
+    <labels>
+      <fullName>quantic_InclusionFilter_plural</fullName>
+      <value>Inclusion filter on {{0}}; {{1}} results</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Inclusion filter on field; 2 results</shortDescription>
+    </labels>
 </CustomLabels>
--- a/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -1064,4 +1064,11 @@
       <protected>false</protected>
       <shortDescription>Inclusion filter on field; 2 results</shortDescription>
     </labels>
+    <labels>
+      <fullName>quantic_InclusionFilter_zero</fullName>
+      <value>Inclusion filter on {{0}}; no results</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Inclusion filter on field; no results</shortDescription>
+    </labels>
 </CustomLabels>
--- a/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -946,6 +946,41 @@
       <shortDescription>Show less</shortDescription>
     </labels>
     <labels>
+      <fullName>quantic_DidNotAnswerQuestion</fullName>
+      <value>Didn't answer my question at all</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Didn't answer my question at all</shortDescription>
+    </labels>
+    <labels>
+      <fullName>quantic_PartiallyAnsweredQuestion</fullName>
+      <value>Only answered part of my question</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Only answered part of my question</shortDescription>
+    </labels>
+    <labels>
+      <fullName>quantic_RequestWasNotQuestion</fullName>
+      <value>My request wasn't a question</value>
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>My request wasn't a question</shortDescription>
+    </labels>
+    <labels>
+      <fullName>quantic_Other</fullName>
+      <value>Other</value>
+      <language>en_US</language>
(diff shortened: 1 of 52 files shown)


SimonMilord (approved):

nathanlb (approved):
Nice!

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
```diff
@@ -64,8 +69,20 @@ export default class QuanticCategoryFacetValue extends LightningElement {
...
   }
 
+  get numberOfResults() {
+    return new Intl.NumberFormat(LOCALE).format(this.item.numberOfResults);
+  }
+
   get ariaLabelValue() {
-    return `Inclusion filter on ${this.facetValue}`;
+    const labelName = I18nUtils.getLabelNameWithCount(
+      'inclusionFilter',
```
lbergeron, line 78: I'd suggest to use the same casing as the custom label declaration.

```suggestion
      'InclusionFilter',
```

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -78,7 +86,15 @@ export default class QuanticFacetValue extends LightningElement {
   }
 
   get ariaLabelValue() {
-    return `Inclusion filter on ${this.formattedFacetValue}; ${this.numberOfResults} result`;
+    const labelName = I18nUtils.getLabelNameWithCount(
+      'inclusionFilter',
```
lbergeron, line 90:
```suggestion
      'InclusionFilter',
```

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1050,4 +1050,18 @@
       <protected>false</protected>
       <shortDescription>People also ask</shortDescription>
     </labels>
+    <labels>
+      <fullName>quantic_InclusionFilter</fullName>
```
lbergeron, line 1054: You should also define a custom label suffixed with `_zero`. Even if a facet should not display a value matching no items, it is possible to get into this situation when loading a search page from a bookmark, for instance.

lbergeron (commented):
Please add a label suffixed with `_zero`.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
mmitiche: good catch! thank you!

lvu285:
can you do a quick check for CategoryFacet, to make sure it works too, thanks!
(your example is for normal facet)

mmitiche:
> can you do a quick check for CategoryFacet, to make sure it works too, thanks!
> (your example is for normal facet)

Yes sure I've already manually tested the category facet as well!

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
mmitiche: this key will be used to get the right label from the labels object which is declared as the following:
```
  labels = {
     inLabel,
     inclusionFilter,
     inclusionFilter_plural,
   };
```
That's why I'm using a lower case first letter.

lbergeron: Ah right, my bad :facepalm:

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
lbergeron: nevermind

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
```diff
@@ -1,6 +1,9 @@
-import {api, LightningElement} from 'lwc';
-
+import LOCALE from '@salesforce/i18n/locale';
 import inLabel from '@salesforce/label/c.quantic_InLabel';
+import inclusionFilter from '@salesforce/label/c.quantic_InclusionFilter';
+import inclusionFilter_plural from '@salesforce/label/c.quantic_InclusionFilter_plural';
```
lbergeron, line 4: The `_zero` label should also be included.

packages/quantic/force-app/main/default/lwc/quanticFacetValue/quanticFacetValue.js
```diff
@@ -1,5 +1,8 @@
-import {LightningElement, api} from 'lwc';
 import LOCALE from '@salesforce/i18n/locale';
+import inclusionFilter from '@salesforce/label/c.quantic_InclusionFilter';
+import inclusionFilter_plural from '@salesforce/label/c.quantic_InclusionFilter_plural';
```
lbergeron, line 3: The `_zero` label should also be included.

lbergeron (changes requested):
The `_zero` label has been added but is not used by components.

packages/quantic/force-app/main/default/lwc/quanticCategoryFacetValue/quanticCategoryFacetValue.js
mmitiche: oops my bad, just added it, thank you!

lbergeron (approved):

PR #2752, opened 2023-02-27
ci: cache `build:stencil`

btaillon:
https://coveord.atlassian.net/browse/KIT-2293

# Changes
In this PR, I decided to enable caching for all `build:stencil` actions. This requires, for each affected project, to do the following:
* Define a target in `project.json`.
* Define the dependencies of the target.
  * e.g., `build:stencil` depends on `^build`, where the caret means "all dependencies".
* Define the `inputs` of the target.
  * e.g., if the inputs are `{projectRoot}/src/**`, then the cache will only be re-used if nothing was changed in `src`.
* Define the `outputs` of the target.
  * e.g., if the outputs are `{projectRoot}/dist`, then only the `dist` directory will be included in the cache.
* Choose an executor for the target.
  * e.g., for `build:stencil`, I decided to use the `nx:run-commands` executor.

In addition to the `build:stencil` targets, I also defined other targets in `project.json` files just to define their dependencies and enable parallelism. For now, these aren't cached and these use `nx:run-script` executors.

# Speed test
As a test, I deleted my cache then ran `npm run build` at the root of the repo twice in a row. The first build took 6 minutes, the second took 3 minutes.

# TODO next
* All other tasks (in separate PRs, just like this one).
* Enable caching in GitHub Actions.
* Improve dev scripts so we can run as many projects as we want, all at the same time.
* Separate storybook and validations from `build`


packages/atomic-hosted-page/project.json
```diff
@@ -1 +1,42 @@
...
+      "outputs": [
+        "{projectRoot}/dist",
+        "{projectRoot}/www",
+        "{projectRoot}/loader",
+        "{projectRoot}/src/components.d.ts"
+      ],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "node --max_old_space_size=6144 ../../node_modules/@stencil/core/bin/stencil build",
+        "cwd": "packages/atomic-hosted-page"
```
btaillon, line 23: The default `cwd` is the repo root. We can't use variables like `{projectRoot}` or `{workspaceRoot}` here, unfortunately.

packages/atomic-hosted-page/package.json
```diff
@@ -21,9 +21,8 @@
   ],
   "scripts": {
     "clean": "rimraf -rf dist/*",
-    "build": "npm run build:stencil && npm run validate:definitions",
-    "build:stencil": "node --max_old_space_size=6144 ../../node_modules/@stencil/core/bin/stencil build",
-    "dev": "npm run build:stencil -- --dev --watch --serve",
+    "build": "nx build",
+    "dev": "node --max_old_space_size=6144 ../../node_modules/@stencil/core/bin/stencil build --dev --watch --serve",
```
btaillon, line 25: Any task whose command was moved to `project.json`, I'm removing from the scripts here.

It would be nice, eventually, to have only useful tasks here.

packages/samples/stencil/project.json
```diff
@@ -1,3 +1,29 @@
...
+    ]
+  },
+  "targets": {
+    "cached:build:stencil": {
+      "dependsOn": ["^build"],
+      "inputs": ["^production", "production"],
+      "outputs": ["{projectRoot}/www", "{projectRoot}/src/components.d.ts"],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "node --max_old_space_size=6144 ../../../node_modules/@stencil/core/bin/stencil build",
```
btaillon, line 18: I added the max memory size here just like we do for all other Stencil build tasks.

Eventually, I think we should stop referencing `node_modules` directly. Since this project is inside `samples/`, it needs to go one level higher than the `build:stencil` targets.

packages/atomic-hosted-page/package.json
```diff
@@ -21,9 +21,8 @@
   ],
   "scripts": {
     "clean": "rimraf -rf dist/*",
-    "build": "npm run build:stencil && npm run validate:definitions",
-    "build:stencil": "node --max_old_space_size=6144 ../../node_modules/@stencil/core/bin/stencil build",
-    "dev": "npm run build:stencil -- --dev --watch --serve",
+    "build": "nx build",
```
btaillon, line 24: If, from the root of the repo, you do `nx run atomic-hosted-page:build`, and then, you cd into `packages/atomic-hosted-page` and run `npm run build`, the cache will **not** be re-used.

Explanation: I was able to verify that Nx's hash calculations are impacted by the current CWD.
* Changing the command itself never seems to make a difference.
  * From the Atomic package, `nx build` and `nx run atomic:build` share the same cache, which is different from the cache of `nx run atomic:build` at the root of the repo.
  * The cache is always added to the root `node_modules`, never in sub-packages.
* I didn't find the cause of this in their source code, they don't seem to explicitly cache the CWD path.
* Nx does cache the dependency graph, but when I do `nx graph --file=somefile.json`, changing the CWD doesn't cause any change.
* I was able to "fix" this by creating my own small script which executes a command with a different CWD, but I decided not to include it in this PR. I will revisit this later.

nx.json
```diff
@@ -3,52 +3,52 @@
...
-      ]
+      "dependsOn": ["build"]
     },
     "lint:check": {},
     "lint:fix": {},
     "publish:npm:release": {
-      "dependsOn": [
-        "build"
-      ]
+      "dependsOn": ["build"]
```
louis-bompart, line 42: - [ ] ensure nx.json is prettified
- [ ] ensure nx.json is protected by the lint step on CI

louis-bompart: can be split in another PR, but still, that's a req.

packages/atomic-hosted-page/package.json
louis-bompart: ngl: Don't care much about caching on local builds. What matter is the CI/CD, where the CWD is predictable (no hooman, no troubles).
So, in short... we don't care?

packages/atomic-hosted-page/project.json
```diff
@@ -1 +1,42 @@
...
+      "inputs": ["^production", "production"],
+      "outputs": [
+        "{projectRoot}/dist",
+        "{projectRoot}/www",
+        "{projectRoot}/loader",
+        "{projectRoot}/src/components.d.ts"
+      ],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "node --max_old_space_size=6144 ../../node_modules/@stencil/core/bin/stencil build",
```
louis-bompart, line 22: Can we use `workspaceRoot` to simplify this path

packages/atomic/project.json
```diff
@@ -1 +1,73 @@
...
+      "sharedTypescriptSources",
+      "{projectRoot}/stencil.config.ts",
+      "{projectRoot}/collection-manifest.json",
+      "{projectRoot}/tailwind.config.js",
+      "{projectRoot}/svg.transform.js",
+      "{projectRoot}/stencil-plugin/**",
+      "!{projectRoot}/src/components.d.ts",
+      "!{projectRoot}/src/external-builds/**",
+      "!{projectRoot}/src/generated/**",
+      "!{projectRoot}/src/components/search/atomic-search-interface/lang/**"
```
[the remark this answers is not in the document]
louis-bompart: > external-builds/** is generated by Headless's build. generated/** and lang/** are generated by build:locales.
Let's fix the problem at the root (in another PR).
 1. Build the locales in headless.
 2. Bundle the local of headless into atomic _from_ atomic.
 
 Also why the locales are build by headless? that's doesn't seems to fit with headless nature.

packages/atomic-hosted-page/project.json
btaillon: unfortunately not, `{workspaceRoot}` isn't supported in `options` (AFAIK, from a while ago). I'll try it again.

packages/atomic-hosted-page/package.json
btaillon: One flaw I forgot to mention is that running `npm run -w @coveo/somepackage` changes the CWD.

So, if you just did `npm run build` at the repo root, and then you did `npm run build -w @coveo/atomic`, then the cache would miss.

I guess that whether this is important depends on whether some people do this stuff often. Using `nx run` instead of `npm run` solves the issue.

packages/atomic/project.json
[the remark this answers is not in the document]
btaillon: My bad, `build:locales` is a script in Atomic, I meant to say that:
* `headless:build` generates:
  * `packages/atomic/src/external-builds/**`.
* `atomic:"build:locales"` generates:
  * `packages/atomic/src/generated/**`
  * `packages/atomic/src/components/search/atomic-search-interface/lang/**`.

nx.json
btaillon: our lint step uses ESLint, which doesn't work with JSON without a plugin. I prefer to stick to just using Prettier, so I'll just add Prettier to the lint-staged rules.

packages/atomic-hosted-page/package.json
louis-bompart: Fair, worst case we can change it in the future

louis-bompart (approved):
lgtm!

PR #2753, opened 2023-02-28
feat(atomic): add browser cond. export

louis-bompart:
This will allow users to use [conditions](https://nodejs.org/api/packages.html#resolving-user-conditions) to resolve `@coveo/atomic` differently.
We already did it for CJS/ESM with the `main` and `module` fields, but this adds another [community condition](https://nodejs.org/api/packages.html#community-conditions-definitions): browser.

`browser` will point towards the tree-shakeable version of the CLI, allowing the user to transpile/bundle/build its project using Atomic and have treeshaking, while having user-friendly import (i.e.  not `@coveo/atomic/dist/components/index.js` but `@coveo/atomic`)

### DIFF
--- a/packages/atomic/package.json
+++ b/packages/atomic/package.json
@@ -10,6 +10,13 @@
   },
   "main": "dist/index.cjs.js",
   "module": "dist/index.js",
+  "exports": {
+    "." : {
+      "browser":"./dist/components/index.js",
+      "import":"./dist/index.js",
+      "require":"./dist/index.cjs.js"
+    }
+  },
   "es2015": "dist/esm/index.mjs",
   "es2017": "dist/esm/index.mjs",
   "types": "dist/types/index.d.ts",


y-lakhdar:
I tried loading atomic in a vite-project and when export `./dist/components/index.js` for browsers, the dist size increases considerably.

## Before 
image

## After
image

louis-bompart:
> I tried loading atomic in a vite-project and when export `./dist/components/index.js` for browsers, the dist size increases considerably.
> 
> ## Before
> image
> 
> ## After
> image

Discussed in Slack, issue was that resolve.alias was not set ;)

louis-bompart:
Further investigation show that there's more to it (especially regarding types).
Back to draft.

louis-bompart:
Further further investigation shows that there's preliminary work to be done. Closing.

PR #2754, opened 2023-02-28
feat(headless,atomic): add useCustomDNS configuration option for headless

olamothe:
Add support for organization custom DNS 

Add a new configuration option on the core headless engine, which means it is available for search, service, commerce use cases.

The option `useCustomDNS` is undefined by default and optional (a required option would be a breaking change).

When left to undefined, the engine will log a warning that indicate to a developer that "hey, by the way, there is this new feature that allows you to use a custom DNS for your organization, and you should probably use it"

I've started a discussion about this in Slack, mostly because it has an impact for the Salesforce integration. 

I don't want to add a warning message that an implementer would see in the context of Quantic, but being unable to actually fix it.

I've left some "todos" in JSDOC/warning messages. We should link a documentation page about this that explains the changes in details and the benefits of doing so.

Also to note is the "bad" interaction between the `platformUrl` engine option and this new one.

If someones manually specify a `platformUrl`, we should respect it and it should take precedence over `useCustomDNS`. 

It is a legit use case to specify your own platform proxy if you desire.

In the next major version of Headless, we could envision either switching the default when left unspecified (from false to true), or just completely removing the option and automatically using the custom DNS when there is no platform URL manually configured.

https://coveord.atlassian.net/browse/KIT-2262


packages/headless/src/app/engine.ts
```diff
@@ -163,7 +166,27 @@ export function buildEngine<
   thunkExtraArguments: ExtraArguments
 ): CoreEngine<StateFromReducersMapObject<Reducers>, ExtraArguments> {
   const engine = buildCoreEngine(options, thunkExtraArguments);
-  const {accessToken, organizationId, platformUrl} = options.configuration;
+  const {accessToken, organizationId} = options.configuration;
+  let {platformUrl, useCustomDNS} = options.configuration;
+
+  if (shouldWarnAboutCustomDNS(options)) {
+    // @v3 make useCustomDNS true by default.
+    engine.logger.warn(
```
olamothe, line 174: @jpmarceau / @srosenberg-coveo : This is the warning I was describing above.

packages/headless/src/api/search/search-api-client.ts
```diff
@@ -74,7 +75,13 @@ export class SearchAPIClient implements FacetSearchAPIClient {
     req: PlanRequest
```
btaillon, line 75: Some requests may use `${orgId}.admin.org.coveo.com`, which I think is the default for anything that's not Search API nor Analytics related. I recommend double-checking that we have at-least one functional (or e2e) test that sends a query per request type.

packages/headless/src/app/engine-configuration.ts
```diff
@@ -49,6 +49,7 @@ export interface EngineConfiguration {
    * Allows configuring options related to analytics.
    */
   analytics?: AnalyticsConfiguration;
+  useCustomDNS?: boolean;
```
btaillon, line 52: # Problems
I'm not sure how I feel about `useCustomDNS` as a name 🤔 

"Custom" makes me think that this is the option I would need if I want to customize the URLs (e.g., use my own proxy).

As for "DNS", maybe it's because I have less back-end experience, but that makes me think it may somehow allow me to choose a different DNS (e.g., CloudFlare instead of Bell).

As for "use`, it feels reminiscent of React hooks.

Additionally, I'm not sure if I find it intuitive that `platformUrl` overrides `useCustomDNS`.

# Solution?

Let's envision what we'd like configuration to look like in the future (e.g., v3).

Using a proxy (`platformUrl`) feels like a less common use case, and it strongly relates to `useCustomDNS`, so I think that we should group them together. Also, I often hear of the feature to customize your URL referred to as "proxy", so maybe we could use that name.

I think that configuration in v3 could look something like:
```ts
interface Proxy { // Same interface as the return type of customDNSUrl()
  platform: string;
  search: string;
  analytics: string;
}

interface EngineConfiguration {
  // ... all the same options we currently have on `master`, except `platformUrl` is removed.
  proxy?: Proxy // Defaults to `customDNSUrl()`
}
```

If we went with an approach like this for v3, then for now, all we would have to do is:
1. Deprecate `platformUrl` (bonus: this would show in people's code editors that they need to prepare for the new config).
2. Add `proxy`, which would default for now to our current URL logic.
3. Export `customDNSUrl()` and instruct implementers who want to try the new DNS early to set `proxy` to `customDNSUrl()`.

As for the name `customDNSUrl()`, I'm not sure what I prefer, it could be `getOrganizationProxy`.

ThibodeauJF: Same with Benjamin I also don't like "custom" DNS, it feels like it's the customer own unofficial proxy or something... Perhaps use org DNS or something that sounds more "official"

I don't have a better idea right now 😄

louis-bompart: For the name, we're one of the last stop for this feature release, maybe someone before us thought of a/the user-friendly/public name?

olamothe: I can start a survey with the team owning this feature in the platform in Slack.

olamothe: As for the change for the `proxy` option as you describe (regardless of the name we end up using), there's one problem that I wanted to circumvent, which is that I wanted 3 different state for the warning:

1- There's this new option that you are not leveraging, you should look into it, so we warn you.
2- There's this new option that you know about, but you've deliberately opted out (because it's not possible, because it's not worth it for you etc.). So we don't warn you.
3- There's this new option that you know about and you've opted in, so we don't warn you.


Hence why the optional `boolean` type was interesting because of the 3 different types available.

olamothe: If we keep the suggested optional object approach, the same thing could be replicated, mind you.

We'll still have an option conflict for v2 which we will have to warn/handle (same as here):

If someone configures both the `proxy` object and the `platformUrl` string.

packages/headless/src/api/platform-client.ts
```diff
@@ -134,6 +127,28 @@ function coveoCloudURL<E extends PlatformEnvironment>(
   return `https://${subdomain}${urlEnv}${urlRegion}.cloud.coveo.com`;
 }
 
+/**
+ * Return the unique DNS entry for a given organization identifier.
+ * @param orgId The organization identifier.
+ * @param env Optional. The environment (prod, hipaa, staging, dev) that the organization belongs to. Default to `prod`.
```
jpmarceau, line 133:
```suggestion
 * Returns the unique DNS entry for a target organization identifier.
 * @param orgId The organization identifier.
 * @param env Optional. The environment (prod, hipaa, staging, dev) that the organization belongs to. Defaults to `prod`.
```

packages/headless/src/api/platform-client.ts
```diff
@@ -134,6 +127,28 @@ function coveoCloudURL<E extends PlatformEnvironment>(
...
+
+  const platform = `https://${orgId}.org${envSuffix}.coveo.com`;
+  const analytics = `https://${orgId}.analytics.org${envSuffix}.coveo.com/rest/v15`;
+  const search = `${platform}/rest/search/v2`;
+  return {platform, analytics, search};
+}
+
+/**
+ * Return the base Coveo platform URL, based on environment and region.
+ * @deprecated Coveo now offers unique DNS for each organization. Consider switching to customDNSUrl utility function instead.
```
jpmarceau, line 147:
```suggestion
 * Returns the base Coveo platform URL, based on environment and region.
 * @deprecated Coveo now offers unique DNS for each organization. Consider switching to customDNSUrl utility function instead.
```

packages/headless/src/api/platform-client.ts
```diff
@@ -144,6 +159,13 @@ export function platformUrl<E extends PlatformEnvironment>(
   return coveoCloudURL('platform', options);
 }
 
+/**
+ * Return the Coveo analytics platform URL, based on environment and region.
+ * @deprecated Coveo now offers unique DNS for each organization. Consider switching to customDNSUrl utility function instead.
```
jpmarceau, line 164:
```suggestion
 * Returns the Coveo analytics platform URL, based on environment and region.
 * @deprecated Coveo now offers unique DNS for each organization. Consider switching to customDNSUrl utility function instead.
```

packages/headless/src/app/engine-configuration.ts
btaillon: > As for the change for the `proxy` option as you describe (regardless of the name we end up using), there's one problem that I wanted to circumvent, which is that I wanted 3 different state for the warning:

From your comments, I don't understand what's the issue with the `proxy` object, it seems to be able to handle all three cases (as you imply in your second comment) 🤔 

> If someone configures both the proxy object and the platformUrl string.

In both cases, I'd assume that the new option (`useCustomDNS` / `proxy`) would be prioritized, I don't think that's an issue.

btaillon: > `useOrgEndpoints`?

I do like "OrgEndpoints"

erocheleau: @btaillon 
We cannot default to `customDNSUrl()` right now, so it would need to be an interface param that accepts `undefined`.
The problem as stated by @olamothe is what are you going to do in the code?

Right now we're in this state:
If `proxy` is `undefined` we need to default to the current endpoints, and display a warning.
If you pass your own custom `proxy` object, then no issue.
But how do you have customers giving us the orgId endpoints `proxy` without asking them to have to construct that object? 
They would have to know to call a method like `getOrgEndpoints()` every time they want to initialize?

To me it seems a lot simpler to just have a Boolean with "three" states as Oli described?

packages/headless/src/api/platform-client.ts
```diff
@@ -134,6 +127,28 @@ function coveoCloudURL<E extends PlatformEnvironment>(
...
+ * @param orgId The organization identifier.
+ * @param env Optional. The environment (prod, hipaa, staging, dev) that the organization belongs to. Default to `prod`.
+ * @returns
+ */
+export function customDNSUrl(orgId: string, env: PlatformEnvironment = 'prod') {
+  const envSuffix = env === 'prod' ? '' : env;
+
+  const platform = `https://${orgId}.org${envSuffix}.coveo.com`;
+  const analytics = `https://${orgId}.analytics.org${envSuffix}.coveo.com/rest/v15`;
+  const search = `${platform}/rest/search/v2`;
```
erocheleau, line 141: Do you want to put the `v2` into a const for when we have `v4` you know? ;)

packages/headless/src/api/search/search-api-client.ts
erocheleau: Which requests calls `.admin.` ?
I'm curious

erocheleau:
@olamothe Did you mean to include this new param also on the insight-engine and case-assist-engine?
I'm pretty sure it's the same thing for the base Url, just a different path

packages/headless/src/api/search/search-api-client.ts
btaillon: I recall that anything related to Insights is under `.admin.`

packages/headless/src/app/engine-configuration.ts
btaillon: > They would have to know to call a method like getOrgEndpoints() every time they want to initialize?

That's what I was thinking. We're already asking them to import `buildSearchEngine`, I don't find it unreasonable to expect them to import and call `getOrgEndpoints` if they want to use the org endpoints early (before v3).

Once we upgrade to v3, we could make `proxy` default to `getOrgEndpoints`.

btaillon: > To me it seems a lot simpler to just have a Boolean with "three" states as Oli described?

Currently, yes. Maybe we could even stick to that for now and then change it in v3.

However, in v3, I feel like it'll be confusing to have two separate parameters (`platformUrl` and `useOrgEndpoints`) controlling the same thing.

packages/headless/src/app/engine-configuration.ts
```diff
@@ -49,6 +49,14 @@ export interface EngineConfiguration {
    * Allows configuring options related to analytics.
    */
   analytics?: AnalyticsConfiguration;
+  /**
+   * Specify that the custom DNS URL scheme should be used to communicate with the Coveo platform.
+   *
+   * This is a recommended setting for most implementation, as this will simplify the overall configuration of a Coveo deployment in a multi region scenario.
+   *
+   * TODO: Insert Coveo documentation link
```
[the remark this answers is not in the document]
olamothe: I'll remove the TODO. 

> Maybe the simplest thing we could do is to not include links during the release, and only add them after I publish the docs?

Yep !

packages/headless/src/app/engine-configuration.ts
olamothe: Second problem about requiring a function to configure the object: You now need users to send us the same org id config property 2 times:

```
buildSearchEngine({configuration: {
  organizationId: 'my-org',
  accessToken: 'my-token',
  endpoints: getOrgEndpoints('my-org')
}});
```

olamothe:
Update: 

`organizationEndpoints` has been decided as the name for this feature, with the help of @jpmarceau , and approved by Team Foundation that is ultimately in charge of the whole active-active feature of the platform. So I've mass renamed `customDNS` -> `organizationEndpoints`



@btaillon Your proposition to use an object instead of a boolean to configure causes these issues: 

1- Not all engine will require the same object shape (there should be no "search" property or notion for a case assist engine, for example). 

So this potentially means generic types/generic function/different functions to configure this object. 

Which exposes more complexity to end users: It is harder to reason about when you see generic types (from an implementer POV) than a simple boolean, IMO.

2- Any function that we would expose to configure that object would need to have a parameter for `organizationId`: So we'd end up requiring the same configuration option when building an engine: 
```
buildSearchEngine({configuration: {
  organizationId: 'my-org',
  accessToken: 'my-token',
  endpoints: getOrgEndpoints('my-org')
}});
```

Which seems silly ? 🤔 

So we'd potentially have to also deprecate `organizationId` option, which is very impactful for all our users.

packages/headless/src/api/search/search-api-client.ts
olamothe: No, insights panel is under `/search`.

Insight panel is working correctly with these changes.

olamothe:
> @olamothe Did you mean to include this new param also on the insight-engine and case-assist-engine? I'm pretty sure it's the same thing for the base Url, just a different path

With these changes, insight panel is working correctly in the local page we have in this repo.

Haven't tested case assist, I don't have a setup for that.

from Kevin Larose:


```
There is a list of routes in the proxy that are allowed for the custom DNS.
Customer Service api is part of it. We keep using the phrase only "search" is allowed on that DNS but the definition of search was adapted for active-active.

    allowedRoutesPerDomain:
      search:
        routeIds:
          - search
          - ml-user-profile
          - case-assist
          - insight-panel-query
          - searchinterface
          - searchinterface-handshake
          - hostedpages
          - insight-panel-interface
          - ipx-interface
          - commerceService
          - catalogs
          - commerceProductListings
```

jpmarceau (approved):
I'll adjust my docs PR 👍

packages/headless/src/app/engine-configuration.ts
btaillon: I see. In this case, I'm actually wondering if it would make more sense to use a callback instead of an object, which feels like something I see a lot in other libraries when they want to allow deeper customization.

Example:
```ts
buildSearchEngine({configuration: {
  organizationId: 'my-org',
  accessToken: 'my-token',
  configureEndpoints: getOrgEndpoints
}});
```
OR
```ts
buildSearchEngine({configuration: {
  organizationId: 'my-org',
  accessToken: 'my-token',
  configureEndpoints: ({ organization }) => getOrgEndpoints({ organization, environment: 'stg' })
}});
```

Where engines' configurations could have a signature like:
```ts
export type EndpointsConfigurator<T extends AnyEndpoints> = (options: {
  organization: string;
  environment: PlatformEnvironment;
  region: PlatformRegion;
}) => T;
// ...
export interface EngineConfiguration {
  // ...
  configureEndpoints: EndpointsConfigurator<SearchEndpoints>;
}
// ...
export interface CaseAssistEngineConfiguration {
  // ...
  configureEndpoints: EndpointsConfigurator<CaseAssistEndpoints>;
}
```

In this case, `getOrgEndpoints` could return endpoints for all use cases and it wouldn't really matter, since the engine would only actually store the needed ones.

packages/headless/src/app/engine-configuration.ts
```diff
@@ -49,6 +51,14 @@ export interface EngineConfiguration {
...
    */
   analytics?: AnalyticsConfiguration;
+  /**
+   * Whether to use the organization endpoints to communicate with the Coveo platform.
+   *
+   * We recommend setting this to `true` for most implementations, since it has resiliency benefits and simplifies the overall configuration for multi-region deployments.
+   *
+   * Cannot be set to `true` while the `platformUrl` is populated. In such cases, `useOrganizationEndpoints` is ignored.
+   */
+  useOrganizationEndpoints?: boolean;
```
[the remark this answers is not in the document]
olamothe: The `getOrganizationEndpoints` utility function accept a parameter for the environment. It defaults to `prod`, but if you pass `hipaa` you'll get the proper endpoints.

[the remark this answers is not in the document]
jpmarceau: Sorry I think I wasn't clear: when I _initialize my engine_, do I need to tell Headless that it's supposed to use the orgipaa domain? If I just tell Headless my orgId, I give it a token and I set `useOrganizationEndpoints` to `true`, I don't understand how Headless could understand that it's dealing with a HIPAA org?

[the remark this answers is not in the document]
olamothe: You would do it like this:
```
const {platformUrl} = getOrganizationEndpoints('orgid', 'hipaa');

buildSearchEngine({
 configuration: {
      orgId,
      accessToken,
      platformUrl
  }
})
```

packages/headless/src/app/case-assist-engine/case-assist-engine.ts
```diff
@@ -122,6 +122,11 @@ function createCaseAssistAPIClient(
   logger: Logger
 ) {
   return new CaseAssistAPIClient({
+    useOrganizationEndpoints: isNullOrUndefined(
+      configuration.useOrganizationEndpoints
+    )
+      ? false
+      : configuration.useOrganizationEndpoints,
```
btaillon, line 129: equivalent to this:
```suggestion
    useOrganizationEndpoints: configuration.useOrganizationEndpoints ?? false,
```

PR #2758, opened 2023-03-06
fix: prevent unnecessary query suggestions

y-lakhdar:
https://coveord.atlassian.net/browse/KIT-2327

Prevent suggestion queries to be fired when `disable-search` is set to `true`.


packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
```diff
@@ -478,7 +478,7 @@ describe('Search Box Test Suites', () => {
       SearchBoxSelectors.submitButton().click({force: true});
     });
 
-    SearchBoxAssertions.assertHasSuggestionsCount(0);
+    SearchBoxAssertions.assertNoSuggestionGenerated();
```
btaillon, line 481: Just to make sure: when you revert your changes in `atomic-search-box.tsx`, does this test fail?

Since there's normally a delay between typing and suggestions appearing, there's a chance that the test may pass anyway. I suspect that the delay introduced by `.type(...)` already prevents this, but I want to make sure.

btaillon (approved):

packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
y-lakhdar: Yes, the test fails with the following error
```
Timed out retrying after 4000ms: Expected [ <button#atomic-search-box-2kao5-left-suggestion-qs-test.flex.px-4.min-h-[40px].items-center.text-left.text-neutral-dark.hover:bg-neutral-light.cursor-pointer>, 1 more... ] not to exist in the DOM, but it was continuously found. Queried from element: <document>
```

louis-bompart (approved):

ThibodeauJF (approved):

olamothe (approved):

PR #2761, opened 2023-03-07
feat(atomic-hosted-page): add new atomic-simple-builder component

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2335

Goes along with https://github.com/coveo/search-interface-service/pull/786

This pulls the new route, which takes a page made with the builder and inserts it into a customer's page, just like the hosted pages.

It was either adding a different config, or having a conditional config, so I decided to simply share some code and make a new component. Its ownership will be more clear and future-proof.

Open to different names...

Proof:
Screen Shot 2023-03-07 at 2 56 29 PM

### DIFF
--- a/packages/atomic-hosted-page/src/components/atomic-simple-builder/atomic-simple-builder.tsx
+++ b/packages/atomic-hosted-page/src/components/atomic-simple-builder/atomic-simple-builder.tsx
@@ -57,7 +57,7 @@ export class AtomicSimpleBuilder implements ComponentInterface {
 
     try {
       const pageResponse = await fetch(
-        `${platformUrl}/rest/organizations/${options.organizationId}/searchinterfaces/${options.interfaceId}/hostedpage`,
+        `${platformUrl}/rest/organizations/${options.organizationId}/searchinterfaces/${options.interfaceId}/hostedpage/v1`,
         {
           headers: {
             Authorization: `Bearer ${options.accessToken}`,
--- a/packages/atomic-hosted-page/src/components/atomic-simple-builder/atomic-simple-builder.tsx
+++ b/packages/atomic-hosted-page/src/components/atomic-simple-builder/atomic-simple-builder.tsx
@@ -1,12 +1,12 @@
 import {Schema, StringValue} from '@coveo/bueno';
 import {Component, ComponentInterface, Method, Element} from '@stencil/core';
-import {processHostedPage} from './hosted-pages';
+import {processHostedPage} from '../atomic-hosted-page/hosted-pages';
 
-interface AtomicHostedPageInitializationOptions {
+interface AtomicSimpleBuilderInitializationOptions {
   /**
-   * The unique identifier of the hosted page.
+   * The unique identifier of the search interface.
    */
-  pageId: string;
+  interfaceId: string;
   /**
    * The unique identifier of the target Coveo Cloud organization (e.g., `mycoveocloudorganizationg8tp8wu3`)
    */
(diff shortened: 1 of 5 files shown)


packages/atomic-hosted-page/src/pages/simple-builder.html
```diff
@@ -0,0 +1,32 @@
...
+          TODO: switch for following values when route in prod
+          interfaceId: '0c9eaa4a-8dab-4c11-83d6-24dd1feb961e',
+          accessToken: 'xx71d82ce9-f6fe-4be0-9209-c0e43f6fae81',
+          organizationId: 'searchuisamples',
+        */
+        await builder.initialize({
+          interfaceId: '45f9a933-b329-491e-bb15-64105eeea7e7',
+          accessToken: 'xx4bf05218-829b-4839-ab16-6ef044146909',
+          organizationId: 'hostedsearchpagetest5ftf64ka',
+          platformUrl: 'https://platformdev.cloud.coveo.com',
```
ThibodeauJF, line 24: If you run the SIS service locally to test this without cors issues:
```
npm install -g local-cors-proxy
lcp --proxyUrl http://localhost:8222
```

louis-bompart (approved):
all seems fine to me

olamothe (approved):
I think two component is fine.

I was writing something along the line of being able to offer one component, that allows you to seamlessly switch between the two modes (simple-builder vs hosted-page).

But I realized that when we create an hosted page from one built with the simple-builder, the search page ID in the DB will change, right ?

And so that it's just impossible to have one config that will allow you to switch between the two modes.

packages/atomic-hosted-page/src/components/atomic-simple-builder/atomic-simple-builder.tsx
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * The access token to use to authenticate requests against the Coveo Cloud endpoints. Typically, this will be an API key or search token that grants the privileges to execute queries and push usage analytics data in the target Coveo Cloud organization.
+   */
+  accessToken: string;
+  /**
+   * The Plaform URL to use. (e.g., https://platform.cloud.coveo.com)
+   * The platformUrl() helper method can be useful to know what url is available.
+   * @defaultValue `https://platform.cloud.coveo.com`
+   */
+  platformUrl?: string;
```
btaillon, line 23: We should add a TODO for v3 to support CustomDNS

packages/atomic-hosted-page/src/components/atomic-simple-builder/atomic-simple-builder.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * A Web Component used to inject a [Coveo Search Interface made with the simple builder](https://docs.coveo.com/en/m7e92019/adobe/build-the-search-solution-using-a-coveo-ui-library-directly#search-interface-builder) in the DOM.
+ * Pulls from the [Search Interfaces API](https://platform.cloud.coveo.com/docs?urls.primaryName=Search%20Interface%20Service#/)
+ * @internal
+ */
+@Component({
+  tag: 'atomic-simple-builder',
```
btaillon, line 32: `atomic-simple-builder` makes me think that we build the page in Atomic rather than it being hosted somewhere else.

I would prefer a name like `atomic-hosted-simple-page` for this component, and the other component could keep its current name.

btaillon (approved):

ThibodeauJF:
Yes @olamothe 2 components for 2 endpoints (and a different ID as you mention). This way it's future-proof if one of those endpoint changes or new options are added to one but not the other, etc.

PR #2763, opened 2023-03-08
fix(headless): add warning around partial state for search parameter manager

olamothe:
Context:

Search parameter manager is a module that helps synchronize an object (typically parsed from an URL) representing what should be applied to headless as an initial state.

The code in that module cannot assume the feature(s) that are available on the engine when it tries to synchronize the parameters. 

It's typically the job of the various controller, and more precisely the `loadXYZActions` function we expose) to modify the engine and add the proper reducer at runtime.

The parameter manager also cannot load every reducer pre-emptively, since it would defeat the whole system we have in place to help tree shake Headless to it's minimum payload size.

This PR adds a warning that tries to be a helpful as possible to developers, telling them that a part of the search parameters can't be synchronized and which `load` function to use to modify the engine. 

https://coveord.atlassian.net/browse/KIT-2253

### DIFF
--- a/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
+++ b/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
@@ -1,4 +1,5 @@
 import {RecordValue, Schema} from '@coveo/bueno';
+import {Logger} from 'pino';
 import {CoreEngine} from '../../../app/engine';
 import {partitionIntoParentsAndValues} from '../../../features/facets/category-facet-set/category-facet-utils';
 import {FacetValueRequest} from '../../../features/facets/facet-set/interfaces/request';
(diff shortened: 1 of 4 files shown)


packages/headless/src/integration-tests/field-suggestions.test.ts
```diff
@@ -198,7 +198,7 @@ describe('field suggestions', () => {
     });
 
     it('can update captions', async () => {
-      const rawValue = 'pdf';
```
olamothe, line 201: Unrelated fix: Seems this test started failing, most probably due to some content change in the index.

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
```diff
@@ -124,20 +125,21 @@ export function validateParams(
...
+    ...getFacets(state, logger),
+    ...getCategoryFacets(state, logger),
+    ...getNumericFacets(state, logger),
+    ...getDateFacets(state, logger),
   };
 }
 
-function getQ(state: Partial<SearchParametersState>) {
+function getQ(state: Partial<SearchParametersState>, logger: Logger) {
   if (state.query === undefined) {
```
btaillon, line 141: If the key exists in the state, even if its value is `undefined`, it could be sign that the value was updated to `undefined`.

I think we should check instead whether the key exists in the state like so:
```suggestion
  if ('query' in state) {
```

btaillon (approved):

olamothe:
There is a problem with this approach, where we end up warning due to no fault of the developer working on an implementation.

Will re-open with this fixed

PR #2766, opened 2023-03-09
fix(quantic): invalid inline links disabled in the smart snippet answer

mmitiche:
[SFINT-4856](https://coveord.atlassian.net/browse/SFINT-4856)

### Issue:
Sometimes anchor tags without an `href` are returned in the smart snippet answer, these links are considered as invalid links, thus we shouldn’t display them as links and we shouldn't log analytics when these invalid links are clicked:
image-20230222-223034

### Solution:
The invalid links are no longer displayed as links and analytics are no longer sent when these invalid are being clicked.

https://user-images.githubusercontent.com/86681870/224106845-4a39889c-ece0-4d18-a138-dc8c34fd4f90.mov

#### Before:
ULinSS

#### After:
ULinSSFix

[SFINT-4856]: https://coveord.atlassian.net/browse/SFINT-4856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
LGTM!

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetAnswer/quanticSmartSnippetAnswer.js
```diff
@@ -59,34 +59,59 @@ export default class QuanticSmartSnippetAnswer extends LightningElement {
     // eslint-disable-next-line @lwc/lwc/no-inner-html
     snippetAnswerElement.innerHTML = this?.answer;
     this.bindAnalyticsToSmartSnippetInlineLinks();
+    this.disableInvalidInlineLinks();
```
erocheleau, line 62: I think this could be merged into something like:

```
this.handleLinksInSmartSnippets();

handleLinksInSmartSnippets() {
  this.inlineLinks.forEach((link) => {
    if(link?.href) {
      // Bind analytics
    } else {
      link.classList.add(disabledCSSClass);
    }
  }
}
```

PR #2767, opened 2023-03-09
feat: use querySelector with interface element

dguerinCoveo:
We are putting the IPX inside a shadow DOM, thus making its element invisible to document.querySelector. We need to use query selector on the search interface instead. 

Everything else seems to be working fine in the IPX when inside a shadow DOM


packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
```diff
@@ -99,7 +108,7 @@ export class AtomicIPXButton {
   }
 
   private get ipxModal() {
-    return document.querySelector('atomic-ipx-modal')!;
+    return this.bindings.interfaceElement.querySelector('atomic-ipx-modal')!;
```
ThibodeauJF, line 111: If this is something we do in the base modal & others as well, might be a good fix there too  🙏

ThibodeauJF: (For the use case of having multiple interfaces in 1 page)

dguerinCoveo: I see the use of document.querySelector in the following files: 

- [atomic-recs-list](https://github.com/coveo/ui-kit/blob/feat/SVCINT-2155/packages/atomic/src/components/recommendations/atomic-recs-list/atomic-recs-list.tsx#L236)
- [atomic-search-interface](https://github.com/coveo/ui-kit/blob/feat/SVCINT-2155/packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx#L263-L264)
- [default-init](https://github.com/coveo/ui-kit/blob/feat/SVCINT-2155/packages/atomic/.storybook/default-init.ts#L19)
- [recs-init](https://github.com/coveo/ui-kit/blob/feat/SVCINT-2155/packages/atomic/.storybook/recommendation/recs-init.ts#L19)
- [atomic-external](https://github.com/coveo/ui-kit/blob/feat/SVCINT-2155/packages/atomic/src/components/search/atomic-external/atomic-external.tsx#L39)

I could make the changes there as well, but I feel these changes might have sneaky impacts and I think I would not be the best person to assess those impacts. If its alright with you, I'd like to just modify the behaviour of the ipx-button so that I can deploy this and use it in the SIS quickly 😅

louis-bompart: You can proceed with the change only to this bit for this PR I think, but if you can open a PR with the change in the other places, we would for sure appreciate 🙏

louis-bompart (approved):
lgtm

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
dguerinCoveo: Deal!
https://coveord.atlassian.net/browse/SVCINT-2159

btaillon (approved):
Looks good!

mikegron (approved):

pathiery (approved):

PR #2768, opened 2023-03-09
feat: add interactiveResult controller for ProductListingEngine

thudon-coveo:
The ProductListingEngine was missing the interactiveResult controller.


louis-bompart (approved):

packages/headless/src/controllers/product-listing/result-list/headless-product-listing-interactive-result.ts
```diff
@@ -0,0 +1,58 @@
...
+  props: InteractiveResultProps
+): InteractiveResult {
+  let wasOpened = false;
+
+  const logAnalyticsIfNeverOpened = () => {
+    if (wasOpened) {
+      return;
+    }
+    wasOpened = true;
+    engine.dispatch(logDocumentOpen(props.options.result));
```
olamothe, line 49: What's important is the analytics getting logged for product listing, when a document is clicked.

If everything is good on that front, 🚢

olamothe (approved):

packages/headless/src/product-listing.index.ts
```diff
@@ -170,3 +170,10 @@ export type {
   ContextValue,
 } from './controllers/product-listing/context/headless-product-listing-context';
 export {buildContext} from './controllers/product-listing/context/headless-product-listing-context';
+
+export type {
+  InteractiveResultOptions,
+  InteractiveResultProps,
+  InteractiveResult,
+} from './controllers/product-listing/result-list/headless-product-listing-interactive-result';
+export {buildInteractiveResult} from './controllers/product-listing/result-list/headless-product-listing-interactive-result';
```
samisayegh, line 179: If we want to include this controller in the docs, we'll need to add it to [this array](https://github.com/coveo/ui-kit/blob/master/packages/headless/doc-parser/use-cases/product-listing.ts#L5)

packages/headless/src/controllers/product-listing/result-list/headless-product-listing-interactive-result.ts
samisayegh: This is a great point, the `logDocumentOpen` is a good action to use. In the case of commerce though, ML expects ~5 commerce fields in the `customData` object listed in [this sheet](https://docs.google.com/spreadsheets/d/1eegdQ4TYTYjfDmgTiqHdX97aGifCC2PhgG8kDYDJFfA/edit#gid=846427451) (see the [UA] product click tab).

The `logDocumentOpen` action today does not send `id`, `brand` etc. I think this is _okay_ for this PR because professional services are aware of the limitation, and are using middleware to enrich events with additional fields in the case of `search`. Doing the same in product listing would be consistent ...

In a follow-up PR, we could improve the `logDocumentOpen` action by adding the relevant commerce standard fields to both search and PLP clicks if/when we find them in the result. Notes:
- This is a little hacky so open to other suggestions. A commerce click action could be ideal, but requires initing the search engine with `commerce` option? Or provide a `buildCommerceSearchEngine` ...
- The solution only works if the catalog is using the standard fields. This should be a fair assumption since other functionality depends on it, but if it's not the case, middleware would still be needed.
- Ideally, only a product id (sku) would be enough, and the other fields could be enriched server-side. This would simplify the implementation for direct api users.

tl;dr I think this is good for now, and we can improve on it in the future.

samisayegh (approved):

packages/headless/src/product-listing.index.ts
thudon-coveo: Thanks !

drosenfeld-coveo (approved):

packages/headless/src/product-listing.index.ts
samisayegh: Yea, if CI is erroring, it's possible that documentation is missing on the controller.
If this is the case, it should be possible to recreate the error locally by running `npm run build`.

PR #2770, opened 2023-03-09
feat(Quantic): added smartSnippet and smartSnippetSuggestions components to exampleS…

SimonMilord:
[SFINT-4820](https://coveord.atlassian.net/browse/SFINT-4820)

Added the Quantic Smart Snippet and the Quantic Smart Snippet Suggestions components to the Example Search Page component.

A lot of lines were changed mostly due to prettier upon save. Lines 132 - 140 is where its at!

Example in a community (the ¶ is normal btw :p )

Screenshot 2023-03-09 at 4 05 38 PM

[SFINT-4820]: https://coveord.atlassian.net/browse/SFINT-4820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/exampleSearch/exampleSearch.html
```diff
@@ -1,64 +1,165 @@
...
+              ></c-quantic-query-error>
+              <c-quantic-breadcrumb-manager
+                engine-id={engineId}
+              ></c-quantic-breadcrumb-manager>
+              <c-quantic-did-you-mean
+                engine-id={engineId}
+              ></c-quantic-did-you-mean>
               <c-quantic-no-results engine-id={engineId}></c-quantic-no-results>
-              <c-quantic-result-list engine-id={engineId}></c-quantic-result-list>
+              <div class="slds-var-m-vertical_small">
```
mmitiche, line 132: Why the margin is only around the QuanticSmartSnippet component? it's possible that a query returns some  `smart snippet suggestions` but no `smart snippet`, in that case maybe we need margin around the QuanticSmartSnippetSuggestions as well.
It would be nice to see this situation as well in a screenshot.

packages/quantic/force-app/main/default/lwc/exampleSearch/exampleSearch.html
```diff
@@ -1,64 +1,167 @@
...
+                <c-quantic-smart-snippet
+                  engine-id={engineId}
+                  maximum-snippet-height="250"
```
mmitiche, line 135: Why do we need to set this property?

SimonMilord: Oh I believe I tested the collapsible feature with a lower value initially but then set it to default instead of removing it... Will remove it :)

mmitiche (approved):

nathanlb (commented):
Most of this is formatting. Are you sure your IDE is configured with the right config?

nathanlb:
Mine is also "on save" and it doesn't reformat the html. Maybe it's mine that's poorly configured. @mmitiche does yours reformat like this? 😅

mmitiche:
> Mine is also "on save" and it doesn't reformat the html. Maybe it's mine that's poorly configured. @mmitiche does yours reformat like this? 😅


The HTML should be ignored because of this line: https://github.com/coveo/ui-kit/blob/59deaa78a4b082daec45c19fcfe501edde4229b5/.prettierignore#L2
On my side when the whole UI-KIT repo is opened in vs code, the html files are not reformatted, but when only the Quantic sub folder of UI-KIT is opened in vs code   the HTML is formatted like this yeah.
Personally I don't like how prettier is formatting the HTML, it breaks a lot of lines x)

nathanlb (approved):

nathanlb:
> > Mine is also "on save" and it doesn't reformat the html. Maybe it's mine that's poorly configured. @mmitiche does yours reformat like this? 😅
> 
> The HTML should be ignored because of this line:
> 
> https://github.com/coveo/ui-kit/blob/59deaa78a4b082daec45c19fcfe501edde4229b5/.prettierignore#L2
> 
> 
> On my side when the whole UI-KIT repo is opened in vs code, the html files are not reformatted, but when only the Quantic sub folder of UI-KIT is opened in vs code the HTML is formatted like this yeah.
> Personally I don't like how prettier is formatting the HTML, it breaks a lot of lines x)

I agree that i prefer the original formatting but we shouldn't rely on prettier failing when the monorepo is open 😅. If we agree to keep the other formatting we should indeed alter the prettier config... however looks like the current config uses the lwc parser which would technically be the recommended formatting
```
  overrides: [
      {
        files: '**/lwc/**/*.html',
        options: {parser: 'lwc'},
      },
  ],
```
TBH I would go with what we prefer.

PR #2771, opened 2023-03-09
test(headless): fix failing tests for headless field suggestions

olamothe:
Most probably some content changed in the index.

https://coveord.atlassian.net/browse/KIT-2347


louis-bompart (approved):
Perfect timing, was about to investigate on #2764. LGTM!

PR #2772, opened 2023-03-10
fix(quantic): fixed the sizing of the smartSnippet headings to match suggestions

SimonMilord:
[SFINT-4824](https://coveord.atlassian.net/browse/SFINT-4824)

Before:

image

After:
Screenshot 2023-03-10 at 11 48 59 AM

[SFINT-4824]: https://coveord.atlassian.net/browse/SFINT-4824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticSmartSnippetSuggestions/quanticSmartSnippetSuggestions.html
```diff
@@ -1,22 +1,38 @@
...
         <template for:each={suggestions} for:item="suggestion">
-          <lightning-accordion-section data-id={suggestion.questionAnswerId} key={suggestion.questionAnswerId}
-            name={suggestion.question} label={suggestion.question} onclick={handleSuggestionClick}>
+          <lightning-accordion-section
+            data-id={suggestion.questionAnswerId}
+            key={suggestion.questionAnswerId}
+            name={suggestion.question}
+            label={suggestion.question}
+            onclick={handleSuggestionClick}
+            class="accordion__summary-heading"
```
SimonMilord, line 20: added CSS class for styling hook here

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetSuggestions/quanticSmartSnippetSuggestions.css
```diff
@@ -1,3 +1,7 @@
 .smart-snippet-suggestions__title {
   font-weight: 700;
 }
+
+.accordion__summary-heading {
+  --slds-c-accordion-heading-font-size: 1.1em;
```
mmitiche, line 6: We usually prefer REM in quantic.
```suggestion
  --slds-c-accordion-heading-font-size: 1.1rem;
```

mmitiche (approved):
LGTM!
Just switch EM to REM 👌

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetSuggestions/quanticSmartSnippetSuggestions.css
SimonMilord: noted ! :)

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetSuggestions/quanticSmartSnippetSuggestions.css
```diff
@@ -1,3 +1,7 @@
 .smart-snippet-suggestions__title {
   font-weight: 700;
 }
+
+.accordion__summary-heading {
+  --slds-c-accordion-heading-font-size: 1.1rem;
```
erocheleau, line 6: I'm wondering if instead of overriding this to a `1.1rem` value, we could instead set a css class on the container to influence the inner font-size?

For example by adding `class="slds-text-body_small"` maybe to the `lightning-accordion-section` maybe the content of it would be more aligned with the quantic text size?

erocheleau (commented):
Please make sure it's the only way to set the font-size for this component before adding a "magic" 1.1 rem value somewhere.

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetSuggestions/quanticSmartSnippetSuggestions.css
SimonMilord: yes I did think of using the slds class on the lightning-accordion-section, but it doesnt target the header of the section, just the inside content text. The current solution was the only way of targeting the header I could find.

erocheleau (approved):

PR #2778, opened 2023-03-13
fix(atomic-react): fixed IIFE export

btaillon:
https://coveord.atlassian.net/browse/KIT-2352

The IIFE export from `@coveo/atomic-react`, when added to a script tag, would give an error which said that `require` is undefined.

The error was caused by Rollup not identifying Headless, `react-dom/client` and `react-dom/server` as external dependencies, which caused Rollup to try to bundle them with `atomic-react`.

# Changes
## Exclude Headless and ReactDOM from bundle
I excluded `@coveo/headless`, `react-dom/client` and `react-dom/server` from the bundle, since in our doc, we already required users to include them in their page anyway.

Prior to these changes, `atomic-react.js` contained 132,773 lines (1,552 KB minified). After these changes, `atomic-react.js` contains 493 lines (12 KB minified).

## Add IIFE samples project
I added a new IIFE samples project so that we can test our IIFE exports.

This project, instead of reading dependencies via `require`/`import`, copies the CDN assets and uses them directly. 

I decided to make it a new project instead of adding it to existing projects (like atomic-react), since our other sample projects generally use bundlers. For any IIFE page we do now or in the future, I wanted to avoid using any bundlers, since they may import modules or bundle them differently than just vanilla script tags.

## Add smoke test
I added a smoke test so that we can ensure AtomicReact keeps working.

## Fixed CDN error
Specifically for major versions of AtomicReact, we were publishing them to Atomic's CDN instead of AtomicReact's. I fixed the typo that caused this.

## Made every package use the same Cypress version
~~When I created the new project, I noticed I got an error in the command line and in the CI which says that verification timed out.~~

~~The cause is apparently because Cypress initially does a verification when it's first installed, and it tends to fail when re-using an NPM cache or when the timeout is just too small.~~

~~The reason why Cypress was running this verification for me is because the version matcher allowed newer versions of Cypress to be installed, and NPM installed a newer version than the rest of the repo. For now, I decided to make every package use the same Cypress version. This should also save installation time when we do `npm i`.~~

Edit: I'm undoing this change, since newer versions of Cypress use a different configuration file and directory structure than older versions, and our projects don't all match the same structure. For now, I'll just try to minimize the amount of different versions.


packages/atomic-react/README.md
```diff
@@ -295,90 +296,94 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+  '@types/react-dom': '18.0.6',
+  ncp: '2.0.0',
+  react: '18.2.0',
+  'react-dom': '18.2.0',
+  rollup: '^2.79.0',
+  'rollup-plugin-polyfill-node': '^0.10.2',
+  'rollup-plugin-terser': '^7.0.2'
+}
+peerDependencies = {
+  '@coveo/headless': '^2.0.0',
```
btaillon, line 325: **Caveat**: currently, our `peerDependencies` indicate that any version of Headless would work, but in reality, only `2.8.10` would work. This will be fixed with the new versioning process.

packages/atomic-react/README.md
```diff
@@ -295,90 +296,94 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+    crossorigin
+    src="https://unpkg.com/react@18/umd/react.production.min.js"
+  ></script>
+  <script
+    crossorigin
+    src="https://unpkg.com/react-dom@18/umd/react-dom.production.min.js"
+  ></script>
+  <script
+    crossorigin
+    src="https://unpkg.com/react-dom@18/umd/react-dom-server-legacy.browser.production.min.js"
```
btaillon, line 353: **Caveat**: `react-dom/umd/react-dom-server.browser.production.min.js` does not include the function we use in AtomicReact, which is `renderToString`. `react-dom/server.browser.js` does include everything we need, but it imports them via `require`, which throws an error.

packages/atomic-react/README.md
```diff
@@ -295,90 +296,94 @@ import {AtomicSearchInterface} from '@coveo/atomic-react';
...
+    crossorigin
+    src="https://static.cloud.coveo.com/atomic-react/v2.1/iife/atomic-react.min.js"
+  ></script>
 </head>
 ```
 
-Once this is done, you can start using `CoveoAtomicReact` directly with an inline script tag: 
+Once this is done, you can start using `CoveoAtomicReact` directly with an inline script tag:
 
 ```html
```
btaillon, line 388: I made some changes to shorten this sample a little bit, I can revert them if you prefer.

packages/atomic-react/rollup.config.mjs
```diff
@@ -1,31 +1,46 @@
...
+import {defineConfig} from 'rollup';
 import nodePolyfills from 'rollup-plugin-polyfill-node';
 import {terser} from 'rollup-plugin-terser';
 
+/** @type {import("rollup").GlobalsOption} */
+const globals = {
+  react: 'React',
+  'react-dom': 'ReactDOM',
+  'react-dom/client': 'ReactDOM',
+  'react-dom/server': 'ReactDOMServer',
```
btaillon, line 14: Rollup interprets `react-dom`, `react-dom/client` and `react-dom/server` as being completely different packages.

packages/samples/iife/build-assets.mjs
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ *
+ * @param {readonly V[]} values
+ * @param {(value: V, index: number) => K} getKey
+ * @template V
+ * @template {PropertyKey} K
+ * @returns {Record<K, V[]>}
+ */
+function aggregate(values, getKey) {
```
btaillon, line 18: Copied from Atomic's utils.

packages/samples/iife/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@coveo/iife-samples",
```
olamothe, line 2: Should we add something in scripts:deploy -> do nothing ?

So that we don't deploy any of those sample packages

packages/samples/iife/build-assets.mjs
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * @typedef CDNAsset
+ * @property {string} sourceDirectory Directory containing the required assets, relative to the root of the workspace.
+ * @property {string[]} relativeDestinations Where the directory needs to be accessible from. For example: `atomic/storybook`.
+ */
+
+/**
+ * @returns {CDNAsset[]}
+ */
+function getDeploymentPipelineAssets() {
```
btaillon, line 38: I decided to use the deployment pipeline config as a source of truth as to what gets exported to the CDN. This should help us catch errors.

olamothe (approved):

packages/samples/iife/package.json
btaillon: no need, it's already marked as `private`, and it doesn't contain a publish script.

packages/samples/iife/www/atomic-react.html
```diff
@@ -0,0 +1,255 @@
...
+    <script crossorigin type="module" src="/cdn/atomic/atomic.esm.js"></script>
+    <link rel="stylesheet" href="/cdn/atomic/themes/coveo.css" />
+    <script crossorigin src="/cdn/atomic-react/iife/atomic-react.min.js"></script>
+
+    <script crossorigin src="/cdn/deps/@babel/standalone/babel.min.js"></script>
+    <title>Atomic React test page</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="text/babel" data-plugins="transform-react-jsx">
```
btaillon, line 22: I decided to use babel here for two reasons.

1. It allows me to copy and paste the entire NextJS sample almost as-is, since it transforms JSX to vanilla JS.
2. It's what we were already recommending in our doc, and I think it's a good idea.

louis-bompart (approved):

PR #2785, opened 2023-03-16
fixed <path> closing tags for local server

SimonMilord:
[SFINT-4932](https://coveord.atlassian.net/browse/SFINT-4932)

Fixed the <path> closing tag causing error when starting the local dev server. 
replaced : `<path />` by : `<path></path>`

Reducing the need to fix it every time we start the local dev server

[SFINT-4932]: https://coveord.atlassian.net/browse/SFINT-4932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mmitiche:
The changes look good! However I'm bit worried about formatting of the HTML files with prettier. Cause for now the majority of the HTML files in Quantic is not formatted with prettier and now we are starting to format some and we really do not want to end up with a repo where half the HTML is formatted with method A and the other i formatted with method B.

That's what I'd suggest:
- Remove the formatting changes from this PR.
- Open a new Jira  ticket specific to formatting the HTML in Quantic. This ticket would include having a small discussion or just a draft PR shared with the team to really decide as a team which formatting method is the one we we wanna follow and then we implement it in every HTML file.

What do you think @nathanlb ?

SimonMilord:
> The changes look good! However I'm bit worried about formatting of the HTML files with prettier. Cause for now the majority of the HTML files in Quantic is not formatted with prettier and now we are starting to format some and we really do not want to end up with a repo where half the HTML is formatted with method A and the other i formatted with method B.
> 
> That's what I'd suggest:
> 
> * Remove the formatting changes from this PR.
> * Open a new Jira  ticket specific to formatting the HTML in Quantic. This ticket would include having a small discussion or just a draft PR shared with the team to really decide as a team which formatting method is the one we we wanna follow and then we implement it in every HTML file.
> 
> What do you think @nathanlb ?

I removed the formatting from the PR and created the [JIRA](https://coveord.atlassian.net/browse/SFINT-4935)

maybe we can slot it in the next sprint planning or something

erocheleau:
> > The changes look good! However I'm bit worried about formatting of the HTML files with prettier. Cause for now the majority of the HTML files in Quantic is not formatted with prettier and now we are starting to format some and we really do not want to end up with a repo where half the HTML is formatted with method A and the other i formatted with method B.
> > That's what I'd suggest:
> > 
> > * Remove the formatting changes from this PR.
> > * Open a new Jira  ticket specific to formatting the HTML in Quantic. This ticket would include having a small discussion or just a draft PR shared with the team to really decide as a team which formatting method is the one we we wanna follow and then we implement it in every HTML file.
> > 
> > What do you think @nathanlb ?
> 
> I removed the formatting from the PR and created the [JIRA](https://coveord.atlassian.net/browse/SFINT-4935)
> 
> maybe we can slot it in the next sprint planning or something

Totally agree on my side. Even if I am not a huge fan of pre-commit hooks with prettier, it's one OR the other and not half hehe :) 

Also I believe there is a specific prettier module for LWC

erocheleau (approved):

PR #2786, opened 2023-03-17
feat(quantic): quantic folded result list created

mmitiche:
[SFINT-4845](https://coveord.atlassian.net/browse/SFINT-4845)

This PR contains:

- The Quantic Folded Result List component.
- An implementation of the folding in the ExampleSearch component.
- E2E to cover the folding feature.

Demo:

https://user-images.githubusercontent.com/86681870/225882818-ca0425f6-79df-4097-9fc4-294fc43dcb3a.mov

[SFINT-4845]: https://coveord.atlassian.net/browse/SFINT-4845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/cypress/e2e/default-2/result-list/folded-result-list/folded-result-list.cypress.ts
```diff
@@ -0,0 +1,180 @@
...
+      });
+
+      scope('when showing less results', () => {
+        Actions.toggleMoreChidResults();
+        Expect.resultsEqual(resultHierarchy);
+        Expect.logShowLessFoldedResults();
+      });
+    });
+
+    describe('when no child result has been found after loading all collections', () => {
```
mmitiche, line 114:
```suggestion
    describe('when no child result has been found after loading all the collection', () => {
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticFoldedResultList/exampleQuanticFoldedResultList.html
```diff
@@ -0,0 +1,52 @@
...
+        <c-action-perform-search
+          slot="actions"
+          disabled={notConfigured}
+          engine-id={engineId}
+        ></c-action-perform-search>
+        <c-action-next-results
+          slot="actions"
+          disabled={notConfigured}
+          engine-id={engineId}
+        ></c-action-next-results>
```
mmitiche, line 19:
```suggestion
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticFoldedResultList/exampleQuanticFoldedResultList.js
```diff
@@ -0,0 +1,72 @@
...
+// @ts-ignore
+import childTemplate from './resultTemplates/childTemplate.html';
+
+export default class ExampleQuanticFoldedResultList extends LightningElement {
+  @api engineId = 'quantic-folded-result-list-engine';
+  @track config = {};
+  isConfigured = false;
+
+  pageTitle = 'Quantic Folded Result List';
+  pageDescription = 'The QuanticFoldedResultList component is responsible for displaying query results by applying one or more result templates.';
```
mmitiche, line 13:
```suggestion
  pageDescription = 'The QuanticFoldedResultList component is responsible for displaying query results by applying one or more result templates. This component can display query results that have a parent-child relationship with any level of nesting.';
```

packages/quantic/force-app/main/default/lwc/quanticFoldedResultList/quanticFoldedResultList.js
```diff
@@ -0,0 +1,158 @@
...
+   */
+  @api fieldsToInclude =
+    'date,author,source,language,filetype,parents,sfknowledgearticleid';
+  /**
+   * The name of the field on which to do the folding.
+   * @api
+   * @type {string}
+   * @defaultValue `'foldingcollection'`
+   */
+  @api collectionField;
```
lbergeron, line 43: Shouldn't `collectionField` be initialized to `foldingcollection`? The same goes for `parentField`, `childField`, and `numberOfFoldedResults`.

packages/quantic/force-app/main/default/lwc/quanticResultTemplate/quanticResultTemplate.js
```diff
@@ -30,6 +30,13 @@ export default class QuanticResultTemplate extends LightningElement {
    */
   @api resultPreviewShouldNotBeAccessible = false;
   
+  /**
+   * Indicates whether the bottom border of the result template should be hidden..
+   * @api
+   * @type {boolean}
+   */
+  @api hideBottomBorder = false;
```
lbergeron, line 38: This property is tightly coupled with the child result templates. Later on, the component styling might change in a way that makes removing the border unnecessary. Setting this property would then break the component styling.

Instead of exposing a property like this, what would you think of inferring the value by checking if a component is configured in the `children` slot?

lbergeron (commented):
That's very nice! Well done!

packages/quantic/cypress/page-objects/search.ts
```diff
@@ -196,14 +198,18 @@ export function mockSearchNoResults(useCase?: string) {
...
-      ];
-      res.body.totalCount = 1;
-      res.body.totalCountFiltered = 1;
+      if (!results) {
+        res.body.results = [
+          {title: 'Result', uri: 'uri', raw: {urihash: 'resulthash'}},
+        ];
+      } else {
+        res.body.results = results;
+      }
```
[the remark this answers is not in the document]
mmitiche: way better thanks👌

packages/quantic/force-app/main/default/lwc/quanticFoldedResultList/quanticFoldedResultList.js
mmitiche: didn't give these properties any initial values because the default values are used when these properties are `undefined`, but yeah you are right it would be safer to set these initial values directly in Quantic, I will add them.

packages/quantic/force-app/main/default/lwc/quanticFoldedResultList/quanticFoldedResultList.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  updateResultPerPage() {
+    this.numberOfResults = this.resultsPerPage.state.numberOfResults;
+  }
+
+  get fields() {
+    if (this.fieldsToInclude.trim() === '') return [];
+    return this.fieldsToInclude.split(',').map((field) => field.trim());
+  }
```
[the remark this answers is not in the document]
mmitiche: sure!

packages/quantic/force-app/main/default/lwc/quanticResultTemplate/quanticResultTemplate.js
mmitiche: yeah I agree, wasn't even confident about this property, not sure tho how we can detect if a template is configured in the `children` slot.

One solution I thought about is avoid adding this `hideBottomBorder` property and completely remove the border bottom of the QuanticResultTemplate component.
Then if we want to have a border bottom we would only need to do the following:
```
<div class="slds-border-bottom">
   <c-quantic-result-template>....</c-quantic-result-template>
</div>
```
But I think this solution also  falls in the breaking change category.

lbergeron: A named slot elements can be accessed like this (see [LWC doc](https://lwc.dev/guide/composition#access-elements-passed-via-slots)):

```javascript
const childTemplates = this.querySelectorAll('*[slot="children"]');
```

In your case, you could have something like this:
```javascript
get hasChildTemplates() {
  return this.querySelectorAll('*[slot="children"]').length > 0;
  // or use `querySelector` to avoid retrieving all elements
  return !!this.querySelector('*[slot="children"]');
}
```

packages/quantic/force-app/main/default/lwc/quanticFoldedResultList/quanticFoldedResultList.js
```diff
@@ -0,0 +1,161 @@
...
+    this.numberOfResults = this.resultsPerPage.state.numberOfResults;
+  }
+
+  get fields() {
+    if (this.fieldsToInclude.trim() === '') return [];
+    return this.fieldsToInclude
+      .split(',')
+      .map((field) => field.trim())
+      .filter((field) => field.length > 0);
+  }
```
lbergeron, line 156: The advantage of filtering the empty fields at the end is that the separate `if` is not necessary anymore.

```suggestion
  get fields() {
    return this.fieldsToInclude
      .split(',')
      .map((field) => field.trim())
      .filter((field) => field.length > 0);
  }
```

packages/quantic/force-app/main/default/lwc/quanticResultList/quanticResultList.js
```diff
@@ -141,7 +141,10 @@ export default class QuanticResultList extends LightningElement {
 
   get fields() {
     if (this.fieldsToInclude.trim() === '') return [];
-    return this.fieldsToInclude.split(',').map((field) => field.trim());
+    return this.fieldsToInclude
+      .split(',')
+      .map((field) => field.trim())
+      .filter((field) => field.length > 0);
   }
```
lbergeron, line 148: The `if` is not necessary anymore.

```suggestion
get fields() {
    return this.fieldsToInclude
      .split(',')
      .map((field) => field.trim())
      .filter((field) => field.length > 0);
  }
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticFoldedResultList/resultTemplates/childTemplate.html
```diff
@@ -0,0 +1,20 @@
+<template>
+  <c-quantic-result-template is-any-preview-open={isAnyPreviewOpen} hide-bottom-border
```
lbergeron, line 2: The `hide-bottom-border` attribute should not be set anymore.

```suggestion
  <c-quantic-result-template is-any-preview-open={isAnyPreviewOpen}
```

packages/quantic/force-app/main/default/lwc/exampleSearch/resultTemplates/childTemplate.html
```diff
@@ -0,0 +1,29 @@
+<template>
+  <c-quantic-result-template is-any-preview-open={isAnyPreviewOpen} hide-bottom-border result-preview-should-not-be-accessible={resultPreviewShouldNotBeAccessible}>
```
lbergeron, line 2: Remove the `hide-bottom-border` attribute.

```suggestion
  <c-quantic-result-template is-any-preview-open={isAnyPreviewOpen} result-preview-should-not-be-accessible={resultPreviewShouldNotBeAccessible}>
```

lbergeron (approved):
Spotted a few things to fix, but everything else looks good to me!

lvu285:
Nice job done!
Only small issue related to Accessibility, when use keyboard to click "Load more result", the focus of the mouse I think should be at the top of the folding again, as right now it's still at the "Hide items" button

mmitiche:
> Nice job done! Only small issue related to Accessibility, when use keyboard to click "Load more result", the focus of the mouse I think should be at the top of the folding again, as right now it's still at the "Hide items" button

Thanks @lvu285!
For the accessibility thing I'm opening a new ticket to address it, in the same ticket we will try to address another a11y concern related to aria live.

PR #2787, opened 2023-03-17
feat: recommendations error component

pathiery:
[SVCINT-2174](https://coveord.atlassian.net/browse/SVCINT-2174)

This PR adds an `atomic-recs-error` component, which is bound to the `recommendations` field in the stencil store.
It work very similarly to `atomic-query-error` and can be used to display an error screen when recommendations fail to load.

image

[SVCINT-2174]: https://coveord.atlassian.net/browse/SVCINT-2174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/recommendations/atomic-recs-error/atomic-recs-error.tsx
```diff
@@ -0,0 +1,64 @@
+// import {QueryError, QueryErrorState, buildQueryError} from '@coveo/headless';
+import {
+  buildRecommendationList,
+  RecommendationList,
+  RecommendationListState,
+} from '@coveo/headless/recommendation';
+import {Component, h, State} from '@stencil/core';
+import {AriaLiveRegion} from '../../../utils/accessibility-utils';
+// import {AriaLiveRegion} from '../../../utils/accessibility-utils';
```
btaillon, line 9: You left some comments here and at line 1

olamothe (approved):
It would be a good idea to create the accompanying storybook file.

This will ensure that the proper documentation appear on docs.coveo.com/atomic under: https://docs.coveo.com/en/atomic/latest/reference/recommendation-components/

With an interactive sample.

You should be able to pretty much just copy `atomic-query-error.stories.tsx` and adapt it by using `recsStory` utility function instead of `defaultStory` (check in `atomic-recs-list-stories.tsx`), and the proper tagname `atomic-recs-error`.

PR #2788, opened 2023-03-17
chore(atomic): isolated storybook dependencies (#2788

btaillon:
https://coveord.atlassian.net/browse/KIT-2363

# The problem
Storybook being a dependency of Atomic caused multiple issues when re-generating the lockfile:
* Some Storybook packages are getting hoisted and some don't, which causes some packages to become unable to find their peers and fail the build.
* Storybook requires React v16, both for `addon-docs` and for some deeper dependencies. This causes React v16 to be hoisted to the root of the repo, which in turn causes other issues. Simply adding React v18 as a dependency at the root of the repo wouldn't work due to peer dependencies errors with Storybook.

# The solution
I tried several fixes to the above problems, and so far, this is the one that works best (shout out to @louis-bompart for coming up with it!).

In this PR, I'm separating the Storybook dependencies from Atomic by creating a new package that contains the Storybook dependencies.

This new package:
* Is called `@coveo/atomic-storybook`.
* Is private.
  * We don't want Lerna to bump its version.
* Is isolated.
  * Meaning it's **not** part of the workspaces, and requires its own `npm install`, which I added to the root `package.json`'s `postinstall`.
* Depends on `@coveo/atomic`.
* Contains the former `build:storybook` and `dev:storybook` scripts (now named `build` and `dev`).
* Contains the Storybook configuration (the `.storybook` directory).
* Contains the Storybook output (the `storybook-static` directory).

# Possible changes or improvements

## Adding `@coveo/atomic-storybook` to workspaces.
We could decide to add `@coveo/atomic-storybook` to workspaces, like our other packages. However, I tested this, and it requires us to prevent Storybook dependencies from getting hoisted.

We can prevent Storybook dependencies from getting hoisted by adding these dependencies to the root `package.json`:
```json
{
    "@storybook/channel-postmessage": "npm:noist@1.0.0",
    "@storybook/channels": "npm:noist@1.0.0",
    "@storybook/client-logger": "npm:noist@1.0.0",
    "@storybook/core-events": "npm:noist@1.0.0",
    "@storybook/csf": "npm:noist@1.0.0",
    "@storybook/csf-tools": "npm:noist@1.0.0",
    "@storybook/html": "npm:noist@1.0.0",
    "@storybook/node-logger": "npm:noist@1.0.0",
    "@storybook/postinstall": "npm:noist@1.0.0",
    "@storybook/semver": "npm:noist@1.0.0"
}
```

## Adding an alias for `.stories` files.
Currently, `.stories` files need to reference `atomic-storybook`'s `.storybook` directory, which causes them to have an extremely long chain of `../`. I could add a Typescript alias to remove the need for this, but it might require messing around with more than one `tsconfig` file.

## Making `@coveo/atomic` depend on `@coveo/atomic-storybook` instead.
This is the solution I originally tested, and it worked, but it required more changes than this.

Pros:
* No circular\* dependency
  * \*We have what *looks* like a circular dependency right now, since `@coveo/atomic-storybook` is responsible for building the stories inside `@coveo/atomic`, but those stories need to use helpers from `@coveo/atomic-storybook`.

Cons:
* More scripts in Atomic
  * The current solution reduces the amount of scripts defined in Atomic. It feels cleaner and better supports parallelization.
* More complicated / doesn't play well with Storybook
  * The CWD still needs to be `@coveo/atomic-storybook` in order for Storybook to work properly. Some gymnastics are required.

## Caching Storybook
I could add some inputs/outputs to the Storybook scripts so that they are cached. They take a while to run.


utils/atomic-storybook/.storybook/main.js
```diff
@@ -1,20 +1,31 @@
...
     '@storybook/addon-controls',
     '@storybook/addon-viewport',
     '@storybook/addon-a11y',
     './preset.js',
   ],
+  /** @type {import('webpack')['config']['getNormalizedWebpackOptions']} */
   webpack: (config) => {
     return {
       ...config,
+      resolve: {...config.resolve, modules: module.paths},
```
btaillon, line 28: `module.paths` here contains the `node_modules` directory of `@coveo/atomic-storybook`. Without it, stories would resolve their dependencies relative to `@coveo/atomic` instead.

louis-bompart (approved):
lgtm. I would not invest more 'QoL' on this, given that it might be essentially scrapped when the next major of storybook roll out

utils/atomic-storybook/project.json
```diff
@@ -0,0 +1,58 @@
+{
+  "targets": {
+    "install:isolated": {
+      "dependsOn": [],
+      "inputs": [],
+      "outputs": [],
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "npm i --package-lock=false",
```
btaillon, line 9: Including the lockfiles causes issues with the `commit-generated-files` actions.

louis-bompart: 👎 Keep the lockfile, and modify the action so that the second lockfile is allowed

btaillon: will do

packages/atomic/src/components/common/atomic-icon/atomic-icon.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import bgIcons from '@salesforce-ux/design-system/design-tokens/dist/bg-standard.common';
 import {html} from 'lit-html';
-import defaultStory from '../../../../.storybook/default-story';
+import defaultStory from '../../../../../../utils/atomic-storybook/.storybook/default-story';
```
olamothe, line 3: Yeah, I would like if you could check if those imports could be simplified with a TS config change.

If it's too complicated and start randomly breaking other things, then so be it.

But the import chain is a bit extreme right now 😉

olamothe (approved):

scripts/bin/install-isolated.mjs
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env node
+import {execute} from '../exec.mjs';
+
+if (!['install', 'ci'].includes(process.env.npm_command)) {
+  throw `Unexpected npm_command environment variable (value was "${process.env.npm_command}"). This script should only be run from a preinstall or postinstall script.`;
+}
+
+await execute('npm', [process.env.npm_command]);
```
louis-bompart, line 8:
```suggestion
await execa('npm', [process.env.npm_command]);
```

louis-bompart (changes requested):
`execute` plain doesn't work on windows.
(that's why we have the tidbit `appendCmdIfWindows` in the Coveo CLI)
But you should probably use `execa` here.

btaillon:
> `execute` plain doesn't work on windows. (that's why we have the tidbit `appendCmdIfWindows` in the Coveo CLI) But you should probably use `execa` here.

I'm gonna replace `spawn` in `exec.mjs` with `execa`, since I'd like to avoid having too much duplication of ways to execute commands. `exec` defaults to forwarding the shell and waiting for all streams to finish before resolving.

louis-bompart:
> > `execute` plain doesn't work on windows. (that's why we have the tidbit `appendCmdIfWindows` in the Coveo CLI) But you should probably use `execa` here.
> 
> I'm gonna replace `spawn` in `exec.mjs` with `execa`, since I'd like to avoid having too much duplication of ways to execute commands. `exec.mjs` defaults to forwarding the shell and waiting for all streams to finish before resolving.

Maybe in another PR to keep the scope tight?

btaillon:
> > > `execute` plain doesn't work on windows. (that's why we have the tidbit `appendCmdIfWindows` in the Coveo CLI) But you should probably use `execa` here.
> > 
> > 
> > I'm gonna replace `spawn` in `exec.mjs` with `execa`, since I'd like to avoid having too much duplication of ways to execute commands. `exec.mjs` defaults to forwarding the shell and waiting for all streams to finish before resolving.
> 
> Maybe in another PR to keep the scope tight?

Yeah. Anyway, it seems better to follow execa's scripts doc instead and call it directly: https://github.com/sindresorhus/execa/blob/HEAD/docs/scripts.md

louis-bompart (approved):

PR #2789, opened 2023-03-17
chore: update lockfile

btaillon:
https://coveord.atlassian.net/browse/KIT-2338

When bumping versions package versions using NPM's `npm version` command, the lockfile gets re-generated, and some packages that were hoisted are no longer hoisted, and some packages that weren't hoisted become hoisted.

Some packages seem to only look for their dependencies in the same `node_modules` directory as themselves, therefore the above behaviour causes a lot of bugs.

In this PR, I'm trying to add all preliminary changes for the version bumping commits, including:
* Re-generating the lockfile and fixing the errors that ensue.
  * Migrating to lockfile v3 (see [documentation](https://docs.npmjs.com/cli/v9/using-npm/config?v=true#lockfile-version)).
  * Upgrading the version of `node` from v16 to v18.
  * Upgrading NPM to v9.
  * Fixing a version conflict in the NextJS samples by bumping its `next` version.
  * Making Headless's `esbuild` resolve `web-encoding` relative to its own `node_modules` instead of the root `node_modules`.
* Removing the `link:angular` script.
  * Seems no longer necessary, so I'm removing it.
* Removing Snyk and Yarn
  * We're going to use a new process for security anyway. Snyk apparently doesn't play well with lockfile v3 (if I recall correctly).


packages/samples/angular/package.json
```diff
@@ -4,8 +4,7 @@
   "scripts": {
     "ng": "ng",
     "dev": "npm run build:assets && npm run dev:angular",
-    "dev:angular": "ng serve",
-    "link:angular": "node scripts/link-hoisted-dependencies.mjs",
+    "dev:angular": "ng serve --host=0.0.0.0",
```
btaillon, line 7: Angular's e2e test was timing out with an error that implied it didn't finish building within 90 seconds, despite it having actually finished within 30 seconds. Re-running it wouldn't fix the issue.

I saw a post that had a similar error for a seemingly unrelated issue, and their solution was to add `--host=0.0.0.0`. I was trying a lot of stuff in parallel, so I gave it a shot, and it worked. I'm still not quite sure what the root of the issue is, but if you guys think it's worth investigating, I can look into it.

PR #2790, opened 2023-03-17
Fix: add z-index to atomic-ipx-modal

pathiery:
[SVCINT-2156](https://coveord.atlassian.net/browse/SVCINT-2156)

There can be issues with clipping in some host websites. The solution is to set the z-index of the modal to the maximum value by default, thus letting us be ON TOP.

This makes sense since we always wanna appear on top, and host websites can choose to override these settings as they please.


dguerinCoveo (approved):
I think its a good way to fix the problem. Curious to see what the DX team thinks about it

olamothe (approved):

PR #2791, opened 2023-03-20
fix: hide footer when empty

pathiery:
[SVCINT-2179](https://coveord.atlassian.net/browse/SVCINT-2179)

This PR adds a check inside `atomic-ipx-modal` that looks for elements mounted inside the `footer` slot. If none are found, a prop is set on the `atomic-ipx-body` which hides the footer wrapper.

I tried to find a pure CSS solution, but there doesn't seem to be one... [yet](https://github.com/WICG/webcomponents/issues/701#issuecomment-368430226) 

[SVCINT-2179]: https://coveord.atlassian.net/browse/SVCINT-2179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/ipx/atomic-ipx-body/atomic-ipx-body.tsx
+++ b/packages/atomic/src/components/ipx/atomic-ipx-body/atomic-ipx-body.tsx
@@ -33,6 +33,8 @@ export class AtomicIPXBody implements InitializableComponent<AnyBindings> {
 
   @Prop() isOpen = true;
 
+  @Prop({reflect: true}) displayFooterSlot = true;
+
   public componentDidLoad() {
     const id = this.host.id || randomID('atomic-ipx-body-');
     this.host.id = id;
@@ -65,7 +67,9 @@ export class AtomicIPXBody implements InitializableComponent<AnyBindings> {
         </div>
         <footer
           part="footer-wrapper"
-          class="border-neutral border-t bg-background z-10 flex flex-col items-center w-full"
+          class={`border-neutral border-t bg-background z-10 flex flex-col items-center w-full ${
+            this.displayFooterSlot ? 'visible' : 'invisible'
+          }`}
         >
           <div part="footer" class="max-w-lg">
             <slot name="footer"></slot>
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/ipx/atomic-ipx-body/atomic-ipx-body.tsx
```diff
@@ -65,7 +67,9 @@ export class AtomicIPXBody implements InitializableComponent<AnyBindings> {
         </div>
         <footer
```
louis-bompart, line 68: Why not
```tsx
this.displayFooterSlot && <footer></footer>
```

pathiery: sure !

jbvouma (approved):

pathiery:
> Nice and simple, we just need to do it for ipx-embedded too

So hum, it's already there since we're using a script to copy the contents of the modal version into the embedded version

dguerinCoveo (approved):

mikegron (approved):

packages/atomic/src/components/ipx/atomic-ipx-embedded/atomic-ipx-embedded.tsx
```diff
@@ -37,6 +37,12 @@ export class AtomicIPXEmbedded implements InitializableComponent<AnyBindings> {
 
   @Event() animationEnded!: EventEmitter<never>;
 
+  @State() private hasFooterSlotElements = true;
+
+  public componentWillLoad(): void | Promise<void> {
+    this.hasFooterSlotElements = !!this.host.querySelector('[slot="footer"]');
```
olamothe, line 43: Could this be calculated from inside `atomic-ipx-body` instead ? I'm not sure it's possible, but it would be better since you would not need the `displayFooterSlot` attribute, it could all be handled internally in that component.

PR #2794, opened 2023-03-22
doc(quantic): added internal jsdoc tag

nathanlb:
Added an "Internal" category to be used to exclude quantic components from the generated doc.


mmitiche (approved):
nice 👌

PR #2795, opened 2023-03-22
fix(atomic): made smart snippet source title re-render when changed (#2795

btaillon:
https://coveord.atlassian.net/browse/KIT-2237

I also refactored the default snippets test logic, and I made sure that analytics tests are contained within `beforeEach` blocks because they were flaky.

### DIFF
--- a/packages/atomic/src/components/search/atomic-smart-snippet/atomic-smart-snippet-source.tsx
+++ b/packages/atomic/src/components/search/atomic-smart-snippet/atomic-smart-snippet-source.tsx
@@ -71,6 +71,7 @@ export class AtomicSmartSnippetSource implements InitializableComponent {
           <atomic-result-text
             field="title"
             default="no-title"
+            key={this.source.uniqueId}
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/search/atomic-smart-snippet/atomic-smart-snippet-source.tsx
```diff
@@ -71,6 +71,7 @@ export class AtomicSmartSnippetSource implements InitializableComponent {
           <atomic-result-text
             field="title"
             default="no-title"
+            key={this.source.uniqueId}
```
btaillon, line 74: Some logic was already added in the result list to ensure results are re-rendered when their uniqueId is changed.

I'm was a bit tempted to introduce a result provider component that would handle re-rendering, but it would be overkill for this PR.

olamothe (approved):

olamothe (approved):

PR #2796, opened 2023-03-23
fix(quantic): fixed a problem when deselecting breadcrumb values 

mmitiche:
[SFINT-4938](https://coveord.atlassian.net/browse/SFINT-4938)
## The issue: 
When you have two facet values in the Quantic Breadcrumb component, and you remove the first value, removing the second value will make the first value reappear like the following:

https://user-images.githubusercontent.com/86681870/227228447-2bdd4c9f-bc8d-4b45-95ed-ad22d472a0d7.mov

## The source of the issue:
The facet values in the Quantic Breadcrumb component were displayed in the HTML like the following:
```
<template for:each={breadcrumb.values} for:item="breadcrumbValue">
     <li key={breadcrumbValue.value}>
          <c-quantic-pill ondeselect={breadcrumbValue.deselect}></c-quantic-pill>
     </li>
</template>
```
The problem is that the key passed to the `li` tag is invalid.

the object `breadcrumbValue` has the following structure: 
```
{
    formattedValue: 'some formatted value',
    value: 
    {
         value: 'some value',
         state: 'selected',
    }
}
```
which makes `breadcrumbValue.value` an object, which is an invalid value to give for the `key` attribute.

## The solution:
- #### `breadcrumbValue.value.value` used instead of `breadcrumbValue.value`.
- #### New e2e test created to avoid facing this issue again. 

https://user-images.githubusercontent.com/86681870/227233211-18b211fa-6a2a-4b55-9f03-9fe54c521439.mov

[SFINT-4938]: https://coveord.atlassian.net/browse/SFINT-4938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/cypress/e2e/facets-2/breadcrumb-manager/breadcrumb-manager.cypress.ts
+++ b/packages/quantic/cypress/e2e/facets-2/breadcrumb-manager/breadcrumb-manager.cypress.ts
@@ -27,6 +27,7 @@ describe('quantic-breadcrumb-manager', () => {
     useCase: string;
     categoryDivider: string;
     collapseThreshold: number;
+    displayFacetValuesAs: 'link' | 'checkbox';
   }
 
   function visitBreadcrumbManager(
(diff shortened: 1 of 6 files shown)


packages/quantic/cypress/e2e/facets-2/breadcrumb-manager/breadcrumb-manager.cypress.ts
```diff
@@ -209,6 +210,41 @@ describe('quantic-breadcrumb-manager', () => {
...
+              .wait(InterceptAliases.UA.Facet.Select)
+              .then((interception) => {
+                Expect.facetBreadcrumb.firstBreadcrumbValueLabelContains(
+                  interception.request.body.customData.facetValue
+                );
+              });
+            Expect.facetBreadcrumb.displayBreadcrumb(true);
+            Expect.facetBreadcrumb.labelContains('File Type');
+            Expect.facetBreadcrumb.displayValues(true);
+            Expect.facetBreadcrumb.numberOfValues(1);
```
erocheleau, line 231: Would it make sense to make a function with these 4 expects if it's repeated multiple times?

erocheleau (approved):
Looks good to me 👍

packages/quantic/cypress/e2e/facets-2/breadcrumb-manager/breadcrumb-manager.cypress.ts
mmitiche: No not really, we are not really repeating the exact same thing, the expects are called with different parameters each time.

mmitiche:
> Looks good, great job with the e2e test. I am assuming we can't change the name of value right? value.value is not the most descriptive 😂

yeah unfortunately this is coming from the Headless state so we can't really change the name.

PR #2797, opened 2023-03-23
feat(quantic): Creation of a quanticQuickviewContent component

SimonMilord:
[SFINT-4936](https://coveord.atlassian.net/browse/SFINT-4936)

**Description**

This is a fix for the [youtube iframe bug](https://coveord.atlassian.net/browse/SFINT-4769)
Creation of a quanticQuickviewContent component in LWC. We decided to render only the youtube iframe in the content modal and not the _excerpts_ @obonneau let us know if you feel the excerpts for youtube content is necessary. We can add it later, it is just slightly more work to get it formatted properly.
 
The new quickview component handles the youtube iframes properly while still handling other types of documents the same way it had before. The use of render method provides us with the flexibility to add more templates in the future, should other similar exceptions arise.

**DEMO:**

https://user-images.githubusercontent.com/73316533/227291707-30b85045-7db3-42d4-97dc-e4a8a6f3fc9a.mov

[SFINT-4936]: https://coveord.atlassian.net/browse/SFINT-4936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.html
mmitiche, line 1: It would really help the reviewers to only see the lines that actually changed without the lines that changed duo to the formatting, especially that we decided to address the formatting of the HTML files in a different  ticket [SFINT-4935](https://coveord.atlassian.net/browse/SFINT-4935)

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.css
```diff
@@ -0,0 +1,17 @@
+@import 'c/quanticResultQuickview';
+@import 'c/quanticResultActionStyles';
+
+.iframe-wrapper {
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
```
mmitiche, line 8: These two lines can be replaced by the following SLDS CSS classes: `slds-grid` and `slds-grid_vertical`
https://www.lightningdesignsystem.com/utilities/grid/#Top-to-Bottom

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.html
```diff
@@ -1,65 +1,95 @@
...
-          <footer onclick={stopPropagation} class="slds-modal__footer slds-p-top_none slds-p-bottom_none">
-            <slot name="footer"></slot>
-          </footer>
-        </template>
+        <c-quantic-quickview-content
+          id="quickview__content-container"
+          class="quickview__content-container slds-modal__content slds-wrap"
+          result={result}
+          content-url={contentURL}
+          height="100%"
```
mmitiche, line 89: pretty sure we don't need this.
```suggestion
```

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -107,10 +117,10 @@ export default class QuanticResultQuickview extends LightningElement {
 
   connectedCallback() {
     getHeadlessEnginePromise(this.engineId)
-      .then((engine) => {
+      .then((/** @type {import("coveo").SearchEngine<any>} */ engine) => {
         this.initialize(engine);
       })
-      .catch((error) => {
+      .catch((/** @type {{ message: any; }} */ error) => {
```
mmitiche, line 123: Why are you adding this typings?  does it show you a red underline when you try to access `error.message` ?

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -174,12 +183,15 @@ export default class QuanticResultQuickview extends LightningElement {
...
 
   addRecentResult() {
-    getHeadlessEnginePromise(this.engineId).then((engine) => {
-      const {pushRecentResult} = this.headless.loadRecentResultsActions(engine);
-      engine.dispatch(
-        pushRecentResult(JSON.parse(JSON.stringify(this.result)))
-      );
-    });
+    getHeadlessEnginePromise(this.engineId).then(
+      (/** @type {{ dispatch: (arg0: any) => void; }} */ engine) => {
```
mmitiche, line 187: same remark for all the typings, I don't think they are necessary.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
```diff
@@ -3,6 +3,10 @@
 .quickview__content-container {
   overflow: auto;
   border: none;
+  display: flex;
+}
+.slds-modal__content {
+  height: 90%;
```
mmitiche, line 9: Also you can add the styles of this class to the already existing class `quickview__content-container` since both are used in the same element instead of creating a new class.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.html
```diff
@@ -1,65 +1,95 @@
...
-          class="quickview__content-container slds-modal__content slds-p-around_large slds-wrap" src={contentURL}
-          height="100%" onload={onIframeLoaded}></iframe>
-        <template if:false={isLoading}>
-          <footer onclick={stopPropagation} class="slds-modal__footer slds-p-top_none slds-p-bottom_none">
-            <slot name="footer"></slot>
-          </footer>
-        </template>
+        <c-quantic-quickview-content
+          id="quickview__content-container"
+          class="quickview__content-container slds-modal__content slds-wrap"
```
mmitiche, line 86: Are these classes really affecting the look of the `c-quantic-quickview-content` component?

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.html
```diff
@@ -0,0 +1,33 @@
...
+      <iframe
+        id="quickview__content-iframe"
+        class="quickview__content-container slds-modal__content slds-p-around_large slds-wrap"
+        src={youtubeURL}
+        height="100%"
+        width="100%"
+        onload={handleIframeLoaded}
+        allowfullscreen
+        referrerpolicy="same-origin"
+        sandbox="allow-same-origin allow-scripts allow-popups allow-top-navigation allow-top-navigation-by-user-activation allow-presentation"
```
mmitiche, line 21: Do we need all this directives to make it work?
Let's keep only what we need.

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,47 @@
...
+
+export default class QuanticQuickviewContent extends LightningElement {
+  /**
+   * The [result item](https://docs.coveo.com/en/headless/latest/reference/search/controllers/result-list/#result).
+   * @api
+   * @type {Result}
+   */
+  @api result;
+  /**
+   * contentUrl.
```
mmitiche, line 19: a better description would be appreciated.

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,47 @@
...
+// @ts-ignore
+import defaultTemplate from './quanticQuickviewDefault.html';
+// @ts-ignore
+import youtubeTemplate from './quanticQuickviewYoutube.html';
+
+/** @typedef {import("coveo").Result} Result */
+
+const allowedHost = ['https://youtube.com/'];
+
+export default class QuanticQuickviewContent extends LightningElement {
```
mmitiche, line 11: The js doc to document the component would be nice! take the other components as a reference.
Also this I think this component can be tagged as `@internal` since it's really an internal component and we don't really want it to appear in the Quantic documentation.

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.css
```diff
@@ -0,0 +1,3 @@
+@import 'c/quanticResultQuickview';
+@import 'c/quanticResultActionStyles';
```
mmitiche, line 2: why exactly do we need the result actions styles here?

mmitiche (changes requested):
Great job finding the solution 👌

I'm mainly requesting changes to address some refinement  and  toremove the new exposed properties of the QuanticResultQuickview component, cause public properties are quite dangerous. They are known for being the number one reason causing breaking changes x)

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
SimonMilord: it showed a warning, but not an error, I can remove those :P

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -79,6 +79,18 @@ export default class QuanticResultQuickview extends LightningElement {
...
+   * @api
+   * @type {String}
+   */
+  @api contentUrl;
+  /**
+   * loading state of quickview content.
+   * @api
+   * @type {Boolean}
+   */
+  @api isloading = false;
```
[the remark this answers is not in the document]
SimonMilord: you are right, I changed it to have them in the childcomponent. We dont need them here

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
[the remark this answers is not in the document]
SimonMilord: this was following a discussion with @erocheleau we had agreed on 90% for the modal content
otherwise it gives us something like this: 
image

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.html
SimonMilord: yes they are needed, for example if I remove the slds-modal__content it removes the background for the modal where the content is

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.html
SimonMilord: I removed the sandbox, it is not needed. However, I originally put the permissions because I was getting errors in the console. I was able to avoid console errors after removing the sandbox by allowing youtube as a trusted site in the CSP Trusted Sites settings of the scratch org.

https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/csp_trusted_sites.htm

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.css
SimonMilord: good catch will remove

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,47 @@
...
+  @api result;
+  /**
+   * contentUrl.
+   * @api
+   * @type {String}
+   */
+  @api contentUrl;
+
+  /** @type {Boolean} */
+  @track isLoading = true;
```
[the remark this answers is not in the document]
SimonMilord: oh you are right, it seems to work even without 😅

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewDefault.html
```diff
@@ -0,0 +1,30 @@
...
+        if:true={isLoading}
+        alternative-text="Loading"
+        size="large"
+      ></lightning-spinner>
+      <iframe
+        id="quickview__content-iframe"
+        class="quickview__content-container slds-modal__content slds-p-around_large slds-wrap"
+        src={contentURI}
+        height="100%"
+        width="100%"
```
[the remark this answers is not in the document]
SimonMilord: yeah just read a bit on this, it seems better to include in css rather than the iframe tag properties.

More flexible, better for accessibility and avoid certain issues with putting them directly in the tag (ie performance loading issues, display issues)

packages/quantic/cypress/e2e/default-2/result-quickview/result-quickview.cypress.ts
```diff
@@ -42,7 +42,7 @@ describe('quantic-result-quickview', () => {
         Expect.displayTitle(true);
         Expect.displayDate(true);
         Expect.displayContentContainer(true);
-        Expect.displaySpinner(false);
+        Expect.displaySpinner(true);
```
SimonMilord, line 45: we want the spinner to display when loading

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.css
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.iframe-wrapper iframe {
+  border: none;
+}
+
+.wrapper {
+  height: 100%;
```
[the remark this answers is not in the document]
SimonMilord: I refactored as much as I could while making sure the modal content looks good

mmitiche (changes requested):
As mentioned in the comments, the `footer` slot needs to remain in the QuanticResultQuickview component, otherwise it would be a breaking change in already existing implementations using this this component.

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,55 @@
...
+
+  /** @type {Boolean} */
+  isLoading = true;
+
+  handleIframeLoaded() {
+    this.isLoading = false;
+  }
+
+  render() {
+    if (this.result?.uniqueId.includes(allowedHost[0])) {
```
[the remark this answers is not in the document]
SimonMilord: makes sense!

packages/quantic/cypress/e2e/default-2/result-quickview/result-quickview-selectors.ts
```diff
@@ -37,6 +37,9 @@ export const ResultQuickviewSelectors: ResultQuickviewSelector = {
   contentContainer: () =>
     ResultQuickviewSelectors.get().find('.quickview__content-container'),
   spinner: () =>
-    ResultQuickviewSelectors.get().find('.quickview__spinner-container'),
+    ResultQuickviewSelectors.get()
```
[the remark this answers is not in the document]
SimonMilord: yes this also works

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -218,7 +214,20 @@ export default class QuanticResultQuickview extends LightningElement {
     });
   }
 
+  handleLoadingStateChange(event) {
```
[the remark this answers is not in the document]
mmitiche: you should have something like: 
```
this.addEventListener('renderfacet', this.handleLoadingStateChange);
```
in the connected callback of this component in order to properly listen to the event.

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,57 @@
...
+   * @type {String}
+   */
+  @api contentUrl;
+
+  /** @type {Boolean} */
+  isLoading = true;
+
+  handleIframeLoaded() {
+    this.isLoading = false;
+    const event = new CustomEvent('loadingstatechange', {detail: { isLoading: this.isLoading}});
```
mmitiche, line 38: It would be nice if we can add a unit test that confirms that the QuanticQuickviewContent component is properly dispatching this event.

Take a look at this test in the QuanticResultAction component, it can give you an idea on how we can achieve such a test: https://github.com/coveo/ui-kit/blob/758f2f0f0f5a1d6dbf94c279563c04d999cc699a/packages/quantic/force-app/main/default/lwc/quanticResultAction/__tests__/quanticResultAction.test.js#L232-L247

mmitiche: Also since we always send the value `false` in the key `isLoading` of the payload, we can reduce complexity by completely omitting this payload.

SimonMilord: yeah I removed the payload and changed the event, thats a good point

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -125,6 +125,10 @@ export default class QuanticResultQuickview extends LightningElement {
           },
         }
       );
+      this.addEventListener(
```
mmitiche, line 128: This should not be included in the  `if (this.isResultAction)` condition block.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -218,6 +226,18 @@ export default class QuanticResultQuickview extends LightningElement {
     });
   }
 
+  handleLoadingStateChange() {
+    this._isLoading = false;
+  }
```
mmitiche, line 231: Just to play it safe it would be better to stop the propagation of the event after handling it.
```suggestion
  handleLoadingStateChange(event) {
    event.stopPropagation();
    this._isLoading = false;
  }
```

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.html
```diff
@@ -0,0 +1,21 @@
+<template>
+  <div class="iframe-wrapper iframe-wrapper__youtube slds-grid slds-grid_vertical">
+    <div
+      onclick={stopPropagation}
```
mmitiche, line 4: I don't see the function `stopPropagation` defined in the JS file, we need to define it.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
SimonMilord: I must have made a mis-type here, it was meant to be under the condition

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.html
mmitiche, line 1: Now that I see the full component, I really wonder if having two templates `quanticQuickviewDefault.html` and `quanticQuickviewYoutube.html` is the best solution .
Cause After all the difference between the two templates is mainly just the attribute `src` passed to the iframe.
We can create a getter that returns the right src instead, this would simplify the component by removing one of the template and 2 of the CSS files.

We usually prefer rendering multiple template when the difference is huge between the templates.
In that case the difference is very subtle.

What do you think?

mmitiche (approved):
Nice job!
Raised one last concern about the usage of multiple templates, I'll leave that to you and the second reviewer.

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.html
SimonMilord: I think this is a valid point. However, we originally had decided to use multiple templates because of the following reasons:

- the current youtube template might change in the future where we might end up adding excerpts later on which will make the template different than the default one
- the use of the templates with the render function makes it more flexible in case we have to handle new types of results in the future which would need their own template.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.css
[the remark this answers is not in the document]
erocheleau: It's based off the existing quickview, it takes maximum 90% of the content.

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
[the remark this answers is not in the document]
erocheleau: Yea at some point they made @track decoration obsolete I think.
Feels weird to rely on it being automagic instead of specifying it though :)

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/__tests__/quanticQuickviewContent.test.js
```diff
@@ -0,0 +1,111 @@
...
+
+      setupSimulation(element, 'loadingstatechange');
+
+      const iframe = element.shadowRoot.querySelector(
+        youtubeTestCase.selectors.iframeSelector
+        );
+      iframe.dispatchEvent(new CustomEvent('load'));
+
+      expect(functionsMocks.listener).toHaveBeenCalledTimes(1);
+      expect(iframe).not.toBeNull();
```
erocheleau, line 108: I would put this expect before the listener one no? 
If the iframe is null it will for sure never call the listener, but the iframe could be non-null and never call the listener?

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,60 @@
...
+    this.isLoading = false;
+    this.dispatchEvent(new CustomEvent('loadingstatechange'));
+  }
+
+  stopPropagation(evt) {
+    evt.stopPropagation();
+  }
+
+  render() {
+    if (this.result?.uniqueId.includes(allowedHost.youtube)) {
```
erocheleau, line 46: Interesting, wouldn't have thought to use that as the check, but I think it works well :+1:

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewDefault.css
```diff
@@ -0,0 +1 @@
+@import './quanticQuickviewContent.css';
```
erocheleau, line 1: Do you really need to make it's own css file? Does it not load the `quanticQuickviewContent.css` for both templates if you don't make those specific css files?

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewYoutube.html
erocheleau: I think I would keep both templates too, but if possible remove the 2 specific CSS files?

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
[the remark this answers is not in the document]
erocheleau: Yea, also, I think I would call it `doneloading`.

erocheleau (commented):
Looking good, small comments otherwise this is really close to :go:

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,60 @@
...
+/** @typedef {import("coveo").Result} Result */
+
+const allowedHost = {youtube: 'https://youtube.com/'};
+
+/**
+ * The `QuanticQuickviewContent` component renders a content template based on the type of the content (Youtube, other).
+ * @category Result Template
+ * @example
+ * <c-quantic-quickview-content result={result} content-url={contentURL}></c-quantic-quickview-content>
+ * @internal
```
nathanlb, line 17: While this is actually pretty cool and the doc generator could be configured to use a tag like this instead, currently the way to omit a component from the reference doc JSON is to set `@category Internal`

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
```diff
@@ -0,0 +1,60 @@
...
+import youtubeTemplate from './quanticQuickviewYoutube.html';
+
+
+/** @typedef {import("coveo").Result} Result */
+
+const allowedHost = {youtube: 'https://youtube.com/'};
+
+/**
+ * The `QuanticQuickviewContent` component renders a content template based on the type of the content (Youtube, other).
+ * @category Result Template
```
nathanlb, line 14:
```suggestion
 * @category Internal
```

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewDefault.css
SimonMilord: I do unfortunately. LWC doesnt support the import of CSS files in js and I didn't want to add inline styles. Salesforce links the CSS files with corresponding HTML files. Its seems the best solution for now is 1 shared css file with imports in name-specific css files. Unless you know another way?

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/__tests__/quanticQuickviewContent.test.js
```diff
@@ -0,0 +1,111 @@
...
+    result: {
+      uniqueId:
+        '42.38148$https://community.fitbit.com/board:charge/thread:5332832/message:5333023',
+    },
+    contentUrl: contentURLMock,
+  },
+  selectors: {
+    templateSelector: '.iframe-wrapper__default',
+  },
+};
```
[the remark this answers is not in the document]
SimonMilord: good point, ill add JSON files

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewDefault.css
nathanlb: I'm also surprised that each .html template needs its own .css 🤔 but I haven't checked so if you say it's necessary so maybe it is.

nathanlb (approved):
Maybe @erocheleau wants to finish his review before you merge but LGTM

packages/quantic/force-app/main/default/lwc/quanticQuickviewContent/quanticQuickviewContent.js
nathanlb: You can remove this now that the @category has been updated.

erocheleau (approved):
Looks good, just to remove the `@internal` in the jsdoc since you have the `@category Internal` :)

PR #2799, opened 2023-03-28
feat(headless,atomic): add organizationEndpoints configuration option for headless

olamothe:
Second take on the organization endpoints support (formerly known as custom DNS).

# Headless changes

- Add a new configuration option on all engines, called `organizationEndpoints`.

- Is it an object, which contains 
    `{platform: string, analytics: string}` for the core engine, and `{search: string}` for the search engine.

- It is optional. When the object is left undefined, Headless will log a warning to nudge people to opt in the feature.

- `platformUrl` configuration option on the core engine is now marked deprecated. It will continue to work as before for the current major version (2). When it is used however, Headless will log a warning nudging people to migrate to `organizationEndpoints` instead.

- In the next major version, we can -> Default `organizationEndpoints` to what `getOrganizationEndpoints` returns for production, and remove `platformUrl`.

- Values for `platform`, `analytics`, `search` can be set individually, or with a `getOrganizationEndpoints` utility function. When set individually, this would allow an implementer to use a proxy if they desire to do so. They could route only search requests in their proxy, and not proxy any UA request, if need be.

- The `getOrganizationEndpoints` utility function accept 2 params: The organizationId, and the environment (which defaults to `prod`. It returns an object that match the `organizationEndpoints` configuration on Headless engine.

# Atomic changes

- Re-expose the `getOrganizationEndpoints` utility function from Headless on the search interface. This is just a quality of life improvement for implementers, as they don't necessary want to have to imports from Headless directly in their search page.

# What it looks like from an implementer POV with Headless

```typescript
// Simple search engine in Headless, hooked to prod platform

const mySearchEngine = buildSearchEngine({
    configuration: {
         accessToken: '12345',
         organizationId: 'foo',
         organizationEndpoints: getOrganizationEndpoints('foo')
     }
})
```

```typescript
// Simple search engine in Headless, hooked to hipaa platform

const mySearchEngine = buildSearchEngine({
    configuration: {
         accessToken: '12345',
         organizationId: 'foo',
         organizationEndpoints: getOrganizationEndpoints('foo', 'hipaa')
     }
})
```

```typescript
// Simple search engine in Headless, with all search request going through a custom proxy,
// and other requests going directly to Coveo platform

const mySearchEngine = buildRecommendationEngine({
    configuration: {
         accessToken: '12345',
         organizationId: 'foo',
         organizationEndpoints: {
             ...getOrganizationEndpoints('foo'),
             search: 'https://myproxy.com'
         }
     }
})
```

# What it looks like from an implementer POV with Atomic

```typescript
// Search interface re-expose the organizationEndpoints utility function for ease of use
// All the above variation would still work.

const searchInterface = document.querySelector('atomic-search-interface');

// Initialization
  accessToken: '12345',
  organizationId: 'foo',
  organizationEndpoints: await searchInterface.getOrganizationEndpoints('foo'),
});
```

# What it looks like when you get warnings: 

```typescript
// This is what implementers are currently using in Headless v2
// This will send request to `https://platform.cloud.coveo.com`
// This config will now log a warning: 'Consider using organizationEndpoints [...]'
// This warning will ultimately contain a link to an article on docs.coveo.com

const mySearchEngine = buildSearchEngine({
    configuration: {
         accessToken: '12345',
         organizationId: 'foo',
     }
})
```

```typescript
// This is what implementers are currently using in Headless v2 when they want to use a proxy
// This will send request to `https://myproxy.com`
// This config will now log a warning: 'platformUrl is deprecated [...] Consider using organizationEndpoints [...]'
// This warning will ultimately contain a link to an article on docs.coveo.com

const mySearchEngine = buildSearchEngine({
    configuration: {
         accessToken: '12345',
         organizationId: 'foo',
         platformUrl: 'https://myproxy.com'
     }
})
```

https://coveord.atlassian.net/browse/KIT-2262


packages/headless/src/api/platform-client.ts
```diff
@@ -134,6 +127,32 @@ function coveoCloudURL<E extends PlatformEnvironment>(
   return `https://${subdomain}${urlEnv}${urlRegion}.cloud.coveo.com`;
 }
 
+/**
+ * Returns the unique endpoint(s) for a given organization identifier.
+ * @param orgId The organization identifier.
+ * @param env Optional. The environment (prod, hipaa, staging, dev) that the organization belongs to. Defaults to `prod`.
+ * @returns
```
jpmarceau, line 134: Won't there always be three endpoints? Namely, platform, analytics and search?
```suggestion
 * Returns the unique platform, analytics and search endpoints for a given organization identifier.
 * @param orgId The organization identifier.
 * @param env Optional. The environment (prod, hipaa, staging, dev) that the organization belongs to. Defaults to `prod`.
 * @returns The unique platform, analytics and search endpoints for a given organization identifier.
```

olamothe: I prefer not to be specific with the object returned.

It's possible that we will need to return different endpoints down the line. 
We can add new keys without breaking the contract.

packages/atomic/src/components/recommendations/atomic-recs-interface/atomic-recs-interface.tsx
```diff
@@ -203,6 +205,13 @@ export class AtomicRecsInterface
     );
   }
 
+  @Method() public async getOrganizationEndpoints(
```
jpmarceau, line 208: I think we need this for the atomic-search-interface docs:
```suggestion
  /**
   * Returns the unique, organization-specific endpoint(s) using options for `orgId` (required) and `env` (optional — possible values:`prod`, `hipaa`, `staging`,`dev`. Defaults to `prod`).
   */
  @Method() public async getOrganizationEndpoints(
```

jpmarceau: Also, I think we need to modify line 157/159 for `initialize`:

Initializes the connection with the headless search engine using options for `accessToken` (required), `organizationId` (required), `renewAccessToken`, `organizationEndpoints` (recommended), and `platformUrl` (deprecated).

packages/headless/src/app/engine-configuration.ts
```diff
@@ -9,12 +9,37 @@ import {
...
+ *
+ * For example https://orgid.org.coveo.com
+ *
+ * The getOrganizationEndpoints() helper function can be useful to create the appropriate object
+ */
+export interface CoreEngineOrganizationEndpoints {
+  /**
+   * The base platform endpoint.
+   *
+   * For example, https://{orgid}.org.coveo.com
```
jpmarceau, line 26:
```suggestion
   * For example: `https://{orgid}.org.coveo.com`
```

packages/headless/src/app/engine-configuration.ts
```diff
@@ -9,12 +9,37 @@ import {
...
+   */
+  platform: string;
+  /**
+   * The base analytics service endpoint.
+   *
+   * For example, https://{orgid}.analytics.org.coveo.com
```
jpmarceau, line 32:
```suggestion
   * For example: `https://{orgid}.analytics.org.coveo.com`
```

jpmarceau (approved):
I'll adjust the docs 👍

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -338,6 +340,13 @@ export class AtomicSearchInterface
     this.engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
   }
 
+  @Method() public async getOrganizationEndpoints(
+    organizationId: string,
+    env: PlatformEnvironment = 'prod'
+  ) {
+    return getOrganizationEndpointsHeadless(organizationId, env);
+  }
```
btaillon, line 348: This utility neither reads something from the search interface nor writes anything to it, so I'm not sure if we should expose it through the search interface like this.

Instead, perhaps we should add it to the global `CoveoAtomic` object and our `atomic/src/index.ts` file. We could even decide to re-export everything from `atomic/src/index.ts` in the global `CoveoAtomic` object.

I'm overall hesitant though, since we don't usually direct people towards using the global object.

btaillon (approved):
looks good!

packages/headless/src/api/platform-client.ts
olamothe: > For now I'm thinking we can live with comments in this file only.

Yep. Also, for anyone using TS, it's all going to be safely typed

packages/headless/src/app/search-engine/search-engine-configuration.ts
```diff
@@ -12,14 +12,23 @@ import {
...
-export interface SearchEngineConfiguration extends EngineConfiguration {
+export interface SearchEngineConfiguration
+  extends EngineConfiguration<SearchEngineOrganizationEndpoints> {
   /**
-   * The global headless engine configuration options specific to the SearchAPI.
+   * The base Search API endpoint.
+   *
+   * For example, https://orgid.org.coveo.com/rest/search/v2
    */
   search?: SearchConfigurationOptions;
```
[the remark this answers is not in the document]
olamothe: Oh yeah good catch, it's a bad copy paste

packages/headless/src/app/engine.ts
```diff
@@ -269,3 +298,13 @@ function createMiddleware<Reducers extends ReducersMapObject>(
     analyticsMiddleware,
   ].concat(options.middlewares || []);
 }
+
+function shouldWarnAboutOrganizationEndpoints(
+  options: EngineOptions<ReducersMapObject>
+) {
+  return isNullOrUndefined(options.configuration.organizationEndpoints);
```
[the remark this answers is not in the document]
olamothe: 👍 Ill change to undefined check only.

packages/headless/src/utils/url-utils.ts
```diff
@@ -33,3 +33,25 @@ export class URLPath {
...
+  );
+
+export const isCoveoOrganizationEndpointUrl = (
+  url: string,
+  organizationId: string
+) => {
+  return url.match(
+    new RegExp(`^https://${organizationId}\\.org(dev|stg|hipaa)?\\.coveo.com`)
+  );
+};
```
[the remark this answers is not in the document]
olamothe: How about `matchCoveoOrganizationEndpointUrl` ?

And good idea with the object return

packages/atomic/src/components.d.ts
louis-bompart, line 1: I don't understand why we need the dichotomy of `PlatformEnvironment`

packages/headless/src/api/platform-client.test.ts
```diff
@@ -28,11 +29,47 @@ describe('url helper', () => {
...
-      );
+      if (!expectation.multiRegionSubDomain) {
+        expect(analyticsUrl(expectation)).toEqual(
+          expectation.analytics.split('/rest')[0]
+        );
+      }
     });
   });
+
+  it('return the correct #getOrganizationEndpoints()', () => {
```
louis-bompart, line 40: nitpik, this is a good candidate for an `it.each`
e.g. https://github.com/coveo/cli/blob/cf0b98146c5603220e81245c4fbea6c6d8fb9f66/packages/ui/atomic/health-check/src/index.spec.ts#L70-L79

packages/headless/src/api/platform-client.test.ts
```diff
@@ -28,11 +29,47 @@ describe('url helper', () => {
 
   it('return the correct #analyticsUrl()', () => {
     allValidPlatformCombination().forEach((expectation) => {
-      expect(analyticsUrl(expectation)).toEqual(
-        expectation.analytics.split('/rest')[0]
-      );
+      if (!expectation.multiRegionSubDomain) {
```
louis-bompart, line 32: nitpik: conditional expectation is a code smell
https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md
I would  rework this with a `it.each` instead to move the loop outside the it

packages/headless/src/app/engine.test.ts
```diff
@@ -106,4 +116,59 @@ describe('engine', () => {
 
     expect(stateListener).not.toHaveBeenCalled();
   });
+
+  it('should correctly log warnings when dealing with the organizationEndpoints and platformUrl option', () => {
```
louis-bompart, line 120: nitpick: this seems a good candidate for a `it.each`

packages/headless/src/app/engine.ts
```diff
@@ -163,7 +173,26 @@ export function buildEngine<
   thunkExtraArguments: ExtraArguments
 ): CoreEngine<StateFromReducersMapObject<Reducers>, ExtraArguments> {
   const engine = buildCoreEngine(options, thunkExtraArguments);
-  const {accessToken, organizationId, platformUrl} = options.configuration;
+  const {accessToken, organizationId} = options.configuration;
+  const {organizationEndpoints} = options.configuration;
+  let {platformUrl} = options.configuration;
+
+  if (shouldWarnAboutOrganizationEndpoints(options)) {
+    // @v3 make organizationEndpoints the default.
```
louis-bompart, line 181: - [ ] Either add a mention to the V3 epic or of  its jira

packages/headless/src/utils/url-utils.test.ts
```diff
@@ -41,4 +41,32 @@ describe('URLPath', () => {
...
+        'https://platform.cloud.coveo.com',
+        'https://platform-eu.cloud.coveo.com',
+        'https://platformhipaa.cloud.coveo.com',
+        'https://analytics.cloud.coveo.com',
+        'https://search.cloud.coveo.com',
+        'https://completely.random.com',
+      ].forEach((url) =>
+        expect(isCoveoOrganizationEndpointUrl(url, 'myorg')).toBeFalsy()
+      );
+    });
```
louis-bompart, line 70: `it.each`

louis-bompart (commented):
lgtm overall:
- some questions left 🤔 
- some suggestions for the regex 🔍 
- some suggestions for the test structure 🤡

packages/atomic/src/components.d.ts
louis-bompart: Alright, I checked out the branch to try to understand.

I think this kind of 'duplications' could be fixed when we have a better division with 'headless core' for the 'common stuff between use cases of headless'.

Given that its a type and not an enum, lgtm, no compat issue foreseen.

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -280,7 +280,7 @@ export class AtomicSearchInterface
   }
 
   /**
-   * Initializes the connection with the headless search engine using options for `accessToken` (required), `organizationId` (required), `renewAccessToken`, and `platformUrl`.
+   * Initializes the connection with the headless search engine using options for accessToken (required), organizationId (required), renewAccessToken, organizationEndpoints (recommended), and platformUrl (deprecated).
```
louis-bompart, line 283: nitpik: uses `@deprecated`

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
```diff
@@ -340,6 +340,9 @@ export class AtomicSearchInterface
     this.engine.executeFirstSearchAfterStandaloneSearchBoxRedirect(analytics);
   }
 
+  /**
+   * Returns the unique, organization-specific endpoint(s) using options for `orgId` (required) and `env` (optional — possible values:`prod`, `hipaa`, `staging`,`dev`. Defaults to `prod`).
+   */
```
louis-bompart, line 345:
```suggestion
  /**
   * Returns the unique, organization-specific endpoint(s)
   * @param {string} organizationId 
   * @param {'prod'|'hipaa'|'staging'|'dev'} [env=Prod]
   */
```

packages/atomic/src/components/search/atomic-search-interface/atomic-search-interface.tsx
olamothe: This function expect a type that is derived from Headless, which is itself already marking `platformUrl` as deprecated. 

So it's already doing what you suggest.

louis-bompart (approved):
LGTM! :shipit:

PR #2800, opened 2023-03-29
feat(headless): create product listing analytics provider

thudon-coveo:
[https://coveord.atlassian.net/browse/COMSDZ-399](https://coveord.atlassian.net/browse/COMSDZ-399)

# Headless
- Create an analytics provider for product listing to map  responseId to searchUID.
- Also added related action.


packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
```diff
@@ -0,0 +1,32 @@
...
+  partialDocumentInformation,
+  ProductListingAction,
+  validateResultPayload,
+} from '../analytics/analytics-utils';
+
+export const logProductListing = (): ProductListingAction =>
+  makeAnalyticsAction(
+    'analytics/productListing/load',
+    AnalyticsType.Search,
+    (client) => client.makeInterfaceLoad(),
```
thudon-coveo, line 17: I'm not very familiar with analytics, should I add a `makeListingInterfaceLoad` function to [coveo.analytics.js](https://github.com/coveo/coveo.analytics.js) or is this the generic one ?

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
```diff
@@ -0,0 +1,32 @@
...
+    (getState) => new ProductListingAnalyticsProvider(getState)
+  );
+
+export const logProductListingOpen = (result: Result): ClickAction =>
+  makeAnalyticsAction(
+    'analytics/productListing/open',
+    AnalyticsType.Click,
+    (client, state) => {
+      validateResultPayload(result);
+      return client.makeDocumentOpen(
```
thudon-coveo, line 27: Same thing here, should I add a `makeListingOpen` ?

btaillon:
will review soon

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
```diff
@@ -0,0 +1,32 @@
+import {Result} from '../..';
```
btaillon, line 1: we don't normally end file names with `.actions.ts`, it looks like it was a typo in the case of `product-recommendations`.

The name should be `product-listing-analytics-actions.ts`.

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
btaillon: `makeInterfaceLoad` will log the actionCause `interfaceLoad`, and `makeRecommendationInterfaceLoad` (used by `product-recommendations`) will log the action cause `recommendationInterfaceLoad`.

Ultimately, I guess it's up to the UA team which is better. I don't know if product listing is different enough from search that it needs a different event, or if there's enough users tracking the current event with product listing that changing it would be bad.

btaillon (approved):

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
samisayegh: I think reusing `makeDocumentOpen` is correct, since the click event ultimately looks the same, whether it is from a search page or PLP.

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const logProductListing = (): ProductListingAction =>
+  makeAnalyticsAction(
+    'analytics/productListing/load',
+    AnalyticsType.Search,
+    (client) => client.makeInterfaceLoad(),
+    (getState) => new ProductListingAnalyticsProvider(getState)
+  );
+
+export const logProductListingOpen = (result: Result): ClickAction =>
```
samisayegh, line 21: Should we call this `logDocumentOpen` for consistency? The `ProductListingOpen` naming suggests the PLP was opened, but this event is more for a click on a specific product,

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
samisayegh: With the search engine, we provide an `executeFirstSearch` method on the engine. Calling the method dispatches a) a search with b) interfaceLoad as the analytics event.

I think the same event can be reused with the first product listing search. Would name it the same for consistency though,

packages/headless/src/api/analytics/product-listing-analytics.ts
```diff
@@ -0,0 +1,55 @@
...
+      queryText: '',
+      responseTime: 0,
+      results: this.mapResultsToAnalyticsDocument(),
+      numberOfResults: this.numberOfResults,
+    };
+  }
+
+  public getSearchUID() {
+    const newState = this.getState();
+    return newState.productListing?.responseId || this.initialState.responseId;
```
samisayegh, line 38: The MVP 🚀

packages/headless/src/api/analytics/product-listing-analytics.ts
```diff
@@ -0,0 +1,55 @@
...
+    return this.state.productListing?.products.map((p) => ({
+      documentUri: p.documentUri,
+      documentUriHash: p.documentUriHash,
+      permanentid: p.permanentid,
+    }));
+  }
+
+  private get numberOfResults() {
+    return (
+      this.state.productListing?.products.length ||
```
samisayegh, line 51: The productListing attribute should [always be present](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/app/product-listing-engine/product-listing-engine.ts#L32) on the PLP engine. I think we can simplify here, and remove the `initialState` property.

Updating the type on line 11 to [ProductListingEngineState](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/app/product-listing-engine/product-listing-engine.ts#L34) should to the trick, but this might require moving the type to it's own file. Moving the `ProductListingSection` out of the `Partial` could be another option,

samisayegh (approved):
How does it look when dispatching the event?

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
jdevost: +1

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
jdevost: +1

packages/headless/src/features/product-listing/product-listing-analytics.actions.ts
thudon-coveo: Thanks for letting me know !

packages/headless/src/api/analytics/product-listing-analytics.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  public getSearchUID() {
+    const newState = this.getState();
+    return newState.productListing?.responseId || this.initialState.responseId;
+  }
+
+  private mapResultsToAnalyticsDocument() {
+    return this.state.productListing?.products.map((p) => ({
+      documentUri: p.documentUri,
+      documentUriHash: p.documentUriHash,
```
jdevost, line 44: Is the `urihash` returned with the product listing results by default now?
I had to add it in `additionalFields` when I did it manually.

packages/headless/src/api/analytics/product-listing-analytics.ts
thudon-coveo: I'm not sure I understand the first proposition here about the [ProductListingEngineState](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/app/product-listing-engine/product-listing-engine.ts#L34). You suggest to replace `ProductListingSection` by `ProductListingEngineState`  or something like
 type StateNeededByProductListingAnalyticsProvider = `ProductListingEngineState`

samisayegh: Yea, maybe that could be one option, but if it's too complicated (e.g. due to circular references) then just moving `ProductListingSection` out of the `Partial` could be good too. I imagine the other uses-cases are doing this, so maybe best to stay consistent for now,

packages/headless/src/api/analytics/product-listing-analytics.ts
thudon-coveo: I opened a PR for it, waiting to see how the team will react https://github.com/coveo/commerce-service/pull/915

PR #2804, opened 2023-03-30
feat(quantic): modified quanticResultLink to support salesforceResult in salesforce console Tab

SimonMilord:
[SFINT-4772](https://coveord.atlassian.net/browse/SFINT-4772)

**Description:** 

I want the Salesforce result to open in a Salesforce Console tab instead of a new browser tab.

**Acceptance criteria**

- We should be opening the salesforce results in a new _subTab_ using navigation mixin of LWC.
     _We open in a subTab since primary tab would only be available if we override the current tab, which is not exactly what 
     we want._

**Technical details**

- Use a pure LWC approach to open results (with the new navigation mixin)
- Handle the different corner-cases of when to open a console tab or a browser tab, etc. 
- Handle the exception for knowledge articles which uses `sfkavid` instead of `sfid`

**DEMO: **
(notice: for the demo I hardcoded the recordId of another case in my org. This is why the tab that opens is not the result from the HIP but another case instead).

https://user-images.githubusercontent.com/73316533/228963567-f5ea1a44-502f-4092-b006-68c27c0da15d.mov

[SFINT-4772]: https://coveord.atlassian.net/browse/SFINT-4772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -47,7 +46,14 @@ export default class QuanticResultLink extends LightningElement {
    * @defaultValue `'search'`
    */
   @api useCase = 'search';
-
+  /**
+   * A list of fields to include in the query results, separated by commas.
+   * @api
+   * @type {string}
+   * @defaultValue `'date,author,source,language,filetype,parents,sfknowledgearticleid'`
```
nathanlb, line 53: need to update the documented default value as well. They should always be in sync.

```suggestion
   * @defaultValue `'date,author,source,language,filetype,parents,sfknowledgearticleid,sfid,sfkbid,sfkavid'`
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.html
```diff
@@ -1,3 +1,8 @@
 <template>
-  <a href={result.clickUri} target={target}>{displayedTitle}</a>
-</template>
\ No newline at end of file
+  <template if:true={isSalesforceLink}>
+    <a href="#" onclick={handleOpenTabSalesforceLink}>{displayedTitle}</a>
```
nathanlb, line 3: from an accessibility standpoint it might be better to add an aria-label that describes that it'll be opened in a salesforce tab. I think an anchor tag with no href isn't considered accessible.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +99,47 @@ export default class QuanticResultLink extends LightningElement {
     );
   };
 
+  checkResultType() {
```
nathanlb, line 102: instead of this function that sets a state variable i think it would be simpler to just have a getter

```
get isSalesforceLink() {
  return this.result?.raw?.sfid !== undefined
}
```

nathanlb (commented):
Good start. I think some tests are coming soon and there are a few improvement suggestions I'd like your view on.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +99,47 @@ export default class QuanticResultLink extends LightningElement {
...
+  }
+
+  handleOpenTabSalesforceLink(event) {
+    event.stopPropagation();
+
+    const targetPageRef = {
+      type: 'standard__recordPage',
+      attributes: {
+        recordId: this.getRecordIdAttribute(),
+        objectApiName: this.getObjectApiNameAttribute(),
```
erocheleau, line 115: I've tested it and you should be able to remove this `objectApiName`
Test it with knowledge as well, but basically because we are providing Salesforce with a `recordId` the `objectApiName` seems to be irrelevant?

All in all, we can't rely on the `objecttype` from our fields, because people tend to change that, or to use the "label" of the object name instead of the API name.

erocheleau:
```suggestion
        
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +99,47 @@ export default class QuanticResultLink extends LightningElement {
...
+      this.result.raw.sfkavid !== undefined
+    ) {
+      idToUse = this.result.raw.sfkavid;
+    }
+    return idToUse || '';
+  }
+
+  getObjectApiNameAttribute() {
+    return this.result?.raw?.objecttype;
+  }
```
erocheleau, line 137:
```suggestion

```

erocheleau: I believe we can safely remove this.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +99,47 @@ export default class QuanticResultLink extends LightningElement {
...
+      attributes: {
+        recordId: this.getRecordIdAttribute(),
+        objectApiName: this.getObjectApiNameAttribute(),
+        actionName: 'view',
+      },
+    };
+    this[NavigationMixin.Navigate](targetPageRef);
+  }
+
+  getRecordIdAttribute() {
```
erocheleau, line 122: This method cannot afford to return `''` as it is now as you are taking the return value directly.

Right now is we don't have the `sfid` nor the `sfkbid|sfkavid` this will return an empty string and you will try to open a `standard__recordPage` with an empty ID.

I suggest that you get the return of this method first, and if you see this returns undefined (don't return empty string, it is less clear that the method didn't find what it was looking for than if it returns undefined), you instead treat the result link as a regular result link and simply open the clickUri as if it was a normal result.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
mmitiche, line 137: Like Nathan already mentioned, a getter would be nice to have here as well:

```suggestion
  get objectType() {
    return this.result?.raw?.objecttype;
  }
```

erocheleau (changes requested):
Need better handling if your sfid is null in `handleOpenTabSalesforceLink`

If at least the `sfid` cannot be null by this point, then remove the `|| ''` part.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +99,47 @@ export default class QuanticResultLink extends LightningElement {
...
+    let idToUse = this.result.raw.sfid;
+
+    // Knowledge article uses the knowledge article version id to navigate.
+    if (
+      this.result.raw.sfkbid !== undefined &&
+      this.result.raw.sfkavid !== undefined
```
mmitiche, line 128: Why not just this:
```suggestion
      this.result?.raw?.sfkbid &&
      this.result?.raw?.sfkavid
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +99,47 @@ export default class QuanticResultLink extends LightningElement {
...
+
+    const targetPageRef = {
+      type: 'standard__recordPage',
+      attributes: {
+        recordId: this.getRecordIdAttribute(),
+        objectApiName: this.getObjectApiNameAttribute(),
+        actionName: 'view',
+      },
+    };
+    this[NavigationMixin.Navigate](targetPageRef);
```
mmitiche, line 119: Is there a case where this operation fails? like if the recordId or the objectApiName are incorrect or missing for example.
We should fallback the usual behaviour of the QuanticResultLink component if this operation fails.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
SimonMilord: 😄 I overthought that one

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
SimonMilord: yeah double checked with knowledge articles, seems to work fine

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
SimonMilord: indeed

SimonMilord: as brought up by @erocheleau I tested the NavigationMixin call without the objectApiName and it works well, so we dont need this at all in the end

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
SimonMilord: ok I will change the getter to this:

`return sfid !== undefined && sfid !== null;`

so that by the time it gets the recordId the sfid is for sure not null or undefined. If it is `isSalesforceLink` is false and the link will simply default to an a tag with `href={result.clickUri}`

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
SimonMilord: addressed [here](https://github.com/coveo/ui-kit/pull/2804#:~:text=ok%20I%20will,href%3D%7Bresult.clickUri%7D)

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,114 @@
...
+      const linkSalesforce = element.shadowRoot.querySelector('a');
+      linkSalesforce.click();
+
+      const {pageReference} = getNavigateCalledWith();
+
+      expect(spy).toHaveBeenCalled();
+      expect(pageReference.attributes.recordId).toBe(MOCK_RECORD_ID);
+    });
+
+    describe('is a knowledge article', () => {
```
nathanlb, line 81:
```suggestion
    describe('when the result is a knowledge article', () => {
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,114 @@
...
+      it('should also open the result link in a salesforce console subTab', async () => {
```
nathanlb, line 82:
```suggestion
      it('should open the result link in a salesforce console subTab', async () => {
```
IMO it's best to treat the tests as independent from each other, even in the naming. Ideally the order in which they're declared in the spec file shouldn't matter.

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+      const element = createTestComponent({...mockSalesforceResult});
+      await flushPromises();
+
+      const linkSalesforce = element.shadowRoot.querySelector('a');
+      linkSalesforce.click();
+
+      const {pageReference} = getNavigateCalledWith();
+
+      expect(spy).toHaveBeenCalled();
```
nathanlb, line 77: Why do we care to check that the `getHeadlessEnginePromise` mock was called? isn't it called when the component initializes, regardless of the click?

nathanlb (approved):
I like 👍

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,108 @@
+// @ts-nocheck
```
mmitiche, line 1: Why do we need to completely disable all the type script checks?

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,108 @@
...
+
+      const element = createTestComponent({...mockSalesforceResult});
+      await flushPromises();
+
+      const linkSalesforce = element.shadowRoot.querySelector('a');
+      linkSalesforce.click();
+
+      const {pageReference} = getNavigateCalledWith();
+
+      expect(pageReference.attributes.recordId).toBe(MOCK_RECORD_ID);
```
mmitiche, line 74: Instead of comparing with the variable you created, it would be better to compare with `mockSalesforceResult.result.raw.sfid`
This way we would have only one source of truth instead of two, your `MOCK_RECORD_ID ` variable and  the `salesforceResult` JSON file.

Same remark for the other tests.

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,108 @@
...
+        const linkSalesforce = element.shadowRoot.querySelector('a');
+        linkSalesforce.click();
+
+        const {pageReference} = getNavigateCalledWith();
+
+        expect(pageReference.attributes.recordId).toBe(MOCK_RECORD_ID);
+      });
+    });
+  });
+  describe('when the result is NOT of type salesforce', () => {
```
mmitiche, line 94:
```suggestion

  describe('when the result is NOT of type salesforce', () => {
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/data/knowledgeArticleResult.json
```diff
@@ -0,0 +1,9 @@
+{
+  "result": {
+    "clickUri": "http://coveo.com/",
```
mmitiche, line 3: It doesn't matter but it's better x)
```suggestion
    "clickUri": "https://coveo.com/",
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,108 @@
...
+    it('should open the result link in the current browser tab', async () => {
+      mockHeadless();
+
+      const element = createTestComponent({...mockDefaultResult});
+      await flushPromises();
+
+      const link = element.shadowRoot.querySelector('a');
+      link.click();
+
+      expect(link).toHaveAttribute('href', 'http://coveo.com/');
```
mmitiche, line 104: Same as above
```suggestion
      expect(link).toHaveAttribute('href', mockDefaultResult.result.clickUri);
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.html
```diff
@@ -1,3 +1,8 @@
 <template>
-  <a href={result.clickUri} target={target}>{displayedTitle}</a>
-</template>
\ No newline at end of file
+  <template if:true={isSalesforceLink}>
+    <a href="#" onclick={handleOpenTabSalesforceLink} aria-label="opens in a salesforce console subTab">{displayedTitle}</a>
```
mmitiche, line 3: `"opens in a salesforce console subTab"` needs to be localized.
We need to create a label and a french translation for it. check [CustomLabels.labels-meta.xml](https://github.com/coveo/ui-kit/blob/master/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml) and [fr.translation-meta.xml](https://github.com/coveo/ui-kit/blob/master/packages/quantic/force-app/main/translations/fr.translation-meta.xml)

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +88,40 @@ export default class QuanticResultLink extends LightningElement {
...
+
+  getRecordIdAttribute() {
+    let idToUse = this.result.raw.sfid;
+
+    // Knowledge article uses the knowledge article version id to navigate.
+    if (this.result?.raw?.sfkbid && this.result?.raw?.sfkavid) {
+      idToUse = this.result.raw.sfkavid;
+    }
+    return idToUse;
+  }
```
mmitiche, line 111: a bit more simple imo
```suggestion
  get recordId() {
    // Knowledge article uses the knowledge article version id to navigate.
    if (this.result?.raw?.sfkbid && this.result?.raw?.sfkavid) {
      return this.result.raw.sfkavid;
    }
    return this.result.raw.sfid;
  }
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +88,40 @@ export default class QuanticResultLink extends LightningElement {
...
+
+  /**
+   * Checks if the Result type is Salesforce.
+   */
+  get isSalesforceLink() {
+    const sfid = this.result?.raw?.sfid;
+    return sfid !== undefined && sfid !== null;
```
mmitiche, line 118:
```suggestion
    return !!this.result?.raw?.sfid
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -1,9 +1,8 @@
...
-} from 'c/quanticHeadlessLoader';
-import {ResultUtils} from 'c/quanticUtils';
+import { getHeadlessBundle, getHeadlessEnginePromise } from 'c/quanticHeadlessLoader';
+import { ResultUtils } from 'c/quanticUtils';
+import { NavigationMixin } from 'lightning/navigation';
+import { LightningElement, api } from 'lwc';
+
 
 /** @typedef {import("coveo").Result} Result */
 /** @typedef {import("coveo").SearchEngine} SearchEngine */
```
mmitiche, line 8: It would be nice to update the description of this component in line `11` in order to reflect the new capability that got added to the component.

Because this description is the one will be displayed in our documentation.

packages/quantic/package.json
```diff
@@ -54,6 +54,7 @@
     "@octokit/graphql-schema": "10.74.2",
     "@salesforce/eslint-config-lwc": "3.3.3",
     "@salesforce/sfdx-lwc-jest": "1.1.2",
+    "@testing-library/jest-dom": "^5.16.5",
```
mmitiche, line 57: Is this necessary in order to test the navigation?

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
SimonMilord: this is good ? 

>  The `QuanticResultLink` component creates a clickable link from a result that points to the original item.
>  If the result has is of type `Salesforce` it will open the link in a new salesforce console subtab.
>  If the result is of type other than `Salesforce`, it will open the link in the browser tab.

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,108 @@
...
+    jest.clearAllMocks();
+  }
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  describe('when the result is of type salesforce', () => {
+    it('should open the result link in a salesforce console subTab', async () => {
+      let MOCK_RECORD_ID = '123';
```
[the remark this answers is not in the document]
SimonMilord: got confused with the syntax from the tests in APEX 🤣

packages/quantic/package.json
SimonMilord: yes in order to be able to use the function : `toHaveAttribute` in the last test

cmartel-coveo (approved):
Talked to Simon M., all good!

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1072,6 +1072,11 @@
       <shortDescription>Inclusion filter on field; no results</shortDescription>
     </labels>
     <labels>
+      <fullName>quantic_OpensInSalesforceSubTab</fullName>
+      <value>Opens in a salesforce console subTab</value>
```
mmitiche, line 1076: I think we can write subtab here, that's not JS x)
```suggestion
      <value>Opens in a salesforce console subtab</value>
```

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1072,6 +1072,11 @@
...
+      <language>en_US</language>
+      <protected>false</protected>
+      <shortDescription>Opens in a salesforce console subTab</shortDescription>
```
mmitiche, line 1079:
```suggestion
      <shortDescription>Opens in a salesforce console subtab</shortDescription>
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/data/knowledgeArticleResult.json
mmitiche: do you prefer `http`? x)

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
mmitiche: I see, well personally I prefer not to disable the whole ts checks because they can be useful sometimes but feel free to go with the option you judge best.

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,105 @@
...
+  return new Promise((resolve) => setTimeout(resolve, 0));
+}
+
+// Helper function to mock headless for this suite of tests.
+function mockHeadless() {
+  jest.spyOn(mockHeadlessLoader, 'getHeadlessBundle').mockReturnValue(
+    new Promise((resolve) => {
+      resolve();
+    })
+  );
```
mmitiche, line 38: The return value of `getHeadlessBundle` is not supposed to be a promise, it's supposed to return the following:
```
{
    buildInteractiveResult: jest.fn(),
}
```
```suggestion
  jest.spyOn(mockHeadlessLoader, 'getHeadlessBundle').mockReturnValue(
   {
    buildInteractiveResult: jest.fn(),
   }
  );
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -1,20 +1,22 @@
...
+import { NavigationMixin } from 'lightning/navigation';
+import opensInSalesforceSubTab from '@salesforce/label/c.quantic_OpensInSalesforceSubTab';
+
 
 /** @typedef {import("coveo").Result} Result */
 /** @typedef {import("coveo").SearchEngine} SearchEngine */
 
 /**
  * The `QuanticResultLink` component creates a clickable link from a result that points to the original item.
+ * If the result has is of type `Salesforce` it will open the link in a new salesforce console subtab.
```
mmitiche, line 13:
```suggestion
 * If the result is of type `Salesforce` it will open the link in a new salesforce console subtab.
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
mmitiche: LGTM!

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -90,10 +95,37 @@ export default class QuanticResultLink extends LightningElement {
...
+      type: 'standard__recordPage',
+      attributes: {
+        recordId: this.getRecordIdAttribute(),
+        actionName: 'view',
+      },
+    };
+    this[NavigationMixin.Navigate](targetPageRef);
+  }
+
+  getRecordIdAttribute() {
```
mmitiche, line 110: just as an FYI, LWC has this special way to implement getters: https://salesforcediaries.com/2020/10/12/understand-getter-in-lightning-web-component/

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -573,6 +573,8 @@
       <name>quantic_InclusionFilter_zero</name>
     </customLabels>
     <customLabels>
+      <label>Ouvre dans un sous-onglet de la console Salesforce</label>
+      <name>quantic_OpensInSalesforceSubTab</name>
```
mmitiche, line 577: I see that there was two `customLabels ` missing, that may have occurred after a conflict after pulling from master, please confirm that all the tags are close properly in this file.
```suggestion
      <name>quantic_OpensInSalesforceSubTab</name>
    </customLabels>
    <customLabels>
```

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
mmitiche, line 1079: the labels tag is not closed properly, please confirm that labels are properly closed in this file.
```suggestion
      <shortDescription>Opens in a salesforce console subTab</shortDescription>
    </labels>
    <labels>
```

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/data/knowledgeArticleResult.json
SimonMilord: thought I changed it my bad

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
SimonMilord: good catch

packages/quantic/force-app/main/translations/fr.translation-meta.xml
SimonMilord: mmm odd, I will repush and see for conflicts, seems fine on my end for the closures

packages/quantic/force-app/test/jest-mocks/lightning/navigation/navigation.js
erocheleau, line 1: You have taken this file as-is right?

packages/quantic/package.json
erocheleau: We don't need that.
If someday we choose to add it, we will but not through 1 PR that does something else.
It will be it's own task

erocheleau (changes requested):
Please remove the testing-library import

And move the complexity of handling the click in the JS file instead of having conditional rendering in the HTML template.

packages/quantic/package.json
SimonMilord: ok ill remove it

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,113 @@
...
+      beginDelayedSelect: jest.fn(),
+      cancelPendingSelect: jest.fn(),
+    }),
+  });
+
+  jest.spyOn(mockHeadlessLoader, 'getHeadlessEnginePromise').mockReturnValue(
+    new Promise((resolve) => {
+      resolve();
+    })
+  );
```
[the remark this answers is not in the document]
erocheleau: I meant that the `getHeadlessEnginePromise` doesn't it need to return something in order for the mock to be valid and the code of your component to run? Or it's just waiting after the promise has resolved?

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
```diff
@@ -0,0 +1,113 @@
...
+      });
+    });
+  });
+  describe('when the result is NOT of type salesforce', () => {
+    it('should open the result link in the current browser tab', async () => {
+      const element = createTestComponent({...mockDefaultResult});
+      await flushPromises();
+
+      const link = element.shadowRoot.querySelector('a');
+      link.click();
```
erocheleau, line 105: Why do you need to click it btw?
Your test only look at the attributes `href` and `target`.
Those are not impacted after you've clicked on it right?

erocheleau (approved):
Looks good to me, do you prefer it like this with the click handler in pure JS?

packages/quantic/force-app/main/default/lwc/quanticResultLink/__tests__/quanticResultLink.test.js
SimonMilord: you are right, for the other 2 tests I need the click but not this one

SimonMilord:
> Looks good to me, do you prefer it like this with the click handler in pure JS?

yeah I think its a bit more clean than the if:true templates polluting the HTML file unnecessarily

packages/quantic/jest.config.js
```diff
@@ -13,11 +13,16 @@ module.exports = {
   ],
   moduleNameMapper: {
     '^@salesforce/i18n/': '<rootDir>/typings/lwc/customlabels.d.ts',
-    '^lightning/modal$': '<rootDir>/force-app/test/jest-mocks/lightning/modal/modal',
-    '^lightning/modalBody$': '<rootDir>/force-app/test/jest-mocks/lightning/modalBody/modalBody',
-    '^lightning/modalFooter$': '<rootDir>/force-app/test/jest-mocks/lightning/modalFooter/modalFooter',
-    '^lightning/modalHeader$': '<rootDir>/force-app/test/jest-mocks/lightning/modalHeader/modalHeader',
-    '^c/quanticResult$': '<rootDir>/force-app/test/jest-mocks/quantic/quanticResult/quanticResult',
```
mmitiche, line 20: We shouldn't delete this, it's needed for the the unit tests of the Quantic Result Children component.

packages/quantic/.sf/config.json
mmitiche, line 1: We actually need this file to still be in Quantic.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -1,20 +1,23 @@
...
+
+import opensInBrowserTab from '@salesforce/label/c.quantic_OpensInBrowserTab';
+import opensInSalesforceSubTab from '@salesforce/label/c.quantic_OpensInSalesforceSubTab';
 
 /** @typedef {import("coveo").Result} Result */
 /** @typedef {import("coveo").SearchEngine} SearchEngine */
 
 /**
  * The `QuanticResultLink` component creates a clickable link from a result that points to the original item.
+ * If the result is of a Salesforce record or a Salesforce Knowledge article it will open the link in a new salesforce console subtab.
```
mmitiche, line 14:
```suggestion
 * If the result is a Salesforce record or a Salesforce Knowledge article it will open the link in a new salesforce console subtab.
```

PR #2805, opened 2023-03-31
chore(IPX): remove focus trap on main modal

dguerinCoveo:
Since the [changes to not prevent focus outside the scope of the focus-trap](https://github.com/coveo/ui-kit/pull/2750), we do not need to have a focus trap on the main ipx modal anymore. 

https://coveord.atlassian.net/browse/SVCINT-2161


btaillon (approved):

btaillon:
normally, I would recommend adding E2E tests, but unfortunately we get very flaky results when asserting focus with Cypress.

jbvouma (approved):

pathiery (approved):
@dguerinCoveo how much (if any) of our internal css are we breaking with this change ?

dguerinCoveo:
> 

Nothing will break that is not already broken style-wise in the IPX in the loader

mikegron (approved):

PR #2806, opened 2023-03-31
feat(atomic)!: enable atomic modal in shadow dom

dguerinCoveo:
Instead of putting the `atomic-modal-opened` class on document.body, put it inside the interface element so that the classname can be set even when inside a shadow dom. 

I checked also for other things that would be problematic inside a shadow dom and did not find anything else.

https://coveord.atlassian.net/browse/SVCINT-2159

### DIFF
--- a/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
+++ b/packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
@@ -21,7 +21,7 @@ import {once, randomID} from '../../../utils/utils';
 import {AnyBindings} from '../interface/bindings';
 
 /**
- * When the modal is opened, the class `atomic-modal-opened` is added to the body, allowing further customization.
+ * When the modal is opened, the class `atomic-modal-opened` is added to the interfaceElement, allowing further customization.
  *
  * @part backdrop - The transparent backdrop hiding the content behind the modal.
  * @part container - The modal's outermost container with the outline and background.
@@ -74,14 +74,14 @@ export class AtomicModal implements InitializableComponent<AnyBindings> {
 
     if (isOpen) {
       this.wasEverOpened = true;
-      document.body.classList.add(modalOpenedClass);
+      this.bindings.interfaceElement.classList.add(modalOpenedClass);
       await this.waitForAnimationEnded();
       if (watchToggleOpenExecution !== this.currentWatchToggleOpenExecution) {
         return;
       }
       this.focusTrap!.active = true;
     } else {
-      document.body.classList.remove(modalOpenedClass);
+      this.bindings.interfaceElement.classList.remove(modalOpenedClass);
       if (isIOS()) {
         await this.waitForAnimationEnded();
       }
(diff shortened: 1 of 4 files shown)


dguerinCoveo:
> :suspect: `!` wowow: we don't do breaking change lightly in this repo (it's not like `platform-client`)
> 
> * If your change is absolutetly in use by customers, you will have to work around it and schedule a work order for our V3
> * If the current behaviour is not the expected one, then its a fix, and not a breaking change (can't break what's broken)

I see, I'll work around it then. I suppose its not really possible to know if its used by a customer right?

btaillon:
> > :suspect: `!` wowow: we don't do breaking change lightly in this repo (it's not like `platform-client`)
> > 
> > * If your change is absolutetly in use by customers, you will have to work around it and schedule a work order for our V3
> > * If the current behaviour is not the expected one, then its a fix, and not a breaking change (can't break what's broken)
> 
> I see, I'll work around it then. I suppose its not really possible to know if its used by a customer right?
> 
> The current behaviour works as expected for most customers, so I'd say its not a fix. It could cause problems if there were multiple search interfaces, but since it might break perfectly fine implementations, I propose we just update the style of two components. I don't know if I should put a comment to maybe delete the styling of `document.body` in the release of v3 though

Yeah, a comment prefixed with `TODO:` with a note to remove in v3 would be a good idea. Alternatively, you can create a dedicated Jira in KIT-2157, but this is a pretty simple change so I don't think it's necessary.

btaillon (approved):

pathiery (approved):

olamothe (approved):

packages/atomic/src/components/common/atomic-modal/atomic-modal.tsx
jdevost, line 1: I know it may be unrelated, but could you include the following change in this PR? 

[Line 101] `if (e.key.toLowerCase() === 'escape') {`
and use `.key?`, like this:

`if (e.key?.toLowerCase() === 'escape') {`


I have a case where this generate a JavaScript error: 
image

dguerinCoveo: Of course! ✅

packages/atomic/src/components/ipx/atomic-ipx-modal/atomic-ipx-modal.tsx
```diff
@@ -53,13 +53,17 @@ export class AtomicIPXModal implements InitializableComponent<AnyBindings> {
     const modalOpenedClass = 'atomic-ipx-modal-opened';
 
     if (isOpen) {
+      //TODO: remove the addition of a class to the body in atomicV3
```
mikegron, line 56: A tu un jira pour ca lol

mikegron (approved):

packages/atomic/src/components/ipx/atomic-ipx-modal/atomic-ipx-modal.tsx
dguerinCoveo: > Yeah, a comment prefixed with TODO: with a note to remove in v3 would be a good idea. Alternatively, you can create a dedicated Jira in [KIT-2157](https://coveord.atlassian.net/browse/KIT-2157), but this is a pretty simple change so I don't think it's necessary.
:trollface:

PR #2807, opened 2023-04-03
Fix undefined bound state

NNUrbinaGa:
The subscribed state has to be mapped when the subscription starts, not just when a change is triggered

This is causing race conditions, when the subscribed state is set before the component initialises it remains undefined, and never syncs

### DIFF
--- a/packages/atomic/src/utils/initialization-utils.tsx
+++ b/packages/atomic/src/utils/initialization-utils.tsx
@@ -280,7 +280,7 @@ export function BindStateToController(
           component
         );
       }
-
+      this[stateProperty] = this[controllerProperty].state;
       unsubscribeController = this[controllerProperty].subscribe(() => {
         this[stateProperty] = this[controllerProperty].state;
         options?.onUpdateCallbackMethod &&


btaillon:
Most controllers call the listener immediately upon subscription, with some exceptions, which is likely to be a change we need to do in the future.

Which controller did you encounter this issue with?

NNUrbinaGa:
notifications, it causes this error 
```
TypeError: Cannot destructure property 'notifications' of 'this.notifyTriggerState' as it is undefined.
    at p.render (atomic-notifications.tsx:99:12)
    at s.render (initialization-utils.tsx:200:31)
```

btaillon:
Created a Jira: https://coveord.atlassian.net/browse/KIT-2382

louis-bompart:
Like, @btaillon explained, the fix to this issue will probably be to rework more code.
Thus, we're unlikely to merge this PR. I cannot convert it to an issue, so I created one on your behalf (#2826) @NNUrbinaGa so you can keep track of the resolution nonetheless.

PR #2808, opened 2023-04-03
chore(atomic): update i18next (#2808

btaillon:
https://coveord.atlassian.net/browse/KIT-2334

I spent a while trying to add typing for localization following the new docs (https://www.i18next.com/overview/typescript). Unfortunately, I kept getting `TS2589: Type instantiation is excessively deep and possibly infinite.` when `i18n.t` was called with a `ns` in its options parameter. After a while, I decided it wasn't worth the effort and removed typing by declaring `resources` as a string record.
* If you wish to see what this looks like or reproduce the issue, I created a branch here: https://github.com/coveo/ui-kit/pull/2809

To do before merging:
- [x] Verify that exported type definitions aren't significantly changed.
- [x] Verify that plurals aren't broken.


packages/atomic/src/components/common/interface/i18n.ts
```diff
@@ -61,5 +61,12 @@ export function initi18n(atomicInterface: BaseAtomicInterface<AnyEngineType>) {
     interpolation: {
       escape: (str) => DOMPurify.sanitize(str),
     },
+    compatibilityJSON: 'v3',
```
olamothe, line 64: If I understand correctly from i18n doc, this is what makes plural support the same as the version currently used on master, is that correct ?

packages/atomic/src/locales.json
```diff
@@ -1212,6 +1212,14 @@
     "en": "Apply custom start and end dates for the {{label}} facet",
     "fr": "Appliquer les dates de début et de fin manuelles pour la facette {{label}}"
   },
+  "past-minute": {
```
olamothe, line 1215: Why is this needed now ?

packages/atomic/src/components/common/interface/i18n.ts
```diff
@@ -61,5 +61,12 @@ export function initi18n(atomicInterface: BaseAtomicInterface<AnyEngineType>) {
...
   });
 }
+
+declare module 'i18next' {
+  interface CustomTypeOptions {
+    resources: Record<string, Record<string, string>>;
```
olamothe, line 70: Where is this type used ?

packages/atomic/src/components/common/interface/i18n.ts
btaillon: Yes

packages/atomic/src/locales.json
btaillon: I forgot to mention that.

While testing with typings, I noticed that this was missing. It's not strictly needed for the PR.

packages/atomic/src/components/common/interface/i18n.ts
btaillon: It's used by `i18n.t(...)`.

olamothe (approved):

packages/atomic/src/components/common/interface/i18n.ts
olamothe: So, is it really useful then ? Or mostly an artefact of trying to add typing to the available keys ?

btaillon: Yeah, we need it to compile, since i18next now tries to add typings by default.

PR #2811, opened 2023-04-03
feat(SFINT-4880): Added default searchhub as an example

erocheleau:
New line to provide an example of how to set a searchHub as part of the Quantic token creation through the Globals method from the Coveo for Salesforce package.

With link to the documentation page.

### DIFF
--- a/packages/quantic/force-app/coveo-token/main/default/classes/CoveoTokenProvider.cls
+++ b/packages/quantic/force-app/coveo-token/main/default/classes/CoveoTokenProvider.cls
@@ -6,7 +6,12 @@ global with sharing class CoveoTokenProvider implements ITokenProvider {
      */
     @AuraEnabled(continuation=true cacheable=false)
     global static String getHeadlessConfiguration() {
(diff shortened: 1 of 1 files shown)


packages/quantic/force-app/coveo-token/main/default/classes/CoveoTokenProvider.cls
mmitiche, line 1: We should set the searchHub when creating the search token with the `generateSearchToken` method and not with the `getEndpointData` method: 
```
    Map<String, Object> coveoEndpointData = CoveoV2.Globals.getEndpointData();
    String searchToken = CoveoV2.Globals.generateSearchToken(new Map<String, Object> {
      'searchHub' => 'example_coveo'
    });
```

erocheleau: 🤦 I spent so long thinking if I should setup prettier for apex in Quantic that I didn't realize I put the line in the wrong spot 😂

mmitiche (approved):

SimonMilord (approved):

louis-bompart (approved):

PR #2812, opened 2023-04-04
fix(headless): reset preview content on `fetchMoreResults` and `fetchPage`

louis-bompart:
https://coveord.atlassian.net/browse/KIT-2389
Sames as #2657, but for `fetchMoreResults` and `fetchPage`

This fix the issue because it will make:
https://github.com/coveo/ui-kit/blob/1a358caee72aa50c0bd45a98c0760becbd39829c/packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx#L208-L211
be/stay/return false (as it should).

## Testing:
 - [x] UT
 - [x] Manual tests:
    -  `git checkout "@coveo/headless@2.10.1" -f ; npm run build ; npm run dev`
       - Click on Quickview, close quickview, click on load more results: the quickview reopen.
       - Edit `packages/atomic/www/index.html` to replace the `atomic-load-more-results` by `atomic-results-per-page`
       - Click on Quickview, close quickview, click on load more results: the quickview reopen.
    -  `git checkout "fix/KIT-2389" -f ; npm run build ; npm run dev`
       - Click on Quickview, close quickview, click on load more results: the quickview does not reopen.
       - Edit `packages/atomic/www/index.html` to replace the `atomic-load-more-results` by `atomic-results-per-page`
       - Click on Quickview, close quickview, click on load more results: the quickview does not reopen.

### DIFF
--- a/packages/headless/src/features/result-preview/result-preview-slice.test.ts
+++ b/packages/headless/src/features/result-preview/result-preview-slice.test.ts
@@ -4,7 +4,8 @@ import {buildMockSearch} from '../../test/mock-search';
 import {buildMockSearchResponse} from '../../test/mock-search-response';
 import {logInterfaceLoad} from '../analytics/analytics-actions';
 import {executeSearch} from '../insight-search/insight-search-actions';
-import {fetchMoreResults} from '../search/search-actions';
+import {logPageNext} from '../pagination/pagination-analytics-actions';
+import {fetchMoreResults, fetchPage} from '../search/search-actions';
 import {
   fetchResultContent,
   nextPreview,
@@ -105,6 +106,20 @@ describe('ResultPreview', () => {
         'fifth',
         'eight',
       ]);
+    });
+  });
+
+  describe('on #fetchPage.fulfilled', () => {
+    it('re-initialize the state to initial when a query returns correctly', () => {
+      state.content = 'content';
+      state.contentURL = 'url';
+      state.isLoading = true;
+      state.uniqueId = 'uniqueId';
+      const action = fetchPage.fulfilled(buildMockSearch(), '', logPageNext());
+
+      const finalState = resultPreviewReducer(state, action);
+
+      expect(finalState).toEqual(getResultPreviewInitialState());
     });
   });
 
--- a/packages/headless/src/features/result-preview/result-preview-slice.test.ts
+++ b/packages/headless/src/features/result-preview/result-preview-slice.test.ts
@@ -73,6 +73,18 @@ describe('ResultPreview', () => {
   });
 
   describe('on #fetchMoreResults.fulfilled', () => {
+    it('re-initialize the state to initial when a query returns correctly', () => {
+      state.content = 'content';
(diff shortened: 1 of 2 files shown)


packages/headless/src/features/result-preview/result-preview-slice.test.ts
```diff
@@ -73,6 +73,18 @@ describe('ResultPreview', () => {
   });
 
   describe('on #fetchMoreResults.fulfilled', () => {
+    it('re-initialize the state to initial when a query returns correctly', () => {
```
mrrajamanickam-coveo, line 76: is there a test for the reverse condition? where the state isn't re-initialized or when a query doesn't return correctly?

packages/headless/src/features/result-preview/result-preview-slice.ts
```diff
@@ -1,4 +1,5 @@
 import {createReducer} from '@reduxjs/toolkit';
+import {WritableDraft} from 'immer/dist/internal';
```
olamothe, line 2: I would not import types from internal of immer, unless absolutely necessary.

```typescript
const resetPreviewContentState = (state:ResultPreviewState) => {

[... ]
```
Should work just fine.

packages/headless/src/features/result-preview/result-preview-slice.ts
```diff
@@ -30,18 +34,13 @@ export const resultPreviewReducer = createReducer(
...
-        state.uniqueId = uniqueId;
-        state.contentURL = contentURL;
-
+        resetPreviewContentState(state);
         state.resultsWithPreview = getUniqueIdsOfResultsWithHTMLVersion(
           action.payload.response.results
         );
       })
       .addCase(fetchMoreResults.fulfilled, (state, action) => {
+        resetPreviewContentState(state);
```
olamothe, line 43: I think we would also need to reset preview content state on `fetchPage.fullfilled`. 

I think you would be able to reproduce the same kind of problem reported on discuss with the results per page component, for example.

louis-bompart: Mh good catch! I thought about the `atomic-pager`, but I didn't do anything because I didn't see any reducer in place, so I assumed it was 'fine.

I now tested and confirm that the incorrect behavior is indeed present with the `atomic-pager`

packages/headless/src/features/result-preview/result-preview-slice.test.ts
louis-bompart: No, but IMHO, this is normal: We're testing the execution of the reducer associated with the `fetchMoreResults` actions, so testing the case where a query doesn't return correctly or 'fetchMoreResults' doesn't occur is a bit like having a function, and saying that to unit test it you need test the case where you don't call it, I found that a bit odd, because that means you don't have any 'Act' in an Arrange/Act/Assert test.

olamothe (approved):

PR #2814, opened 2023-04-05
fix(atomic): use img instead of atomic-icon for ipx button icon

jbvouma:
[SVCINT-2203](https://coveord.atlassian.net/browse/SVCINT-2203)

We had issues with some SVGs where it wasn't showing some images correctly (too big). 
I found out that using the `img` tag instead of the `atomic-icon` was resolving it.

[SVCINT-2203]: https://coveord.atlassian.net/browse/SVCINT-2203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/images/close.svg
jbvouma, line 1: Added the XML namespace `xmlns="http://www.w3.org/2000/svg"` because it wasn't working without it when encoded in Base64.

btaillon (commented):
It's normal for the svg in `atomic-icon` to expand to fill it, that was done on purpose so that we can set the size of `atomic-icon` and the svg inside scales.

Instead of trying  to use a different method of displaying svgs in IPX, I think we should try to use `atomic-icon`, but just apply some styling so that it doesn't expand too much.

btaillon:
> > It's normal for the svg in `atomic-icon` to expand to fill it, that was done on purpose so that we can set the size of `atomic-icon` and the svg inside scales.
> > Instead of trying to use a different method of displaying svgs in IPX, I think we should try to use `atomic-icon`, but just apply some styling so that it doesn't expand too much.
> 
> This is just the case of the button icon. I tried to adjust styling but it still wasn't working for some cases. The img seems to work every time.
> 
> If you can try to access to this [IPX builder](https://searchstg.cloud.coveo.com/rest/organizations/qaregression2/ipxinterface/v1/interfaces/f6bfb136-5534-4110-aa64-72c42b187195/edit) and try to play with it. I did spend all day yesterday trying to adjust the CSS but didn't find a style for it.

Can you give me an example image that doesn't work?

When I try with https://elinux.org/images/c/cb/Raspberry_Pi_Logo.svg, it originally shows up way too big, but upon adding the `viewBox` property to the SVG to match its `width` and `height`, the icon scales correctly.

I suspect that the issue may be that we need to fix the `viewBox` of icons provided by the user.

jbvouma:
> viewBox

Sorry for the late follow up. I was off yesterday. You are probably right about the viewbox but I'm not sure whether we want to force this update or just specify it in the IPX doc 🤔.

I tested with two others [SVG_Logo](https://upload.wikimedia.org/wikipedia/commons/4/4f/SVG_Logo.svg) and [Providenca](https://upload.wikimedia.org/wikipedia/commons/1/12/777_Providenca_okulo.svg).

mikegron:
I agree with @jbvouma on that one @btaillon .. We don't want to modify the SVG and img just works so there's no advantage of using `atomic-icon`? Isn't that component used mostly to display assets?

btaillon:
> > viewBox
> 
> Sorry for the late follow up. I was off yesterday. You are probably right about the viewbox but I'm not sure whether we want to force this update or just specify it in the IPX doc 🤔.
> 
> I tested with two others [SVG_Logo](https://upload.wikimedia.org/wikipedia/commons/4/4f/SVG_Logo.svg) and [Providenca](https://upload.wikimedia.org/wikipedia/commons/1/12/777_Providenca_okulo.svg).

> I agree with @jbvouma on that one @btaillon .. We don't want to modify the SVG and img just works so there's no advantage of using `atomic-icon`? Isn't that component used mostly to display assets?

Your current code already modifies the icon though, it changes the `fill` property.

The main advantage of `atomic-icon` is consistency across the UI; an icon that works in one place should also work in all others. We allow the customization of certain icons, such as the `atomic-result-badge` and the `atomic-pager`.

Another small advantage of `atomic-icon` is that it sanitizes the icon using DOMPurifier.

It's not strictly necessary, we could have behaviour that's different specifically for the IPX button, but I prefer to have as few exceptions as possible.

PR #2816, opened 2023-04-06
fix(quantic): transforming the numberOfFoldedResults property  to a numeric value

mmitiche:
[SFINT-4971](https://coveord.atlassian.net/browse/SFINT-4971)

## The issue:

before this fix we were obliged to create a property in the JS file containing a numeric value to be able to pass it to the component.

```
// exampleSearch.js
numberOfFoldedResults = 3;
```
```
// exampleSearch.html
  <c-quantic-folded-result-list
          number-of-folded-results={numberOfFoldedResults}
          engine-id={engineId}
  ></c-quantic-folded-result-list>
```

## The solution:

Added one line to transform the value of the property `numberOfFoldedResults` of the `QuanticFoldedResultList` component to a numeric value in order to facilitate its usage.

With that fix, this property can be passed easily like so:
```
  <c-quantic-folded-result-list
          number-of-folded-results="3"
          engine-id={engineId}
  ></c-quantic-folded-result-list>
```

[SFINT-4971]: https://coveord.atlassian.net/browse/SFINT-4971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticFoldedResultList/quanticFoldedResultList.js
```diff
@@ -99,7 +99,7 @@ export default class QuanticFoldedResultList extends LightningElement {
       options: {
         fieldsToInclude: this.fields,
         folding: {
-          numberOfFoldedResults: this.numberOfFoldedResults,
+          numberOfFoldedResults: Number(this.numberOfFoldedResults),
```
erocheleau, line 102: What about using something like this?

```
_numberOfFoldedResults = 2;
@api
get numberOfFoldedResults() {
    return this._numberOfFoldedResults;
}
set numberOfFoldedResults(value) {
    const num = parseInt(value, 10);
    this._numberOfFoldedResults = (num !== NaN) ? num : 2;
}
```

That way you ensure it's a number, otherwise you go back to the default?

erocheleau (commented):
Just a proposition

packages/quantic/force-app/main/default/lwc/quanticFoldedResultList/quanticFoldedResultList.js
mmitiche: Yeah this could work perfectly as well 👌
My only small concern is about consistency, because  we currently do not have this strategy of falling back to the default value  when an invalid property is passed to a Quantic component. 

IMO the end user should be very aware that the value  introduced to a property is invalid. Falling back to the default value   can make invalid properties, easily, go under the radar/unnoticed which can be a bit problematic.

Going with the implementation in this PR was my first choice mainly because  we currently using the same logic with other numeric public properties like, for example, in the Quantic Facet component here: https://github.com/coveo/ui-kit/blob/71c783e9d989573f40959ff0aeb93b5c5afdab5c/packages/quantic/force-app/main/default/lwc/quanticFacet/quanticFacet.js#L239

SimonMilord: this is a good point: 

> IMO the end user should be very aware that the value introduced to a property is invalid. Falling back to the default value can make invalid properties, easily, go under the radar/unnoticed which can be a bit problematic.

SimonMilord (approved):
👌

packages/quantic/force-app/main/default/lwc/quanticFoldedResultList/quanticFoldedResultList.js
erocheleau: Ok well the fallback was just an idea, but do you prefer using the `parseInt(value, 10)` to the `Number(value)` method?

I would prefer it I think because using the Number constructor could give you non Integer AFAIK, so you'd have 3.5 numberOfFoldedResults potentially?

erocheleau (approved):

PR #2817, opened 2023-04-06
ci(atomic): include headless build in e2e test directory

btaillon:
https://coveord.atlassian.net/browse/KIT-2399

### DIFF
--- a/packages/atomic/stencil.config.ts
+++ b/packages/atomic/stencil.config.ts
@@ -107,9 +107,7 @@ export const config: Config = {
       copy: [
         {src: 'pages', keepDirStructure: false},
         {src: 'themes'},
-        isDevWatch
-          ? {src: 'external-builds', dest: 'build/headless'}
-          : {src: ''},
+        {src: 'external-builds', dest: 'build/headless'},
         {
           src: '../../../node_modules/@salesforce-ux/design-system/assets/icons/{doctype,standard}/*.svg',
           dest: 'build/assets',


packages/atomic/stencil.config.ts
```diff
@@ -107,9 +107,7 @@ export const config: Config = {
       copy: [
         {src: 'pages', keepDirStructure: false},
         {src: 'themes'},
-        isDevWatch
-          ? {src: 'external-builds', dest: 'build/headless'}
-          : {src: ''},
+        {src: 'external-builds', dest: 'build/headless'},
```
olamothe, line 110: But does that mean we will now ship this on all our official build 🤔  ?

btaillon: Nope, this is only part of the `www` build.

olamothe: Would just changing insights.html to this achieve the same thing ?
```
await insightInterface.initialize({
          insightId: 'fc2ee6e0-8bda-4883-bfc6-ffc1b0ce34c7',
          accessToken: 'xx9446b04d-45d4-44a6-880c-22c6c04573ff',
});
```

btaillon: yes, but we also have pages in Atomic that use `external-builds`, it's just that we weren't running e2e tests on them.

packages/atomic/src/components/insight/index.ts
```diff
@@ -76,4 +76,6 @@ export {
   buildInteractiveResult as buildInsightInteractiveResult,
   InteractiveResult as InsightInteractiveResult,
   loadQuerySetActions as loadInsightQuerySetActions,
+  getOrganizationEndpoints as getOrganizationEndpointsInsight,
+  PlatformEnvironment as PlatformEnvironmentInsight,
```
louis-bompart, line 80: Mh, before we create a duplicate export, is it too late to create an 'atomic-core' bundle or something akin where functionality shared across the board lives?

btaillon: We're gonna have to do changes to bundling later anyway, so I don't feel like this should be done later. For now, I think we should leave it as-is.

louis-bompart: However, removing something exported will be a breaking change.

btaillon: Ah, I see what you mean.

Currently, `@coveo/atomic/insight` already exports a lot of "core" stuff, therefore I don't think we need to make an exception here. I think it'll have to be a breaking change along with everything else, when the time comes.

louis-bompart: Good for me! I just want to be sure that this is a conscious decision

olamothe (approved):

louis-bompart:
@btaillon , Co-Author mention are made in the commit footer/trailler, not the header :P

GitHub should add it automatically in the squash message too IIRC

btaillon:
@louis-bompart Yeah, I didn't look and clicked "Squash and merge" too quickly.

PR #2818, opened 2023-04-06
fix(quantic): condition that detects if more child results are available in collection fixed

mmitiche:
[SFINT-4972](https://coveord.atlassian.net/browse/SFINT-4972)

## The issue

https://user-images.githubusercontent.com/86681870/230460153-5139f3b3-768c-41a2-b655-ac6ace78342b.mov

## The explanation

After loading all the results of a given folded collection we were using the following condition to check whether more child results have been found in the collection:
```
get moreResultsFound() {
    return this?.collection?.children?.length > this.firstChildrenPartition?.length
}
```

Turns out that this condition was not sufficient, for example imagine the following collection:
```
A 
  |  B
```
After loading all the results the same collections becomes:
```
A 
  |  B
     |  C
```

in that case an additional grand child have been found but the condition `moreResultsFound` is `false` because we are only comparing the length of the children of the parent which is not sufficient.
## The solution

I updated the condition in order to make it detect any  change between all the child results of a collection and the first partition children.
```
  get moreResultsFound() {
    return (
      JSON.stringify(this?.collection?.children) !==
      JSON.stringify(this.firstChildrenPartition)
    );
```

https://user-images.githubusercontent.com/86681870/230460226-7a354252-0635-4006-ba22-87256a4f8ef7.mov

[SFINT-4972]: https://coveord.atlassian.net/browse/SFINT-4972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
very nice!

erocheleau (changes requested):
The test looks good but the JSON.stringify is probably not the best solution

erocheleau (approved):

PR #2819, opened 2023-04-07
feat(atomic): added a responsive tab container

nathanlb:
This PR adds a reusable container for atomic tabs (insight & ipx) so that overflow doesn't create a scroll bar but instead moves the overflow tabs to a popover.
Tabs selected from the popover then get displayed alongside the shown tabs.

It also proposes a change in the IPX search section layout to use flexbox instead of grid. It seemed more cooperative this way. The end result is intended to be the exact same as before.

Tests are in progress: 
  - [x] Done

**Example in Insight Panel**

https://user-images.githubusercontent.com/16785453/230664156-48baeb44-3b4d-4cf7-9477-efd19b4fc186.mov

**Example in IPX**

https://user-images.githubusercontent.com/16785453/230664170-74187f2e-11ab-4040-ab81-f04d08b63990.mov


packages/atomic/src/utils/dom-utils.tsx
```diff
@@ -6,3 +6,45 @@ export function rectEquals(r1: DOMRect, r2: DOMRect) {
...
+    right: parseFloat(styles.paddingRight.length ? styles.paddingRight : '0'),
+    bottom: parseFloat(
+      styles.paddingBottom.length ? styles.paddingBottom : '0'
+    ),
+    left: parseFloat(styles.paddingLeft.length ? styles.paddingLeft : '0'),
+  };
+}
+
+/**
+ * Returns the absolute width of an element.
```
louis-bompart, line 27:
```suggestion
 * Returns the absolute height of an element.
```

packages/atomic/src/components/common/tabs/tab-popover.pcss
```diff
@@ -0,0 +1,18 @@
...
+
+tab-popover {
+  position: absolute;
+}
+
+.visibility-hidden {
+  visibility: hidden;
+}
+
+[part~='overflow-tabs'] {
```
louis-bompart, line 11: Why `~=` and not `=` ?

packages/atomic/src/utils/dom-utils.tsx
nathanlb: removed those utils as I'm not using them in the end but good catch 👍

packages/atomic/src/components/common/tabs/tab-popover.pcss
nathanlb: no reason that I can think of. changed to `=`

olamothe (commented):
There's a couple of things I'm wondering:

* I'm not certain there is a need to create full fledge web component for this, and possibly just simple functional component could have worked. In that case, we would not need to work with a more complex web component lifecyle, and no need to hide with `@internal`/"pollute" the global scope of the page with new elements. I'm not 100% sure everything would work correctly (ie: not having to rely on `@state` from stencil), but it would be interesting to at least test it. We should, as a general rule, try to limit ourselves when we create web component to stuff that implementer will actually need to interact with/configure in their HTML page directly. For abstract/only internal components, whenever possible, we should try to use functional components as much as possible.

* There is already a very similar pattern that was already created for segmented facet, which had been approved by UX (Johnny) as the way we wanted to display "long list of filters horizontally". This component (`atomic-segmented-facet-scrollable`) could potentially just be slightly tweaked/adapted, and would work properly for tabs in IPX/insight panel, without the need for a lot of additional  code. It's not a popover, but.... 🤷

nathanlb:
> There's a couple of things I'm wondering:
> 
> * I'm not certain there is a need to create full fledge web component for this, and possibly just simple functional component could have worked. In that case, we would not need to work with a more complex web component lifecyle, and no need to hide with `@internal`/"pollute" the global scope of the page with new elements. I'm not 100% sure everything would work correctly (ie: not having to rely on `@state` from stencil), but it would be interesting to at least test it. We should, as a general rule, try to limit ourselves when we create web component to stuff that implementer will actually need to interact with/configure in their HTML page directly. For abstract/only internal components that we create and inject in the DOM automatically, we should try to use functional components as much as possible.
> * There is already a very similar pattern that was already created for segmented facet, which had been approved by UX (Johnny) as the way we wanted to display "long list of filters horizontally". This component (`atomic-segmented-facet-scrollable`) could potentially just be slightly tweaked/adapted, and would work properly for tabs in IPX/insight panel, without the need for a lot of additional  code. It's not a popover, but.... 🤷

Good questions!

I did indeed start off by trying to create a simpler functional component but the lack of a lifecycle was problematic and I couldn't get it to work at all. Getting to this point took a lot of different attempts. These components are internal at least so they wouldn't affect the quantity of exposed components right? Maybe someone smarter than me can figure it out.

Unfortunately I don't think the segmented (horizontal) facet is close to being a suitable replacement of the current usage of tabs in service for the following reasons:
- They are very different... 😅 so from a migration standpoint it would likely represent another barrier to adoption of our no-code hosted interfaces given clients won't be able to recreate their existing tabs in the new hosted solution.
- The UX designs for atomic insight panel and IPX explicitly has tabs like this.
- The insight panel builder and IPX builder both have tab configurations and leverage a component to configure tabs that is not compatible with the horizontal facet, making this kind of switch costly at this point.
- For the insight panel use case we're trying to have the atomic tabs display and behaviour matching those of the quantic tabs which has this popover design.

IMO given right now this rudimentary tab component is only leveraged by hosted interfaces, it's usage by developers will be internal and at a minimum. Also, the current display behaviour of just having a horizontal scrollbar is quite subpar. The hosted insight panel is rendered in quantic for the end user but the hosted IPX will be atomic and this would be a significant improvement.

What do you think?

olamothe:
> I did indeed start off by trying to create a simpler functional component but the lack of a lifecycle was problematic and I couldn't get it to work at all.

Alright, that's fine then ! I wasn't sure it was possible in the first place (hence my question), and if it complicates things then it pretty much defeats the purpose of using functional components in the first place.

> Unfortunately I don't think the segmented (horizontal) facet is close to being a suitable replacement of the current usage of tabs in service for the following reasons:

Yes, this is not what I'm suggesting (using segmented facet instead of tabs), I was just suggesting adopting the scrollable horizontal container we've created for it. Inside the container, it would be exactly the tabs component as you have right now.

> The UX designs for atomic insight panel and IPX explicitly has tabs like this.

I'd like to challenge the UX team on this one, is it @johnnyCoveo ? 
I'm wondering if it's a conscious decision, or just an oversight.

I fail to understand why there would be 2 different way to displays horizontal filter kinda thing, be it in quantic or atomic, or IPX/insight panel/full search page.

It's pretty much nearly the same end user interaction in my mind, albeit the way the different filters works under the hood are different.

So I'll reach out to UX and come back

packages/atomic/src/components/common/radio-button.tsx
```diff
@@ -53,12 +53,12 @@ export const RadioButton: FunctionalComponent<RadioButtonProps> = (props) => {
 
   return (
     <input
-      {...attributes}
       type="radio"
       onChange={(e) =>
         (e.currentTarget as HTMLInputElement).checked && props.onChecked?.()
       }
       onMouseDown={onMouseDown}
+      {...attributes}
```
lbergeron, line 61: Out of curiosity, why are you appending the attributes after the other ones? Unless I am mistaken this change has no effect.

packages/atomic/src/components/common/tabs/tab-bar.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const containerRelativeRightPosition =
+      this.host.getBoundingClientRect().right;
+    const selectedTabRelativeRightPosition =
+      this.selectedTab?.getBoundingClientRect().right;
+
+    return this.tabsFromSlot.filter((element) => {
+      const isBeforeSelectedTab = selectedTabRelativeRightPosition
+        ? selectedTabRelativeRightPosition >
+          element.getBoundingClientRect().right
+        : null;
```
lbergeron, line 67: Why are you setting `null` here? `isBeforeSelectedTab` has 3 possible values when only 2 are needed. I'd set it to `false` if possible.

packages/atomic/src/components/common/tabs/tab-bar.tsx
```diff
@@ -0,0 +1,186 @@
...
+    });
+  }
+
+  private get displayedTabs() {
+    return this.tabsFromSlot.filter((el) => !this.overflowingTabs.includes(el));
+  }
+
+  private get lastDisplayedTab() {
+    const displayedTabs = this.displayedTabs;
+    return displayedTabs[displayedTabs.length - 1];
```
lbergeron, line 90: Just found out the [at function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) does handle negative indices.

```suggestion
    return this.displayedTabs.at(-1);
```

johnnyCoveo:
@olamothe @nathanlb I think what Nathan have proposed works in the context of an Insight panel and the IPX.

I think that it is important to make a distinction between tabs and horizontal facets, they have different levels in terms of information hierarchy, and having them visually treated differently will help the user better understand it.

Also, in the Insight panel and IPX context, which will be mostly on desktop, a pattern such as clicking on "more" and get more options if very common and easy to control, versus a horizontal scroll, which is more easy to manipulate on mobile (drag and scroll), but not necessarily on desktop since it is not intuitive to drag and scroll.

packages/atomic/src/components/common/radio-button.tsx
nathanlb: good question. Stencil was giving me warnings saying that a certain prop needed to be after others so I did this and there no more warnings

packages/atomic/src/components/common/tabs/tab-bar.tsx
nathanlb: TIL. It's a good suggestion but unfortunately it doesn't look like the project's target version 😅 

Screen Shot 2023-04-13 at 1 26 21 PM

lbergeron: Ah... sorry :sweat_smile:

packages/atomic/src/components/common/radio-button.tsx
lbergeron: The change would have been breaking if one of the 3 attributes declared on the element were also part of the `attributes` object. I wonder if the same warning would have been raised.

olamothe: It's much more logical to have the spread attributes after the others indeed. I did a quick scan of everywhere this radio button is used, and there should not be any breaking changes 👍

packages/atomic/src/components/common/tabs/tab-popover.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  @Method()
+  public togglePopover() {
+    this.isOpen = !this.isOpen;
+  }
+
+  private renderDropdownButton() {
+    const label = this.bindings?.i18n.t('more');
+    const ariaLabel = this.bindings?.i18n.t('popover', {label});
```
olamothe, line 78: I think the `popover` key might not be localized properly, since it talks about `facet`

packages/atomic/src/components/common/tabs/tab-popover.tsx
```diff
@@ -0,0 +1,175 @@
...
+    const buttonClasses = ['relative', 'pb-1', 'mt-1', 'mr-6', 'font-semibold'];
+
+    return (
+      <Button
```
olamothe, line 82: I would have been neat to be able to re-use the `IconButton` component, since it should properly handle accessibility concerns. 

However I guess it does not work exactly as you want here with what you are trying to do here with the label 🤔

packages/atomic/src/components/common/tabs/tab-popover.tsx
```diff
@@ -0,0 +1,175 @@
+import {
+  createPopperLite as createPopper,
+  preventOverflow,
+  Instance as PopperInstance,
+} from '@popperjs/core';
```
olamothe, line 5: I forgot we already had that as a dependency 😅

olamothe (approved):

packages/atomic/src/components/common/tabs/tab-popover.tsx
nathanlb: oups! good catch. thanks

packages/atomic/src/components/common/tabs/tab-popover.tsx
nathanlb: Yeah and it has a different look we'd need to override. I think this is simpler in this case.

dguerinCoveo (approved):
awesome!

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -154,6 +161,10 @@
         display: none;
       }
 
+      atomic-ipx-tabs {
+        width: 100%;
+      }
+
```
dguerinCoveo, line 167: We will have to keep in mind to update those styles in the SIS builder as well when updating the atomic version

PR #2824, opened 2023-04-12
fix(headless): add warning on configuration mismatch between organizationId and organizationEndpoint

olamothe:
This will log a warning if an implementer configure the engine like so:

```typescript
buildSearchEngine({
     organizationId: 'some-org-id',
     organizationEndpoints: getOrganizationEndpoints('another-org-id')
})
```
https://coveord.atlassian.net/browse/KIT-2403

### DIFF
--- a/packages/headless/src/app/engine.ts
+++ b/packages/headless/src/app/engine.ts
@@ -197,7 +197,7 @@ export function buildEngine<
     )
   ) {
     engine.logger.warn(
-      `There is a mismatch between the \`organizationId\` option (${options.configuration.organizationId}) and the organization configured on the \`organizationEndpoints\` option (${options.configuration.organizationEndpoints?.platform}). This could lead to issues that are complex to troubleshoot. Please make sure both values are equivalent.`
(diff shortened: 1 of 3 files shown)


packages/headless/src/app/engine.ts
```diff
@@ -190,6 +191,16 @@ export function buildEngine<
     );
   }
 
+  if (
+    shouldWarnAboutMismatchBetweenOrganizationIDAndOrganizationEndpoints(
+      options
+    )
+  ) {
+    engine.logger.warn(
```
olamothe, line 199: Error message to review, thanks ! 🙇

packages/headless/src/app/engine.ts
```diff
@@ -190,6 +191,16 @@ export function buildEngine<
...
+      `There is a mismatch between the \`organizationId\` option (${options.configuration.organizationId}) and the organization configured on the \`organizationEndpoints\` option (${options.configuration.organizationEndpoints?.platform}). This could lead to issues that are complex to troubleshoot. Please make sure both values are equivalent.`
```
jpmarceau, line 200:
```suggestion
      `There is a mismatch between the \`organizationId\` option (${options.configuration.organizationId}) and the organization configured in the \`organizationEndpoints\` option (${options.configuration.organizationEndpoints?.platform}). This could lead to issues that are complex to troubleshoot. Please make sure both values match.`
```

jpmarceau (approved):
👍

louis-bompart (approved):

PR #2827, opened 2023-04-12
ci: add new versioning process

btaillon:
https://coveord.atlassian.net/browse/KIT-2330

# Purpose
In this PR, I copied and slightly modified the logic from the CLI project responsible for versioning. This new logic replaces Lerna for:
* Bumping the version of each package.
  * Unlike the CLI, we're appending the commit hash instead of an incrementing number.
* Generating changelogs for each package.
  * The new format for changelogs doesn't seem compatible with our current format though.
* Publishing each package to NPM.
  * The new process publishes each commit without creating a version bump commit.
  * Unlike the CLI, we're using the `alpha` tag instead of the `beta` tag.

For now, unlike the CLI, this logic only publishes to a fake NPM registry whose instance is different in each CI run.

# What's next
The next steps are gonna be:
1. Add the commit/release logic.
  	* I would copy the release logic from the CLI project. It would be disabled until the next step.
2. Remove Lerna and old publishing logic.
	  * Also remove the old prerelease logic.
	  * Instead of bumping the version on every commit, we would bump the version by starting a GitHub workflow.
	  	* I would therefore enable `start_environment_automatically` in the deployment pipeline configuration so that it stops asking us before moving on to the next environment.
	  * There may be ways to play it safe, like committing to another branch and commenting the lock/unlock logic, but I don't think it's worth the time investment.
    	* Worst case scenario, we can publish manually or temporarily revert the changes.
3. (Optional, could be done later) Remove build from Jenkins
  	* I would make Jenkins stop building the project and instead download and publish the NPM packages to the CDN.

# Details
More specific changes I made:
* Added two new packages to `utils/`:
  * `verdaccio-starter`
    * copy and pasted from https://github.com/coveo/cli/tree/2f2273c/utils/verdaccio-starter (with most of its options removed)
    * used to start a fake NPM registry so we can confirm that the version bumping process works before removing the old one.
  * `release`
    * copy and pasted from https://github.com/coveo/cli/tree/2f2273c/utils/release
    * used to store executable scripts for the new release process.
      * eventually, we can migrate scripts from the `scripts` directory to `utils/` packages so we don't need to have a long `../` chain.
    * for now, I only needed to copy the `npm-publish-package.mjs` script.
    * I created the `install-npm-proxy.mjs` script which starts Verdaccio and prevents the CI from publishing to NPM.
* Added prerelease step to `master` branch CI.
  * for now, it doesn't actually publish to NPM, but it generates a downloadable artifact of all the packages it would've published so I can verify them.

Another detail is that, when the CI bumps the package versions, bad entries are added to the `package-lock` which request Atomic from the NPM registry.
* I tried to fix this using `npm update`, but calling it on every package (whether in topological order or in the reverse order) doesn't fix the package-lock.
* This may be fixable with Arborist, but not worth the time investment IMO.


nx.json
```diff
@@ -49,6 +49,9 @@
     },
     "promote:sfdx": {
       "dependsOn": ["build"]
+    },
+    "release:phase1": {
+      "dependsOn": ["^release:phase1"]
     }
```
btaillon, line 55: Ensures packages are bumped in the correct order.

.github/actions/setup/action.yml
```diff
@@ -20,6 +20,9 @@ runs:
           packages/atomic/docs
           packages/atomic/loader
           packages/atomic/www
+          packages/atomic/src/components/search/atomic-search-interface/lang/
+          packages/atomic/src/external-builds/
+          packages/atomic/src/generated/
```
btaillon, line 25: Had some errors which I thought were because some files weren't being cached. Regardless, those files should be cached just like any other build artifact.

.github/workflows/masterbot.yml
```diff
@@ -182,3 +182,41 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup
+      - name: Release
+        run: |
+          npm run nx:graph
+          npm run release:install-npm-proxy
+          npm run release:phase1
+        env:
+          # NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
btaillon, line 207: Not needed yet.

package.json
```diff
@@ -19,7 +19,13 @@
     "bump:version:major-prerelease": "node ./scripts/deploy/manual-major-prerelease.mjs",
     "promote:npm:beta": "nx run-many --target=promote:npm:beta",
     "promote:npm:latest": "nx run-many --target=promote:npm:latest",
-    "notify:docs": "node ./scripts/notify-docs/published-ui-kit.mjs"
+    "notify:docs": "node ./scripts/notify-docs/published-ui-kit.mjs",
+    "release": "npm run nx:graph && npm run release:phase0 && npm run release:phase1 && npm run release:phase2",
+    "nx:graph": "nx graph --file=topology.json",
+    "release:install-npm-proxy": "npx -p=@coveo/release install-npm-proxy",
+    "release:phase0": "exit 38",
```
btaillon, line 26: Temporary.

Number corresponds with `ENOSYS 38 Function not implemented` in case you were wondering.

package.json
```diff
@@ -19,7 +19,13 @@
     "bump:version:major-prerelease": "node ./scripts/deploy/manual-major-prerelease.mjs",
     "promote:npm:beta": "nx run-many --target=promote:npm:beta",
     "promote:npm:latest": "nx run-many --target=promote:npm:latest",
-    "notify:docs": "node ./scripts/notify-docs/published-ui-kit.mjs"
+    "notify:docs": "node ./scripts/notify-docs/published-ui-kit.mjs",
+    "release": "npm run nx:graph && npm run release:phase0 && npm run release:phase1 && npm run release:phase2",
+    "nx:graph": "nx graph --file=topology.json",
```
btaillon, line 24: Copied from CLI

utils/release/package.json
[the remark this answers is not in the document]
btaillon: Yeah, once we know for sure the process works for `ui-kit` too (e.g., we finally released to prod), we should refactor with the CLI and extract some stuff to monorepo tools.

utils/verdaccio-starter/src/index.ts
```diff
@@ -0,0 +1,78 @@
+// eslint-disable-next-line node/no-unpublished-import
```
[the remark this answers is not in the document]
btaillon: image

utils/verdaccio-starter/project.json
```diff
@@ -0,0 +1 @@
+{}
```
btaillon, line 1: This entire project is copy and pasted from the CLI, but I removed most of the code in `index.ts`.

utils/verdaccio-starter/src/index.ts
```diff
@@ -0,0 +1,78 @@
...
+import {ConfigRuntime} from '@verdaccio/types';
```
louis-bompart, line 2:
```suggestion
import type {ConfigRuntime} from '@verdaccio/types';
```

Should fix the node/no-unpublished-import iirc

utils/verdaccio-starter/src/index.ts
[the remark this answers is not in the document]
louis-bompart: https://github.com/coveo/ui-kit/pull/2827/files/10d791d71cc9020f29841b1325abe9ea40e3f507#r1170373213

louis-bompart (approved):
LGTM, but check that your copy of utils/* from the CLI are up-to-date, I've done some changes recently due to some CLI release trouble

btaillon:
Closing until I fix this.

PR #2829, opened 2023-04-13
chore(workflows): don't use always() in workflows

pballandras:
Related Issue
DT-5750

## Hi, what's this?
👋 We have come across a problem with using `if: always()` in workflows. The basic gist is that these steps become un-cancellable, possibly creating all sorts of unexpected problems. This is less frequent with GitHub hosted runners, but it can happen.

This PR introduces and workaround this problem by explicitly defining the different states the step can face.
Namely:
* `success()`
* `cancelled()`
* `failure()`

This way, you get the `always()` behaviour without it's possible caveats.

## I don't want this change
In a way, this is more of a "we really recommend this" kind of PR. That is because you may never face the problem I'm talking about, but when you do, you'll want this workaround.

If you decide against merging this, please make sure you mention this if you need support in the future to help us identify this issue faster.

## Final notes
Please review and merge this PR. Since this is a CI change, you don't need to release this change.

This PR is created as part of a large batch. I would appreciate it if you could merge this yourself to ease the amount of followup on my end.

## References
* https://coveo.slack.com/archives/CBP3GTFTQ/p1679495407426389 (Coveo internal)
* https://github.com/orgs/community/discussions/26303

### DIFF
--- a/.github/actions/e2e-atomic-screenshots/action.yml
+++ b/.github/actions/e2e-atomic-screenshots/action.yml
@@ -18,7 +18,7 @@ runs:
         install: false
         record: false
     - uses: actions/upload-artifact@v3
-      if: always()
+      if: cancelled() || failure() || success()
       with:
         name: cypress-result-lists-screenshots
         path: packages/atomic/cypress/screenshots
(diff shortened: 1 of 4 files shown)


louis-bompart (approved):
LGTM!

btaillon (approved):

pballandras:
I'm merging this now, CI is fixed 🎉

PR #2834, opened 2023-04-15
feat(atomic): add option to disable search button when there is no query input

mrrajamanickam-coveo:
https://coveord.atlassian.net/browse/KIT-2286

* In addition to the existing `disableSearch` option to disable the search action 
* This PR adds another option `disableSearchWhenNoQuery` to disable search when there is no query input
 - The search button is enabled only when is any query input and disabled when there is no query


packages/atomic/src/components.d.ts
```diff
@@ -1462,9 +1462,13 @@ export namespace Components {
          */
         "clearFilters": boolean;
         /**
-          * Whether to prevent the user from triggering a search from the component. Perfect for use cases where you need to disable the search conditionally, like when the input is empty.
```
mrrajamanickam-coveo, line 1465: @jpmarceau @srosenberg-coveo Please review the doc changes. 
* Starting the sentence with `Whether` (conjunction) feels weird. Although it is not invalid grammatically I think it reads weird. 
* But I do see that this pattern is used in many places in this repo. So inclined to follow the same for consistency. Let me know your thoughts ..

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -153,11 +153,18 @@ export class AtomicSearchBox {
...
+   * To be used when you need to disable the search conditionally.
+   * For the case when you need to disable the search on no query input refer to {@link disableSearchWhenNoQuery}.
    */
   @Prop({reflect: true}) public disableSearch = false;
 
+  /**
+   * Disable search button when there is no query input.
+   */
+  // TODO(major release): Consider combining with `disableSearch` prop and/or defaulting to true.
+  @Prop({reflect: true}) public disableSearchWhenNoQuery = false;
```
olamothe, line 166: An alternative, which could be more flexible, would be to change the option to something akin to "minimumInputContent", which would disable search when there is less than X characters in the search box.

This is a semi frequent request from implementers, that is being able to disable search when there is less than 2, or 3 characters. 

0 character is the most frequent need, but I think we would be able to cover a bit more ground with an option that would accept a number instead.

packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
```diff
@@ -473,15 +478,42 @@ describe('Search Box Test Suites', () => {
...
+      QuerySummaryAssertions.assertHasPlaceholder();
+      CommonAssertions.assertConsoleError(false);
+    });
+  });
+
+  describe('with disableSearchWhenNoQuery set to true', () => {
+    beforeEach(() => {
+      new TestFixture()
+        .with(addSearchBox({props: {'disable-search-when-no-query': 'true'}}))
+        .with(addQuerySummary())
```
olamothe, line 495: Is the query summary component needed in this test suite ?

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -153,11 +153,18 @@ export class AtomicSearchBox {
...
+   * Prevent the user from triggering a search from the component.
+   * To be used when you need to disable the search conditionally.
+   * For the case when you need to disable the search on no query input refer to {@link disableSearchWhenNoQuery}.
    */
   @Prop({reflect: true}) public disableSearch = false;
 
+  /**
+   * Disable search button when there is no query input.
+   */
+  // TODO(major release): Consider combining with `disableSearch` prop and/or defaulting to true.
```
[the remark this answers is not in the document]
olamothe: In any case, if we want to deprecate/modify default behaviour, I would remove the TODO from the codebase, and instead create a ticket in https://coveord.atlassian.net/browse/KIT-2157 with all the details needed.

packages/atomic/src/components.d.ts
jpmarceau: Hello @mrrajamanickam-coveo 👋 
Yeah, `Whether...` is a pattern we use tyically in reference docs, not just in ui-kit. We use it specifically for booleans. If we start with a verb (e.g., "Prevent ..."), it can seem, at first glance, like we're documenting a function rather than a boolean.
Writing "Whether..." for booleans seems common in the industry, so I don't find it weird 🙂 For example, [Algolia](https://www.algolia.com/doc/api-reference/api-parameters):
Screen Shot 2023-04-17 at 12 07 36 PM

Do you think that makes sense?

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
btaillon: I'm starting to think we should ask UX about this.

I can kind of imagine some purpose for disabling the search box while it's empty, e.g.:
* In case somebody accidentally presses enter.
* Because it just doesn't make much logical sense to have search results without a query.

However, I can't really think of a purpose for disabling search under a certain number of characters, and I'm not sure if that's something we even want to allow since I can only imagine it being an annoyance.

Concerning the naming of this parameter, I think we could still make it a boolean, but instead make it positive (`enable` instead of `disable`).

For example: `@Prop({reflect:true}) public allowEmptyQueries = true`

packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
```diff
@@ -473,15 +478,42 @@ describe('Search Box Test Suites', () => {
...
-    QuerySummaryAssertions.assertHasPlaceholder();
-    CommonAssertions.assertConsoleError(false);
+    it('there are no search suggestions or errors', () => {
+      SearchBoxAssertions.assertHasSuggestionsCount(0);
+      QuerySummaryAssertions.assertHasPlaceholder();
+      CommonAssertions.assertConsoleError(false);
+    });
+  });
+
+  describe('with disableSearchWhenNoQuery set to true', () => {
```
mrrajamanickam-coveo, line 491: TODO: add test to check disableQuery prop overrides this prop

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
olamothe: > However, I can't really think of a purpose for disabling search under a certain number of characters, and I'm not sure if that's something we even want to allow since I can only imagine it being an annoyance.

Often time, the reason is $  😉 .

Some client (for good and/or bad reasons) want to limit their consumption against the Coveo platform, because that's what contracts are based on. And so they go on some crusade to find everything they can configure/tweak to achieve that goal. And trying to block/prevent "incomplete" search requests is something we see from time to time.

As for if there's good UX reasons behind it, I don't think so.

packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
mrrajamanickam-coveo: looks like its not, will remove - thanks for catching - meant to check after I got the test working.

packages/atomic/src/components.d.ts
mrrajamanickam-coveo: Thanks for the explanation @jpmarceau 
Will revert to using `Whether ..`

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
mrrajamanickam-coveo: What do you think about having this prop control triggering search suggestions based on query input length instead of the search action itself. That way if the user wanted to search for small words less than the set length, they can still do so (without being confused/annoyed about why the search is not enabled) but it would also satisfy the goal of limiting resource consumption?

packages/atomic/src/components.d.ts
```diff
@@ -1470,13 +1470,17 @@ export namespace Components {
...
+          * Whether to prevent the user from triggering a search from the component. Perfect for use cases where you need to disable the search conditionally. For the specific case when you need to disable the search based on length of query input refer to {@link minInputToEnableSearch}.
```
jpmarceau, line 1473:
```suggestion
          * Whether to prevent the user from triggering a search from the component. Perfect for use cases where you need to disable the search conditionally. For the specific case when you need to disable the search based on the length of the query, refer to {@link minInputToEnableSearch}.
```

packages/atomic/src/components.d.ts
```diff
@@ -1470,13 +1470,17 @@ export namespace Components {
...
          */
         "disableSearch": boolean;
         /**
           * Whether to interpret advanced [Coveo Cloud query syntax](https://docs.coveo.com/en/1814/) in the query. You should only enable query syntax in the search box if you have good reasons to do so, as it requires end users to be familiar with Coveo Cloud query syntax, otherwise they will likely be surprised by the search box behaviour.  When the `redirection-url` property is set and redirects to a page with more `atomic-search-box` components, all `atomic-search-box` components need to have the same `enable-query-syntax` value.
          */
         "enableQuerySyntax": boolean;
+        /**
+          * Minimum length of query input to enable the search button. E.g. Setting it to 3 would enable search only after 3 characters have been typed in query input.
```
jpmarceau, line 1481:
```suggestion
          * The minimum query length required to enable the search button. For example, setting this parameter to `3` would enable search only after 3 characters have been typed in the search box.
```

packages/atomic/src/components.d.ts
```diff
@@ -4131,13 +4135,17 @@ declare namespace LocalJSX {
          */
         "clearFilters"?: boolean;
         /**
-          * Whether to prevent the user from triggering a search from the component. Perfect for use cases where you need to disable the search conditionally, like when the input is empty.
+          * Whether to prevent the user from triggering a search from the component. Perfect for use cases where you need to disable the search conditionally. For the specific case when you need to disable the search based on length of query input refer to {@link minInputToEnableSearch}.
```
jpmarceau, line 4138:
```suggestion
          * Whether to prevent the user from triggering a search from the component. Perfect for use cases where you need to disable the search conditionally. For the specific case when you need to disable the search based on the length of the query, refer to {@link minInputToEnableSearch}.
```

packages/atomic/src/components.d.ts
```diff
@@ -4131,13 +4135,17 @@ declare namespace LocalJSX {
...
          */
         "disableSearch"?: boolean;
         /**
           * Whether to interpret advanced [Coveo Cloud query syntax](https://docs.coveo.com/en/1814/) in the query. You should only enable query syntax in the search box if you have good reasons to do so, as it requires end users to be familiar with Coveo Cloud query syntax, otherwise they will likely be surprised by the search box behaviour.  When the `redirection-url` property is set and redirects to a page with more `atomic-search-box` components, all `atomic-search-box` components need to have the same `enable-query-syntax` value.
          */
         "enableQuerySyntax"?: boolean;
+        /**
+          * Minimum length of query input to enable the search button. E.g. Setting it to 3 would enable search only after 3 characters have been typed in query input.
```
jpmarceau, line 4146:
```suggestion
          * The minimum query length required to enable the search button. For example, setting this parameter to `3` would enable search only after 3 characters have been typed in the search box.
```

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -154,10 +154,17 @@ export class AtomicSearchBox {
 
   /**
    * Whether to prevent the user from triggering a search from the component.
-   * Perfect for use cases where you need to disable the search conditionally, like when the input is empty.
+   * Perfect for use cases where you need to disable the search conditionally.
+   * For the specific case when you need to disable the search based on length of query input refer to {@link minInputToEnableSearch}.
```
jpmarceau, line 158:
```suggestion
   * For the specific case when you need to disable the search based on the length of the query, refer to {@link minInputToEnableSearch}.
```

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -154,10 +154,17 @@ export class AtomicSearchBox {
...
    */
   @Prop({reflect: true}) public disableSearch = false;
 
+  /**
+   * Minimum length of query input to enable the search button.
+   * E.g. Setting it to 3 would enable search only after 3 characters have been typed in query input.
```
jpmarceau, line 164:
```suggestion
   * The minimum query length required to enable the search button.
   * For example, setting this parameter to `3` would enable search only after 3 characters have been typed in the search box.
```

jpmarceau (approved):
I have some string adjustment suggestions, but this is already quite clear. Well done Mohan 👍

louis-bompart (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
btaillon: yeah, we can add the option to stop suggestions later

btaillon (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
mrrajamanickam-coveo: https://coveo.slack.com/archives/G7HD9P48H/p1681832351521629

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -744,6 +766,7 @@ export class AtomicSearchBox {
             ref={(el) => (this.inputRef = el as HTMLInputElement)}
             bindings={this.bindings}
             value={this.searchBoxState.value}
+            title={this.tooltipText}
```
mrrajamanickam-coveo, line 769: wonder if `title` and `ariaLabel` could be combined in some way because[ looks like aria label is prioritized over title by screen readers](https://stackoverflow.com/a/27953625) so the title message wouldn't be accessible

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -154,9 +154,16 @@ export class AtomicSearchBox {
 
   /**
    * Whether to prevent the user from triggering a search from the component.
-   * Perfect for use cases where you need to disable the search conditionally, like when the input is empty.
+   * Perfect for use cases where you need to disable the search conditionally.
```
mrrajamanickam-coveo, line 157: @jpmarceau Please review the updated text and the new locale entry below.

mrrajamanickam-coveo:
@olamothe @louis-bompart @btaillon Please review the latest changes.

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -154,9 +154,16 @@ export class AtomicSearchBox {
...
+   * For the specific case when you need to disable the search based on the length of the query, refer to {@link minimumQueryLength}.
    */
-  @Prop({reflect: true}) public disableSearch = false;
+  @Prop({reflect: true, mutable: true}) public disableSearch = false;
+
+  /**
+   * The minimum query length required to enable search.
+   * For example, to disable the search for empty queries set this to `1`.
```
jpmarceau, line 164:
```suggestion
   * For example, to disable the search for empty queries, set this to `1`.
```

jpmarceau (approved):
👍

packages/atomic/src/components/common/search-box/search-box-common.tsx
```diff
@@ -91,7 +91,13 @@ export class SearchBoxCommon {
     item.query && this.props.clearSuggestions();
   }
 
-  public getSearchInputLabel() {
+  public getSearchInputLabel(minimumQueryLength = 0) {
```
mrrajamanickam-coveo, line 94: Not very happy with having to pass in `minimumQueryLength` param which is a child class only prop into the parent class func. Let me know of any suggestions..

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
mrrajamanickam-coveo: refactored to reuse the same msg for both title and arialabel

srosenberg-coveo (approved):

packages/atomic/src/components/common/search-box/search-box-common.tsx
olamothe: Nothing quick come to mind... 🤔

mrrajamanickam-coveo:
All TODOs have been addressed and PR is ready for final review.

mrrajamanickam-coveo:
> Waiting on code owner review from coveo/service-core.
@olamothe who can I ping for this ?

louis-bompart:
> > Waiting on code owner review from coveo/service-core.
> > @olamothe who can I ping for this ?

Odd. @nathanlb? Do your team either monitor request review, or do you use [auto-assignment]?(https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team#about-auto-assignment)

nathanlb (approved):

PR #2838, opened 2023-04-19
feat(organizationendpoints): support organization endpoints for hosted-pages components

olamothe:
* Add `admin` endpoint in Headless > `getOrganizationEndpoints` utility function
* Leverage it to load the pages inside `<atomic-hosted-pages>` and `<atomic-simple-builder>`

https://coveord.atlassian.net/browse/KIT-2405
https://coveord.atlassian.net/browse/CDX-1410


louis-bompart (approved):
Fixes https://coveord.atlassian.net/browse/CDX-1410 too, added it to desc.

PR #2839, opened 2023-04-20
docs(headless): add links in orgEndpoints warnings

jpmarceau:
https://coveord.atlassian.net/browse/KIT-2411


olamothe (approved):
I would also add this link to the JSDoc for the related options: 

https://github.com/coveo/ui-kit/blob/master/packages/headless/src/app/engine-configuration.ts#L69

and 

https://github.com/coveo/ui-kit/blob/master/packages/headless/src/app/engine-configuration.ts#L88

PR #2841, opened 2023-04-21
chore: create dependency-review.yml

JPLachance:
+ Configure Dependency Review

Dependency review helps you understand dependency changes and the security impact of these changes at every pull request. It provides an easily understandable visualization of dependency changes with a rich diff on the "Files Changed" tab of a pull request.

The warning for .github/workflows/dependency-review.yml is expected.

https://coveord.atlassian.net/browse/DEF-657

J:DEF-657


louis-bompart (approved):

JPLachance:
I'm one click away from removing the setting that prevents me from merging. Can I do it? I'll put it back after.

btaillon:
> I'm one click away from removing the setting that prevents me from merging. Can I do it? I'll put it back after.

I can force merge it if you want.

btaillon (approved):

PR #2842, opened 2023-04-24
docs(atomic): Update docs for field property of atomic-result-badge

karoomoogon-coveo:
https://coveord.atlassian.net/browse/DOC-12604

Multi-value field was causing errors in code sandbox when trying to be displayed within field properties of badges.
Added an example for multi-value field.


packages/atomic/src/components/search/result-template-components/atomic-result-badge/atomic-result-badge.tsx
```diff
@@ -25,6 +26,13 @@ import {ResultContext} from '../result-template-decorators';
  * </atomic-result-badge>
  * ```
  *
+ * The content of a multi-value field can be displayed as follows:
```
jpmarceau, line 29:
```suggestion
 * The contents of a multi-value field can be displayed as in the following example:
```

jpmarceau (approved):
Thanks Krishna 👍

louis-bompart (approved):

PR #2843, opened 2023-04-24
docs(headless): random improvements

spiracha-coveo:
Made some minor updates to strings for certain controllers to link to a better article added a string where there wasn't any before


jpmarceau (approved):
👍

PR #2845, opened 2023-04-25
fix: pre-process svg inside of atomic-ipx-button

pathiery:
[SVCINT-2203](https://coveord.atlassian.net/browse/SVCINT-2203)

Updated the `atomic-ipx-button` component to pre-process svg.
It will: 
- Load the provided svg in the dom
- Check if it is actually an svg
- Compute the svg's width and height if set
- Strip all the styles from the svg (including width and height) since we want the svg to adapt to the button size
- If the svg had a width and height set, set the svg's viewBox to the given width and height

image

[SVCINT-2203]: https://coveord.atlassian.net/browse/SVCINT-2203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
+++ b/packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
@@ -8,6 +8,8 @@ import {
 import {Button} from '../../common/button';
 import {Bindings} from '../../search/atomic-search-interface/atomic-search-interface';
 
+const numberOrPixelValuePattern = new RegExp(/^(?=.*(?:\d+|px)$).*$/);
+
 /**
  * @internal
  */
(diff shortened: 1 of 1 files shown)


olamothe:
We could (should?) modify CODEOWNERS -> 

```
/packages/atomic/src/components/ipx @coveo/service-integrations
```

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
```diff
@@ -126,10 +128,38 @@ export class AtomicIPXButton implements InitializableComponent {
   private getIcon(icon: string) {
     const initialDiv = document.createElement('div')!;
     initialDiv.innerHTML = icon;
-    initialDiv
-      .querySelector('svg')
-      ?.setAttribute('fill', 'var(--atomic-on-primary)');
-
+    const iconElement = initialDiv.querySelector('svg');
+    if (!iconElement) {
+      return initialDiv.innerHTML; // handle null case from .querySelector
```
olamothe, line 133: nitpick: Personally find the comments superfluous and that they don't add much, since what the code does on a line by line basis is kinda straightforward anyway. But they don't really explain the initial problem/initial bug that you are trying to fix here.

What would be more explanatory about "what" the code should be fixing would be to add a cypress test for that initial bug, if possible.

dguerinCoveo (approved):
lgtm

olamothe (approved):

packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
pathiery: That makes a lot of sense, let me see what I can do !

pathiery: As I thought, it's really hard to track this kind of thing via cypress since it's purely visual. I'll move to a single line which gives a broader explanation and doesn't interfere with code reading as much

PR #2848, opened 2023-04-25
chore(headless): update coveo.analytics dependency

olamothe:
Newer version of coveo.analytics (the dependency that is core for all analytics events sent through Headless) changed the way "client ID" are generated.

It now relies on this: https://github.com/uuidjs/uuid 

That dependency itself internally relies on global crypto module being available (specifically `window.crypto.getRandomValues`), or throws an error.

Trying to make a long story short, Headless is officially usable in Node, for server side rendering.

A client just trying to perform a simple query with Headless in that scenario would get a fatal error originating from `uuid` package.

So for Node 18-20, we need to make this globally available by requiring the module.

As a bonus for Node 16 support, we  can just do a bit of plumbing to map `crypto.webcrypto.getRandomValues` to `crypto.getRandomValues`.

We could also do that polyfill directly in `coveo.analytics`, but I think it makes sense to patch at Headless level, since... well, that's where we need support for Node JS.

https://coveord.atlassian.net/browse/KIT-2426

### DIFF
--- a/packages/headless/src/api/analytics/analytics-crypto-polyfill.ts
+++ b/packages/headless/src/api/analytics/analytics-crypto-polyfill.ts
@@ -1,0 +1,19 @@
+interface CryptoModuleInNode {
+  getRandomValues?: typeof crypto.getRandomValues;
+  webcrypto?: {getRandomValues: typeof crypto.getRandomValues};
+}
+
+const getGlobalCrypto = () => global.crypto as CryptoModuleInNode;
+
+export const polyfill = () => {
+  if (!getGlobalCrypto()) {
+    global.crypto = require('crypto');
+  }
+
+  if (!getGlobalCrypto().getRandomValues && getGlobalCrypto().webcrypto) {
+    global.crypto.getRandomValues =
+      getGlobalCrypto().webcrypto!.getRandomValues.bind(
+        getGlobalCrypto().webcrypto
+      );
+  }
+};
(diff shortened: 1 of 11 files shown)


erocheleau (changes requested):
Just missing to also add the same polyfill to case-assist.index.ts

packages/headless/src/api/analytics/analytics-crypto-polyfill.ts
```diff
@@ -0,0 +1,19 @@
+interface CryptoModuleInNode {
+  getRandomValues?: typeof crypto.getRandomValues;
+  webcrypto?: {getRandomValues: typeof crypto.getRandomValues};
+}
+
+const getGlobalCrypto = () => global.crypto as CryptoModuleInNode;
+
+export const polyfill = () => {
```
mrrajamanickam-coveo, line 8: adding a TODO / doc about when this can be removed in future would be useful

mrrajamanickam-coveo: Naming it specifically would be helpful too even though this is the only polyfill e.g `polyfillCryptoNode17`

louis-bompart: We could: https://nodejs.org/api/globals.html#crypto_1
Starting with Node 20, crypto as a global property is enabled by default.
Assuming we keep on supporting 2 majors version of Node at a time, when Node 22 will roll out in ~Y24Q4, we could remove the polyfill.

erocheleau (approved):

PR #2849, opened 2023-04-26
chore(atomic): upgrade storybook to v7

olamothe:
All storybook related dependency should be at latest with this PR.

https://coveord.atlassian.net/browse/KIT-2351


utils/atomic-storybook/package.json
```diff
@@ -3,25 +3,38 @@
...
+    "@storybook/react": "^7.0.7",
+    "@storybook/react-webpack5": "^7.0.7",
+    "@storybook/testing-library": "^0.0.14-next.2",
+    "lit-analyzer": "^2.0.0-pre.2",
+    "lodash": "4.17.21",
     "monaco-editor-webpack-plugin": "6.0.0",
-    "lit-analyzer": "1.2.1",
+    "prop-types": "^15.8.1",
     "react": "16.14.0",
     "react-dom": "16.14.0",
```
louis-bompart, line 30: Can you check if we can 'bump' react up so that it's synced up/deduped with our other project.
@btaillon migrated storybook out due to this, so if we manage to sync them up, we could move storybook back into the fold and reduce the complexity

olamothe:
I would propose to have the move of storybook back to atomic package as a separate PR, to keep it shorter. 

What do you guys think ?

louis-bompart:
> I would propose to have the move of storybook back to atomic package as a separate PR, to keep it shorter.
> 
> What do you guys think ?

agreed, baby-steps.
Updating storybook and removing react from there is already a leap.

louis-bompart (approved):

Issue #2840, opened 2023-04-21
[Bug]: Type issues with pino-std-serializers


vikrant-harness:
### Which product are you using?

Headless

### product version

2.8.3

### bug description

Getting this typecheck error while using coveo headless 2.8.3

error TS2688: Cannot find type definition file for ‘pino-std-serializers’.
  The file is in the program because:
    Entry point for implicit type library ‘pino-std-serializers’
Found 1 error.

### Steps to reproduce

Added coveo headless 2.8.3 to the repo

used coveo headless to generate engine config and search box controllers

running typecheck gives the above error 

manually adding pino-std-serializers solves the issue but we do not want to add something we do not use as dev dependency 

### Relevant log output

```shell
error TS2688: Cannot find type definition file for ‘pino-std-serializers’.
  The file is in the program because:
    Entry point for implicit type library ‘pino-std-serializers’
Found 1 error.
```

y-lakhdar:
Hi @vikrant-harness,

Are you using yarn in your project? I was able to reproduce the issue with the latest version of Headless when using yarn instead of npm.

When running `yarn install`, the `index.d.ts` is missing from the `pino-std-serializers` package. I am still investigating the root cause, but based on the error message you provided, I see several alternatives to avoid the issue:

1. Extend the `typeRoots` property by adding the path of `pino-std-serializers` definition file.
   
```json
// tsconfig.json
...
"typeRoots": ["./node_modules/@types", "./node_modules/@types/pino-std-serializers/node_modules"],
...

```

2. Manually installing the type definition 
```
yarn add @types/pino-std-serializers -D
```

3. Update to the latest version of Yarn (Yarn Berry) as it provides improved type-checking features that can help resolve these type of issues.

I would recommend the third option to avoid adding "patch solutions". I will check if I can get more info on the root cause

vikrant-harness:
Hi @y-lakhdar , The first solution worked for us and as for the second and third are currently not an option for us , Can we find the root cause of the issue as ideally we should not be adding this as well .

y-lakhdar:
Hi @vikrant-harness , 
I have addressed the issue by bumping [pino](https://www.npmjs.com/package/pino)'s version.
Fixed in [2.16.1](https://github.com/coveo/ui-kit/compare/@coveo/headless@2.16.0...@coveo/headless@2.16.1)

y-lakhdar:
Let me know if you still experience the issue

PR #2851, opened 2023-04-27
chore: bump pino version to v8

y-lakhdar:
Update pino version to fix issue #2840.

https://coveord.atlassian.net/browse/CDX-1442

### DIFF
--- a/packages/headless/src/app/logger.ts
+++ b/packages/headless/src/app/logger.ts
@@ -11,7 +11,7 @@ export interface LoggerOptions {
    * Changes the shape of the log object. This function will be called every time one of the log methods (such as `.info`) is called.
    * All arguments passed to the log method, except the message, will be pass to this function. By default it does not change the shape of the log object.
    */
-  logFormatter?: (object: object) => object;
+  logFormatter?: (object: Record<string, unknown>) => Record<string, unknown>;
   /**
    * Function which will be called after writing the log message in the browser.
    */
--- a/package-lock.json
+++ b/package-lock.json
@@ -14046,7 +14046,6 @@
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/abort-controller/-/abort-controller-3.0.0.tgz",
       "integrity": "sha512-h8lQ8tacZYnR3vNQTgibj+tODHI5/+l06Au2Pcriv/Gmet0eaj4TwWH41sO9wnHDiQsEj19q0drzdWdeAHtweg==",
-      "peer": true,
       "dependencies": {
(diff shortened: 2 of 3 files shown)


packages/headless/src/app/logger.ts
```diff
@@ -11,7 +11,7 @@ export interface LoggerOptions {
    * Changes the shape of the log object. This function will be called every time one of the log methods (such as `.info`) is called.
    * All arguments passed to the log method, except the message, will be pass to this function. By default it does not change the shape of the log object.
    */
-  logFormatter?: (object: object) => object;
+  logFormatter?: (object: Record<string, unknown>) => Record<string, unknown>;
```
louis-bompart, line 14: 🤔 https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UBKEDGB7ATgEwDwGdhMBLAOwHMAaAV1IGtT0B3UgPgG4AobNAGwENM0DKUJQAZunQAudACMAVmmBQuk9FHz9gxfOOIR8sdkA
Technically breaking change, `{}` does not satisfies `Record<string, unknown>`.

olamothe: We do end up exporting the logger, and so you are right that this is a breaking change.

However in this case I'm "fairly certain" this would have no impact/no one relies on that formatter.

I'm curious why we need to change it though @y-lakhdar  ? 🤔

y-lakhdar: In the latest version of pino, the log property accepts a Record<string, unknown> only  and...
```
Type 'object' is not assignable to type 'Record<string, unknown>'.
```

I'll check to see if we can avoid doing a breaking change by using `{}` instead since it is accepted by pino

btaillon: Chances of this actually causing issues seems pretty low, and fix looks like it'll be intuitive in most cases, so YOLO I think

btaillon (approved):

packages/headless/src/app/logger.ts
louis-bompart: (to be clear, it won't keep me up at night, I just flagged the thing just in case)

louis-bompart:
LGTM for Dependency Review bypass:
Only vulnerable vector on the package-lock is `"packages/quantic/node_modules/sfdx-cli/node_modules/json5"` which is a dev dep.

louis-bompart (approved):

louis-bompart:
Reverting some tweaks and update of quantic to keep it lean, as we accept the vuln on our devdeps

PR #2852, opened 2023-04-27
ci: remove snyk remnant, add codeql

louis-bompart:
- Remove Snyk
- Add CodeQL


.github/workflows/codeql-analysis.yml
louis-bompart, line 1: We can either keep the default codeQL and let it do its thing and remove this file, or disable the default and use this

olamothe: Why wouldn't we want the default ? What's different between the two ?

louis-bompart: In our case, not much (if anything): https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages#about-the-codeql-analysis-workflow-and-compiled-languages

btaillon (approved):

PR #2856, opened 2023-05-01
fix(atomic): fixed leaky styling from tab components

nathanlb:
Some ipx and atomic tabs components were leaking global css outside of the shadow DOM. To fix I did 2 things:
1. Enabled shadow dom to `tab-popover` component
2. Removed unnecessary css from `ipx-tabs` and `insight-tabs` components which was leaking due to those components not using shadow-dom. It's necessary to the proper functioning of the `tab-bar` component they have as a child that these do not use a shadow-dom.

Added a rudimentary test to the insight test suite to check that only the expected number of script tags are added. I want to create a better test that validates script tags added by any component but currently the atomic insight use case doesn't have a good per-component test setup. That'll require more work I deem out of scope of this fix. We'll do it in the near future though, especially before we start offering a production-grade atomic insight panel.


louis-bompart (approved):
lgtm

olamothe (approved):

jelmedini (approved):

PR #2857, opened 2023-05-01
fix(SFINT-4968): fix the InsightTokenProvider

erocheleau:
It was requesting classes that were not Global from the C4SF package to generate a token.


mmitiche (approved):
LGTM 👌

SimonMilord (approved):
Looks good!

PR #2858, opened 2023-05-01
feat(headless): support did you mean with query suggestion fallback 

olamothe:
This feature is an experimentation we want to run with some client.

It replaces the "classic" did you mean feature typically powered by the index. The classic algo essentially relies on the distance (number of permutation) between the user query and which keywords are actually present in the index corpus.

The new algo uses what has already been developed for query suggestions by the ML team, with more modern support for typo tolerance, based on analytics data.
 
This feature requires 3 things in the platform:

* Feature flag enabled for an organization in launch darkly.
* A working/valid machine learning model for query suggestions
* A query parameter (`enableFallbackSearchOnEmptyQueryResults`). The query parameter can be enforced at the query pipeline level (Override query parameter feature)

For this experimentation, we can rely on the query pipeline override feature, so we don't really need to add support for the query parameter in Headless just yet.

When the feature is activated, the search API will send a new parameter in the response: `changedQuery`. We can use it to decide what to do when processing the request, and dispatch the appropriate actions to make sure end users correctly see the `Query was corrected to XYZ` in the front end, proper analytics getting logged, and that the search box content update itself automatically.

https://coveord.atlassian.net/browse/KIT-2434


packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -263,6 +257,55 @@ export class AsyncSearchThunkProcessor<RejectionType> {
...
+      },
+      automaticallyCorrected: true,
+      originalQuery,
+      analyticsAction: logDidYouMeanAutomatic(),
+    };
+  }
+
+  private processFallbackQueryResultsDidYouMeanAutoCorrection(
+    originalFetchedResponse: FetchedResponse
+  ): ExecuteSearchThunkReturn {
```
mrrajamanickam-coveo, line 290: No promise reject for this one?

packages/headless/src/api/search/search/query-corrections.ts
```diff
@@ -30,3 +30,8 @@ export interface WordCorrection {
    */
   correctedWord: string;
 }
+
+export interface ChangedQuery {
```
louis-bompart, line 34: JSDOC to mark this as 'experimental' and not officially supported.

louis-bompart (commented):
Code LGTM, but it needs either:
 - to be merged/released somewhere else than the main/official release/branch
 - everything related must have a disclaimer.
 
 Otherwise doing breaking change in there would be a bit more haphazardous

packages/headless/src/features/search/search-actions-thunk-processor.ts
olamothe: No, in this code path we are not retrying any query, it's already been corrected at the API level and we have the proper results already. We just need to adjust client side state to what the query has been corrected to.

For the classic did you mean feature, client side code is in charge of doing the retry with the correction (which means there's an async request + request failure possibility)

btaillon (approved):

mrrajamanickam-coveo (approved):

PR #2859, opened 2023-05-02
feat(quantic): expose the fetchClassificationOnChange & fetchDocumentSuggestionOnChange properties

SimonMilord:
[SFINT-4827](https://coveord.atlassian.net/browse/SFINT-4827)

**As a User**
I want have the ability to make the Quantic Case Classification component automatically fetch the new case classifications and document suggestions when its value changes.

**Acceptance criteria**

These two new boolean properties should be exposed in the Quantic Case Classification component. 

We should cover this new feature with e2e tests.

**Technical details**

The Headless Case Field controller already supports this feature, we only need to add it to the Quantic Case Classification component.

[SFINT-4827]: https://coveord.atlassian.net/browse/SFINT-4827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**DEMO:**

https://user-images.githubusercontent.com/73316533/236263693-614ef6f5-cae1-4fa2-9f50-5563f1dfc544.mov


packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -889,4 +897,94 @@ describe('quantic-case-classification', () => {
       });
     });
   });
+
+  describe('when the field value changes and updates-to-fetch is not undefined', async () => {
```
mmitiche, line 901: `updates-to-fetch` is  a name of a variable that you are using in the component, however in the description of this test it's a bit hard to understand and not very meaningful.
I would suggest to update this with a very simple description that a person can understand without having to look at what the component does.

mmitiche (changes requested):
The solutions looks good, good job!
just some work needed with the e2e tests.

mmitiche:
In the demo you have attached in this PR, you are showing how new classifications and document suggestions are being fetched after updating some case inputs.
That's not really the purpose of the change you added to the Quantic Case Classification component.

The purpose is to make the Quantic Case Classification component fetch new classification and document suggestions when its value changes.

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
SimonMilord: good point! will change it to something more descriptive

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -889,4 +893,343 @@ describe('quantic-case-classification', () => {
       });
     });
   });
+
+  describe('when a field value changes and the fetchClassificationsOnChange property is set to true', async () => {
```
erocheleau, line 897: This describe text needs to be updated no?

Wouldn't it be:
`when the fetchClassificationsOnChange property is set to true`

erocheleau (approved):
Looks good! Just a title change I think

packages/quantic/cypress/e2e/default-1/case-classification/case-classification-expectations.ts
```diff
@@ -170,6 +170,44 @@ function caseClassificationExpectations(selector: CaseClassificationSelector) {
...
+            'predictions'
+          );
+          expect(
+            caseClassificationBody.fields[field].predictions
+          ).to.deep.equal(predictions);
+        })
+        .logDetail(
+          'should fetch new case classifications after the value changes'
+        );
+    },
```
mmitiche, line 193: I don't see really the point of testing the content of the response here. 
All what we need to test here is that there is indeed a new case classification request sent after selecting a new case classification. the content of the response doesn’t really matter.

It could just be something like this.
```suggestion
    fetchClassificationsAfterValueChange: () => {
      cy.wait(InterceptAliases.CaseClassification)
        .logDetail(
          'should fetch new case classifications after the value changes'
        );
    },
```

packages/quantic/cypress/e2e/default-1/case-classification/case-classification-expectations.ts
```diff
@@ -170,6 +170,44 @@ function caseClassificationExpectations(selector: CaseClassificationSelector) {
...
+            'documents'
+          );
+          expect(documentSuggestionResponseBody.documents).to.deep.equal(
+            mockDocuments
+          );
+        })
+        .logDetail(
+          'should fetch new document suggestions after the value changes'
+        );
+    },
```
mmitiche, line 210: Same remark, I don't see the point of testing the content of the response, what's the reason behind adding these checks on the response?

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -889,4 +893,343 @@ describe('quantic-case-classification', () => {
...
+            coveoDefaultField,
+            allOptions.slice(0, suggestionsCount)
+          );
+          fetchClassifications();
+          Expect.displaySelectTitle(true);
+          Expect.displaySelectInput(false);
+          Expect.numberOfSuggestions(suggestionsCount);
+          Expect.correctSugestionsOrder(allOptions.slice(0, suggestionsCount));
+          Expect.numberOfInlineOptions(0);
+          Expect.logClickedSuggestion(firstSuggestionIndex, true);
```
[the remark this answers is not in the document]
mmitiche: yep exactly this for the auto selection of the classification with the highest confidence.

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -889,4 +893,343 @@ describe('quantic-case-classification', () => {
...
+          const clickedIndex = 1;
+
+          Actions.clickSuggestion(clickedIndex);
+          Expect.logClickedSuggestion(clickedIndex);
+          Expect.logUpdatedClassificationFromSuggestion(
+            coveoDefaultField,
+            clickedIndex
+          );
+          Expect.correctValue(allOptions[clickedIndex].value);
+        });
```
mmitiche, line 1040: Why do we need this section?
I don't think we need a step to select a suggestion because the goal of the describe section is to click on an option from the select input and confirm that a new classification request has been sent.

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -889,4 +893,343 @@ describe('quantic-case-classification', () => {
...
+
+        scope('when fetching suggestions', () => {
+          const firstSuggestionIndex = 0;
+
+          mockCaseClassification(
+            coveoDefaultField,
+            allOptions.slice(0, suggestionsCount)
+          );
+          mockDocumentSuggestion(mockDocuments.documents);
```
mmitiche, line 1108: Why do we need to mock the document suggestions returned?
The only thing that we need to test is that there is a document suggestion request made  whenever a case classification is selected.
The content of the document suggestion doesn't really matter thus I don't think we need to mock the document suggestions.

packages/quantic/force-app/examples/main/lwc/exampleQuanticCaseClassification/exampleQuanticCaseClassification.js
```diff
@@ -60,11 +61,25 @@ export default class ExampleQuanticCaseClassification extends LightningElement {
       description: 'The maximum number of suggestions to be displayed',
       defaultValue: 3,
     },
+    {
+      attribute: 'fetchClassificationOnChange',
+      label: 'Fetch classifications on field value change',
+      description:
+        'Whether the QuanticCaseClassification component should automatically fetch new case classifications when the input field values changes',
```
mmitiche, line 68: we should update the description, it's not about the input value like previously discussed.

packages/quantic/force-app/examples/main/lwc/exampleQuanticCaseClassification/exampleQuanticCaseClassification.js
```diff
@@ -60,11 +61,25 @@ export default class ExampleQuanticCaseClassification extends LightningElement {
...
+      defaultValue: false,
+    },
+    {
+      attribute: 'fetchDocumentSuggestionOnChange',
+      label: 'Fetch document suggestion on field value change',
+      description:
+        'Whether the QuanticCaseClassification component should automatically fetch new document suggestions when the input field values changes',
```
mmitiche, line 75: we should update the description, it's not about the input value like previously discussed.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -107,6 +102,20 @@ export default class QuanticCaseClassification extends LightningElement {
    * @defaultValue `'Select an option'`
    */
   @api messageWhenValueMissing = this.labels.selectOption;
+  /**
+   * Whether the Quantic Case Classification component should automatically fetch new case classifications when its value changes.
```
mmitiche, line 106: we are not really used to mention the name of the component in the property descriptions.
```suggestion
   * Indicates whether the component should automatically fetch new case classifications when its value changes.
```

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -107,6 +102,20 @@ export default class QuanticCaseClassification extends LightningElement {
...
+   * @api
+   * @type {boolean}
+   * @defaultValue `false`
+   */
+  @api fetchClassificationOnChange = false;
+  /**
+   * Whether the Quantic Case Classification component should automatically fetch new document suggestions when its value changes.
```
mmitiche, line 113:
```suggestion
   * Indicates whether the component should automatically fetch new document suggestions when its value changes.
```

packages/quantic/cypress/e2e/default-1/case-classification/case-classification-expectations.ts
SimonMilord: Mmmm I thought there would be value in checking that we do get what we are expecting from the request, but I guess you are right in that it is not to us to test here.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
SimonMilord: ok noted!

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
SimonMilord: yeah if we dont test for response doesnt make sense then to mock documents

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
SimonMilord: You are right, I must have forgotten to remove this while refactoring. Good catch!

mmitiche (approved):
Looks good👌
just one small remark to address.

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -889,4 +893,299 @@ describe('quantic-case-classification', () => {
...
+        });
+
+        scope('when fetching suggestions', () => {
+          const firstSuggestionIndex = 0;
+
+          mockCaseClassification(
+            coveoDefaultField,
+            allOptions.slice(0, suggestionsCount)
+          );
+          mockDocumentSuggestion([{}]);
```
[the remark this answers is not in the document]
SimonMilord: we seem to need the mock for the intercept in the expectation to work with the fetchSuggestions

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -1,9 +1,11 @@
 import {fetchClassifications} from '../../../page-objects/actions/action-fetch-classifications';
+import {fetchSuggestions} from '../../../page-objects/actions/action-get-suggestions';
```
mmitiche, line 2: I don't think this is necessary.

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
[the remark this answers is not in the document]
mmitiche: why do you need to call `fetchSuggestions`? I don't think it's necessary.

packages/quantic/force-app/examples/main/lwc/exampleQuanticCaseClassification/exampleQuanticCaseClassification.html
```diff
@@ -1,16 +1,45 @@
...
+      <c-action-fetch-classifications
+        slot="actions"
+        disabled={notConfigured}
+        engine-id={config.engineId}
+      ></c-action-fetch-classifications>
+      <c-action-fetch-suggestions
+        slot="actions"
+        disabled={notConfigured}
+        engine-id={config.engineId}
+      ></c-action-fetch-suggestions>
```
mmitiche, line 21: Same thing here, don't think this is necessary. 
In the test we only want to confirm that a new request for document suggestion has been made when the value of classification changes, we do not need to mock this document suggestion request. we only need to to confirm that the request has been made.

packages/quantic/force-app/main/default/lwc/quanticCaseClassification/quanticCaseClassification.js
```diff
@@ -316,6 +324,17 @@ export default class QuanticCaseClassification extends LightningElement {
       });
   }
 
+  /**
+   * Whether we want to fetch new case classifications or document suggestions when we update the field.
+   * @returns {Object}
```
mmitiche, line 329: It's nice to be precise with the types when we can.
```suggestion
   * @returns {{caseClassifications: boolean, documentSuggestions: boolean}}
```

mmitiche (approved):
Looking good!

PR #2860, opened 2023-05-02
fix(atomic,headless): fix support for exact phrase stemming in quickview

olamothe:
There was no proper support for "exact phrase" highlighting + stemming in quickview. 

The symptom would be that, for example, the query (in our sample org/test page): `author=Jules Vernes` and `query="the bank"` (ie: exactly the two keywords "the bank" right next to one another), which opens a PDF of the book `Around the World in 80Days`, before this fix, you would see highlights for:

* "The Bank" 
* "the bank"
* "The bank"
* "the Bank"
* "the"
* "bank" 
* etc.

We would just do literal exact match in the document.
With this fix, users will see only one highlighting for "the bank", with all its variation in the document grouped together.

https://coveord.atlassian.net/browse/KIT-2438


packages/headless/src/api/search/search/stemming.ts
```diff
@@ -4,3 +4,12 @@
  * E.g., for the query "indexing consultant", where a result contains "Techzample is a consulting firm specializing in search indexes. It employs consultants that can analyze and improve your search index.", the `TermsToHighlight` might be `{ "indexing": ["indexes", "index"], "consultant": ["consulting", "consultants"] }`.
  */
 export type TermsToHighlight = {[originalTerm: string]: string[]};
+
+/**
+ * Phrases to highlight (with [stemming](https://docs.coveo.com/en/1576)) in the results.
+ *
+ * E.g., for the exact phrase "indexing consultant", where a result contains "Techzample is a consulting firm specializing in search indexes. It employs consultants that can analyze and improve your search index.", the `PhrasesToHighlight` might be `{ "indexing consultant": {"indexing": ["indexes", "index"], "consultant": ["consulting", "consultants"]}}`.
```
mrrajamanickam-coveo, line 11: A note on how it differs from Terms to highlight would be helpful. E.g. I am guessing its `{ "indexing consultant"` vs  ``{ "indexing"`? In that case isn't `"consultant":` a term?

packages/headless/src/api/search/search/stemming.ts
```diff
@@ -4,3 +4,12 @@
...
+ */
+export declare type PhrasesToHighlight = {
+  [originalPhrase: string]: TermsToHighlight;
```
mrrajamanickam-coveo, line 14: Curious, why is the `originalPhrase` an array? Could there be multiple?

olamothe: Yes, you could search for `"the bank" "the city"` which is 2 exact phrases

packages/headless/src/api/search/search/stemming.ts
mrrajamanickam-coveo: Got it, thanks

packages/atomic/cypress/e2e/result-list/result-components/quickview.cypress.ts
```diff
@@ -142,6 +142,25 @@ describe('Quickview Component', () => {
...
+    beforeEach(() => {
+      new TestFixture()
+        .with(addFacet({field: 'author'}))
+        .with(addSearchBox())
+        .withHash('q="the%20bank"&f-author=Jules%20Verne')
+        .with(addQuickviewInResultList())
+        .init();
+    });
+
+    it('should display only highlight button for the phrase with multiple occurrences', () => {
```
mrrajamanickam-coveo, line 155: does this test that the exact phrase is highlighted?

mrrajamanickam-coveo (approved):

packages/atomic/cypress/e2e/result-list/result-components/quickview.cypress.ts
olamothe: It performs a query for the exact phrase "the bank", with author=Jules Verne (line 150) (a pdf book for around the world in 90 days).

Then looks at the content of the highlight button, which says there's 23 occurrences for the exact phrase "the bank"

mrrajamanickam-coveo: Ok got it so relies on the summary text on number of occurrences rather than verifying that the text is present and highlighted?

olamothe: Yeah, otherwise we'd have to replicate the same logic as the component directly in the test: As explained in the summary, the keyword highlighting is not exactly matched, it relies on stemming information from the index. 

And so we'd have to add complex-ish logic in the test file that's very similar to what's in `quickview-word-highlight` if we wanted to verify how keywords are highlighted inside the iframe.

VS verifying that the summary is correct seems much simpler.

PR #2862, opened 2023-05-02
ci: added new release process

btaillon:
https://coveord.atlassian.net/browse/KIT-2415

I copy and pasted the release process from https://github.com/coveo/cli (see first commit), then I made some changes to adapt it to ui-kit.

Here are the changes in this PR:
1. Added a new manually triggered workflow called "Create release", copied from the CLI.
    * Triggering the workflow publishes packages to NPM with the alpha tag and pushes a version bump commit.
    * Package versions are determined with the new versioning system.
    * Unlike the CLI, we don't have/need an incremental release tag (e.g., `release-1`, `release-2`, etc).
    * Jenkins should detect those version bumps just like the old ones and creates a deployment package.
      * *Jenkins normally publishes packages to NPM, but there's already conditions that should make it skip publishing without exiting*.
2. Moved the old version bump workflow to a manually triggered workflow called "Create release (Lerna)".
    * That means releases are no longer automatic, but we can still trigger them if the new workflow doesn't work.

~~**Important**: Before merging this PR, I will need to enable "Requires branches to be up to date before merging". See https://github.com/coveo/cli/pull/1261~~

Edit: I misunderstood this. The CI takes care of enabling/disabling "Requires branches to be up to date before merging".

I will merge [KIT-2330-c ](https://github.com/coveo/ui-kit/pull/2827) before this PR.


.github/workflows/git-lock-fail.yml
```diff
@@ -0,0 +1,13 @@
+name: 'master lock validator'
```
btaillon, line 1: Explained here: https://github.com/coveo/cli/pull/1261

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+        default: ''
+      debug:
+        description: 'Add DEBUG=* to the env if true'
+        type: boolean
+        default: false
+        required: false
+
+jobs:
+  release:
+    environment: 'Release'
```
btaillon, line 17: I created a "Release" environment just like the CLI with the below secrets.

utils/release/npm-publish-package.mjs
```diff
@@ -68,7 +68,7 @@ await (async () => {
   let currentNpmVersion = new SemVer(
     privatePackage
       ? '0.0.0' // private package does not have a npm version, so we default to the 'lowest' possible
-      : await describeNpmTag(packageJson.name, 'latest')
```
btaillon, line 71: Our old process waited for QA approval before updating the `latest` tag, so I decided to keep that behaviour by making the release process here publish with the `beta` tag instead.

Pre-release versions are published with the `alpha` tag.

olamothe (approved):

PR #2869, opened 2023-05-04
fix: prevent atomic-focus-detector css glitch on searchbox render

y-lakhdar:
## Issue
When the atomic-search-box is rendered, the atomic-focus-detector children are added to the DOM before the CSS class `display: content` is being applied. This causes the children to glitch

Before the render method is called on atomic-focus-detector
image

after the render method is called on atomic-focus-detector
image

## Proposed Solution
This is more a patch than anything else, but the simplest solution without any CSS glitches.


packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
```diff
@@ -388,7 +388,10 @@ export class AtomicInsightSearchBox {
   public render() {
     return (
       <SearchBoxWrapper disabled={this.disableSearch}>
-        <atomic-focus-detector onFocusExit={() => this.clearSuggestions()}>
+        <atomic-focus-detector
+          style={{display: 'contents'}}
```
olamothe, line 392: Can we remove the `display: contents` applied inside `atomic-focus-detector` then ?

olamothe (approved):

packages/atomic/src/components/insight/atomic-insight-search-box/atomic-insight-search-box.tsx
y-lakhdar: I don't know... I would keep it as a fallback in case someone forgets to add the property at this level.

btaillon (approved):

louis-bompart (approved):

PR #2870, opened 2023-05-05
feat(insight): added button for the full search page

drosenfeld-coveo:
[SVCC-2949](https://coveord.atlassian.net/browse/SVCC-2949)

Added triangular button component that could be used to link to the full search page

Screen Shot 2023-05-05 at 1 40 57 PM

without decimals

Screen Shot 2023-05-08 at 1 32 20 PM

[SVCC-2949]: https://coveord.atlassian.net/browse/SVCC-2949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.pcss
nathanlb, line 1: fractions of pixels are usually not noticeable so I tend to round them off

nathanlb (approved):
Looks pretty good to me!

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.pcss
drosenfeld-coveo: it looked better to me with the fractions but maybe it was just in my head 😕

nathanlb: Mind sharing a screenshot? You might be right

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  @Prop({mutable: true}) public tooltip = '';
+
+  public render() {
+    return (
+      <IconButton
+        partPrefix="full-search"
+        style="outline-neutral"
+        icon={ArrowFull}
+        ariaLabel="full-search"
```
btaillon, line 24: Don't forget to add the corresponding string to `locales.json`

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.tsx
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @internal
+ */
+@Component({
+  tag: 'atomic-insight-full-search-button',
+  styleUrl: 'atomic-insight-full-search-button.pcss',
+  shadow: true,
+})
+export class AtomicInsightFullSearchButton {
+  @Prop({mutable: true}) public clickCallback: () => void = () => {};
```
btaillon, line 14: Will this always just be a redirection to another page, or are there some instances where this will basically just be an expand button in a single-page application?

If it's always just a redirection, then I think that we should use the `a` tag instead of the `button` tag, which allows people to right click and press "Open in new tab", but also conveys to accessibility tools that clicking it will exit the current page.

nathanlb: As of now the only usage will be in the builder preview to reflect a real functioning button rendered by quantic in salesforce so it will be no-op. When we start providing Atomic insight panels outside of the builder then it's still unclear what this button will be used for or if it will be allowed at all.

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.pcss
drosenfeld-coveo: added to description

nathanlb: I'm usually a stickler for pixel perfection but both images look the same to me. 😅 I'll leave it up to you

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.tsx
olamothe: I would propose then just removing that property, it does not seem meaningful to add one yet ?

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.pcss
```diff
@@ -0,0 +1,28 @@
+@import '../../../global/global.pcss';
+
+:host {
+  &::part(full-search-button) {
```
olamothe, line 4: Wouldn't it be possible to use tailwind classes for a lot of those ?

There's also hardcoded colors which should be variables.

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.pcss
```diff
@@ -0,0 +1,28 @@
...
+    background-color: #949aa2;
+    height: 27px;
+    width: 27.75px;
+    clip-path: polygon(0 0, 100% 100%, 100% 0);
+    border-radius: 0;
+    border: none;
+    position: absolute;
```
olamothe, line 11: The fact that this is absolutely positioned means that there's only one place where it can be rendered, and you cannot really move it around to place it differently in the interface.

I would propose to not have that kind of precise hardcoded styles directly on the component itself, and to instead use non-component section to do that kinda thing.

Another way to go about this could be to have a new slot on the insight-interface to place the button.

drosenfeld-coveo: what did you have in mind exactly for the slot? 🤔

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.pcss
btaillon: IMO we should use `rem` instead of `px`. If somebody had to make their browser text size larger to be able to read text, they probably need this button to be bigger too.

packages/atomic/src/components/insight/atomic-insight-full-search-button/atomic-insight-full-search-button.pcss
```diff
@@ -0,0 +1,20 @@
+@import '../../../global/global.pcss';
+
+:host {
+  --full-search-background-color: #949aa2;
```
btaillon, line 4: The contrast ratio of this color against a perfectly white background is really bad, I think we should use a darker grey here.

Also, if possible, I think we should avoid using locally-defined colors and re-use variables from `coveo.css`/`accessibility.css`. I think it's fine to re-defined variables though.

Example:
```suggestion
  --full-search-background-color: var(--atomic-neutral-dark);
```

drosenfeld-coveo: The colours need to match the button in salesforce so I used their colours: https://github.com/coveo/salesforceintegration/tree/develop/force-app/main/default/lwc/openFullSearchButton

drosenfeld-coveo: I changed the colours but we want it to darken wen you mouse over so i added a darker colour, let me know what you think

PR #2871, opened 2023-05-05
ci: fix new versioning strategy for ui-kit

btaillon:
https://coveord.atlassian.net/browse/KIT-2330

The new versioning strategy for ui-kit is adapted from the CLI's versioning strategy. The CLI's versioning strategy is to:
* **Phase 0**: Lock the main branch
* **Phase 1**: For each individual workspace:
  1. bump the package version using `npm version`
  2. generate changelogs
  3. regenerate (a.k.a., "**reify**") the root `package-lock.json`
  4. publish the package to NPM
* **Phase 2**: Commit changes and unlock the main branch.

Unfortunately, this strategy has some incompatibilities with ui-kit. In this PR, I address these two issues:
1. Regenerating the `package-lock.json` right after bumping `@coveo/atomic-angular` causes errors.
2. We do not want to publish the `@coveo/atomic-angular`, but we do want to bump its version.
  * When we build the `@coveo/atomic-angular` workspace, it generates a `dist` directory which contains a new `package.json` with the same version and dependencies as itself. We want to publish that directory instead of the workspace.
  * The CI's strategy publishes the same workspace it just bumped.

My solution was to change the strategy to be like so:
* **Phase 0**: Unchanged (Lock the main branch)
* **Phase 1**: For each individual workspace, bump the package version and generate changelogs
  * We bump the package by modifying `package.json` directly instead of calling `npm version`.
    * An advantage of this is that the `package-lock.json` at the very end of the process no longer modifies a bunch of other packages. It **only** updates versions. We get `package-lock.json` diffs of like <100 lines instead of >30,000 lines like Lerna.
    * This is essentially what `npm version` does anyway. The different is that, when we reify later on, we do so in topological order, whereas `npm version` would do so in alphabetical order. 
* **Phase 2**: For each individual workspace, publish the package to NPM
  * Where `@coveo/atomic-angular-builder` redefines the task to:
    1. Rebuild the `dist/` directory.
    2. Execute the publish script in the `dist/` directory.
* **Phase 3**: Regenerate the root `package-lock.json`
* **Phase 4**: Unchanged (Commit changes and unlock the main branch)


.github/workflows/masterbot.yml
```diff
@@ -202,7 +202,7 @@ jobs:
         run: |
           npm run nx:graph
           npm run release:install-npm-proxy
-          npm run release:phase1
```
btaillon, line 205: Phases 1-3 always need to be executed together, so I combined them.

packages/atomic-angular/angular.json
```diff
@@ -35,5 +35,8 @@
       }
     }
   },
-  "defaultProject": "@coveo/atomic-angular"
+  "defaultProject": "@coveo/atomic-angular",
+  "cli": {
+    "analytics": false
+  }
```
btaillon, line 41: At some point, angular prompted me to choose whether I wanted analytics or not, and it automatically appended that line.

It's not the first time it does this, so I'm including it in this PR once and for all.

packages/atomic-angular/project.json
```diff
@@ -1,3 +1,32 @@
...
+    "release:phase2": {
+      "dependsOn": ["release:phase1", "build-new-version"],
+      "executor": "nx:run-commands",
+      "options": {
+        "parallel": false,
+        "commands": [
+          "cp ../../../../../.npmrc .",
+          "npm link ../../../../../node_modules/@coveo/release",
+          "npx -p=@coveo/release npm-publish"
+        ],
```
btaillon, line 27: `packages/atomic-angular/projects/atomic-angular/dist` is not a workspace. It's even excluded from the repo by `.gitignore`. Therefore, when we execute a script inside of it, it doesn't inherit from the root `.npmrc` and it doesn't automatically link `@coveo/release`.

packages/headless/esbuild.mjs
```diff
@@ -2,6 +2,7 @@ import {build} from 'esbuild';
 import alias from 'esbuild-plugin-alias';
 import {readFileSync, promises} from 'node:fs';
 import {resolve} from 'node:path';
+import resolveModule from 'resolve';
```
btaillon, line 5: When the package version was being bumped by `npm version`, the generated `package-lock` would sometimes hoist/unhoist certain packages, which caused `coveo.analytics` to move from the root `node_modules` to the Headless' `node_modules` and back randomly.

I found this simple fix, which isn't necessary anymore but feels cleaner than directly referencing `node_modules`, therefore I'm keeping it.

utils/release/bump-package.mjs
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env node
```
btaillon, line 1: This script contains the first half of the former `npm-publish-package.mjs`.

The main difference is that I added the `modifyPackageJson` function and I replaced `npmBumpVersion` in favor of manually modifying `package.json`.

utils/release/npm-publish-package.mjs
```diff
@@ -1,222 +1,24 @@
 #!/usr/bin/env node
```
btaillon, line 1: This is the second half of the former `npm-publish-package.mjs`. Essentially, nothing changed.

utils/release/reify.mjs
```diff
@@ -0,0 +1,102 @@
...
+import {dirname, resolve} from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+/**
+ * Current strategy: reify each package (along with their dependants transitively) in topological (parents->dependants) order.
+ *
+ * Other (untested) strategies that may work:
+ * * Strategy A: reify each package individually in topological (parents->dependants) order.
+ * * Strategy B: reify all packages at once.
+ */
```
btaillon, line 13: This script is probably overly complex, since strategy A or strategy B may also work.

However, I've wasted enough time on this, this works very well and we may eventually switch to `pnpm` anyway, so YOLO.

utils/release/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "allowJs": true,
     "checkJs": true,
     "skipLibCheck": true,
-    "noEmit": true
+    "noEmit": true,
+    "lib": ["ESNext"]
```
btaillon, line 10: Needed to use `flatMap`.

packages/atomic-angular/project.json
louis-bompart: 🤔 why not add it to the ws in the root package.json? (and leave it gitignored)

packages/headless/esbuild.mjs
louis-bompart: I reckon you should be able to manage with just good ol' `require`
Do something like
```ts
import {createRequire} from 'node:module';
import {dirname, resolve} from 'node:path';
const require = createRequire(import.meta.url);

// ...
'coveo.analytics': (() => {
  const moduleDir = dirname(require.resolve('coveo.analytics.js/package.json'));
  const {module} = require('coveo.analytics.js/package.json');
  return resolve(moduleDir, module);
})()

louis-bompart (commented):
lgtm, let's shave https://www.npmjs.com/package/resolve tho

packages/atomic-angular/project.json
btaillon: 1. It has the same name as the parent package. If I rename the parent, a bunch of stuff would probably break and need to be fixed.
2. `dist` is generated by `npm run build`, and workspaces are included (AFAIK) in `package-lock.json`, therefore the `package-lock.json` would be invalid until somebody runs `npm run build`. Semantically, I don't like this, and I'm worried it could get flaky.

packages/headless/esbuild.mjs
btaillon: TIL we can use `require` in ESM like that. Will do.

louis-bompart (approved):

Issue #2863, opened 2023-05-02
[Bug]: Loosing browser history when selecting a tab for the first time


marbor3:
### Which product are you using?

Headless

### product version

2.13.1

### bug description

When navigation to a page with tabs, and selecting a tab for the first time, browser history gets reset.
Continues navigation thrught the tabs works good, but I can go back only to a point when the first tab was selected

### Steps to reproduce

Setup headless react sample from https://github.com/coveo/ui-kit/tree/master/packages/samples/headless-react
Switch to any page, like about or standalone searchbox.
Then swtich to Search page, `#tab=all` will get preselected
Try to navigate using browser back button - retriggers search but you cannot go back to where you were.

On a page without tabs component this works

marbor3:
Works like a charm!. Thanks @y-lakhdar

PR #2873, opened 2023-05-09
fix: replaceState on firstSearchExecuted

y-lakhdar:
https://coveord.atlassian.net/browse/KIT-2453

Fix for issue #2863
Having a tab in the search-interface will cause the url hash to automatically change from `/` to `#tab=<default_tab>`.
This addition to the url history will prevent the user to go before `/`.

The solution is to prevent having these 2 states in the history and instead to replace `/` with `#tab=<default_tab>`.


olamothe (approved):

packages/samples/headless-react/src/components/url-manager/url-manager.ts
```diff
@@ -21,6 +26,19 @@ export function bindUrlManager(engine: SearchEngine) {
...
     const hash = `#${urlManager.state.fragment}`;
+
+    if (!statusControllers.state.firstSearchExecuted) {
+      // The purpose of using `replaceState()` instead of `pushState()` in this case is to ensure that the URL reflects the current state of the search page on the first interface load.
+
+      // If `pushState()` were used instead, you could possibly enter in a history loop, which could result in having to click the back button multiple times without being able to  return to a previous page.
+      // This situation happens when you have a component like the Tab component that adds a new state to the browser history stack.
+
+      // Replacing the current state of the browser history with a new state, effectively updating the URL without adding a new entry to the history stack.
+      // For more info, visit https://docs.coveo.com/en/headless/latest/usage/synchronize-search-parameters-with-the-url/
```
jpmarceau, line 37:
```suggestion
      // If `pushState()` were used instead, users could possibly enter in a history loop, having to click the back button multiple times without being able to return to a previous page.
      // This situation happens with components such as the Tab component, which adds a new state to the browser history stack.
      // This replaces the current state of the browser history with a new state, effectively updating the URL without adding a new entry to the history stack.
      // See https://docs.coveo.com/en/headless/latest/usage/synchronize-search-parameters-with-the-url/
```

jpmarceau (approved):

packages/samples/headless-react/src/components/url-manager/url-manager.ts
jpmarceau: I'll add details to the handwritten docs too https://docs.coveo.com/en/headless/latest/usage/synchronize-search-parameters-with-the-url/ 👍

btaillon (approved):

PR #2874, opened 2023-05-09
chore(headless): initial work to cleanup headless bundles

olamothe:
So I've done quite a bit of testing around the work that will be necessary to make our bundles leaner.

One thing I've realized is everything around three shaking will require quite a lot of manual work to help bundle tools (esbuild, webpack, rollup)  identify functions/modules/classes that do not have side effects ( https://esbuild.github.io/api/#tree-shaking // https://esbuild.github.io/api/#pure ). Right now, I'd say only about 5 to 10% of Headless package is tree shakable, which is far lower that what is should be.

In the first phase, what I'd like to achieve is to simply changes the way we do imports so that we can shake off some weights from each produced bundles, notwithstanding any tree shaking capabilities.

This in itself will be a lot of work, as we need to juggle some imports that are used in hundreds of files.

This PR contains: 
* Adding "metafile" when we build packages. These files contains information about how esbuild built the project. This is what I've been using to manually identify the import chain so I can start refactoring how we do imports internally. I've put them in the `dist` bundle, which means we would now ship those. This can be helpful I think, to add tooling down the line around all of this, when we do a release. Downside is that we add weights to npm install. An alternative would be to put them in a different folder, which we could git ignore. Open to any idea you guys would have.

* Starting to do some minor import changes, in order to remove `productRecommendation` use case from `searchPage` use case.

https://coveord.atlassian.net/browse/KIT-2452

### DIFF
--- a/packages/headless/esbuild.mjs
+++ b/packages/headless/esbuild.mjs
@@ -1,6 +1,6 @@
 import {build} from 'esbuild';
 import alias from 'esbuild-plugin-alias';
-import {readFileSync, promises} from 'node:fs';
+import {readFileSync, promises, writeFileSync} from 'node:fs';
 import {resolve} from 'node:path';
 import {umdWrapper} from '../../scripts/bundle/umd.mjs';
 import {apacheLicense} from '../../scripts/license/apache.mjs';
@@ -64,13 +64,16 @@ const browserEsmForAtomicDevelopment = Object.entries(useCaseEntries).map(
     const outDir = getUseCaseDir('../atomic/src/external-builds', useCase);
     const outfile = `${outDir}/headless.esm.js`;
 
-    return buildBrowserConfig({
-      entryPoints: [entryPoint],
-      outfile,
-      format: 'esm',
-      watch: devMode,
-      minify: false,
-    });
+    return buildBrowserConfig(
+      {
+        entryPoints: [entryPoint],
+        outfile,
+        format: 'esm',
+        watch: devMode,
+        minify: false,
+      },
+      outDir
+    );
   }
 );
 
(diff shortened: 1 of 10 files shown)


packages/headless/esbuild.mjs
```diff
@@ -131,39 +141,49 @@ function buildBrowserConfig(options) {
     ],
     ...options,
   });
+  outputMetafile(`browser.${options.format}`, outDir, out.metafile);
```
btaillon, line 144: Similarly to the Stencil definitions file, I think it may be useful to include the metafile in Git/PRs. It looks like it would make it easy to identify changes in the bundle at a pretty low effort cost.

packages/headless/src/features/configuration/search-configuration-actions-loader.ts
```diff
@@ -1,6 +1,7 @@
 import {PayloadAction} from '@reduxjs/toolkit';
 import {RecommendationEngine} from '../../app/recommendation-engine/recommendation-engine';
-import {configuration, pipeline, searchHub} from '../../app/reducers';
+import {configuration} from '../../app/reducers';
+import {pipeline, searchHub} from '../../app/reducers';
```
btaillon, line 4: Did you mean to split those?

Suggestion to revert the change:
```suggestion
import {configuration, pipeline, searchHub} from '../../app/reducers';
```

btaillon (approved):

packages/headless/esbuild.mjs
olamothe: I want to work on this at a later stage: But yes, eventually I want to add reports directly in PR to see changes. 

I've explored some tools that will allow us to do this (eg: https://esbuild.github.io/analyze/)

PR #2875, opened 2023-05-09
refactor(headless): move configuration reducers to shared file for all types of engines

olamothe:
Part of the work to make specific Headless bundle smaller.

One main culprit for pollution is the single `reducers.ts` file, which gets referenced by many files in the project.

It contains all feature for different use cases, and having everything imported through a singular files is problematic.

We will need to break up that singular file into multiple ones. 

The `configuration` is a feature that is used by all types of engine: all use cases need to support managing state property related to organizationId, accessToken, platform URL, etc.

So we can export it through a `common-reducers` file, without fear of polluting with something that not all use cases will utilize.

My end goal is to have something like this:

- common-reducers.ts -> reducers used by every type of engines 
- search-reducers.ts -> all reducers which are needed specifically for search page use cases
- insight-reducers.ts -> all reducers which are needed specifically for insight panel use cases
- recommendation-reducers.ts -> etc...
- product-recommendation-reducers.ts -> etc...
- case-assist-reducers.ts -> etc...

It will be possible for some feature to be exported by multiple different "use cases" reducers. These will be reserved for feature which can overlap **some** use cases, but not **all** of them.

For example:

```ts
import {pagination as paginationFromSearchUseCase} from './search-reducers';
import {pagination as paginationFromInsightUseCase} from './insight-reducers'

// true
paginationFromSearchUseCase === paginationFromInsightUseCase

// There should not be any need to import both types of pagination in a single file, this is just for demonstration
```

# Edit

After working on this some more on a separate branch, the concept of having files such as `search-reducers.ts` and `insight-reducers.ts ` etc (that simply import/re-export stuff) is un-necessary and complicate things.

We can just import from specific feature/slice file directly.

https://coveord.atlassian.net/browse/KIT-2455


btaillon (approved):
Neat! Might be worth writing a one-time script to auto-refactor as many reducers as possible at once and open a big PR.

y-lakhdar (approved):

PR #2876, opened 2023-05-09
fix(atomic-storybook): add missing dependency to target dev when runn…

aprudhomme-coveo:
…ing atomic-storybook:dev

"dev" target was missing a dependency. When running `npx nx run atomic-storybook:dev` before every building, assets.json was never created.

Since "dev" and "build" are basically the same except for the command, I wonder if there is a way to refactor the two so they use the same dependencies.

https://coveord.atlassian.net/browse/KIT-2457

### DIFF
--- a/utils/atomic-storybook/project.json
+++ b/utils/atomic-storybook/project.json
@@ -46,7 +46,11 @@
       }
     },
     "dev": {
-      "dependsOn": ["^build", "validate:storybook"],
+      "dependsOn": [
+        "^build",
+        "validate:storybook",
+        "build:storybook:assets-list"
+      ],
       "inputs": [],
       "outputs": [],
       "executor": "nx:run-commands",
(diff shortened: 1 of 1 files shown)


louis-bompart:
@aprudhomme-coveo , looks like your PR name might've been a bit long, try to make it concise ;)

louis-bompart (commented):
LGTM! @btaillon , I wonder do you know if there'd be a way to refactor the 'dependsOn' between the build task and this one? 🤔

btaillon:
> LGTM! @btaillon , I wonder do you know if there'd be a way to refactor the 'dependsOn' between the build task and this one? 🤔

Could extract the common dependencies in a `nx:noop` target. It could be named something like `prepare-and-validate`.

btaillon (approved):

PR #2878, opened 2023-05-10
feat(atomic): refactored smart snippets components for reusability

nathanlb:


packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet/smart-snippet-common.tsx
```diff
@@ -0,0 +1,168 @@
...
+  getMaximumHeight: () => number;
+  getSmartSnippetState: () => SmartSnippetState;
+  getSmartSnippet: () => SmartSnippet;
+  getSnippetStyle: () => string | undefined;
+  getFeedbackSent: () => boolean;
+  setModalRef: (ref: HTMLAtomicSmartSnippetFeedbackModalElement) => void;
+  setFeedbackSent: (isSent: boolean) => void;
+}
+
+export class SmartSnippetCommon {
```
btaillon, line 24: I know we've been extracting a lot of components into `*Common` already, but I think this pattern has made it difficult to maintain components, and I think we should revisit that.

What I find especially difficult to maintain with `*Common` components is that:
1. They don't have any semantic meaning besides "extracts of my parent classes". Their individual purpose/scope is ambiguous, we can't get the big picture without tabbing between them and their parent component.
2. We lose the distinction between props, state, events, public methods, private methods, etc.

I'm honestly wondering if it would've been easier to maintain if we had just copy and pasted every component and only extracted distinguishable parts (e.g., `atomic-smart-snippet` is already composed of `atomic-smart-snippet-expandable-answer`, `SmartSnippetFeedbackBanner`, `atomic-smart-snippet-source`, `atomic-smart-snippet-modal` and a few `Heading`).

Thoughts?

nathanlb: It's definitely a good question. One advantage of the current method I see has to do with the split ownership of the insight vs. search components. You guys are far more focused on the health of atomic components and seem to be bringing improvements more regularly than we would. Yet we benefit from those changes when they're applied to the common code. Could there be a danger then that the search smart snippets get a bug fix or a UX improvement that the insight version doesn't because of that and falls behind?

Your argument centers around difficulties in maintaining the components and you do far more of that overall than we do so the issue is likely more apparent to you than it is to me.

btaillon: Good point.

The biggest risk I see is for components where it's difficult to extract anything from. We may forget that they're part of other use cases too, and we may forget to update all of them. Components with extracted parts (like the smart snippet) should be a tad safer since their changes would affect all use cases, which would be especially apparent when refactoring or using "Find all references" (which I personally do very often).

Overall though, I think I'd still rather avoid `*Common` components so far. We could make more of an effort to divide every component into distinct/scoped parts, and make them case-agnostic.

nathanlb: Agreed
Honestly there are usually only 2 things that differ between insight and search components:
1. The headless imports: They come from different places but the api of the controllers and the shape of the interfaces are usually the same.
2. The bindings, these don't differ much either AFAIK.

That's why in Quantic we reuse the same component for each use case and just give it a different bundle based on the use case. It expects a certain API which we make sure is consistent between insight and search and it just works.

Maybe there's something to be done like that where the components are more generic as you say.

btaillon: Does this need to be merged this week? I'm wondering if this is something we should discuss after Innovate.

nathanlb: We can wait until tomorrow to discuss but it is blocking for a key feature we need to deliver this quarter so this can't hang in limbo for long. We're following your team's initial guidance on how to contribute insight components to Atomic so we were expecting this to be a solved problem 😅.

olamothe: My two cents on this topic: 
* I would say go ahead @nathanlb  and merge this as is.

* @btaillon : if we want to change how we approach "common" components pattern, it's not up to Nathan's team to do these changes, especially not in a PR like this. It would be up to our team to come up with the new recipe that we think will work better, and refactor existing code to match that new approach to validate that it's a better way of doing things. 

* And we definitely don't want to be blocking other teams from advancing their projects while any of this would be happening. I know I would * hate * to be on the receiving end of this 😁

olamothe (approved):

packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet/smart-snippet-common.tsx
nathanlb: 🙏 Thanks! I'm very willing to participate in upcoming discussions to find the optimal formula if you think that'll help. Ideally in a non-blocking way as you said.

btaillon (approved):

PR #2883, opened 2023-05-10
fix(headless): serialize and deserialize endInclusive parameters for range facets (numeric, dates)

### LINKED ISSUE #2877: [Bug]: NumericFacet endInclusive defaults to false when refreshing a page

Headless

### product version

2.13.1

### bug description

Hello!

I wanted to report an edge case we've found when testing Numeric Facets.

If you select the last generated range - so the one with `endInclusive` set to `true` in URL the information about inclusion is not persisted.
When you refresh the page facet selected gets set with `endInclusive: false` in some cases resulting with loosing last matches.

### Steps to reproduce

1. Open `headless samples` project
2. Select last item of a Numeric Facet, for example `42.7 MB to 48.8 MB inclusively (2 results)`
3. Check request parameters - in facets you can see the selected one is being sent with `endInclusive: true`
4. Refresh the page, you can see that options for the facet changed - now the second from the end is saying it's `inclusive` and if you check the call properties you'll see that `42.7 MB to 48.8 MB inclusively (2 results)` was called now with `endInclusive: false`

### Relevant log output

_No response_

olamothe:
Fixes #2877 

~Add optional serialization and deserialization of search parameter `endInclusive` for range facets.~

~It's optional in the sense that, by default `endInclusive` is false (at the API and index level).~
~So it's un-necessary to persist `endInclusive=false` in the URL.~

Following Benjamin suggestion, changed the format to use `..` as the delimiter to signify `endInclusive=false` and `...` for `endInclusive=true`

Previous URL format was : `(nf/df)-somefacetfield=start..end`
With this change format becomes ~`(nf/df)-somefacetfield=start..end(optional)..endInclusive`~ `(nf/df)-somefacetfield=start...end`

https://coveord.atlassian.net/browse/KIT-2460


olamothe:
> I'm also seeing other languages like Perl and Pascal which use two dots or three dots to signify an inclusive range.
> Given that, I think we could use triple-dots to mean an inclusive range.

Very good idea ! I'll apply those changes

packages/headless/src/features/search-parameters/search-parameter-serializer.ts
```diff
@@ -172,9 +178,22 @@ function processObjectValues(key: string, values: string[]) {
...
 function buildNumericRanges(ranges: string[]) {
   return ranges
-    .map((str) => str.split(rangeDelimiter).map(parseFloat))
-    .filter((range) => range.length === 2 && range.every(Number.isFinite))
-    .map(([start, end]) => buildNumericRange({start, end, state: 'selected'}));
+    .map((str) => {
+      const isEndInclusive = str.indexOf(rangeDelimiterInclusive) !== -1;
+      const [startAsString, endAsString] = str.split(
+        isEndInclusive ? rangeDelimiterInclusive : rangeDelimiterExclusive
+      );
```
louis-bompart, line 185: This bit of code is repeated between date and numeric, do you think we should refactor it?

louis-bompart (approved):
LGTM, excellent comment @btaillon, that's a really neat idea.

PR #2885, opened 2023-05-11
fix(atomic): use admin endpoints to redirect with relevance inspector

olamothe:
https://coveord.atlassian.net/browse/KIT-2463

### DIFF
--- a/packages/atomic/src/components/search/atomic-relevance-inspector/atomic-relevance-inspector.tsx
+++ b/packages/atomic/src/components/search/atomic-relevance-inspector/atomic-relevance-inspector.tsx
@@ -1,3 +1,4 @@
+import {PlatformEnvironment, getOrganizationEndpoints} from '@coveo/headless';
 import {Component, h, Prop, Event, EventEmitter} from '@stencil/core';
 import {Button} from '../../common/button';
 import {Bindings} from '../atomic-search-interface/atomic-search-interface';
@@ -22,8 +23,15 @@ export class AtomicRelevanceInspector {
   closeRelevanceInspector: EventEmitter<null> | undefined;
 
   public render() {
-    const {platformUrl, organizationId} =
+    const {organizationId, platformUrl} =
       this.bindings.engine.state.configuration;
+    const envMatch = platformUrl.match(/(hipaa|dev|stg)/);
+    const env = envMatch ? envMatch[1] : 'prod';
+
+    const {admin} = getOrganizationEndpoints(
+      organizationId,
+      env as PlatformEnvironment
+    );
     const {searchResponseId} = this.bindings.engine.state.search;
     return (
       <atomic-modal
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/search/atomic-relevance-inspector/atomic-relevance-inspector.tsx
```diff
@@ -22,8 +23,15 @@ export class AtomicRelevanceInspector {
   closeRelevanceInspector: EventEmitter<null> | undefined;
 
   public render() {
-    const {platformUrl, organizationId} =
+    const {organizationId, platformUrl} =
       this.bindings.engine.state.configuration;
+    const envMatch = platformUrl.match(/(hipaa|dev|stg)/);
```
btaillon, line 28: If somebody uses organization URLs and their organization happens to contain `hipaa`, `dev` or `stg` in the name, I think this would get matched?

I think this needs a more strict regular expression. Alternatively, Headless could provide a way of getting the current admin endpoint OOTB.

olamothe: > Alternatively, Headless could provide a way of getting the current admin endpoint OOTB.

It does, that's what the `getOrganizationEndpoints` utility function does. We still need to pass it the environment though

btaillon: I meant that it could be stored in the reducer just like the `platformUrl` and `analytics.apiBaseUrl` (even though we may not use it in Headless *yet*).

packages/atomic/src/components/search/atomic-relevance-inspector/atomic-relevance-inspector.tsx
```diff
@@ -57,4 +55,27 @@ export class AtomicRelevanceInspector {
...
+  private extractEnvironmentFromPlatformURL() {
+    const {platformUrl} = this.bindings.engine.state.configuration;
+
+    const envMatch = platformUrl.match(
+      /^https:\/\/(platform(?<envFromPlatform>dev|stg|hipaa)|[a-z0-9]+\.org(?<envFromOrg>dev|stg|hipaa))/
+    );
+    if (!envMatch || !envMatch.groups) {
+      return 'prod';
+    }
+    return envMatch.groups.envFromPlatform || envMatch.groups.envFromOrg;
```
btaillon, line 68: This is a bit difficult to read IMO.

What about something like:
```ts
const platformUrl = new URL(this.bindings.engine.state.configuration.platformUrl);
const platformUrlMatch = platformUrl.host.match(
  /^platform(?<env>dev|stg|hipaa)?\.coveo\.com/
);
if (platformUrlMatch) {
  return platformUrlMatch.groups?.env ?? 'prod';
}
const organizationUrlMatch = platformUrl.host.match(
  /^[^\.]+\.org(?<env>dev|stg|hipaa).coveo.com/
);
...
```

btaillon (approved):

packages/atomic/src/components/search/atomic-relevance-inspector/atomic-relevance-inspector.tsx
olamothe: Ok, in this particular case I don't think it's a good idea to store that value in Headless, it's pretty corner case that we ourselves use it; most clients won't have any need for that value when building any UI with Headless.

PR #2890, opened 2023-05-12
ci: fixed CI repo strings and extracted common constants

btaillon:
https://coveord.atlassian.net/browse/KIT-2470

I went too fast and didn't pay enough attention, I repeatedly missed some constants that referred to the CLI repo.

I decided to extract some common constants in a new file & directory (`./common/constants.mjs`) to avoid this happening again.

I was also tempted to extract secrets/env variables, but those are searchable anyway.

### DIFF
--- a/utils/release/tsconfig.json
+++ b/utils/release/tsconfig.json
@@ -7,6 +7,7 @@
     "checkJs": true,
     "skipLibCheck": true,
     "noEmit": true,
+    "noUnusedLocals": true,
     "lib": ["ESNext"]
   },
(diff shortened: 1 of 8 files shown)


utils/release/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "checkJs": true,
     "skipLibCheck": true,
     "noEmit": true,
+    "noUnusedLocals": true,
```
btaillon, line 10: just a bonus

louis-bompart (approved):

PR #2892, opened 2023-05-12
ci: replaced automatic auth with app auth

btaillon:
https://coveord.atlassian.net/browse/KIT-2472


louis-bompart (approved):
LGTM, ensure the scripts has access to the secret in the workflow yml

PR #2895, opened 2023-05-12
feat(atomic): add a part on `facet-value-checkbox > atomic-icon`

aprudhomme-coveo:
[KIT-2412](https://coveord.atlassian.net/browse/KIT-2412)

### Summary
This PR adds a part for  the atomic-icon within facet checkboxes.

### Possible problem
Currently, the tick icon in the SVG has a default stroke color of white and I haven't found a way to override it. If a user wanted to change the tick icon to another color, they would not be able to.

### Two solutions 

- Add custom icons to the checkbox
- Remove directly the default stroke in the tick SVG

[KIT-2412]: https://coveord.atlassian.net/browse/KIT-2412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/common/checkbox.tsx
+++ b/packages/atomic/src/components/common/checkbox.tsx
@@ -9,6 +9,7 @@ export interface CheckboxProps {
   class?: string;
   text?: string;
   part?: string;
+  iconPart?: string;
   ariaLabel?: string;
   ariaCurrent?: string;
   ref?(element?: HTMLElement): void;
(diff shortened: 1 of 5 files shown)


btaillon (approved):

louis-bompart:
LGTM! Straight to the point ;) 

Regarding the problem you raised up, it'll be nice to fix it up in a following PR/Jira.
Indeed, if I check the discuss that started all this, it asks for "a way to override the default colour of the check mark icon of the facet checkbox?"

I think that changing the whole SVG is a big ask for the user, we should probably modify the SVG instead to make it adaptable with our CSS part

louis-bompart (approved):

PR #2897, opened 2023-05-15
feat(headless, folding): Adapted the headless folded result list controller to support insight use

SimonMilord:
[SFINT-4851](https://coveord.atlassian.net/browse/SFINT-4851)

**As a User**
I want to use the folding feature when I’m using Quantic for the insight use case. I need to be able to build a Headless folded result list controller using an insight engine.

**Acceptance criteria**
Headless folded result list controller should work with the insight use case.

**Technical details**
We need to create a core headless folded result list controller headless that accepts both a search engine and an insight engine. This core controller would then be used to build a folded result list controller for both the search use case and the insight case.

**STEPS** (maybe will make it easier to understand what is going on in the PR): 

1. Move analytics actions in their own file
2. Added analytics actions for insight folding
3. Updated exports files (index.ts and insight.index.ts) for folding
4. Added Core controller
5. Added InsightFoldedResultList controller
6. Refactored existing controller to use SearchEngine. Cleaned up file
7. Added Insight-folding-actions and its loader
8. wrote unit tests for the core controller
9. wrote unit tests for insight analytics

**DEMO (search & insight usecases):**

https://github.com/coveo/ui-kit/assets/73316533/f7c00372-8143-4802-8cc3-93b0cd00c999

[SFINT-4851]: https://coveord.atlassian.net/browse/SFINT-4851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.test.ts
```diff
@@ -0,0 +1,145 @@
+import {search, configuration, folding, query} from '../../../app/reducers';
```
olamothe, line 1: You'll see the `reducers` file has been deleted in master as of today.

This is part of an initiative to reduce bundle size for our clients. You will need to imports reducers from their respective files instead 

ie: 
```typescript
import {searchReducer as search} from '../../../features/search/search-slice'`
// etc.
```

package.json
```diff
@@ -22,6 +22,7 @@
     "notify:docs": "node ./scripts/notify-docs/published-ui-kit.mjs"
   },
   "dependencies": {
+    "coveo.analytics": "^2.27.12",
```
mmitiche, line 25: This should not be here.
`npm i coveo.analytics@xyz` should be executed from ui-kit/packages/headless and the dependency  should be updated in `ui-kit/packages/quantic/package.json`

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,275 @@
...
+  moreResultsAvailable: boolean;
+}
+
+export interface FoldedResultAnalyticsClient {
+  logShowMoreFoldedResults: (result: Result) => ClickAction;
+  logShowLessFoldedResults: () => CustomAction;
+}
+
+/**
+ * Creates a `FoldedResultList` controller instance.
```
mmitiche, line 158:
```suggestion
 * Creates a core `FoldedResultList` controller instance.
```

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
```diff
@@ -148,111 +42,18 @@ export interface FoldedResultListState extends SearchStatusState {
  */
 export function buildFoldedResultList(
   engine: SearchEngine,
-  props: FoldedResultListProps = {}
+  props: FoldedResultListProps
```
mmitiche, line 45: why do we need to remove this default value?

packages/headless/src/features/insight-folding/insight-folding-actions-loader.ts
mmitiche, line 1: I don't think the folder`features/insight-folding` should be created.
You can directly add the file `folding-insight-actions-loader.ts` inside the folder `features/folding`.

packages/headless/src/features/insight-search/insight-search-request.ts
```diff
@@ -55,6 +56,10 @@ export const buildInsightSearchRequest = (
       sortCriteria: state.sortCriteria,
     }),
     tab: state.configuration.analytics.originLevel2,
+    filterField: state.folding?.fields.collection,
+    childField: state.folding?.fields.parent,
+    parentField: state.folding?.fields.child,
+    filterFieldRange: 100,
```
mmitiche, line 62: This should not always be 100. The value of this parameter should be 100 only in the `loadCollection` action.
```suggestion
    filterFieldRange: state.folding.filterFieldRange,
```

mmitiche (commented):
taking shape 👌
A bit more work needed for the load collection action and the unit test but nice job!

package.json
SimonMilord: Yeah I forgot to cleanup this, will remove

packages/headless/src/controllers/folded-result-list/headless-folded-result-list.ts
mmitiche: I think this is a breaking change, we shall not touch the signature of the `buildFoldedResultList`.

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,275 @@
...
+   * @defaultValue `2`
+   */
+  numberOfFoldedResults?: number;
+}
+
+export interface FoldedResultListOptions extends ResultListOptions {
+  folding?: FoldingOptions;
+}
+
+export interface FoldedResultListProps {
```
mmitiche, line 71: You  need to name this `CoreFoldedResultListProps` because that's not the interface that we need to pass to the search/insight folding result controller.
The interface we need to pass to the search/insight folding result controller doesn’t need to contain `loadCollectionActionCreator`, it will just be: 
```
export interface FoldedResultListProps {
  /**
   * The options for the `FoldedResultList` controller.
   * */
  options?: FoldedResultListOptions;
}
```

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,275 @@
...
+  folding?: FoldingOptions;
+}
+
+export interface CoreFoldedResultListProps {
+  /**
+   * The options for the `FoldedResultList` controller.
+   * */
+  options?: FoldedResultListOptions;
+  /**
+   *
```
SimonMilord, line 77: just noticed missing description here, will add

packages/headless/src/api/search/search-metadata.ts
```diff
@@ -6,7 +6,8 @@ export type SearchApiMethod =
   | 'recommendations'
   | 'html'
   | 'productRecommendations'
-  | 'fieldDescriptions';
+  | 'fieldDescriptions'
+  | 'query';
```
mmitiche, line 10: I don't  think the new value here is necessary, it's not a new Search API method that you are using for folding, you are doing a search query so I think you can just use the value `search` in you request.

packages/headless/src/api/service/insight/insight-api-client.ts
```diff
@@ -81,10 +82,12 @@ export class InsightAPIClient {
   }
 
   async query(
-    req: InsightQueryRequest
+    req: InsightQueryRequest,
+    options?: SearchOptions
   ): Promise<InsightAPIResponse<SearchResponseSuccess>> {
     const response = await PlatformClient.call({
       ...buildInsightQueryRequest(req),
+      requestMetadata: {method: 'query', origin: options?.origin},
```
mmitiche, line 90:
```suggestion
      requestMetadata: {method: 'search', origin: options?.origin},
```

packages/headless/src/controllers/result-list/headless-result-list.ts
```diff
@@ -18,7 +18,7 @@ export type {ResultListOptions, ResultListProps, ResultListState, ResultList};
  * @returns A `ResultList` controller instance.
  */
 export function buildResultList(
-  engine: SearchEngine,
+  engine: CoreEngine,
```
mmitiche, line 21: This is the function that builds the Headless Search Result controller, I don't think it should receive a CoreEngine. it's specific to search.

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,275 @@
...
+export function buildCoreFoldedResultList(
+  engine: CoreEngine,
+  props: CoreFoldedResultListProps,
+  analyticsClient: FoldedResultAnalyticsClient
+): CoreFoldedResultList {
+  if (!loadFoldingReducer(engine)) {
+    throw loadReducerError;
+  }
+
+  const controller = buildResultList(engine, props);
```
mmitiche, line 174: Instead of making `buildResultList` use a core engine which is not ideal, you can use the `buildCoreResultList`.

packages/headless/src/features/folding/insight-folding-actions.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export const foldingOptionsSchemaDefinition: SchemaDefinition<
+  Required<RegisterFoldingActionCreatorPayload>
+> = {
+  collectionField: new StringValue({emptyAllowed: false, required: false}),
+  parentField: new StringValue({emptyAllowed: false, required: false}),
+  childField: new StringValue({emptyAllowed: false, required: false}),
+  numberOfFoldedResults: new NumberValue({min: 0, required: false}),
+};
```
mmitiche, line 58: You can import these interfaces from `folding-actions.ts` instead.

packages/headless/src/features/folding/insight-folding-actions.ts
```diff
@@ -0,0 +1,106 @@
...
+export const registerFolding = createAction(
+  'folding/register',
+  (payload: RegisterFoldingActionCreatorPayload) =>
+    validatePayload(payload, foldingOptionsSchemaDefinition)
+);
+
+export type StateNeededByLoadCollection = ConfigurationSection &
+  FoldingSection &
+  QuerySection &
+  InsightConfigurationSection;
```
mmitiche, line 69: Same for this interface, we can import it and save some lines of code.

packages/headless/src/features/insight-search/insight-search-request.ts
```diff
@@ -55,6 +56,10 @@ export const buildInsightSearchRequest = (
       sortCriteria: state.sortCriteria,
     }),
     tab: state.configuration.analytics.originLevel2,
+    filterField: state.folding?.fields.collection,
+    childField: state.folding?.fields.parent,
+    parentField: state.folding?.fields.child,
+    filterFieldRange: state.folding?.filterFieldRange,
```
mmitiche, line 62:
```suggestion
     ...(state.folding && {
          filterField: state.folding.fields.collection,
         childField: state.folding.fields.parent,
         parentField: state.folding.fields.child,
         filterFieldRange: state.folding.filterFieldRange,
    }),
```

mmitiche (commented):
Nice job! few things to address but it's taking shape👌

packages/headless/src/controllers/result-list/headless-result-list.ts
SimonMilord: you are right, must have forgotten this one as the insight has a InsightSearchEngine. Good catch

packages/headless/src/features/folding/insight-folding-actions.ts
SimonMilord: touché :D

packages/headless/src/features/folding/insight-folding-actions.ts
SimonMilord: This one I put there since we add the InsightConfigurationSection

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,277 @@
...
+): CoreFoldedResultList {
+  if (!loadFoldingReducer(engine)) {
+    throw loadReducerError;
+  }
+
+  const controller = buildCoreResultList(engine, props);
+  const {dispatch} = engine;
+  const getState = () => engine.state;
+
+  const options = props?.options?.folding
```
btaillon, line 180: Not sure why this optional chain is needed, `props` is required.
```suggestion
  const options = props.options?.folding
```

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  return {
+    ...controller,
+
+    loadCollection: (collection) => {
+      dispatch(
+        props.loadCollectionActionCreator(
+          collection?.result?.raw[
+            engine.state.folding.fields.collection
+          ] as string
```
btaillon, line 199: Same as above. I don't see why this optional chain is needed, since `collection` and `result` both seem required.

```suggestion
          collection.result.raw[
            engine.state.folding.fields.collection
          ] as string
```

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,277 @@
...
+    },
+  };
+}
+
+function loadFoldingReducer(
+  engine: CoreEngine
+): engine is CoreEngine<
+  SearchSection & ConfigurationSection & FoldingSection & QuerySection
+> {
+  engine.addReducers({search, configuration, folding, query});
```
btaillon, line 256: Are all of these reducers needed for the insight use case?

packages/headless/src/features/folding/folding-actions.ts
```diff
@@ -73,7 +66,8 @@ export const registerFolding = createAction(
 
 export type StateNeededByLoadCollection = ConfigurationSection &
   FoldingSection &
-  QuerySection;
+  QuerySection &
+  InsightConfigurationSection;
```
btaillon, line 70: Am I correct that this `loadCollection` action should only be used by the search use case, and the insight use case uses a different one?

If so, you should revert this change.
```suggestion
  QuerySection;
```

packages/headless/src/features/folding/insight-folding-actions.ts
```diff
@@ -0,0 +1,68 @@
...
+export const registerFolding = createAction(
+  'folding/register',
+  (payload: RegisterFoldingActionCreatorPayload) =>
+    validatePayload(payload, foldingOptionsSchemaDefinition)
+);
+
+// export type StateNeededByLoadCollection = ConfigurationSection &
+//   FoldingSection &
+//   QuerySection &
+//   InsightConfigurationSection;
```
btaillon, line 31: Looks like this should be uncommented and used instead of the one from `../folding/folding-actions`.

packages/headless/src/features/insight-search/insight-search-request.ts
```diff
@@ -10,7 +11,7 @@ import {MappedSearchRequest, mapSearchRequest} from '../search/search-mappings';
 
 type StateNeededBySearchRequest = ConfigurationSection &
   InsightConfigurationSection &
-  Partial<InsightAppState>;
+  Partial<InsightAppState & FoldingSection>;
```
btaillon, line 14: `InsightAppState` already includes `FoldingSection`

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
SimonMilord: you are right, ill remove it

packages/headless/src/features/folding/folding-actions.ts
SimonMilord: loadCollection is used in both useCases, but the insight one has the InsightConfigurationSection on top of the what the folding-action. I thought I could declare it in one and import it, but following your other comment in insight-folding-actions.ts it would be better maybe to declare it in each file

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
SimonMilord: for the reducers I have used the same ones previous in use by the `headless-folded-result-list` controller.
It seems to work fine without the query reducer but I believe the 3 other ones to be necessary

packages/headless/src/api/service/insight/insight-api-client.ts
mmitiche: It's more about the action that will be executed on the Search API. In the insight use case it's a bit special because the requests are not sent directly sent to the search API, they are sent to Customer Service API which performs the request to the search API. so the finality of the request is a search request.
That's why imo the value should be `search`.

mmitiche (commented):
I see that the Atomic Folding  e2e test is failing, is this due to some flakiness or there is something wrong with these changes?
Folding

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
btaillon: I'm a bit suspicious of the search reducer. Does the insight use case typically use the search reducer?

SimonMilord: I dont think the insight usecase uses it. I tested it on my end, both insight and search use cases seem to work even without the search reducer.
I would have thought it would be needed since the core controller contains both the search and insight usecases and the previous "search only" iteration of the controller seemed to add it to its engine

SimonMilord: I dug a bit more into this, turns out you were right search doesnt seem to be needed. `query`, however does.

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.test.ts
```diff
@@ -108,14 +150,14 @@ describe('FoldedResultList', () => {
...
 
-    it('dispatches analytics and folding/loadCollection when calling loadCollection', () => {
-      foldedResultList.loadCollection(foldedResultList.state.results[0]);
-      expect(engine.actions.pop()?.type).toEqual(
-        'analytics/folding/showMore/pending'
-      );
-      expect(engine.actions.pop()?.type).toEqual(
-        'folding/loadCollection/pending'
-      );
+    it('#loadCollection dispatches #loadCollection AND #logShowMoreFoldedResults analytics', () => {
```
mmitiche, line 153: #loadCollection dispatches #loadCollection sounds a bit odd, I believe we can improve the description of this test.

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.test.ts
```diff
@@ -108,14 +150,14 @@ describe('FoldedResultList', () => {
...
+      const expectedLogShowMoreAction = 'analytics/folding/showMore/pending';
+      const expectedLoadCollectionAction = 'folding/loadCollection/pending';
+
+      foldedResultList.loadCollection(foldedResultList?.state?.results[0]);
+      expect(foldedResultList.state.results).not.toBe(undefined);
```
mmitiche, line 158: why do we need this expect here? we are testing  the dispatch of `loadCollection` and `logShowMoreFoldedResults` no?

mmitiche (approved):
GG 👌

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.test.ts
SimonMilord: oh I tried to add this before I moved the test, I can remove it as I thought it could fix the test failing originally.

erocheleau (commented):
Just one suggestion where I think we could simplify the code

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,274 @@
...
+   * The options for the `FoldedResultList` controller.
+   * */
+  options?: FoldedResultListOptions;
+  /**
+   * The action creator to build the `loadCollection` action.
+   */
+  loadCollectionActionCreator: (
+    collectionId: CollectionId
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ) => AsyncThunkAction<any, CollectionId, any>;
```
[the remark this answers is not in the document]
SimonMilord: I believe so yes, other examples using the actionCreator pattern are also passing any as types in the AsyncThunkAction : 

image

packages/headless/src/features/folding/folding-insight-analytics-actions.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    );
+    expect(mockToUse.mock.calls[0][2]).toStrictEqual(expectedCaseContext);
+  });
+
+  it('should log #logShowLessFoldedResults', async () => {
+    await engine.dispatch(logShowLessFoldedResults());
+
+    const mockToUse = mockLogShowLessFoldedResults;
+    expect(mockToUse).toBeCalledTimes(1);
+    expect(mockToUse.mock.calls[0][0]).toStrictEqual(expectedCaseContext);
```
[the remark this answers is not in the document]
SimonMilord: no we don't! :D

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/**
+ * The `CoreFoldedResultList` headless controller re-organizes results into hierarchical collections (a.k.a. threads).
+ */
+export interface CoreFoldedResultList extends Controller {
```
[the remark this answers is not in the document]
SimonMilord: good point, I will refactor these, only thing is the props which I will have to export from the files because of the actionCreator pattern. It does clean up the files nicely though thanks!

packages/headless/src/api/service/insight/insight-api-client.ts
```diff
@@ -81,10 +82,12 @@ export class InsightAPIClient {
   }
 
   async query(
-    req: InsightQueryRequest
+    req: InsightQueryRequest,
+    options?: SearchOptions
   ): Promise<InsightAPIResponse<SearchResponseSuccess>> {
     const response = await PlatformClient.call({
       ...buildInsightQueryRequest(req),
+      requestMetadata: {method: 'search', origin: options?.origin},
```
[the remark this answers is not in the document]
SimonMilord: we need this optional param to be passed in order to pass the `origin` param as it is used as a second parameter in the `search` method of the `loadCollection` action. Its use is to pass the SearchOrigin which can take these values:

`export type SearchOrigin =
  | 'mainSearch'
  | 'facetValues'
  | 'instantResults'
  | 'foldingCollection';`

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
erocheleau: Yea okay if I understand what you told me correctly Simon and I don't remember why we need that in the props there, but @mmitiche is correct.

Name this one `CoreFoldedResultListProps`

packages/headless/src/controllers/insight/folded-result-list/headless-insight-folded-result-list.ts
```diff
@@ -0,0 +1,52 @@
...
+export type {
+  FoldedResultListOptions,
+  FoldedResultListState,
+  FoldedResultList,
+  FoldingOptions,
+  FoldedCollection,
+  FoldedResult,
+};
+
+export interface FoldedResultListProps {
```
erocheleau, line 23: You should rename this something like `InsightFoldedResultListProps` to avoid confusion.

erocheleau, line 23: This should be something like : `InsightFoldedResultListProps`

erocheleau (approved):
Just a rename otherwise it looks good

PR #2899, opened 2023-05-15
feat(atomic): remove white stroke on checkbox tick svg when using part

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2475

This PR enable the user to change the tick svg color when using css parts.


packages/atomic/src/components/common/checkbox.tsx
```diff
@@ -17,6 +17,8 @@ export interface CheckboxProps {
 }
 
 export const Checkbox: FunctionalComponent<CheckboxProps> = (props) => {
+  const isIconPart = props.iconPart !== undefined && props.iconPart !== null;
```
louis-bompart, line 20: You may want to use `isNullOrUndefined` from `@coveo/bueno`
Source:
https://github.com/coveo/ui-kit/blob/7f9b9cb3469e1852765b00d40adca999ba4a0d4c/packages/bueno/src/values/value.ts#L46-L48
E.g in Atomic:
https://github.com/coveo/ui-kit/blob/7f9b9cb3469e1852765b00d40adca999ba4a0d4c/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx#L1
https://github.com/coveo/ui-kit/blob/7f9b9cb3469e1852765b00d40adca999ba4a0d4c/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx#L537-L538
And some mind juggling later, here's a silly oneliner implementation without equal sign:
```js
const isNullOrUndefined = (a) => !(!(a ?? true) || a)
```

louis-bompart (commented):
lgtm, maybe use bueno

packages/atomic/src/components/common/checkbox.tsx
```diff
@@ -54,6 +54,7 @@ export const Checkbox: FunctionalComponent<CheckboxProps> = (props) => {
       onMouseDown={(e) => props.onMouseDown?.(e)}
     >
       <atomic-icon
+        style={{stroke: 'white'}}
```
aprudhomme-coveo, line 57: Using a inline style gives (0, 1, 0) priority and using a part selector gives (0, 1, 1) priority.

aprudhomme-coveo: This is quite bizarre because it works in atomic testing and in the storybook. When I specify a new stroke in the part, it overrides it visually but in the browser dev tools it is crossed out. I am looking into it to understand better.

```
atomic-rating-facet::part(value-checkbox-icon) {
        stroke: aqua;
        background-color: red;
      }
```
      

Screenshot 2023-05-16 at 1 29 47 PM

Screenshot 2023-05-16 at 1 29 59 PM

louis-bompart: As you said the browser is doing pretty weird 💩 here.
However, it 'works' so far, so... :shipit:

louis-bompart (approved):

PR #2902, opened 2023-05-16
ci: re-build Headless and Atomic after bumping versions

btaillon:
https://coveord.atlassian.net/browse/KIT-2477

Headless, Atomic and AtomicAngular all require to be built right after their version was bumped but before they are published.

Here's what I did in this PR:
1. Added a new nx task/target called `release:build`
    * Using a custom NX plugin, this task is copied from the build task, but depends on the version being bumped.
2. I modified our `setup` composite action to optionally skip building and optionally use a different registry.


.github/workflows/masterbot.yml
```diff
@@ -164,26 +164,36 @@ jobs:
...
     #   - 'e2e-quantic-test'
     runs-on: ubuntu-latest
-    environment: prerelease
+    environment: 'Prerelease'
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: ./.github/actions/setup
+      - uses: actions/setup-node@v3
```
btaillon, line 181: At some point I also tested modifying the `setup` composite action to have an optional `build` boolean parameter and an optional `registry-url` parameter, but I reverted it because it's probably better to have less options.

btaillon: Essentially, we no longer want to build, because we want the new release process to build from scratch.

.github/workflows/masterbot.yml
```diff
@@ -164,26 +164,36 @@ jobs:
...
+        with:
+          # registry-url: 'https://registry.npmjs.org' # TODO: Uncomment when we enable new versioning process.
```
btaillon, line 183: See documentation for `registry-url`, which also explains why I'm adding `NODE_AUTH_TOKEN` to the whole job instead of one action:
```md
    # Optional registry to set up for auth. Will set the registry in a project level .npmrc and .yarnrc file, 
    # and set up auth to read in from env.NODE_AUTH_TOKEN.
    # Default: ''
    registry-url: ''
```

nx.json
```diff
@@ -7,6 +7,7 @@
       }
     }
   },
+  "plugins": ["@coveo/nx-plugin"],
```
btaillon, line 10: The plugin is executed whenever we do `npm install` to generate Nx's topology.

utils/nx-plugin/index.js
```diff
@@ -0,0 +1,13 @@
+const {
+  addReleaseBuildTargets,
+} = require('./clone-build-targets-for-release.cjs');
+
+module.exports = /** @type {import('@nrwl/devkit').NxPlugin} */ ({
+  name: '@coveo/nx-plugin',
+  processProjectGraph(currentGraph) {
```
btaillon, line 7: This allows us to modify Nx's project graph completely. In our case, we use that to add new targets (tasks).

Alternatively, there's another function that allows us to add targets directly, but it doesn't allow us to read other targets, which we need to do.

utils/nx-plugin/clone-build-targets-for-release.cjs
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * @param {ProjectGraphProjectNode} project
+ */
+function addReleaseBuildTargets(project) {
+  const projectTargets = project.data.targets || {};
+  if (!('build' in projectTargets)) {
+    return;
+  }
+  const {target: buildTarget, localDependencies: buildTargetDependencies} =
+    cloneTarget(projectTargets, 'build');
```
btaillon, line 119: We want to clone both the `build` target itself, and all the **local** (non-parent) targets it depends on.

To understand this, take a look at `atomic`'s targets:
```json
{
  "targets": {
    "cached:build:stencil": {
      "dependsOn": ["^build", "build:locales"],
      "inputs": ["^production", "production"],
      "outputs": [
        "..."
      ],
      "executor": "...",
      "options": {
        "..."
      }
    },
    "validate:definitions": {
      "dependsOn": ["^build", "cached:build:stencil"],
      "inputs": [],
      "outputs": [],
      "executor": "...",
      "options": {
        "..."
      }
    },
    "build": {
      "dependsOn": ["cached:build:stencil", "validate:definitions"],
      "inputs": [],
      "outputs": [],
      "executor": "nx:noop"
    }
  }
}
```

As you can see, if we were to just clone the `build` target, it would queue  `cached:build:stencil` and `validate:definitions`. Subsequently, these targets would queue up `^build` and `build:locales`.

Queueing up `^build` is problematic. It will cause parent projects to be rebuilt by executing their `build` target, which won't wait for the version to be bumped. This could cause Headless to be built before its version is bumped, which would cause Atomic to use an old build of Headless, which contains the wrong Headless `engine.state.version`.

utils/nx-plugin/clone-build-targets-for-release.cjs
```diff
@@ -0,0 +1,130 @@
...
+  const newTargets = {...buildTargetDependencies, 'release:build': buildTarget};
+  for (const dependencyName of Object.keys(buildTargetDependencies)) {
+    renameTarget(newTargets, dependencyName, `release:build:${dependencyName}`);
+  }
```
btaillon, line 123: We want to rename the new targets to avoid overwriting the old ones. In this case, I'm just prefixing them with `release:build:`.

This function will also rename them in the `dependsOn` field of other targets.

utils/nx-plugin/clone-build-targets-for-release.cjs
```diff
@@ -0,0 +1,130 @@
...
+  for (const target of Object.values(newTargets)) {
+    replaceDependency(target, '^build', ['^release:build', 'release:phase1']);
+  }
+  project.data.targets = {...newTargets, ...projectTargets};
```
btaillon, line 127: I'm merging the new targets first because I want to allow us to define custom `release:build`/`release:build:*` targets if we eventually need to.

utils/nx-plugin/index.js
```diff
@@ -0,0 +1,13 @@
+const {
+  addReleaseBuildTargets,
+} = require('./clone-build-targets-for-release.cjs');
+
+module.exports = /** @type {import('@nrwl/devkit').NxPlugin} */ ({
+  name: '@coveo/nx-plugin',
```
mrrajamanickam-coveo, line 6: can we have a more specific name e.g. @coveo/ui-kit-release-bump

olamothe (approved):
It's one of the fear I had when this project about mono repo was starting: It's arguably more complicated than what we originally started with.

Hopefully we'll be able to maintain this until the next time there's a new tool that forces us to scrap it all 😂

utils/nx-plugin/index.js
btaillon: I'm a bit hesitant about this. If we need to change other things with Nx, I'm tempted to re-use the same plugin to avoid polluting the repo with too many projects and having to copy and paste configs.

On the other hand, this currently only has one function and we ideally want to avoid needing nx plugins. If I rename this project, I would also move all the code from `clone-build-targets-for-release.cjs` into this file.

btaillon:
> It's one of the fear I had when this project about mono repo was starting: It's arguably more complicated than what we originally started with.
> 
> Hopefully we'll be able to maintain this until the next time there's a new tool that forces us to scrap it all 😂

Yeah, this is the most nx-specific thing in the repo. Thankfully, its only purpose is to control the order in which things happen. If we eventually decide to scrap Nx, that's something that needs to be reimagined anyway.

PR #2903, opened 2023-05-16
fix(quantic):  properly encoding the uri in the quantic standalone search box

mmitiche:
[SFINT-5019](https://coveord.atlassian.net/browse/SFINT-5019)

## The issue:
An error occurs when the end user performs a query starting with the character % in the quantic standalone search box in a Salesforce community.

https://github.com/coveo/ui-kit/assets/86681870/6ebcaf9a-0599-4e19-852f-9aa6df01edcb

## The solution:
Headless has some logic that will try to deserialize and decode the URI in order to extract the search parameters so just needed to properly encode the URI to prevent this issue from happening again.

https://github.com/coveo/ui-kit/assets/86681870/604fa8ee-6c44-4c99-af8c-888485bb3790

also replaced `encodeURI` by `encodeURIComponent` in the e2e tests, that's a nice article explaining the difference: https://thisthat.dev/encode-uri-vs-encode-uri-component/

[SFINT-5019]: https://coveord.atlassian.net/browse/SFINT-5019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau (approved):
That seems to make sense.
However I think if we have standalone searchbox cypress tests, I would add one to make sure to cover this use-case, like triggering a query starting with `%` for example.

mmitiche:
> That seems to make sense.
> However I think if we have standalone searchbox cypress tests, I would add one to make sure to cover this use-case, like triggering a query starting with `%` for example.

Sure 💯 let me add that real quick

PR #2910, opened 2023-05-17
fix: hide ipx if parent is hidden SVCINT-2357

pathiery:
[SVCINT-2357](https://coveord.atlassian.net/browse/SVCINT-2357)

Embedded IPX was always displayed, even when parent was hidden (since we always set the visible/invisible class)

We now only set the class if `isOpen` is set on the ipx-body, which lets us inherit visibility from the parent in the embedded case.

[SVCINT-2357]: https://coveord.atlassian.net/browse/SVCINT-2357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/ipx/atomic-ipx-body/atomic-ipx-body.tsx
+++ b/packages/atomic/src/components/ipx/atomic-ipx-body/atomic-ipx-body.tsx
@@ -45,12 +45,11 @@ export class AtomicIPXBody implements InitializableComponent<AnyBindings> {
   public render() {
     this.updateBreakpoints();
 
+    const isEmbedded = this.isOpen === undefined;
     return (
       <article
         part="container"
-        class={`${
-          this.isOpen === undefined ? '' : this.isOpen ? 'visible' : 'invisible'
-        }`}
+        class={`${isEmbedded ? '' : this.isOpen ? 'visible' : 'invisible'}`}
         onAnimationEnd={() => this.animationEnded.emit()}
       >
         <style>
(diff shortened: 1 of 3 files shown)


packages/atomic/src/components/ipx/atomic-ipx-body/atomic-ipx-body.tsx
```diff
@@ -48,7 +48,9 @@ export class AtomicIPXBody implements InitializableComponent<AnyBindings> {
     return (
       <article
         part="container"
-        class={`${this.isOpen ? 'visible' : 'invisible'}`}
+        class={`${
+          this.isOpen === undefined ? '' : this.isOpen ? 'visible' : 'invisible'
```
dguerinCoveo, line 52: double ternary 🤯

dguerinCoveo (approved):
Do you think it would be worth it for readability and understandability to add a isEmbedded prop on atomic-ipx-body instead of deriving it from `isOpen === undefined`. 

Or maybe just naming that `cont isEmbedded = isOpen === undefined ` would be enough 🤔

olamothe:
As part of this PR, what do you think about changing CODEOWNERS file and adding this line:

```
/packages/atomic/src/components/ipx @coveo/service-integration 
```

olamothe (approved):

PR #2914, opened 2023-05-18
SVCINT-2169 fix: embedded ipx demo

pathiery:
https://coveord.atlassian.net/browse/SVCINT-2169

Not sure why I set this, but I caused the demo org not to load any recommendations


pathiery:
@y-lakhdar looks like some cypress tests are broken 🤔

btaillon (approved):

PR #2915, opened 2023-05-18
feat(atomic): support fallback for `atomic-result-image` 

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2280

This PR adds a new prop to atomic-result-image that serves as a fallback in these three cases :
- The image url is invalid
- The field value is missing
- The field value is not a string

~~Should I add tests for this ?~~ Added tests for all 3 cases


packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,50 @@ export class AtomicResultImage implements InitializableComponent {
...
 
+  private handleImageError() {
+    let message =
+      'The image url is not valid. You might want to add a "fallback" property.';
+
+    if (this.fallback) {
+      message = 'The image url is not valid or could not be loaded.';
+      this.host.querySelector('img')!.src = this.fallback;
+    }
+    this.bindings.engine.logger.error(message, this.host);
```
mrrajamanickam-coveo, line 52: Could this be a warning if fallback is present and error if not?

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,50 @@ export class AtomicResultImage implements InitializableComponent {
     return Array.isArray(value) ? value[0] : value;
   }
 
+  private handleImageError() {
+    let message =
+      'The image url is not valid. You might want to add a "fallback" property.';
+
+    if (this.fallback) {
+      message = 'The image url is not valid or could not be loaded.';
```
mrrajamanickam-coveo, line 49: Can we have the same first line in the msg in both cases and add `You might want to add a "fallback" property.'` conditionally?

mrrajamanickam-coveo: Can the invalid url be logged with the msg as well?

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,50 @@ export class AtomicResultImage implements InitializableComponent {
...
+
   public render() {
-    const url = this.url;
+    let url = this.url;
 
     if (!url) {
-      this.host.remove();
-      return;
+      if (this.fallback) {
+        this.bindings.engine.logger.error(
```
mrrajamanickam-coveo, line 60: Could this be a warning if the fallback is present?

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,50 @@ export class AtomicResultImage implements InitializableComponent {
...
+  }
+
   public render() {
-    const url = this.url;
+    let url = this.url;
 
     if (!url) {
-      this.host.remove();
-      return;
+      if (this.fallback) {
```
mrrajamanickam-coveo, line 59: Should we validate that fallback url is reachable and doesn't result in the same error?

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,50 @@ export class AtomicResultImage implements InitializableComponent {
...
+        this.bindings.engine.logger.error(
+          `"${this.field}" is missing. The fallback is being used. Please review the indexing.`,
```
mrrajamanickam-coveo, line 61: log the fallback value in this case?

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,50 @@ export class AtomicResultImage implements InitializableComponent {
...
+        );
+        this.host.remove();
+        return;
+      }
     }
 
-    if (typeof url !== 'string') {
+    if (url && typeof url !== 'string') {
       this.bindings.engine.logger.error(
         `Expected "${this.field}" to be a text field.`,
```
mrrajamanickam-coveo, line 77: Don't we want to use the fallback in this case?

mrrajamanickam-coveo (commented):
And yes please add tests

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
aprudhomme-coveo: Yes great idea ! I will change it.

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
aprudhomme-coveo: We could but it would require a lot of nesting. I would need to validate in every edge case where I need the fallback if the fallback is valid. I think users will most likely use a local image file path so it won't be a problem for them.

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
aprudhomme-coveo: yes good idea, I did not know about the warning function 😁

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
aprudhomme-coveo: yes good idea, I did not know about the warning function 😁

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
aprudhomme-coveo: Yes you are right, I changed my code to cover this case as well.

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,56 @@ export class AtomicResultImage implements InitializableComponent {
...
+      }
+    }
+
+    if (url && typeof url !== 'string') {
+      const message = `Expected "${this.field}" to be a text field. Please review the indexing.`;
+      url = this.handleInvalidUrl(message, this.fallback);
+      if (!url) {
+        return;
+      }
+    }
```
btaillon, line 88: I don't think we need this many errors. I think a warning to tell the user "hey, there's a fallback feature you can use" is handy, but beyond that, I think errors/warnings would be redundant and would pollute the logs.

I also prefer not to concatenate messages over different scopes, I think it makes it more difficiult to interpret the possible errors.

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +41,57 @@ export class AtomicResultImage implements InitializableComponent {
...
       return;
     }
 
-    if (typeof url !== 'string') {
-      this.bindings.engine.logger.error(
-        `Expected "${this.field}" to be a text field.`,
-        this.host
-      );
+    if (this.fallback) {
+      image.src = this.fallback;
```
btaillon, line 56: Hadn't thought of it until now, but we should avoid modifying attributes directly, since they'll be reset whenever there's a re-render of the component.

Instead of dynamically modifying `image.src`, I would recommend adding a `@State()` property, perhaps something like `@State() useFallback = false`. Then, in the `render` function, you'd just have to check whether `useFallback` is true to determine which URL to use.

aprudhomme-coveo: I did not know changing a state would recall the render function. Thanks

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -36,27 +42,54 @@ export class AtomicResultImage implements InitializableComponent {
...
 
-  public render() {
-    const url = this.url;
+  private logWarning(message: string) {
+    this.bindings.engine.logger.warn(message, this.host);
+  }
 
-    if (!url) {
+  private handleImageError(target: EventTarget | null) {
+    const image = target as HTMLImageElement;
```
y-lakhdar, line 50: Could `image.src` be undefined? Would it be safer to get the image url from `this.field` instead?

y-lakhdar (approved):

lvu285 (approved):

PR #2930, opened 2023-05-23
ci: add retry at test suite level for cypress tests in CI

mrrajamanickam-coveo:
- Currently retry happens at the CI step level which executes multiple test suites. 
  - As a result multiple test suites, even passing ones, are rerun when a test in the group fails. 
  - This increases retry time and at times introduces more flakiness when another test suite which passed previously fails now.
- Scoping the retry to the test suite containing the failed test should reduce the retry time and not introduce more flakiness into retry. 

https://coveord.atlassian.net/browse/KIT-2495


olamothe (approved):
That's nice !

PR #2932, opened 2023-05-24
test: fix flaky search box instant results tests and smart snippet feedback modal tests

mrrajamanickam-coveo:
Fix flakiness in couple of test suites by using force click for modal (as Cypress thinks its not visible) and adding a wait between 2 search boxes. 

https://coveord.atlassian.net/browse/KIT-2441


packages/atomic/cypress/e2e/search-box/standalone-search-box.cypress.ts
```diff
@@ -30,6 +30,7 @@ describe('Standalone Search Box Test Suites', () => {
   }
 
   function setupStandardSearchBox() {
+    cy.wait(1000); // Flakiness workaround to wait after clicking in standalone search box
```
btaillon, line 33: I'm not sure why this `wait` is necessary, since it runs before the new fixture is built. What failed here?

mrrajamanickam-coveo: `setupStandardSearchBox()` is called after standalone search box is setup with redirect and clicked. So need to wait for the page to be redirected otherwise it fails when interacting with the standard search box.

btaillon: I see. Then I think that this function should either be renamed to something else (like redirectToStandardSearchBoxAfterClick) or the `cy.wait` should be moved out of this function.

packages/atomic/cypress/e2e/smart-snippet-feedback-modal.cypress.ts
```diff
@@ -11,12 +11,12 @@ import {SmartSnippetSelectors} from './smart-snippet-selectors';
 describe('Smart Snippet Feedback Modal Test Suites', () => {
   function setupOpenModal() {
     new TestFixture().with(addSmartSnippet()).init();
-    SmartSnippetSelectors.feedbackDislikeButton().click();
-    SmartSnippetSelectors.feedbackExplainWhy().click();
+    SmartSnippetSelectors.feedbackDislikeButton().click({force: true});
+    SmartSnippetSelectors.feedbackExplainWhy().click({force: true});
```
btaillon, line 15: I wonder why Cypress thought these weren't visible. I'm worried that adding `force: true` would occasionally click on a non-working button.

btaillon (approved):

packages/atomic/cypress/e2e/search-box/standalone-search-box.cypress.ts
mrrajamanickam-coveo: Makes sense, renamed function and added TODO

packages/atomic/cypress/e2e/smart-snippet-feedback-modal.cypress.ts
mrrajamanickam-coveo: Yes, I am not very happy with the solution either. Think it might have something to do with the modal and its apperance with an animation. 
I did try `waitForAnimations` but it didn't fix the issue. 
To your concern about it clicking on an non-working button, it would fail further actions and subsequently the test itself. 
I have added a TODO to explore this further as needed.

mrrajamanickam-coveo:
1 test (unrelated to this PR) from facet.cypress.ts  is failing - I will try to stabilize it as part of https://github.com/coveo/ui-kit/pull/2934

PR #2938, opened 2023-05-26
ci: removed Lerna

btaillon:
https://coveord.atlassian.net/browse/KIT-2316

# 2023-05-23 - Cleanup Lerna

## Removed `beta` tag update
- 🗑 Removed the `promote:npm:beta` script
  - We now publish non-pre-release versions directly with the `beta` tag.
- 🗑 Removed `JenkinsfileStagingRelease`
  - Its only purpose was to run `promote:npm:beta`.

## Removed old release & pre-release processes
- 🗑 Removed `atomic-angular`'s `build:npm-package` script
  - Was used to run `atomic-angular/modify-package-json.js`, which was meant to append release scripts to the re-generated `package.json`.
  - 🗑 Removed `atomic-angular/modify-package-json.js`
- 🗑 Removed `scripts/deploy/bump-version.mjs`
  - Was used as an entry point to execute either `scripts/release.mjs` or `scripts/prerelease.mjs`.
  - It contained code to avoid bumping versions when new commits were added to `master`.
- 🗑 Removed `scripts/publish.mjs`
  - Was used to publish the current version to NPM with the desired tag. The new "release" utils package handles this already.
- 🗑 Removed the `bump:version` script
  - Was used to call `scripts/deploy/bump-version.mjs`.
- 🗑 Removed the `manual-package.yml` and `workflow-old.yml` manually triggered GitHub workflows
  - Were used to trigger the `bump:version` script.

### Removed the old release bump process
- 🗑 Removed `scripts/release.mjs`
  - Was responsible for calling Lerna.

### Removed the old pre-release bump process
- 🗑 Removed `scripts/prerelease.mjs`
  - Was responsible for bumping pre-release packages.
- 🗑 Removed `scripts/deploy/manual-major-prerelease.mjs`
  - Was used to bump a package to the next major pre-release version and allow it to be automatically bumped by `scripts/prerelease.mjs`.
    - This old process would probably not work anymore.
  - The `masterbot` workflow is already executed on `prerelease/**` branches. The new pre-release process should be able to publish major new pre-releases easily. The only flaw is that they will use the `alpha` tag, and the deployment pipeline won't run.
- 🗑 Removed the `bump:version:major-prerelease` script
  - Was used to call `scripts/deploy/manual-major-prerelease.mjs`.
- 🗑 Removed old pre-release docs from `CONTRIBUTING.md`

### Removed the old NPM release process
- 🗑 Removed the `publish:npm:release` script
  - Was used to call `scripts/publish.mjs`.

### Removed the old NPM pre-release process
- 🗑 Removed the `publish:npm:pre-release` script
  - Was used to call `scripts/publish.mjs`.


.deployment.config.json
louis-bompart, line 1: While at it... We have the 🟢 to remove everything related to snyk instead of just nooping.
Care to include it or should I open a PR for that? (lazyness 101, IK)

louis-bompart (approved):
N I C E

package.json
```diff
@@ -63,7 +57,6 @@
     "gts": "3.1.0",
     "husky": "5.2.0",
     "jest": "27.5.1",
-    "lerna": "5.6.2",
```
aprudhomme-coveo, line 66: Stupid question but did we completely replace lerna by nx ? Was lerna only used for publishing up to this point ?

CONTRIBUTING.md
```diff
@@ -87,9 +87,9 @@ Does the same thing as the `test` script, except:
 - Promotes a remote package.
 - Should be run after QA validations.
 
-### `promote:npm:beta` and `promote:npm:latest` scripts
+### `promote:npm:latest` script
 
-- Promotes an NPM package.
+- Promotes an NPM package to `latest`.
```
aprudhomme-coveo, line 92: Another random question but can anyone trigger a release ? Does it need special authentification ? I looked but am not able to find the code for it.

aprudhomme-coveo (approved):

package.json
btaillon: Yeah, Lerna was replaced by NPM Workspaces, Nx and our new process.

CONTRIBUTING.md
btaillon: > can anyone trigger a release ? Does it need special authentification ?

Anybody who is a collaborator to the repo is allowed to trigger the release workflow from [Actions > Create release](https://github.com/coveo/ui-kit/actions/workflows/release.yml).

Additionally, the workflow is run in the "Release" environment as you can see in `release.yml`, which requires approval immediately after triggering the workflow.
image

aprudhomme-coveo: But what about the publish:npm:release in @atomic ? Would calling this start a release process for all of ui-kit ?

packages/headless/contributors/adding-a-sub-package.md
```diff
@@ -8,7 +8,7 @@ Headless provides exports through multiple sub packages. A sub-package groups to
 1. Create an entry file for your sub-package inside the `src` directory (e.g. case-assist.ts).
 2. Configure nodejs and browser bundles inside `rollup.config.js` for your entry file.
 3. Create a new directory with the name of your sub-package at the project root.
-4. Inside the new directory, add a `package.json` file. Add the paths to your bundled files and type definitions. Mark the package `private` to prevent `lerna` from bumping it's version.
```
mrrajamanickam-coveo, line 11: Don't we still need to set private if we don't want the pkg to be published?

.deployment.config.json
```diff
@@ -6,10 +6,10 @@
       "sequential": ["FIRST_MAIN_REGION"]
     },
     "environments_order": {
-      "sequential": $[ENVIRONMENTS]
+      "sequential": ["dev", "stg", "prd"]
```
mrrajamanickam-coveo, line 9: Is ENVIRONMENTS not usable here anymore? Why?

.deployment.config.json
```diff
@@ -6,10 +6,10 @@
...
     },
     "team_jenkins": "searchuibuilds",
-    "start_environment_automatically": false,
+    "start_environment_automatically": true,
```
mrrajamanickam-coveo, line 12: What does this do? Why is this change required?

.deployment.config.json
```diff
@@ -191,16 +191,6 @@
...
-      "id": "stg-release",
-      "team_jenkins": {
-        "disabled": true,
-        "job_name": "ui-kit-stg-release",
-        "stg": {
-          "disabled": false
-        }
-      }
-    },
     {
```
mrrajamanickam-coveo, line 194: Why is this step being removed?

nx.json
```diff
@@ -39,12 +39,6 @@
     },
     "lint:check": {},
     "lint:fix": {},
```
mrrajamanickam-coveo, line 41: So looks like nx is not involved with the release process anymore?

mrrajamanickam-coveo (commented):
Most questions to understand the change

packages/headless/contributors/adding-a-sub-package.md
louis-bompart: Overall, if a package is not meant to be published, it's good practice to mark it as such

CONTRIBUTING.md
btaillon: I'm removing the publish:npm scripts currently. Prior to this PR, it would've only published the current version from all `package.json` files to NPM.

.deployment.config.json
btaillon: It's the other way around, we no longer need this to be `false`.

Having this set to `false` meant that we were getting prompted before a package would progress from `dev` to `stg` and `stg` to `prd`. However, to accept it, we needed to connect to some of our back-end infrastructure, which could take some time. Now, since releases are manually triggered by a maintainer, we don't need this.

.deployment.config.json
btaillon: No longer needed. This executed the staging Jenkinsfile, which only updated the beta tag. Now, we already publish with the beta tag.

nx.json
btaillon: It's still involved, it's just that these scripts are on longer used.

PR #2939, opened 2023-05-26
feat(atomic): changed translation key in `atomic-did-you-mean`

aprudhomme-coveo:
https://coveord.atlassian.net/browse/CDX-1426

Adds a condition that checks if the `no-results-for-dym` key exist. If it does, it uses it for the localized string.

#### Breaking change problem
This avoids a breaking change by leaving the `no-results-for` key as the default value. However, this is bad code. The new key (`no-results-for-dym`) does not show in the `i18n.getResourceBundle()` function. The correct way to do this would be to have a breaking change and create an entirely new key in locales.json. 

#### My Solution
- Merge this bandaid PR 
- Add a spike/story in the [V3 update epic ](https://coveord.atlassian.net/browse/KIT-2157) to look for other places in the code where splitting translation keys would be useful. 
- In the spike/story, add a child issue to revert this PR changes and simply add a new key in locales.json.

### DIFF
--- a/packages/atomic/src/components/search/atomic-did-you-mean/atomic-did-you-mean.tsx
+++ b/packages/atomic/src/components/search/atomic-did-you-mean/atomic-did-you-mean.tsx
@@ -53,8 +53,7 @@ export class AtomicDidYouMean implements InitializableComponent {
 
   private withQuery(
     key:
-      | 'no-results-for'
-      | 'no-results-for-dym'
+      | 'no-results-for-did-you-mean'
       | 'query-auto-corrected-to'
       | 'showing-results-for',
     query: string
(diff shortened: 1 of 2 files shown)


mrrajamanickam-coveo:
> Do I absolutely need to add a new key in locales.json or is this fine ? 

- Looking at the discussion that initiated this change (quoted below), it seems like we would need both keys to differentiate between two different use cases?
- Wouldn't you need the entry in locales.json to test this out? Wonder if the docs team would help with adding locales entries for the new key? @jpmarceau 

>> atomic-did-you-mean and atomic-no-results are using the same translation key “no-results-for”, I couldn’t differentiate the translation between these two components

> If I need to add a new key, my solution was to detect if the no-results-for-dym changed but it is a bit clunky.

Not sure I understand what you mean. Can you please explain.

aprudhomme-coveo:
> Not sure I understand what you mean. Can you please explain.

The main issue is to avoid a breaking change. If I were to add another key,  I would need to make sure that this new key `no-results-for-dym` points to the `no-results-for` key by default. Then, I would need to detect if the user is changing `no-results-for-dym` to remove the pointing to `no-results-for`.

jpmarceau:
> * Wouldn't you need the entry in locales.json to test this out? Wonder if the docs team would help with adding locales entries for the new key? @jpmarceau

Hello @mrrajamanickam-coveo , what we typically do is:
1. You guys add the locales to the best of your ability in your PRs. If you can't write French, then just use placeholders.
2. We review them in the PR and suggest adjustments as required.

olamothe:
I would add a new key for `no-results-for-did-you-mean` and reference it in the did you mean component.

packages/atomic/src/components/search/atomic-did-you-mean/atomic-did-you-mean.tsx
```diff
@@ -52,7 +52,11 @@ export class AtomicDidYouMean implements InitializableComponent {
   }
 
   private withQuery(
-    key: 'no-results-for' | 'query-auto-corrected-to' | 'showing-results-for',
+    key:
+      | 'no-results-for'
+      | 'no-results-for-dym'
```
btaillon, line 57: I prefer to avoid abbreviations and use the full words.
`no-results-for-did-you-mean`

olamothe (approved):

PR #2941, opened 2023-05-26
build(lockfile): add lockfile=3 gate

louis-bompart:
Copied from coveo/cli, migrated to ESM, workflow adapted to fit in ui-kit existing workflow.

https://coveord.atlassian.net/browse/KIT-2502

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,41 +1,86 @@
 {
-  "ignorePatterns": ["node_modules", "stencil-generated", "dist", "www"],
+  "ignorePatterns": [
+    "node_modules",
+    "stencil-generated",
+    "dist",
+    "www"
+  ],
   "env": {
     "jest": true,
     "es6": true
   },
-  "extends": ["prettier"],
+  "extends": [
+    "prettier"
+  ],
   "parserOptions": {
     "ecmaVersion": 2020,
     "sourceType": "module"
   },
   "overrides": [
     {
-      "files": ["**/*.ts", "**/*.tsx"],
+      "files": [
+        "**/package-lock.json"
+      ],
+      "parser": "eslint-plugin-json-es",
+      "plugins": [
+        "package-lock"
+      ],
+      "rules": {
+        "package-lock/lock-file-version": [
+          "error",
+          {
+            "version": 3
+          }
+        ]
+      }
+    },
+    {
+      "files": [
+        "**/*.ts",
+        "**/*.tsx"
+      ],
       "parser": "@typescript-eslint/parser",
-      "plugins": ["@typescript-eslint"],
-      "extends": ["./node_modules/gts"],
+      "plugins": [
+        "@typescript-eslint"
+      ],
+      "extends": [
+        "./node_modules/gts"
+      ],
       "parserOptions": {
         "jsxPragma": "h"
       },
       "rules": {
         "@typescript-eslint/no-unused-vars": [
           "warn",
-          {"ignoreRestSiblings": true}
+          {
+            "ignoreRestSiblings": true
+          }
         ],
         "@typescript-eslint/explicit-module-boundary-types": "off",
         "@typescript-eslint/ban-types": "off",
         "@typescript-eslint/no-empty-interface": [
           "error",
-          {"allowSingleExtends": true}
+          {
+            "allowSingleExtends": true
+          }
         ]
       }
     },
     {
-      "files": ["**/*.js", "**/*.jsx"],
-      "extends": ["eslint:recommended"],
+      "files": [
+        "**/*.js",
+        "**/*.jsx"
+      ],
+      "extends": [
+        "eslint:recommended"
+      ],
       "rules": {
-        "no-unused-vars": ["error", {"argsIgnorePattern": "^_"}]
+        "no-unused-vars": [
+          "error",
+          {
+            "argsIgnorePattern": "^_"
+          }
+        ]
       },
       "env": {
         "node": true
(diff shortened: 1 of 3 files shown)


.github/workflows/prbot.yml
```diff
@@ -16,6 +16,13 @@ jobs:
           node-version-file: '.nvmrc'
       - run: npm ci
       - run: npm run pr:report
+  lockfile-version:
+    name: 'Check lockfile version'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - run: node scripts/check-lockfile-versions.js $(find ./ -xdev -wholename '**/package-lock.json')
```
mrrajamanickam-coveo, line 25: could we not have this in lint-staged also / instead ?

mrrajamanickam-coveo: On a related note is there any way to detect if the pkg lock has been deleted and regenerated (inadvertently during a merge conflict e.g.)

mrrajamanickam-coveo: wouldn't having an engine in pkg json and engine strict setting prevent users from using the wrong node/npm version?

mrrajamanickam-coveo: we already have node-glob as a dep. Can we use that instead of find?

scripts/check-lockfile-version.mjs
```diff
@@ -0,0 +1,22 @@
...
+import {resolve} from 'node:path';
+
+const lockFileVersionFinder = /(?<="lockfileVersion":\s)\d(?=,)/;
+
+const lockFilePaths = process.argv[2].toString().split('\n');
+let err = '';
+for (const lockFilePath of lockFilePaths) {
+  console.log(lockFilePath);
+  const lockFileText = readFileSync(resolve(lockFilePath)).toString();
+  // Package-lock.json are bulky and the info we need is right at the top of the file, so regex-match to go fast.
```
mrrajamanickam-coveo, line 14: can readSync be used with `length` arg instead for better perf?

.github/workflows/prbot.yml
louis-bompart: > could we not have this in lint-staged also / instead
Sure.
- [ ] Move/add gate on lint-staged

> On a related note is there any way to detect if the pkg lock has been deleted and regenerated (inadvertently during a merge conflict e.g.)

Mhh I don't see how that'd be possible. We could do more checks on the lockfile, such as checking if some package are "where they should be", but I reckon the cost of coding and maintaining the rules ain't worth the additional gain.

> wouldn't having an engine in pkg json and engine strict setting prevent users from using the wrong node/npm version?

We could enforce `npm9` indeed, it would help it would not prevent users from comitting wrong lockfileversion.
One could've set something like lockfile-version=2 in its global npmrc, and thus cause the issue to persist even with the latest npm version

>we already have node-glob as a dep. Can we use that instead of find?

Sure, this makes sense if we put the gate on the pre-commit

scripts/check-lockfile-version.mjs
louis-bompart: If I want to keep the script generic, not really.
The first thing I thought of was going async and use a pipe to detect when we have the data, but same thing: cannot be sure that the data I got at a given time contains everything, so I can have let's say `endofthepackagename",\n\t"lockfi` followed by `"leVersion": 3,`

At this stage, I think the complexity is not worth the gain in speed.

.github/workflows/prbot.yml
louis-bompart: - [ ]  add lockfile-version=3 in npmrc

y-lakhdar (approved):

scripts/check-lockfile-version.mjs
mrrajamanickam-coveo: Makes sense 👍

.github/workflows/prbot.yml
mrrajamanickam-coveo: Sounds good 👍

mrrajamanickam-coveo:
@louis-bompart How about (as a followup even) prevent any local updates to pkg lock file (with lint-staged) and only update it in CI. That will ensure that the correct version is used and also the lock file hasn't been inadvertently deleted and regenerated leading to issues?

.github/workflows/prbot.yml
btaillon, line 25: might be easier to have a bash script than add a new separate js script.
example:
```suggestion
      - run: |
        exitCode=0
        for lockfile in $(git ls-files {**/,}package-lock.json); do
          version=$(cat $lockfile | jq .lockfileVersion)
          if [ $version -ne 3 ]; then
            echo "Expected lockfileVersion of $lockfile to be 3." >&2
            exitCode=1
          fi
        done
        exit $exitCode
```

btaillon: Also, we should probably use ls-files instead of find since`find` searches through `node_modules`

btaillon (approved):

louis-bompart:
> @louis-bompart How about preventing any local updates to pkg lock file (with lint-staged) and only update it in CI. That will ensure that the correct version is used and also the lock file hasn't been inadvertently deleted and regenerated leading to issues?

Mh, I'm really not found of that, it reverse on its head the meaning of a lockfile (they are mostly meant to ensure a ci install the exact same deps as the user machine), and adding dependencies would become a real pain.

.github/workflows/prbot.yml
louis-bompart: > FYI just came across https://www.npmjs.com/package/eslint-plugin-package-lock which seems to do the same check. Would be better if we can use it as it would be a eslint plugin and would go with the lint workflow. What do you think?

simple as pie, this is nice. reworking with this.

PR #2944, opened 2023-05-29
fix: undefined error when checking mismatch with new endpoint url 

mrrajamanickam-coveo:
Check for mismatch only when new endpoint url is defined to avoid undefined error

```
TypeError: Cannot read properties of undefined (reading 'match')
    at matchCoveoOrganizationEndpointUrlAnyOrganization (url-utils.ts:61:21)
    at shouldWarnAboutMismatchBetweenOrganizationIDAndOrganizationEndpoints (engine.ts:330:17)
    at buildEngine (engine.ts:196:5)
    at buildSearchEngine (search-engine.ts:124:18)
    at AtomicSearchInterface.initEngine (atomic-search-interface.tsx:394:21)
    at atomic-search-interface.tsx:286:51
    at CommonAtomicInterfaceHelper.onInitialization (interface-common.tsx:100:5)
    at AtomicSearchInterface.internalInitialization (atomic-search-interface.tsx:520:38)
    at AtomicSearchInterface.initialize (atomic-search-interface.tsx:286:17)
    at index.js:1979:97 
```

##### TODO

- [x] Add/update test 

https://coveord.atlassian.net/browse/KIT-2504


mrrajamanickam-coveo:
Test failures (unrelated - known flaky)

```
1) Color Facet Test Suites
       when selecting the "Show more" button
         when selecting the "Show less" button
           verify analytics
             should log the facet show less results to UA:

      AssertionError: expected { Object (language, userAgent, ...) } to have property 'eventValue' of 'showLessFacetResults', but got 'showMoreFacetResults'
      + expected - actual

      -'showMoreFacetResults'
      +'showLessFacetResults'
```

packages/headless/src/app/engine.ts
```diff
@@ -192,18 +192,17 @@ export function buildEngine<
...
-    )
-  ) {
-    engine.logger.warn(
-      `There is a mismatch between the \`organizationId\` option (${options.configuration.organizationId}) and the organization configured in the \`organizationEndpoints\` option (${options.configuration.organizationEndpoints?.platform}). This could lead to issues that are complex to troubleshoot. Please make sure both values match.`
-    );
-  }
-
   if (organizationEndpoints?.platform) {
     platformUrl = organizationEndpoints.platform;
+    if (
```
olamothe, line 197: Seems like a better fix would instead be adding a missing null check in `shouldWarnAboutMismatchBetweenOrganizationIDAndOrganizationEndpoints` 🤔  ?

```typescript
function shouldWarnAboutMismatchBetweenOrganizationIDAndOrganizationEndpoints(
  options: EngineOptions<ReducersMapObject>
) {
  if (isUndefined(options.configuration.organizationEndpoints)) {
    return false;
  }
  const {platform} = options.configuration.organizationEndpoints;

  if (isUndefined(platform) {
    return false;
  }

  const match = matchCoveoOrganizationEndpointUrlAnyOrganization(
    options.configuration.organizationEndpoints.platform
  );
  return match && match.organizationId !== options.configuration.organizationId;
}
```

packages/headless/src/app/engine.ts
```diff
@@ -177,22 +177,22 @@ export function buildEngine<
...
   const {organizationEndpoints} = options.configuration;
-  let {platformUrl} = options.configuration;
-
-  if (shouldWarnAboutOrganizationEndpoints(options)) {
-    // @v3 make organizationEndpoints the default.
-    engine.logger.warn(
-      'The `organizationEndpoints` options was not explicitly set in the Headless engine configuration. Coveo recommends setting this option, as it has resiliency benefits and simplifies the overall configuration for multi-region deployments. See [Organization endpoints](https://docs.coveo.com/en/mcc80216).'
-    );
-  }
+  const platformUrl = organizationEndpoints?.platform
```
mrrajamanickam-coveo, line 180: refactored platform url init to one place with org end point overriding the platform url

packages/headless/src/app/engine.ts
```diff
@@ -177,22 +177,22 @@ export function buildEngine<
...
     );
   }
 
-  if (
+  if (shouldWarnAboutOrganizationEndpoints(options)) {
+    // @v3 make organizationEndpoints the default.
+    engine.logger.warn(
+      'The `organizationEndpoints` options was not explicitly set in the Headless engine configuration. Coveo recommends setting this option, as it has resiliency benefits and simplifies the overall configuration for multi-region deployments. See [Organization endpoints](https://docs.coveo.com/en/mcc80216).'
+    );
+  } else if (
```
mrrajamanickam-coveo, line 195: sine warning about mismatch makes sense only when org end point is defined

packages/headless/src/app/engine.ts
```diff
@@ -318,17 +314,20 @@ function shouldWarnAboutOrganizationEndpoints(
 }
 
 function shouldWarnAboutPlatformURL(options: EngineOptions<ReducersMapObject>) {
-  return !isNullOrUndefined(options.configuration.platformUrl);
+  return isNullOrUndefined(
+    options.configuration.organizationEndpoints?.platform
```
mrrajamanickam-coveo, line 318: if org end point url is defined do we care about the older platform url since it would not be used anyways?

mrrajamanickam-coveo (commented):
Made some refactoring - let me know if I am missing something

packages/headless/src/app/engine.ts
mrrajamanickam-coveo: Giving this some more thought and the logical flow - made some refactoring. Let me know if I am missing something.

packages/headless/src/app/engine.ts
```diff
@@ -177,34 +177,31 @@ export function buildEngine<
   const engine = buildCoreEngine(options, thunkExtraArguments);
   const {accessToken, organizationId} = options.configuration;
   const {organizationEndpoints} = options.configuration;
-  let {platformUrl} = options.configuration;
+  const platformUrl = organizationEndpoints?.platform
+    ? organizationEndpoints.platform
+    : options.configuration.platformUrl!;
```
louis-bompart, line 182:
```suggestion
  const platformUrl = organizationEndpoints?.platform || options.configuration.platformUrl!
```
Or 
```suggestion
  const platformUrl = organizationEndpoints?.platform ?? options.configuration.platformUrl!
```

louis-bompart (approved):

packages/headless/src/app/engine.ts
mrrajamanickam-coveo: Thanks Louis. I implemented the suggestion but without the `!` at end as I noticed 
- it does not exist in master - I must have added in inadvertently during refactoring
- there is schema validation down that would catch it, so it is redundant

packages/headless/src/app/engine.ts
olamothe: I think we should still warn: if an implementer set an option that's discarded/irrelevant/not needed anymore, it's still meaningful to warn and inform them.

Plus, next major, we will actually want to remove entirely the `platformUrl` option, so it also a preventive warning that it's going away, and it might actually throw error in next version.

mrrajamanickam-coveo: Makes sense 👍

y-lakhdar (approved):

olamothe (approved):

PR #2947, opened 2023-05-31
fix(atomic-react): bad esm-export

### LINKED ISSUE #2928: [Bug]: ESM support

Atomic

### product version

2.2.33

### bug description

The `@coveo/atomic-react` is built & published to npm in a way that isn't compliant with [ESM](https://nodejs.org/api/esm.html#introduction). In fact, `@coveo/atomic` seems to have some more minor issues about its configuration aswell.

### Steps to reproduce

[Publint](https://github.com/bluwy/publint) scan of `@coveo/atomic-react`: https://publint.dev/@coveo/atomic-react@2.2.33
[Publint](https://github.com/bluwy/publint) scan of `@coveo/atomic`: https://publint.dev/@coveo/atomic@2.30.1

**Suggestion**: run `publint` on `prepublishOnly` script for each package published to npm.

### Relevant log output

_No response_

louis-bompart:
Fixes #2928 
Dangerous change, discussion on Slack, because we need to draw a line in the sand.

My proposal is to only fix the error because it just plain doesn't work and might 💥 stuff.
V3 however, I think we could go wild and simply wipe the slate and redefine the export contract of all public packages.

### DIFF
--- a/packages/atomic-react/package.json
+++ b/packages/atomic-react/package.json
@@ -27,9 +27,9 @@
       "require": "./dist/cjs/index.js",
       "types": "./dist/index.d.ts"
     },
-    "./*": {
-      "types": "./dist/*.d.ts",
-      "default": "./dist/*.js"
+    "./**/*": {
+      "types": "./dist/**/*.d.ts",
+      "default": "./dist/**/*.js"
     }
   },
   "module": "./dist/index.js",
--- a/packages/atomic-react/package.json
+++ b/packages/atomic-react/package.json
@@ -21,7 +21,7 @@
     "build:assets": "ncp ../atomic/dist/atomic/assets dist/assets && ncp ../atomic/dist/atomic/lang dist/lang "
   },
   "main": "./dist/cjs/index.js",
-  "module": "./dist/index.esm.js",
+  "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "files": [
     "dist/",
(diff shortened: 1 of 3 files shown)


packages/atomic-react/package.json
```diff
@@ -21,7 +21,18 @@
     "build:assets": "ncp ../atomic/dist/atomic/assets dist/assets && ncp ../atomic/dist/atomic/lang dist/lang "
   },
   "main": "./dist/cjs/index.js",
-  "module": "./dist/index.esm.js",
+  "exports": {
```
mrrajamanickam-coveo, line 24: Would these be helpful (for now or future)
https://www.npmjs.com/package/package-json-exports
https://www.npmjs.com/package/eslint-import-resolver-exports

louis-bompart: - https://www.npmjs.com/package/package-json-exports

While I understand the motivation, I don't think it'd be necessarily a good fit or idea.
Indeed, this is an easy way to end up with a lot of exports and entrypoints, which imo has adverse consequences for every party involved:
 - For the maintainer, they have more entrypoints when doing changes, and also more chance to mess up.
 - For the user, they may need to have lines and lines of import, or have 'choices' to make "should I import X from subModuleA or subModuleB??". This could lead to an unpleasant UX I think.
 
 Right now, gut feeling, we're going toward having entrypoints per 'use-cases'. Those entrypoints are to be manually crafted.
 
 - https://www.npmjs.com/package/eslint-import-resolver-exports
Yes this one (and the one it refer) would be useful to ensure we don't break an entrypoint when moving files around.
Nice one :D

louis-bompart:
Won't do in v2

PR #2951, opened 2023-05-31
ci: remove force push from version bump logic

btaillon:
https://coveord.atlassian.net/browse/KIT-2510

We can't force push to `master` with a deploy key. We can, however, update the remote `master` branch using Octokit.

### DIFF
--- a/utils/release/git-publish-all.mjs
+++ b/utils/release/git-publish-all.mjs
@@ -12,11 +12,11 @@ import {
   gitCommitTree,
   gitUpdateRef,
   gitPublishBranch,
-  gitSetRefOnCommit,
-  gitPush,
+  gitPull,
 } from '@coveo/semantic-monorepo-tools';
 import {createAppAuth} from '@octokit/auth-app';
 import {spawnSync} from 'child_process';
+import {randomUUID} from 'crypto';
 import {readFileSync} from 'fs';
 import {Octokit} from 'octokit';
 import {dedent} from 'ts-dedent';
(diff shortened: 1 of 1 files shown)


utils/release/git-publish-all.mjs
```diff
@@ -113,13 +111,15 @@ async function commitChanges(commitMessage, octokit) {
...
-  await gitPush({remote: GIT_SSH_REMOTE, refs: [mainBranchName], force: true});
+  await octokit.rest.git.updateRef({
+    owner: REPO_OWNER,
+    repo: REPO_NAME,
+    ref: `heads/${mainBranchName}`,
+    sha: commit.data.sha,
+    force: true, // Needed since the remote main branch contains a "lock" commit.
+  });
+  await gitCheckoutBranch(mainBranchName);
+  await gitPull();
```
louis-bompart, line 122: Mhh is the pull remote->local linear?
I think because of the lock commit, it won't but maybe I'm wrong? 🤔

btaillon: I ran the following test to see if this works:
1. Clone the local repo twice into two different directories (A and B).
2. From repo A, commit and push something (this simulates the "lock" commit).
3. Push the commit.
4. From repo A, do `git reset --hard HEAD^1`.
5. From repo B (which is now behind by 1 commit), commit and push something else.
6. From repo B, try to push the commit (this expectedly fails because the history is non-linear).
7. From repo B, force push the new commit (this expectedly works).
8. From repo A, do `git pull` (without any parameter).

Although vscode showed that repo A was one behind the upstream, and it probably remembered the commit we reset from, `git pull` worked and pulled the commit from repo B.

louis-bompart: Ok gotcha, I forgot that the git lock was reseting the local master to the commit before the lock commit, that's what confused me.

louis-bompart (approved):
LGTM!

PR #2955, opened 2023-06-01
fix(headless): prevent overriding queries from a different origin

btaillon:
https://coveord.atlassian.net/browse/KIT-2480

Somebody reported an issue where the instant results search in Atomic was interrupting the main search. I'm no longer able to reproduce the issue in neither our UI nor theirs, but this PR should logically fix the issue.


packages/headless/src/utils/abort-controller-polyfill.ts
```diff
@@ -0,0 +1,13 @@
+export function createAbortController(): AbortController | null {
+  // For nodejs environments only, we want to load the implementation of AbortController from node-abort-controller package.
+  // For browser environments, we need to make sure that we don't use AbortController as it might not be available (Locker Service in Salesforce)
+  // This is not something that can be polyfilled in a meaningful manner.
+  // This is a low level browser API after all, and only JS code inside a polyfill cannot actually cancel network requests done by the browser.
+
+  if (typeof window === 'undefined') {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const nodeAbort = require('node-abort-controller');
```
olamothe, line 9: Unrelated, but we will ultimately want to remove this inline require: Webpack and co really dont like it and warn about it at compile time.

packages/headless/src/api/search/search-api-client.ts
```diff
@@ -263,20 +265,34 @@ export class SearchAPIClient implements FacetSearchAPIClient {
...
-      return new nodeAbort() as AbortController;
+  private clearAbortController(
+    origin: SearchOrigin | 'unknown',
+    controllerToClear: AbortController
+  ) {
+    if (
+      this.searchAbortControllers[origin]?.signal !== controllerToClear.signal
+    ) {
+      return;
     }
```
btaillon, line 294: Added this to prevent possible race conditions. I tried to write unit tests to verify the race condition, but they always failed, even with that fixed, and it took too long to dig into what was causing them to fail. I suspect the thing that caused them to fail was some weird behaviour I observed with Promises and mock timers in Jest, since stepping through the code executed in an order that didn't seem logical.

packages/headless/src/api/search/search-api-client.ts
```diff
@@ -263,20 +265,34 @@ export class SearchAPIClient implements FacetSearchAPIClient {
     };
   }
 
-  private getAbortControllerInstanceIfAvailable(): AbortController | null {
```
btaillon, line 266: Moved this function to `abort-controller-polyfill.ts`.

packages/headless/src/api/search/search-api-client.ts
```diff
@@ -124,25 +125,26 @@ export class SearchAPIClient implements FacetSearchAPIClient {
     };
   }
 
-  private searchAbortController: AbortController | null = null;
+  private searchAbortControllers: Partial<
```
olamothe, line 128: A refactor opportunity would be to create an `APIClientAbortController` class, which could hold the state implemented here, a constructor which would wrap `createAbortController` function, and, a reference to the `PlatformClient.call` function which could then automatically `clear` itself once the request finishes.

What do you think ?

btaillon: Yeah, I thought about making a dedicated class, this feels pretty intuitive to abstract and would look kinda clean, but I'm not sure if it would actually simplify the code that much. I also feel like it's one more thing we'd have to step through while debugging. It does shorten the file a bit though, and I like reducing visual pollution like this.

I'm open to changing it, I'm curious what others think too.

packages/headless/src/utils/abort-controller-polyfill.ts
btaillon: Removing it now

louis-bompart (approved):

olamothe (approved):

packages/headless/src/api/search/search-api-client.ts
btaillon: Refactored into APICallsQueue

packages/headless/src/utils/abort-controller-polyfill.ts
```diff
@@ -0,0 +1,13 @@
+export function createAbortController(): AbortController | null {
+  // For nodejs environments only, we want to load the implementation of AbortController from node-abort-controller package.
+  // For browser environments, we need to make sure that we don't use AbortController as it might not be available (Locker Service in Salesforce)
+  // This is not something that can be polyfilled in a meaningful manner.
+  // This is a low level browser API after all, and only JS code inside a polyfill cannot actually cancel network requests done by the browser.
```
mrrajamanickam-coveo, line 5: "only JS code inside a polyfill cannot actually" 
did you mean "only JS code inside a polyfill can actually"

PR #2956, opened 2023-06-01
chore: update supported node version to >=16

btaillon:
https://coveord.atlassian.net/browse/KIT-2514

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -43,8 +43,5 @@
   "devDependencies": {
     "rimraf": "3.0.2",
     "vite": "2.9.15"
-  },
-  "engines": {
-    "node": ">=10"
   }
 }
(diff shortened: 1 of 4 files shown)


packages/auth/package.json
```diff
@@ -43,8 +43,5 @@
   "devDependencies": {
     "rimraf": "3.0.2",
     "vite": "2.9.15"
-  },
-  "engines": {
-    "node": ">=10"
   }
```
btaillon, line 46: `@coveo/auth` uses the fetch method from `browser-fetch.ts`, which only contains:
```ts
export type BrowserFetch = typeof fetch;

export function getBrowserFetch(): BrowserFetch {
  const isBrowser = typeof window !== 'undefined';
  return isBrowser ? fetch : () => Promise.resolve(new Response());
}
```

Given that it doesn't support Node.js (yet?), I'm not sure if we should or should not have `engines` here. Opinions?

PR #2959, opened 2023-06-05
feat(quantic): Added check for collection children in quanticResultChildren component

SimonMilord:
[SFINT-5050](https://coveord.atlassian.net/browse/SFINT-5050)

This is a PR I opened to make this fix built in to Quantic rather than in the templates of Salesforce Integration.
We add a check for collection children to make sure that if the result is not a folding result, it does not show the "no related items" message unnecessarily. We basically want this message only if the user has loaded a folded collection and no items were found.

**As a Dev**
I want to make this component display its content when it detects that the result has child results, otherwise it just displays nothing. This allows to make this validation built in in quantic.

**As an Admin/user**
I will not have to implement this logic in my code, it will be built in the Quantic Library.

**Acceptance criteria**
Should not display “no related items” in non-folding results

Should support folding properly when the result is a folding result.

**Technical details**
Use the collection.children to validate

DEMO:

image

[SFINT-5050]: https://coveord.atlassian.net/browse/SFINT-5050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultChildren/__tests__/quanticResultChildren.test.js
```diff
@@ -142,7 +142,7 @@ describe('c-quantic-result-children', () => {
   });
 
   describe('when there are no child results available in the collection', () => {
-    it('should not display any child results and display the no more child results message', async () => {
```
SimonMilord, line 145: When there are no children in the collection, we no longer show the message, no children results will be displayed.

packages/quantic/force-app/main/default/lwc/quanticResultChildren/__tests__/quanticResultChildren.test.js
```diff
@@ -156,12 +156,8 @@ describe('c-quantic-result-children', () => {
       const childResult = element.shadowRoot.querySelector(
         selectors.quanticResult
       );
-      const noMoreChildrenMessage = element.shadowRoot.querySelector(
-        selectors.noMoreChildrenMessage
-      );
 
       expect(childResult).toBeNull();
-      expect(noMoreChildrenMessage).not.toBeNull();
```
mmitiche, line 164: Doing like this you are not testing that the `noMoreChildrenMessage` is not displayed.
Like we need to test if it is displayed, we also need to test if it's not displayed by doing:
```
 expect(noMoreChildrenMessage).toBeNull();
```

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.html
```diff
@@ -1,31 +1,48 @@
...
+          ></lightning-button>
+        </div>
+      </template>
 
-    <template if:true={shouldDisplayNoChildrenMessage}>
-      <div data-cy="result-children__no-more-children-message"
-        class="result-children__no-more-children-message slds-grid slds-grid_align-end slds-var-p-vertical_x-small slds-text-color_weak">
-        {labels.noRelatedItems}
-      </div>
+      <template if:true={shouldDisplayNoChildrenMessage}>
```
mmitiche, line 38: Instead of applying a new condition on all the content you could just update this condition `shouldDisplayNoChildrenMessage` that controls the display of the noChildrenMessage for more simplicity.

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.js
```diff
@@ -182,6 +182,10 @@ export default class QuanticResultChildren extends LightningElement {
     return !this.areMoreResultsAvailable;
   }
 
+  get hasCollectionChildren() {
```
mmitiche, line 185: As mentioned above, we can update the logic that controls the display of the noChildrenMessage in the already existing condition: `shouldDisplayNoChildrenMessage`

erocheleau: Yea now since you are in this same component, you could just add a new condition to the `shouldDisplayNoChildrenMessage` however make sure no unnecessary HTML will be renderer if the component doesn't have any children.

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.html
SimonMilord: yeah thats also an idea :p

packages/quantic/force-app/main/default/lwc/quanticResultChildren/__tests__/quanticResultChildren.test.js
SimonMilord: yeah I guess with the additional check in the `shouldDisplayNoChildrenMessage` it makes more sense to test that it is not shown

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.js
```diff
@@ -173,7 +173,10 @@ export default class QuanticResultChildren extends LightningElement {
   }
 
   get shouldDisplayNoChildrenMessage() {
```
mmitiche, line 175: I'm pretty sure we can simplify the condition to this:
```
  get shouldDisplayNoChildrenMessage() {
    if (!this.isFirstLevelChildCollection || !this.areAllChildResultsLoaded) {
      return false;
    }
    return !this.moreResultsFound;
  }
```
In other words, we display the `NoChildrenMessage` only in one case, which is when the end user has clicked the `load all items` but no more results have been found.

SimonMilord: nice! you're right, will refactor

mmitiche (approved):
LGTM!

packages/quantic/cypress/e2e/default-2/result-list/folded-result-list/folded-result-list.cypress.ts
```diff
@@ -139,11 +139,11 @@ describe('quantic-folded-result-list', () => {
     });
 
     describe('when the result has no child result', () => {
```
erocheleau, line 141: I think it's more "when the result has no child collection" no?

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.js
erocheleau: ohhh yea ok I was about to write that the two were not equivalent at all, but you're right, it's only whenever we tried to load children and there was none to load

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.js
```diff
@@ -173,13 +174,10 @@ export default class QuanticResultChildren extends LightningElement {
...
-    if (!this.isFirstLevelChildCollection) {
+    if (!this.isFirstLevelChildCollection || !this.areAllChildResultsLoaded) {
       return false;
     }
-    if (this.areAllChildResultsLoaded) {
-      return !this.moreResultsFound;
-    }
-    return !this.areMoreResultsAvailable;
+    return !this.moreResultsFound;
```
erocheleau, line 180: Whenever a result doesn't even have a collection, this won't work no?

If a result doesn't even have a collection `this.moreResultsFound` should be `false`, so this will return `true`?

erocheleau (commented):
Something is fishy about the condition to display the no more results

packages/quantic/cypress/e2e/default-2/result-list/folded-result-list/folded-result-list.cypress.ts
SimonMilord: the child result is in the collection, but yeah I can change it to collection if you prefer. Can change that in my FTs PR though

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.js
mmitiche: If a result doesn't even have a collection `this.areAllChildResultsLoaded` will always be `false`. It means that we will always  return `false` following the first block of the condition.

erocheleau: I think it's a bit confusing the way these conditions are paired now, essentially would this make sense instead?

```
if (!this.isFirstLevelChildCollection) {
    return false;
}
return this.areAllChildResultsLoaded && !this.moreResultsFound;
```

erocheleau (approved):
Just check out my suggestion of condition, I think it's more readable if it's equivalent.

If it's not equivalent, you should consider it approved.

packages/quantic/force-app/main/default/lwc/quanticResultChildren/quanticResultChildren.js
SimonMilord: Yeah I think this is more clear IMO since moreResultsFound is tied to the loading

PR #2961, opened 2023-06-05
refactor(headless): add types to request and responses to support automatic facets

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2523

This PR adds necessary types to support the automatic facets feature. 

I used `refactor` to not trigger semantic versioning.

### DIFF
--- a/packages/headless/src/api/search/search-api-params.ts
+++ b/packages/headless/src/api/search/search-api-params.ts
@@ -124,6 +124,11 @@ export interface ExcerptLength {
 export interface AuthenticationParam {
   authentication?: string;
 }
+export interface AutomaticFacetsParams {
+  generateAutomaticFacets?: {
+    desiredCount?: number;
+  };
+}
 
 export const baseSearchRequest = (
   req: BaseParam & AuthenticationParam,
--- a/packages/headless/src/api/search/search-api-params.ts
+++ b/packages/headless/src/api/search/search-api-params.ts
@@ -126,7 +126,7 @@ export interface AuthenticationParam {
 }
 export interface AutomaticFacetsParams {
   generateAutomaticFacets?: {
-    desiredCount?: number;
+    desiredCount: number;
   };
 }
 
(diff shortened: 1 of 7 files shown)


packages/headless/src/features/facets/facet-set/interfaces/response.ts
```diff
@@ -5,3 +5,8 @@ export interface FacetValue extends BaseFacetValue {
 }
 
 export type FacetResponse = BaseFacetResponse<FacetValue>;
+
+export type AutomaticFacetResponse = Omit<
+  BaseFacetResponse<FacetValue>,
+  'facetId'
```
aprudhomme-coveo, line 11: The facets returned by the automatic feature follow the same pattern as `BaseFacetResponse<FacetValue>` except they don't have a `facetId`.

packages/headless/src/api/search/search/search-response.ts
```diff
@@ -13,6 +14,7 @@ import {SecurityIdentity} from './security-identity';
 import {PhrasesToHighlight, TermsToHighlight} from './stemming';
 
 export interface SearchResponseSuccess {
+  generateAutomaticFacets: AutomaticFacets;
```
olamothe, line 17: Double check the optionality of the value in the response.

I think `generatedAutomaticFacets` won't be returned by the API if none are requested ? And if so, it should be `generateAutomaticFacets?: AutomaticFacets`

It's very important that all the types defined here match very well how the API behaves, otherwise it will start creeping up as bug in a lot of different places.

packages/headless/src/api/search/search-api-params.ts
```diff
@@ -124,6 +124,11 @@ export interface ExcerptLength {
 export interface AuthenticationParam {
   authentication?: string;
 }
+export interface AutomaticFacetsParams {
+  generateAutomaticFacets?: {
+    desiredCount?: number;
```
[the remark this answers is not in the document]
aprudhomme-coveo: I made a mistake, desired count is not optional and there is no default value. Fixed it !

packages/headless/src/api/search/search/search-response.ts
aprudhomme-coveo: I made it optional to match how the API behaves

PR #2962, opened 2023-06-05
fix: correct org id in react samples folded results list page

mrrajamanickam-coveo:
### Before:  
"Folded result list example" resulted in error
image
```
{
  "statusCode": 400,
  "message": "Unknown query pipeline: name or id \"Search\"",
  "type": "UnknownQueryPipelineOtherException",
...
```

### After:
Folded results are displayed
image

https://coveord.atlassian.net/browse/KIT-2525


y-lakhdar (approved):

packages/samples/atomic-react/package.json
```diff
@@ -29,7 +29,7 @@
   "scripts": {
     "build": "tsc --noEmit && tsc --module commonjs --noEmit",
     "dev": "npm run build:assets && npm run dev:webpack",
-    "dev:webpack": "webpack serve",
+    "dev:webpack": "nx run atomic:build && webpack serve",
```
mrrajamanickam-coveo, line 32: Webpack would error if run without running build

packages/samples/atomic-react/src/pages/FoldedResultListPage.tsx
```diff
@@ -58,7 +60,10 @@ function MyTemplate(result: FoldedResult) {
         )}
       </AtomicResultSectionBadges>
       <AtomicResultSectionVisual>
-        <img loading="lazy" src="https://picsum.photos/350" />
+        <AtomicResultImage
+          field="ec_images"
+          fallback="https://picsum.photos/350"
```
mrrajamanickam-coveo, line 65: Guess `ec_images` doesn't make sense for folded results. But this could demonstrate usage of `fallback`?

packages/samples/atomic-react/package.json
btaillon: I recommend not to use the NX CLI to run parent tasks in scripts, since it doesn't add them to the dependency graph.

Instead, you should either:
* Modify AtomicReact's `project.json` to tell it that "dev" is dependent on building the parent packages. For example:
	```json
	{
	  "name": "atomic-react",
	  "targets": {
	    "dev": {
	      "dependsOn": ["^build"],
	      "inputs": [],
	      "outputs": [],
	      "executor": "nx:run-script",
	      "options": {
	        "script": "dev"
	      }
	    }
	  }
	}
	```
* Modify `nx.json` to require all `dev` tasks to build their parent task first. At the moment, I think this is better?

btaillon (changes requested):

packages/samples/atomic-react/src/components/AtomicPageWrapper.tsx
```diff
@@ -56,22 +58,34 @@ export const AtomicPageWrapper: FunctionComponent<Props> = ({
   children,
   options = {},
 }) => {
+  const searchConfig = options.expression
+    ? {}
+    : {
+        pipeline: 'Search',
+        searchHub: 'MainSearch',
+      };
```
btaillon, line 66: This will be re-evaluated every time the component re-renders, even though it's generally only going to be used once. I recommend moving this expression inside the `useMemo` below.

packages/samples/atomic-react/src/pages/FoldedResultListPage.tsx
btaillon: Sure, good idea!

packages/samples/atomic-react/package.json
mrrajamanickam-coveo: Cool - good idea! But couple of things
* this would only work with `nx run` and not `npm run` (which is to be expected I guess, but still something that could be missed when looking at package.json scripts and running them)
* building all parent deps is cool - but in this case of atomic-react is builds beuno and headless too in addition to atomic which is what is needed. And there doesn't seem to be a cache / dependency change detection - the parent builds are run every time. Is this expected?

packages/samples/atomic-react/src/components/AtomicPageWrapper.tsx
mrrajamanickam-coveo: Makes sense, it would also go together with the existing options.

mrrajamanickam-coveo: Actually the source containing folded results errors when search config is specified
```
{
  "statusCode": 400,
  "message": "Unknown query pipeline: name or id \"Search\"",
  "type": "UnknownQueryPipelineOtherException",
```

packages/samples/atomic-react/package.json
btaillon: it's expected. I did want to eventually add caching to Bueno and Headless but the story about Nx was starting to take too long so I left them as technical debt tickets.

Personally, I'd prefer if we choose one of those two extremes, no middle ground:
* Only build/debug AtomicReact, don't try to build Atomic.
* Build the whole stack (using Nx dependencies).

packages/samples/atomic-react/src/components/AtomicPageWrapper.tsx
btaillon: Yeah, you need to set the pipeline to default (empty string).

btaillon (approved):

packages/samples/atomic-react/src/components/AtomicPageWrapper.tsx
mrrajamanickam-coveo: Updated props as discussed.

btaillon (approved):

mrrajamanickam-coveo:
@olamothe Can you please review the latest changes as the PR is blocked pending your approval.

PR #2964, opened 2023-06-07
COMSDZ-501: Atomic product listing result list

thudon-coveo:


packages/atomic/src/components/product-listing/atomic-product-listing/atomic-product-listing.tsx
```diff
@@ -0,0 +1,292 @@
...
+  //     return this.productListingState.products;
+  //   }
+
+  //   return this.productListingState.products.slice(
+  //     this.currentIndex,
+  //     this.currentIndex + this.numberOfRecommendationsPerPage
+  //   );
+  // }
+
+  private get productRecommendationsAsResuls(): Result[] {
```
thudon-coveo, line 228: Since the ProductListingEngine doesn't use the Result interface like the other engines I had to do this to use the common stuff. This is something I'd like to change in Headless in the future if possible. Would be easier if we were like everyone else.

packages/atomic/src/components/product-listing/atomic-product-listing-result-template/atomic-product-listing-result-template.tsx
```diff
@@ -0,0 +1,69 @@
+import {ResultTemplate, ResultTemplateCondition} from '@coveo/headless';
```
olamothe, line 1: Similar problem as described above with mixing imports from `@coveo/headless` which contains all the code for "classic" search interface (and so potentially polluting the `product-listing` use case with un-necessary code).

Take a look at what was done for `atomic-insight-result-template`

packages/atomic/src/components/product-listing/atomic-product-listing-interface/store.ts
```diff
@@ -22,66 +21,48 @@ export interface SortDropdownOption {
...
+    getUrl() {
+      return commonStore.get('url');
+    },
   };
 }
 
-export function makeDesktopQuery(mobileBreakpoint: string) {
-  return `only screen and (min-width: ${mobileBreakpoint})`;
-}
+// export function makeDesktopQuery(mobileBreakpoint: string) {
```
olamothe, line 66: Remove ?

packages/atomic/src/components/product-listing/atomic-product-listing-result-template/atomic-product-listing-result-template.tsx
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Note: Any `<script>` tags defined inside of a `<template>` element will not be executed when results are being rendered.
+ * @MapProp name: mustMatch;attr: must-match;docs: The field and values that define which result items the condition must be applied to. For example, a template with the following attribute only applies to result items whose `filetype` is `lithiummessage` or `YouTubePlaylist`: `must-match-filetype="lithiummessage,YouTubePlaylist"`;type: Record<string, string[]> ;default: {}
+ * @MapProp name: mustNotMatch;attr: must-not-match;docs: The field and values that define which result items the condition must not be applied to. For example, a template with the following attribute only applies to result items whose `filetype` is not `lithiummessage`: `must-not-match-filetype="lithiummessage";type: Record<string, string[]> ;default: {}
+ */
+@Component({
```
olamothe, line 16: For all components that you are creating as part of this whole task about product listing, I would suggest adding a JSDOC comment `@internal`. 

You should see this pattern used for `insight` use case/components.

This will mark the component still under development/not yet ready for production, and will allow you to merge in master continuously (and not block any release) while the work is still ongoing.

You could do breaking changes/completely rework all components and it would be fine.


The components will be "available" to use if you know they exists (ie: you could install Atomic from NPM/production CDN and use the components for a demo on your side or something similar), but they won't appear in our doc, they won't generate code for Angular or React packages, etc.

Then, as a work item, when everything is good to go in production and is stable, you could go and remove all the `@internal` tags in one PR.

packages/atomic/src/components/product-listing/atomic-product-listing-result/atomic-product-listing-result.tsx
```diff
@@ -0,0 +1,201 @@
...
+        <Host class={resultComponentClass}>
+          <div
+            class="result-root"
+            ref={(ref) => (this.resultRootRef = ref)}
+          ></div>
+        </Host>
+      );
+    }
+    return (
+      // deepcode ignore ReactSetInnerHtml: This is not React code
```
olamothe, line 166: This is no longer needed/we don't use Snyk deepcode anymore.

packages/atomic/src/components/product-listing/atomic-product-listing/atomic-product-listing.tsx
```diff
@@ -0,0 +1,292 @@
...
+  /**
+   * The non-localized label for the list of products.
+   */
+  @Prop({reflect: true}) public label?: string;
+
+  /**
+   * The [heading level](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) to use for the heading label, from 1 to 6.
+   */
+  @Prop({reflect: true}) public headingLevel = 0;
+  // @Watch('numberOfProductsPerPage')
```
olamothe, line 108: Is the intention to work on all the commented code in this PR in another PR ?

packages/atomic/src/components/product-listing/atomic-product-listing/atomic-product-listing.tsx
```diff
@@ -0,0 +1,292 @@
+import {Result} from '@coveo/headless';
```
olamothe, line 1: Same concern about mixing imports

packages/atomic/src/components/product-listing/atomic-product-listing/atomic-product-listing.tsx
olamothe: Why does it need to respect the `Result` interface coming from the classic search use case though ?

A product recommendation is not really equal to a standard search result 🤔

packages/atomic/src/components/product-listing/atomic-product-listing-result-template/atomic-product-listing-result-template.tsx
thudon-coveo: Sounds good, thanks

packages/atomic/src/components/product-listing/atomic-product-listing/atomic-product-listing.tsx
thudon-coveo: Forgot to remove it sorry.

packages/atomic/src/components/product-listing/atomic-product-listing-interface/store.ts
thudon-coveo: I kept it in case we need it in the future but might as well remove it for now.

packages/atomic/src/components/product-listing/atomic-product-listing-interface/atomic-product-listing-interface.tsx
```diff
@@ -246,17 +255,13 @@ export class AtomicProductListingInterface
...
     }
 
-    if (!this.url) {
-      console.error(
-        'You need to set a url to fetch a product listing.',
-        this.url
-      );
-      return;
+    if (this.url) {
+      console.warn('url was manually set to :', this.url);
```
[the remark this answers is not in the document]
olamothe: Okay, it's useful when you are working on localhost, for example ?

And so it would still be useful for any developer trying to implement/develop their solution (not just for our own demo) ?

And those dev would need to do some changes before going in production.

I think we should beef up the warning then, be more verbose about all of this.

If you put yourself in the show of a random developer working on their Coveo project, I'm not sure that the message `url was manually set to XYZ` would help them understand what's the context.

packages/atomic/src/components/product-listing/atomic-product-listing/atomic-product-listing.tsx
olamothe: I see... I guess we can merge it like that then, but an alternative would be to rework `ResultListCommon` to be more lax on the types needed/more specific on the properties it needs, versus requiring a "full" result with all the properties.

packages/atomic/src/components/product-listing/atomic-product-listing-interface/atomic-product-listing-interface.tsx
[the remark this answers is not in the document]
thudon-coveo: That's exactly the use cases where it would be needed.

I'll write a more meaningful warning.

olamothe (approved):

olamothe:
Closing those PR as there was a focus shift regarding those component in the commerce unit.

Re-open if necessary.

PR #2970, opened 2023-06-08
feat(atomic): added folding-result-list for insight panel

nathanlb:
Added internal insight panel folded result list component which includes refactor of search page folding components.

Screen Shot 2023-06-07 at 2 40 14 PM


lbergeron (approved):
Great job! It looks all good to me!

packages/atomic/src/components/insight/result-templates/atomic-insight-result-children/atomic-insight-result-children.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+    if (!content) {
+      return;
+    }
+
+    const key =
+      child.result.uniqueId +
+      child.children.map((child) => child.result.uniqueId);
+    return (
+      <atomic-result
```
btaillon, line 159: Should this be `atomic-insight-result` instead?

btaillon (approved):

PR #2971, opened 2023-06-08
docs(storybook): fix controls for array properties 

olamothe:
https://coveord.atlassian.net/browse/KIT-2540

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,8 @@
     "node_modules",
     "stencil-generated",
     "dist",
-    "www"
+    "www",
+    "!.storybook"
   ],
   "env": {
     "jest": true,
@@ -16,13 +17,9 @@
   },
   "overrides": [
     {
-      "files": [
-        "**/package-lock.json"
-      ],
+      "files": ["**/package-lock.json"],
       "parser": "eslint-plugin-json-es",
-      "plugins": [
-        "package-lock"
-      ],
+      "plugins": ["package-lock"],
       "rules": {
         "package-lock/lock-file-version": [
           "error",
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -3,7 +3,8 @@
     "node_modules",
     "stencil-generated",
     "dist",
-    "www"
+    "www",
+    "!.storybook"
```
olamothe, line 7: This was causing linting to fail (because dot folder are ignored by default)

utils/atomic-storybook/.storybook/map-props-to-args.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable node/no-extraneous-import */
```
olamothe, line 1: Atomic/bueno are not installed as direct dependency in storybook, they are copied at build time (if I remember that part right).

Hence why we need to disable the tool here

utils/atomic-storybook/.storybook/map-props-to-args.ts
```diff
@@ -47,12 +48,14 @@ export const mapPropsToArgTypes = (componentTag: string): ArgTypes => {
...
     )
     .forEach((prop) => {
+      const {type, options} = determineControlTypeFromJsonValues(prop.values);
       ret[prop.name] = {
         description: `<pre>${prop.docs}</pre>`,
         table: {
           defaultValue: {summary: prop.default},
         },
-        control: determineControlTypeFromJsonValues(prop.values),
+        options,
```
olamothe, line 57: The structure changed in Storybook v7

utils/atomic-storybook/.storybook/map-props-to-args.ts
btaillon: ah, actually, I think this is a leftover from back when I tested adding `@coveo/atomic` as a dependency. I may have had to remove it to avoid npm hoisting dependencies, but you might be able to re-add `@coveo/atomic` as a dependency to `@coveo/atomic-storybook`.

btaillon (approved):

PR #2976, opened 2023-06-09
feat(headless): modify facet-manager controller to support automatic facets

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2524

### Changes
- Change the automaticFacets type to FacetResponse since I make my own facetId when receiving the facets anyway.
- Added a state containing an array of FacetResponse[ ] inside of the facet-manager controller
- Added a reducer for the action setDesiredCount (if we ever add the feature into the core facet manager we could simply load the action in the `loadFacetManagerReducers` 
- Modified the facet-set files to add automatic facets in the facet-set 
- Modified the search-request to send the right facets where they are supposed to go depending on the `generated` boolean value

- [x] add tests


packages/headless/src/features/facets/automatic-facets/automatic-facets-actions-loader.ts
```diff
@@ -0,0 +1,33 @@
+import {PayloadAction} from '@reduxjs/toolkit';
+import {SearchEngine} from '../../../app/search-engine/search-engine';
+import {setDesiredCount} from './automatic-facets-actions';
+import {automaticFacetsReducer as automaticFacets} from './automatic-facets-slice';
+
+/**
+ * The action creators for managing automatic facets.
```
jpmarceau, line 7:
```suggestion
 * The action creators to manage automatic facets.
```

packages/headless/src/controllers/facet-manager/headless-facet-manager.ts
```diff
@@ -1,18 +1,74 @@
...
+export interface FacetManager extends CoreFacetManager {
+  /**
+   * The state of the FacetManager.
+   */
+  state: FacetManagerState;
+}
+
+export interface FacetManagerProps {
+  /**
+   * The desired count of automatic facets.
```
jpmarceau, line 31: Here and throughout, might I suggest `target`/`targetCount` rather than `desired`/`desiredCount`?
This could just be personal preference, but `desired` always feels a bit weird to me in a technical context 🤓 

I would also suggest `number`/`targetNumber` rather than `count`/`targetCount`. I think that `number` is more common in technical contexts because it's more formal.

packages/headless/src/features/facets/automatic-facets/automatic-facets-actions-loader.ts
```diff
@@ -0,0 +1,33 @@
...
+   * Sets the desired count of automatic facets.
+   *
+   * @param desiredCount - Desired count of automatic facets.
+   * @returns A dispatchable action.
+   */
+  setDesiredCount(desiredCount: number): PayloadAction<number>;
+}
+
+/**
+ * Loads the automatic facets actions and adds the reducer to the search engine.
```
jpmarceau, line 20:
```suggestion
 * Loads the automatic facet actions and adds the reducer to the search engine.
```
(if you're curious, this is because `facet` is here operating as an adjective for `actions`, and is therefore not conjugated)

packages/headless/src/controllers/facet-manager/headless-facet-manager.ts
```diff
@@ -1,18 +1,74 @@
...
+   * The default value is 0.
+   */
+  desiredCount: number;
+}
 
 /**
  * Creates a `FacetManager` instance.
  *
  * @param engine - The headless engine.
+ * @param props - The facet manager props
```
jpmarceau, line 41:
```suggestion
 * @param props - The facet manager props.
```

packages/headless/src/features/facets/automatic-facets/automatic-facets-actions-loader.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param engine - The headless search engine
+ * @returns An object holding the automatic facets action creators
```
jpmarceau, line 23:
```suggestion
 * @param engine - The headless search engine.
 * @returns An object holding the automatic facets action creators.
```

packages/headless/src/features/facets/automatic-facets/automatic-facets-state.ts
```diff
@@ -0,0 +1,13 @@
+export type AutomaticFacetsState = {
+  /**
+   * The desired count of facets.
+   * The default value is 0.
```
jpmarceau, line 4: I think:
```suggestion
   * 
   * @defaultValue 0
```

packages/headless/src/controllers/facet-manager/headless-facet-manager.ts
```diff
@@ -1,18 +1,74 @@
...
+      const modifiedAutomaticFacets = getAutomaticFacets()?.map((facet) => ({
+        ...facet,
+        facetId: `generated_${facet.field}`,
+      }));
+
+      return {
+        ...core.state,
+        automaticFacets: modifiedAutomaticFacets,
+      };
+    },
```
louis-bompart, line 72: I would create an util `buildAutomaticFaceManager` or something that do that., and the in
`buildFacetManager` return a `{...core, ...automatic}`

packages/headless/src/features/facets/facet-set/facet-set-selectors.ts
```diff
@@ -19,9 +19,15 @@ export const baseFacetResponseSelector = (
...
 
   if ('search' in state && state.search) {
-    return state.search.response.facets.find(
+    const fromFacetResponse = state.search.response.facets.find(
       (response) => response.facetId === id
     );
+    const fromGeneratedFacetResponse =
+      state.search.response.generateAutomaticFacets?.facets.find(
+        (response) => `generated_${response.field}` === id
+      ) as FacetResponse;
```
louis-bompart, line 28: Why do you need a cast here?

louis-bompart (commented):
Looks 👌 to me. I don't see a lot of new tests associated with the new stuff tho, only adaption of the existing one.
Is this normal?

packages/headless/src/controllers/facet-manager/headless-facet-manager.ts
olamothe: `desiredCount` matches the API parameter though. I'd prefer if we had the same verbiage for both

jpmarceau: Oh I see, looks good then 👍

jpmarceau (approved):

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -63,6 +63,21 @@ export const facetSetReducer = createReducer(
...
+        generatedKeys.forEach((id) => {
+          const field = id.replace(/^generated_/, '');
+          const request = buildFacetRequest({
+            facetId: id,
+            field,
+          });
+          request.generated = true;
+          state[id] = getFacetSetSliceInitialState(request);
+        });
+
```
aprudhomme-coveo, line 80: this is the fix for the url-manager. Basically creating facets slice for the generated facets contained in the url

packages/headless/src/features/facets/automatic-facets/automatic-facets-actions.ts
```diff
@@ -0,0 +1,10 @@
+import {NumberValue} from '@coveo/bueno';
+import {createAction} from '@reduxjs/toolkit';
+import {validatePayload} from '../../../utils/validate-payload';
+
+const desiredCountDefinition = new NumberValue({min: 0});
```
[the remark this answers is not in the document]
aprudhomme-coveo: You are right, I will put 1 instead and make sure the generateAutomaticFacet argument is not sent otherwise.

aprudhomme-coveo:
Hi @jpmarceau I have a question. Since this feature is only available to a few beta clients via the search-api, should I add in the desiredCount doc comments something like : "This is a beta only feature, GA is planned in september." ?

jpmarceau:
> Hi @jpmarceau I have a question. Since this feature is only available to a few beta clients via the search-api, should I add in the desiredCount doc comments something like : "This is a beta only feature, GA is planned in september." ?

Hello @aprudhomme-coveo , in my opinion, yes. I would only capitalize `September` (English convention).
Or you could mark the controllers as internal, to avoid generating docs at all, but that might be too much.
I feel like this is mostly a product decision. I'd be curious to know what @mdey1 thinks?

packages/headless/src/controllers/facet-manager/headless-facet-manager.ts
aprudhomme-coveo: The code you selected is not entirely for the `automatic` feature. There is also the `...core.state` that needs to be returned in the state. I think extracting this in a function would be confusing and hard since the `get` part is necessary for constantly updating the state. I still improved the code a bit for better clarity.

mdey1:
> > Hi @jpmarceau I have a question. Since this feature is only available to a few beta clients via the search-api, should I add in the desiredCount doc comments something like : "This is a beta only feature, GA is planned in september." ?
> 
> Hello @aprudhomme-coveo , in my opinion, yes. I would only capitalize `September` (English convention). Or you could mark the controllers as internal, to avoid generating docs at all, but that might be too much. I feel like this is mostly a product decision. I'd be curious to know what @mdey1 thinks?

From my understanding, we do not want any public documentation on this feature just yet. If it's documentation for within the library that's alright, but not for the public doc site (So no storybook, no mention of it, etc.). @ariouxcoveo is leading this initiative for the customers, so I will let her confirm as well :)

olamothe:
> > > Hi @jpmarceau I have a question. Since this feature is only available to a few beta clients via the search-api, should I add in the desiredCount doc comments something like : "This is a beta only feature, GA is planned in september." ?
> > 
> > 
> > Hello @aprudhomme-coveo , in my opinion, yes. I would only capitalize `September` (English convention). Or you could mark the controllers as internal, to avoid generating docs at all, but that might be too much. I feel like this is mostly a product decision. I'd be curious to know what @mdey1 thinks?
> 
> From my understanding, we do not want any public documentation on this feature just yet. If it's documentation for within the library that's alright, but not for the public doc site (So no storybook, no mention of it, etc.). @ariouxcoveo is leading this initiative for the customers, so I will let her confirm as well :)

We should mark all the JSDOC with @ beta tag

packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts
```diff
@@ -403,6 +415,9 @@ export function buildCoreFacet(
 
     get state() {
       const request = getRequest();
+      if (!request) {
+        return deletedFacetState;
+      }
```
aprudhomme-coveo, line 420: small patch to support when facet get deleted

louis-bompart: Do you have another example of 'static' state being used somewhere in Headless?

I'm wondering if it should be in the function or module scope 🤔

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -281,3 +317,18 @@ function restoreFacetValueToIdleState(
 ): FacetValueRequest {
   return {...facetValue, state: 'idle'};
 }
+
+function deleteOldAutomaticFacets(
```
louis-bompart, line 321: what does old means here? i.e. when does an automatic facets become old?

louis-bompart (commented):
LGTM!, A few 🤔 ❓

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
aprudhomme-coveo: For the upper comment about the `static` state. I decided to put the logic for this inside of the `facetRequestSelector` in the `facet-set-selector.ts` instead. Thank you for the comment.

packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts
aprudhomme-coveo: Instead of this, I decided to put the logic for this inside of the `facetRequestSelector` in the `facet-set-selector.ts` instead. Thank you for the comment.

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
aprudhomme-coveo: I could replace the name for `removeUnusedAutomaticFacets` or `cleanupAutomaticFacets` what do you prefer ?

PR #2979, opened 2023-06-12
fix(headless): handle no question answer in insight api client

nathanlb:
Adding same empty questionAnswer object to response to match what is expected by the smart snippets reducer


packages/headless/src/features/question-answering/question-answering-slice.ts
```diff
@@ -82,6 +82,9 @@ export const questionAnsweringReducer = createReducer(
         state.feedbackModalOpen = false;
       })
       .addCase(executeSearch.fulfilled, (state, action) => {
+        if (!action.payload.response.questionAnswer) {
```
olamothe, line 85: Possible that since it goes through a different endpoint (search-interface as opposed to directly SearchAPI) that's where the difference comes from ?

If it's true that `questionAnswer` is possibly undefined when coming back from the API, a better fix would be to change `search-response.ts` to mark it as possibly undefined ( `questionAnswer?:  QuestionAnswers`), and fix all the compilation problem that will arise from the type change.

nathanlb: I confirmed it is indeed possible that `questionAnswer` is undefined in a searchapi response. It's the case when no smart snippets model is associated to the pipeline used. I'll try your suggestion and see what those compilation issues are. Just want to make sure we aren't doing anything breaking.

nathanlb: If we previously marked that part of the state as always being defined and we make it optional that could be breaking no?

nathanlb: What do you think?

olamothe: With the new changes, this could be removed, right ?

olamothe (approved):

packages/headless/src/features/question-answering/question-answering-slice.ts
olamothe, line 85: WIth these new changes, this is no longer necessary I assume ?

PR #2980, opened 2023-06-12
fix(headless): wrong type in product listing interactive result

thudon-coveo:
https://coveord.atlassian.net/browse/KIT-2542


packages/headless/src/api/search/search/product-recommendation.ts
```diff
@@ -101,8 +101,8 @@ export interface ProductRecommendation {
 
 // Change this list when changing the fields exposed by `ProductRecommendation`
 export const ProductRecommendationDefaultFields: string[] = [
-  'uri',
-  'urihash',
+  'documentUri',
+  'documentUriHash',
```
thudon-coveo, line 105: This is unrelated to the problem but it was bugging me.

packages/headless/src/features/product-listing/product-listing-analytics.ts
```diff
@@ -18,15 +19,89 @@ export const logProductListing = (): ProductListingAction =>
...
+    documentAuthor: '',
+    documentPosition: resultIndex + 1,
+    documentTitle: productRec.ec_name || '',
+    documentUri: productRec.documentUri,
+    documentUriHash: productRec.documentUriHash,
+    documentUrl: productRec.clickUri,
+    rankingModifier: '',
+    sourceName: '',
+    queryPipeline: getPipelineInitialState(),
+  };
```
thudon-coveo, line 106: I wanted to reuse the `PartialDocumentInformation` but it leaves some blank properties. I'm open to either create a new interface for product recs or trying to make PartialDocumentInformation more generic.

olamothe: Either of those would be breaking changes I reckon. If my making `PartialDocumentInformation` more generic you mean removing properties from the interface.

At face value, I think the proper approach would be to create a new interface specifically for product recs. But this would have to be done in a major version, and tracked here: https://coveord.atlassian.net/browse/KIT-2157

As for now, I think the only thing that can be done in a backward compatible manner is to do what you've done here.

packages/headless/src/api/search/search/product-recommendation.ts
thudon-coveo: These fields were already returned as `documentUri` and `documentUriHash`.

I'll bring back the old ones and keep these ones then to not break anything.

packages/headless/src/features/product-listing/product-listing-analytics.ts
btaillon: I think creating a new interface would be a good idea.

packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -525,3 +531,22 @@ function findPositionWithUniqueId(
...
+export const validateProductRecommendationPayload = (
+  productRec: ProductRecommendation
+) =>
+  new Schema(productRecommendationPartialDefinition).validate(
+    partialProductRecommendationPayload(productRec)
+  );
+
+function partialProductRecommendationPayload(
+  productRec: ProductRecommendation
+): Partial<Result> {
```
btaillon, line 544: What happens when you directly validate `productRec` without this?

olamothe (approved):

packages/headless/src/features/product-listing/product-listing-analytics.ts
thudon-coveo: The new interface would be in coveo.analytics and will require bigger changes. I think I'd go with Oli for now and keep it as is.

PR #2981, opened 2023-06-12
feat(atomic): mass translation of atomic with gpt model 

olamothe:
Mass translate all entries in `locales.json` for Atomic, using GPT4.

I added the script I used in the repo, as well as the prompt. I took care to ask GPT4 not to retranslate strings that were already translated.

The script continually write in a `temporary.json` file as a way to "save" the progress if it fails while executing.
It's because GPT can sometimes fumble on an entry and you have to tweak the prompt to fix it.

This takes about 30 minutes to execute for our repo, with each entry taking couple seconds to execute.

I think we might have saved multiple weeks of work with this one 😎 

I'll also start a thread with legal to get the green flag.

https://coveord.atlassian.net/browse/KIT-2545


utils/translation-gpt/index.mjs
```diff
@@ -0,0 +1,120 @@
+import {readFileSync, writeFileSync} from 'fs';
+import {env} from 'process';
+
+const fileToTranslate = JSON.parse(
```
olamothe, line 4: For `.mjs` file, to import JSON you have to add `assert {type: 'json'}` at the end of the import line.

And the `assert` keyword was tripping up es-lint/prettier.
So I used a wonky import instead.

utils/translation-gpt/temporary.json
```diff
@@ -0,0 +1 @@
+{}
```
btaillon, line 1: should probably add this to `.gitignore`.

btaillon (commented):
I have some worries about merging this without validating that nothing particularly bad made it into these translations. Even if these translations had been done by a human translator, there's no context behind these entries besides their English source entries.

I'm curious what we'd get if we removed all french translations, re-generated them with ChatGPT, then looked at the diff.

olamothe:
> I have some worries about merging this without validating that nothing particularly bad made it into these translations. Even if these translations had been done by a human translator, there's no context behind these entries besides their English source entries.
> 
> I'm curious what we'd get if we removed all french translations, re-generated them with ChatGPT, then looked at the diff.

The process I've done to validate some of the strings (not all of them) is this:

Take the translated input, translate it back to english, and verify that it matches one to one with the original english string. If not, validate that it's "close enough": you can ask GPT to tell you if the meaning between two sentences are similar.

I can add a script to add that validation for every single string if you think that's worthwhile.

btaillon:
> > I have some worries about merging this without validating that nothing particularly bad made it into these translations. Even if these translations had been done by a human translator, there's no context behind these entries besides their English source entries.
> > I'm curious what we'd get if we removed all french translations, re-generated them with ChatGPT, then looked at the diff.
> 
> The process I've done to validate some of the strings (not all of them) is this:
> 
> Take the translated output, translate it back to english, and verify that it matches one to one with the original english string. If not, validate that it's "close enough": you can ask GPT to tell you if the meaning between two sentences is similar.
> 
> I can add a script to add that validation for every single string if you think that's worthwhile.

I think we should add a script, yes. Maybe write some doc on how to re-do this process in the future, since I think we'll have to do it a decent number of times as we add more strings.

btaillon (approved):

fbeaudoincoveo (approved):
I checked a few things in French, Italian, and German. Overall it doesn't seem too bad! Sometimes the translation is a bit awkward for longer strings but people would probably understand 😁 

I could go through the entire thing but it would probably take me several hours so I don't think it's worth it.

utils/translation-gpt/index.mjs
```diff
@@ -0,0 +1,120 @@
...
+    ', '
+  )} and to place the variable with braces correctly in the new string if it applies.
+  For each language code, answer in valid JSON format, with the the main key being ${mainKey} and each individual language code is a property of ${mainKey}. 
+  The value of each property of ${mainKey} should be the translated string.
+  The string to translate is: 
+  "${toTranslate}"`;
+
+async function main() {
+  for (const translationKey of Object.keys(fileToTranslate)) {
+    if (fileAlreadyTranslated[translationKey]) {
```
mrrajamanickam-coveo, line 51: What happens if we want to add (or delete) a language to the list of supported languages?

utils/translation-gpt/index.mjs
```diff
@@ -0,0 +1,120 @@
...
+        headers: {
+          'Content-Type': 'application/json',
+          'api-key': env.COVEO_AZURE_OPEN_AI_KEY,
+        },
+        body: JSON.stringify({
+          messages: [
+            {
+              role: 'user',
+              content: prompt(
+                translationKey,
```
mrrajamanickam-coveo, line 82: Can we use these vars in the definition as they are more readable

utils/translation-gpt/index.mjs
```diff
@@ -0,0 +1,120 @@
...
+        'Troubleshoot the issue, then restart to script to continue translation.'
+      );
+    }
+  }
+
+  writeFileSync(
+    '../../packages/atomic/src/locales.json',
+    readFileSync('temporary.json')
+  );
+  writeFileSync('temporary.json', JSON.stringify({}, null, 2));
```
mrrajamanickam-coveo, line 117: or remove the file ?

mrrajamanickam-coveo:
> I can add a script to add that validation for every single string if you think that's worthwhile.

We can add it as an option or even another script that goes through and validates all translations (including existing ones)

utils/translation-gpt/index.mjs
```diff
@@ -0,0 +1,120 @@
...
+  'sv',
+  'th',
+  'tr',
+  'zh',
+  'zh-cn',
+  'zh-tw',
+];
+
+const prompt = (mainKey, toTranslate, languages) =>
+  `I will give you a string to translate. The string can possible contain variables between braces similar to this: {{variable}}.
```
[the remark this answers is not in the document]
louis-bompart: Not sure about changing text in the prompt.
With GPT, I reckon every comma is part of the secret sauce?

utils/translation-gpt/index.mjs
olamothe: - Add: Rerun the script, it will ask GPT for the new language for each entry in `locales.json`
- Delete: Nothing

utils/translation-gpt/index.mjs
```diff
@@ -0,0 +1,120 @@
...
+  'zh',
+  'zh-cn',
+  'zh-tw',
+];
+
+const prompt = (mainKey, toTranslate, languages) =>
+  `I will give you a string to translate. The string can possible contain variables between braces similar to this: {{variable}}. 
+  Those variables are meant to represent dynamic values.
+  If there are no variable, then do a normal translation.
+  I want you to translate it in the language code ${languages.join(
```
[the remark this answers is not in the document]
olamothe: ISO-639 I think, but GPT understood perfectly what was the intention with the current verbiage.

PR #2983, opened 2023-06-14
fix(headless): add missing controllers for product listing facets

thudon-coveo:
https://coveord.atlassian.net/browse/KIT-2546

### DIFF
--- a/packages/headless/src/features/analytics/analytics-utils.ts
+++ b/packages/headless/src/features/analytics/analytics-utils.ts
@@ -534,19 +534,4 @@ function findPositionWithUniqueId(
 
 export const validateProductRecommendationPayload = (
   productRec: ProductRecommendation
-) =>
-  new Schema(productRecommendationPartialDefinition).validate(
-    partialProductRecommendationPayload(productRec)
-  );
-
-function partialProductRecommendationPayload(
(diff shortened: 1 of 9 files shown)


packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -156,6 +160,11 @@ export type ProductListingAction<
   StateNeededByProductListingAnalyticsProvider
 >;
 
+export type ProductListingClickAction = PreparableAnalyticsAction<
+  {analyticsType: AnalyticsType.Click},
+  StateNeededBySearchAnalyticsProvider
+>;
```
btaillon, line 166: What is this for? I would expect `ProductListingAction<AnalyticsType.Click>` to be sufficient.

Also, should `StateNeededBySearchAnalyticsProvider` be `StateNeededByProductListingAnalyticsProvider`?

thudon-coveo: Yes I made a mistake here and I forgot to use it. Thanks for noticing.

thudon-coveo: So I removed it and used what you suggested instead.

olamothe (approved):

PR #2986, opened 2023-06-14
feat(quantic): quantic recommendation interface component created

mmitiche:
[SFINT-5070](https://coveord.atlassian.net/browse/SFINT-5070)

- Quantic Recommendation Interface component created.
- This component will initialize  the Headless Recommendation engine.
- This component will also dispatch the Headless action that will fetch the recommendation.

#### Design Doc: 
[Design Doc - Recommendation Feature in Quantic ](https://docs.google.com/document/d/13DPgvszxnwvm6ZcZhTrUlmP2jFIhLQu5nELq-lHCAmI/edit?usp=sharing)

[SFINT-5070]: https://coveord.atlassian.net/browse/SFINT-5070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticRecommendationInterface/quanticRecommendationInterface.js
```diff
@@ -0,0 +1,102 @@
...
+      } else {
+        setInitializedCallback(this.initialize, this.engineId);
+      }
+    });
+  }
+
+  /**
+   * @param {RecommendationEngine} engine
+   */
+  initialize = (engine) => {
```
SimonMilord, line 92: shouldn't we set the engine here? 
`this.engine = engine;`

mmitiche: Why would we?
We usually store the engine value when we need to use it again later in other methods. This does not apply in this component.

packages/quantic/force-app/main/default/lwc/quanticRecommendationInterface/quanticRecommendationInterface.js
```diff
@@ -0,0 +1,102 @@
...
+  engineOptions;
+  /** @type {boolean} */
+  initialized = false;
+
+  connectedCallback() {
+    loadDependencies(this, HeadlessBundleNames.recommendation).then(() => {
+      if (!getHeadlessBindings(this.engineId)?.engine) {
+        getHeadlessConfiguration().then((data) => {
+          if (data) {
+            this.engineOptions = {
```
erocheleau, line 64: Have you found out @mmitiche how we would set the `originContext` value in the Usage Analytics events sent by this?

We should be setting it to something like `quanticRecommendations`

erocheleau (approved):
Looks good to me

packages/quantic/force-app/main/default/lwc/quanticRecommendationInterface/quanticRecommendationInterface.js
mmitiche: Not yet no, but I will investigate this 👌

PR #2987, opened 2023-06-15
chore: fix scope of cspell plugin only to the relevant section

mrrajamanickam-coveo:
* Turn spell checker plugin only in the required eslint config section and not globally for all files
* Currently eslint spell checker plugin is turned on in the global scope making it run on package-lock.json etc.
https://coveord.atlassian.net/browse/KIT-2550

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,7 @@
     "jest": true,
     "es6": true
   },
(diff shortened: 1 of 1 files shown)


aprudhomme-coveo (approved):
Nice, this fix will also not turn it off for the .html files right ?

PR #2995, opened 2023-06-21
feat(headless): added generated answer (Gen-Q&A) component

nathanlb:
## [SVCC-3016](https://coveord.atlassian.net/browse/SVCC-3016)

The information related to this PR is included in much greater detail in this [design doc](https://docs.google.com/document/d/1JfpL134w_Pyf3JVRYlfs7D5YbrBXLO8hx2S5Gx4aqx0/edit?usp=sharing).
Please leave comments and questions either in this PR or in the design doc.

The controller is marked as `@internal` as it is still in beta stage.

**Demo Video** 
of the Atomic component that will leverage this controller.

https://github.com/coveo/ui-kit/assets/16785453/f7540f3b-eb85-4894-877a-f6f4ba94d079

[SVCC-3016]: https://coveord.atlassian.net/browse/SVCC-3016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/package.json
```diff
@@ -48,12 +48,14 @@
...
     "@reduxjs/toolkit": "1.8.5",
+    "@types/event-source-polyfill": "^1.0.1",
     "@types/pino": "6.3.12",
     "@types/redux-mock-store": "1.0.3",
     "abab": "2.0.6",
-    "coveo.analytics": "2.28.5",
+    "coveo.analytics": "2.28.7",
     "cross-fetch": "3.1.5",
     "dayjs": "1.11.5",
+    "event-source-polyfill": "^1.0.31",
```
olamothe, line 58: How much thoughts/analysis have we put into checking how the polyfill works, it's size, who's the developer, when and why it's needed, etc.

Any dependency we add here has to be very solid.

It seems that https://caniuse.com/eventsource is supported by all browsers we care about ?
And so we'd only need it for Node JS (or do we ?)

olamothe, line 58: How much thoughts/analysis have we put into checking how the polyfill works, it's size, who's the developer, when and why it's needed, etc.

Any dependency we add here has to be very solid.

It seems that https://caniuse.com/eventsource is supported by all browsers we care about ?
And so we'd only need it for Node JS (or do we ?)

packages/headless/src/api/analytics/search-analytics.ts
```diff
@@ -75,6 +75,21 @@ export class SearchAnalyticsProvider
...
   }
 
+  public getBaseMetadata() {
+    const state = this.getState();
+    const baseObject = super.getBaseMetadata();
+
+    const generativeQuestionAnsweringId =
+      state.search?.response?.extendedResults?.generativeQuestionAnsweringId;
+
+    if (generativeQuestionAnsweringId) {
```
olamothe, line 85: This whole function seems like a hack because `coveo.analytics` does not accept the proper type ?

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,152 @@
...
+export interface AsyncThunkGeneratedAnswerOptions<
+  T extends Partial<SearchAppState>
+> extends AsyncThunkOptions<
+    T,
+    ClientThunkExtraArguments<SearchAPIClient, GeneratedAnswerAPIClient>
+  > {}
+
+const buildStreamingUrl = (url: string, orgId: string, streamId: string) =>
+  new URLPath(
+    `${url}/rest/organizations/${orgId}/machinelearning/streaming/${streamId}`
```
olamothe, line 32: Is that service built with the mindset that it will be used for "everything" ML streaming related ? 🤔 

My question is mostly around naming of this thing (`GeneratedAnswerAPIClient`) and the path ( generic `machinelearning/streaming` which does not really contain anything related to "generated answer for search pages").

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (!streamId) {
+      this.logger.error('No stream ID found');
+      return;
+    }
+
+    let retryCount = 0;
+    let timeout: ReturnType<typeof setTimeout>;
+    let source: EventSourcePolyfill;
+
+    const checkAndRetry = (e: unknown): EventSourcePolyfill | void => {
```
olamothe, line 64: I haven't gone into details of everything happening here, but I * think * all the retry logic could be replaced/improved with `exponential-backoffs` dependency which is already in used in the project.

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,152 @@
...
+      }
+    };
+
+    const stream = (): EventSourcePolyfill | void => {
+      try {
+        source = new EventSourcePolyfill(
+          buildStreamingUrl(url, organizationId, streamId).href,
+          {
+            headers: {
+              Authorization: `Bearer ${accessToken}`,
```
olamothe, line 101: The stream API has a way to know that a given access token has access to retrieve a particular stream ID ?

Do you know how that works ?

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -0,0 +1,164 @@
...
+  let lastRequestId: string;
+  let lastStreamId: string;
+
+  const onMessage = (payload: GeneratedAnswerMessagePayload) =>
+    dispatch(updateMessage(payload));
+
+  const onCitations = (payload: GeneratedAnswerCitationsPayload) =>
+    dispatch(updateCitations(payload));
+
+  const onError = (error: SSEErrorPayload) => {
```
olamothe, line 81: Most of what's happening here to check what's the status on the source, update state by dispatching actions, actually looks like logic that should be done in a reducer, and not a controller 🤔

packages/headless/package.json
nathanlb: This is indeed my least favorite part of this project. EventSource is supported by all browsers but the reason we need something else is that it is extremely simple and limited as it doesn't even allow us to specify an authorization header. The auth check is a requirement we have so already we need some type of polyfill that expands the base features.

The lib is the best I could find after some research. the class we import is this one https://github.com/Yaffle/EventSource/blob/master/src/eventsource.js which is 33kb. The library has a decent amount of usage and 30-ish contributors but I don't know what our standards are normally.

Writing our own version of an SSE client was off the table for this beta feature but could probably be done later if that's what we need.

What do you think?

packages/headless/src/api/analytics/search-analytics.ts
nathanlb: This is fair 😅. I went with the way we add the headless version in the payload but it is indeed hacky. I'll see about adding it in coveo.analytics.

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: it could be used for other types of streaming in the future but it is unlikely that the specs would be maintained so pretty much everything in this client is specific to "generated answer for search pages".

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
nathanlb: I agree though the source object is not serializable AFAIK so I couldn't put it in the state. So the controller is where I chose to keep its reference but if you can think of a better place I'd like to hear it.

Also the EventSource object needs to be given the event handlers which do the action dispatching. They're defined here mainly because the controller is the object that has both access to the source and the dispatch function. I could alternatively pass the dispatch function to the client when starting the streaming to that it can do some dispatching. Do you think that's better?

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: The stream API does indeed check access. For now it only checks that the user has query_execute rights but in a few weeks it will also make sure that the user reading the stream is the same one that generated it.

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: I'll try it out!

nathanlb: Hmmm i tried replacing it but what i can't seem to replicate is that this mechanism is intended to detect when it's been too long between message events. The backOff function expects the promise function that can be rejected which doesn't happen with this SSE stream.

packages/headless/src/api/analytics/search-analytics.ts
nathanlb: I was asked to set it as custom event data here and it's supposed to be temporary until we have the new UA protocol. I could add a getter to the provider given to the analytics client and set it in there but is it really worth it? For now headless is the only lib that will use this.

packages/headless/src/api/generated-answer/generated-answer-event-payload.ts
```diff
@@ -0,0 +1,33 @@
+export enum GeneratedAnswerStreamFinishReason {
```
olamothe, line 1: A bit nitpicky, but I generally find string enum un-necessary in TypeScript when you can simply use string literals.

ie:
```typescript
export type GeneratedAnswerStreamFinishReason = 'COMPLETED' | 'ERROR';
```

Similar comment for `GeneratedAnswerPayloadType`

packages/headless/package.json
olamothe: Is this something we should discuss with the team in charge of the API then, and arrange for a different way to negotiate authentication ? Possibly using a `auth` event to send the token ?

Clients not using Headless will face the same problem, and so we're forcing a polyfill on everyone, because it does not follow "normal web standards for event source".

Is @AndyCoveo team's in charge of this service ? What is your opinion on this one ?

packages/headless/src/api/analytics/search-analytics.ts
olamothe: "Supposed to be temporary" 😉 

Alright then 😄

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
olamothe: Or could this be done in the asyncThunk (streamAnswer) ?
You can dispatch action from inside an asyncThunk.

packages/headless/package.json
olamothe: Could also look at full blown web socket 😁

olamothe: I had a live discussion with Andy.
I'll talk with security also.

What it boils down to is that I find it weird that we'll have in production an API that is essentially not usable with the standard as it is defined on MDN/what browser implements.

Which polyfill we end up using if that's really the only course of action is a secondary concern.

packages/headless/src/api/analytics/search-analytics.ts
nathanlb: Our UA plans are the thing i'm able to be the least certain about so I don't want to risk being too confident 😅

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
nathanlb: If i can share the reference to `source` between the asyncthunk and the controller this may indeed work and would clean up the controller. I'll give it a try

nathanlb: I tried it and it is much better! thanks for the suggestion

packages/headless/package.json
louis-bompart: I'm strongly against including the polyfill. Details in Slack.

packages/headless/src/api/generated-answer/generated-answer-event-payload.ts
olamothe: >  i'm just usually weary of typing constant strings out like this.

vs code auto complete will work no problem 😂

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -0,0 +1,132 @@
...
+    source &&
+    (source.readyState === source.OPEN ||
+      source.readyState === source.CONNECTING);
+
+  const subscribeToSearchRequests = () => {
+    const strictListener = () => {
+      const state = getState();
+      const requestId = state.search.requestId;
+      const streamId =
+        state.search.extendedResults.generativeQuestionAnsweringId;
```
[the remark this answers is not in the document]
erocheleau: I vote to drop the "Question" part of the name to make it shorter :)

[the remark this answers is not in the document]
erocheleau: Actually just noticed too... but the SmartSnippet feature is actually called `Question Answering` in the code, so to avoid confusion, I would definitely stick with `Generated Answer`

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -0,0 +1,132 @@
...
+    get state() {
+      return getState().generatedAnswer;
+    },
+
+    retry() {
+      dispatch(executeSearch(logRetryGeneratedAnswer()));
+    },
+
+    likeGeneratedAnswer() {
+      dispatch(logLikeGeneratedAnswer());
```
erocheleau, line 114: There's no state modification other than analytics events on like/dislike ?

Once again, looks very similar to smart snippets, all that's done is to keep in the state if it's liked/disliked so the thumbs up/down can display a state if they are liked/disliked already by the current user/

image

I'm assuming we're not to that point yet :)

packages/headless/src/features/generated-answer/generated-answer-actions.ts
```diff
@@ -0,0 +1,125 @@
...
+const booleanValue = new BooleanValue({required: true});
+const citationSchema = {
+  id: stringValue,
+  title: stringValue,
+  uri: stringValue,
+  permanentid: stringValue,
+  clickUri: optionalStringValue,
+};
+
+export interface SSEErrorPayload {
```
erocheleau, line 38: What's a SSE? :D

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -0,0 +1,60 @@
...
+    'generatedAnswer/logOpenGeneratedAnswerSource',
+    AnalyticsType.Custom,
+    (client, state) => {
+      const generativeQuestionAnsweringId =
+        state.search?.response?.extendedResults
+          ?.generativeQuestionAnsweringId ?? '';
+      return client.makeOpenGeneratedAnswerSource({
+        generativeQuestionAnsweringId,
+        permanentId: citation.permanentid,
+        id: citation.id,
```
erocheleau, line 29: What's the difference between the two Ids?

Each citation will have an individual ID even if the permanentID of the document behind it is the same?

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -0,0 +1,60 @@
...
+      });
+    }
+  );
+
+export const logLikeGeneratedAnswer = (): CustomAction =>
+  makeAnalyticsAction(
+    'generatedAnswer/logLikeGeneratedAnswer',
```
erocheleau, line 36: Looks like all other analytics actions have this format: `'analytics/generatedAnswer/likeGeneratedAnswer'` 

Don't know if it matters a lot to be consistent or not in this

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -0,0 +1,60 @@
...
+          ?.generativeQuestionAnsweringId ?? '';
+      return client.makeLikeGeneratedAnswer({
+        generativeQuestionAnsweringId,
+      });
+    }
+  );
+
+export const logDislikeGeneratedAnswer = (): CustomAction =>
+  makeAnalyticsAction(
+    'generatedAnswer/logDislikeGeneratedAnswer',
```
erocheleau, line 50:
```suggestion
    'analytics/generatedAnswer/dislike',
```
Just a suggestion

packages/headless/src/features/generated-answer/generated-answer-state.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface GeneratedAnswerState {
+  isLoading: boolean;
+  answer?: string;
+  citations: GeneratedAnswerCitation[];
+  error?: {
+    message?: string;
+    code?: number;
+    isRetryable?: boolean;
+  };
+}
```
erocheleau, line 12: Assuming later we could keep a state here of 

```
liked: boolean;
disliked: boolean;
```

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    const handleStreamPayload = (
+      payloadType: GeneratedAnswerPayloadType,
+      payload: string
+    ) => {
+      if (payloadType === 'genqa.citationsType') {
+        onMessage(JSON.parse(payload) as GeneratedAnswerMessagePayload);
+      } else if (payloadType === 'genqa.messageType') {
+        onCitations(JSON.parse(payload) as GeneratedAnswerCitationsPayload);
+      }
```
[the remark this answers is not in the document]
nathanlb: yeah my bad i just changed that from another comments suggestion and didnt swap the right ones. The client is still missing tests...

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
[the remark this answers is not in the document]
nathanlb: I agree and it's `generatedAnswer` everywhere except in the param the search api returns us that contains the stream ID. That one is called `generativeQuestionAnsweringId` and it's the only place.

packages/headless/src/features/generated-answer/generated-answer-actions.ts
nathanlb: Server Sent Event. I could call it `GeneratedAnswerStreamErrorPayload` instead

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
nathanlb: Yeah the citation id appears to be something to identify the citation used by ML. Given it's a different instance than the document itself I think it makes sense to keep both.

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
nathanlb: Consistent is better

packages/headless/src/features/generated-answer/generated-answer-state.ts
nathanlb: yeah i'll add that now

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,149 @@
...
+          onError({
+            message: errorMessage,
+            code: RETRYABLE_STREAM_ERROR_CODE,
+          });
+        };
+
+        return source;
+      } catch (e) {
+        return checkAndRetry(e);
+      }
```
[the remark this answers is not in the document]
nathanlb: Hmmm good point. If we can't parse the response because the JSON is malformed then it's an error with our backend. I think we should terminate the stream set an error message in the state. I don't think retrying would yield a proper JSON magically. What do you think?

packages/headless/src/features/generated-answer/generated-answer-slice.ts
```diff
@@ -0,0 +1,42 @@
...
+  updateError,
+  updateMessage,
+} from './generated-answer-actions';
+import {getGeneratedAnswerInitialState} from './generated-answer-state';
+
+export const generatedAnswerReducer = createReducer(
+  getGeneratedAnswerInitialState(),
+  (builder) =>
+    builder
+      .addCase(updateMessage, (state, {payload}) => {
```
lbergeron, line 19: Is it possible to have the `updateMessage` or `updateCitations` actions called after `updateError`? If yes, we may want to clear `state.error`.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -0,0 +1,132 @@
...
+    },
+
+    dislikeGeneratedAnswer() {
+      dispatch(logDislikeGeneratedAnswer());
+    },
+
+    logCitationClick(citation: GeneratedAnswerCitation) {
+      dispatch(logOpenGeneratedAnswerSource(citation));
```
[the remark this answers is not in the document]
nathanlb: Hmm yeah i guess if it's already an established pattern I can do the same. If the clicked citation should be in the state and I guess this more easily prevents citation data from being tampered with given it stays close to the headless core.

packages/headless/src/features/generated-answer/generated-answer-slice.ts
nathanlb: We could do it for safety. But to answer your question, no currently when the onError is called we also terminate the stream. But there's nothing forcing those two actions to be coupled it's just what the flow is now.

packages/headless/package.json
nathanlb: Replaced the polyfill with https://www.npmjs.com/package/@microsoft/fetch-event-source

packages/headless/src/features/generated-answer/generated-answer-slice.ts
nathanlb: I reset the error state in those reducers 👍

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
erocheleau: All good was just making sure it was clear

packages/headless/src/api/generated-answer/generated-answer-client.ts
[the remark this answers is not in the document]
nathanlb: So, using the new eventsource library the retry logic is different and errors are automatically retried unless fatal up to a max of 3 as prescribed. In this case, parsing an invalid json would throw an error and cause a retry.
I like your gut feeling but what i implemented here is kind of the opposite. We know what not to retry and automatically retry the rest a maximum of 3 times. Based on the information the backend team could give me about the possible error scenarios this seemed the most efficient. javascript errors like parsing would get swept up in it and I think error handling is something we can improve if problematic.

erocheleau (approved):
As far as my limited knowledge of EventSource goes, this looks fine now :) 

Huge feature, GG

packages/headless/src/api/generated-answer/generated-answer-event-payload.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface GeneratedAnswerCitation {
+  id: string;
+  title: string;
+  uri: string;
+  permanentid: string;
+  clickUri?: string;
+}
+
+export interface GeneratedAnswerMessagePayload {
+  textDelta: string;
```
btaillon, line 16: Is this always text or can it be HTML?

If it can be HTML, then I would replace `text` with `html`.

nathanlb: you mean you would call it `htmlDelta` instead? This is defined by the schema returned by the API so the ML team and I settled on textDelta a few weeks ago. So far it is text. Could it potentially be rich text down the road with some markup tags? possibly. But we can't really change this in the backend at this point in time.

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,166 @@
...
+    ) => {
+      if (payloadType === 'genqa.messageType') {
+        actions.updateMessage(
+          JSON.parse(payload) as GeneratedAnswerMessagePayload
+        );
+      } else if (payloadType === 'genqa.citationsType') {
+        actions.updateCitations(
+          JSON.parse(payload) as GeneratedAnswerCitationsPayload
+        );
+      }
```
mrrajamanickam-coveo, line 93: Is it worth adding a warn/error for unknown payload types?

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,166 @@
...
+          if (data.finishReason === 'ERROR') {
+            clearTimeout(timeout);
+            actions.updateError({
+              message: data.errorMessage,
+              code: data.errorCode,
+            });
+            abortController.abort();
+            return;
+          }
+          if (data.payload && data.payloadType) {
```
mrrajamanickam-coveo, line 134: Would the finishReason be empty or some other value for this case?

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  const subscribeToSearchRequests = () => {
+    const strictListener = () => {
+      const state = getState();
+      const requestId = state.search.requestId;
+      const streamId =
+        state.search.extendedResults.generativeQuestionAnsweringId;
+
+      if (lastRequestId !== requestId) {
```
mrrajamanickam-coveo, line 86: When would this happen? In case of retry? A comment might be helpful.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  return {
+    ...controller,
+
+    get state() {
+      return getState().generatedAnswer;
+    },
+
+    retry() {
+      dispatch(executeSearch(logRetryGeneratedAnswer()));
```
mrrajamanickam-coveo, line 115: Are we retriggering the entire search for retry? If yes would it be possible to distinguish this in Analytics and guessing it is not possible to just retry gen-qa without retriggering the search?

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  it('should subscribe to state updates', () => {
+    expect(engine.subscribe).toHaveBeenCalledTimes(1);
+  });
+
+  it('#retry dispatches #executeSearch', () => {
+    generatedAnswer.retry();
+    const action = engine.findAsyncAction(executeSearch.pending);
+
+    expect(action).toBeTruthy();
```
mrrajamanickam-coveo, line 60: can we assert falsy before retry()?

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -0,0 +1,70 @@
...
+    'analytics/generatedAnswer/retry',
+    AnalyticsType.Search,
+    (client) => client.makeRetryGeneratedAnswer()
+  );
+
+export const logOpenGeneratedAnswerSource = (
+  citationId: string
+): CustomAction =>
+  makeAnalyticsAction(
+    'analytics/generatedAnswer/openSource',
```
mrrajamanickam-coveo, line 23: nit: openSource -> clickSource to avoid confusion :) ?

packages/headless/src/features/generated-answer/generated-answer-slice.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    const finalState = generatedAnswerReducer(baseState, likeGeneratedAnswer());
+
+    expect(finalState).toEqual({
+      ...getGeneratedAnswerInitialState(),
+      liked: true,
+      disliked: false,
+    });
+  });
+
+  it('#dislikeGeneratedAnswer should set the answer as liked in the state', () => {
```
mrrajamanickam-coveo, line 197:
```suggestion
  it('#dislikeGeneratedAnswer should set the answer as disliked in the state', () => {
```

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,166 @@
...
+class RetryableError extends Error {}
+class FatalError extends Error {}
+
+interface DispatchActions {
+  updateMessage: (payload: GeneratedAnswerMessagePayload) => void;
+  updateCitations: (payload: GeneratedAnswerCitationsPayload) => void;
+  updateError: (payload: GeneratedAnswerErrorPayload) => void;
+  setIsLoading: (isLoading: boolean) => void;
+  resetAnswer: () => void;
+}
```
btaillon, line 48: Instead of taking 5 actions, I think we should:
1. Just like the search API client, not use redux terminology (e.g., actions).
2. Take similar callbacks to a [`WriteableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream#examples) (as if `streamGeneratedAnswer` was the equivalent of calling `someStream.pipeTo`).
  * write(chunk)
  * close()
  * abort()

Ideally, I would've preferred to use the `ReadableStream` and `WriteableStream` directly, but they aren't supported by Node 12, which our docs say we support.

packages/headless/src/api/generated-answer/generated-answer-event-payload.ts
btaillon: alright then, just gotta make sure that the API never returns HTML in this.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
nathanlb: AFAIK this listener will run on every state update of which there are many that could not involve any change in the `requestId`. My condition here is just to execute this code when a new request has been made and processed by redux.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
nathanlb: That's exactly what the `logRetryGeneratedAnswer` function does. The analytics event logged with this retry will specify an `actionCause` indicating that this is a retry for gen-qa.
Unfortunately it is not possible to try regenerating a stream without re-executing a query to the search API like this.

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
nathanlb: I hadn't even thought that just 'openSource' could lead to confusion. 😅

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
nathanlb: sure thing 👍

nathanlb: But I don't think we do that in any other tests asserting that an action can be found

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -0,0 +1,166 @@
...
+            response.headers.get('content-type') === EVENT_STREAM_CONTENT_TYPE
+          ) {
+            return;
+          }
+          const isClientSideError =
+            response.status >= 400 &&
+            response.status < 500 &&
+            response.status !== 429;
+          if (isClientSideError) {
+            throw new FatalError();
```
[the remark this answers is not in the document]
nathanlb: indeed

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: I don't see any harm in adding it

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: I can change the terminology for sure.
I'm not sure I understand how you envision the client working like a writeable stream. `fetchEventSource` already works in a way that resembles it. but there's a lot of stuff happening around it that IMO doesn't make sense to handle in the redux action over right here in the client.

packages/headless/src/api/generated-answer/generated-answer-event-payload.ts
nathanlb: If there is html it will be stringified for sure.

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: Maybe i'm just not understanding your suggestion.

nathanlb: I agree it's odd to pass in essentially redux actions but it's hard to find a context that has both access to redux and the mechanics of the stream. I could move most of the logic out of here and into the action but then the client doesn't really have much reason for existing as I think it'll basically be empty. Here it at least encapsulated the streaming complexity away from redux definitions.

packages/headless/src/api/generated-answer/generated-answer-client.ts
mrrajamanickam-coveo: Would it be more readable then to switch on finishReason ?

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
mrrajamanickam-coveo: Yes I noticed the other tests and thought the same. Since these are unit tests with state reset before each test, guess it is fine to leave them as they are.

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: I personally prefer this as it is flatter and we can `return` and exit the callback.

mrrajamanickam-coveo: How about using `else if` to make the flow explicit and that the blocks are related ?

mrrajamanickam-coveo (approved):

packages/headless/src/api/generated-answer/generated-answer-client.ts
nathanlb: Sure if you prefer it i can do that for you. Thanks for your review!

louis-bompart (approved):

PR #2996, opened 2023-06-21
ci: auto release every wednesday

louis-bompart:
image
We'll see where it goes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,8 @@
 name: Create release
 on:
+  schedule:
+    # At 13:00 (UTC) every Wednesday.
+    - cron:  '0 13 * * 3'
   workflow_dispatch:
     inputs:
       version:


btaillon:
What's the benefit of this?

louis-bompart:
> What's the benefit of this?

I don't have to think about pushing a button.

.github/workflows/release.yml
```diff
@@ -1,5 +1,8 @@
 name: Create release
 on:
+  schedule:
```
btaillon, line 3: Potential issues:
* The `release` environment requires approval, so this might essentially do nothing.
* We don't run tests before releasing, so this could deploy something broken. I wrote the release workflow assuming that we want to release "no matter what" (e.g., some tests failed, and we acknowledge they're just flaky or some API is down).

btaillon (approved):
If the release requires approval and this is purely to get notified to release, then I'm fine with it.

.github/workflows/release.yml
louis-bompart: As you say, it'll 'block' at the env approval stage, which is the goal: It'll do an auto message in the channel. Maintenance guy can then do a quick check: is it green, nothing blocking, ok, just press the :shipit: button.

louis-bompart:
> If the release requires approval and this is purely to get notified to release, then I'm fine with it.

And it saves a click! 🤑 /s

olamothe (approved):
Alright sure, lets see how it goes !

PR #2997, opened 2023-06-21
feat(headless): Add new insight analytics actions

erocheleau:
[SFINT-5082]

logCaseSendEmail, logFeedItemTextPost 

Will have to wait until this PR has been released to bump the version of coveo.analytics.js used in ui-kit

https://github.com/coveo/coveo.analytics.js/pull/425

(These two click actionCause have also already been added to the UA valid event schema repo)

[SFINT-5082]: https://coveord.atlassian.net/browse/SFINT-5082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/features/analytics/insight-analytics-actions-loader.ts
+++ b/packages/headless/src/features/analytics/insight-analytics-actions-loader.ts
@@ -27,7 +27,11 @@ import {
   logPagePrevious,
 } from '../pagination/pagination-insight-analytics-actions';
 import {SmartSnippetFeedback} from '../question-answering/question-answering-analytics-actions';
-import {logCopyToClipboard} from '../result-actions/result-actions-insight-analytics-actions';
+import {
+  logCopyToClipboard,
+  logCaseSendEmail,
+  logFeedItemTextPost,
+} from '../result-actions/result-actions-insight-analytics-actions';
 import {logResultsSort} from '../sort-criteria/sort-criteria-insight-analytics-actions';
 import {
   LogStaticFilterToggleValueActionCreatorPayload,
@@ -215,6 +219,22 @@ export interface InsightAnalyticsActionCreators {
    * @returns A dispatchable action.
    */
   logCopyToClipboard(result: Result): InsightAction<AnalyticsType.Click>;
+
+  /**
+   * The event to log when the Case Send As Email result action is clicked.
+   *
+   * @param result - The result.
+   * @returns A dispatchable action.
+   */
+  logCaseSendEmail(result: Result): InsightAction<AnalyticsType.Click>;
+
+  /**
+   * The event to log when the Feed Item Text Post result action is clicked.
+   *
+   * @param result - The result.
+   * @returns A dispatchable action.
+   */
+  logFeedItemTextPost(result: Result): InsightAction<AnalyticsType.Click>;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


erocheleau:
coveo.analytics has been released and updated in this PR. It's now ready for review.

packages/headless/src/features/analytics/insight-analytics-actions-loader.ts
```diff
@@ -215,6 +219,22 @@ export interface InsightAnalyticsActionCreators {
...
+   */
+  logCaseSendEmail(result: Result): InsightAction<AnalyticsType.Click>;
+
+  /**
+   * The event to log when the Feed Item Text Post result action is clicked.
+   *
+   * @param result - The result.
+   * @returns A dispatchable action.
+   */
+  logFeedItemTextPost(result: Result): InsightAction<AnalyticsType.Click>;
```
mmitiche, line 237: Isn't the name a bit weird ? x) `logPostToFeed` would be better I think.
I know it's named like that because  that's the name of the event in `coveo.analytics` but we need to break this loop of weird names at some point.
 
Same remark for first action, maybe simply `logSendAsEmail`.

mmitiche (approved):
LGTM!
Just a small remark about the names of the actions. I personally prefer to name the dispatch-able actions in a more simple and meaningful way to always keep the library easy to use and break the loop of, maybe, badly named actions in the past.

packages/headless/src/features/analytics/insight-analytics-actions-loader.ts
dmbrooke: I'm not sure if I agree with this. If the purpose of this function is to log an event triggered by a Feed Item Text Post result action, then `logFeedItemTextPost` leaves very little to the imagination as to what it could do...

I don't feel _**super**_ strongly about this, but I'm curious to hear everyone's thoughts 🙂

dmbrooke (approved):
Looks good!

packages/headless/src/features/analytics/insight-analytics-actions-loader.ts
erocheleau: Yea exactly, in my mind, I want to make it as simple as possible to figure out the full loop of this.

The analytics `actionCause` is `FeedItem.TextPost` for example or `Case.SendEmail`. Those we can't change to stay compatible with existing Insight Panels.

It's also the name of the method in the coveo.analytics insight client `logFeedItemTextPost()` and `logCaseSendEmail()`.

It's also the API name of the action that's triggered in Salesforce `Case.SendEmail` and `FeedItem.TextPost`.

So if you `ctrl+f` in the codebase you're going to find them easily.

Whereas `logPostToFeed` is an arbitrary name we had given to the "button" only. That name we can change if we want to, to become more similar to FeedItem.TextPost, but it's pretty ugly haha. So the only place where that will appear is in the SFINT code `resultPostToFeed` and `resultSendAsEmail`. All the rest is consistent and based on a name we cannot change in the analytics event cause.

mmitiche: Agreed then 👌

PR #2998, opened 2023-06-21
feat(headless): support date range facet exclusion

dmbrooke:


packages/headless/src/features/facets/range-facets/generic/range-facet-utils.ts
```diff
@@ -9,6 +10,10 @@ export const isRangeFacetValueSelected = (selection: RangeFacetValue) => {
   return selection.state === 'selected';
 };
 
+export const isRangeFacetValueExcluded = (selection: RangeFacetValue) => {
```
y-lakhdar, line 13: I think you forgot to add this method to the `headless-core-range-facet`

dmbrooke: Whoops,  good catch! I'll add that now ✅

packages/headless/src/features/facets/range-facets/date-facet-set/date-facet-selectors.ts
```diff
@@ -35,3 +35,14 @@ export const dateFacetSelectedValuesSelector = (
...
 };
+
+export const dateFacetExcludedValuesSelector = (
+  state: SearchSection & DateFacetSection,
+  facetId: string
+): DateFacetValue[] => {
+  const facetResponse = dateFacetResponseSelector(state, facetId);
+  if (!facetResponse) {
+    return [];
+  }
```
louis-bompart, line 46:
```suggestion
  const facetResponse = dateFacetResponseSelector(state, facetId) || [];
```

packages/headless/src/features/facets/range-facets/numeric-facet-set/numeric-facet-selectors.ts
```diff
@@ -35,3 +35,14 @@ export const numericFacetSelectedValuesSelector = (
...
 };
+
+export const numericFacetExcludedValuesSelector = (
+  state: SearchSection & NumericFacetSection,
+  facetId: string
+): NumericFacetValue[] => {
+  const facetResponse = numericFacetResponseSelector(state, facetId);
+  if (!facetResponse) {
+    return [];
+  }
```
louis-bompart, line 46: same as above `|| []`

louis-bompart (approved):
few suggestions

packages/headless/src/features/facets/range-facets/date-facet-set/date-facet-selectors.ts
dmbrooke: I like this idea, but it wouldn't _quite_ work exactly as suggested because the filtering logic below depends on the `values` property of this response. Because of this, it would either have to be
```
const facetResponse = dateFacetResponseSelector(state, facetId) ?? {
    values: [],
  };
```

OR update all of the logic within this method to

```
const facetResponseValues =
    dateFacetResponseSelector(state, facetId)?.values ?? [];
  return facetResponseValues.filter((value) => value.state === 'selected');
```

I think I personally prefer the first one 🙂

PR #3000, opened 2023-06-26
feat(quantic): Added disableDynamicNavigation in quanticRefineToggle and quanticRefineModalContent components

SimonMilord:
[SFINT-5087](https://coveord.atlassian.net/browse/SFINT-5087)

This PR is linked with this ticket: https://coveord.atlassian.net/browse/SFINT-5044

We are passing the disableDynamicNavigation property to support a new (yet to be added) Dynamic Navigation Experience config in the Hosted Insight Panel

[SFINT-5087]: https://coveord.atlassian.net/browse/SFINT-5087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mmitiche (commented):
We need E2E Cypress tests to cover this new logic.
We already have test suites for these two components: [cypress/e2e/default-1/refine-modal-content](https://github.com/coveo/ui-kit/tree/master/packages/quantic/cypress/e2e/default-1/refine-modal-content) and  [cypress/e2e/default-1/refine-toggle](https://github.com/coveo/ui-kit/tree/master/packages/quantic/cypress/e2e/default-1/refine-toggle)

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.html
```diff
@@ -28,7 +28,7 @@
           ></lightning-button>
         </template>
       </div>
-      <c-quantic-facet-manager engine-id={engineId}>
+      <template lwc:if={disableDynamicNavigation}>
```
mmitiche, line 31: To avoid duplicating the same logic in the same file and ending up with a big HTML file with some conditions in it, what do you think about having two separate HTML templates to cover these two situations, one template will use the facet manager and the second won't.

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -52,6 +47,13 @@ export default class QuanticRefineModalContent extends LightningElement {
    * @type {boolean}
    */
   @api hideSort;
+  /**
+   * Indicates whether the Dynamic Navigation for facet ordering should be disabled.
+   * @api
+   * @type {boolean}
+   * @defaultValue `false`
+   */
+  @api disableDynamicNavigation = false;
```
mmitiche, line 56: Since the DNE feature includes more than just the facet ordering like documented here: https://docs.coveo.com/en/3383/leverage-machine-learning/about-dynamic-navigation-experience-dne
What do you think about making the name of this property very specific to what the property does.
`disableFacetOrdering` for example.

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.html
SimonMilord: not a bad idea, I was wondering if that would be an issue in the future though if we start making a template for each config, what if we have config A and config B, then we would need template A & !B, template A & B, template !A & B , template !A & !B ?

mmitiche: Not every boolean property will make us create a new template in a component.
 Only the properties that yield big code duplication in the HTML are good candidates for this multiple templates rendering method.

This property seems like a good candidate  since  without multiple templates we would do something like this: 
```
  <template lwc:if={condition}>
     Majority of the HTML of the component  
</template>
  <template ekse={condition}>
     Majority of the HTML of the component  duplicated
</template>
``` 
I think having two separate templates in this situation can make the component a lot easier to read.

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -267,4 +277,11 @@ export default class QuanticRefineModalContent extends LightningElement {
   setInitializationError() {
     this.hasInitializationError = true;
   }
+
+  render() {
+    if (this.disableFacetOrdering === true) {
```
mmitiche, line 282:
```suggestion
    if (this.disableFacetOrdering) {
```

packages/quantic/cypress/e2e/default-1/refine-modal-content/refine-modal-content.cypress.ts
```diff
@@ -89,4 +90,18 @@ describe('quantic-refine-content', () => {
...
+    it('should not render the facets inside the quantic-facets-manager', () => {
+      visitRefineContent({
+        disableFacetOrdering: true,
+      });
+
+      scope('when loading the page', () => {
+        Expect.displayFiltersTitle();
+        Expect.displayClearAllFiltersButton(false);
+        Expect.displayFacetManager(false);
+      });
```
mmitiche, line 104: We should also make sure that the duplicated facets are displayed  properly as well in the case where the `disableFacetOrdering` property is set to true.

packages/quantic/cypress/e2e/default-1/refine-toggle/refine-toggle.cypress.ts
mmitiche, line 1: We should also add `Expect.displayFacetManager(true);` to the test suites where the property  `disableFacetOrdering` is set to false.

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineModalContent/exampleQuanticRefineModalContent.js
```diff
@@ -15,6 +15,13 @@ export default class ExampleQuanticRefineModalContent extends LightningElement {
       description: 'Whether the Quantic Sort component should be hidden.',
       defaultValue: false,
     },
+    {
+      attribute: 'disableFacetOrdering',
+      label: 'Disable Facet Ordering',
+      description:
+        'Whether the Dynamic Navigation for facet ordering should be disabled.',
```
mmitiche, line 22: This description should be updated to accurately reflect the new property name.

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineToggle/exampleQuanticRefineToggle.js
```diff
@@ -60,6 +61,13 @@ export default class ExampleQuanticRefineToggle extends LightningElement {
       description: 'The title of the toggle button.',
       defaultValue: 'Sort & Filters',
     },
+    {
+      attribute: 'disableFacetOrdering',
+      label: 'Disable Facet Ordering',
+      description:
+        'Indicates whether the Dynamic Navigation for facet ordering should be disabled.',
```
mmitiche, line 68: same remark about the description.

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -52,6 +55,13 @@ export default class QuanticRefineModalContent extends LightningElement {
    * @type {boolean}
    */
   @api hideSort;
+  /**
+   * Indicates whether the Dynamic Navigation for facet ordering should be disabled.
```
mmitiche, line 59:
```suggestion
   * Indicates whether to disable the facet ordering feature according to [the dynamic navigation experience](https://docs.coveo.com/en/3383/leverage-machine-learning/about-dynamic-navigation-experience-dne).
```

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -75,6 +71,13 @@ export default class QuanticRefineToggle extends LightningElement {
    * @type {string}
    */
   @api title = this.labels.sortAndFilters;
+  /**
+   * Indicates whether the Dynamic Navigation for facet ordering should be disabled.
```
mmitiche, line 75:
```suggestion
   * Indicates whether to disable the facet ordering feature according to [the dynamic navigation experience](https://docs.coveo.com/en/3383/leverage-machine-learning/about-dynamic-navigation-experience-dne).
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticRefineModalContent/exampleQuanticRefineModalContent.js
```diff
@@ -15,6 +16,12 @@ export default class ExampleQuanticRefineModalContent extends LightningElement {
       description: 'Whether the Quantic Sort component should be hidden.',
       defaultValue: false,
     },
+    {
+      attribute: 'disableFacetOrdering',
```
mmitiche, line 20: You are not passing this property in the HTML file of the ExampleQuanticRefineToggle component.
I wonder, how the E2E of this component are passing?

packages/quantic/cypress/e2e/default-1/refine-toggle/refine-toggle.cypress.ts
mmitiche, line 1: These new tests you added need to be redesigned.

If you observe the actual structure of the the current tests you will find it as follows : 
-  describe('when no facets are registered', () => {})
 -  describe('when facets with inputs are used in the search interface', () => {})
 -  describe('when facets without inputs are used in the search interface', () => {})

In the new describe you added, aka describe('when the disableFacetOrdering property is set to true', () => {}), you didn't use any of the functions `addFacets`  or `addFacetsWithInputs` which means that the tests were performed without any facets. and when there is no facets it means that there will be no Facet Manager regardless of the value passed to the property `disableFacetOredering`. that's why the tests were passing even without this property passed in the HTML.

The new test you added needs to be inside the two describes: 
 -  describe('when facets with inputs are used in the search interface', () => {})
 -  describe('when facets without inputs are used in the search interface', () => {})
 we wanna test the existence of the facet manger when there is fact some facets displayed.

packages/quantic/cypress/e2e/default-1/refine-toggle/refine-toggle.cypress.ts
```diff
@@ -145,6 +146,7 @@ describe('quantic-refine-toggle', () => {
             Expect.refineToggleTitleContains(defaultRefineToggleTitle);
             Expect.displayRefineToggleIcon(true);
             Expect.displayModal(false);
+            Expect.displayFacetManager(true);
```
mmitiche, line 149: No need to check when Loading the page. checking this in the scope `when opening the refine modal` would make more sense and would be enough.

mmitiche (changes requested):
The tests need to be redesigned, please confirm that the tests fail as expected if you don't pass the `disableFacetOredering` property or if you put  `Expect.displayFacetManager(true);` in a scenario where the facet manager should not be displayed and vice versa.

packages/quantic/cypress/e2e/default-1/refine-toggle/refine-toggle.cypress.ts
SimonMilord: yeah makes sense! thx 👍

packages/quantic/cypress/e2e/default-1/refine-toggle/refine-toggle.cypress.ts
SimonMilord: I have redesigned the tests to move the disableFacetOrdering tests in the describes. I also tested that the tests fail if I remove the attribute in the toggle HTML, which they did. Now they are passing and we are adding the facets properly.
👌 Thank you for the review :D

mmitiche (approved):
Nice!

mmitiche (approved):

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -32,7 +31,7 @@ import {LightningElement, api} from 'lwc';
  * @category Search
  * @category Insight Panel
  * @example
- * <c-quantic-refine-modal-content engine-id={engineId} hide-sort></c-quantic-refine-modal-content>
+ * <c-quantic-refine-modal-content engine-id={engineId} hide-sort disable-facet-ordering={disableFacetOrdering}></c-quantic-refine-modal-content>
```
erocheleau, line 34:
```suggestion
 * <c-quantic-refine-modal-content engine-id={engineId} hide-sort disable-facet-ordering></c-quantic-refine-modal-content>
```

This should be what's valid in the html.

erocheleau: Just having the attribute there on the HTML is equivalent to writing `disable-facet-ordering="true"`.

erocheleau (approved):
Other than me wondering why there's so many files changed for just 1 small thing, everything looks good?

Might want to set the default value like I suggested so you don't have to type "true" all the time.

SimonMilord:
> Other than me wondering why there's so many files changed for just 1 small thing, everything looks good?
> 
> Might want to set the default value like I suggested so you don't have to type "true" all the time.

🤣  11/18 files are due to e2e tests and I believe 4 files were affected by the templates used thats why... Got to love e2e...

packages/quantic/cypress/e2e/default-1/refine-modal-content/refine-modal-content-expectations.ts
```diff
@@ -52,11 +52,11 @@ function refineContentExpectations(selector: RefineContentSelector) {
         .logDetail(`${should(display)} the clear all filters button`);
     },
 
-    displayFacetManager: () => {
+    displayFacetManager: (display: boolean) => {
```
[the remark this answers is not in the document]
SimonMilord: yeah makes sense, will set to true by default

packages/quantic/cypress/e2e/default-1/refine-toggle/refine-toggle.cypress.ts
```diff
@@ -76,6 +77,7 @@ describe('quantic-refine-toggle', () => {
             Expect.displayRefineModalTitle(true);
             Expect.refineModalTitleContains(customRefineModalTitle);
             Expect.displayModalContent(true);
+            Expect.displayFacetManager(false);
```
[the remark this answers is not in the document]
SimonMilord: yeah not when no facets are registered using default options

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -75,6 +71,13 @@ export default class QuanticRefineToggle extends LightningElement {
    * @type {string}
    */
   @api title = this.labels.sortAndFilters;
+  /**
+   * Indicates whether to disable the facet ordering feature according to [the dynamic navigation experience](https://docs.coveo.com/en/3383/leverage-machine-learning/about-dynamic-navigation-experience-dne).
+   * @api
+   * @type {boolean}
+   * @defaultValue `false`
+   */
+  @api disableFacetOrdering = false;
```
erocheleau, line 80: This should be called `disableDynamicNavigation` to be consistent everywhere.

erocheleau (changes requested):

packages/quantic/force-app/main/default/lwc/quanticRefineModalContent/quanticRefineModalContent.js
```diff
@@ -52,6 +51,13 @@ export default class QuanticRefineModalContent extends LightningElement {
    * @type {boolean}
    */
   @api hideSort;
+  /**
+   * Indicates whether to disable the facet ordering feature according to [the dynamic navigation experience](https://docs.coveo.com/en/3383/leverage-machine-learning/about-dynamic-navigation-experience-dne).
+   * @api
+   * @type {boolean}
+   * @defaultValue `false`
+   */
+  @api disableFacetOrdering = false;
```
erocheleau, line 60: Same as the `quanticRefineToggle`, it should be called `disableDynamicNavigation`

erocheleau (changes requested):

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
SimonMilord: the reason we named it that way was to be more specific since the DNE includes more features than facetOrdering 
[see here](https://github.com/coveo/ui-kit/pull/3000#:~:text=Since%20the%20DNE,disableFacetOrdering%20for%20example)

erocheleau (approved):

PR #3002, opened 2023-06-26
feat(atomic): add result action bar on hover

jelmedini:
[SVCC-2942](https://coveord.atlassian.net/browse/SVCC-2942)

Add result action bar component for the insight panel use case
Acceptance Criteria:

- The result actions button bar should show when hovering over a result
- Hovering over the buttons should display a tooltip with the name of the result action
- The result action should require an icon and a label
- Hovering over a result should change the background color. 

When Hover :
Screenshot 2023-06-28 at 3 33 52 PM

Video Demo:

https://github.com/coveo/ui-kit/assets/119955059/1dd70a7c-55d2-4723-ba3b-dece04391d6a

Screenshot 2023-06-26 at 7 57 16 PM

[SVCC-2942]: https://coveord.atlassian.net/browse/SVCC-2942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathanlb:
Haven't checked the code yet but the video looks real good! Couple styling comments:
1) Nice tooltip!
2) Could we have the action bar aligned completely to the right? Looks like there's an empty space there.
3) Maybe we should use a shade of grey that is slightly lighter than the default badge color so that we can still see it?

248925819-3906d18e-98bd-4b84-ad46-39a7c4a36983

packages/atomic/src/components.d.ts
```diff
@@ -613,6 +613,18 @@ export namespace Components {
...
     }
+    interface AtomicInsightResultAction {
+        /**
+          * Specify the result action icon to display.
+         */
+        "icon": string;
+        /**
+          * The text tooltip to show on the result action icon
+         */
+        "label"?: string;
```
lbergeron, line 624: Sometimes result actions are toggle buttons. Did you consider supporting different labels depending on the button state? In addition, the operation performed when activating a toggle button is not instant. We need to be able to support a loading/processing state as well.

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
```diff
@@ -139,9 +154,21 @@ export class AtomicInsightResult {
...
+      'atomic-insight-result-action'
+    );
     if (this.loadingFlag && this.store) {
       this.store.unsetLoadingFlag(this.loadingFlag);
     }
+    if (resultActionBarElement && resultActionElement?.length) {
+      this.host.parentElement!.style.padding = '0px';
+      this.host.classList.add('withActionBar');
+      this.host.addEventListener('mouseover', this.handleMouseOver);
+      this.host.addEventListener('mouseout', this.handleMouseOut);
```
lbergeron, line 170: The event handlers should be disconnected when the component unloads.

lbergeron (commented):
It looks quite nice!

packages/atomic/src/components/insight/atomic-insight-result-action-bar/atomic-insight-result-action-bar.pcss
```diff
@@ -0,0 +1,29 @@
+@import '../../../global/global.pcss';
+
+atomic-insight-result-action-bar {
+  display: flex;
```
nathanlb, line 4: I think a lot of this css can be replaced by tailwind classes which is the preferred way of styling in atomic :)

packages/atomic/cypress/integration-insight-panel/insight-panel.cypress.ts
```diff
@@ -67,6 +67,15 @@ describe('Insight Panel test suites', () => {
         .should('have.attr', 'href');
     });
 
+    it('should display result action bar', () => {
+      InsightPanelsSelectors.resultActionBar()
```
nathanlb, line 71: Would we be able to trigger the hover and assert that the buttons are actually visible?

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
```diff
@@ -139,9 +154,21 @@ export class AtomicInsightResult {
...
+      'atomic-insight-result-action-bar'
+    );
+    const resultActionElement = this.host.shadowRoot?.querySelectorAll(
+      'atomic-insight-result-action'
+    );
     if (this.loadingFlag && this.store) {
       this.store.unsetLoadingFlag(this.loadingFlag);
     }
+    if (resultActionBarElement && resultActionElement?.length) {
+      this.host.parentElement!.style.padding = '0px';
```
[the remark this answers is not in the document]
jelmedini: the way how the results are displayed make it impossible to add the background color properly without touching the css of the parent elements, with the padding and also the separators between each result etc...

packages/atomic/src/components.d.ts
```diff
@@ -3499,6 +3525,18 @@ declare namespace LocalJSX {
...
+    interface AtomicInsightResultAction {
+        /**
+          * Specify the result action icon to display.
+         */
+        "icon"?: string;
+        /**
+          * The text tooltip to show on the result action icon
+         */
+        "label"?: string;
+    }
```
louis-bompart, line 3537: :suspect: why does those interface keys are wrapped in double quotes?

packages/atomic/src/components/insight/atomic-insight-result-action-bar/atomic-insight-result-action-bar.pcss
louis-bompart: - [ ] tailwind magic

packages/atomic/cypress/integration-insight-panel/insight-panel.cypress.ts
nathanlb: ah i see thanks

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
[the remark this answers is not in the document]
louis-bompart: Mhhh, could you add a wrapper div and use margin or similar?
A component affecting modifying its parent is really iffy.

If it works, it means the parent dom is not Shadowed from the child, which means that goes both way: a CSS rule from a parent might conflict with that, and well: 💥 (I think at least, I'm a bit rusty CSS wise)

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
```diff
@@ -123,6 +123,21 @@ export class AtomicInsightResult {
       .join('');
   }
 
+  private handleMouseOver = () => {
+    const resultActionBarElement = this.host.shadowRoot?.querySelector(
```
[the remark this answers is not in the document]
jelmedini: isn't creating a variable much better ? `private resultActionBarElement?: HTMLElement | null;` it is set iafter once inside `ComponentDidLoad` and can use it after directly

louis-bompart (commented):
Some :suspect: stuff to address

packages/atomic/src/components.d.ts
jelmedini: they are all wrapped in double quotes 🤔

nathanlb: Indeed they are as this is the generated components file 😅

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
```diff
@@ -139,9 +149,33 @@ export class AtomicInsightResult {
...
+    );
+    const resultActionElement = this.host.shadowRoot?.querySelectorAll(
+      'atomic-insight-result-action'
+    );
     if (this.loadingFlag && this.store) {
       this.store.unsetLoadingFlag(this.loadingFlag);
     }
+    if (this.resultActionBarElement && resultActionElement?.length) {
+      this.host.parentElement!.style.padding = '0px';
+      this.host.parentElement
```
nathanlb, line 163: I think all the querySelector instances could be put into private getters and it would improve readability nicely

jelmedini: sure!

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
```diff
@@ -123,6 +123,34 @@ export class AtomicInsightResult {
...
+
+  private get resultActions() {
+    return this.host.shadowRoot?.querySelectorAll(
+      'atomic-insight-result-action'
+    );
+  }
+
+  private get firstChildActionBar() {
+    return this.host.parentElement
+      ?.querySelectorAll('atomic-insight-result')[0]
```
nathanlb, line 150: are we missing an optional chainer here? for when the array is empty

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.pcss
```diff
@@ -5,3 +5,21 @@ atomic-result-section-children {
   border-left: 1px solid var(--atomic-neutral);
   padding-left: 1rem;
 }
+
+:host(.resultHovered) {
+  @apply relative;
+
+  .with-sections {
+    background-color: #f2f2f2;
```
nathanlb, line 13: Colors should come from variables so if there isn't an existing one in `coveo.css` that works we can add one.

nathanlb (approved):
Still a couple comments but this is looking great 👍

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
jelmedini: the getter is called after checking that we have `this.resultActions?.length` but I could add the optional chainer also here :)

nathanlb: Yeah IMO a getter should not be implemented based on where it's called. Once it's a getter is should be a secure and reliable way to get a value on its own. If we add code in the future that uses the getter as well it would be unnecessarily difficult to remember this requirement to check for `this.resultActions?.length` before using it.

lvu285:
Nice work!
#1. Can you plz check to make sure that the action bar should not overlap with the item title when title doens't have `badge` + the title is very long (similar scenario like folding item _ but title is longer). Maybe it should be more consistent with the case when result has badge
#2. Can you plz check to make sure if only 1 action (1 icon) display, it still displays nicely _ include round corner
#3. From the comment of Nathan, I can see that we have the tooltips for these icons, just want to double confirm that we have it, as on the demo video, I don't see it. 
Thanks

packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.pcss
```diff
@@ -5,3 +5,21 @@ atomic-result-section-children {
...
+}
+
+:host {
+  .with-sections {
+    @apply p-6;
+  }
+}
+
+:host::before {
+  @apply my-0 mx-6 !important;
```
louis-bompart, line 24: :finnadie: `!important` is a pain to deal with and often a code smell.
Can you try to increase the specificity of your CSS selector instead?

louis-bompart (changes requested):
I'd prefer to see this `!important` gone.

jelmedini:
> Nice work! #1. Can you plz check to make sure that the action bar should not overlap with the item title when title doens't have `badge` + the title is very long (similar scenario like folding item _ but title is longer). Maybe it should be more consistent with the case when result has badge #2. Can you plz check to make sure if only 1 action (1 icon) display, it still displays nicely _ include round corner #3. From the comment of Nathan, I can see that we have the tooltips for these icons, just want to double confirm that we have it, as on the demo video, I don't see it. Thanks

hey @lvu285 !
- for point 1: the video isn't updated but the result bar action is in between each result, EXCEPT the first result, as it is hidden if we do so, the layout component overflow is set to auto can't change that! but here it is how it looks like in screenshots.
- point 2: yes I will include how it looks like when we have only one button
- for tooltips it's actually implemented the same way as edit and toggle button as it is the same button component, they aren't quite intuitive I agree you need to be a little bit precise with the cursor.
Screenshot 2023-07-05 at 4 46 08 PM

Screenshot 2023-07-05 at 4 44 21 PM

Screenshot 2023-07-05 at 4 43 52 PM

lvu285:
awesome, only left 1 small comment, the rest is great. Thanks Jihan!
image

> 
> hey @lvu285 !
> 
> * for point 1: the video wasn't updated but the result bar action is in between each result, EXCEPT the first result, as it is hidden if we do so, the layout component overflow is set to auto can't change that! but here it is how it looks like in screenshots. (I updated the demo video also)
> * point 2: yes I will include how it looks like when we have only one button
> * for tooltips it's actually implemented the same way as edit and toggle button as it is the same button component, they aren't quite intuitive I agree you need to be a little bit precise with the cursor.

louis-bompart (changes requested):
Good chunks of logic could/should be handled through CSS https://github.com/coveo/ui-kit/pull/3019

louis-bompart (approved):

PR #3005, opened 2023-06-28
fix(quantic): Changed the quantic_OpensInSalesforceSubTab to something more generic

SimonMilord:
[SFINT-5084](https://coveord.atlassian.net/browse/SFINT-5084)

In this PR:

The `quantic_OpensInSalesforceSubTab` localized label currently contains the value `Opens in a Salesforce console subtab`, which is a bit wrong.

We should not mention  “Salesforce console” in this label because it doesn't make sense to display this label for a result that’s being displayed in a Salesforce community. 

- I have changed the label to `Opens in a Salesforce subtab` to be more generic and make more sense in Salesforce Communities.

[SFINT-5084]: https://coveord.atlassian.net/browse/SFINT-5084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1093,11 +1093,11 @@
       <shortDescription>No related items</shortDescription>
     </labels>
     <labels>
-      <fullName>quantic_OpensInSalesforceSubTab</fullName>
```
mmitiche, line 1096: The old name of the label seems good to me.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1093,11 +1093,11 @@
...
-      <value>Opens in a Salesforce console subtab</value>
+      <fullName>quantic_OpensInSubTab</fullName>
+      <value>Opens in a subtab</value>
       <language>en_US</language>
       <protected>false</protected>
-      <shortDescription>Opens in a Salesforce console subtab</shortDescription>
```
mmitiche, line 1100: in this label we have a problem with the word `console`. because when we have a Salesforce it doesn't necessary mean that we are in a Salesforce console, we can be in a Salesforce community.
I would suggest to use the value `Opens in a Salesforce sub tab`.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
SimonMilord: ok this was to match the new label value more closely but since we keep 'Salesforce' I will revert it!

mmitiche (approved):
👌

nathanlb (approved):

PR #3006, opened 2023-06-29
fix(SFINT-5094): Attempt to fix the quantic package release

erocheleau:


packages/quantic/scripts/build/util/sfdx.ts
```diff
@@ -26,18 +26,20 @@ export function sfdx<T = SfdxResponse>(command: string): Promise<T> {
...
       (error, stdout) => {
-        if (error) {
-          let jsonOutput;
+        let jsonOutput: unknown;
+        if (stdout) {
           try {
             jsonOutput = JSON.parse(strip(stdout));
-          } catch (e) {
+            if (error) {
+              reject(jsonOutput || error);
```
nathanlb, line 34: maybe we should throw here instead so that the stdout gets logged on line 39

nathanlb: or just add logging here as well given we reject here with different info

louis-bompart (approved):

packages/quantic/scripts/build/util/sfdx.ts
erocheleau: added

packages/quantic/scripts/build/util/sfdx.ts
```diff
@@ -26,18 +26,22 @@ export function sfdx<T = SfdxResponse>(command: string): Promise<T> {
...
+              console.error(jsonOutput);
+              console.error(error);
```
louis-bompart, line 36:
```suggestion
              console.error(jsonOutput);
              console.error(error);
              reject(jsonOutput || error);
```

PR #3009, opened 2023-07-03
feat(quantic): aria-live logic added to insight, folding and recommendations components

mmitiche:
[SFINT-5083](https://coveord.atlassian.net/browse/SFINT-5083)

- Aria live regions logic added to insight, folding and recommendations components.

#### Aria live regions before: the loading region was not registered
Folding-1

#### Aria live regions after: the loading region is now properly registered
Folding-2

#### The screen reader properly announcing the loading of the results:

https://github.com/coveo/ui-kit/assets/86681870/e86e7f51-d938-4354-b9e1-58c737183fb1

#### Aria live regions before: No region is being  registered
insight-vefore

#### Aria live regions after:all the regions regions are now properly registered
Insight-after

#### The screen reader properly announcing the content of the error aria live region:

https://github.com/coveo/ui-kit/assets/86681870/524fb122-2630-42f6-a5ad-97f5adc0eddd

  

[SFINT-5083]: https://coveord.atlassian.net/browse/SFINT-5083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -636,4 +636,12 @@
...
         <name>quantic_UnableToInitializeComponent</name>
     </customLabels>
+        <customLabels>
+        <label>Chargement des résultats de la recherche</label>
+        <name>quantic_LoadingResults</name>
+    </customLabels>
+        <customLabels>
+        <label>Chargement des recommandations</label>
+        <name>quantic_LoadingRecommendations</name>
+    </customLabels>
```
SimonMilord, line 646:
```suggestion
    </customLabels>
    <customLabels>
        <label>Chargement des résultats de la recherche</label>
        <name>quantic_LoadingResults</name>
    </customLabels>
    <customLabels>
        <label>Chargement des recommandations</label>
        <name>quantic_LoadingRecommendations</name>
    </customLabels>
```

packages/quantic/force-app/main/default/lwc/quanticInsightInterface/quanticInsightInterface.js
```diff
@@ -93,4 +111,39 @@ export default class QuanticInsightInterface extends LightningElement {
...
+  }
+
+  handleAriaLiveMessage(event) {
+    /** @type {import('quanticAriaLive/quanticAriaLive').IQuanticAriaLive} */
+    const ariaLiveRegion = this.querySelector('c-quantic-aria-live');
+    if (ariaLiveRegion) {
+      ariaLiveRegion.updateMessage(
+        event.detail.regionName,
+        event.detail.message,
+        event.detail.assertive
```
SimonMilord, line 134: question: Is loading considered critical here? I believe assertive should be used as little as possible as it can be annoying. Have you considered `polite` instead here? 

From the [ARIA docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#:~:text=aria%2Dlive%3D%22assertive,be%20used%20sparingly.)

SimonMilord (approved):
Just a question about the `assertive` but other than that looks good 👌

packages/quantic/force-app/main/default/lwc/quanticInsightInterface/quanticInsightInterface.js
mmitiche: We are not using assertive aria live regions here.

Here we are only calling the method `updateMessage` of the QuanticAriaLive component and we are passing to it, as parameters, the the payload of the sent event.

If we take a step back and follow the creation of the event we see that in the payload of it the value of `assertive` is set to `false`. this would make the QuanticArialive component create a polite aria live region and not an assertive region.

erocheleau (approved):
Other than the failing build, everything looks good.

PR #3011, opened 2023-07-04
Kit 2552 alphanumeric facet desc order

dmbrooke:


packages/headless/src/features/facets/facet-set/interfaces/options.ts
```diff
@@ -1,6 +1,8 @@
-import {FacetRequest} from './request';
+import {FacetOptions} from '../../../../product-listing.index';
 
-export type FacetOptionalParameters = Pick<
-  FacetRequest,
-  'filterFacetCount' | 'injectionDepth' | 'numberOfValues' | 'sortCriteria'
+export type FacetOptionalParameters = Required<
```
dmbrooke, line 3: The `sortCriteria` in the FacetRequest interface can now either be `FacetSortCriterion` OR `SpecificSortCriteriaExplicitAlphanumeric`, which was leading to type problems that I couldn't resolve in a few places, so I had to update this type such that it remained the same as before.

louis-bompart (approved):
lgtm!
for funsies, maybe modify one of our sample to use this?

packages/headless/src/features/facets/facet-set/interfaces/options.ts
```diff
@@ -1,6 +1,8 @@
-import {FacetRequest} from './request';
+import {FacetOptions} from '../../../../product-listing.index';
```
aprudhomme-coveo, line 1: Should you not import FacetOptions from the normal controllers instead of product-listing ?

like this import {FacetOptions} from '../../../../controllers';

dmbrooke: Oh yes, that would make more sense 🙂

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -1,5 +1,6 @@
 import {Schema, StringValue} from '@coveo/bueno';
 import {AnyFacetValuesCondition, AnyFacetValueRequest} from '@coveo/headless';
+import {SpecificSortCriteriaExplicitAlphanumeric} from '@coveo/headless/dist/definitions/features/facets/facet-api/request';
```
olamothe, line 3: That's a red flag, we should not be relying on types not exposed by Headless at a top level in Atomic.

It means the type is not properly exposed by Headless.

Secondly, (and more importantly) I don't think that's even something that should be leaking from Headless in the first place, or that we should need/reference in consumer projects (ie: Atomic).

My feeling is that the only thing we should expose from Headless is `FacetSortCriterion`, which is just the finite list of sort criterion.

And then internally in Headless, when we do receive `alphanumericDescending`, we map it to that type: `SpecificSortCriteriaExplicitAlphanumeric`.

In v3, if we want to only support those types of `Explicit` sort criteria, it would be a breaking change to manage.

Does it make sense ?

packages/headless/src/features/facets/facet-set/facet-set-actions.ts
```diff
@@ -60,7 +61,7 @@ export interface RegisterFacetActionCreatorPayload {
    *
    * @defaultValue `automatic`
    */
-  sortCriteria?: FacetSortCriterion;
+  sortCriteria?: FacetSortCriterion | SpecificSortCriteriaExplicitAlphanumeric;
```
olamothe, line 64: To re-emphasize on my above point, another way to look at it is that everything "external facing" (controllers, actions) should not even expose anything regarding that type of sort criteria. It adds confusion/complexity to have something polymorphic for sort (ie: 2 different ways to specify sort, even if the API does support it).

The only place that new type should be added is at the "building the request that get sent to the search API" level, where we detect that what's in the store is the string `alphanumericDescending`, and build the proper object with `type` / `order`.

dmbrooke: This makes sense. So just to confirm, _**all instances of sortCriteria**_ should be of type `FacetSortCriterion` _**except for in the `FacetRequest` interface**_, in which case `SpecificSortCriteriaExplicitAlphanumeric` is also allowed?

If this is the case, I'll update the logic here such that there is no mention of the FacetRequest type, but `buildFacetRequest` will still "quietly" update sortCriteria to accommodate the new Search API specs.

Does this sound good to you? 🙂

olamothe (approved):
💯 Good job !

packages/headless/src/features/search/search-request.ts
```diff
@@ -100,13 +101,28 @@ function getAllEnabledFacets(state: StateNeededBySearchRequest) {
...
+    ...getSpecificFacetRequests(state.facetSet ?? {}),
     ...getRangeFacetRequests(state.numericFacetSet ?? {}),
     ...getRangeFacetRequests(state.dateFacetSet ?? {}),
     ...getFacetRequests(state.categoryFacetSet ?? {}),
   ];
 }
 
+function getSpecificFacetRequests<T extends FacetSetState>(state: T) {
+  return getFacetRequests(state).map((request) => {
+    /* The Search API does not support 'alphanumericDescending' as a string value and instead relies on a new sort mechanism to specify sort order.
```
olamothe, line 113: Have you created the corresponding ticket in the V3 epic for UI kit ?

dmbrooke: Yup 🙂 
https://coveord.atlassian.net/browse/KIT-2578

packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts
```diff
@@ -464,8 +464,17 @@ export function buildCoreFacet(
...
+      let sortCriterion!: FacetSortCriterion;
+
+      if (typeof request.sortCriteria === 'object') {
+        sortCriterion =
+          request.sortCriteria.order === 'descending'
+            ? 'alphanumericDescending'
+            : 'alphanumeric';
+      } else {
+        sortCriterion = request.sortCriteria;
+      }
```
btaillon, line 476: IMO this logic should instead be part of the `facet-set-slice`, since it already contains a `request` object meant to be a match with what we'll send to the API.

btaillon: what if we decided not to support the `alphanumericDescending` string at all (both in the slice and the controller), and required it to be an object? Would that be possible?

While changing the type in `FacetRequest` to accept both the string or an object would be breaking, I don't think the cases where it's actually breaking would be fairly narrow (e.g., somebody casts it to `string` or does some string operations on it). Given that this is a feature and not a fix, I think we also have a bit more leeway to do something potentially (but unlikely) breaking.

packages/headless/src/features/facets/facet-set/interfaces/options.ts
```diff
@@ -1,6 +1,8 @@
-import {FacetRequest} from './request';
+import {FacetOptions} from '../../../../controllers';
```
btaillon, line 1: we should avoid importing from `index.ts` files since this may mess with tree-shaking. you should instead import this from the exact file that defines it.

packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts
MellowM00se: After some discussion over slack, we have decided to continue with the current approach to save headache's from supporting multiple types at once on both our/my end as well as the implementers end. We can support this model end-to-end as a part of V3 once it's been more robust on the Search API's ✅

PR #3013, opened 2023-07-04
feat(headless): add support for automatic facets via the facet manager

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2522

This PR adds support for automatic facets by doing these things:
- Modifies the `headless-facet-manager` to support the request and response of automatic facets
- Adds a new controller for singular automatic facets called `headless-automatic-facet`
- Adds a new automatic-facet-set reducer

After this PR, the roadmap is as follows :
- implement in atomic
-  make react samples 
- add support for breakbox
- add support for url manager
- add a facet search


packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
```diff
@@ -0,0 +1,121 @@
...
+      const response = getFacetResponse();
+      const field = response ? response.field : '';
+      const values = response ? response.values : [];
+      const label = response ? response.label : '';
+
+      return {
+        field,
+        label,
+        values,
+      };
```
louis-bompart, line 118: nit: here you evaluate 3 times your if. Could be evaluated once:

```suggestion
      const defaultState = 

      return response ? {
        field: response.field,
        label: response.label,
        values: response.value,
      } : {field: '', values: [], label: ''};
```

packages/headless/src/features/history/history-state.ts
```diff
@@ -37,6 +38,8 @@ export function extractHistory(state: Partial<HistoryState>): HistoryState {
     dateFacetSet: state.dateFacetSet || getDateFacetSetInitialState(),
     categoryFacetSet:
       state.categoryFacetSet || getCategoryFacetSetInitialState(),
+    automaticFacetSet:
+      state.automaticFacetSet ?? getAutomaticFacetSetInitialState(),
```
louis-bompart, line 42: 🤔 curious, why nullish here? (wondering because all others are ||)

louis-bompart (commented):
Looks sound! Is there a a way to test it?

packages/headless/src/controllers/facet-manager/headless-facet-manager.ts
```diff
@@ -1,18 +1,83 @@
...
-export function buildFacetManager(engine: SearchEngine): FacetManager {
-  return buildCoreFacetManager(engine);
+export function buildFacetManager(
+  engine: SearchEngine,
+  props?: FacetManagerProps
+): FacetManager {
+  if (!props?.desiredCount) {
+    return buildCoreFacetManager(engine);
+  }
+  if (!loadAutomaticFacetSetActions(engine)) {
```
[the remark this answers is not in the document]
aprudhomme-coveo: You are right. It makes no sense. I added the `loadFacetManagerReducers` it to copy how other controllers do it. Eventually, it will have more reducers.

packages/headless/src/features/history/history-state.ts
aprudhomme-coveo: Nullish is the right operator in this case. I think the others were made before the nullish operator came out or was better known. Sonarlint tells me to change the '||' to '??', that's why I used the nullish operator instead.

packages/headless/src/controllers/facet-manager/headless-facet-manager.ts
```diff
@@ -1,18 +1,100 @@
...
 /**
  * Creates a `FacetManager` instance.
  *
  * @param engine - The headless engine.
+ * @param props - The optional facet manager props for the automatic facets feature. Automatic facets are currently in beta testing and should be available soon.
+ * @returns A `FacetManager` controller instance.
  */
-export function buildFacetManager(engine: SearchEngine): FacetManager {
-  return buildCoreFacetManager(engine);
+export function buildFacetManager(
```
olamothe, line 60: That's one function for which we definitely should annotate JSDOC with `@beta`

packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
```diff
@@ -0,0 +1,121 @@
...
+
+      const defaultState = {field: '', values: [], label: ''};
+
+      return response
+        ? {
+            field: response.field,
+            label: response.label,
+            values: response.values,
+          }
+        : defaultState;
```
olamothe, line 118: This should also work ?

```suggestion
      return response || defaultState;
```

olamothe (approved):
Clean ! Nicely done

packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
louis-bompart: While it might, if response returns more stuff and we want to only returns field, label and values.
So yes it's compatible, but it doesn't do the same thing.

I don't know tho what response could/would be?

aprudhomme-coveo: There is the useless `indexScore` in the response so I would prefer to only keep what is absolutely necessary.

PR #3016, opened 2023-07-07
fix(quantic): optimized the time needed to perform a quick view open operation

mmitiche:
https://coveord.atlassian.net/browse/SFINT-5081

### The issue
Opening the Quick view of certain results take several seconds , approximately 30 seconds: 

https://github.com/coveo/ui-kit/assets/86681870/8a28f871-58e1-4a36-aaa8-117524ad2376

### The source of the issue:
While trying to open the Quick view there was two operations that were taking a very long time to be executed.
the two operations are:
- JSON.parse(JSON.stringify(result)) performed when binding the result link of the Quick view the adequate analytics.
- JSON.parse(JSON.stringify(result)) performed when trying to add the clicked result to the recent result.

This operation was taking a very long delay in results that contained a lot nested objects, the delay was mainly observed when parsing the objects that contains a lot keyword highlights  data in the result object.

To accelerate this I made some optimization to the Quick view components and stopped using JSON.parse(JSON.stringify()).
These two methods  where used because objects in LWC were always proxied: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy
And we were deep cloning the result in order to get the a new object without the Proxy by fear that the Proxy objects wouldn't work with the Headless library.

Turns out, that these steps were not really necessary, Headless is able to perform its logic normally with an object wrapped in a Proxy.
I confirmed that the Headless state is working properly and that both the Result Links and the Recent Results component are behaving as expected.

### After:

https://github.com/coveo/ui-kit/assets/86681870/2ff6a40a-9733-4644-be52-678dc4dda9e4


packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -179,12 +182,10 @@ export default class QuanticResultQuickview extends LightningElement {
...
-    getHeadlessEnginePromise(this.engineId).then((engine) => {
-      const {pushRecentResult} = this.headless.loadRecentResultsActions(engine);
-      engine.dispatch(
-        pushRecentResult(JSON.parse(JSON.stringify(this.result)))
-      );
-    });
+    const {pushRecentResult} = this.headless.loadRecentResultsActions(
+      this.engine
+    );
+    this.engine.dispatch(pushRecentResult(Object.create(this.result)));
```
SimonMilord, line 188: nice!

SimonMilord (approved):
👌

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
erocheleau, line 188: Interesting, I don't know enough about the potential caveats of doing this, but if it works and is extensively tested, should be ok

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
```diff
@@ -101,7 +101,7 @@ export class ResultUtils {
     selector = undefined
   ) {
     const interactiveResult = controllerBuilder(engine, {
-      options: {result: JSON.parse(JSON.stringify(result))},
+      options: {result: result},
```
erocheleau, line 104: This probably has more places than just the quickview where it's used? Are you sure it still works everywhere else it was used?

erocheleau (approved):

packages/quantic/force-app/main/default/lwc/quanticUtils/quanticUtils.js
mmitiche: yes! the other places where this used were also tested.

erocheleau:
If you feel comfortable go ahead @mmitiche , otherwise maybe ask others from ui-kit if they would have counter-indications against using the `Object.create` ?

PR #3017, opened 2023-07-07
feat(headless): isUserActionView property added to the result interface

mmitiche:
[SFINT-5101](https://coveord.atlassian.net/browse/SFINT-5101)

The property `isUserActionView` is now added to the Result interface in Headless.

I described this property as documented in the search-ui repo: https://github.com/coveo/search-ui/blob/a39c8bec0515809812b9917917167e2751dfd0d0/src/rest/QueryResult.ts#L109-L112

In Quantic we rely on this property to display a `viewed by customer badge` component on the results viewed by a given user.

[SFINT-5101]: https://coveord.atlassian.net/browse/SFINT-5101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/api/search/search/result.ts
+++ b/packages/headless/src/api/search/search/result.ts
@@ -174,7 +174,7 @@ export interface Result {
   rankingModifier?: string;
 
   /**
-   * Whether the result item has been previously viewed by one of the users specified in the `canSeeUserProfileOf` section of the search token [search token](https://docs.coveo.com/en/13/api-reference/search-api#tag/Search-V2/operation/token) generated to perform the search request.
+   * Whether the result item has been previously viewed by one of the users specified in the `canSeeUserProfileOf` section of the [search token](https://docs.coveo.com/en/13/api-reference/search-api#tag/Search-V2/operation/token) generated to perform the search request.
    */
   isUserActionView: boolean;
 }
--- a/packages/headless/src/api/search/search/result.ts
+++ b/packages/headless/src/api/search/search/result.ts
@@ -174,7 +174,7 @@ export interface Result {
   rankingModifier?: string;
 
   /**
-   * Whether the result item was previously viewed by the user specified in the request of the query.
+   * Whether the result item has been previously viewed by one of the users specified in the `canSeeUserProfileOf` section of the search token [search token](https://docs.coveo.com/en/13/api-reference/search-api#tag/Search-V2/operation/token) generated to perform the search request.
    */
   isUserActionView: boolean;
 }
--- a/packages/headless/src/api/search/search/result.ts
+++ b/packages/headless/src/api/search/search/result.ts
@@ -172,4 +172,9 @@ export interface Result {
    * TopResult
    */
   rankingModifier?: string;
+
+  /**
+   * Whether the result item was previously viewed by the user specified in the request of the query.
+   */
+  isUserActionView: boolean;
 }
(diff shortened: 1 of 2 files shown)


packages/headless/src/api/search/search/result.ts
```diff
@@ -172,4 +172,9 @@ export interface Result {
    * TopResult
    */
   rankingModifier?: string;
+
+  /**
+   * Whether the result item was previously viewed by the user specified in the request of the query.
```
dmbrooke, line 177:
```suggestion
   * Whether the user who triggered the request has previously viewed the result item.
```

At least I _**believe**_ that's what this description is saying 🙂

dmbrooke (approved):
LGTM!

packages/headless/src/api/search/search/result.ts
btaillon: What exactly toggles this value? I believe there should be a link here to some docs.

Also, if someone views a result then immediately triggers another query with the same result, is this value already going to be `true`, or is there a delay? If there's a delay, I'm not sure that we should include this at all, or at-least we should include a very visible warning. Implementers will think that this value should be `true` in cases where it isn't and they will open support tickets.

btaillon (changes requested):
Re-request changes once you've replied to my message and I'll check it out.

packages/headless/src/api/search/search/result.ts
erocheleau: You can see the field being part of the Search API response schema here: 
image

btaillon: I see. Then, I think the description of this property should be edited to either:
* Specify that this is affected by the `canSeeUserProfileOf` parameter specifically.
* A link to some docs that includes the above clarification.

I really want to avoid any potential confusion around this.

mmitiche, line 177:
```suggestion
   * Whether the result item has been previously viewed by one of the users specified in the `canSeeUserProfileOf` section of the search token [search token](https://docs.coveo.com/en/13/api-reference/search-api#tag/Search-V2/operation/token) generated to perform the search request.
```

packages/headless/src/api/search/search/result.ts
```diff
@@ -172,4 +172,9 @@ export interface Result {
    * TopResult
    */
   rankingModifier?: string;
+
+  /**
+   * Whether the result item has been previously viewed by one of the users specified in the `canSeeUserProfileOf` section of the search token [search token](https://docs.coveo.com/en/13/api-reference/search-api#tag/Search-V2/operation/token) generated to perform the search request.
```
mmitiche, line 177:
```suggestion
   * Whether the result item has been previously viewed by one of the users specified in the `canSeeUserProfileOf` section of the [search token](https://docs.coveo.com/en/13/api-reference/search-api#tag/Search-V2/operation/token) generated to perform the search request.
```

packages/headless/src/api/search/search/result.ts
mmitiche: description of the property updated 👌

erocheleau (approved):

packages/headless/src/api/search/search/result.ts
mmitiche: Do you think this is good enough @btaillon ?

btaillon (approved):
Good description!

PR #3019, opened 2023-07-07
embrace CSS

louis-bompart:


packages/atomic/src/components/insight/atomic-insight-result/atomic-insight-result.tsx
```diff
@@ -123,40 +123,6 @@ export class AtomicInsightResult {
       .join('');
   }
```
jelmedini, line 125: This is a good refacto 🙏 , magic ✨ 
but one only thing, deleting this selectors make the css global and not only when we have result actions implemented and on hover. for example the hover on each result changes color all the time even if we have no result action to show. I personnaly don't find it disturbing to be global, what do you think ? 

case on hover with no result action:

Screenshot 2023-07-07 at 11 40 11 AM

louis-bompart: Good catch, missed it.
Yeah, I think having global is fine too. Maybe validate with the rest of the gang?

If that's a no-go, right now there's no way to do it in CSS only (we need the `:has` pseudo-selector for that, and Firefox does not support it yet). I would then put a CSS class on the host element "hasResultAction" and a TODO to refactor when Firefox supports it).

https://developer.mozilla.org/en-US/docs/Web/CSS/:has

louis-bompart:
@jelmedini I think it's preferable for the position attribute to be as consistent as possible.
So I added cd21616

jelmedini (approved):
@louis-bompart just checked, and actually the effect inside SFINT is like this, we have a color change on hover. I also validated with the UX this change. so we are good to keep the background color hover in general :)

louis-bompart:
> @louis-bompart just checked, and actually the effect inside SFINT is like this, we have a color change on hover. I also validated with the UX this change. so we are good to keep the background color hover in general :)

👍 feel free to merge this PR into yours :)

PR #3021, opened 2023-07-07
feat(atomic): add the singular automatic facet component

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2529

This PR adds support for automatic facets via the facet manager by creating a special component <atomic-automatic-facet> that will only be used internally by the facet manager. 

Right now the component is missing these features, it will be added in next prs:

- url manager support
- breadbox support
- facet search support
- more prop support like isCollapsed and displayValuesAs


packages/headless/src/features/facets/automatic-facet-set/interfaces/request.ts
```diff
@@ -2,5 +2,6 @@ import {FacetValue} from '../../../../product-listing.index';
 
 export type AutomaticFacetRequest = {
   field: string;
+  label: string;
   currentValues: FacetValue[];
 };
```
aprudhomme-coveo, line 7: Talked with jérémie to confirm that I was using the right request and response. I was missing the `label` in the request.

packages/atomic/src/pages/index.html
```diff
@@ -356,7 +356,7 @@
                     </atomic-result-section-excerpt>
                   </atomic-table-element>
                   <atomic-table-element label="author">
-                    <atomic-result-text field="author"></atomic-result-text>
+                    <atomic-result-multi-value-text field="author"></atomic-result-multi-value-text>
```
aprudhomme-coveo, line 359: for some reason, when selecting some automatic facets, it logged an error in the console because of multi value being used in a normal result-text.

packages/atomic/src/components/search/facets/atomic-automatic-facet/atomic-automatic-facet.tsx
```diff
@@ -0,0 +1,147 @@
...
+  @Prop({reflect: true}) public facetId!: string;
+  @Prop({reflect: true}) public facet!: AutomaticFacet;
+  @Prop({reflect: true}) public searchStatus!: SearchStatus;
+  @Prop({reflect: true, mutable: true}) public isCollapsed = false;
+
+  @FocusTarget()
+  private headerFocus!: FocusTargetController;
+
+  private get numberOfSelectedValues() {
+    return this.facet.state.values.filter(({state}) => state === 'selected')
```
louis-bompart, line 61: Here...

packages/atomic/src/components/search/facets/atomic-automatic-facet/atomic-automatic-facet.tsx
```diff
@@ -0,0 +1,147 @@
...
+      .length;
+  }
+
+  private renderValue(facetValue: FacetValue, onClick: () => void) {
+    const displayValue = getFieldValueCaption(
+      this.facet.state.field,
+      facetValue.value,
+      this.bindings.i18n
+    );
+    const isSelected = facetValue.state === 'selected';
```
louis-bompart, line 71: ... and there, you essentially do the same thing.
Maybe you should refactor isSelected as a function:
```ts
const isSelected = (facetValue: FacetValue) => facetValue.state === 'selected';
```

packages/atomic/src/pages/index.html
louis-bompart: would be interesting to know why :P

aprudhomme-coveo: Last friday, I was playing with the platform api to modify fields in the index and I kept using the `author` field for my examples. I must have changed a parameter in the `author` field. 🤔 🤔 Jlessard was also making changes to the searchuisamples org with the api. Anyway, this problem is not caused by the automatic facets. Its something about a change in the certain field.

olamothe (approved):
The code is good 💯  

However as we discussed in person, it might be worth it to create an `atomic-automatic-facets` component which would do the role of what the manager is doing right now.

We could merge most of what's here, except the change to the atomic-facet-manager, and rework that in a subsequent PR

packages/atomic/src/components/search/facets/atomic-automatic-facet/atomic-automatic-facet.tsx
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * An automatic facet is a facet generated by the automatic facets feature. This component should
+ * not be used directly. It is internally used by the facet manager to automatically render updated
+ * automatic facets based on the search results.
+ *
+ * @part facet - The wrapper for the entire facet.
+ *
+ * @part label-button - The button that displays the label and allows to expand/collapse the facet.
+ * @part label-button-icon - The label button icon.
+ * @part clear-button - The button that resets the actively selected facet values.
```
dmbrooke, line 30: Seeing as facet value exclusion is in the works, we should perhaps keep this more generic. This way, we don't need to re-document things if/when we support facet value exclusion for this component 🙂 
```suggestion
 * @part clear-button - The button that resets all facet values to their idle state.
```

dmbrooke (approved):
A question and a few minor suggestions for documentation strings, but this looks and works great! 🎉

packages/atomic/src/components/search/facets/atomic-automatic-facet/atomic-automatic-facet.tsx
```diff
@@ -0,0 +1,150 @@
...
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+
+/**
+ * An automatic facet is a facet generated by the automatic facets feature. This component should
+ * not be used directly. It is internally used by the facet manager to automatically render updated
+ * automatic facets based on the search results.
+ *
+ * @part facet - The wrapper for the entire facet.
+ *
+ * @part label-button - The button that displays the label and allows to expand/collapse the facet.
```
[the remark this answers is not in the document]
aprudhomme-coveo: I agree this is better but I don't like to have the same part in different components but with a different documentation string. We should make another issue to improve all facet parts documentation and make the changes in another pr. Or maybe we could also try to extract them in a variable and reuse the string. I am not sure how we could use variables inside comments.

https://github.com/coveo/ui-kit/blob/ddd2465081c0541c6b0c0032fcd57c074094f39f/packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx#L40 
https://github.com/coveo/ui-kit/blob/ddd2465081c0541c6b0c0032fcd57c074094f39f/packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx#L64

packages/atomic/src/components/search/facets/atomic-automatic-facet/atomic-automatic-facet.tsx
aprudhomme-coveo: same thing here.

packages/atomic/cypress/e2e/facets/automatic-facet/automatic-facet-assertions.ts
```diff
@@ -0,0 +1,55 @@
...
+    BaseFacetSelector.labelButton().should('not.be.empty');
+  });
+}
+
+export function assertValueAtIndex(
+  FacetWithCheckboxSelector: FacetWithCheckboxSelector & BaseFacetSelector,
+  index: number
+) {
+  it(`should go to index ${index}`, () => {
+    FacetWithCheckboxSelector.selectedCheckboxValue().eq(index);
```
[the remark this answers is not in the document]
aprudhomme-coveo: I agree this looks weird lol. The eq cypress function is specifically used get a A DOM element at a specific index in an array of elements. Here is the doc : https://docs.cypress.io/api/commands/eq

packages/atomic/src/components/search/facets/atomic-automatic-facet/atomic-automatic-facet.tsx
```diff
@@ -0,0 +1,150 @@
...
+import {FacetContainer} from '../../../common/facets/facet-container/facet-container';
+import {FacetHeader} from '../../../common/facets/facet-header/facet-header';
+import {FacetValueCheckbox} from '../../../common/facets/facet-value-checkbox/facet-value-checkbox';
+import {FacetValueLabelHighlight} from '../../../common/facets/facet-value-label-highlight/facet-value-label-highlight';
+import {FacetValuesGroup} from '../../../common/facets/facet-values-group/facet-values-group';
+import {Hidden} from '../../../common/hidden';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+
+/**
+ * An automatic facet is a facet generated by the automatic facets feature. This component should
```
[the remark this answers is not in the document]
aprudhomme-coveo: I did a design doc 1 month ago. The part about atomic/headless is a bit outdated but the context/searchAPI part is the same. there is also a bunch of links in the epic if you are curious. Basically, right now we are sending requests for every facets. Some clients have thousands of facets. We would like to have facets generated by the index that are dynamic and that depend on the search query. 

There is no mention on the doc site because this is a beta search api feature. From what I understand, there is only 1 client using this feature (Bunnings). The GA will probably be in september.

I will add a sentence to clear up the difference between normal and automatic facets. 

https://coveord.atlassian.net/browse/KIT-2503
https://docs.google.com/document/d/1E6OdgIdxovs2_wcTyemsKNtOXU7N9VTP9CpgKPacyho/edit#heading=h.uyanedjmkl4o

packages/atomic/src/components/search/facets/atomic-automatic-facet/atomic-automatic-facet.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * @beta - This component is part of the automatic facets feature.
+ * Automatic facets are currently in beta testing and should be available soon.
+ *
+ * An automatic facet is a special type of facet generated by the automatic facets feature. Unlike traditional facets that
+ * need to be explicitly defined and requested in the search query, automatic facets are dynamically generated by the index
+ * in response to the search query. It's important to note that this component should not be used directly. It is internally used
+ * by the `atomic-automatic-facets` component to automatically render updated facets.
+ *
```
aprudhomme-coveo, line 29: Is this better @dmbrooke ? What do you think ?

PR #3022, opened 2023-07-07
Boilerplate for unified API support + working Recs/Badges

olamothe:
* Add a new engine: Tentative name is `CommercePlacements`

* The new engine supports badging and Qubit recommendations features. I removed the support for personalized content.

* Created a `unified-api-client`, which inherits from most of the generic features we are using for all other service clients (throttling/backoff-retry/auth token renewal/etc.). It mostly defines types that are linked to the unified API.

* Add a `placement` section in the state:
  *  It is an object, which contains `recs` and `badge`
  * Recs and badge are themselves object with the key being the `placementId`, and the content being the actual data were interested in (array of products/of badge, callbackData, ie: the response coming from the unified API)


packages/headless/src/integration-tests/commerce-placements.test.ts
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable @cspell/spellchecker */
+import {buildCommercePlacementsEngine} from '../app/commerce-placements-engine/commerce-placements-engine';
+import {
+  getBadge,
+  getRecs,
+} from '../features/placement-set/placement-set-action';
+import {getOrganizationEndpoints} from '../recommendation.index';
+
+describe('commerce placements', () => {
```
olamothe, line 9: Some simple end to end tests just to verify that the boilerplate is working.

As explained in the channel however, I am simply unable to perform a request with the "correct values" that would actually return recommendations or badge: the request is not error-ing, but there's no products or badges that ever get returned.

I assume it's due to incorrect parameters, but so far, I haven't been able to "guess" the correct combination that would return something meaningful.

packages/headless/src/api/commerce/unified-api/unified-api-client.ts
```diff
@@ -0,0 +1,100 @@
...
+      return buildAPIResponseFromErrorOrThrow(response);
+    }
+    const body = await response.json();
+
+    return response.ok
+      ? {success: body as Recommendations}
+      : {error: body as UnifiedCommerceAPIError};
+  }
+
+  async getBadges(
```
samisayegh, line 68: Does this endpoint return multiple badges? Or only one?

packages/headless/src/api/commerce/unified-api/unified-api-error-response.ts
```diff
@@ -0,0 +1,43 @@
+import {DisconnectedError} from '../../../utils/errors';
+
+export interface UnifiedCommerceAPIError {
+  errorCode: number;
+  message: string;
+  ignored?: boolean;
```
samisayegh, line 6: What does `ignored` do?

packages/headless/src/api/platform-client-error-response.ts
```diff
@@ -0,0 +1,46 @@
+import {DisconnectedError} from '../utils/errors';
+
+export interface PlatformAPIErrorWithStatusCode {
+  statusCode: number;
+  message: string;
+  type: string;
```
samisayegh, line 6: There's a lot of duplication between this file, and [this one](https://github.com/coveo/ui-kit/pull/3022/files#diff-d23883d1c9e76f00ba1cd5375015ac95f75215fabd518080b290fcd9ff541463R1). Why is it needed? Is it just due to the `type` parameter?

packages/headless/src/api/commerce/unified-api/unified-api-params.ts
```diff
@@ -0,0 +1,88 @@
...
+export interface PlacementParam {
+  placementId: string;
+}
+
+export interface ImplementationParam {
+  implementationId?: string;
+}
+
+export interface ModeParam {
+  mode?: 'LIVE' | 'SAMPLE' | 'PREVIEW' | 'NONE';
```
olamothe, line 14: I think those need to be all upper case ? I vaguely remember getting an error back otherwise (ie: requesting `mode=live`)

packages/headless/src/features/placement-set/placement-set-action.ts
```diff
@@ -0,0 +1,101 @@
...
+});
+
+async function augmentPlacementPayload<
+  Payload extends GetPlacementCreatorPayload
+>(
+  payload: Payload,
+  state: CommercePlacementsAppState
+): Promise<Payload & VisitorParam> {
+  const visitorID = await getVisitorID(state.configuration.analytics);
+  return {...payload, visitor: {id: visitorID}};
```
Spuffynism, line 100: We will want to use `clientId` here instead of the visitor id as per https://coveo.slack.com/archives/C04NW83ADNW/p1688739000887759

packages/headless/src/features/placement-set/placement-set-interface.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface Badge {
+  id: string;
+  message: string;
+}
+
+export interface Recommendations extends Placement {
+  products: Product[];
+}
+
+export interface Product {
```
Spuffynism, line 20: Can we reuse the standard fields as defined here?
https://github.com/coveo/ui-kit/blob/4579bcef4a080d14498a1ee8c45dd10b8a307493/packages/headless/src/features/fields/fields-state.ts#L48

packages/headless/src/features/placement-set/placement-set-interface.ts
```diff
@@ -0,0 +1,40 @@
+export interface Placement {
+  callbackData: string;
+  campaignId: string;
+  visitorId: string;
```
Spuffynism, line 4:
```suggestion
  clientId: string;
```

packages/headless/src/integration-tests/commerce-placements.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      })
+    );
+
+    await engine.dispatch(
+      getRecs({
+        previewOpts: {
+          campaignId: 'CHLR7FPrRTi5bq_oQcCUcA',
+        },
+        placementId: 'kTLIHvXYTyWvuej4qJJaiA',
+        implementationId: 's1rAcVo_SmCVoSpEZ',
```
samisayegh, line 38: How did you get these values for `campaignId`, `placementId` and `implementationId`? Is it easy for an implementer to get them?

packages/headless/src/integration-tests/commerce-placements.test.ts
Spuffynism: This is probably because there isn't any badging or recs configured on the property. Maybe we can use the barca org in dev? It should have some recs and badging configured in the merch hub.

Otherwise, we might want to create a test org for the unified api. This org could be used in unified commerce api functional tests as well.

samisayegh (commented):
The PR looks good. With the pivot to focus on search, listings and Coveo recs though, I'm questioning whether it still makes to add these Qubit capabilities to headless?

packages/headless/src/integration-tests/commerce-placements.test.ts
olamothe: At the moment, it's trial and error, and no, it's not easy to get them or to know what they are for. Baby steps :D

First goal is to go without any magic, and see where we can simplify/how we could make this easier.

Probably some of these should be global engine config, not on a per request kinda thing.

packages/headless/src/api/commerce/unified-api/unified-api-client.ts
olamothe: The strict definition of the endpoint is an array of badges.

In practice however, I'm not certain if it's always an array with a single badge 🤔

packages/headless/src/api/platform-client-error-response.ts
olamothe: Yep I should refactor that.

packages/headless/src/api/commerce/unified-api/unified-api-error-response.ts
olamothe: It's an artefact of the duplication for errors definitions. It will go away when I refactor as you proposed in your other comment.

As for what it does, it's there because some errors are not really errors that an implementer need to know about/need to take action.

For example, if the end user spams the search button, there will be a lot of request that gets aborted and pollute the console. 

So some type of errors are flagged as ignored, and the logger will simply silence those.

olamothe:
> The PR looks good. With the pivot to focus on search, listings and Coveo recs though, I'm questioning whether it still makes to add these Qubit capabilities to headless?

Wasn't that the output of the meeting last week though ? 🤔 Have recommendations powered by the unified API + Headless on Barca demo by end of the month.

packages/headless/src/app/commerce-placements-engine/commerce-placements-engine.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface CommercePlacementsEngine<State extends object = {}>
+  extends CoreEngine<
+    State & CommercePlacementEngineState,
+    CommercePlacementsThunkExtraArguments
+  > {}
+
+export interface CommercePlacementsEngineOptions
+  extends ExternalEngineOptions<CommercePlacementEngineState> {}
+
+export function buildCommercePlacementsEngine(
```
fbeaudoincoveo, line 36: For the sake of uniformity with our other build engine functions (e.g., `buildRecommendationEngine`, `buildProductRecommendationEngine`), I believe "Placement" should be singular in the function name, i.e.,

`buildCommercePlacementEngine`

packages/headless/src/app/commerce-placements-engine/commerce-placements-engine.ts
```diff
@@ -0,0 +1,74 @@
...
+  options: CommercePlacementsEngineOptions
+): CommercePlacementsEngine {
```
fbeaudoincoveo, line 38: Same thing here, `CommercePlacementsEngine` -> `CommercePlacementEngine`

(you may want to do a pass to rename everything that's going to be publicly exported accordingly)

packages/headless/src/api/commerce/unified-api/unified-api-params.ts
fbeaudoincoveo: Confirmed, I got that error too

fbeaudoincoveo (approved):

packages/headless/src/features/placement-set/placement-set-interface.ts
olamothe: I'll see if can try something to tie those two types together, yeah.

olamothe: I'll be able to do something like: 

```typescript
type Product = {
  [K in typeof StandardCommerceFields[number]]: K extends
    | 'ec_images'
    | 'ec_thumbnails'
    ? string[]
    : string;
} & {
  additionalFields: string;
  childResults: Product[];
  clickUri: string;
  documentUri: string;
  permanentId: string;
  totalNumberOfChildResults: number;
};
```

packages/headless/src/api/commerce/unified-api/unified-api-client.ts
louis-bompart, line 1: What is the rationale for having the API wrapping defined in headless instead of [coveo/platform-client](github.com/coveo/platform-client) ?

olamothe: We are not using the platform client at all in Headless, for any of the services we consume.

If we were to import it today, the size of Headless would balloon up by more than 7x 😰

packages/headless/src/features/placement-set/placement-set-slice.ts
```diff
@@ -0,0 +1,81 @@
...
+        state.skus.search = action.payload.skus;
+      })
+      .addCase(setLocale, (state, action) => {
+        state.view.currency = action.payload.currency;
+        state.view.locale = action.payload.locale;
+      })
+      .addCase(setView, (state, action) => {
+        state.view.type = action.payload.type ?? '';
+        state.view.subtype = action.payload.subtype;
+        state.view.type !== 'category' ? (state.skus.plp = []) : {};
```
Spuffynism, line 49: I understand that this action cleans up skus that are not to be used with this view type.

However, can we get in a problematic state if the `setSearchSkus` is called after the `setView` action with a `payload.type` of `product` , for example (search skus should never be used with the product view type)?

fbeaudoincoveo: You're correct, this approach could lead to weird states.

One thing I could do instead would be to have actions that would let the implementer set the correct SKUs at the same time as they set the rest of the current view context.

I'd keep the "setCartSkus" and "setRecsSkus" actions, because those two are relevant regardless of the current view context.

However, I'd replace the generic "setView" action and the other "set...Skus" actions with:

- usePdpContext(sku: string)
- usePlpContext(skus: string[], subtypes?: string[])
- useSearchContext(skus: string[])
- useCartContext()
- useCheckoutContext()
- useConfirmationContext()
- useHomeContext()

Each of those actions would set the proper view AND skus.

E.g., calling `usePlpContext(['123', '456'], ['category1', 'category2'])` would set state.view.type to 'category', state.view.subtypes to ['category1', 'category2'], and state.skus.plp to ['123', '456']. It would also automatically clean up any other useless skus in the state.

I think I like this approach better, because it means the implementer has fewer actions to dispatch, which makes it less error-prone.

This makes me think I could also add a "useCustomContext" action that would allow for wild stuff we haven't considered. However, that would imply that any string would be acceptable as a view "type" value... and I'm not convinced this generic custom action would actually be of any use, so I won't add it just now.

Spuffynism: I like this! Less stuff for an implementer to trip over :)

packages/headless/src/controllers/placement-manager/headless-placement-manager.ts
```diff
@@ -0,0 +1,332 @@
...
+
+export interface PlacementManagerOptions {
+  /**
+   * The unique identifier of the implementation to request Placement content for.
+   */
+  implementationId?: string;
+  /**
+   * The product SKUs to use as seeds when requesting Placement content.
+   */
+  skus?: PlacementManagerSkusOptions;
```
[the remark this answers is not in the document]
fbeaudoincoveo: Oh right, they could potentially be any content from the index...

PR #3023, opened 2023-07-11
fix(headless): fix read-write error with facet descending order

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2583

This somehow got past all of the E2E tests as well as testing locally before merging...

Basically, I was trying to write to a property that was read-only, so that obviously causes problems 🤦


aprudhomme-coveo (approved):
So before this fix, the feature was working properly but there was a warning in the console ? Is that right ?

dmbrooke:
> So before this fix, the feature was working properly but there was a warning in the console ? Is that right ?

Nope, the whole page would get stuck loading as soon as it tried sending out the request for the descending order facet 😱 
```
TypeError: Cannot assign to read only property 'sortCriteria' of object '#<Object>'
```
This is why I'm so surprised that I was even able to merge this without getting blocked by E2E tests... Perhaps I'll do some investigation to see if there any improvements I can make on that front...

y-lakhdar (approved):

PR #3024, opened 2023-07-12
docs: add documentation for translation scripts

olamothe:
* Add some doc about how to run the translation scripts.
* Fix the translate script so that it does not crash if there is no temporary file
* Generate the missing translations in locales.json

https://coveord.atlassian.net/browse/KIT-2572


louis-bompart (approved):

utils/translation-gpt/validate.mjs
```diff
@@ -28,9 +34,8 @@ const prompt = (translations) =>
 
 async function main() {
   let allIdentifiedProblems = {};
-  writeFileSync('temporary.json', JSON.stringify(allIdentifiedProblems));
 
-  for (const translationKey of Object.keys(fileToValidate).slice(1, 20)) {
```
mrrajamanickam-coveo, line 33: Guessing it will validate all entries? Wonder if it would be possible to validate just the recently added ones to save time/resources using content in temporary.json file?

aprudhomme-coveo (approved):

y-lakhdar (approved):

PR #3026, opened 2023-07-12
feat(smart-snippet): allow link _target modifs

dguerinCoveo:
Allows the modification of the _target attribute in the <a> tags inside smart snippets and smart snippets suggestions. Uses the same mechanism as result links.

Demo:
https://github.com/coveo/ui-kit/assets/64476861/0e4b9c31-cc7a-4048-908e-3d3f88f385fa

### DIFF
--- a/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-suggestions/smart-snippet-suggestions-common.tsx
+++ b/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-suggestions/smart-snippet-suggestions-common.tsx
@@ -13,6 +13,7 @@ import {AnyBindings} from '../../interface/bindings';
 
 interface SmartSnippetSuggestionProps {
   id: string;
+  attributes?: Attr[];
   getHost: () => HTMLElement;
   getBindings: () => AnyBindings;
   getHeadingLevel: () => number;
@@ -128,6 +129,7 @@ export class SmartSnippetSuggestionCommon {
         {
           <atomic-smart-snippet-source
             source={source}
+            attributes={this.props.attributes}
             onSelectSource={() =>
               this.props
                 .getQuestionsList()
--- a/packages/atomic/src/components/search/smart-snippets/atomic-smart-snippet-suggestions/atomic-smart-snippet-suggestions.tsx
+++ b/packages/atomic/src/components/search/smart-snippets/atomic-smart-snippet-suggestions/atomic-smart-snippet-suggestions.tsx
@@ -11,6 +11,7 @@ import {
   BindStateToController,
 } from '../../../../utils/initialization-utils';
 import {randomID} from '../../../../utils/utils';
+import {getAttributesFromLinkSlot} from '../../../common/result-link/attributes-slot';
 import {SmartSnippetSuggestionCommon} from '../../../common/smart-snippets/atomic-smart-snippet-suggestions/smart-snippet-suggestions-common';
 import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
 
@@ -91,6 +92,7 @@ export class AtomicSmartSnippetSuggestions implements InitializableComponent {
 
     this.smartSnippetSuggestionListCommon = new SmartSnippetSuggestionCommon({
       id: this.id,
+      attributes: getAttributesFromLinkSlot(this.host, 'attributes'),
       getHost: () => this.host,
       getBindings: () => this.bindings,
       getHeadingLevel: () => this.headingLevel,
--- a/packages/atomic/src/pages/examples/ipx.html
+++ b/packages/atomic/src/pages/examples/ipx.html
@@ -174,6 +174,17 @@
         top: 10px;
         left: 5px;
       }
+      atomic-did-you-mean,
+      atomic-notifications,
+      atomic-smart-snippet,
+      atomic-smart-snippet-suggestions {
+        padding-bottom: 1rem;
+      }
+
+      atomic-smart-snippet::part(source-url),
+      atomic-smart-snippet-suggestions::part(source-url) {
+        word-wrap: break-word;
+      }
     </style>
   </head>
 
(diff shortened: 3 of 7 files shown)


packages/atomic/src/components/common/result-link/attributes-slot.ts
```diff
@@ -5,16 +5,41 @@ export function getAttributesFromLinkSlot(host: HTMLElement, slotName: string) {
   if (!attributesSlot) {
     return;
   }
+  return extractAttributeFromLink(attributesSlot, slotName);
+}
+
+export function getLinkAttributesFromSlotChild(
```
dguerinCoveo, line 11: Unfortunately, Stencil does not recognize the HTMLSlotElement during tests so this method cannot be unit tested properly. I added a cypress test though
https://github.com/ionic-team/stencil/issues/2830

packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-suggestions/smart-snippet-suggestions-common.tsx
```diff
@@ -143,7 +144,9 @@ export class SmartSnippetSuggestionCommon {
                 .getQuestionsList()
                 .cancelPendingSelectSource(relatedQuestion.questionAnswerId)
             }
-          ></atomic-smart-snippet-source>
+          >
+            <slot name="attributes" slot="source-attributes"></slot>
```
btaillon, line 148: I now realize that in the example I gave you, it looks like I reversed this.

I meant for the slot in `atomic-smart-snippet` to be named "source-attributes" and the one in `atomic-smart-snippet-source` to be named "attributes".

So this:
```suggestion
            <slot name="source-attributes" slot="attributes"></slot>
```

dguerinCoveo: Changed!

packages/atomic/src/components/common/result-link/attributes-slot.ts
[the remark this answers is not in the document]
dguerinCoveo: Yeah I still need them. Its because the slot is forwarded from `atomic-smart-snippet` to `atomic-smart-snippet-source`. So the actual `<a>` tag is still a children of `atomic-smart-snippet` even if the slot becomes the child of `atomic-smart-snippet-source`. That is why we need to use the `slot.assignedElements()` method to fetch its children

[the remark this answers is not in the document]
dguerinCoveo: Maybe the first iteration was the best? To just forward the attributes through `atomic-smart-snippet-source`?

louis-bompart (approved):

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -231,8 +242,13 @@
...
           <atomic-layout-section section="results">
-            <atomic-smart-snippet></atomic-smart-snippet>
-            <atomic-smart-snippet-suggestions></atomic-smart-snippet-suggestions>
+            <atomic-generated-answer></atomic-generated-answer>
+            <atomic-smart-snippet>
+              <a slot="attributes" target="_blank"></a>
+            </atomic-smart-snippet>
+            <atomic-smart-snippet-suggestions>
+              <a slot="attributes" target="_blank"></a>
+            </atomic-smart-snippet-suggestions>
```
btaillon, line 251: Don't forget to add [noopener](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener) to avoid potential vulnerabilities in old browsers (just in case).
```suggestion
            <atomic-generated-answer></atomic-generated-answer>
            <atomic-smart-snippet>
              <a slot="attributes" target="_blank" rel="noopener"></a>
            </atomic-smart-snippet>
            <atomic-smart-snippet-suggestions>
              <a slot="attributes" target="_blank" rel="noopener"></a>
            </atomic-smart-snippet-suggestions>
```

packages/atomic/src/components/search/smart-snippets/atomic-smart-snippet-suggestions/atomic-smart-snippet-suggestions.tsx
```diff
@@ -91,6 +92,7 @@ export class AtomicSmartSnippetSuggestions implements InitializableComponent {
 
     this.smartSnippetSuggestionListCommon = new SmartSnippetSuggestionCommon({
       id: this.id,
+      attributes: getAttributesFromLinkSlot(this.host, 'attributes'),
```
btaillon, line 95: I would prefer if this was a getter instead, to avoid race conditions.

For example:
```suggestion
      getAttributes: () => getAttributesFromLinkSlot(this.host, 'attributes'),
```

btaillon: Also, I would prefer if we used a different name than `attributes`, since the `atomic-smart-snippet`/`atomic-smart-snippet-suggestions` contain a lot of things that technically have attributes.

I think a name like "source-anchor-attributes" would work.

btaillon: Lastly, the jsdoc of the smart snippet and smart snippet suggestions should be updated to contain the `@slot` with a description. You can copy the one from `atomic-result-link` and adapt it, then tag `@srosenberg-coveo` for review.

packages/atomic/src/components/search/smart-snippets/atomic-smart-snippet-suggestions/atomic-smart-snippet-suggestions.tsx
```diff
@@ -30,6 +31,9 @@ import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
  * </atomic-smart-snippet-suggestions>
  * ```
  *
+ * @slot source-anchor-attributes - Lets you pass [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) down to anchor elements, overriding other attributes.
+ * To be used exclusively in anchor elements, such as: `<a slot="source-anchor-attributes" target="_blank"></a>`.
```
dguerinCoveo, line 35: @srosenberg-coveo what do you think of that jsdoc entry?

mikegron (approved):

dguerinCoveo:
@nathanlb should I update the atomic-insight-smart-snippet* component as well while I'm there?

btaillon (approved):

PR #3027, opened 2023-07-12
fix: unsubscribe component on disconnected callback

olamothe:
This was essentially caused by a closure/scope issue. 

When you look at what happens internally in redux when the `unsubscribe` function is called, there's a closure flag (`isSubscribed`) that changes when unsubscribe is called for the first time. Any subsequent call to the same function will be a no-op that returns immediately.

We need to make sure we are referencing different instances of the unsubscribe function when a component gets removed from the page. A simple solution is to simply tack it on the component instance.

https://coveord.atlassian.net/browse/KIT-2569


aprudhomme-coveo (approved):
lgtm !

PR #3028, opened 2023-07-12
feat(atomic, headless): add new component that builds automatic facets

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2588

This PR moves the generation of automatic facets to a new component and extracts it from the facet manager. Originally the name for it was `automatic-facets`. I felt it was confusing when looking at the folders because of the already present `automatic-facet`. For now, I went with `automatic-facet-builder`. Keep in mind it could be used inside of the `facet-manager` and would create `automatic-facet`.

These are other name suggestions for the new component, what do you guys prefer ? :

- automatic-facets
- automatic-facet-manager
- automatic-facet-builder
- automatic-facet-generator
- automatic-facet-renderer


packages/atomic/src/components/search/atomic-facet-manager/atomic-facet-manager.tsx
```diff
@@ -100,6 +100,10 @@ export class AtomicFacetManager implements InitializableComponent {
 
     children.forEach((child) => {
       this.isPseudoFacet(child) && facets.push(child);
+
+      if (child.tagName === 'ATOMIC-AUTOMATIC-FACET-BUILDER') {
+        facets.push(child as BaseFacetElement);
+      }
     });
 
     return facets;
```
aprudhomme-coveo, line 109: Fix in the get facet to make sure that the automatic facets gets always pushed to the bottom of the facet manager.

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
```diff
@@ -0,0 +1,96 @@
...
+   */
+  @Prop() public desiredCount!: number;
+
+  /**
+   * @beta - This prop is part of the automatic facets feature.
+   * Automatic facets are currently in beta testing and should be available soon.
+   *
+   * Specifies whether the automatic facets are collapsed.
+   */
+  @Prop({reflect: true, mutable: true}) public areCollapsed = true;
```
aprudhomme-coveo, line 56: Also, this pr adds a new prop to the component called areCollapsed to control wether or not the individual automatic facets are collapsed by default or not. I added it to help with the sorting of the facet manager so that the index.html looked good. Eventually, I want to better support the automatic facets components with the facet manager feature `collapseFacetsAfter`

packages/atomic/cypress/e2e/facets/automatic-facet-builder/automatic-facet-builder.cypress.ts
```diff
@@ -0,0 +1,45 @@
...
+      {
+        shadow: () => cy.get(automaticFacetBuilderComponent).shadow(),
+      },
+      true
+    );
+  });
+
+  it('should throw an error when areCollapsed property is invalid', () => {
+    new TestFixture()
+      .with(addAutomaticFacetBuilder({'are-collapsed': 'potato'}))
```
mrrajamanickam-coveo, line 28: isn't there a test case for the positive case - when valid value is passed

packages/atomic/src/components/search/atomic-facet-manager/atomic-facet-manager.tsx
mrrajamanickam-coveo: Guess it wouldn't be possible for a child to have true for both isPseudoFacet and the tagName check?

mrrajamanickam-coveo: Wonder if this would read better with a `.filter()` instead of conditional push with forEach?

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
```diff
@@ -0,0 +1,96 @@
...
+import {Component, Prop, State, h} from '@stencil/core';
+import {
+  BindStateToController,
+  InitializableComponent,
+  InitializeBindings,
+} from '../../../../utils/initialization-utils';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+
+/**
+ * @beta - This component is part of the automatic facets feature.
```
mrrajamanickam-coveo, line 18: `@beta` should this be `@internal` ?

mrrajamanickam-coveo: Not sure if it is customary to mention "Automatic facets are currently in beta testing and should be available soon." - guess it is implied when marked

aprudhomme-coveo: @beta was the tag Oli told me to use. I already used it in other places in the code that is not present in this pr. I think it is fine as it is but I asked Sam to review the strings. @srosenberg-coveo

packages/atomic/src/components/search/atomic-facet-manager/atomic-facet-manager.tsx
aprudhomme-coveo: Yes it would be better. I didnt think about improving the function as a whole !

aprudhomme-coveo:
> These are other name suggestions for the new component, what do you guys prefer ? :

>automatic-facets
>automatic-facet-manager
>automatic-facet-builder
>automatic-facet-generator
>automatic-facet-renderer

Also @mrrajamanickam-coveo, what do you think about these names ? What is the best in your opinion ?

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
dmbrooke: Just to confirm, this prop only effects the _initial_ state of the facets? So if this prop is `true`, all child facets will be rendered as collapsed at first, but can be individually expanded after the fact?

dmbrooke: I'm still quite new to Stencil, but I don't **_think_** you need either of the `reflect` nor `mutable` options for this prop as I don't believe it is ever reassigned/changed at render-time 🤔 

Also more variable name nitpicking, I think `collapse` or `collapsed` is better. It's purpose is still clear to the dev without there being the clunkiness of it only being grammatically correct in certain contexts (e.g., if only a single automatic facet is returned by the index, it's not right to say that the facet "are collapsed")

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
```diff
@@ -0,0 +1,96 @@
...
+  public searchStatus!: SearchStatus;
+
+  /**
+   * @beta - This prop is part of the automatic facets feature.
+   * Automatic facets are currently in beta testing and should be available soon.
+   *
+   * The desired count of automatic facets.
+   * Must be a positive integer.
+   */
+  @Prop() public desiredCount!: number;
```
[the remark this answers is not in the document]
dmbrooke: Ah rats, I see that it's labelled as `desiredCount` all throughout Headless as well. Never mind then, it's more work than it's worth to rename this 😅

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
```diff
@@ -0,0 +1,96 @@
...
+      areCollapsed: new BooleanValue({default: true, required: false}),
+    }).validate({
+      desiredCount: this.desiredCount,
+      areCollapsed: this.areCollapsed,
+    });
+  }
+
+  public render() {
+    const automaticFacets =
+      this.automaticFacetBuilderState.automaticFacets?.map((facet) => {
```
dmbrooke, line 81: Just to confirm, the Search API will return `undefined` if there aren't any relevant automatic facets? I just ask because I was surprised to see that the `automaticFacets` property is nullable (hence `?.map`) when I would expect it to be an empty array if there wasn't anything to render 🙂

packages/headless/src/controllers/facets/automatic-facet-builder/headless-automatic-facet-builder.test.ts
```diff
@@ -0,0 +1,38 @@
...
+      search,
+    });
+  });
+
+  it('should dispatch #setDesiredCount', () => {
+    expect(engine.actions).toContainEqual(setDesiredCount(props.desiredCount));
+  });
+
+  it('should return automatic facets as empty array if the response is empty', () => {
+    expect(automaticFacets.state.automaticFacets).toEqual([]);
```
dmbrooke, line 36: I believe this confirms my previous suspicion that `automaticFacets` _**should not**_ be nullable then, right? It changes very little, but if the expected outcome for an empty response is an empty array (rather than nothing at all), then there's no situation (that I can think of) where `automaticFacets` could be `null`/`undefined` in the state 👀 

If I'm overlooking something though, please let me know!

packages/headless/src/controllers/facets/automatic-facet-builder/headless-automatic-facet-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export interface AutomaticFacetBuilderState {
+  /**
+   * @beta - This property is part of the automatic facets feature.
+   * Automatic facets are currently in beta testing and should be available soon.
+   *
+   * The list of automatic facet controllers.
+   */
+  automaticFacets?: AutomaticFacet[];
```
dmbrooke, line 36: If my other comments are correct, otherwise ignore 😅 
```suggestion
  automaticFacets: AutomaticFacet[];
```

packages/headless/src/controllers/facets/automatic-facet-builder/headless-automatic-facet-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+  props: AutomaticFacetBuilderProps
+): AutomaticFacetBuilder {
+  if (!loadAutomaticFacetBuilderReducers(engine)) {
+    throw loadReducerError;
+  }
+  const {dispatch} = engine;
+  dispatch(setDesiredCount(props.desiredCount));
+
+  const controller = buildController(engine);
+  const getAutomaticFacets = () =>
```
dmbrooke, line 71: Why is this a method that returns a value rather than just being the value itself? I would argue you could remove this logic and just update the `automaticFacets` assignment logic in the state getter to something like
```
const automaticFacets =
        engine.state.search.response.generateAutomaticFacets?.facets.map(
          (facet) => buildAutomaticFacet(engine, {field: facet.field})
        );
```

Or if you're not a fan of putting too much logic into a single statement, you could split it up as you're doing right now, just without making it a pointer to a method 🙂

packages/atomic/cypress/e2e/facets/automatic-facet-builder/automatic-facet-builder.cypress.ts
```diff
@@ -0,0 +1,70 @@
...
+    assertConsoleError();
+    assertContainsComponentError(
+      {
+        shadow: () => cy.get(automaticFacetBuilderComponent).shadow(),
+      },
+      true
+    );
+  });
+
+  it('should collapse the facets when areCollapsed property is `true`', () => {
```
dmbrooke, line 47: It's up for debate as to whether this is a worthwhile test, but seeing as the use-case for this prop is to collapse _**all automatic facets**_, I feel like it would be appropriate to use a `desired-count` that is greater than one. This way, we can confirm that it doesn't just collapse the top automatic facet, but all automatic facets.

This might be splitting hairs though, so I'm curious to hear others thoughts on this 🙂

dmbrooke (commented):
Primarily just have some questions and suggestions, but overall this looks great 🙂

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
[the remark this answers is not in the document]
aprudhomme-coveo: Yea, I agree the name is not the best but it is what the searchAPI chose. I would prefer to match what the searchAPI is doing. However, you are not the first person to complain about the variable name. I will notify the searchAPI team about this.

packages/headless/src/controllers/facets/automatic-facet-builder/headless-automatic-facet-builder.test.ts
aprudhomme-coveo: Yea you are right, it is better if it is not nullable everywhere and a empty array is created after the request. I changed it

packages/headless/src/controllers/facets/automatic-facet-builder/headless-automatic-facet-builder.ts
aprudhomme-coveo: I agree it is kinda wierd. I got inspired by other controllers but this is clearer this way imo. I changed it.

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
aprudhomme-coveo: Yes that is correct, I can remove those, I missed it thank you.

packages/atomic/src/components/search/atomic-facet-manager/atomic-facet-manager.tsx
```diff
@@ -95,12 +95,13 @@ export class AtomicFacetManager implements InitializableComponent {
...
     const children = Array.from(this.host.children);
 
-    children.forEach((child) => {
-      this.isPseudoFacet(child) && facets.push(child);
-    });
+    const facets = children.filter(
+      (child) =>
+        this.isPseudoFacet(child) ||
+        child.tagName === 'ATOMIC-AUTOMATIC-FACET-BUILDER'
+    ) as BaseFacetElement[];
```
louis-bompart, line 104: Array.prototype.filter is a generic function in typescript, meaning  you could do:
```suggestion
    const facets = children.filter<BaseFacetElement>(
      (child) =>
        this.isPseudoFacet(child) ||
        child.tagName === 'ATOMIC-AUTOMATIC-FACET-BUILDER'
    );
```

You may need to explicitly declare the return type of your predicate so that it is a typeguard func

louis-bompart (approved):

packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
```diff
@@ -0,0 +1,96 @@
...
```
dmbrooke, line 81: Now that this is no longer nullable 🙂 
```suggestion
      this.automaticFacetBuilderState.automaticFacets.map((facet) => {
```

packages/atomic/src/components/search/atomic-facet-manager/atomic-facet-manager.tsx
aprudhomme-coveo: This doenst work. There needs to be a cast somewhere.

PR #3032, opened 2023-07-13
docs: build storybook on release

olamothe:
The problem for storybook is that when you look at what's available in s3, it's on old version without recent bugfixes done in the storybook project.

I *think* it's because we did not rebuild it on release. I'll try a release to see if it fixes...

https://coveord.atlassian.net/browse/KIT-2574


btaillon:
Closed because I found that the issue is a transpilation error reproductible locally in prod build mode.

PR #3033, opened 2023-07-13
feat(headless): add url manager support for automatic facets

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2584

### DIFF
--- a/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
+++ b/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
@@ -1,6 +1,7 @@
 import {restoreSearchParameters} from '../../../features/search-parameters/search-parameter-actions';
 import {initialSearchParameterSelector} from '../../../features/search-parameters/search-parameter-selectors';
 import {buildMockSearchAppEngine, MockSearchEngine} from '../../../test';
+import {buildMockAutomaticFacetResponse} from '../../../test/mock-automatic-facet-response';
 import {buildMockCategoryFacetRequest} from '../../../test/mock-category-facet-request';
 import {buildMockCategoryFacetSlice} from '../../../test/mock-category-facet-slice';
 import {buildMockCategoryFacetValueRequest} from '../../../test/mock-category-facet-value-request';
@@ -9,6 +10,7 @@ import {buildMockDateFacetSlice} from '../../../test/mock-date-facet-slice';
 import {buildMockDateFacetValue} from '../../../test/mock-date-facet-value';
 import {buildMockFacetRequest} from '../../../test/mock-facet-request';
 import {buildMockFacetSlice} from '../../../test/mock-facet-slice';
+import {buildMockFacetValue} from '../../../test/mock-facet-value';
 import {buildMockFacetValueRequest} from '../../../test/mock-facet-value-request';
 import {buildMockNumericFacetRequest} from '../../../test/mock-numeric-facet-request';
 import {buildMockNumericFacetSlice} from '../../../test/mock-numeric-facet-slice';
@@ -276,6 +278,11 @@ describe('search parameter manager', () => {
 
     const tab = buildMockTabSlice({id: 'a', isActive: true});
     engine.state.tabSet = {a: tab};
+
+    const automaticFacetValues = [buildMockFacetValue({state: 'selected'})];
+    engine.state.automaticFacetSet.facets = {
+      a: buildMockAutomaticFacetResponse({values: automaticFacetValues}),
+    };
 
     engine.state.query.q = 'a';
     engine.state.sortCriteria = 'qre';
--- a/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
+++ b/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
@@ -133,6 +133,7 @@ export function getCoreActiveSearchParameters(
     ...getCategoryFacets(state),
     ...getNumericFacets(state),
     ...getDateFacets(state),
+    ...getAutomaticFacets(state),
   };
 }
 
@@ -189,7 +190,7 @@ function getFacets(state: Partial<SearchParametersState>) {
     return {};
   }
 
-  const f = Object.entries(state.facetSet)
+  const facets = Object.entries(state.facetSet)
     .filter(([facetId]) => state.facetOptions?.facets[facetId]?.enabled ?? true)
     .map(([facetId, {request}]) => {
       const selectedValues = getSelectedValues(request.currentValues);
@@ -197,7 +198,10 @@ function getFacets(state: Partial<SearchParametersState>) {
     })
     .reduce((acc, obj) => ({...acc, ...obj}), {});
 
-  return Object.keys(f).length ? {f} : {};
+  if (Object.keys(facets).length > 0) {
+    console.log('a');
+  }
+  return Object.keys(facets).length ? {f: facets} : {};
 }
 
 function getSelectedValues(values: FacetValueRequest[]) {
@@ -258,4 +262,18 @@ function getDateFacets(state: Partial<SearchParametersState>) {
 
 function getSelectedRanges<T extends RangeValueRequest>(ranges: T[]) {
   return ranges.filter((range) => range.state === 'selected');
+}
+
+function getAutomaticFacets(state: Partial<SearchParametersState>) {
+  if (state.automaticFacetSet?.facets === undefined) {
+    return {};
+  }
+  const af = Object.entries(state.automaticFacetSet.facets)
+    .map(([facetId, {values}]) => {
+      const selectedValues = getSelectedValues(values);
+      return selectedValues.length ? {[facetId]: selectedValues} : {};
+    })
+    .reduce((acc, obj) => ({...acc, ...obj}), {});
+
+  return Object.keys(af).length ? {af} : {};
 }
--- a/packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
+++ b/packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
@@ -1,4 +1,6 @@
 import {createReducer} from '@reduxjs/toolkit';
+import {FacetValue} from '../../../product-listing.index';
+import {restoreSearchParameters} from '../../search-parameters/search-parameter-actions';
 import {executeSearch} from '../../search/search-actions';
 import {
   deselectAllAutomaticFacetValues,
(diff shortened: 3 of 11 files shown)


packages/headless/src/features/search-parameters/search-parameter-analytics-actions.test.ts
```diff
@@ -53,51 +53,65 @@ describe('logParametersChange', () => {
...
+  });
+});
+
+function testFacetLogging(
+  parameter: string,
+  expectIdenticalActionType: (
+    action1: SearchAction,
+    action2: SearchAction
+  ) => void
+) {
```
aprudhomme-coveo, line 76: In this file, the logging of 'cf' facets was not tested so I put the 'f' tests in a function to then be able to call it with different parameters.

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
```diff
@@ -48,6 +51,39 @@ export const automaticFacetSetReducer = createReducer(
         for (const value of facet.values) {
           value.state = 'idle';
         }
+      })
+      .addCase(restoreSearchParameters, (state, action) => {
+        const af = action.payload.af ?? {};
+        const fields = Object.keys(af);
+
+        for (const field of fields) {
```
mrrajamanickam-coveo, line 59: nit: could use `for .. in` directly on `af`

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
```diff
@@ -48,6 +51,39 @@ export const automaticFacetSetReducer = createReducer(
...
+          const response = buildTemporaryAutomaticFacetResponse(field);
+          const values = af[field].map((value) =>
+            buildTemporarySelectedFacetValue(value)
+          );
+          response.values.push(...values);
+
+          state.facets[field] = response;
```
mrrajamanickam-coveo, line 66: so in this case the `field` is present in both the key and value? Is this dup necessary?

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
```diff
@@ -221,6 +223,30 @@ describe('search parameter manager', () => {
...
+      };
+
+      expect(manager.state.parameters.af).toEqual({author: ['a']});
+    });
+
+    it('when there are no facets with selected values, the #af parameter is not included', () => {
+      engine.state.automaticFacetSet.facets = {
+        author: buildMockAutomaticFacetResponse(),
+      };
+      expect('af' in manager.state.parameters).toBe(false);
```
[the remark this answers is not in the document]
aprudhomme-coveo: You are right `expect(manager.state.parameters).not.toContain('af');` is better. I will improve the file as a whole.

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.test.ts
```diff
@@ -163,4 +164,19 @@ describe('automatic-facet-set slice', () => {
       );
     });
   });
+
+  describe('#restoreSearchParameters', () => {
+    it(`when a facet is found in the #af payload,
```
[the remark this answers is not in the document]
aprudhomme-coveo: I agree I will improve every modified test file in this PR.

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
aprudhomme-coveo: Yea, the field is used as a key. I followed the same pattern as the other facet sets and I think it is fine to use it this way.

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
```diff
@@ -259,3 +260,17 @@ function getDateFacets(state: Partial<SearchParametersState>) {
...
+    return {};
+  }
+  const af = Object.entries(state.automaticFacetSet.facets)
+    .map(([facetId, {values}]) => {
+      const selectedValues = getSelectedValues(values);
+      return selectedValues.length ? {[facetId]: selectedValues} : {};
+    })
+    .reduce((acc, obj) => ({...acc, ...obj}), {});
+
+  return Object.keys(af).length ? {af} : {};
```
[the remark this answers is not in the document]
aprudhomme-coveo: right... I just copied what the other functions were doing. Thank you for noticing. I will change them all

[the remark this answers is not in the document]
aprudhomme-coveo: Actually, since the af can be undefind,  i think its best to leave it like that. I tried other methods but this is the cle

packages/headless/src/api/search/search/search-response.ts
```diff
@@ -15,7 +15,7 @@ import {SecurityIdentity} from './security-identity';
 import {PhrasesToHighlight, TermsToHighlight} from './stemming';
 
 export interface SearchResponseSuccess {
-  generateAutomaticFacets: AutomaticFacets;
+  generateAutomaticFacets?: AutomaticFacets;
```
aprudhomme-coveo, line 18: Unrelated change :  I introduced a bug last PR. this and the changes in `headless-automatic-facet-builder.ts` fixes it.

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
```diff
@@ -215,30 +217,54 @@ describe('search parameter manager', () => {
...
+    it('is not included when the parameter is undefined', () => {
       engine.state.sortCriteria = undefined as never;
-      expect('sortCriteria' in manager.state.parameters).toBe(false);
+      expect(manager.state.parameters).not.toContain('sortCriteria');
     });
   });
 
-  it(`given a certain initial state,
-  it is possible to access every relevant search parameter using #state.parameters`, () => {
+  it(`is possible to access every relevant search parameter using #state.parameters
```
mrrajamanickam-coveo, line 266: is there string interpolation happening here?
also nit: remove `,` at end

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
```diff
@@ -296,7 +327,8 @@ describe('search parameter manager', () => {
   });
 
   describe('#synchronize', () => {
-    it('given partial search parameters, it dispatches #restoreSearchParameters with non-specified parameters set to their initial values', () => {
+    it(`it dispatches #restoreSearchParameters with non-specified parameters set to their initial values
```
mrrajamanickam-coveo, line 330: same as above

mrrajamanickam-coveo, line 330: same as above

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.test.ts
```diff
@@ -163,4 +164,19 @@ describe('automatic-facet-set slice', () => {
       );
     });
   });
+
+  describe('#restoreSearchParameters', () => {
+    it(`it sets #values to the selected values in the payload
```
mrrajamanickam-coveo, line 169: string interpolation needed?

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
```diff
@@ -48,6 +51,38 @@ export const automaticFacetSetReducer = createReducer(
...
+): AutomaticFacetResponse {
+  return {
+    field,
+    values: [],
+    moreValuesAvailable: false,
+    label: '',
+    indexScore: 0,
+  };
+}
+function buildTemporarySelectedFacetValue(value: string): FacetValue {
```
mrrajamanickam-coveo, line 82: nit: new line before function def
didn't the lint/prettier catch/fix this?

mrrajamanickam-coveo (commented):
Thanks for the effort to fix many `it` descriptions. Some minor comments.

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
aprudhomme-coveo: Thanks for reviewing again, I dont think we have a rule for this. I can look into how to add a rule for this case as a technical debt.

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
aprudhomme-coveo: The string interpolation is needed for the line break. 

I prefer like this : 
```
it(`is possible to access every relevant search parameter using #state.parameters
   given a certain initial state`, () => {
```
Than this:
```
it('is possible to access every relevant search parameter using #state.parameters given a certain initial state', () => {
```

mrrajamanickam-coveo: That is weird. Are there any other instances where `` are used to break string across lines?

mrrajamanickam-coveo:
Apart from the `` nit looks good to me. But would be good to get the eyes of someone more familiar with headless like @louis-bompart

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
aprudhomme-coveo: Yes there must be other instances in our test files where we do this. It was done this way before I made my changes. If you really dont like it, I will remove them. I just think its more readable this way.

aprudhomme-coveo: I removed them all

mrrajamanickam-coveo: Thanks, I asked because it felt strange (not incorrect) and wanted to make sure if the pattern was used in the codebase already. Although the new line might improve readability, when using string interpolation the reader would expect to find a variable inside if not it might lead to some pause/wonder - which would reduce its readability.

packages/headless/src/features/search-parameters/search-parameter-analytics-actions.test.ts
louis-bompart: mhh do we wrap `it` statements in a function somewhere else in the code?
If so sure let's keep it that way. If not, maybe try describe.each?
https://jestjs.io/docs/api#describeeachtablename-fn-timeout

louis-bompart (approved):

packages/headless/src/features/search-parameters/search-parameter-analytics-actions.test.ts
aprudhomme-coveo: Yes we do it elsewhere in the code. I think it is easier to read this way than wrapping inside a describeEach

PR #3034, opened 2023-07-13
feat(atomic): added cursor effect and streamEnd event

nathanlb:
_"Fear not the large diff for t'is merely the package-lock."_
                                         - Elrond, Lord of Rivendell

## Headless changes:
- Added support for a new "endOfStream" SSE. We log a custom event when received.
- Added `isStreaming` state param used to display the blinking cursor.

## Atomic changes:
- Added blinking cursor at the end of the text while streaming
- Improvements to feedback button colors so the stroke color of the icon is attached to a variable.
  - looks better OOTB
  - more compatible with theming. (didn't look so hot in coveodocs dark theme)
- spacing, line-height, font-color changes based on feedback.

### Video Demo

https://github.com/coveo/ui-kit/assets/16785453/ab0c4adb-2eb1-4f6c-8803-ec927f402916


packages/headless/src/features/generated-answer/generated-answer-actions.ts
```diff
@@ -19,6 +20,7 @@ import {
   SearchSection,
 } from '../../state/state-sections';
 import {validatePayload} from '../../utils/validate-payload';
+import {logGeneratedAnswerStreamEnd} from './generated-answer-analytics-actions';
 import {buildStreamingRequest} from './generated-awswer-request';
```
aprudhomme-coveo, line 24: unrelated to this PR but this file has a mistake in its name.
```suggestion
import {buildStreamingRequest} from './generated-answer-request';
```

aprudhomme-coveo: Also unrelated, in this file, there is a useless import at line 3.

aprudhomme-coveo (commented):
Looks good, is there any way for the user to change the color of the blinking cursor ?

nathanlb:
@aprudhomme-coveo the cursor color is dictated by the css var `--atomic-neutral-dark` so if the user changes it then the cursor will follow. I think that should be sufficient but if it isn't then I believe the element is still accessible for targeted styling using the `generated-text` part.

packages/headless/src/features/generated-answer/generated-answer-actions.ts
nathanlb: BooleanValue? it is used. 😅 which import are you referring to?

aprudhomme-coveo: Oops sorry ! I meant line 3 in `generated-answer-slice.ts`.

dmbrooke (approved):
Looks great!

jelmedini (approved):
Looks really nice !!

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.pcss
```diff
@@ -31,14 +60,24 @@
...
+
+  .citations-label {
+    /* Making sure the label height is the same as the citation for vertical centering. */
+    height: var(--height);
+  }
+
   .citations-container {
     .citation {
-      max-width: 160px;
+      max-width: var(--max-citation-width);
```
lbergeron, line 74: Are you sure about limiting the width of citations this way? In the video, 4 of the 5 citations are labelled `How to Resolve...` making it impossible for a user to know what differs between all of them. There is still some free space on the right. The citations could be a little wider and maybe allow the user to see the difference between each of them.

packages/headless/src/features/generated-answer/generated-answer-actions.ts
```diff
@@ -115,8 +122,17 @@ export const streamAnswer = createAsyncThunk<
...
+        dispatch(
+          logGeneratedAnswerStreamEnd(
+            (JSON.parse(payload) as GeneratedAnswerEndOfStreamPayload)
+              .answerGenerated
+          )
+        );
+        break;
       default:
-        extra.logger.error(`Unknown payloadType: "${payloadType}"`);
+        extra.logger.info(`Unknown payloadType: "${payloadType}"`);
```
lbergeron, line 135: It seems to me this message should be logged as a warning. Considering there is nothing a user can do to prevent this situation from happening, maybe the message should be logged only when the debug mode is enabled.

lbergeron (approved):
It looks very good! Well done!

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.pcss
nathanlb: Yeah I'm not super confident about this being the greatest solution though I think we definitely need to truncate them as most titles are way too long. In my latest commit I made the citation element a css part so an implementer could set their own max-width. Do you have a better idea?
FYI i also increased it from 160 to 170 so what you see in the video is 160.

packages/headless/src/features/generated-answer/generated-answer-actions.ts
nathanlb: Yeah why not. Good idea

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.pcss
lbergeron: I agree with you, most titles are long and should be truncated. Would it be possible to adjust the citation width automatically to take all available width, with a maximum of 4 citations on a single row?

For instance:
- 1 citation: would take 100% width
- 2 citations: would take 50% each
- 4 citations: would take 25% each
- 5 citations: the first 4 would take 25% each, and the last one 100% on a second row (we may want to make them all the same size tough to keep them aligned)

What do you think?

nathanlb: I like the general idea! To account for mobile or IPX style interfaces we would probably also need a min-width as 25% of an iphone width is probably just a few letters.

nathanlb: Screenshot 2023-07-18 at 3 28 19 PM

So close! It would be ideal if the last one could max out at the content width instead of the container width. Might be reaching the limits of flexbox though

nathanlb: Can't get that width to cap at available space OR content-width :/ I spoke with UX and the citations in general will be looked at for improvements in the very near future so I'll hold off on modifying them for now.

lbergeron: No worries. Thanks for looking into it!

PR #3035, opened 2023-07-14
feat(headless): category facet exclusion

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2555

[Another one](https://tenor.com/view/dj-khaled-another-one-one-gif-5057861)

Very similar to my previous facet value exclusion PRs. However, I'm the least familiar with the inner-workings of category/hierarchical facets, so I went through whatever docs I could find and tried to land on something that I made sense.

I had to take a bit of a guess as to how we should handle building the facet request (highlighted by a comment below) because it would assume that the facet state would be `selected`. To address this, I decided to get the lowest active level of the facet (the last value returned by `categoryFacetRequestActiveValuesSelector`) and use this values state for the request. Not entirely sure if this will introduce new edge-cases, but I was honestly struggling a bit understanding how exactly a category should behave when you can switch back and forth between selecting and excluding whenever you want...


packages/headless/src/features/facets/facet-set/facet-set-analytics-actions-utils.ts
```diff
@@ -94,20 +94,22 @@ export const buildFacetStateMetadata = (
...
         facetRequest.facetId
-      ).length;
+      );
 
-      if (!hasSelectedValue) {
+      if (!activeValues.length) {
         return;
       }
 
+      const lowestLeaf = activeValues[activeValues.length - 1];
```
dmbrooke, line 106: Not sure if this is the right way to handle this. I try to grab the absolute lowest level of the "tree" in case one of the parent values is selected, but a child value is excluded. Not sure if this even a possibility to be honest, but I wasn't sure so I assumed it's best to try and handle it.

Curious to hear what everyone thinks on this.

btaillon:
I can see the purpose of digging down into a single category facet value (what's already possible).

However, excluding a single category facet value seems weird to me 🤔  has UX given an opinion on this?

My reflex would be to allow excluding siblings of the currently selected value (or root values if no value is selected).

dmbrooke:
> I can see the purpose of digging down into a single category facet value (what's already possible).
> 
> However, excluding a single category facet value seems weird to me 🤔 has UX given an opinion on this?
> 
> My reflex would be to allow excluding siblings of the currently selected value (or root values if no value is selected).

Yaaaaaa I agree, it does seem awfully weird... I guess the main thing is that I'm trying make it so that anything you can do for selection, you can also do for exclusion (i.e., exclusion shouldn't be held back by the need to select something first, it should be treated as its own thing). If you have a look at the interactive mockup, you can see that the exclusion feature is enabled through a menu, meaning that if we're going to follow the general practices of this mockup for all facet types in Atomic, we need to be able to handle
1. Only selected values in a facet
2. Only excluded values in a facet
3. Both selected and excluded values in a facet

I would hate to have Headless be the bottleneck for implementing certain features that are supported by the platform out-of-the-box, but you're right, it _**does**_ feels odd. I'll ask UX now to see if they have any knee-jerk reactions to this facet type. Moreover, I have a sync with them this afternoon where I can bring this up if we don't feel great with their answer 😅

PR #3039, opened 2023-07-17
docs: add descriptive text for suggested query aria label

mrrajamanickam-coveo:
* Add descriptive text used in aria labels to help distinguish between recent and suggested queries
* between parens label - Normalize parens used for zh in ; Generate entries for all supported langs
--------
* translate gpt scripts
   - move from under `utils` (which contains importable code) to `scripts` (which contains directly runnable code)
* translate gpt script
   - add EOL when writing locales file
   - log info before api is called
 * validate gpt script
   - fix to use `validation` return value from gpt
   - fix prompt to return correct validation result
        - `validation: "Explanation of why it's different from validator or true if similar"` -> `validation: "The meaning of 'recommended search query' is different from 'suggested query'"`
   - add option to validate only given keys

https://coveord.atlassian.net/browse/KIT-2590


scripts/translation-gpt/translate.mjs
```diff
@@ -1,4 +1,5 @@
 import {existsSync, openSync, readFileSync, writeFileSync} from 'fs';
+import {EOL} from 'os';
```
louis-bompart, line 2: Mh watch out, this is OS bound, meaning that the generated text would change depending on the runner OS.
I would propose to append `\n` instead

mrrajamanickam-coveo: Yes I thought making it OS bound might be a good thing? We do use it when generating commit msg e.g.
Guess we want to use EOL in cmd line utils etc but not in file content?

louis-bompart (approved):

aprudhomme-coveo (approved):

PR #3042, opened 2023-07-18
feat(atomic): empty no-result-text prop hides component when no child results found

nathanlb:
Currently when there are no result children in a template that supports them we display a message.

For the builder use-case we use a dynamic templating engine that requires us to support folded results without knowing if they'll ever exist. So we need a way to completely hide the child results section when there are none to be shown.

This justification based solely on the very niche builder use case probably isn't super convincing but I can see regular users also preferring to show nothing rather than a label saying there is nothing.

Also reduced the excessive spacing around the folded results due to result-action changes.

|  Before  | After |
| ------------- | ------------- |
| Screenshot 2023-07-20 at 2 56 52 PM | Screenshot 2023-07-20 at 2 54 48 PM |


packages/atomic/src/components/insight/result-templates/atomic-insight-result-children/atomic-insight-result-children.tsx
```diff
@@ -74,6 +74,10 @@ export class AtomicResultChildren
    * The non-localized copy for an empty result state.
    */
   @Prop() public noResultText = 'no-documents-related';
+  /**
+   * Whether to show a message when no child results are found.
+   */
+  @Prop() public hideNoResults = false;
```
btaillon, line 80: I think we should reverse this and also copy the name of the property above: `showNoResultText = true`.

Alternatively, we could remove this property and decide not to show the text when `noResultText` is an empty string. I'm fine with either solution, they both have their pros and cons.

packages/atomic/src/components/insight/result-templates/atomic-insight-result-children/atomic-insight-result-children.tsx
```diff
@@ -74,6 +74,10 @@ export class AtomicResultChildren
    * The non-localized copy for an empty result state.
    */
   @Prop() public noResultText = 'no-documents-related';
+  /**
+   * Whether to show a message when no child results are found.
+   */
```
btaillon, line 79: I think we should be more specific here and in the description of the property above. We should specify when exactly this label is shown. Does it show up before pressing "show more"? only after? etc.

packages/atomic/src/components/insight/result-templates/atomic-insight-result-children/atomic-insight-result-children.tsx
nathanlb: Since working in Quantic I tend to avoid boolean props defaulting to true. This is the reason they give:

_"Boolean attributes on standard HTML Elements are set to true by adding the attribute to the element. The absence of the attribute defaults the attribute to false. Therefore, the default value of an attribute is always false. Lightning web components use the same principle for boolean properties."_ [Source](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/js_props_boolean)

It makes sense but personally i'm not super conflicted by doing it here.

I had thought about the empty `noResultText` doing the same. I like that it doesn't add any props and it makes sense because otherwise you're just rendering useless empty space. IMO we just need to communicate it clearly in the prop documentation as it isn't exactly intuitive. The HTML might just look a little bit odd.
```
<atomic-result-children no-result-text=""></atomic-result-children>
```
If you're cool with it so am I.

btaillon: I'm cool with the empty string

packages/atomic/src/components/common/result-children/result-children-common.tsx
```diff
@@ -128,7 +128,9 @@ export class ResultChildrenCommon {
     }
 
     if (!collection.moreResultsAvailable && !this.hasChildren) {
-      return this.renderNoResult();
+      return this.props.getNoResultText()?.trim().length
```
dmbrooke, line 131: Perhaps it would be worthwhile to add a Cypress test for this behaviour? 🙂

dmbrooke (approved):
LGTM

packages/atomic/src/components/common/result-children/result-children-common.tsx
nathanlb: wise words! I see we don't have any for the initial display of this message either so i'll add a few.

drosenfeld-coveo (approved):

PR #3043, opened 2023-07-18
feat(headless): breadcrumb manager facet exclusion

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2563

Another fairly straight-forward facet exclusion PR. Most of the work on this one was for updating the UTs for all of the different types of facets to ensure their corresponding breadcrumbs were working as expected.


packages/headless/src/controllers/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
```diff
@@ -204,6 +205,10 @@ export const getBreadcrumbs = <T extends AnyFacetSetState>(
     facetId: string;
     selection: InferFacetSliceValueType<T>;
   }) => void,
+  executeToggleExclude: (payload: {
+    facetId: string;
+    selection: InferFacetSliceValueType<T>;
+  }) => void,
```
btaillon, line 211: There's starting to be quite a few properties here, I think it would be easier to read if we replaced these properties with an object/interface.

btaillon (approved):

packages/headless/src/controllers/insight/breadcrumb-manager/headless-insight-breadcrumb-manager.ts
```diff
@@ -5,24 +5,31 @@ import {logCategoryFacetBreadcrumb} from '../../../features/facets/category-face
...
+  toggleSelectNumericFacetValue,
+} from '../../../features/facets/range-facets/numeric-facet-set/numeric-facet-actions';
 import {logNumericFacetBreadcrumb} from '../../../features/facets/range-facets/numeric-facet-set/numeric-facet-insight-analytics-actions';
-import {numericFacetSelectedValuesSelector} from '../../../features/facets/range-facets/numeric-facet-set/numeric-facet-selectors';
+import {numericFacetActiveValuesSelector} from '../../../features/facets/range-facets/numeric-facet-set/numeric-facet-selectors';
 import {numericFacetSetReducer as numericFacetSet} from '../../../features/facets/range-facets/numeric-facet-set/numeric-facet-set-slice';
 import {executeSearch} from '../../../features/insight-search/insight-search-actions';
 import {searchReducer as search} from '../../../features/search/search-slice';
-import {toggleSelectStaticFilterValue} from '../../../features/static-filter-set/static-filter-set-actions';
+import {toggleExcludeStaticFilterValue, toggleSelectStaticFilterValue} from '../../../features/static-filter-set/static-filter-set-actions';
```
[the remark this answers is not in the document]
dmbrooke: Strange, I normally lint on save. This must be the result of a bulk find/replace that I missed 😅

packages/headless/src/controllers/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
dmbrooke: I mostly agree, but I messed around with this a fair amount and couldn't seem to resolve the typing everywhere. The main problem has to do with the `executeToggleSelect/Exclude` methods, or more specifically, the `selection` parameter for these methods. No matter what I changed, I couldn't seem to please the parser for these parameters. 

Also the further I got into these changes, I noticed that it didn't really "clean" things up as much as I would hope as it would still be the same amount of parameters, but now nested within an object that needed to be further typed because of generics...

I could be misunderstanding and/or incorrectly implementing what you're suggesting though, so I'd be happy to talk about this further on Slack if you really want this change?

louis-bompart: tldr of the discussion? 🍿

louis-bompart (approved):

packages/headless/src/controllers/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
dmbrooke: No discussion happened 😅 @btaillon you're welcome to chime in if you'd like (or ping me on Slack), but if it's not something you feel strongly enough about, I'll go ahead and merge hehe

btaillon: Oh, I just meant refactoring the parameters of `getBreadcrumbs` to take an object instead of ordered parameters. It shouldn't affect validation or anything like that.

dmbrooke: I'm still not certain if I love having it this way over how it was, but I move it to its own type! Let me know if this is what you had in mind 🙂

PR #3045, opened 2023-07-19
fix(atomic): Grid cell results prevent interaction with clickable children

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2597

Originally, the quickview button was un-clickable when using the grid display for `atomic-result-list`. At first, I was thinking about fixing this by simply using z-index such that the button gets rendered above the card, but this could introduce problems for implementers, so I instead went with the [W3C approach](https://design-system.w3.org/components/cards.html). This means we open links programatically, which ensures that when a child container is clicked, it doesn't also trigger the parent container `onClick` event.

One problem I had was having to add the `target` property to the `ResultListCommonProps` interface such that implementers are able to specify a `target` attribute as a part of the `atomic-result-list` tag. Implementing this was fine, but I'm not super happy about adding an attribute/property to a bunch of different components just so we're able to pass that prop all the way down to the grid result list item. Curious to hear if there are any better ways to approach this such that implementers can specify the target without having so much overhead...

To test:
1. Replace the `atomic-result-list` tag in `index.html` with the following:
```
<atomic-result-list display="grid" target="_blank">
```
2. Run the atomic dev setup
3. ???
4. Approve

### DIFF
--- a/packages/atomic/src/components/common/result-list/grid-display-results.tsx
+++ b/packages/atomic/src/components/common/result-list/grid-display-results.tsx
@@ -1,6 +1,5 @@
 import {FunctionalComponent, h} from '@stencil/core';
 import {extractUnfoldedResult} from '../interface/result';
-import {LinkWithResultAnalytics} from '../result-link/result-link';
 import {ResultListDisplayProps} from './result-list-common-interface';
 
 export const GridDisplayResults: FunctionalComponent<ResultListDisplayProps> = (
@@ -14,17 +13,12 @@ export const GridDisplayResults: FunctionalComponent<ResultListDisplayProps> = (
       <div
         part="result-list-grid-clickable-container outline"
         ref={(element) => props.setNewResultRef(element!, index)}
+        onClick={(event) => {
+          event.preventDefault();
+          interactiveResult.select();
+          location.assign(unfoldedResult.clickUri);
+        }}
       >
-        <LinkWithResultAnalytics
-          part="result-list-grid-clickable"
-          onSelect={() => interactiveResult.select()}
-          onBeginDelayedSelect={() => interactiveResult.beginDelayedSelect()}
-          onCancelPendingSelect={() => interactiveResult.cancelPendingSelect()}
-          href={unfoldedResult.clickUri}
-          title={unfoldedResult.title}
-          tabIndex={-1}
-          ariaHidden={true}
-        />
         {props.renderResult({
           key: props.getResultId(result),
           result: result,
--- a/packages/atomic/src/components/common/result-list/grid-display-results.tsx
+++ b/packages/atomic/src/components/common/result-list/grid-display-results.tsx
@@ -16,7 +16,7 @@ export const GridDisplayResults: FunctionalComponent<ResultListDisplayProps> = (
         onClick={(event) => {
           event.preventDefault();
           interactiveResult.select();
-          location.assign(unfoldedResult.clickUri);
+          window.open(unfoldedResult.clickUri, props.target);
         }}
       >
         {props.renderResult({
--- a/packages/atomic/src/components/common/result-list/styles/mixins.pcss
+++ b/packages/atomic/src/components/common/result-list/styles/mixins.pcss
@@ -150,6 +150,7 @@
         &:hover {
           border: 1px solid var(--atomic-neutral);
           box-shadow: 0px 10px 25px var(--atomic-neutral);
+          cursor: pointer;
         }
       }
     }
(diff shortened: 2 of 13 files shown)


packages/atomic/src/components/common/result-list/grid-display-results.tsx
```diff
@@ -14,17 +13,12 @@ export const GridDisplayResults: FunctionalComponent<ResultListDisplayProps> = (
       <div
         part="result-list-grid-clickable-container outline"
         ref={(element) => props.setNewResultRef(element!, index)}
+        onClick={(event) => {
+          event.preventDefault();
+          interactiveResult.select();
+          window.open(unfoldedResult.clickUri, props.target, 'noopener');
+        }}
```
btaillon, line 20: I just realized that we had an accessibility issue here, which isn't affected/fixed by these changes.

# The problem
Neither the new structure nor the old structure conveys to accessibility tools that results are clickable. For the most part, this is fine, since the template should contain another link (usually the title), which has the same purpose and conveys that they are clickable.

However, due to the whole result in grids being clickable already, some implementers may find it redundant to use `atomic-result-link` as a title, and may choose to use `atomic-result-text` instead. Some (or even most/all) accessibility tools should still be able to open the result, but they won't tell the users that any interactions are available.

# The fix
I think a good fix for this would be to add a [skip link](https://webaim.org/techniques/skipnav/) at the beginning of results.

Essentially, this would involve keeping the old `LinkWithResultAnalytics` that you removed below, but completely changing its CSS so it doesn't cover the result and instead behaves like a skip link.

You could also change the behaviour of the `onClick` event here so that it activates the link instead.

btaillon: We should make sure it's possible to open the link in a new tab with ctrl+click and mouse3 even if the `target` is `_self`.

packages/atomic/src/components/insight/result-lists/atomic-insight-result-list/atomic-insight-result-list.tsx
```diff
@@ -73,6 +74,12 @@ export class AtomicInsightResultList
    */
   @Prop({reflect: true}) imageSize: ResultDisplayImageSize = 'icon';
 
+  /**
+   * Where to open the result link.
+   * @defaultValue `_blank`
+   */
+  @Prop() target: ResultTarget = '_blank';
```
btaillon, line 81: Instead of naming this `target`, I think it should include more context (e.g., that it only works in grid cells, and that it refers to the `target` property of anchor elements). Could be something like `gridCellLinkTarget`.

Also, I think the jsdoc above should:
* Include a link to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target
* Specify that the property only works for `display="grid"`.

packages/atomic/src/components/common/result-list/styles/mixins.pcss
```diff
@@ -150,18 +150,27 @@
...
           box-shadow: 0px 10px 25px var(--atomic-neutral);
+          cursor: pointer;
         }
       }
     }
   }
 
-  [part='result-list-grid-clickable'] {
-    content: '';
+  [part='result-list-grid-clickable']:focus-visible {
```
btaillon, line 159: Due to this change, I think the final commit should be a `feat` instead of a `fix`.

Like "feat(atomic): added skip link to grid results".

packages/atomic/src/components/common/result-list/grid-display-results.tsx
```diff
@@ -14,6 +14,17 @@ export const GridDisplayResults: FunctionalComponent<ResultListDisplayProps> = (
...
+          window.open(
+            unfoldedResult.clickUri,
+            event.ctrlKey || event.metaKey
+              ? '_blank'
+              : props.gridCellLinkTarget,
+            'noopener'
+          );
+        }}
       >
         <LinkWithResultAnalytics
```
btaillon, line 29: I think this link should also contain the target and `noopener`, but it's not a strong preference.

btaillon (approved):

btaillon:
Fantastic job!

packages/atomic/src/components/common/result-list/grid-display-results.tsx
dmbrooke: I don't see why not 🙂 I'll quickly add it

mrrajamanickam-coveo (approved):

dmbrooke:
@coveo/service-core I believe I need an approval from one of you because I had to make a small change to all result lists (including the insight result list) to accommodate a new prop 😅 A review/approval would be much appreciated!

nathanlb (approved):
A suggestion for your consideration but if you prefer this I don't mind that much.

PR #3049, opened 2023-07-20
feat: add loadClickAnalyticsActions for PLP to fix interactive result

thudon-coveo:
Jira ticket: https://coveord.atlassian.net/browse/COMSDZ-605

The issue was with the `interactiveResult` was with the missing `loadClickAnalyticsActions` for product listing. We needed to align with Product Recommendation and not Search.

### DIFF
--- a/packages/headless/src/features/product-listing/product-listing-analytics.ts
+++ b/packages/headless/src/features/product-listing/product-listing-analytics.ts
@@ -46,7 +46,7 @@ const documentIdentifier = (
   }
   return {
     contentIDKey: 'permanentid',
-    contentIDValue: productRecommendation.permanentid || '',
+    contentIDValue: productRecommendation.permanentid,
   };
 };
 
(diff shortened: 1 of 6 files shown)


packages/headless/src/features/product-listing/product-listing-analytics.ts
```diff
@@ -20,89 +19,91 @@ export const logProductListing = (): ProductListingAction =>
...
+): DocumentIdentifier => {
+  if (!productRecommendation.permanentid) {
+    console.warn(
+      'Missing field permanentid on productRecommendation. This might cause many issues with your Coveo deployment. See https://docs.coveo.com/en/1913 and https://docs.coveo.com/en/1640 for more information.',
+      productRecommendation
+    );
   }
+  return {
+    contentIDKey: 'permanentid',
+    contentIDValue: productRecommendation.permanentid || '',
```
dmbrooke, line 49: I don't think you need this fallback since you already check for falsey values above

```suggestion
    contentIDValue: productRecommendation.permanentid,
```

dmbrooke (approved):
Looks good!

packages/headless/src/features/product-listing/product-listing-analytics.ts
thudon-coveo: You're right thanks !

PR #3053, opened 2023-07-25
fix(test): fix false positives from remove component assertion

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2497

I originally thought this was a problem with passing a selector, but it's actually more likely that there is a problem with Cypress itself for the `.should('not.exist')` case. Although no bug has been reported for this on Cypress' side (at least not that I've found), I've found a few threads where others are experiencing a very similar problem where `not.exist` isn't behaving as it should. After lots of research/tinkering (too much, honestly), I've decided to instead just check for a specific console error that is thrown whenever a component is placed outside of the result template and removed from the DOM. It's not ideal, but it's better than what was there before 😅 

To test:
1. Add a `assertRemovesComponent` call to a test where we know a component _**should**_ exist and ensure it fails
2. Add a `assertRemovesComponent` call to a test where we know a component _**won't**_ exist and ensure it passes

### DIFF
--- a/packages/atomic/cypress/e2e/result-list/result-components/result-badge.cypress.ts
+++ b/packages/atomic/cypress/e2e/result-list/result-components/result-badge.cypress.ts
@@ -42,6 +42,7 @@ describe('Result Badge Component', () => {
           .init();
       });
 
+      CommonAssertions.assertRemovesComponent();
       CommonAssertions.assertConsoleError();
     });
   });
--- a/packages/atomic/cypress/e2e/result-list/result-components/result-badge.cypress.ts
+++ b/packages/atomic/cypress/e2e/result-list/result-components/result-badge.cypress.ts
@@ -55,6 +55,8 @@ describe('Result Badge Component', () => {
             .with(addResultBadgeInResultList({field: 'thisfielddoesnotexist'}))
             .init();
         });
+
+        CommonAssertions.assertConsoleError();
       });
 
       describe('when the field value is a string', () => {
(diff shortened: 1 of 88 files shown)


louis-bompart (approved):

aprudhomme-coveo (approved):
Looks good ! Easy fix :)

packages/atomic/cypress/e2e/result-list/result-components/result-badge.cypress.ts
```diff
@@ -42,9 +42,6 @@ describe('Result Badge Component', () => {
           .init();
       });
 
-      CommonAssertions.assertRemovesComponent(() =>
```
mrrajamanickam-coveo, line 45: Why are the following being removed? Is it part of the "To Test" of the PR?

dmbrooke: These assertions were specifically added by @btaillon a while back to confirm that this method was giving us false positives 🙂 

You can see where he added them in the Jira description:
https://coveord.atlassian.net/browse/KIT-2497

dmbrooke:
@aprudhomme-coveo @louis-bompart @mrrajamanickam-coveo I've changed this PR a fair amount since the last set of reviews. I decided to take a "better than nothing" approach where we now check that the correct "removed from DOM" error is logged in the console where necessary as opposed to checking the DOM itself (which has proven to be extremely unreliable). You're welcome to have another look, but if there's nothing else to say, I'll merge this PR this afternoon 🙂 

Just thought I'd give the heads up!

PR #3055, opened 2023-07-25
feat(atomic): add placeholders for automatic facet

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2615

This PR adds placeholders for automatic facet. In other facet, the placeholder is in the actual facet component. Since the automatic facet component is only accessed after the query, the only way to have placeholders in this case is to show them depending on the `desiredCount` in the `automaticFacetBuilder`.

Image

### DIFF
--- a/packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
+++ b/packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
@@ -12,6 +12,7 @@ import {
   InitializableComponent,
   InitializeBindings,
 } from '../../../../utils/initialization-utils';
+import {FacetPlaceholder} from '../../../common/facets/facet-placeholder/facet-placeholder';
 import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
 
 /**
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/search/facets/atomic-automatic-facet-builder/atomic-automatic-facet-builder.tsx
```diff
@@ -91,6 +93,11 @@ export class AtomicAutomaticFacetBuilder implements InitializableComponent {
         );
       }
     );
+    if (!this.searchStatus.state.firstSearchExecuted) {
+      return Array.from({length: this.desiredCount}, () => (
+        <FacetPlaceholder numberOfValues={8} isCollapsed={this.areCollapsed} />
```
aprudhomme-coveo, line 98: This hardcoded `8` value will surely change once the search api documentation gets published. There will be ways to ask for a specific number of values. I put 8 since it is the default value for other facet requests.

louis-bompart: :+1: add a `//TODO` so we don't forget ;)

louis-bompart (approved):

PR #3059, opened 2023-07-27
folding bug fixed

mmitiche:
[SFINT-5127](https://coveord.atlassian.net/browse/SFINT-5127)

### Before

http://somup.com/c0iObJzmGp

### After

https://github.com/coveo/ui-kit/assets/86681870/237f015c-bcbe-4047-9e81-cb95abe8a8f9

### Explanation
When a new search query is made, the results are fetched again which means the state of the results is reinitialized: if a result had all his collection loaded, after the search query this loaded collection is discarded from the state and the attribute `areMoreResultsAvailable ` is reset to `true`.

That's why this attribute `areMoreResultsAvailable` needs to be taken in consideration when checking whether the no more child results message should be displayed.


SimonMilord (approved):
👌

erocheleau (approved):

PR #3062, opened 2023-07-28
fix(headless): better history support for automatic facets

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2616

### DIFF
--- a/packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
+++ b/packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
@@ -1,5 +1,6 @@
 import {createReducer} from '@reduxjs/toolkit';
 import {FacetValue} from '../../../product-listing.index';
+import {change} from '../../history/history-actions';
 import {restoreSearchParameters} from '../../search-parameters/search-parameter-actions';
 import {executeSearch} from '../../search/search-actions';
 import {
@@ -54,6 +55,7 @@ export const automaticFacetSetReducer = createReducer(
       })
       .addCase(restoreSearchParameters, (state, action) => {
         const af = action.payload.af ?? {};
+        state.set = {};
 
         for (const field in af) {
           const response = buildTemporaryAutomaticFacetResponse(field);
@@ -64,7 +66,11 @@ export const automaticFacetSetReducer = createReducer(
 
           state.set[field] = {response};
         }
-      });
+      })
+      .addCase(
+        change.fulfilled,
+        (state, action) => action.payload?.automaticFacetSet ?? state
+      );
   }
 );
 
--- a/packages/headless/src/features/history/history-slice.ts
+++ b/packages/headless/src/features/history/history-slice.ts
@@ -1,10 +1,12 @@
 import {createReducer} from '@reduxjs/toolkit';
 import {undoable} from '../../app/undoable';
 import {SortState} from '../../controllers/sort/headless-sort';
+import {FacetValue} from '../../product-listing.index';
 import {arrayEqual} from '../../utils/compare-utils';
 import {AdvancedSearchQueriesState} from '../advanced-search-queries/advanced-search-queries-state';
 import {ContextState} from '../context/context-state';
 import {DictionaryFieldContextState} from '../dictionary-field-context/dictionary-field-context-state';
+import {AutomaticFacetSetState} from '../facets/automatic-facet-set/automatic-facet-set-state';
 import {CategoryFacetSetState} from '../facets/category-facet-set/category-facet-set-state';
 import {partitionIntoParentsAndValues} from '../facets/category-facet-set/category-facet-utils';
 import {AnyFacetSetState} from '../facets/generic/interfaces/generic-facet-section';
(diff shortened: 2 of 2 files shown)


packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
```diff
@@ -56,16 +57,45 @@ export const automaticFacetSetReducer = createReducer(
...
+              );
+              const numberOfResults = facetValueObject?.numberOfResults;
+              return buildTemporarySelectedFacetValue(value, numberOfResults);
+            }),
+            ...idleValues.map(restoreFacetValueToIdleState),
+          ];
 
           state.set[field] = {response};
         }
       })
```
aprudhomme-coveo, line 87: Originally, I did not know that restoreSearchParameters would get called when going back and forward with the browser history. I only thought that restoreSearchParameters would get called when the page was reloading. I thought that the facet set would always be empty so I added bugs without knowing. This is the fix.

packages/headless/src/features/history/history-slice.ts
```diff
@@ -152,6 +155,32 @@ const isCategoryFacetsEqual = (
...
+
+    const currentSelectedValues = (
+      current.set[key].response.values as FacetValue[]
+    ).filter((value) => value.state === 'selected');
+    const nextSelectedValues = (value.response.values as FacetValue[]).filter(
+      (value) => value.state === 'selected'
+    );
+
+    if (
+      JSON.stringify(currentSelectedValues) !==
```
mrrajamanickam-coveo, line 175: Would they need to be sorted before comparing?

dmbrooke (changes requested):
Didn't do too much for code review, but I found an edge-case that should be handled. It doesn't seem to let me go back in history to undo the _final selected_ facet value. So if I have a single value selected on an automatic facet, clicking back in the browser does not clear said selected value. However, if you click a value from one facet, and then click another value from a _**different**_ facet, clicking back will unselect the latest value.

Perhaps the logic in `packages/headless/src/features/history/history-slice.ts` needs to be updated to catch this problem. But aside from this one case, everything works great!

aprudhomme-coveo:
> dn't do too much for code review, but I found an edge-case that should be handled. It doesn't seem to let me go back in history to undo the _final selected_ facet value. So if I have a single value selected on an automatic facet, clicking back in the browser does not clear said selected value. However, if you click a value from one facet, and then click another value from a _**different**_ facet, clicking back will unselect the latest value.
> 
> Perhaps the logic in `packages/headless/src/features/history/history-slice.ts` needs to be updated to catch this problem. But aside from this one case, everything works great!

@dmbrooke @mrrajamanickam-coveo hey guys, I completely changed the logic in automatic facet set slice to make the history work. I dont know why I pushed this while it was clearly not working lol... I also added comments before the two complicated loops to help understand to goal of them.

Basically, there is two slightly different use case where the restoreSearchParameters action is called : 
1. When the page is reloaded (there is not set of automatic facets no request has been been done)
2. When you undo/redo the page (there is a facet set because a previous request has been done)

Before this PR, the code was only covering the first case. The new logic with the 2 loops solve both cases.

packages/headless/src/features/history/history-slice.ts
aprudhomme-coveo: I dont think so. Even if I sorted them in some way,  they could never be the same. Doing it this way is fine and is how all the other facets do it. Basically, if the set is different that means a query was done and we to trigger the snapshot action.

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
```diff
@@ -55,7 +56,21 @@ export const automaticFacetSetReducer = createReducer(
...
       .addCase(restoreSearchParameters, (state, action) => {
         const af = action.payload.af ?? {};
+        const currentFields = Object.keys(state.set);
 
+        //unselected facets in set that are not in af
+        for (let i = 0; i < currentFields.length; i++) {
```
dmbrooke, line 62:
```suggestion
        for (const field of currentFields) {
```

packages/headless/src/features/history/history-slice.ts
```diff
@@ -152,6 +155,32 @@ const isCategoryFacetsEqual = (
...
+  next: AutomaticFacetSetState
+) => {
+  for (const [key, value] of Object.entries(next.set)) {
+    if (!current.set[key]) {
+      return false;
+    }
+
+    const currentSelectedValues = (
+      current.set[key].response.values as FacetValue[]
+    ).filter((value) => value.state === 'selected');
```
dmbrooke, line 169: Automatic facets will likely support facet exclusion in the future, so might as well save ourselves the hassle now 😅 

```suggestion
    ).filter((value) => value.state !== 'idle');
```

packages/headless/src/features/history/history-slice.ts
```diff
@@ -152,6 +155,32 @@ const isCategoryFacetsEqual = (
...
+    const nextSelectedValues = (value.response.values as FacetValue[]).filter(
+      (value) => value.state === 'selected'
```
dmbrooke, line 171:
```suggestion
      (value) => value.state !== 'idle'
```

PR #3064, opened 2023-07-29
feat: scaffold initial headless sub-package for server-side rendering utils

mrrajamanickam-coveo:
* Create `@coveo/headless/ssr` sub-package containing utils for framework agnostic server-side rendering
* Add `headless-ssr` package under `examples` exercising the new sub-package with Next.js

## TODO

- [x] PR comments
  - [x] Fix broken SSR rendering 
  - [ ] ~~[Serialize/deserialize action refactor](https://github.com/coveo/ui-kit/pull/3064#discussion_r1281148592)~~
  - [x] Fix async eslint issue
  - [x] Split types into their own files
  - [ ] ~~Create a branch to merge WIP nextjs PRs~~
  - [x] Pass in result list instead of engine 
- [x] Address in-line TODO comments
- [ ] ~~Add a cypress smoke test for the example page (with `next build`)~~  Will be done in a followup PR
  - [ ] assert for console error (React errors in console on infinite loops in effects e.g.)
  - [ ] check if SSR render is working as expected - check rendered page with JS disabled 
- [ ] ~~Fix License issues reported in CI (below)~~ All identified packages have valid acceptable license (listed on their repo)

https://coveord.atlassian.net/browse/KIT-2600


packages/headless/src/app/ssr-engine/ssr-engine.ts
```diff
@@ -0,0 +1,465 @@
...
+        SearchEngineDefinition<TControllerDefinitions>['fetchInitialState']
+      >
+    ) =>
+      new Promise<
+        EngineSnapshot<
+          {type: string},
+          InferControllerSnapshotsMapFromDefinitions<TControllerDefinitions>
+        >
+        // eslint-disable-next-line no-async-promise-executor
+      >(async (resolve, reject) => {
```
louis-bompart, line 412: I think you can replace the returned promise with an async IIFE without any params, replace the resolves by returns statement, and the reject by throw statements

packages/headless/src/ssr.index.ts
```diff
@@ -0,0 +1,10 @@
+import {polyfillCryptoNode} from './api/analytics/analytics-crypto-polyfill';
+
+polyfillCryptoNode();
+
+// 3rd Party Libraries
+export type {Unsubscribe, Middleware} from '@reduxjs/toolkit';
```
louis-bompart, line 6: What's the rationale here? Why should we expose/export those 3rd party types?

packages/headless/ssr/package.json
```diff
@@ -0,0 +1,10 @@
+{
```
louis-bompart, line 1: - [ ] @louis-bompart check this later, :suspect: monorepo structure (3rd level pjson)

packages/samples/headless-ssr/package-lock.json
```diff
@@ -0,0 +1,3610 @@
+{
```
louis-bompart, line 1: You should definitely not have a lock file here.

louis-bompart (commented):
Preliminary review done, it looks good overall, I'll do a second pass later today.
The only thing I ticked on is the subpackage. I don't think a subpackage is a package per se, in the sense that it does not have a package JSON etc: https://nodejs.org/api/packages.html#subpath-exports

packages/samples/headless-ssr/app/common/engine.ts
```diff
@@ -0,0 +1,16 @@
+import {getSampleSearchEngineConfiguration} from '@coveo/headless';
```
btaillon, line 1: These files should probably not be nested under `app`, since that signifies that they are pages.

You should move `common` and `components` to the root of the project. If you want, you can add a `src` directory which contains `app`, `common` and `components`.

packages/samples/headless-ssr/app/components/search-page.tsx
```diff
@@ -0,0 +1,24 @@
...
+    useState<SearchHydrationResult | null>(null);
+
+  useEffect(() => {
+    hydrateInitialState(engineSnapshot).then(({engine, controllers}) => {
+      setHydrationResult({engine, controllers});
+    });
+  }, [engineSnapshot]);
+
+  if (hydrationResult === null) {
+    return <div>Hydrating engine ..</div>;
```
btaillon, line 19: I think we should avoid having a loading state while the engine is hydrating, since this would cause the page to temporarily disappear (AFAIK).

Instead, the approach I would recommend is to temporarily render the engine/controller state from the `engineSnapshot`.

packages/samples/headless-ssr/app/components/search-page.tsx
```diff
@@ -0,0 +1,24 @@
...
+  } else {
+    const {engine, controllers} = hydrationResult;
+    return <ResultList engine={engine} controllers={controllers}></ResultList>;
```
btaillon, line 22: For the result list, since we didn't add the `defineResultList` function yet, I think it would be easier to just give it `engine.state.search.results`, which should (IMO) be part of both the hydrated engine and the snapshot.

packages/samples/headless-ssr/README.md
```diff
@@ -0,0 +1,17 @@
+## About
+
+- This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
+- It demonstrates usage of the `@coveo/headless/ssr` utils for Server-Side Rendering with headless.
+- Although NextJS is used to demonstrate usage for convenience, the utils are not specific to NextJS/React.
```
btaillon, line 5: One caveat with using React for this sample is that there's a chance customers may see it and follow it for their React implementation, not knowing that we also offer React utilities.

I think we could just add a note about it here once we add the React utils.

packages/samples/headless-ssr/package-lock.json
btaillon: Just gotta add this project to the `workspaces` in the root `package.json`

btaillon (commented):
Looks great! I think we should change the base branch to something else than `master` though. I would prefer to avoid publishing any of the utils until we've added more of them and have more samples. The base branch could be something like `feature/ssr-utils`.

packages/samples/headless-ssr/app/components/search-page.tsx
mrrajamanickam-coveo: Changed to use suspense when loading as it would be good to demonstrate using suspense and it keeps it simpler

packages/headless/src/app/ssr-engine/ssr-engine.ts
```diff
@@ -0,0 +1,465 @@
...
+
+export interface ControllerSnapshot<TState> {
+  initialState: TState;
+}
+
+export interface ControllerSnapshotsMap {
+  [customName: string]: ControllerSnapshot<unknown>;
+}
+
+type AnyAction = {type: string};
```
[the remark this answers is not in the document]
mrrajamanickam-coveo: Done

packages/samples/headless-ssr/app/components/search-page.tsx
mrrajamanickam-coveo: This demonstrates using `buildResultList()` with the hydrated engine. Isn't that desirable? 
I wonder if demonstrating only access of the hydrated engine and not the snapshot directly would make it simpler and reduce confusion for implementers?

packages/headless/src/app/ssr-engine/ssr-engine.ts
louis-bompart: Mh true, nvm

packages/samples/headless-ssr/src/app/page.tsx
```diff
@@ -0,0 +1,13 @@
...
+import SearchPage from '@/src/components/search-page';
+import {Suspense} from 'react';
+
+// Entry point SSR function
+export default async function Search() {
+  const engineSnapshot = await fetchInitialState({controllers: {}});
+  return (
+    <Suspense fallback={<div>Hydrating engine ..</div>}>
+      <SearchPage engineSnapshot={engineSnapshot}></SearchPage>
+    </Suspense>
```
btaillon, line 11: Looks like this isn't working. When I checkout your branch, the server-side-rendered page is blank.

I believe the reason is because Suspense only works with some APIs (see the "Note" below https://react.dev/reference/react/Suspense#displaying-a-fallback-while-content-is-loading). It also works with the `use` hook, but [it's not recommended by Next.js yet](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#fetching-data-on-the-client).

btaillon: Anyway, this is a pretty React-specific concept, and I think we should demonstrate using the snapshot's state as well as the hydrated state.

packages/headless/src/app/ssr-engine/ssr-engine.ts
```diff
@@ -0,0 +1,465 @@
...
+        // eslint-disable-next-line no-async-promise-executor
+      >(async (resolve, reject) => {
+        const middleware: Middleware = () => (next) => (action) => {
+          next(action);
+          if (action.type === 'search/executeSearch/fulfilled') {
+            resolve({
+              controllers: mapObject(controllers, (controller) => ({
+                initialState: controller.state,
+              })) as InferControllerSnapshotsMapFromDefinitions<TControllerDefinitions>,
+              searchFulfilledAction: JSON.parse(JSON.stringify(action)),
```
btaillon, line 420: In my local tests, I sent the action because it's easier, but I'm not entirely sure if that's what we should send.
* Pros
   * We don't need to add an extra action and clone the handler of `executeSearch.fulfilled` in every slice.
* Cons
  * To the user, it may look hack-ish? They don't need to know we're sending an action, they just need to serialize it.
  * The user may try to mess around with the action, like replacing parts of the search response.

A possible solution would be to serialize and deserialize the action ourselves. However, if we serialize the action to a string, then Typescript won't stop the implementer from passing some unrelated string to the `hydrateInitialState` function.

Example:
```ts
export type EngineDigest<TEngine extends CoreEngine> = string;

export interface EngineDefinitionHydrateOptionsWithoutProps<TEngine extends CoreEngine> {
  digest: EngineDigest<CoreEngine>;
}
// ...
const {digest} = await engineDefinition.fetchInitialState();
return <MyPage digest={digest} />;
// ...
(props: MyPageProps) => {
  // ...
  const {engine, controllers} = await engineDefinition.hydrateInitialState({ digest: props.digest });
}
```

btaillon: I experimented with using symbols in Typescript to anonymize the type of a property, and it works, but it's finicky...
image

packages/samples/headless-ssr/app/components/search-page.tsx
mrrajamanickam-coveo: When I checked SSR was working.. must have got broken during refactoring. 
Added a TODO to create a e2e test to catch this. 
Modified to render results directly using snapshot if hydration result is null. 
But still getting a blank page. Removing loading page didn't help either.
Will investigate more tomorrow.

btaillon:
I just realized that, when I told you that you should move the engine definition out of the `app` directory, I was wrong.

With the pages router, each ts file would be a page, but not with the app router.

Additionally, since implementations may have multiple search pages with different controllers, it would probably be better to move the definitions closer to the page (like you had originally, I believe).

packages/headless/src/app/ssr-engine/ssr-engine.ts
btaillon: Closing this, since I created a separate Jira for it.

packages/headless/src/app/ssr-engine/ssr-engine.ts
```diff
@@ -0,0 +1,465 @@
+/**
```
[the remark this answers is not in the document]
btaillon: > Didn't find the need for `search-engine` types file

My concern with this one, is that we're only supporting the SearchEngine for now, but we'll eventually need to support the ProductListingEngine, and maybe more.

> The dep graph is not as clean as you have outlined - e.g. `hydrate-initial-state` types file has `import` `{EngineDefinition} from './core-engine-ssr-types';` because `EngineDefinition` depends on multiple types that cannot be moved into common. This is the only exception for now from what I see.

Is it only because of `InferHydrationResult`? If so, I think we should move that particular type to `core-engine-ssr-types`, since it's a circular dependency anyway.

Alternatively, you could replace `InferHydrationResult` to instead be:
```ts
type InferHydrationResult<
  T extends HydrateInitialStateWithProps<
    CoreEngine,
    ControllersMap,
    AnyAction,
    ControllersPropsMap
  >
> = Awaited<ReturnType<T['hydrateInitialState']>>;
```

> Since the files are nested under `ssr-engine/types/` wonder if the ..`ssr-types.ts` suffix is redundant?

Good point. I would be fine removing those suffix. Even `ssr` seems redundant in `ssr-engine.ts` honestly, since it's in the `ssr-engine` directory.

packages/headless/src/app/ssr-engine/utils.ts
```diff
@@ -0,0 +1,11 @@
+export function mapObject<TKey extends string, TInitialValue, TNewValue>(
```
btaillon, line 1: Could be moved into `headless/src/utils`

packages/samples/headless-ssr/src/components/search-page.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const [hydrationResult, setHydrationResult] =
+    useState<SearchHydrationResult | null>(null);
+
+  useEffect(() => {
+    hydrateInitialState(engineSnapshot).then(({engine, controllers}) => {
+      setHydrationResult({engine, controllers});
+    });
+  }, [engineSnapshot]);
+
+  const {engine} = hydrationResult ?? engineSnapshot;
```
btaillon, line 18: One thing I now realize is missing from the RFC, is that we don't have the engine state in the snapshot, only the controller states.

Consequently, I believe `engine` here is `undefined` in `engineSnapshot`, and that's why you can't render anything.

I think that may be fine as-is though, since including the `engine` state in the hydration result would (in Next.js) cause the same issue we discussed in the RFC: we'd send a 75kb+ payload from the server to the client that's mostly unused.

Also, if an implementer needs to obtain a part of the engine state, they can easily build their own `define` controller function which extracts the needed state, and add that to their definition.

btaillon: For now, my recommendation would be to either:
1. Do the workaround I suggested in the comment above. Create a `defineResultList` function inside the sample which just returns `{state: engine.state.search.results}`.
2. Create the `defineResultList` function in `headless/ssr` right away and reject the ticket for it in the backlog.

packages/headless/src/app/ssr-engine/ssr-engine.ts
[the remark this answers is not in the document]
mrrajamanickam-coveo: * Create `search-engine` file 
* Removed redundant suffix from file names
* Reg moving `EngineDefinition` from core-engine to common, the issue is `EngineDefinitionWithProps`, `EngineDefinitionWithoutProps` which depend on `FetchInitialStateWithoutProps`, `HydrateInitialStateWithoutProps` and `BuildWithoutProps` ... so moving all of them into common would make common very long and other files too sparse.

mrrajamanickam-coveo:
> I just realized that, when I told you that you should move the engine definition out of the `app` directory, I was wrong.
> 
> With the pages router, each ts file would be a page, but not with the app router.
> 
> Additionally, since implementations may have multiple search pages with different controllers, it would probably be better to move the definitions closer to the page (like you had originally, I believe).
> 
> I will adjust the RFC.

But the refactored layout is also readable - isn't it? Looks like nextjs is ok with either layout https://nextjs.org/docs/app/building-your-application/routing/colocation#store-project-files-outside-of-app
With folders under `app` being routable if they contain special files etc, think it is better to have non routable pages outside just to be safe.

packages/headless/src/app/ssr-engine/ssr-engine.ts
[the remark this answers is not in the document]
btaillon: > * Reg moving `EngineDefinition` from core-engine to common, the issue is `EngineDefinitionWithProps`, `EngineDefinitionWithoutProps` which depend on `FetchInitialStateWithoutProps`, `HydrateInitialStateWithoutProps` and `BuildWithoutProps` ... so moving all of them into common would make common very long and other files too sparse.

I'm not sure what you mean by this.

I think `EngineDefinition` should stay in `core-engine`, but `InferHydrationResult` should either be moved to `core-engine` or be modified as to not need `EngineDefinition` (like my example).

packages/headless/src/app/ssr-engine/types/common.ts
```diff
@@ -0,0 +1,110 @@
...
+  [customName: string]: unknown;
+}
+
+export interface ControllersMap {
+  [customName: string]: Controller;
+}
+
+export interface ControllerSnapshot<TState> {
+  state: TState;
+}
```
btaillon, line 25: One thing I updated in the RFC after all was to rename "Snapshot" to "InitialState". I think it would be more intuitive since it's returned by `fetchInitialState`. Should probably update it in this PR too

btaillon (approved):
Looks great!

Only suggestion may be to move the files in the sample around to be kinda like what you had before. I updated the examples in the React RFC to use a similar structure, you can copy that if you want.

packages/headless/src/app/ssr-engine/ssr-engine.ts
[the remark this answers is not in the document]
mrrajamanickam-coveo: `EngineDefinition` depends on `EngineDefinitionWithProps` and `EngineDefinitionWithoutProps` which in turn depend on FetchInitialStateWithoutProps, HydrateInitialStateWithoutProps and BuildWithoutProps

packages/headless/src/app/ssr-engine/types/common.ts
mrrajamanickam-coveo: Ok, done.

packages/headless/src/app/ssr-engine/ssr-engine.ts
```diff
@@ -0,0 +1,142 @@
...
+        SearchEngineDefinition<TControllerDefinitions>['fetchInitialState']
+      >
+    ) =>
+      new Promise<
+        EngineInitialState<
+          {type: string},
+          InferControllerInitialStateMapFromDefinitions<TControllerDefinitions>
+        >
+      >((resolve, reject) => {
+        let initialControllers: ControllersMap;
```
mrrajamanickam-coveo, line 85: Managed to fix the eslint no-async-promise-executor warn by refactoring to extract the controller out. Don't love this but not sure there is a better solution 🤷‍♂️.

PR #3066, opened 2023-07-31
fix(atomic): center magnifying glass icon in search box submit button

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2626


mrrajamanickam-coveo (approved):
Thanks for identifying and fixing the issue 👍 
Can you please add some description to the PR. Before/after screenshots would be helpful too.

PR #3067, opened 2023-07-31
feat(headless): send clientId instead of visitorId in case assist

nathanlb:
The customer service api now accepts `clientId` or `visitorId`.
We want to favour `clientId` to phase out the other so after this PR the case assist client will send that instead.

### DIFF
--- a/packages/headless/src/api/service/case-assist/get-case-classifications/get-case-classifications-request.ts
+++ b/packages/headless/src/api/service/case-assist/get-case-classifications/get-case-classifications-request.ts
@@ -1,9 +1,9 @@
 import {
   BaseParam,
+  ClientIDParam,
   ContextParam,
   DebugParam,
   LocaleParam,
-  VisitorIDParam,
 } from '../../../platform-service-params';
 import {
   baseCaseAssistRequest,
@@ -14,7 +14,7 @@ import {
 
 export type GetCaseClassificationsRequest = BaseParam &
   CaseAssistIdParam &
-  VisitorIDParam &
+  ClientIDParam &
   LocaleParam &
   FieldsParam &
   ContextParam &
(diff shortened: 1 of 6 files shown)


packages/headless/src/api/service/case-assist/get-case-classifications/get-case-classifications-request.ts
```diff
@@ -14,7 +14,7 @@ import {
 
 export type GetCaseClassificationsRequest = BaseParam &
   CaseAssistIdParam &
-  VisitorIDParam &
+  ClientIDParam &
```
lbergeron, line 17: Isn't removing `VisitorIDParam` from the type a breaking change? An error would be raised for the snippet below.

```typescript
buildGetCaseClassificationsRequest({ ..., visitorId: 'some-id' });
```

lbergeron (commented):
I don't think we can safely remove `VisitorIDParam` from the request types.

packages/headless/src/api/service/case-assist/get-case-classifications/get-case-classifications-request.ts
nathanlb: Given the client is not exported I don't think there can be any instance of such usage outside of Headless. At least I couldn't find any hit of it.

nathanlb: I could easily leave it in though if we do find it is breaking.

lbergeron: > Given the client is not exported I don't think there can be any instance of such usage outside of Headless.

If it is not exported by the Headless bundle, then we're good.

lbergeron (approved):
It turns out the modified request types are not exposed by Headless so removing `VisitorIDParam` from the type is not an issue.

mmitiche (approved):
LGTM!

packages/headless/src/api/service/case-assist/get-case-classifications/get-case-classifications-request.ts
dmbrooke: As far as I can tell, this is only used within Headless itself, so it should be okay 🙂

dmbrooke (approved):

PR #3069, opened 2023-08-01
fix(atomic): add automatic facets to atomic refine modal

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2623

After adding this, I tested the edge cases for the atomic refine modal where there is either no automatic facets or no normal facets and I found a bug [KIT-2627](https://coveord.atlassian.net/browse/KIT-2627) whenever there was no normal facets. I reproduced the bug in the master branch so my changes have no impact on it. I will fix it in a subsequent PR.

To open the refine modal:

1. Put browser window in mobile mode when the 'sort and filter' button appears at the right.
2. Click on it to open the modal

[KIT-2627]: https://coveord.atlassian.net/browse/KIT-2627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
+++ b/packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
@@ -9,6 +9,11 @@ import {
   buildSort,
   SortState,
   buildQuerySummary,
+  AutomaticFacetBuilder,
+  AutomaticFacetBuilderState,
+  buildAutomaticFacetBuilder,
+  buildSearchStatus,
+  SearchStatus,
 } from '@coveo/headless';
 import {
   Component,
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -209,6 +249,17 @@ export class AtomicRefineModal implements InitializableComponent {
...
+          class="flex flex-col"
+          style={{
+            gap: 'var(--atomic-refine-modal-facet-margin, 20px)',
+            marginTop: isFacetElements
+              ? 'var(--atomic-refine-modal-facet-margin, 20px)'
+              : '',
+          }}
+        >
+          {automaticFacets}
+        </div>
```
aprudhomme-coveo, line 262: This is the same style that is used for the normal facets.

[link](https://github.com/coveo/ui-kit/blob/4a0965688b83e211ed47efcd7e7c8912747f2663/packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx#L113)

louis-bompart (approved):

PR #3071, opened 2023-08-01
fix(atomic): css parts not working for automatic facets in refine modal

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2630

The issue is well documented in the jira. 

This PR also solves another found issue where parts were not working for automatic facets within the refine modal.

[KIT-2623]: https://coveord.atlassian.net/browse/KIT-2623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/common/interface/store.ts
```diff
@@ -136,7 +136,8 @@ export function createAtomicCommonStore<
     getFacetElements() {
       return stencilStore
         .get('facetElements')
-        .filter((element) => isInDocument(element));
+        .filter((element) => isInDocument(element))
+        .filter((element) => isVisible(element));
```
btaillon, line 140: Does this copy dependant facets?

E.g., if you have facet A and facet B, and facet B only shows up when facet A has a selected value, are you able to:
1. Open the refine modal
2. Select a value in facet A
3. See facet B

aprudhomme-coveo: Good catch, no it doesnt copy dependant facets. I will try to adjust my solution

packages/atomic/src/components/common/refine-modal/refine-modal-common.tsx
```diff
@@ -120,13 +120,17 @@ export function getClonedFacetElements(
...
-
-  allFacetsInOrderInDOM.forEach((facetElement, index) => {
+  let index = 0;
+  allFacetsInOrderInDOM.forEach((facetElement) => {
     const clone = facetElement.cloneNode(true) as BaseFacetElement;
-    clone.isCollapsed = facetShouldBeInitiallyCollapsed(
-      index,
-      collapseFacetsAfter
-    );
+    if (!facetElement.classList.contains('atomic-hidden')) {
```
btaillon, line 126: I think we should avoid checking whether a facet has `atomic-hidden` and/or checking whether a facet has `display: none`. Instead, I think we should expose something more explicit, like a method on the `atomic-facet` element (e.g., `isVisible()`).

btaillon: For examples on how to do methods on custom elements, see `setRenderFunction` on `atomic-result-list`.

packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -209,7 +209,8 @@ export class AtomicRefineModal implements InitializableComponent {
       return;
     }
     const amountToCollapseNext =
-      this.collapseFacetsAfter - this.bindings.store.getFacetElements().length;
+      this.collapseFacetsAfter -
+      this.bindings.store.getVisibleFacetElementsLength();
     const automaticFacets = isAutomaticFacets
       ? this.automaticFacetBuilderState.automaticFacets.map((facet, index) => {
           const isCollapsed = index >= amountToCollapseNext;
```
btaillon, line 216: Out of scope for this PR, but I haven't noticed until now. Is it possible currently to style automatic facet parts?

For example, it's currently possible with CSS to change the padding of facet values, even when they're in a refine modal, by using `::part(SOMETHING)`. Does this also work with automatic facets?

Part of why I'm asking, is because I vaguely recall now that this is the reason why the regular facets render a `slot` element.

aprudhomme-coveo: it is because of the shadow dom ! I am working on a better implementation

aprudhomme-coveo:
@btaillon I found a pretty creative solution to make the `<slot>` pattern work by making a stencil component and appending it to the atomic-refine-modal like we already append the slotted div for the static facets.

 Also, instead of creating a `isVisible` function in atomic-facet, category-facet, numeric-facet and more. I just used this `window.getComputedStyle(facetElement).display !== 'none'` instead of checking the `'atomic-hidden'`. 

Let me know what you think, the props now work correctly and the modal collapses the right amount of visible facets !!

packages/atomic/src/utils/utils.ts
```diff
@@ -186,6 +186,11 @@ export function isInDocument(element: Node) {
   return false;
 }
 
+export function isVisible(element: Node) {
+  const style = window.getComputedStyle(element as HTMLElement);
+  return style.display !== 'none';
+}
```
btaillon, line 192: It's not just checking for the `atomic-hidden` class that I'd like to avoid, I'd like to avoid anything that isn't explicitly provided by the facets.

My worry is that future changes may hide the facet in a different way or an implementer may add their own styling which would affect this function.

packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -249,21 +216,27 @@ export class AtomicRefineModal implements InitializableComponent {
...
+  private getAutomaticFacetSlotContent() {
+    const isFacetElements =
+      this.bindings.store.getVisibleFacetElementsLength() > 0;
+    const amountToCollapseNext =
+      this.collapseFacetsAfter -
+      this.bindings.store.getVisibleFacetElementsLength();
+
+    const divSlot = document.createElement(
+      'atomic-automatic-facet-slot-content'
+    );
```
btaillon, line 233: Neat solution!

I wonder, would it be possible to use the normal automatic facet builder instead?

packages/atomic/src/utils/utils.ts
aprudhomme-coveo: Okay, I am working on the isVisible Atomic method and for now, I have to add it to all of these facets.

image

Also, since public methods have to be async in stencil, it gets really hard to use them inside of stencil components. I tried multiple ways and nothing is working. 😢 

I think adding a isVisible to every component would almost be harder to remember to change then having two places where we check if the display is none :/

aprudhomme-coveo:
@btaillon I ended not doing the isVisible stuff. This PR will only solve the css part issue for automatic facets. For the collapseFacetsAfter problem, I will address it in another pr. 

I understood the underlying issue. Right now the sort function just puts the hidden ones at the bottom so they dont matter for the collapseFacetsAfter. I need to put the automatic facets in between the static ones and the hidden ones somehow but this should be much easier changes than what I was doing. Thanks for making me think a lot lol.

PR #3074, opened 2023-08-02
fix(atomic): make gridCellLinkTarget optional for insight result lists

dmbrooke:
Accidentally auto-merged before pushing my final commit 🤦


louis-bompart (approved):
🤔 mh, while it's a good idea to allow user to not have the gridCellLinkTarget, would it be a breaking change to remove the default behavior?

Essentially what is a bug:
 - The fact that we cannot change or ignore the gridCellLinkTarget?
 - The fact that by default, when we click a cell it is considered part of the link to the result?

dmbrooke:
> 🤔 mh, while it's a good idea to allow user to not have the gridCellLinkTarget, would it be a breaking change to remove the default behavior?

The files that I removed this field from are files where the `ResultDisplayLayout` is hardcoded to being `list`. Because these lists will never be displayed in `grid` format, the `gridCellLinkTarget` will never actually be used, so we should be safe to remove them 🙂

btaillon (approved):

nathanlb (approved):

PR #3077, opened 2023-08-03
feat(headless): add categoryFacet/updateBasePath action

louis-bompart:
Clean tests
Add a reducer/action to modify the basePath of category facets

KIT-2644


btaillon (approved):

packages/headless/src/features/facets/category-facet-set/category-facet-set-actions.ts
```diff
@@ -178,3 +178,24 @@ export const updateCategoryFacetSortCriterion = createAction(
...
+   */
+  basePath: string[];
+}
+
+export const updateCategoryFacetBasePath = createAction(
+  'categoryFacet/updateBasePath',
+  (payload: UpdateCategoryFacetBasePathActionCreatorPayload) =>
+    validatePayload(payload, {
+      facetId: facetIdDefinition,
+      basePath: new Value<string[]>(),
```
olamothe, line 199: I think the proper validation for this one is already done at line 95 in this file:

```typescript
new ArrayValue({required: false, each: requiredNonEmptyString}),
```

Can just extract and re-use

olamothe (approved):

packages/headless/src/features/facets/category-facet-set/category-facet-set-actions.ts
louis-bompart: From what I saw with the type it was accepting any array of string.
  But if we have a validator already for this kind of thing, I'll refactor it for sure 👍

louis-bompart: I used `ArrayValue({ each: requiredNonEmptyString}), I think the `required:false` does not make sense when you call this method and if it happens is likely an implementer error.

However, I took your advice and apply it to the other instances where we validatePayload (in this file at least)

PR #3078, opened 2023-08-03
fix(ci): update query summary result count in smoke tests

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2648


dmbrooke:
Converting this to draft because it's almost definitely a problem with the electronicscoveodemocomo0n2fu8v org itself as I was able to confirm that everything works as expected with atomic-react when instead pointing to the search-ui samples. Perhaps we should improve the query summary component to address this problem by display the number of results that are actually returned by the search API rather than what is set in the `numberOfResults` option.

btaillon:
I would prefer not to merge this, since this is a problem with a configuration in the org. Currently looking into whether I can reverse the change.

PR #3081, opened 2023-08-04
docs(atomic): fixed Storybook transpilation error with callback

btaillon:
https://coveord.atlassian.net/browse/KIT-2574

Managed to reproduce the  issue in prod builds only:
image

Explored various solutions such as upgrading dependencies, modifying `preview.js`, etc. Amongst the few solutions that worked, this ended up being the simpler solution, especially since we plan to upgrade in the future anyway.


packages/atomic/src/components/common/atomic-timeframe/atomic-timeframe.stories.tsx
```diff
@@ -14,7 +14,7 @@ const {defaultModuleExport, exportedStory} = defaultStory(
         }
       </style>
     `,
-    parentElement: () => {
+    parentElement: function () {
```
btaillon, line 17: `parentElement() {` doesn't work, and `parentElement: () => {` doesn't work either. Looks like the transpiler tried to simplify something and messed up.

olamothe (approved):

olamothe:
> Managed to reproduce the issue in prod builds only:

💯  Nice find !

PR #3082, opened 2023-08-05
ci: add cypress config for devtools and new headless

mrrajamanickam-coveo:
Add cypress config for atomic e2e tests 
* To use the [new headless mode in Chrome](https://developer.chrome.com/articles/new-headless/)
* To auto-open dev tools when running tests locally in headed mode - this should make it easier to detect errors/warnings in console
* Note: This is limited to just atomic e2e test config for now - which form majority of the e2e tests. The config can be copied over / reused (if possible) across e2e tests of all packages if desired/needed.

https://coveord.atlassian.net/browse/KIT-2576


packages/atomic/cypress/plugins/index.js
```diff
@@ -15,7 +15,27 @@
 /**
  * @type {Cypress.PluginConfig}
  */
-module.exports = (_on, _config) => {
+module.exports = (on, _config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+
+  // auto open devtools in headed mode (does not appear in screenshots, recording)
```
olamothe, line 22: Personally I'd prefer no magic with the auto open dev tool (it's kind of annoying... ?) but if other prefer it then so be it !

olamothe (approved):

PR #3083, opened 2023-08-07
chore(quantic): Updated lwc-jest to 1.4.1 & conditional rendering checks to lwc:if

SimonMilord:
[SFINT-5132](https://coveord.atlassian.net/browse/SFINT-5132)

In this PR:

- updated LWC-Jest from 1.1.2 to 1.4.1 (this is needed since it allows the new conditional rendering in LWC (ie: lwc:if) to be supported in the unit tests)
- Fixed 2 tests that were failing following the update
- Fixed the import of CSS files that were no longer found using the c namespace
- Changed the uses of if:true for the new lwc:if since Salesforce will deprecate the uses of if:true, if:false

**NOTE:**

I wanted to change the if:false uses as well, however, I couldn't find a simple way to do it. We will likely have to do it in a seperate PR since it will maybe require more changes.

According to the [documentation](https://www.apexhours.com/conditional-rendering-in-lwc-using-lwcif-and-lwcelse/#:~:text=You%20can%E2%80%99t%20precede,your%20JavaScript%20class.) we can't just use : `lwc:if={!condition}` so I could think of the 2 following solutions:

1) Write an empty lwc:if and then another tag with lwc:else with my actual elements (simple and quicker but ugly)

ie: 
`
<template lwc:if={myStatement}>
    
</template>
`

2) Create a new getter for the false condition and replace the if:false by lwc:if with the new getter condition (more complex + time consuming, cleaner)

[SFINT-5132]: https://coveord.atlassian.net/browse/SFINT-5132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
@@ -1,1 +1,1 @@
-@import 'c/quanticResultActionStyles';
+@import '../quanticResultActionStyles/quanticResultActionStyles.css';
--- a/packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
@@ -1,1 +1,1 @@
-@import '../quanticResultActionStyles/quanticResultActionStyles.css';
+@import 'c/quanticResultActionStyles';
--- a/packages/quantic/force-app/main/default/lwc/quanticResultDate/quanticResultDate.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultDate/quanticResultDate.js
@@ -8,9 +8,7 @@ import {LightningElement, api} from 'lwc';
  * The `QuanticResultDate` component displays a given result date field value.
  * @category Result Template
  * @example
- * <template lwc:if={result.raw.date}>
- *   <c-quantic-result-date result={result} label="Date" field="date"></c-quantic-result-date>
- * </template>
+ * <c-quantic-result-date result={result} label="Date" field="date"></c-quantic-result-date>
  */
 export default class QuanticResultDate extends LightningElement {
   /**
--- a/packages/quantic/force-app/main/default/lwc/quanticResultDate/quanticResultDate.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultDate/quanticResultDate.js
@@ -8,7 +8,7 @@ import {LightningElement, api} from 'lwc';
  * The `QuanticResultDate` component displays a given result date field value.
  * @category Result Template
  * @example
- * <template if:true={result.raw.date}>
+ * <template lwc:if={result.raw.date}>
  *   <c-quantic-result-date result={result} label="Date" field="date"></c-quantic-result-date>
  * </template>
  */
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.css
@@ -1,1 +1,1 @@
-@import 'c/searchBoxStyle';
+@import '../searchBoxStyle/searchBoxStyle.css';
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.css
@@ -1,1 +1,1 @@
-@import '../searchBoxStyle/searchBoxStyle.css';
+@import 'c/searchBoxStyle';
--- a/packages/quantic/force-app/main/default/lwc/quanticSmartSnippetAnswer/__tests__/quanticSmartSnippetAnswer.test.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticSmartSnippetAnswer/__tests__/quanticSmartSnippetAnswer.test.js
@@ -13,7 +13,7 @@ const selectors = {
 };
 
 const exampleAnswer =
-  '<div class="smart-snippet-answer" data-cy="smart-snippet-answer" x-test_quanticsmartsnippetanswer=""><div x-test_quanticsmartsnippetanswer=""><p x-test_quanticsmartsnippetanswer="">Example smart snippet answer</p><a href="#" x-test_quanticsmartsnippetanswer="">Example inline link</a></div></div>';
+  '<div x-test_quanticsmartsnippetanswer=""><p x-test_quanticsmartsnippetanswer="">Example smart snippet answer</p><a href="#" x-test_quanticsmartsnippetanswer="">Example inline link</a></div>';
 const exampleAnswerWithInvalidLink =
   '<div><a>Example invalid inline link</a></div>';
 const exampleActions = {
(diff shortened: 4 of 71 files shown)


packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
```diff
@@ -1 +1 @@
-@import 'c/quanticResultActionStyles';
\ No newline at end of file
+@import '../quanticResultActionStyles/quanticResultActionStyles.css';
```
SimonMilord, line 1: These 2 modules imports were not working with the c namespace anymore following the update of jest-lwc (probably due to a dependency update that came along with the lwc-jest update.

I tried including the paths in the jsconfig.json but it was not working. (I believe because we are dealing with CSS files). `FacetStyles` is imported with c namespace and working still somehow. Not sure what causes this, read the release notes and nothing jumped to me so its probably a sub-dependency such as @babel

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetAnswer/__tests__/quanticSmartSnippetAnswer.test.js
```diff
@@ -13,7 +13,7 @@ const selectors = {
 };
 
 const exampleAnswer =
-  '<div><p>Example smart snippet answer</p><a href="#">Example inline link</a></div>';
+  '<div x-test_quanticsmartsnippetanswer=""><p x-test_quanticsmartsnippetanswer="">Example smart snippet answer</p><a href="#" x-test_quanticsmartsnippetanswer="">Example inline link</a></div>';
```
SimonMilord, line 16: New LWC-Jest update now adds these x-test_componentsName properties to elements in the HTML. Therefore the innerHTML was no longer matching

packages/quantic/package.json
```diff
@@ -56,7 +56,7 @@
     "@octokit/graphql-schema": "10.74.2",
     "@salesforce/eslint-config-lwc": "3.3.3",
     "@salesforce/eslint-plugin-lightning": "1.0.0",
-    "@salesforce/sfdx-lwc-jest": "1.1.2",
+    "@salesforce/sfdx-lwc-jest": "1.4.1",
```
btaillon, line 59: This change means that the corresponding entry in the `overrides` section of the root `package.json` no longer does anything (and it probably isn't needed anymore). I think we should remove it

SimonMilord: good catch!

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
erocheleau: Have you checked that the import is still working correctly after this change?

packages/quantic/force-app/main/default/lwc/quanticResultDate/quanticResultDate.js
```diff
@@ -8,7 +8,7 @@ import {LightningElement, api} from 'lwc';
  * The `QuanticResultDate` component displays a given result date field value.
  * @category Result Template
  * @example
- * <template if:true={result.raw.date}>
+ * <template lwc:if={result.raw.date}>
```
erocheleau, line 11: Why do we have a `lwc:if` in our example of the usage of the component?

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.css
```diff
@@ -1 +1 @@
-@import 'c/searchBoxStyle';
+@import '../searchBoxStyle/searchBoxStyle.css';
```
erocheleau, line 1: Is it still importing correctly?

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetAnswer/__tests__/quanticSmartSnippetAnswer.test.js
erocheleau: Does it make more sense to change the test to check for example that there's a `div > p` with text matching `Example smart snippet answer` etc?

packages/quantic/force-app/main/default/lwc/quanticSmartSnippetAnswer/__tests__/quanticSmartSnippetAnswer.test.js
```diff
@@ -119,12 +116,11 @@ describe('c-quantic-smart-snippet-answer', () => {
...
 
       const answer = element.shadowRoot.querySelector(
-        selectors.smartSnippetAnswer
+        'div .inline-link--disabled'
       );
 
       expect(answer).not.toBeNull();
-      // eslint-disable-next-line @lwc/lwc/no-inner-html
-      expect(answer.innerHTML).toBe(expectedRenderedAnswer);
+      expect(answer).toBeTruthy();
```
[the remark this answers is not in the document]
erocheleau:
```suggestion
      expect(answer).not.toBeNull();
```
Why do we have both?

erocheleau (approved):
You can create a different jira to change the not conditions later, I think it's good for now.

Just make sure all the imports still work etc. I'm worried that changing these imports, they no longer resolve correctly in the actual components when they are executed in Salesforce, not just in the tests.

mmitiche:
Regarding the  `if:false` uses  I would go for option 2 cause having empty templates is something we really do not wanna have in all our Quantic components.
But like you mentioned this will need to be in a separate PR.

packages/quantic/force-app/main/default/lwc/quanticResultDate/quanticResultDate.js
SimonMilord: not sure why, but I'll remove it, doesn't make much sense here

SimonMilord:
> Also I see that the [Setup e2e tests on Quantic](https://github.com/coveo/ui-kit/actions/runs/5811716567/job/15755906676#logs) job is failing, is it related ti changes in this PR?

will re-run and see, I dont believe I changed anything directly tied to that 🤷

SimonMilord:
> Also I see that the [Setup e2e tests on Quantic](https://github.com/coveo/ui-kit/actions/runs/5811716567/job/15755906676#logs) job is failing, is it related ti changes in this PR?

SimonMilord:
> Regarding the `if:false` uses I would go for option 2 cause having empty templates is something we really do not wanna have in all our Quantic components. But like you mentioned this will need to be in a separate PR.

I was leaning towards option 2 also, I did create a JIRA for that.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.css
SimonMilord: yeah nope, will have to revert and figure out something else

packages/quantic/force-app/main/default/lwc/quanticResultAction/quanticResultAction.css
SimonMilord: yeah its not working in the end, I will revert it to the c namespace and try to figure out something
found a [similar issue](https://developer.salesforce.com/forums/?id=9062I000000ISjwQAG) on salesforce exchange but no resolution

mmitiche:
> > Also I see that the [Setup e2e tests on Quantic](https://github.com/coveo/ui-kit/actions/runs/5811716567/job/15755906676#logs) job is failing, is it related ti changes in this PR?
> 
> will re-run and see, I dont believe I changed anything directly tied to that 🤷

I think this is related
path

PR #3086, opened 2023-08-08
fix(atomic): address search box suggestion accessibility problem

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2645

We changed the isActive logic on search box suggestions a while ago to accommodate instant results. However, this added strange behaviour to all other query suggestions where the first suggestions gets highlighted by default as if it were being hovered. This leads to some accessibility problems

To fix this, I just reverted the `isSelected` logic and everything still seems to work as expected, only now without the awkward highlight 🤷 

To test:
1. Run atomic locally
2. Click search box on main sample page. Ensure that the top suggestion isn't already highlighted and that the suggestions still work as normal
3. Go to the fashion/instant results sample page. Click the search box and ensure that instant results appear, leveraging the top suggestion, without needing to actually highlight said suggestion.


btaillon (approved):
simple!

aprudhomme-coveo (approved):

dmbrooke:
> ie: Could we write an automated test that just focus on the search box, and check that the first suggestion is not highlighted because of some css classes ?

Good idea! I'll work on that now 👍

louis-bompart (approved):

PR #3088, opened 2023-08-08
perf(headless): remove non-selected values in automatic facet request

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2632

After talking with Jeremi Robert from the search api, he explained to me that the api only takes the selected values from the request and disregards the rest. 

This PR makes the request slimmer by removing the facets that have no selected values and removes every non-selected values from the currentValues object.


mrrajamanickam-coveo:
Are there no tests that can be added for this ?

louis-bompart (approved):

PR #3089, opened 2023-08-08
fix(atomic,headless): add maximum value of 10 for 'desiredCount'

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2635

Current Behavior:
The search goes through with the incorrect desiredCount attribute
image

Expected Behavior:
The search goes through without the generateAutomaticFacet section and there is an error that is logged.
image


packages/atomic/src/components/search/facets/atomic-automatic-facet-generator/atomic-automatic-facet-generator.tsx
```diff
@@ -70,7 +70,7 @@ export class AtomicAutomaticFacetGenerator implements InitializableComponent {
 
   private validateProps() {
     new Schema({
-      desiredCount: new NumberValue({min: 1, required: true}),
+      desiredCount: new NumberValue({min: 1, max: 10, required: true}),
```
mrrajamanickam-coveo, line 73: Is the max value coming from the API? If yes, wonder if it would be better documented if extracted out as a const with a note.

mrrajamanickam-coveo (approved):
Would be good to have a brief desc in PR (even a copy/paste from JIRA) for those who don't have access to JIRA - in the spirit of OSS and also for convenience.

packages/atomic/src/components/search/facets/atomic-automatic-facet-generator/atomic-automatic-facet-generator.tsx
```diff
@@ -70,7 +70,7 @@ export class AtomicAutomaticFacetGenerator implements InitializableComponent {
...
       areCollapsed: new BooleanValue({default: true, required: false}),
```
mrrajamanickam-coveo, line 74: Would be good to add some tests along the lines of what is described in the JIRA.

louis-bompart (approved):

btaillon (approved):

packages/atomic/src/components/search/facets/atomic-automatic-facet-generator/atomic-automatic-facet-generator.tsx
aprudhomme-coveo: Ideally, we could have a bunch of files with every default value in the search api as a const and use them in both Atomic and Headless. I guess it would be a whole new package ? it could even be something outside of ui-kit that is linked with the search api changes.

aprudhomme-coveo: I can add a some const in the files for now but it would be very helpful to have that with max, min and default values

aprudhomme-coveo: https://github.com/coveo/ui-kit/blob/7a5b21cf2af4db8864319bcc0184589c92d6f0d2/packages/headless/src/controllers/core/facets/_common/facet-option-definitions.ts#L62

PR #3090, opened 2023-08-08
feat(atomic): change default label to the field for automatic facets

aprudhomme-coveo:
https://coveord.atlassian.net/browse/KIT-2631

After the discussion with the others involved in the automatic facet generator, we came to the conclusion that it would be better to have the field instead of ‘No label’.


mrrajamanickam-coveo:
Can we add some tests ?

louis-bompart (approved):

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -389,7 +389,7 @@ export class AtomicBreadbox implements InitializableComponent {
       )
       .map(({value, facetId, field, label}) => ({
         facetId,
-        label: this.bindings.i18n.t(label ? label : 'no-label'),
+        label: label ? label : field,
```
mrrajamanickam-coveo, line 392: Why is i18n not required anymore here?

mrrajamanickam-coveo (approved):

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
aprudhomme-coveo: i18n was required because of the 'no-label' value. We have a entry for no-label in locales.json. Now, since both possible values from the condition come from the index, there is no need to check if there is an entry in locales.json.

PR #3093, opened 2023-08-08
test: add e2e smoke tests for headless-ssr utils

mrrajamanickam-coveo:
* Add e2e cypress smoke tests using the headless-ssr sample page
   - Add a corresponding new GH action to run these tests
* Modify sample page to include a timestamp 
   - This is then parsed by the tests to verify the page is rendered in both SSR and CSR
image

## Check before merging

- [ ] Cypress v12 is being used
- [ ] ~~Next CI cache is being used~~ Next Cache needs to be [moved to root for GH cache action to work](https://github.com/actions/setup-python/issues/476). Can revisit when next build times are significant.

https://coveord.atlassian.net/browse/KIT-2643


packages/samples/headless-ssr/cypress/e2e/ssr.cy.ts
```diff
@@ -0,0 +1,56 @@
...
+    cy.get(msgSelector).should('have.text', numResultsMsg);
+    cy.get('li').should('have.length', numResults);
+  });
+
+  it('renders result list in SSR and then in CSR', () => {
+    const interceptAlias = 'searchResults';
+    cy.intercept('/').as(interceptAlias);
+
+    cy.visit('/');
+    cy.wait(`@${interceptAlias}`).then((intercept) => {
```
louis-bompart, line 32: You should be able to use an async/await in the `it` function.

packages/samples/headless-ssr/cypress/support/commands.ts
louis-bompart, line 1: errr. what is it good for?

louis-bompart (approved):
Look nice!
Wondering how we should proceed regarding Cypress:
- use the same version as the other packages but then have another package to update when we do the update batch
- use the latest version and saves one update, but be out-of-sync with the other package, causing us to install cypress twice

packages/samples/headless-ssr/cypress/support/commands.ts
olamothe: IIRC those are auto generated by Cypress when you initialize a new project.

We can delete those indeed.

packages/samples/headless-ssr/cypress/e2e/ssr.cy.ts
```diff
@@ -0,0 +1,56 @@
...
+      );
+      expect(ssrTimestamp).to.not.be.undefined;
+      cy.get(timestampSelector).should((timeStampElement) => {
+        const csrTimestamp = Date.parse(timeStampElement.text());
+        expect(csrTimestamp).to.be.greaterThan(ssrTimestamp);
+      });
+    });
+  });
+
+  it('should pass the web-vitals audits', () => {
```
olamothe, line 48: Is this something we want to really fail tests/PR ? 

Or simply be aware of/warned of ?

olamothe (approved):

packages/samples/headless-ssr/cypress/e2e/ssr.cy.ts
btaillon: I'm not too worried about having to adjust this over time.

btaillon (approved):

packages/samples/headless-ssr/cypress/e2e/ssr.cy.ts
mrrajamanickam-coveo: `cy.wait` returns a `Chainable<Interception>` and not a `Promise`. Using `await` results in the error: `Type of 'await' operand must either be a valid promise or must not contain a callable 'then' member.`

PR #3095, opened 2023-08-09
feat(quantic): retry prompt added to the quantic generated answer component

mmitiche:
[SFINT-5124](https://coveord.atlassian.net/browse/SFINT-5124)

New logic to handle the retryable error  added to the Quantic Generated Answer component.

The retryable occurs when the ML api returns a server code error `5**` or `429 Too Many Requests` code error . In that case Headless automatically tries to refetch the stream three times. if we still get the response status code errors we display a message to the user and `Try again` button.
This button will trigger a new search query and reinitiate the process of answer generation.

RetryPromptGenAI


adroletCoveo (approved):
Looks good to me! Thanks Mehdi

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -146,10 +177,17 @@ export default class QuanticGeneratedAnswer extends LightningElement {
     }`;
   }
 
+  get hasRetryableError() {
+    return !this?.searchStatusState?.hasError && this.state?.error?.isRetryable;
```
SimonMilord, line 181: could you explain this line? Not sure I get it. If the searchStatusState has no errors and the state has a retryable error then we display the try again button? I understand that if there is a searchStatusState error then we can't try again and the user has to try something else? If so how do we handle such situation?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -137,7 +163,12 @@ export default class QuanticGeneratedAnswer extends LightningElement {
   }
 
   get shouldDisplayGeneratedAnswer() {
-    return !!this.answer || this.isStreaming || !!this.citations?.length;
+    return (
+      !!this.answer ||
+      this.isStreaming ||
+      this.citations?.length > 0 ||
+      this.hasRetryableError
+    );
```
erocheleau, line 171: This is quickly becoming really complex xD

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: Good question, I have the same :)

mmitiche: yeah sure!
The first part of the condition `this?.searchStatusState?.hasError` is to check if there is an error with the Search Query, if it has the value `true`  the Search interface will display the Quantic Query Error  component which will guide the end user to the problem or/and how to fix it.  
So if `this?.searchStatusState?.hasError` has the value true, we will not even need to display the generated answer. 

The second part of the condition `this.state?.error?.isRetryable` where `this.state` represents the state of the Generated Answer Headless controller, this happens when we try get the streaming containing the generated answer from the ML api but the response returned has a code status of `5XX` or `429`.

Does it make sense?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/retryPrompt.html
erocheleau, line 1: Shouldn't this basically be a different component?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: Well in the case where `this?.searchStatusState?.hasError` is `true` is it even possible that `this.state?.error?.isRetryable` is true?

Aka could we leverage just one of the two state values to determine the same condition?

erocheleau (approved):
Some questions but overall everything looks okay.

Just maybe the retry should become its own component?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
mmitiche: Tbh I don't think that it's possible to have a situation where `this.state?.error?.isRetryable` is `true` and `his?.searchStatusState?.hasError` is `true` in the same time but the logic behind it is as follows:
if, hypothetically, we have a retryable error then we show a button that will reexecute the exact same query. If the query caused an error, then reexecuting it would be pointless that's  why are entering a retryable error state only if there was no  error with the search query.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/retryPrompt.html
mmitiche: Um don't think so, this will not make the Quantic Generated Answer component any simpler, cause even if the retry prompt is in a different component the Quantic Generated Answer will still have to display some elements based on the `hasRetryableError` condition.

if the retry prompt is a different component, what should be the logic in it? other than just an HTML displaying a text and a button

PR #3099, opened 2023-08-09
feat(headless): added defineResultList SSR controller

btaillon:
https://coveord.atlassian.net/browse/KIT-2655

What's important to pay attention to in this PR is the directory structure.

Since the SSR engine file follows the same structure as use case engines like the product listing engine, I did the same for controllers and I created a `ssr` directory in the `controllers` directory. Just like use case controllers and the ssr engine, I added ssr in the name.

### DIFF
--- a/packages/headless/src/controllers/ssr/result-list/headless-ssr-result-list.ts
+++ b/packages/headless/src/controllers/ssr/result-list/headless-ssr-result-list.ts
@@ -1,0 +1,23 @@
+import {SearchEngine} from '../../../app/search-engine/search-engine';
+import {ControllerDefinitionWithoutProps} from '../../../app/ssr-engine/types/common';
+import {
+  ResultList,
+  ResultListProps,
+  buildResultList,
+} from '../../result-list/headless-result-list';
+
+export type {
+  ResultListOptions,
+  ResultListProps,
+  ResultListState,
+  ResultList,
+} from '../../result-list/headless-result-list';
+
+/**
+ * @internal
+ */
+export const defineResultList = (
+  props?: ResultListProps
+): ControllerDefinitionWithoutProps<SearchEngine, ResultList> => ({
+  build: (engine) => buildResultList(engine, props),
+});
--- a/packages/headless/src/ssr.index.ts
+++ b/packages/headless/src/ssr.index.ts
@@ -16,3 +16,11 @@ export type {
 } from './app/ssr-engine/types/core-engine';
 
 export {defineSearchEngine} from './app/ssr-engine/ssr-engine';
+
+export type {
+  ResultList,
+  ResultListOptions,
+  ResultListProps,
+  ResultListState,
+} from './controllers/ssr/result-list/headless-ssr-result-list';
+export {defineResultList} from './controllers/ssr/result-list/headless-ssr-result-list';
--- a/packages/samples/headless-ssr/src/common/engine.ts
+++ b/packages/samples/headless-ssr/src/common/engine.ts
@@ -1,41 +1,17 @@
+import {getSampleSearchEngineConfiguration} from '@coveo/headless';
 import {
-  buildController,
-  Controller,
-  getSampleSearchEngineConfiguration,
-  Result,
-  SearchEngine,
-} from '@coveo/headless';
-import {
-  ControllerDefinitionWithoutProps,
   defineSearchEngine,
+  defineResultList,
   InferInitialState,
   InferLiveState,
 } from '@coveo/headless/ssr';
-
-// Custom controller to fetch results from snapshot
-//  as snapshot doesn't have an engine that can be accessed directly.
-function defineCustomResultList(): ControllerDefinitionWithoutProps<
-  SearchEngine,
-  Controller & {state: {results: Result[]}}
-> {
-  return {
-    build(engine: SearchEngine) {
-      return {
-        ...buildController(engine),
-        get state() {
-          return {results: engine.state.search.results};
-        },
-      };
-    },
-  };
-}
 
 const engineDefinition = defineSearchEngine({
   configuration: {
     ...getSampleSearchEngineConfiguration(),
     analytics: {enabled: false},
   },
-  controllers: {resultList: defineCustomResultList()},
+  controllers: {resultList: defineResultList()},
 });
 
 export type SearchInitialState = InferInitialState<typeof engineDefinition>;


packages/samples/headless-ssr/src/components/result-list.tsx
```diff
@@ -1,18 +1,39 @@
...
+} from '@coveo/headless';
+import {useEffect, useState, FunctionComponent} from 'react';
 
-import {Result} from '@coveo/headless';
+interface ResultListProps {
+  initialState: ResultListState;
+  controller?: ResultListController;
+}
+
+export const ResultList: FunctionComponent<ResultListProps> = (props) => {
```
btaillon, line 12: I copied most of this from the existing `result-list.fn.tsx` sample in `headless-react-samples`, except:
* I made the `controller` prop optional.
* I added the `initialState` prop.
* I removed `ResultLink` and `QuickView`.

btaillon: Changes here are necessary in order to subscribe to changes.

louis-bompart (commented):
Could we add some E2E based on what @mrrajamanickam-coveo did?

packages/samples/headless-ssr/src/components/result-list.tsx
```diff
@@ -1,18 +1,39 @@
...
+    [controller]
+  );
+
+  if (!state.results.length) {
+    return <div>No results</div>;
+  }
 
-export default function ResultList({results}: {results: Result[]}) {
   return (
     <div>
```
mrrajamanickam-coveo, line 26: The e2e tests check for number of results etc - so they needed to be added back.
I also copied the initial draft from the existing result-list.fn.tsx sample in headless-react-samples but then modified it to simplify the markup to the bare essentials necessary for illustration. 
Added the `clickUri` as an example of an interactive element.
If we were able to import the existing markup and reuse it - yes there would be advantages to keep it the same. But I do not see any advantage in trying to align the markup with the existing headless-react-samples.

packages/samples/headless-ssr/src/components/result-list.tsx
```diff
@@ -1,18 +1,39 @@
...
+  initialState: ResultListState;
+  controller?: ResultListController;
+}
+
+export const ResultList: FunctionComponent<ResultListProps> = (props) => {
+  const {initialState, controller} = props;
+  const [state, setState] = useState(initialState);
+
+  useEffect(
+    () => controller?.subscribe(() => setState({...controller.state})),
```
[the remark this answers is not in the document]
btaillon: AFAIK, normally the controller should be a dependency specifically because it's being used?

Specifically, we want the effect to be re-started whenever the `controller` pointer's value changes (e.g., when the engine finishes hydrating and it goes from `undefined` to a value).

If the `controller` changes for any reason, we also want to unsubscribe from the old controller, which is handled here by returning the unsubscribe function returned by `subscribe`.

I'll add e2e tests.

packages/samples/headless-ssr/cypress/e2e/ssr-e2e-utils.ts
```diff
@@ -0,0 +1,17 @@
+export const ConsoleAliases = {
+  error: '@consoleError',
+  warn: '@consoleWarn',
+  log: '@consoleLog',
+};
+
+export function stubConsole() {
+  cy.window().then((win) => {
+    cy.stub(win.console, 'error').as(ConsoleAliases.error.substring(1));
```
mrrajamanickam-coveo, line 9: Wouldn't `cy.spy` be better as the msgs will still appear on console?

packages/samples/headless-ssr/src/components/hydration-metadata.tsx
```diff
@@ -0,0 +1,33 @@
+import {FunctionComponent} from 'react';
+import {SearchCSRState, SearchSSRState} from '../common/engine';
+
+export interface HydrationMetadataProps {
+  initialState: SearchSSRState;
```
mrrajamanickam-coveo, line 5: Could we rename `initial` -> `SSR` here and other occurrences in this PR for consistency?

packages/samples/headless-ssr/src/components/hydration-metadata.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const HydrationMetadata: FunctionComponent<HydrationMetadataProps> = ({
+  initialState,
+  hydrationResult,
+}) => (
+  <>
+    <div>
+      Hydrated:{' '}
+      <span id="hydrated-indicator">{hydrationResult ? 'yes' : 'no'}</span>
```
mrrajamanickam-coveo, line 16: nit: guess it could be left as `{hydrationResult}` and then the test checks for `true`?

packages/samples/headless-ssr/src/components/hydration-metadata.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {SearchCSRState, SearchSSRState} from '../common/engine';
+
+export interface HydrationMetadataProps {
+  initialState: SearchSSRState;
+  hydrationResult?: SearchCSRState;
+}
+
+export const HydrationMetadata: FunctionComponent<HydrationMetadataProps> = ({
+  initialState,
+  hydrationResult,
```
mrrajamanickam-coveo, line 11: `hydrationResult` -> `hydratedResults` or `CSRResults` ?

packages/samples/headless-ssr/src/components/hydration-metadata.tsx
```diff
@@ -0,0 +1,33 @@
...
+}) => (
+  <>
+    <div>
+      Hydrated:{' '}
+      <span id="hydrated-indicator">{hydrationResult ? 'yes' : 'no'}</span>
+    </div>
+    <span id="hydrated-msg">
+      Rendered engine with{' '}
```
mrrajamanickam-coveo, line 19: terminology Nit: wouldn't it be 'Rendered page' or 'Hydrated engine' (or the verbose version 'Rendered page with hydrated engine')?

packages/samples/headless-ssr/src/components/result-list.tsx
```diff
@@ -18,28 +18,13 @@ export const ResultList: FunctionComponent<ResultListProps> = (props) => {
...
-          </li>
-        ))}
-      </ul>
-      <div>
-        Rendered on{' '}
-        <span id="timestamp" suppressHydrationWarning>
-          {new Date().toISOString()}
-        </span>
-      </div>
-    </div>
```
mrrajamanickam-coveo, line 24: Nice simplification 👍

packages/samples/headless-ssr/src/components/search-page.tsx
```diff
@@ -6,16 +6,17 @@ import {
...
 
 export default function SearchPage({
   initialState,
 }: {
   initialState: SearchSSRState;
 }) {
-  const [hydrationResult, setHydrationResult] = useState<SearchCSRState | null>(
-    null
-  );
+  const [hydrationResult, setHydrationResult] = useState<
```
mrrajamanickam-coveo, line 17: would `hydration` -> `CSR` make it more readable here (and in other places as well)?

packages/samples/headless-ssr/src/components/hydration-metadata.tsx
```diff
@@ -0,0 +1,33 @@
...
+      {
+        (hydrationResult ?? initialState).controllers.resultList.state.results
```
mrrajamanickam-coveo, line 21: Logically wouldn't the num of results make more sense to be part of result list cmp rather than hydration metadata cmp?

mrrajamanickam-coveo (commented):
Nice refactoring.

packages/samples/headless-ssr/cypress/e2e/ssr-e2e-utils.ts
btaillon: yes, good point.

packages/samples/headless-ssr/src/components/hydration-metadata.tsx
btaillon: This message is also rendered when we don't have the hydrated engine yet and only the SSR state, so I prefer not to say "Hydrated engine".

I do like "Rendered page", I'll change it to that.

olamothe (approved):

olamothe (approved):

PR #3100, opened 2023-08-09
fix(atomic): fix grid result breadcrumbs

dmbrooke:
Before, when a breadcrumb was placed inside of a grid result, clicking the ellipsis to expand the result would also click the result link. This should be fixed now though. I also added a test to confirm this.

To test:
1. Go to full commerce demo locally
2. Confirm that clicking the ellipsis on a breadcrumb doesn't also bring you to the result URL


louis-bompart (approved):
Subtle ahah. lgtm

PR #3101, opened 2023-08-10
feat(atomic, headless): add numberOfValues property in automatic facet generator

ghost:
https://coveord.atlassian.net/browse/KIT-2633 

### Context : 
The numberOfValues property determines the amount of values the facet contains. For the automatic facet generator, this attribute determines the amount of values for every of it's facets. In this example, I used `6` as the value.
image

### This PR does 3 main things :

- Adds the new property numberOfValues for the AutomaticFacetGenerator in both Atomic and Headless
- Changes the headless automatic facet generator to use the option pattern so that they can be set with the same action
- Adds new test in headless that test how the minimum, maximum and default value affect the facet generator state. (found a nice way to test [this](https://github.com/coveo/ui-kit/pull/3089) in headless-automatic-facet-generator.test.ts)

### To test :
Atomic:

image

Headless:

In the headless sample example, play with this [controller](https://github.com/coveo/ui-kit/blob/1b1782f3d70bc913cab94ae8d615292e0096859f/packages/samples/headless-react/src/pages/SearchPage.tsx#L281-L285)


louis-bompart (approved):

packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts
```diff
@@ -24,8 +24,9 @@ export const automaticFacetSetReducer = createReducer(
           state.set[response.field] = {response};
         });
       })
-      .addCase(setDesiredCount, (state, action) => {
-        state.desiredCount = action.payload;
+      .addCase(setOptions, (state, action) => {
```
olamothe, line 27: One thing we should do for this action is make each option not mandatory, and if not provided we keep the previous value in store.

ie:

```typescript
// desiredCount=3, numberOfValues=6

dispatch({desiredCount: 5})

// desiredCount=5, numberOfValues=6
```
```

olamothe (approved):

PR #3104, opened 2023-08-11
fix(atomic): automatic facets not showing in refine modal when no static facets

ghost:
https://coveord.atlassian.net/browse/KIT-2665

Current behavior:
Automatic facets don't show in refine modal unless there is at least 1 static facet.

Expected behavior:
Automatic facets should show no matter what.


packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -187,12 +187,6 @@ export class AtomicRefineModal implements InitializableComponent {
   }
 
   private renderFilters() {
-    const isFacetElements = this.bindings.store.getFacetElements().length > 0;
-
-    if (!isFacetElements) {
-      return;
-    }
-
```
aprudhomme-coveo, line 195: I think this was added just to improve performance. Everything works fine without it.

aprudhomme-coveo: ah yes, you are right ! I completely missed that 😅

packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -188,8 +188,10 @@ export class AtomicRefineModal implements InitializableComponent {
 
   private renderFilters() {
     const isFacetElements = this.bindings.store.getFacetElements().length > 0;
+    const isAutomaticFacets =
+      this.bindings.engine.state.automaticFacetSet?.set !== undefined;
 
-    if (!isFacetElements) {
+    if (!isFacetElements && !isAutomaticFacets) {
```
mrrajamanickam-coveo, line 194: Wonder why this wasn't caught by tests? Any test gaps that can be fixed?

btaillon, line 194: Don't really like the word `is` here, which usually refers to one thing.

Maybe `hasFacetElements` and `hasAutomaticFacets`?

btaillon (approved):

packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
alexprudhomme: I added corresponding tests !

PR #3105, opened 2023-08-11
feat(headless): exposes category facet values hierarchically

louis-bompart:
https://coveord.atlassian.net/browse/KIT-2668
Implement a new set of fields in the CategoryFacet state around the idea of "showing" the actual hierarchical state of the facet values, instead of "hiding/simplifying" it with `parents` and `values`

One suggestion that was made during the 🤔 process was to move the `showMore/showLess` state/concerns from the Facet to the FacetValue.

I decided against it after trying to use the resulting fields, as it was not adding any pertaining value for the current implementation of the category facet.

It could/would have value if we were to allow users to ask for more or less values on facet values other than the selected value or the whole facet.
I quickly evaluated if that'd be a nice UI/UX wise, and I think that it wouldn't be nice:
- I think we would reach a UI and choice overdose on the enduser side, which will lead to a less intuitive experience.
- On the implementer side, I think that unless we see a real need, it is best to have only the showMore & showLess method/actions at the facet level, instead of having it a the facetValue level, as it creates a whole other level of complexity. Unless the added complexity can be justified by an experience enhancement for the end-user, I reckon it is best to KISS


packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
```diff
@@ -141,10 +141,33 @@ export interface CoreCategoryFacetState {
...
+  /**
+   * The facet's values, represented in their hierarchical form
+   * if there's an hierarchy, flat otherwise.
+   */
+  valuesAsTrees: CategoryFacetValue[];
+
+  /**
+   * The facet's selected value if any, undefined otherwise.
+   */
+  activeValue: CategoryFacetValue | undefined;
```
btaillon, line 153: should this be:
```suggestion
  activeValue?: CategoryFacetValue;
```

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
```diff
@@ -141,10 +141,33 @@ export interface CoreCategoryFacetState {
   /** The facet ID. */
   facetId: string;
 
-  /** The facet's parent values. */
+  /**
+   * The facet's values, represented in their hierarchical form
+   * if there's an hierarchy, flat otherwise.
```
btaillon, line 146:
```suggestion
   * if there's a hierarchy, flat otherwise.
```

btaillon: I find it a bit unclear what "if there's a hierarchy" means here, and what "flat otherwise" means too.

I think we should give some examples of what this value may contain.

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.test.ts
```diff
@@ -295,7 +295,7 @@ describe('category facet', () => {
         expect(categoryFacet.state.canShowMoreValues).toBe(true);
       });
 
-      it('if #moreValuesAvailable is true, #state.canShowMore is true', () => {
+      it('if #moreValuesAvailable is false, #state.canShowMore is false', () => {
```
btaillon, line 298: should add more tests IMO

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
```diff
@@ -353,19 +377,36 @@ export function buildCoreCategoryFacet(
...
+      let activeValue: CategoryFacetValue | undefined = undefined;
+      const valueToVisit = [...valuesAsTrees];
+      while (valueToVisit.length > 0) {
+        const currentValue = valueToVisit.pop()!;
+        if (currentValue.state === 'selected') {
+          activeValue = currentValue;
+          break;
+        }
+        valueToVisit.push(...currentValue.children);
+      }
```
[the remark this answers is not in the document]
louis-bompart: Agreed for refacto
- [ ] refacto into func

However, nay for recursion, I dread the stack overflow.

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.test.ts
louis-bompart: Entirely agree 💯
I should have put that test are Todo in the pr desc

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
louis-bompart: Hmm ts playground inc to indicate why no

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
louis-bompart: - [ ] add JSDOC example of possible values, flat and hierarchical

olamothe (approved):
Not much to add vs what Benjamin already pointed out.

The main one will be more test added around this

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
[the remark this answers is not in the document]
y-lakhdar: I also thought about the recursion, but I prefer the simplicity of this approach.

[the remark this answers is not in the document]
louis-bompart: > Should this be kept in the store ? 🤔 ie: computed directly in the reducer and added to the state.
> 
> On the one hand, this is a computed values so there's an argument that we should not do that.
> 
> On the other hand, someone not using the controller/trying to read from the store directly would have to repeat this non trivial computation on their end to achieve the same logic.

I thought of this, and decided to proceed like this because:

1. `parents` and `values` were doing similar computation already. (However, this is a weak argument: the fact we did it in the past doesn't mean we should persist if this is a bad idea)
2. What would make sense for the user? I think it makes sense for a facet as an entity to expose its active value(s) if any.
3. The alternative is likely to create more trouble/confusion for the end-user

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
louis-bompart: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYjgG2QG8AoZZdKbdAfgC5kBnMKUAcwG4SBfEk0SLEQpUUAKogAJhBigIk4mQpV0DZqxBsAPgFcpMuZJ59pCfHCgoEuZsmpZcBBhgd58JU+cvJrIW9nF9WRB5Z0DpYPk+exw3JjgsRlkIRjRwgxCFEgCJCMN4xOTUl1iCIA

If you have `prop: Type | undefined` in an interface, you guarantee that the key `prop` is present in an object that respect the interface
If you have `prop?: type` however, you do not offer this guarantee.

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
louis-bompart: Comment has been rewritten/simplified

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
btaillon: yeah I know that the union type means the key is required, my question is more so about whether we should have the key be present when there's no selected value. It's not something I feel strongly about though.

PR #3107, opened 2023-08-14
feat: add headless react util defineSearchEngine

mrrajamanickam-coveo:
Adds React util to define search engine that provides context provider, hooks to hydrate engine and controllers.

## TODO
- [ ] Add more unit tests 

https://coveord.atlassian.net/browse/KIT-2606


packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,108 @@
...
+} from 'react';
+import {
+  ContextCSRState,
+  ContextState,
+  ControllerHook,
+  InferControllerHooksMapFromDefinition,
+  ReactSearchEngineDefinition,
+} from './types';
+
+export function capitalize(s: string) {
```
olamothe, line 25: Does this really need to be exported from the module ?

packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,108 @@
...
+  TControllers extends ControllerDefinitionsMap<TEngine, Controller>,
+  TKey extends keyof TControllers,
+>(
+  context: Context<ContextState<TEngine, TControllers>>,
+  key: TKey
+): ControllerHook<InferControllerFromDefinition<TControllers[TKey]>> {
+  return () => {
+    const ctx = useContext(context);
+    if (ctx === null) {
+      throw 'Missing SSR state or CSR provider.';
```
olamothe, line 49: Since those will appear for implementers, I think we should limit the amount of acronyms in use.

Also, if that happens what (if anything) could an implementer do to potentially troubleshoot and fix the problem ? 

If there's something that can be done, the message should try to explain it.

packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+    const subscribe = useCallback(
+      (listener: () => void) =>
+        isCSRContext(ctx) ? ctx.controllers[key].subscribe(listener) : () => {},
+      [ctx]
+    );
+    const getSSRState = useCallback(() => ctx.controllers[key].state, [ctx]);
+    // TODO: Eval need for `useSyncMemoizedStore() instead of useSyncExternalStore()`
```
olamothe, line 57: Should these be linked to some investigation JIRA ? 
If only to explain why these alternative would be better ?

packages/headless-react/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc --project ./tsconfig.build.json",
+    "test": "jest",
+    "test:watch": "jest --watch --colors --no-cache",
+    "lint": "eslint ."
+  },
+  "dependencies": {
```
olamothe, line 28: I assume there was discussion about this that I might have missed, but should this be a peer dependency instead 🤔  ?

My first intuition would be that implementers would already install/need `@coveo/headless` no matter what, and that they can optionally install this as an... "addon" kinda thing ?

olamothe (approved):

packages/headless-react/src/search-engine.tsx
dmbrooke: +1

I think most (if not all) TODOs should be linked to a JIRA as to not fall through the cracks

dmbrooke (approved):
Don't have anything that hasn't already been said. Looks great 🎉

packages/headless-react/src/types.ts
```diff
@@ -0,0 +1,72 @@
...
+import {InferControllerFromDefinition} from '@coveo/headless/dist/definitions/app/ssr-engine/types/common';
+import {
+  ControllerDefinitionsMap,
+  InferControllersMapFromDefinition, // TODO: Rename Initial/Live to SSR/CSR in headless/ssr package
+  InferControllerInitialStateMapFromDefinitions as InferControllerSSRStateMapFromDefinitions,
+  EngineDefinition,
+} from '@coveo/headless/ssr';
+import {FunctionComponent, PropsWithChildren} from 'react';
+
+export type ContextSSRState<
```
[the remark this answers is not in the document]
btaillon: Some of these types (including this one) aren't useful outside `headless-react`. They are extracted mostly due to their complexity AFAIK.

btaillon (approved):
Fantastic!

packages/headless-react/src/types.ts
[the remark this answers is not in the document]
mrrajamanickam-coveo: Yes I noticed that we don't have separate type files in other packages. It started with `headless/ssr` which had lot more types that it had to be broken into multiple files. Don't see any strong pros/cons either way in this case for now.

packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export function capitalize(s: string) {
+  return `${s.slice(0, 1).toUpperCase()}${s.slice(1)}`;
+}
+
+function isCSRContext<
+  TEngine extends CoreEngine,
+  TControllers extends ControllerDefinitionsMap<TEngine, Controller>,
+>(
+  ctx: Exclude<ContextState<TEngine, TControllers>, null>
```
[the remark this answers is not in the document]
mrrajamanickam-coveo: Can you please explain the reasoning behind?

mrrajamanickam-coveo (commented):
Added couple more unit tests (increasing coverage from ~20% to ~68%.
There are still couple of things to take care of in terms of tests and review comments. 
Will continue working on them ..

packages/headless-react/src/search-engine.tsx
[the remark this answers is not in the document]
btaillon: My reasoning is very subjective of course. I believe that given your experience, your judgement would be better than mine on this.

Here's my reasoning though:
# Reason 1: it's more intuitive
Instinctively, when I see a type like this, I assume it's most likely an interface, and `null` isn't assignable to an interface.

# Reason 2: less presumptuous
I usually think of `null` or `undefined` as meaning "not specified" or "optional". Usually, whether something is optional depends on where the type is needed in the first place.

For example, if we had a type like `Person` and a method called `setParent(child, newParent)`, it contextually makes sense that `child` isn't nullable/optional but `newParent` is nullable/optional.

# Reason 3: it's easier to extend
To make something non-optional into something optional, we can either use unions or the `someParam?: SomeType` syntax.

To make something optional into something non-optional, we can either use `Exclude<SomeType, null | undefined>` or `Required<ParentOfSomeType>`.

Making something non-optional into something optional seems easier/prettier to me than the other way around.

# Counter point
Arguably that `ContextState` is very strictly tied to the React context. Although we assert that it's not null, in reality, it can be null until that assertion.

Additionally, we could argue that `ContextState` should technically be something like `type ContextState = SSRState | CSRState | null`.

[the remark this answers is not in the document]
mrrajamanickam-coveo: Nm, this is obsolete now with the refactor to use singleton context

packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,108 @@
...
+      const {state: _, subscribe: __, ...remainder} = controller;
+      return mapObject(remainder, (member) =>
+        typeof member === 'function' ? member.bind(controller) : member
+      ) as Omit<typeof controller, 'state' | 'subscribe'>;
+    }, [ctx]);
+    return {state, methods};
+  };
+}
+
+export function defineSearchEngine<
```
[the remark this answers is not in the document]
mrrajamanickam-coveo: Added a brief doc for now. We do have https://coveord.atlassian.net/browse/KIT-2607 for revising all docs before release.

mrrajamanickam-coveo (commented):
Finally resolved errors related to SSR hooks - thanks @btaillon for all the debugging help!
Added e2e tests.
JSDOM Unit tests are still problematic - tempted to scrap them or skip and investigate later.

mrrajamanickam-coveo (commented):
Addressed all Review comments, TODOs. Please review.
Updated react samples and e2e tests to include search box controller from @btaillon.
Disabled jest tests due to jsdom issues (KIT-2692).

mrrajamanickam-coveo:
CI test failure is unrelated to changes in this PR

[Run e2e tests on Atomic (cypress/e2e/result-list/**/*)](https://github.com/coveo/ui-kit/actions/runs/5957618617/job/16160674988?pr=3107#logs)

```
41 passing (2m)
  2 failing

  1) Result MultiValueText Component
       when used inside a result template
         when the field value exists & is a string array
           and the field has an array of 2 values
             with max-values-to-display=1
               should display a "1 more" label:
     TypeError: doc.createTreeWalker is not a function
      at findShadowRoots (http://localhost:3333/__cypress/runner/cypress_runner.js:169004:22)
```

packages/headless-react/src/client-wrapper.ts
```diff
@@ -0,0 +1,31 @@
+'use client';
+
+import {useEffect, useReducer, useRef} from 'react';
+
+/**
+ * Alternate for `useSyncExternalStore` which runs into infinite loops when hooks are used in `getSnapshot`
+ * https://github.com/facebook/react/issues/24529
+ */
+export function useSyncMemoizedStore<T>(
```
btaillon, line 9: I think we should have unit tests for this.

Some tests I would like to have:
1. When changing the `subscribe` function, unsubscribes then re-subscribes to the new function.
2. When changing `getSnapshot`, should replace the current snapshot by calling `getSnapshot` again.
  * This may not pass currently. One way to make it pass may be to replace `useRef(...)` with `useMemo(() => createRef(...), [getSnapshot])`.
3. Passing the value returned from `useSyncMemoizedStore` to the deps array of a hook (like `useEffect` or `useMemo`) only causes a re-render when the listener of `subscribe` is called.

packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export class MissingEngineProviderError extends Error {
+  constructor(
+    message = 'Unable to find Context. Please make sure you are wrapping your component with either `SSRStateProvider` or `CSRProvider` component that can provide the required context.'
+  ) {
+    super(message);
+  }
+}
+
+function capitalize(s: string) {
```
btaillon, line 29: Could be extracted to a separate `utils` file

packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function buildControllerHook<
+  TEngine extends CoreEngine,
+  TControllers extends ControllerDefinitionsMap<TEngine, Controller>,
+  TKey extends keyof TControllers,
+>(
+  singletonContext: {
+    get(): Context<ContextState<TEngine, TControllers> | null>;
+  },
```
btaillon, line 73: I feel like we should add a type of singletons.

For example:
```ts
export type Singleton<T> = { get(): T };
```

packages/headless-react/src/search-engine.tsx
```diff
@@ -0,0 +1,139 @@
...
+  InferControllerHooksMapFromDefinition,
+  ReactSearchEngineDefinition,
+} from './types';
+
+export class MissingEngineProviderError extends Error {
+  constructor(
+    message = 'Unable to find Context. Please make sure you are wrapping your component with either `SSRStateProvider` or `CSRProvider` component that can provide the required context.'
+  ) {
+    super(message);
+  }
```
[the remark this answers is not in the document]
btaillon: I also wonder if there's a way we could include the component itself in the error object.

Something like:
```suggestion
  constructor(
    public readonly node: ReactNode,
  ) {
    super(
      'Unable to find Context. Please make sure you are wrapping your component with either `SSRStateProvider` or `CSRProvider` component that can provide the required context.'
    );
  }
```

I don't know of a way to obtain the current React node to print it though, and I don't know what would be the best practice.

packages/headless-react/src/search-engine.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+      expect(errorSpy).toHaveBeenCalled();
+    });
+
+    test('SSRProvider', async () => {
+      const ssrState = await fetchInitialState();
+      render(<SSRStateProvider controllers={ssrState.controllers} />);
+      expect(errorSpy).not.toHaveBeenCalled();
+    });
+
+    test('CSR Provider', async () => {
```
btaillon, line 106:
```suggestion
    test('CSRProvider', async () => {
```

packages/headless-react/src/search-engine.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+      useEngine,
+      SSRStateProvider,
+      CSRProvider,
+    ].forEach((returnValue) => expect(typeof returnValue).toBe('function'));
+
+    expect(controllers).toEqual({});
+    expect(rest).toEqual({}); // No other return values
+  });
+
+  test('creates a hook based on given controller', () => {
```
btaillon, line 52: I think we should also add some tests to verify the hooks work correctly.

E.g.:
1. When using the SSRStateProvider, returns a `state` but no `methods`.
2. When using the CSRProvider
   a. Returns a `state` and `methods`.
   b. Subscribes to the controller and updates the `state` on change.
   c. Reacts correctly when a method is called (e.g., `searchBox.submit()` should update the state).
3. When switching from SSRStateProvider to CSRProvider
   a. Should cause a re-render / update the state.

packages/samples/headless-ssr/src/app/common/result-list.tsx
```diff
@@ -0,0 +1,17 @@
+import {Result} from '@coveo/headless';
+
+interface ResultListCommonProps {
+  results: Result[];
```
btaillon, line 4: I think this would be more copy and paste-able (and possibly easier to read as a sample) if we changed the props of common components to match the values returned by the hook (which also happen to be the props I used in the generic ssr samples).

For example:
```suggestion
  state: ResultListState;
  methods: ResultListMethods;
```

packages/samples/headless-ssr/src/app/common/result-list.tsx
```diff
@@ -0,0 +1,17 @@
+import {Result} from '@coveo/headless';
```
btaillon, line 1: unfortunately, I think we'll eventually need to export "*Methods" types from either Headless or the HeadlessReact package.

For example:
```ts
export type ResultListMethods = Omit<ResultList, 'state' | 'subscribe'>;
```

btaillon: To be discussed of course (not for this PR).

packages/samples/headless-ssr/src/app/react/components/search-page.tsx
```diff
@@ -0,0 +1,50 @@
...
+    });
+  }, [ssrState]);
+
+  const content = (
+    <>
+      <SearchBox />
+      <ResultList />
+      <HydrationMetadata ssrState={ssrState} csrResult={csrResult} />
+    </>
+  );
```
btaillon, line 32: Unlike the other sample, I think we should take all these these components (except `HydrationMetadata`) as `children`, since we don't need to manually give them the state as parameters anymore. It was part of my motivation behind generating providers.

The `SearchPage` component here could be renamed `SearchPageProvider` or something like that.

btaillon (approved):
that looks fantastic!

packages/samples/headless-ssr/src/app/react/components/search-page.tsx
mrrajamanickam-coveo: Done.

mrrajamanickam-coveo:
Fixed Jest env issues. Will add more tests as commented in followup PR https://coveord.atlassian.net/browse/KIT-2694

mrrajamanickam-coveo:
Working on fixing test failures in CI (that passed locally) 

```
 FAIL  src/search-engine.test.tsx
  ● Test suite failed to run

    src/search-engine.test.tsx:93:23 - error TS2339: Property 'toHaveLength' does not exist on type 'ArrayLikeMatchers<HTMLElement>'.

    93       expect(results).toHaveLength(numResults);
                             ~~~~~~~~~~~~
```

```
 11 passing (31s)
  1 failing

  1) react Headless SSR utils
       should pass the web-vitals audits:
     Error: cy.vitals() - A threshold has been crossed.

ttfb web-vital is 21.70000000001164, and is over the 20 threshold. Fail.
      at Context.eval (webpack://@coveo/headless-ssr-samples/../../../node_modules/cypress-web-vitals/src/reportResults.js:83:0)

```

packages/headless/src/ssr.index.ts
```diff
@@ -6,17 +6,26 @@ polyfillCryptoNode();
 export type {Unsubscribe, Middleware} from '@reduxjs/toolkit';
 
 // ⚠️ NOTE: All exported SSR types, APIs should be marked as `@internal` until MVP is complete
+export {mapObject} from './utils/utils';
```
btaillon, line 9: I would prefer not to export generic utilities like `mapObject` from the library. I would be ok copy and pasting it into the new library.

packages/samples/headless-ssr/src/app/react/components/search-page.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export default function SearchPage({ssrState}: SearchPageProviderProps) {
+  return (
+    <SearchPageProvider ssrState={ssrState}>
+      <SearchBox />
+      <ResultList />
+    </SearchPageProvider>
+  );
+}
```
btaillon, line 60: this could be moved into `page.tsx`

packages/headless/src/ssr.index.ts
mrrajamanickam-coveo: Makes sense. Added a TODO to use a third-party lib if possible and if it doesn't increase bundle size - on one hand it is just handful of lines of code - but on other hand duplicating and maintaining 2 copies of a small util that might not need to be modified in future vs using a popular third-party that seems to handle more scenarios and had dedicated tests etc ..

packages/samples/headless-ssr/src/app/react/components/search-page.tsx
btaillon: This may be because `SearchBox` and `ResultList` are missing the `"use client"` directive

PR #3112, opened 2023-08-16
feat(atomic): make collapseFacetsAfter feature work with automatic facets (facet-manager + refine-modal)

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2652

This PR solves the problem where the collapseFacetsAfter feature was not working in both facet manager and refine modal.

- add isHidden function to every static facet component that gets registred in the AtomicStore. This function serves the purpose of knowing when to stop collapsing facets to start collapsing the automatic ones.

- Big revamp of refine-modal, let go of the old `getClonedFacetElements` function and created my own called `createFacetSlot` to better implement automatic facets 

- remove AutomaticFacetSlotContent. With this revamp of the refine modal, I incorporated the automatic facets inside of the `<div slot="facets"> `. I am using directly the AutomaticFacetGenerator instead to generate them.

- add listener to a facet manager controller in the refine modal to update the collapse state whenever new facets become visible or invisible (due to facet dependency and DNE maybe) 

- replace the 'are-collapsed' prop by 'collapse-facets-after' in facet generator

### Exemple (in facet-manager but refine-modal is the same thing)
BEFORE
collapseFacetsAfter = 3 
automatic facets are colored
image

AFTER 
collapseFacetsAfter = 3 
automatic facets are colored
image


packages/atomic/src/components/search/facets/atomic-rating-range-facet/atomic-rating-range-facet.tsx
```diff
@@ -344,7 +353,7 @@ export class AtomicRatingRangeFacet
   }
 
   private get valuesToRender() {
-    return this.facetState.values.filter(
+    return this.facet.state.values.filter(
```
alexprudhomme, line 356: Helps with race condition between update of facetState object and the isHidden() call.

packages/atomic/src/components/search/facets/atomic-rating-facet/atomic-rating-facet.tsx
```diff
@@ -356,7 +365,7 @@ export class AtomicRatingFacet
   }
 
   private get valuesToRender() {
-    return this.facetState.values.filter(
+    return this.facet.state.values.filter(
```
alexprudhomme, line 368: Helps with race condition between update of facetState object and the isHidden() call.

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -253,6 +263,101 @@ export function facetShouldBeInitiallyCollapsed(
...
+  });
+}
+export function updateCollapseFacetsAfter(
+  generator: AutomaticFacetGeneratorElement,
+  index: number,
+  collapseFacetsAfter: number
+) {
+  if (!generator) {
+    return;
+  }
```
btaillon, line 310: The type of `generator` is non-optional and non-nullable, and it isn't scalar either, which implies that this condition would always be falsy and this `if` block wouldn't do anything. I recommend adjusting either the code outside of this function or making `generator` nullable.

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -253,6 +263,101 @@ export function facetShouldBeInitiallyCollapsed(
...
+      index,
+      collapseFacetsAfter
+    );
+  });
+}
+export function updateCollapseFacetsAfter(
+  generator: AutomaticFacetGeneratorElement,
+  index: number,
+  collapseFacetsAfter: number
+) {
```
btaillon, line 307: The purpose of this function feels ambiguous, I was only able to really understand it by looking at both the code where it gets called and its internal code.

I would prefer removing it altogether, and simply doing this where it was getting called:
```ts
const remainingFacetsToShow = collapseFacetsAfter - visibleFacets.length;
if (generator) {
  generator.collapseFacetsAfter = remainingFacetsToShow;
}
```

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -253,6 +263,101 @@ export function facetShouldBeInitiallyCollapsed(
...
+export function isTagNameAutomaticFacetGenerator(tagName: string) {
+  return tagName === 'ATOMIC-AUTOMATIC-FACET-GENERATOR';
+}
+
+function isPseudoFacet(el: Element): el is BaseFacetElement {
+  return 'facetId' in el;
+}
+
+export function getFacetsInChildren(
+  collection: HTMLCollection
```
btaillon, line 329: instead of taking a collection as a parameter, I have a slight preference for taking the parent node.

```suggestion
  parent: HTMLElement
```

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -253,6 +263,101 @@ export function facetShouldBeInitiallyCollapsed(
...
+): BaseFacetElement[] {
+  const children = Array.from(collection);
+  const facets = children.filter((child) =>
+    isPseudoFacet(child)
+  ) as BaseFacetElement[];
+
+  return facets;
+}
+export function getGeneratorInChildren(
+  collection: HTMLCollection
```
btaillon, line 339: Same comment as above.
```suggestion
  parent: HTMLElement
```

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -253,6 +263,101 @@ export function facetShouldBeInitiallyCollapsed(
...
+): AutomaticFacetGeneratorElement {
+  const children = Array.from(collection);
+
+  const automaticFacetGenerator = children.find((child) =>
+    isTagNameAutomaticFacetGenerator(child.tagName)
+  ) as AutomaticFacetGeneratorElement;
```
btaillon, line 345: This `as` type is incorrect, since `find` may return `null`.

The replacement I suggested for `isTagNameAutomaticFacetGenerator`, if passed directly as a parameter, should remove the need for this `as`.

```suggestion
  const automaticFacetGenerator = children.find(isAutomaticFacetGenerator);
```

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -253,6 +263,101 @@ export function facetShouldBeInitiallyCollapsed(
...
+
+  return automaticFacetGenerator;
+}
+
+export function sortFacetsViaManager(
```
btaillon, line 350: I would prefer the term `using`:
```suggestion
export function sortFacetsUsingManager(
```

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -342,6 +448,10 @@ export class FacetCommon {
     this.dependenciesManager?.stopWatching();
   }
 
+  private get isHidden() {
+    return !this.facet.state.enabled || !this.facet.state.values.length;
+  }
```
btaillon, line 453: Didn't think about this before, but I think `canBeExpanded` would be more relevant/direct and avoids accidentally breaking this in the future.

packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -126,9 +132,115 @@ export class AtomicRefineModal implements InitializableComponent {
...
+
+    return generator;
+  }
+
+  private queryCurrentFacets(): BaseFacetElement[] {
+    const facetIds = Object.keys(this.bindings.store.getAllFacets());
+    const facetElements: BaseFacetElement[] = [];
+
+    facetIds.forEach((facetIdValue) => {
+      const matchingElement = document.querySelector(
```
btaillon, line 206: **Warning**: we should never use `querySelector` from the root `document`. We should assume there might be multiple search interfaces in the page.

btaillon: Why are we using `this.bindings.store.getAllFacets()` and mapping them to elements instead of directly calling `this.bindings.store.getFacetElements()`?

packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -126,9 +132,115 @@ export class AtomicRefineModal implements InitializableComponent {
...
+
+    updateCollapsedState(visibleFacets, this.collapseFacetsAfter);
+    updateCollapseFacetsAfter(
+      generator,
+      visibleFacets.length,
+      this.collapseFacetsAfter
+    );
+  }
+
+  private addElementStyle(el: HTMLElement) {
```
btaillon, line 238: I think this should have a more descriptive name, like:
```suggestion
  private addFacetColumnStyling(style: CSSStyleDeclaration) {
```

packages/atomic/src/components/search/facets/atomic-automatic-facet-generator/atomic-automatic-facet-generator.tsx
```diff
@@ -60,9 +60,12 @@ export class AtomicAutomaticFacetGenerator implements InitializableComponent {
    * @beta - This prop is part of the automatic facets feature.
    * Automatic facets are currently in beta testing and should be available soon.
    *
-   * Specifies whether the automatic facets are collapsed.
+   * Using the value `0` collapses all facets.
+   * Using the value `-1` disables the feature and keeps all facets expanded. Useful when you want to set the collapse state for each facet individually.
```
btaillon, line 64: I don't really like the pattern of using `-1` as meaning "disabled", although I realize that `undefined` means "default".

Maybe we could take a string instead? I think Stencil may support union types. For example:
```ts
  @Prop() public collapseFacetsAfter?: number | 'never-collapse';
```

packages/atomic/src/components/search/facets/atomic-rating-facet/atomic-rating-facet.tsx
btaillon: The `render` function is triggered whenever a `@State` is updated, therefore I would expect it to only render the `@State`, not a value that was dynamically updated.

This function is used during rendering, which makes sense IMO since it's suffixed with `ToRender`.

I think we should find other means of avoiding the race condition.

packages/atomic/cypress/e2e/facets/manager/facet-manager.cypress.ts
```diff
@@ -41,4 +64,107 @@ describe('Facet Manager Test Suite', () => {
...
+      setup(
+        {collapseFacetsAfter: MINUS_ONE_COLLAPSE_FACETS_AFTER},
+        addFacetManagerWithStaticFacets
+      );
+      assertFacetsNoCollapsedAttribute();
+    });
+  });
+
+  describe('with automatic facets only', () => {
+    it(`should only keep the first ${DEFAULT_COLLAPSE_FACETS_AFTER} facets expanded by default`, () => {
```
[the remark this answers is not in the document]
olamothe: I think it's better to accept some repetition for tests (tests are by nature always very repetitive) VS adding some layer of indirection that makes them harder to read...

packages/atomic/src/components/search/facets/atomic-automatic-facet-generator/atomic-automatic-facet-generator.tsx
olamothe: It's the same pattern that is already in use in the project for the same kind of property (atomic-facet-manager).

Not using the same pattern would be more problematic.

olamothe:
Will re-open when issues can be fixed with e2e

PR #3114, opened 2023-08-17
test(samples/headless-ssr): move existing components under app/ssr

mrrajamanickam-coveo:
In prep for creating a new `/react` route for `headless-react` e2e tests.

https://coveord.atlassian.net/browse/KIT-2683


dmbrooke (approved):

mrrajamanickam-coveo:
Will fix the error - the GH action is waiting for the base URL which now returns a 404 

```
- ready started server on 0.0.0.0:3000, url: http://localhost:3000
http://localhost:3000 timed out on retry 6 of 3, elapsed 92454ms, limit 90000ms
Error: Response code 404 (Not Found)
```

packages/samples/headless-ssr/src/app/ssr/components/search-page.tsx
```diff
@@ -4,7 +4,7 @@ import {
   SearchSSRState,
   SearchCSRState,
   hydrateInitialState,
-} from '@/src/common/engine';
+} from '@/src/app/ssr/common/engine';
```
btaillon, line 7: instead of `ssr`, I think I would prefer something like `generic`/`agnostic`, since:
* This sample is already in the `headless-ssr` sample package, the name may be redundant.
* Technically, the `react` samples are also `ssr`, so a different name may help differentiate it.

btaillon (approved):

packages/samples/headless-ssr/src/app/ssr/components/search-page.tsx
mrrajamanickam-coveo: Was thinking the same, renamed to `generic`

PR #3115, opened 2023-08-17
fix(headless): adapted the usage of the AbortController to make GenQI Quantic component work with locker service

mmitiche:
[SFINT-5154](https://coveord.atlassian.net/browse/SFINT-5154)

#### Before the fix:
We were having the following error whenever we try to use the new Quantic GenAI component in a Salesforce community where locker service is enabled. Unfortunately  AbortController is not available in locker service:

```
{
    "type": "generatedAnswer/streamAnswer/rejected",
    "meta": {
        "arg": {},
        "requestId": "FDt8g6eQPy8R1--btntmQ",
        "rejectedWithValue": false,
        "requestStatus": "rejected",
        "aborted": false,
        "condition": false
    },
    "error": {
        "name": "TypeError",
        "message": "AbortController is not a constructor",
        "stack": "TypeError: AbortController is not a constructor\n    at $d.streamGeneratedAnswer (https://xero--devtestsl2.sandbox.my.site.com/resource/1692234989000/coveoheadless/browser/headless.js:49:2303)\n    at eval (https://xero--devtestsl2.sandbox.my.site.com/resource/1692234989000/coveoheadless/browser/headless.js:57:93211)"
    }
}
```

### After the fix:

Instead of directly instantiating the Abort Controller now we using the `createAbortController` function described here: [abort-controller-polyfill.ts](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/utils/abort-controller-polyfill.ts
), this function checks if we can create a AbortController instance before actually doing it.

This change will only have an impact on Salesforce when locker service is enabled.
In this situation, since `abortController` can't be used, we are no longer aborting the requests. and to avoid any confusions in the event payloads received, we are adding in this PR some conditions to make the Headless state reacts only the messages that were sent with a stream id value that corresponds to the stream id returned in the latest search query.

https://github.com/coveo/ui-kit/assets/86681870/6de0bfc8-2295-4cb2-876d-87ff4f041259

[SFINT-5154]: https://coveord.atlassian.net/browse/SFINT-5154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olamothe:
What's the impact when the request fails, for the component/end user ? Where as before the request was aborted, now there's just "nothing" that will happen. Does it still recover nicely ? What will end user see in those case ?

packages/headless/src/features/generated-answer/generated-answer-actions.ts
```diff
@@ -144,21 +144,46 @@ export const streamAnswer = createAsyncThunk<
...
+      resetAnswer: (streamId: string) => {
+        if (
+          streamId ===
+          state.search.extendedResults.generativeQuestionAnsweringId
+        ) {
+          dispatch(resetAnswer());
+        }
       },
-      close: () => dispatch(setIsStreaming(false)),
-      resetAnswer: () => dispatch(resetAnswer()),
```
nathanlb, line 161: how and why are these removed?

nathanlb:
The solution i suggested kind of sucks. It's far less clean than what we had and I literally thought of it in 2 seconds while writing my message. I'd be surprised if we couldn't do better by taking the time to think about it and investigate further. Locker service completely banning abortController seems ridiculous but what about salesforce isn't ridiculous right? There has to be a workaround better than letting a stream complete silently in the background, dispatching no-op actions and risking dangerous side effects.

mmitiche:
> What's the impact when the request fails, for the component/end user ? Where as before the request was aborted, now there's just "nothing" that will happen. Does it still recover nicely ? What will end user see in those case ?

There is no impact for the end user, cause when the request fails we update the state  to reflect the error regardless of whether the request was aborted or not.

abort

`dispatch(updateError(error))` will make the component behave in its error state, so the end user using the component will feel no impact.

erocheleau (approved):
I think this is good.
No other choice anyway

packages/headless/src/features/generated-answer/generated-answer-actions.ts
```diff
@@ -144,21 +144,46 @@ export const streamAnswer = createAsyncThunk<
   const abortController = extra.streamingClient?.streamGeneratedAnswer(
     request,
     {
-      write: (data: GeneratedAnswerStreamEventData) => {
-        dispatch(setIsLoading(false));
-        if (data.payload && data.payloadType) {
-          handleStreamPayload(data.payloadType, data.payload);
+      write: (data: GeneratedAnswerStreamEventData, streamId: string) => {
```
btaillon, line 147: I don't think it's necessary to pass `streamId` to these callbacks, since it's already part of `request`.

I'm also not sure whether it's necessary to pass `abortController` to these callbacks, since it looks like it's already returned by the function.

Lastly, in the `search-api-client`, we instantiate the `abortController` because the `SearchAPIClient` is responsible for aborting requests, but here the responsibility seems mixed between the action and the `GeneratedAnswerAPIClient`. I would recommend completely isolating the logic to one or the other.
* If you isolate the `AbortController` logic to `GeneratedAnswerAPIClient`, then you should ensure that `GeneratedAnswerAPIClient` stops calling `write` on old streams whenever a new stream is created. This would remove the need for these `if` conditions that compare `streamId` with `generativeQuestionAnsweringId`.
* If you isolate the `AbortController` logic to the action, then only the `signal` of the abort controller should be passed down to the stream. If any event such as `write`/`abort`/etc is called on a different streamId than the current context's, then the current context's `abortController.abort` should be called and the remaining logic of the callback should be skipped.

mmitiche: Hey! thanks for the suggestions!
Removed the stream id from the callbacks parameters, now directly using the `request`.

I also kinda reduced the mixture in responsibility between `GeneratedAnswerAPIClient ` and the `GeneratedAnswerAction` on who will call the abort controller,  now only the API client is calling the abort controller and this abort controller is no longer passed to the callbacks.

I would also want to mention that if you think that further refactoring is needed, I think it would be best to address it in a new PR just because this one was opened specifically to solve the abortController issue  occurring in Salesforce when locker Service is enabled, an issue that needs to be solved ASAP cause a client is waiting for it.

olamothe (approved):

PR #3116, opened 2023-08-17
feat(productlistings): [CAPI-63] [POC] Add facets, pagination, and basic controllers for PLP v2

Spuffynism:
This POC adds basic support for PLP v2 functionality:
- basic engine and controller
- Pagination controller
- Facets controller

We'll most likely want to have a CommerceEngine in its own `commerce` sub-package, to support all commerce experiences offered through headless.

[CAPI-63]

[CAPI-63]: https://coveord.atlassian.net/browse/CAPI-63?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/api/commerce/product-listings/product-listing-params.ts
```diff
@@ -34,6 +38,53 @@ export interface ProductListingRequestParam {
   userContext?: ContextPayload;
 }
 
+// TODO(nico): Use interfaces for nested params
+export interface ProductListingV2RequestParam {
```
olamothe, line 42: Yeah, and I think for some of those parameters they are already defined in the project... 🤔

There will be some overlap with what's defined in `api/search` for example, which we can probably just refactor/tweak a bit to reuse in both places.

packages/headless/src/api/commerce/product-listings/query-listing.http
```diff
@@ -0,0 +1,47 @@
+POST http://localhost:8100/rest/organizations/commercestore/properties/propertyid/commerce/v2/listing
+Content-Type: application/json
+Authorization: Bearer x1fafab91-152b-4c2e-9f16-681bc5113722
```
olamothe, line 3: Is it problematic to have this key in the clear on the internet ?

packages/headless/src/api/commerce/product-listings/query-listing.http
```diff
@@ -0,0 +1,47 @@
+POST http://localhost:8100/rest/organizations/commercestore/properties/propertyid/commerce/v2/listing
```
olamothe, line 1: What's the goal of this file 🤔  ?

packages/headless/src/app/product-listing-engine/product-listing-engine.ts
```diff
@@ -94,6 +117,39 @@ export function buildProductListingEngine(
...
+export function buildProductListingV2Engine(
+    options: ProductListingV2EngineOptions
+): ProductListingV2Engine {
+  const logger = buildLogger(options.loggerOptions);
+  validateConfiguration(options.configuration, logger);
+
+  const productListingClient = createProductListingClient(
+      options.configuration,
+      logger,
+      createSearchAPIClient(options.configuration, logger)
```
olamothe, line 129: That seems like an artefact of something that will simply go away eventually (ie: we will never want to rely on the search API at all down the line).

So in other words, by using the "old" product listing client, and having some flag to target v2 or v1 seems to be causing pollution. Which will just get worse IMO.

I think it's better to just do a clean break, create a `commerceServiceClient` (it won't deal with only product listing down the line, after all), and start from there.

packages/headless/src/controllers/core/facets/facet/headless-core-facet-options.ts
```diff
@@ -64,6 +64,51 @@ export interface FacetOptions {
   sortCriteria?: FacetSortCriterion;
 }
 
+export interface FacetOptionsV2 {
```
olamothe, line 67: There seems to be a lot of overlap with the existing option ? And not use in any place ?

If I understand the reasoning well, I'd say the correct way would be to define the new "limited" facet options in `controllers/product-listing` by using `Pick` or `Omit` from `FacetOptions`.

In any case, this definition should not exist in the `core` version of the controller

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet.ts
```diff
@@ -167,6 +177,84 @@ export function buildFacet(
...
 
+export interface Facets extends Controller {
+  state: FacetsState;
+}
+
+export interface FacetsState {
+  facets: AutomaticFacet[];
+}
+
+// TODO(nico): We might need to extract a "core" automatic facets controller
```
olamothe, line 188: For everything related to automatic facets, I would not worry too much about it for the short term... 

I think it's needlessly complicating the first version of what we are trying to achieve here.

We should definitely support it eventually, but I would not put that in scope for the first few versions.

ie: Just having "normal" static facet working correctly from A to Z is a good enough achievement 😉

packages/headless/src/controllers/product-listing/headless-product-listing.ts
```diff
@@ -25,6 +28,50 @@ const optionsSchema = new Schema({
   }),
 });
 
+
+export enum SortBy {
```
olamothe, line 32: Quite a bit of duplication from stuff that exists elsewhere in the project, that I think we could just refactor a bit and re-use properly.

packages/headless/src/features/product-listing/product-listing-actions.ts
```diff
@@ -1,24 +1,27 @@
...
+    CategoryFacetSection,
+    ConfigurationSection,
+    ContextSection,
+    DateFacetSection,
+    FacetOptionsSection,
+    FacetOrderSection,
+    FacetSection,
+    NumericFacetSection,
+    PaginationSection,
+    ProductListingSection, ProductListingV2Section,
```
olamothe, line 23: Seems like there's some weird code formatter issue ? 🤔

Prettier formatted it like that ?

packages/headless/src/controllers/product-listing/headless-product-listing.ts
```diff
@@ -151,9 +241,66 @@ export function buildProductListing(
   };
 }
 
+export function buildProductListingV2(
```
olamothe, line 244: I guess this controller could actually be called `productList` instead of `listing ? 

In the same vein that we have a `resultList`, it's not that different.

And the main function that are linked to this controller are potentially... a bit illogical (`refresh`).

What we could do is picture the whole thing like this:

- One productListing engine
- One or multiple facets bound to that engine
- One productList bound to that engine
- One pager bound to that engine
- Sort bound to that engine
- etc...

The product listing engine would expose the `refresh()`, `setProductListingId()`, `setURL()`, etc.. functions since those are global operation (ie: it affects every controller bound to that engine).

ProductList could expose function related to "infinite scrolling" (`fetchMoreProducts`), since it's the functionality you would expect on a list of product.

olamothe (commented):
Not bad !

packages/headless/src/api/commerce/product-listings/query-listing.http
Spuffynism: it's an invalid access token, so no. Removed the file since it was included because I forgot to clean up after myself 😇

packages/headless/src/api/commerce/product-listings/query-listing.http
Spuffynism: I was using it to test the API stub. Removed it!

packages/headless/src/app/product-listing-engine/product-listing-engine.ts
Spuffynism: We need the search api client for facet search, as we currently don't support it on the PLP v2 spec.

In the squad, we talked about creating our own facet search endpoint. I'll reignite the conversation and come back with more info!

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet.ts
Spuffynism: With the PLP v2 endpoint, facets are effectively always automatic facets. A customer won't be able to specify which facets they want as the PLP v2 endpoint will return facets on the product listing fetch query. Once the facets are returned from the API, customers will be able to manipulate facets as usual.

This is why the facets are implemented as a variation of automatic facets, and not as regular facets.

packages/headless/src/features/product-listing/product-listing-actions.ts
Spuffynism: Cleaned this up in https://github.com/coveo/ui-kit/pull/3116/commits/f53124a1c1fd10b30e523e2b5535ca5716e59c94

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet.ts
olamothe: I see, I stand corrected then. Disregard the above !

packages/headless/src/controllers/product-listing/headless-product-listing.ts
olamothe: We should do a quick sync about that.

I'll start a design doc and share it with you and @fbeaudoincoveo for now.

Spuffynism: I have the very minimal base for a commerce sub-package here fwiw:
https://github.com/coveo/ui-kit/compare/master...feat-headless-commerce

packages/headless/src/features/product-listing/product-listing-actions.ts
```diff
@@ -168,17 +241,76 @@ export const buildProductListingRequest = async (
...
+    locale: state.productListing.locale || 'en-US',
+    mode: state.productListing.mode || 'live',
+    clientId: state.productListing.clientId || 'some-client-id', // Dummy value since the api requires one
+  };
+
+  return {
+    ...baseParams,
+    ...productListingParams,
+    selectedFacets: facets,
+    context: {
```
fbeaudoincoveo, line 269: Why are we returning a bunch of dummy values? Wouldn't these be retrievable from the state object?

packages/headless/src/features/product-listing/product-listing-state.ts
```diff
@@ -32,4 +33,31 @@ export const getProductListingInitialState = (): ProductListingState => ({
   error: null,
   isLoading: false,
   responseId: '',
+  version: '',
+});
+
+export interface ProductListingV2State {
+  listingId: string;
+  locale: string; // TODO: action for setting this
+  mode: string; // TODO: action for setting this
```
fbeaudoincoveo, line 42: Should be an enum

packages/headless/src/api/commerce/product-listings/product-listing-params.ts
```diff
@@ -34,6 +38,47 @@ export interface ProductListingRequestParam {
...
+  selectedPage?: {
+    page?: number;
+  };
+  selectedSort?: SortCriterion;
+  context: {
+    user: {
+      userId?: string;
+      email?: string;
+      ipAddress: string;
+      userAgent: string;
```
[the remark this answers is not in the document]
fbeaudoincoveo: I mean we could probably do without them

I believe the user agent and IP address can both be extracted from the request headers on the API side

And if the user is logged in, then we should also be able to retrieve their email address from the access token... Unless API key authentication is used in the app, in which case passing the email would be legit.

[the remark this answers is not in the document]
Spuffynism: I've asked your question on the spec: https://docs.google.com/spreadsheets/d/1-WFpnipA3fgwCX3GhvnmJvPSwO_jfiQY-lZytsapfKs/edit?disco=AAAA3Gak_T0

As for the email, API key authentication will be used. This means we can't extract the email from the api key.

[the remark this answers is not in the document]
fbeaudoincoveo: Yeah I understand that API key authentication would be legit in a typical commerce use case since every user would likely see the same results.

However, I think the API should have a mechanism to extract the email from the search token if present, and fall back on the email request parameter otherwise

If some sort of authentication is already implemented in a solution, it's not a huge extra step for the implementer to add search token authentication as well. And I suppose there are also cases like B2B where not all users would see the same results?

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet.ts
```diff
@@ -167,6 +187,77 @@ export function buildFacet(
...
+    throw loadReducerError;
+  }
+  const controller = buildController(engine);
+
+  return {
+    ...controller,
+
+    get state() {
+      const facets =
+        engine.state.productListing.facets.map((facet) =>
```
antigremlin, line 210: I find it really weird to have dynamically generated facet controllers created by `buildAutomaticFacet` inside this controller's state. How could that possibly work? Also, why isn't the AutomaticFacet type not related to Facet?

Should we take a different approach, like having a facet manager that dynamically creates facet controllers by looking at the listings API response?

packages/headless/src/features/product-listing/product-listing-actions.ts
```diff
@@ -119,6 +143,54 @@ export const fetchProductListing = createAsyncThunk<
...
+    StateNeededByFetchProductListingV2
+  >;
+}
+
+export const fetchProductListingV2 = createAsyncThunk<
+  FetchProductListingV2ThunkReturn,
+  void,
+  AsyncThunkProductListingV2Options<StateNeededByFetchProductListingV2>
+>(
+  'productlistingV2/fetch',
```
antigremlin, line 174: Why do we provide multiple actions for search (execute, fetchPage, fetchMoreResults, fetchFacetValues...) and we want to get by with just 'fetch'?

antigremlin (commented):
Just a couple of questions/ideas so far. This is a huge PR to digest even as it mostly clones existing code.

packages/headless/src/controllers/product-listing/facet/headless-product-listing-facet.ts
Spuffynism: This follows how generated automatic facets work!
> How could that possibly work?

A client would use the facets controller to get a list of facet controllers, and would map this list to UI facet components. Maybe I'm not understanding your question properly?

> Also, why isn't the AutomaticFacet type not related to Facet?

AutomaticFacet is used in the context of generated automatic facets, Facets is used in the context of commerce facets for PLP v2. Once we have a commerce sub-package, this separation should become clearer.

> Should we take a different approach, like having a facet manager that dynamically creates facet controllers by looking at the listings API response?

This approach you propose is the one that is currently used: the state from which we build facet controllers is populated from the product listing response.

packages/headless/src/features/product-listing/product-listing-actions.ts
Spuffynism: For PLP v2, only `productlistingV2/fetch` is used as an action for fetch product listings. This is the action reducers will handle.

The other ones you mention are used in the context of search.

Spuffynism:
Closing this in favor of the productized version of this POC:
- https://github.com/coveo/ui-kit/pull/3129
- https://github.com/coveo/ui-kit/pull/3130
- https://github.com/coveo/ui-kit/pull/3135

PR #3117, opened 2023-08-17
feat(headless): added defineSearchParametersManager SSR controller

btaillon:
https://coveord.atlassian.net/browse/KIT-2605

Keeping @olamothe and @alexprudhomme as reviewers in case they want to take a look. You don't have to check this PR.

Unlike the previous controllers, the search parameters manager is a SSR controller definition *with* props, which means that, in addition to whichever configuration it takes statically in its `define*` function, it takes some configuration at build time (when calling `fetchInitialState` or `hydrateInitialState`).

The reason `defineSearchParametersManager` needs props at build time is because we don't necessarily know `document.location` statically, but we should be able to obtain it at build time.

This is very important in Next.js, since:
 * On the server side, Next.js receives the browser location as props in `page.ts` components (see [`page.js` props](https://nextjs.org/docs/app/api-reference/file-conventions/page#props)).
 * On the client side, Next.js can read standard browser APIs (like `document.location`) or through hooks (e.g., [`useSearchParams`](https://nextjs.org/docs/app/api-reference/functions/use-search-params)).
   * Beware: client components are pre-rendered on the server. During the pre-render stage, browser APIs aren't available.


packages/samples/headless-ssr/src/app/page.tsx
```diff
@@ -1,8 +1,28 @@
...
+        initialState: {
+          parameters: coveoSearchParameters,
+        },
+      },
+    },
+  });
   return <SearchPage ssrState={ssrState}></SearchPage>;
 }
+
+export const dynamic = 'force-dynamic';
```
btaillon, line 28: Next.js is supposed to automatically infer that the component should be rendered dynamically since it takes `searchParams` as an option, but for some reason it didn't, which resulted in prod builds not working.

packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
```diff
@@ -0,0 +1,160 @@
...
+        it('should have the correct search parameters', () => {
+          cy.url().should((href) => {
+            const searchState = new URL(href).searchParams.get(searchStateKey);
+            expect(searchState && JSON.parse(searchState)).to.deep.equal({
+              q: query,
+            });
+          });
+        });
+
+        it('only has two history states', () => {
```
mrrajamanickam-coveo, line 45: What is this testing? Is it dependent on previous tests?

packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
```diff
@@ -0,0 +1,160 @@
...
+    });
+
+    describe('after hydration', () => {
+      beforeEach(() => {
+        cy.visit(getInitialUrl());
+        waitForHydration();
+      });
+
+      it("doesn't update the page", () => {
+        cy.wait(1000);
```
mrrajamanickam-coveo, line 100: Why the hard-code waits? Anything else we can wait on?

packages/samples/headless-ssr/src/app/generic/page.tsx
```diff
@@ -1,8 +1,28 @@
...
+        initialState: {
+          parameters: coveoSearchParameters,
+        },
+      },
+    },
+  });
   return <SearchPage ssrState={ssrState}></SearchPage>;
 }
+
+export const dynamic = 'force-dynamic';
```
mrrajamanickam-coveo, line 28: Would be good to add a comment/doc here

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
```diff
@@ -0,0 +1,111 @@
...
+
+import {SearchParameters} from '@coveo/headless';
+import {
+  SearchParameterManager,
+  SearchParameterManagerState,
+} from '@coveo/headless/ssr';
+import {useCallback, useEffect, useMemo, useReducer, useState} from 'react';
+import {CoveoNextJsSearchParametersSerializer} from '../common/search-parameters-serializer';
+
+interface UseSyncSearchParametersProps {
```
mrrajamanickam-coveo, line 11: Can't we use the `useSearchParams()` nextjs hook?

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
```diff
@@ -0,0 +1,111 @@
...
+  ssrState: SearchParameterManagerState;
```
mrrajamanickam-coveo, line 12: Would implementors have to do something similar as the following as well? Can we package some of this as utils for reuse?

packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
btaillon: It's testing that we didn't do `history.push` multiple times, which could happen if:
* The listener of `searchParametersManager.subscribe` is called twice.
* The hook we use to push new history states is something like `useEffect` and has too many dependencies, causing it to mistakenly fire more times than necessary.

packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
btaillon: I hate hard-coding waits, but unfortunately, here we're waiting for something not to happen, so I don't see much of a way around it.

We could defer the wait to some particular event we hope doesn't synchronously update the page, but I think this would be too specific for an e2e test (but that's just my own reflex, I believe your judgement here would be better than mine).

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
btaillon: Not for this one. The purpose of this hook is to sync the search parameters bi-directionally with the controller. The purpose of `useSearchParams` is to read the URL search parameters.

If what you meant to ask is whether `useUrlSearchParams` could be replaced with `useSearchParams`, I just tested it and that works, so I'll replace it. I don't remember why I didn't do it like that in the first place.

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
btaillon: We could, but those would probably be very framework-specific, so we'd probably need to export them into yet another package (e.g., `headless-next`).

packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
```diff
@@ -0,0 +1,160 @@
...
+          beforeEach(() => cy.go('back'));
+
+          it('should remove the search parameters', () => {
+            cy.url().should((href) =>
+              expect(new URL(href).searchParams.size).to.be.equal(0)
+            );
+          });
+
+          it('should update the page', () => {
+            cy.get('.search-box input').should('have.value', '');
```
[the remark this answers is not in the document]
btaillon: doing that right now, was forgetting

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
btaillon: Update: turns out that `useSearchParams` doesn't refresh when we push new history states, woops.

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
```diff
@@ -0,0 +1,111 @@
+'use client';
```
mrrajamanickam-coveo, line 1: Can add unit tests for these hooks and new controllers (using `@testing-library/react`) - maybe as a followup PR?

mrrajamanickam-coveo (approved):

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
btaillon: Sure

PR #3119, opened 2023-08-18
ci: cache all build tasks

btaillon:
https://coveord.atlassian.net/browse/KIT-2710

Improvement for our 20% tech debt initiative.

This PR should allow us to confidently use `nx run` to run any command in the repository. There's still some start-up time that may make it tempting to use `npm run`, but it shouldn't be as bad as before.

In this PR, I made the following changes:
1. Added caching to all `build` tasks in the repository.
   * By default, new packages won't be cached.
2. Added default caching behaviour that should work for most cases.
3. Replaced caching behaviour in CI (this ensures we don't break caching).
   * We now cache Nx's `.cache/nx` directory instead of the output of every package.
     * Jobs now start without the build of every project. Using Nx, they load or build every package they need.
   * We now cache the entire `node_modules` directory instead of NPM's cache.
     * I can revert this change, but I noticed that it results in a huge reduction in CI run time.
   * Releases are still built from scratch.
4. Added docs on how to cache builds.
5. Made `dev` scripts automatically build dependencies.

TODO:
- [x] Revert `ui-kit.code-workspaces` change.
- [x] Fix bundle size report performance


.github/actions/build/action.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Install, build & cache'
```
btaillon, line 1: The different between this and `setup`, is that this is meant to create the cache, while the setup task now requires the cache.

.github/actions/setup/action.yml
```diff
@@ -1,51 +1,43 @@
...
+      shell: bash
     - name: Install npm
       run: npm i -g npm@9
       shell: bash
     - uses: actions/cache@v3
-      id: build-cache
+      id: npm-cache
       with:
         path: |
-          packages/bueno/dist
```
btaillon, line 16: Goodbye to this list

.github/actions/setup/action.yml
```diff
@@ -1,51 +1,43 @@
...
-          packages/atomic-hosted-page/dist
-          packages/atomic-angular/projects/atomic-angular/dist
-          packages/samples/headless-react
-          packages/samples/iife/www/cdn
-          packages/samples/headless-ssr/.next
-          packages/quantic/force-app/main/default/staticresources/coveoheadless
-          packages/quantic/force-app/main/default/staticresources/coveobueno
-          utils/atomic-storybook/storybook-static
-          !node_modules
-          !**/node_modules
```
btaillon, line 37: Say goodbye to this list.

nx.json
```diff
@@ -55,5 +108,6 @@
       "dependsOn": ["release:build", "release:phase1"]
     }
   },
-  "defaultBase": "master"
+  "defaultBase": "master",
+  "$schema": "./node_modules/nx/schemas/nx-schema.json"
```
btaillon, line 112: added by `nx repair`.

package.json
```diff
@@ -5,7 +5,7 @@
     "postinstall": "husky install && patch-package && nx run-many --target=install:isolated",
     "reset:install": "git checkout origin/master package-lock.json && npm ci",
     "clean:install": "rimraf **/node_modules && npm run reset:install",
-    "dev:atomic": "concurrently \"nx run headless:dev\" \"nx run atomic:dev\"",
+    "dev:atomic": "concurrently \"npm run -w @coveo/headless dev\" \"npm run -w @coveo/atomic dev\"",
```
btaillon, line 8: Without this change, this would run Nx twice in parallel, which could cause conflicts.

Nx has some mechanisms to handle dev tasks, but they are more complex. IMO, it wouldn't be a valuable use of our time.

packages/auth/project.json
```diff
@@ -1 +1,18 @@
-{}
+{
+  "name": "auth",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "targets": {
+    "cached:build": {
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": ["npm run build:bundles", "npm run build:definitions"],
+        "parallel": true,
```
btaillon, line 9: I took some easy-wins parallelizing like this.

ui-kit.code-workspace
```diff
@@ -108,7 +108,6 @@
     "jest.autoRevealOutput": "off",
     "npm.packageManager": "npm",
     "files.exclude": {
-      "**/node_modules/": true,
```
btaillon, line 111: will revert

olamothe (approved):
Lots of, but most of it seems straightforward enough.

Plus, if everything builds on CI, then there's not much else to say !
💯

.github/actions/e2e-atomic-angular/action.yml
```diff
@@ -16,9 +16,8 @@ runs:
       name: Run Cypress
       with:
         browser: chrome
-        command-prefix: 'npx cypress-repeat run -n 3 --until-passes --rerun-failed-only --'
-        working-directory: ./packages/samples/angular
-        start: npm run dev
+        command-prefix: 'npx -w @coveo/atomic-angular-samples cypress-repeat run -n 3 --until-passes --rerun-failed-only --'
```
mrrajamanickam-coveo, line 19: Would it be possible to extract common commands such as `cypress-repeat run -n 3 --until-passes --rerun-failed-only` to common nx config?

package.json
mrrajamanickam-coveo: Can't `nx run-many --target=dev` be used with selected workspaces/projects?

packages/samples/atomic-react/project.json
```diff
@@ -1,3 +1,22 @@
...
+      "outputs": ["{projectRoot}/public/dist"],
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": ["tsc --noEmit", "tsc --module commonjs --noEmit"],
+        "parallel": true,
+        "cwd": "packages/samples/atomic-react"
+      }
+    },
+    "build": {
+      "dependsOn": ["cached:build"],
```
mrrajamanickam-coveo, line 18: Can we mark atomic build as a dep here ?

packages/samples/headless-react/project.json
```diff
@@ -1,3 +1,18 @@
...
+    "cached:build": {
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": ["npm run build:client", "npm run build:server"],
+        "parallel": true,
+        "cwd": "packages/samples/headless-react"
+      }
+    },
+    "build": {
+      "dependsOn": ["cached:build"],
```
mrrajamanickam-coveo, line 14: Can we mark headless build as a dependency for all samples/headless-* pkgs?

packages/samples/headless-ssr/next.config.js
```diff
@@ -1,4 +1,4 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {}
+const nextConfig = {experimental: {appDir: true}};
```
mrrajamanickam-coveo, line 2: Why is this required, esp with the changes in this PR? Doesn't nextjs 13 default to app dir?

mrrajamanickam-coveo:
For this PR
>Total duration
[53m 36s](https://github.com/coveo/ui-kit/actions/runs/5980457404/usage)

vs PRs in master are around ~1h 10m
>[1h 8m 19s](https://github.com/coveo/ui-kit/actions/runs/5968020857/usage) 

Am I looking at this correctly?

.github/actions/e2e-atomic-angular/action.yml
btaillon: Yes and no.

Technically, Nx doesn't allow us to create "default" tasks. However, we have an Nx plugin in our workspace (under `utils/nx-plugin`) which already generates Nx tasks, so it wouldn't be much more work to add a default `cypress-repeat` task.

I'll give it a shot.

packages/samples/atomic-react/project.json
btaillon: There's no need. This task depends on the `atomic-react:cached:build` task, which itself depends on the build task of all dependencies (`^build`).

packages/samples/headless-react/project.json
btaillon: Same as above, there's no need since `headless-react-samples:cached:build` already depends on `^build` (by default, inherited from `nx.json`).

packages/samples/headless-ssr/next.config.js
btaillon: Had an error with Next.js, don't think this bit is necessary anymore, forgot to revert it 😅

btaillon:
> For this PR
> 
> > Total duration
> > [53m 36s](https://github.com/coveo/ui-kit/actions/runs/5980457404/usage)
> 
> vs PRs in master are around ~1h 10m
> 
> > [1h 8m 19s](https://github.com/coveo/ui-kit/actions/runs/5968020857/usage)
> 
> Am I looking at this correctly?

Yes, but most of the duration in both cases seems to be due to the first Atomic e2e test taking like 45+ mins to run.

PR #3121, opened 2023-08-22
fix(atomic): change language code for region specific languages

olamothe:
For these regions specific languages, the correct identifier is with the region in upper case. 

The identifier we use in locale.json is what is used to generate the files name (JSON files), that ultimately gets uploaded to the CDN.

For example, if you set the language on the search interface to `zh-tw` it will try to first resolve `zh-TW.json` from the CDN (and fails to find it since the file is actually named `zh-tw.json`)

And so it will fallback to non region specific language (`zh.json`)

https://coveord.atlassian.net/browse/KIT-2659


louis-bompart (approved):

packages/atomic/cypress/e2e/search-interface.cypress.ts
```diff
@@ -261,7 +263,7 @@ describe('Search Interface Component', () => {
     it('should work with lowercase regions', () => {
       setLanguageAndWait('zh-tw');
 
-      QuerySummarySelectors.text().should('contain', '结果数');
```
olamothe, line 264: Funny enough, this test was actually passing, but was incorrect 😏

The setup is to test zh-tw and here it was displaying the simplified chinese version (ie: the bug this PR fixes)

PR #3122, opened 2023-08-22
feat(headless): support facet exclusion in search parameter manager

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2684

Realized part-way through implementing facet exclusion for the core atomic facet that we weren't supporting excluding facet values in the search parameter manager. Therefore, I added support for this with a new parameter labelled `fExcluded`.


packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
```diff
@@ -201,10 +202,30 @@ function getFacets(state: Partial<SearchParametersState>) {
   return Object.keys(f).length ? {f} : {};
 }
 
+function getExcludedFacets(state: Partial<SearchParametersState>) {
```
olamothe, line 205: This seems a near exact copy paste of `getSelectFacets`, but with the mapping function doing a different operation.

This could be refactored to use one function for both

Suggestion: 

```typescript
function getFacets(
  state: Partial<SearchParametersState>,
  valuesSelector: (currentValues: FacetValueRequest[]) => string[],
  out: keyof SearchParameters
) {
  if (state.facetSet === undefined) {
    return {};
  }

  const facets = Object.entries(state.facetSet)
    .filter(([facetId]) => state.facetOptions?.facets[facetId]?.enabled ?? true)
    .map(([facetId, {request}]) => {
      const facetValues = valuesSelector(request.currentValues);
      return facetValues.length ? {[facetId]: facetValues} : {};
    })
    .reduce((acc, obj) => ({...acc, ...obj}), {});

  return Object.keys(facets).length ? {[out]: facets} : {};
}
```

and then on line 132-133:

```typescript
...getFacets(state, getSelectedValues, 'f'),
...getFacets(state, getExcludedValues, 'fExcluded'),
```

packages/headless/src/features/search-parameters/search-parameter-serializer.ts
```diff
@@ -39,7 +39,13 @@ function serializePair(pair: [string, unknown]) {
     return '';
   }
 
-  if (key === 'f' || key === 'cf' || key === 'sf' || key === 'af') {
+  if (
```
olamothe, line 42: This is reaching the limit of readability for an inline boolean check, IMO.

Suggestion: could extract an utility function for `isSpecificFacetKey` (specific meaning, the Search API type for "normal" or "standard" facet)

olamothe (approved):

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
dmbrooke: Good idea, I'll apply right now 🙂

packages/headless/src/features/search-parameters/search-parameter-serializer.ts
olamothe: I was suggesting having something like this:
```typescript
if(isSpecificFacetKey(key)) {
   return isFacetObject(val) ? serializeFacets(key, val) : '';
}
```

PR #3123, opened 2023-08-23
fix(headless): fetch event source library patched to support working in SF Locker Service

mmitiche:
[SFINT-5159](https://coveord.atlassian.net/browse/SFINT-5159)
fetchEventSource library patched in order to support working in Salesforce Locker Service.

The `AbortController` used by the library can't be used in Salesforce Locker Service, that's why we are creating this patch that will update the usage of this AbortController in the library and allow the GenAi feature to work as expected in Quantic.

to patch this library I used the library: [patch-package](https://www.npmjs.com/package/patch-package?activeTab=readme
)

[SFINT-5159]: https://coveord.atlassian.net/browse/SFINT-5159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "!utils/atomic-storybook"
       ],
       "dependencies": {
-        "patch-package": "6.4.7"
+        "patch-package": "8.0.0"
       },
       "devDependencies": {
         "@actions/core": "1.9.1",
@@ -22530,6 +22530,17 @@
       "version": "1.0.0",
       "license": "MIT"
     },
+    "node_modules/json-stable-stringify": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/json-stable-stringify/-/json-stable-stringify-1.0.2.tgz",
+      "integrity": "sha512-eunSSaEnxV12z+Z73y/j5N37/In40GK4GmsSy+tEHJMxknvqnA7/djeYtAgW0GsWHUfg+847WJjKaEylk2y09g==",
+      "dependencies": {
+        "jsonify": "^0.0.1"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/ljharb"
+      }
+    },
     "node_modules/json-stable-stringify-without-jsonify": {
       "version": "1.0.1",
       "license": "MIT"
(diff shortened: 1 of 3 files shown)


louis-bompart (changes requested):
Additionally:
The solution proposed noop the abort controller, which may impact the functionality of the fetch-event-source packaged.
The AbortController should instead be po*n*yfill (polyfill ain't possible with Locker afaik, pony should be OK tho)

erocheleau:
> Additionally: The solution proposed noop the abort controller, which may impact the functionality of the fetch-event-source packaged. The AbortController should instead be po_n_yfill (polyfill ain't possible with Locker afaik, pony should be OK tho)

The proposed change doesn't noop the `AbortController`, it only makes the library used for EventSource still work when the `AbortController` is not available.

As you can see, this is the main line that changes:

`curRequestController = typeof AbortController === 'undefined' ? null : new AbortController();` 

So this change will have no effect whatsoever whenever the AbortController is actually available aka outside of Locker.

This is our way of making a change that only impacts Locker users and not anything else in other environments.
If you have a better solution, I welcome the suggestions, but this is time sensitive too, we have two different people waiting on this fix. And we couldn't find a better solution.

(We tried using a polyfill, but it breaks the usage of the AbortController in headless. And I don't see how we could ponyfill it for the lib only when the null tolerance behavior is already what is used in the rest of headless.)

erocheleau (approved):

louis-bompart:
> The proposed change doesn't noop the AbortController, it only makes the library used for EventSource still work when the AbortController is not available.

As it is set to null, it is effectively nooped IMO. In Headless we did measure the impact and adapted our code to use or not use an AbortController with a ponyfill.

We did not conceive the library there, and my concern is: does the library work as intended if we noop the AbortController?

If it works as intended, then the "noop ponyfill" (i.e. if API unavailable, then ignore) is OK. If not, a true ponyfill that replicate the AbortController API would be required.

louis-bompart:
@mmitiche https://github.com/coveo/ui-kit/pull/3126
Should do the same but only impact Quantic build artifacts.
There's some issue with the build process on Windows, so I was not able to confirm that the Quantic build artifacts were properly cleaned up tho.

mmitiche:
> @mmitiche #3126 Should do the same but only impact Quantic build artifacts. There's some issue with the build process on Windows, so I was not able to confirm that the Quantic build artifacts were properly cleaned up tho.

I will test this on my end, thank you!

louis-bompart (approved):
LGTM with ifs:
 1. If fetch-event-source worked despite the noop of the AbortController
 2. If #3126 build artifacts are the same as this PR, please merge #3126 in there and LGTM, otherwise, assuming 1 is OK, :shipit:

PR #3124, opened 2023-08-23
fix: refactor focus target to prevent race condition

olamothe:
Rework the `FocusTargetController` decorator that was used throughout Atomic in various component, to instead use a simple "standard" class.

The decorator was hooking on the component lifecycle to perform it's job and to instantiate itself: this can cause race condition, as was described in the accompanying JIRA.

Instead, use a `getter` for all components that requires it, to instantiate the class only once, when necessary (ie: most of the time in the render loop).

A lot of file changes, but the most important/meaningful change is in `accessibility-utils`.

https://coveord.atlassian.net/browse/KIT-2599


louis-bompart (approved):
lgtm, simpler too it seems

PR #3127, opened 2023-08-23
feat(headless): add did-you-mean option to disable automatic query correction

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2618

Now, by dispatching the `disableAutomaticQueryCorrection` action, the controller will no longer dispatch another search action with the `logDidYouMeanClick` analytics.

### DIFF
--- a/packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
+++ b/packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
@@ -42,11 +42,12 @@ describe('did you mean', () => {
     expect(engine.actions).toContainEqual(applyDidYouMeanCorrection('bar'));
   });
 
-  it('should allow to disable automatic query correction', () => {
+  it('should allow to update query correction when automatic correction is disabled', () => {
+    engine.state.didYouMean.queryCorrection.correctedQuery = 'bar';
     engine.state.didYouMean.automaticallyCorrectQuery = false;
     initDidYouMean();
 
     dym.applyCorrection();
-    expect(engine.actions).not.toContainEqual(applyDidYouMeanCorrection('bar'));
+    expect(engine.actions).toContainEqual(applyDidYouMeanCorrection('bar'));
   });
 });
--- a/packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
+++ b/packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
@@ -1,6 +1,7 @@
 import {configuration} from '../../../app/common-reducers';
 import {
   applyDidYouMeanCorrection,
+  disableAutomaticQueryCorrection,
   enableDidYouMean,
 } from '../../../features/did-you-mean/did-you-mean-actions';
 import {didYouMeanReducer as didYouMean} from '../../../features/did-you-mean/did-you-mean-slice';
@@ -46,9 +47,25 @@ describe('did you mean', () => {
     expect(engine.actions).toContainEqual(applyDidYouMeanCorrection('bar'));
   });
 
+  it('should dispatch disableAutomaticQueryCorrection at initialization when specified', () => {
+    initDidYouMean({automaticallyCorrectQuery: false});
+
+    dym.applyCorrection();
+    expect(engine.actions).toContainEqual(disableAutomaticQueryCorrection());
+  });
+
+  it('should not dispatch disableAutomaticQueryCorrection at initialization when specified', () => {
+    initDidYouMean({automaticallyCorrectQuery: true});
+
+    dym.applyCorrection();
+    expect(engine.actions).not.toContainEqual(
+      disableAutomaticQueryCorrection()
+    );
+  });
+
   it('should allow to update query correction when automatic correction is disabled', () => {
     engine.state.didYouMean.queryCorrection.correctedQuery = 'bar';
-    initDidYouMean({automaticallyCorrectQuery: true});
+    initDidYouMean({automaticallyCorrectQuery: false});
 
     dym.applyCorrection();
     expect(engine.actions).toContainEqual(applyDidYouMeanCorrection('bar'));
--- a/packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
+++ b/packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
@@ -8,14 +8,18 @@ import {
   buildMockSearchAppEngine,
   MockSearchEngine,
 } from '../../../test/mock-engine';
-import {buildCoreDidYouMean, DidYouMean} from './headless-core-did-you-mean';
+import {
+  buildCoreDidYouMean,
+  DidYouMean,
+  DidYouMeanProps,
+} from './headless-core-did-you-mean';
 
 describe('did you mean', () => {
   let dym: DidYouMean;
   let engine: MockSearchEngine;
 
-  function initDidYouMean() {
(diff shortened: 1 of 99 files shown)


packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
```diff
@@ -41,4 +41,12 @@ describe('did you mean', () => {
...
     expect(engine.actions).toContainEqual(applyDidYouMeanCorrection('bar'));
   });
+
+  it('should allow to disable automatic query correction', () => {
+    engine.state.didYouMean.automaticallyCorrectQuery = false;
+    initDidYouMean();
+
+    dym.applyCorrection();
+    expect(engine.actions).not.toContainEqual(applyDidYouMeanCorrection('bar'));
+  });
```
btaillon, line 51: What you're currently testing is that it disables manual query correction, which should still be allowed when automatic query corrections are disabled.

Automatic query correction happens automatically in Headless without calling a method. Specifically, it happens in `packages/headless/src/features/search/search-actions-thunk-processor.ts` when the method `processQueryCorrectionsOrContinue` is called.

btaillon (changes requested):

olamothe:
Like Benjamin said, this is not what the ticket is about. 

Feel free to ping me on slack we can discuss live what we actually want to do for this one.

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
dmbrooke: Yup, my mistake! Redid my changes to the files you suggested and it should work as expected now (tested locally) 🙂 

Thanks for pointing this out!

packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -172,6 +172,7 @@ export class AsyncSearchThunkProcessor<RejectionType> {
     }
     if (
       state.didYouMean?.enableDidYouMean === false ||
+      state.didYouMean?.automaticallyCorrectQuery === false ||
```
olamothe, line 175: This multi line boolean check is getting hard to reason about.

I would suggest breaking it up with multiple private module functions that break down the logic very verbosely.
Long, verbose, descriptive is good, in this case.

Suggestion: 
```typescript
if(!state.didYouMean) {
   return null
}

const {enableDidYouMean, automaticallyCorrectQuery} = state.didYouMean;
const {results, queryCorrections} = successResponse;

if(!enableDidYouMean) {
   return null;
}

const shouldBeAutomaticallyRetried = results.length === 0 && queryCorrections.length !==  && automaticallyCorrectQuery;

if(!shouldBeAutomaticallyRetried) {
  return null;
}

// [...]
```

packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -180,9 +181,10 @@ export class AsyncSearchThunkProcessor<RejectionType> {
 
     const originalQuery = this.getCurrentQuery();
     const {correctedQuery} = successResponse.queryCorrections[0];
-    const retried = await this.automaticallyRetryQueryWithCorrection(
-      correctedQuery
-    );
+
+    const retried = state.didYouMean?.automaticallyCorrectQuery
```
olamothe, line 185: This additional check here should not be needed at this point of the function: We should return early in the function when we should not auto retry.

Also, returning the same `fetched` object is... weird. 

It's kind of tricking upstream function call to think there was an automatic query correction applied (the `processor` part of this module), when in fact there were not. 

And then we log analytics as if it was a new query performed. That does not seem right.

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.ts
```diff
@@ -93,6 +98,7 @@ export function buildCoreDidYouMean(engine: CoreEngine): DidYouMean {
         hasQueryCorrection:
           state.didYouMean.queryCorrection.correctedQuery !== '' ||
           state.didYouMean.wasCorrectedTo !== '',
+        automaticallyCorrectQuery: state.didYouMean.automaticallyCorrectQuery,
```
olamothe, line 101: I don't think we should change the state of the controller we return.
Also, there should probably be a new option on the controller.

1- The state we return from a controller is meant to be used by UI piece to control the rendering logic. This `automaticallyCorrectQuery` does not seem relevant: A good hint is the fact that there was no UI change in Atomic, for example.

2- To control the `enableAutomaticQueryCorrection()`, it would be much better from an implementer POV if they could simply pass an option on the controller. And when the controller is built, we check what's the option value, and dispatch the appropriate `disableAutomaticQueryCorrection` action.

packages/headless/src/features/did-you-mean/did-you-mean-actions.ts
```diff
@@ -8,6 +8,14 @@ export const enableDidYouMean = createAction('didYouMean/enable');
 
 export const disableDidYouMean = createAction('didYouMean/disable');
 
+export const disableAutomaticQueryCorrection = createAction(
+  'didYouMean/correction/disable'
```
olamothe, line 12: The identifier could be a bit more verbose, to explain the fact that it's controlling the `automatic` part of the correction.

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.ts
```diff
@@ -21,6 +22,12 @@ import {
 
 export type {QueryCorrection, WordCorrection};
 
+export interface DidYouMeanProps {
+  /**
+   * Whether to automatically correct queries that will return no results.
```
olamothe, line 27: We will need more meat for the documentation. 

It needs to explain what is an "automatically corrected query", when it happens, what's the impact on the UX, etc.

I don't think an implementer would be able to understand what this option will actually change in their page when flipped to true/false.

We also need to explain the default behaviour when the option is left undefined.

olamothe (approved):

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
```diff
@@ -41,4 +45,12 @@ describe('did you mean', () => {
     dym.applyCorrection();
     expect(engine.actions).toContainEqual(applyDidYouMeanCorrection('bar'));
   });
+
+  it('should allow to update query correction when automatic correction is disabled', () => {
+    engine.state.didYouMean.queryCorrection.correctedQuery = 'bar';
+    initDidYouMean({automaticallyCorrectQuery: true});
+
+    dym.applyCorrection();
+    expect(engine.actions).toContainEqual(applyDidYouMeanCorrection('bar'));
```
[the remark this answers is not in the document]
dmbrooke: Ahh I see what you're saying. This test was originally intended to show that query corrections can still be applied, even with automatic correction disabled. However, when changing the "negation" of the `automaticallyCorrectQuery` param (it used to be `disableAutomaticYadayada`), I forgot to change this value. So this test is useless in its current state 😅 

I'll update this test to do what it's supposed to and also add 2 other tests to ensure it does or doesn't dispatch the `disableAutomaticQueryCorrection` action on initialization.

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.ts
dmbrooke: Hmm I see what you mean. I'll try to find a clearer description...

louis-bompart (approved):

PR #3128, opened 2023-08-23
chore(atomic): fix build on windows

louis-bompart:
Empirical fix, I didn't trace back exactly where it goes haywire.
Gut feeling: too much parallelism made using fs.promises, which in turn leads to Windows to :rage4: because lower limits for opened files.

Solution: Reuse the synchronous code on Windows, and let Linux be Linux.
I prefer to not invest more, will prolly be fixed when Oxide (Tailwind not-so-secret Rust sauce that should come 🔜 ™️) will land.

https://coveord.atlassian.net/browse/KIT-2691

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,9 +20,6 @@
         "utils/*",
         "!utils/atomic-storybook"
       ],
-      "dependencies": {
-        "patch-package": "8.0.0"
-      },
       "devDependencies": {
         "@actions/core": "1.9.1",
         "@actions/github": "5.1.1",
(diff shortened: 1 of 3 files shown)


louis-bompart:
@lvu285 can you validate that all's well on your side?

mrrajamanickam-coveo (approved):
Looks like potentially useful fix for upstream as well!

btaillon (approved):

louis-bompart:
> Looks like potentially useful fix for upstream as well!

Agreed, I had planned to raised up the issue and discuss it with tailwind gang today.

lvu285:
@louis-bompart confirmed it fix the issue on my Windows. Thanks!

lvu285 (approved):

PR #3134, opened 2023-08-25
ci(headless): remove web-encoding dependency

dmbrooke:
It turns out that the indexing pipeline encodes _**EVERYTHING**_ into UTF-8, so we don't need to decode the response 🎉


packages/headless/src/api/search/html/text-decoder-polyfill.ts
```diff
@@ -0,0 +1,10 @@
+// This polyfill is needed for Jest.
+export const polyfillTextDecoder = () => {
+  if (typeof window !== 'undefined') {
+    return;
+  }
+  if (!global.TextDecoder) {
+    const util = require('util');
```
olamothe, line 7: Unfortunately this cause other problems 😅 

These types of inline imports will cause webpack, for example to complain. What we do for the crypto polyfill is another tech debt issue that we need to tackle.

So, we shouldn't add more debt with this change.

olamothe (changes requested):

packages/headless/src/api/search/html/text-decoder-polyfill.ts
louis-bompart: So assuming we don't want to be too much of a pain for react-native (Do we?), I'll delay this to v3.
Starting v3, we yank out the polyfill altogether and shift the burden to the user.

The reasoning is that TextDecoder is very widely available and has been for a while. React Native is truly an outlier here.

dmbrooke: That's one thing that I was curious about as an option. So many threads seem to point to the idea that it's up to the implementor to ensure that the TextEncoder/Decoder interfaces are populated in whatever environment they're working in...

That being said, my most recent push introduces a new dependency that exists purely to try and resolve the `TextDecoder` interface across as many environments as possible:
https://www.npmjs.com/package/@polkadot/x-textdecoder
and by extension
https://www.npmjs.com/package/@polkadot/x-global

I'm sure there are still some environments in which this will cause problems, but this at least covers most of the bases, right?

dmbrooke: I reached out to Sami and, as I worried, the Search API _**DOES NOT**_ do any decoding, and simply returns the document as it was indexed. Thus, we either need to introduce a breaking change by removing this, or leave `web-encoding` as is 😢

dmbrooke: After reaching out to Sami, I followed up with Marc Barrette from the indexing pipeline team. It turns out, _**everything in the index**_ is UTF-8 encoded, so we don't need to decode anything!

I've removed the dependencies and updated the logic accordingly 🎉

louis-bompart (approved):

mrrajamanickam-coveo (approved):

PR #3136, opened 2023-08-25
fix(headless): toggleSelected should toggle both ways (select&deselect)

louis-bompart:
The category facet is the only facet that does not actually toggle off the value. With that change, now it does.

https://coveord.atlassian.net/browse/KIT-2695

### DIFF
--- a/packages/headless/src/features/facets/category-facet-set/category-facet-set-slice.test.ts
+++ b/packages/headless/src/features/facets/category-facet-set/category-facet-set-slice.test.ts
@@ -589,6 +589,25 @@ describe('category facet slice', () => {
           expect(currentValues).toEqual([parent]);
         });
       });
+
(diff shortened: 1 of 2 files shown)


dmbrooke (approved):
Simple enough and it seems to work exactly as expected 🙂 

LGTM

louis-bompart:
Actually won't do:
After tinkering more with it, this creates issues that are a bit complex to manage and are not worth the risks:
For example,, let's say you have a selected value Y that is the children of a value X, and that Y do have like 5 children.
With the current change, in the UI, you end up with no facet value selected, but all the hierarchy deployed.
Which is not really the goal here.
So for now, I'll leave a note on the Jira, but I revised my opinion and think that with the current UI/UX of the Category Facet, this changes is a net negative.

PR #3137, opened 2023-08-26
fix(headless): consider only selected value ancestry for analytics, history and breadcrumbs 

louis-bompart:
Essentially, this PR replaces *both* `getActiveValueFromValueTree` and `partitionIntoParentsAndValues` (partially) with `findActiveValueAncestry`.

The first one is a method I introduced in #3105, and I could have left it there. However, in an effort to ease the deprecation of `categoryFacetState.parents`, I decided to use it in the CategoryFacetController.

However, `parents` is not entirely based on the notion of whether a value is selected or not (which IMHO is not the best because when a value is selected, it represents user intent) contrary to the new `selectedValueAncestry`, so I decided to create the latter to replace the former definitively in v3.

The second one is the historical method that doesn't work well with "multi-root-values" hierarchies, which leads to similar symptoms (i.e. children's root values or whatnot leaking into the parents' hierarchy, leading to bad analytics, bad history slice, and so on)

The logic of `findActiveValueAncestry` is simple:
 - Use the logic of `getActiveValueFromTree` BUT as you queue up new values, keep a note of what is their parent in a Map, where the child value is the key and the parent value is the value.
 - Once the activeValue is found, "rewind" the ancestry using the Map and compute the ancestry as we go
 
 
 ------
 
 KIT-2696

### DIFF
--- a/packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
+++ b/packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
@@ -23,7 +23,10 @@ import {
   partitionIntoParentsAndValues,
 } from '../../../../features/facets/category-facet-set/category-facet-utils';
 import {CategoryFacetSortCriterion} from '../../../../features/facets/category-facet-set/interfaces/request';
-import {CategoryFacetValue} from '../../../../features/facets/category-facet-set/interfaces/response';
+import {
+  CategoryFacetValue,
+  CategoryFacetValueCommon,
+} from '../../../../features/facets/category-facet-set/interfaces/response';
 import {categoryFacetSearchSetReducer as categoryFacetSearchSet} from '../../../../features/facets/facet-search-set/category/category-facet-search-set-slice';
 import {defaultFacetSearchOptions} from '../../../../features/facets/facet-search-set/facet-search-reducer-helpers';
 import {isFacetLoadingResponseSelector} from '../../../../features/facets/facet-set/facet-set-selectors';
@@ -50,6 +53,7 @@ import {
 } from './headless-core-category-facet-options';
 
 export type {
+  CategoryFacetValueCommon,
   CategoryFacetValue,
   CategoryFacetOptions,
   CategoryFacetSearchOptions,
--- a/packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
+++ b/packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
@@ -19,8 +19,7 @@ import {categoryFacetRequestSelector} from '../../../../features/facets/category
 import {defaultCategoryFacetOptions} from '../../../../features/facets/category-facet-set/category-facet-set-slice';
 import {categoryFacetSetReducer as categoryFacetSet} from '../../../../features/facets/category-facet-set/category-facet-set-slice';
 import {
-  getActiveValueAncestry,
-  getActiveValueFromValueTree,
+  findActiveValueAncestry,
   partitionIntoParentsAndValues,
 } from '../../../../features/facets/category-facet-set/category-facet-utils';
 import {CategoryFacetSortCriterion} from '../../../../features/facets/category-facet-set/interfaces/request';
@@ -393,11 +392,10 @@ export function buildCoreCategoryFacet(
       const isHierarchical =
         valuesAsTrees.some((value) => value.children.length > 0) ?? false;
       const {parents, values} = partitionIntoParentsAndValues(response?.values);
-      const activeValue = getActiveValueFromValueTree(valuesAsTrees);
-      const ancestryMap = new Map<CategoryFacetValue, CategoryFacetValue>();
-      const selectedValueAncestry = activeValue
-        ? getActiveValueAncestry(activeValue, ancestryMap)
-        : [];
+      const selectedValueAncestry = findActiveValueAncestry(valuesAsTrees);
+      const activeValue = selectedValueAncestry.length
+        ? selectedValueAncestry[selectedValueAncestry.length - 1]
+        : undefined;
       const hasActiveValues = !!activeValue;
       const canShowMoreValues =
         activeValue?.moreValuesAvailable ??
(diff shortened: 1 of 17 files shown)


packages/headless/src/features/facets/facet-set/facet-set-analytics-actions-utils.test.ts
```diff
@@ -106,7 +106,7 @@ describe('facet-set-analytics-action-utils', () => {
           field,
           currentValues: [
             buildMockCategoryFacetValueRequest({
-              state: 'selected',
+              state: 'idle',
```
louis-bompart, line 109: I decided to changes this, because I don't think it make sense:
if the child value is selected, the parent value should not.

At least, that's what I saw in the response & requests when playing with atomic in stackblitz

packages/headless/src/features/facets/category-facet-set/category-facet-utils.ts
```diff
@@ -33,16 +33,46 @@ export function partitionIntoParentsAndValues<
...
+
+function getActiveValueAncestry<TValue extends GenericCategoryFacetValue>(
+  activeValue: TValue | undefined,
+  valueToParentMap: Map<TValue, TValue>
+): TValue[] {
+  const activeValueAncestry: TValue[] = [];
+  if (!activeValue) {
+    return [];
+  }
+  let lastParent: TValue | undefined = activeValue;
```
louis-bompart, line 72: `undefined` because Map.get returns undefined if the key ain't in the map.
This is used in the while as our cue to break the loop

packages/headless/src/features/facets/category-facet-set/category-facet-utils.test.ts
louis-bompart, line 1: Not really useful to compare it side by side, two differents test suite really. The former is defunct because the method ain't exported anymore

packages/headless/src/features/facets/category-facet-set/category-facet-utils.ts
```diff
@@ -33,16 +33,46 @@ export function partitionIntoParentsAndValues<
...
   const valueToVisit = [...valuesAsTrees];
   while (valueToVisit.length > 0) {
     const currentValue = valueToVisit.shift()!;
     if (currentValue.state === 'selected') {
       return currentValue;
     }
-    valueToVisit.unshift(...currentValue.children);
+    if (ancestryMap) {
+      for (const childValue of currentValue.children) {
+        ancestryMap.set(childValue as TValue, currentValue);
```
olamothe, line 56: These cast are a bit suspicious 🤔 From what I can understand/grasp, it's because there's two types which do not perfectly overlap. These types of cast can be error prone.

I'm wondering if we could do something by changing `GenericCategoryFacetValue` to be a type where we just pick what's needed to perform the partition/identify the ancestry, instead of trying to create a type that a full union of the two.

Something like this maybe ?
```
type GenericCategoryFacetValue = Pick<
  CategoryFacetValue | CategoryFacetValueRequest,
  'state' | 'children' | 'value'
>;
```

I'm not sure that's doable, but food for thoughts.

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
```diff
@@ -170,10 +170,17 @@ export interface CoreCategoryFacetState {
...
+   * @deprecated use `selectedValueAncestry` instead.
    */
   values: CategoryFacetValue[];
 
+  /**
+   * The selected facet values ancestry.
+   * The first element is the "root" of the selected value ancestry tree.
+   * The last element is the selected value itself.
+   */
+  selectedValueAncestry: CategoryFacetValue[];
```
olamothe, line 182: Do we need the `selected` keyword here ? Since we don't allow exclusion, and the only way you can have ancestor here is by selecting a value in the first place.

Would `valueAncestry` be a correct replacement ?

olamothe (approved):

dmbrooke (approved):
Looks good!

packages/headless/src/features/facets/category-facet-set/category-facet-utils.ts
louis-bompart: I lazied out a bit on this one (friday amarite), because I didn't want to refactor too much the response/request interfaces. This is now done in 827b607

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
louis-bompart: We don't need it.
However, technically, if you select a n-th level value, where n>1, you have multiple ancestor: indeed, each n-th leveled value other than 0 has ancestor.

So even if it's a bit redundant, I prefer to be abundantly clear than fuzzy, so IMHO, I'd keep the selected.

packages/headless/src/features/facets/category-facet-set/category-facet-utils.ts
```diff
@@ -33,16 +33,46 @@ export function partitionIntoParentsAndValues<
...
+  valuesAsTree: T[]
+): T[] {
+  const ancestryMap = new Map<T, T>();
+  const activeValue = getActiveValueFromValueTree(valuesAsTree, ancestryMap);
+  return activeValue ? getActiveValueAncestry(activeValue, ancestryMap) : [];
+}
+
+function getActiveValueFromValueTree<TValue extends GenericCategoryFacetValue>(
+  valuesAsTrees: TValue[],
+  ancestryMap?: Map<TValue, TValue>
```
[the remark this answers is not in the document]
louis-bompart: - [ ] remove opt

packages/headless/src/features/facets/category-facet-set/category-facet-utils.ts
```diff
@@ -33,16 +33,46 @@ export function partitionIntoParentsAndValues<
...
+): TValue | undefined {
   const valueToVisit = [...valuesAsTrees];
   while (valueToVisit.length > 0) {
     const currentValue = valueToVisit.shift()!;
     if (currentValue.state === 'selected') {
       return currentValue;
     }
-    valueToVisit.unshift(...currentValue.children);
+    if (ancestryMap) {
```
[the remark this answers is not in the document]
louis-bompart: Yes, but good catch 🥇 
Both your comments are because I initially left the `getActiveValueFroMValueTree` exposed, and didn't want to break its signature.
This is not the case any more so:
 - I can be sure that the ancestryMap exist
 - I can return it instead

PR #3138, opened 2023-08-28
chore(doc-tools): log failing entrypoint

louis-bompart:
When adding an interface, it's confusing to know all the nooks and crannies where you should expose it.
This helps by telling you where you forgot to export it. (it is quite lazy, but it works).

------

KIT-2698


mrrajamanickam-coveo (approved):
👍

PR #3139, opened 2023-08-28
fix(atomic): support hierarchical facet values with multiple roots

louis-bompart:
Essentially, allow the rendering of this kind of structure and a lot of refactor
image

### DIFF
--- a/packages/atomic/cypress/e2e/facets/category-facet/category-facet-selectors.ts
+++ b/packages/atomic/cypress/e2e/facets/category-facet/category-facet-selectors.ts
@@ -40,6 +40,9 @@ export const CategoryFacetSelectors = {
   values() {
     return this.shadow().find('[part="values"]');
   },
+  linkValue() {
+    return this.shadow().find('[part~="value-link"]');
+  },
   childValue() {
     return this.values().find('[part~="value-link"]');
   },
--- a/packages/atomic/cypress/e2e/facets/category-facet/category-facet-selectors.ts
+++ b/packages/atomic/cypress/e2e/facets/category-facet/category-facet-selectors.ts
@@ -41,7 +41,7 @@ export const CategoryFacetSelectors = {
     return this.shadow().find('[part="values"]');
   },
   childValue() {
-    return this.shadow().find('[part~="value-link"]');
+    return this.values().find('[part~="value-link"]');
   },
   childValueLabel() {
     return this.childValue().find('[part="value-label"]');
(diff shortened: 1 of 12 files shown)


olamothe (approved):
I think everything is good since it's a lot of moving code around. 

I tried to pay special attention that all the CSS `parts` stayed the same (since that would be breaking). Everything seems fine on that front.

packages/atomic/cypress/e2e/facets/category-facet/category-facet-selectors.ts
```diff
@@ -40,9 +40,12 @@ export const CategoryFacetSelectors = {
   values() {
     return this.shadow().find('[part="values"]');
   },
-  childValue() {
+  linkValue() {
     return this.shadow().find('[part~="value-link"]');
   },
+  childValue() {
+    return this.values().find('[part~="value-link"]');
```
olamothe, line 47: Does it seem a problem that the differentiator (css selector) between both type of value is * that * subtle (for implementers) ?

louis-bompart: Good point, I would not bring changes here because, IMO, it may be interesting for v3 to rationalize the parts and their names with 2 things in mind:
 - We can now have the hierarchy described in the PR descriptions
 - We need to support all existing CSS customizations
 
Notably, I do not like the fact that I had to leave a "split" in the rendering logic depending on whether the response is hierarchical or not (as all response are hierarchical for the API I reckon, the flat ones are just 'hierarchical without children') . I would have preferred the rendering to always be "hierarchical", and that the lack of "hierarchy" comes from the lack of children, and nothing else (i.e. render children if any)

This was mostly caused by the existing parts and how they work.

I would suggest for v3:
 1.  "raze' existing parts
 2. ask ourselves "ok what makes sense to exposes"
 3. Compare with v2 & try to fill the obvious gaps
 4. Wait for complaints.

louis-bompart:
Without the CSS changes:
image
image

With the CSS changes:
image
image

This IMO is the least worst solution, but will maybe shambles some UI for some maybe

louis-bompart:
CSS Changes, Plan C:
image
image

> If a value is only a passing node (i.e. it have one children and this children is not a leaf), then do not increase the hierarchical padding

dmbrooke:
> CSS B: always pad left
image image

I like option B the most. I think the left padding makes the hierarchical level of each value more clear.
That being said, when this indentation is mixed with some values having < prepended to them, it throws things off for me. Like for this example:
image
Is C at the same level as D? Or is C at the same level as B, but the < pushes it far enough right to make it align perfectly with D? This ambiguity is still quite confusing to me

PR #3141, opened 2023-08-29
fix(headless): bump coveo.analytics to fix click tracking issue on ios safari

olamothe:
Here is the accompanying PR on coveo.analytics that describe the issue in more details: https://github.com/coveo/coveo.analytics.js/pull/435

https://coveord.atlassian.net/browse/KIT-2699

### DIFF
--- a/packages/quantic/cypress/e2e/common-expectations.ts
+++ b/packages/quantic/cypress/e2e/common-expectations.ts
@@ -1,3 +1,7 @@
+import {
+  CyHttpMessages,
+  Interception,
+} from '../../../../node_modules/cypress/types/net-stubbing';
 import {ComponentErrorSelector, should} from './common-selectors';
 
 export function logUaEvent(
@@ -9,7 +13,7 @@ export function logUaEvent(
   return cy
     .wait(requestAlias)
     .then((interception) => {
-      const analyticsBody = interception.request.body;
+      const analyticsBody = getAnalyticsBodyFromInterception(interception);
 
       Object.keys(bodyData).forEach((key: string) => {
         expect(analyticsBody).to.have.property(key, bodyData[key]);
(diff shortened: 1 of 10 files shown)


louis-bompart (approved):

y-lakhdar (approved):

packages/quantic/cypress/e2e/common-expectations.ts
```diff
@@ -46,3 +48,26 @@ export function ComponentErrorExpectations(selector: ComponentErrorSelector) {
...
+export function getAnalyticsBodyFromRequest(
+  req: CyHttpMessages.IncomingRequest
+) {
+  // Some analytics request won't be application/json, but will be form-url-encoded (specifically, click events).
+  // Need to decode them if that's the case
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let analyticsBody: Record<string, any>;
+  if (typeof req.body === 'string') {
+    analyticsBody = JSON.parse(
```
erocheleau, line 65: Do you want to re-use your utils parseClickEventRequest here?

packages/quantic/cypress/e2e/default-2/result-list/folded-result-list/folded-result-list-expectations.ts
```diff
@@ -30,8 +31,8 @@ function logFoldingActionEvent(
   const {title, uri, position, documentId} = result || {};
   cy.wait(eventName)
     .then((interception) => {
-      const analyticsBody = interception.request.body;
-      const customData = analyticsBody?.customData;
+      const analyticsBody = getAnalyticsBodyFromInterception(interception);
+      const customData = analyticsBody.customData;
```
erocheleau, line 35: `req.body` can be null no? 

in common-expectations.ts :
`analyticsBody = req.body;` if `req.body` is missing, it's possible that `analyticsBody.customData` will throw? Is that what we want?

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet-expectations.ts
```diff
@@ -130,7 +131,7 @@ function smartSnippetExpectations(selector: SmartSnippetSelector) {
       const {title, uri, permanentId} = document;
       cy.wait(InterceptAliases.UA.OpenSmartSnippetSource)
         .then((interception) => {
-          const analyticsBody = interception.request.body;
+          const analyticsBody = getAnalyticsBodyFromInterception(interception);
           const customData = analyticsBody?.customData;
```
erocheleau, line 135: It wasn't changed here to remove the `?` so what's the expected behavior?

erocheleau (approved):
Just small questions about the expectation if the request body can be null or not, and if we want the code to throw in this case or force the code to handle a potential undefined?

packages/quantic/cypress/e2e/default-2/result-list/folded-result-list/folded-result-list-expectations.ts
olamothe: Either it throws here, or fails the test couple lines below. Pretty much same end result.

PR #3143, opened 2023-08-29
feat(atomic): support facet exclude clear

y-lakhdar:
Display "Clear all" button when a value is excluded.
image

https://coveord.atlassian.net/browse/KIT-2687

### DIFF
--- a/packages/atomic/src/components/common/facets/facet-common.tsx
+++ b/packages/atomic/src/components/common/facets/facet-common.tsx
@@ -386,8 +386,7 @@ export class FacetCommon {
   }
 
   private get numberOfSelectedValues() {
-    return this.facet.state.values.filter(({state}) => state === 'selected')
-      .length;
+    return this.facet.state.values.filter(({state}) => state !== 'idle').length;
   }
 
   private renderSearchInput() {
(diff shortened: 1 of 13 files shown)


packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -386,8 +386,7 @@ export class FacetCommon {
   }
 
   private get numberOfSelectedValues() {
```
dmbrooke, line 388: I'm wondering if it it would be more appropriate to update the name of this getter to `numberOfActiveValues` now that it includes excluded values... You would have to change it in quite a few places, but it could easily be done using 1 big find/replace 🙂

dmbrooke (approved):
LGTM

packages/atomic/src/components/common/facets/facet-common.tsx
y-lakhdar: Good point. Would `numberOfSelectedValues` make more sense?

olamothe: > Would numberOfSelectedValues make more sense?

Hum isn't that the name already ?

olamothe (approved):

packages/atomic/src/components/common/facets/facet-common.tsx
y-lakhdar: sorry, I meant `numberOfActiveValues`

btaillon (approved):

PR #3144, opened 2023-08-30
feature(SFINT-5141): Added the expandable textArea searchbox in Quantic

erocheleau:


packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -58,6 +64,14 @@ export default class QuanticSearchBox extends LightningElement {
...
    */
   @api numberOfSuggestions = 5;
+  /**
+   * Whether to render the search box using a [textarea](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) element.
+   * The resulting component will expand to support multi-line queries.
+   * @api
+   * @type {boolean}
+   * @defaultValue false
+   */
+  @api textarea = false;
```
SimonMilord, line 74: should we name this property to something more in-line with the feature? ie: `expandableSearchbox` or `expandableSearchboxEnabled` ?

packages/quantic/force-app/main/default/lwc/quanticSearchBox/templates/defaultSearchBox.html
```diff
@@ -0,0 +1,68 @@
+<!-- sldsValidatorIgnore -->
```
SimonMilord, line 1: Why do we need this?

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
erocheleau: It's in line with the same property name from Atomic.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/templates/defaultSearchBox.html
erocheleau: It was asking if I wanted to ignore the "mobile" slds validation... Guess it just adds this to ignore ALL validations

erocheleau: It just highlights pretty much all the HTML with warnings so it's pretty useless.

erocheleau:
I realize the situation with the duplicated code between the QuanticSearchBox and the QuanticStandaloneSearchBox is not ideal. 

I would like to refactor it using composition/component extension, but this is out of scope for this PR.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.css
```diff
@@ -0,0 +1,3 @@
+.quantic-suggestions-list {
+  margin-top: 0;
```
SimonMilord, line 2: why do we want this here is it to reset default margin or something?

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
```diff
@@ -284,16 +309,43 @@ export default class QuanticStandaloneSearchBox extends NavigationMixin(
...
+
+  onTextAreaInput() {
+    this.adjustTextAreaHeight();
+  };
+
+  adjustTextAreaHeight() {
+    if (!this.textarea) {
+      return;
+    }
+    this.input.value = this.input.value.replace(/\n/g, '');
```
SimonMilord, line 328: nice touch!

packages/quantic/force-app/main/default/lwc/searchBoxStyle/searchBoxStyle.css
```diff
@@ -28,7 +28,11 @@
 .searchbox__submit-button {
   width: 46.75px;
   border-radius: 0;
-  height: 100%;
+  height: 48px;
```
SimonMilord, line 31:
```suggestion
  height: 3rem;
```

packages/quantic/force-app/main/default/lwc/searchBoxStyle/searchBoxStyle.css
```diff
@@ -28,7 +28,11 @@
...
+}
+
+.searchbox__clear-button {
+  height: 48px;
```
SimonMilord, line 35:
```suggestion
  height: 3rem;
```

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -5,8 +5,14 @@ import {
   initializeWithHeadless,
```
SimonMilord, line 5: should we maybe add unit tests for the quanticSearchBox to cover whether to render the textarea or not? It currently has 1 test suite in the file which is pretty basic.

SimonMilord (approved):
Looks good! Just a few comments, but other than that GG.  Point noted about the code duplication, maybe we can add the JIRA # to the PR for future reference (unless it messes up the bot)

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.css
erocheleau: There was a margin in the Quantic Query suggest look which created a gap between the suggestions and the bottom of the searchbox. It looked weird so after discussion with UX we removed it. It's connected in Atomic as well so I dunno why we had it.

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
erocheleau: Yea I dunno if it'll be a problem in the future or not, it doesn't allow you to input structured text, it eats up your line breaks, but the line breaks also break the display of the searchbox when it is collapsed.

packages/quantic/force-app/main/default/lwc/searchBoxStyle/searchBoxStyle.css
erocheleau: Nope, the other heights are all set to exactly 48px so it needs to be consistent. Don't know where this value comes from but it is what it is.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
erocheleau: Well it's tested with the Cypress tests for the standalone searchbox. We probably should be adding tests for the searchbox overall, but since we didn't have any I didn't take the time to add new ones to the searchbox alone, only on the standalone version which is exactly the same code.

I think in the future when we re-factor those two components so they actually are the same thing, we can then write tests to test both.

PR #3146, opened 2023-08-31
feat(Stencil samples): change source badge bg

lgcarrier:


btaillon:
<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td>

image
 <td>

image
</table>

I'm not sure what's the motivation behind this change, since it's only in a sample. Assuming there's a good reason and we wish to meet WCAG, this unfortunately has low contrast against the white background.

See:
image

I think changing the text color to white would be easier.

lgcarrier:
@btaillon The main change is to reduce the visibility of "source badge", which are not very important, by changing their color to have a lower contrast with the background. This way, the users can focus more on the relevant information in the results.
I modified the "sample code" as I think this is the template we are using when creating a new Atomic Hosted Search Page.

dmbrooke (approved):
Simple enough, LGTM 👍

louis-bompart:
The change makes sense, however, it would be preferable that you find a way to do so while maintaining or improving the accessibility, can you please revisit this?

Also, by Atomic Hosted Search Page, do you mean the "Simple Builder" ones or the "Custom Deployment" ones?

btaillon:
> @btaillon The main change is to reduce the visibility of "source badge", which are not very important, by changing their color to have a lower contrast with the background. This way, the users can focus more on the relevant information in the results. I modified the "sample code" as I think this is the template we are using when creating a new Atomic Hosted Search Page.

I see. Here's my suggestion then:
```css
background-color: var(--atomic-background);
border-color: var(--atomic-neutral);
border-width: 1px;
```
image

louis-bompart:
@lgcarrier, as we discussed this will change the sample for folks cloning the repo, but that's all.
I invite you to open a Jira so we can investigate exactly where/what needs to change and have a good tracability

louis-bompart:
Converting to draft; we'll see what we do once we establish the problem and study the solutions

louis-bompart:
Stale

PR #3147, opened 2023-08-31
feat(commerce): create PLP v2 product listing controller

Spuffynism:
The implementation is almost identical to the [PLP v1 product listing controller](https://github.com/coveo/ui-kit/blob/4447884bb17a5c9998a9004ccca4a27208290a9a/packages/headless/src/controllers/product-listing/headless-product-listing.ts), with the exception that we do not expose functionality to set `additionalFields`.

[CAPI-83]

[CAPI-83]: https://coveord.atlassian.net/browse/CAPI-83?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
```diff
@@ -0,0 +1,132 @@
...
+const optionsSchema = new Schema({
+  url: new StringValue({
+    required: true,
+    url: true,
+  }),
+});
+
+export interface ProductListingOptions {
+  /**
+   * The initial URL used to retrieve the product listing.
```
fbeaudoincoveo, line 27:
```suggestion
   * The URL used to retrieve the product listing.
```

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  options?: ProductListingOptions;
+}
+
+/**
+ * The `ProductListing` controller allows the end user to configure and retrieve product listing data.
+ */
+export interface ProductListing extends Controller {
+  /**
+   * Changes the current URL used to retrieve product listing.
```
fbeaudoincoveo, line 44:
```suggestion
   * Changes the URL used to retrieve the product listing.
```

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
```diff
@@ -0,0 +1,132 @@
...
+   * @param url - The new URL.
+   */
+  setUrl(url: string): void;
+
+  /**
+   * Refreshes the product listing.
```
fbeaudoincoveo, line 50:
```suggestion
   * Fetches the product listing content.
```

fbeaudoincoveo (commented):
🥇

packages/headless/src/features/product-listing/v2/product-listing-v2-slice.test.ts
```diff
@@ -13,12 +16,24 @@ describe('product-listing-v2-slice', () => {
...
     state = getProductListingV2InitialState();
   });
+
   it('should have an initial state', () => {
     expect(productListingV2Reducer(undefined, {type: 'foo'})).toEqual(
       getProductListingV2InitialState()
     );
   });
 
+  it('should allow to set the product recommendations skus', () => {
```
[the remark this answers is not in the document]
fbeaudoincoveo: You marked this one as resolved without making any change, however unless I misunderstand what this test is about, it seems to me like the description should be something like "should allow to set the product listing url"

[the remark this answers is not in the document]
Spuffynism: Turns out the PLP v1 implementation had a test with the same faulty name, and I fixed that test name instead of this one which you commented on. 🥴 

Should be good now! 😅

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
```diff
@@ -0,0 +1,132 @@
...
+} from '../../controller/headless-controller';
+
+const optionsSchema = new Schema({
+  url: new StringValue({
+    required: true,
+    url: true,
+  }),
+});
+
+export interface ProductListingOptions {
```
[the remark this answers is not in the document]
Spuffynism: I think it would make more sense to have context labels set in the context controller. What do you think? 

It feels strange to have the context labels set through the product listings controller.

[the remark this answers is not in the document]
fbeaudoincoveo: I might be misunderstanding what labels are intended for, but I thought they were specific to each product listing page, i.e., not "global" context that would be relevant to an entire commerce app.

If my understanding is correct, then I think labels should be on the product listing controller. Otherwise, they should be on the context controller.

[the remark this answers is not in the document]
olamothe: If labels are indeed something that are "unique per product listing page", and that you don't need to have any impact outside of the product list per say (ie: if you have labels X for a specific product listing page, it has no impact on the cart), then they are "local" context that need to be kept local (ie: only on the product listing controller).

I guess, it would help if you could you explain what label are for ?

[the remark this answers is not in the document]
olamothe: Ok, after reading the RFC about what they represent, its definitely something "local" to a specific page, not global to the full website.

As such, it looks like what Fred is suggesting is the right approach.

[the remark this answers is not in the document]
fbeaudoincoveo: After reading the RFC myself, I tend to agree with @samisayegh 's comment here: https://docs.google.com/document/d/1HIlPbSVm9luLfObJvoIyCvTpE37O-ts-bQGO8Np5cKo/edit?disco=AAAA4HbWQX4

I.e., it also seems to me like labels are a pre-emptive escape hatch. This leads me to believe that we should not expose this parameter in Headless out of the gate after all. We can wait until someone actually needs it, which might never happen for all we know.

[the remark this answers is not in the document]
Spuffynism: I'll resolve this conversation. Let's add support for context labels as the need comes up.

Thanks y'all!

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const options = {
+    ...props.options,
+  };
+
+  validateOptions(engine, optionsSchema, options, 'buildProductListing');
+
+  dispatch(
+    setProductListingUrl({
+      url: options.url!,
```
olamothe, line 97: Since the controller will simply "not work" if the URL is not defined (ie: we won't even be able to query the API), then the URL should be a required prop on the controller.

ie: Make props and `options` in `ProductListingProps` non optional/mandatory. This will remove the force cast you need to do here.

And then add some validation that there's actually a valid URL (non-empty string, with the bueno package) if an implementer use JavaScript instead of TS: In this case they would not get any compiler validation that some value is mandatory, and would end up with some cryptic error message down the line.

You can see this pattern in some other controller, if you search for the function `validateOptions` in `validate-payload.ts`

We can leave the `setURL` function below, but it should be "impossible" to arrive at this line here in the code with no URL.

olamothe (approved):

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
```diff
@@ -0,0 +1,132 @@
...
+import {validateOptions} from '../../../utils/validate-payload';
+import {
+  buildController,
+  Controller,
+} from '../../controller/headless-controller';
+
+const optionsSchema = new Schema({
+  url: new StringValue({
+    required: true,
+    url: true,
```
olamothe, line 21: Is it an absolute requirement (from the API point of view) that this must be a valid URL that follows the regex defined in `string-value.ts` (in bueno) ?

If not, I would suggest chaning the check to `required: true` and `emptyAllowed: false`

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
Spuffynism: Good point! I was following the existing product listing controller implementation by having the url be optional but there's no need to do that.

Commit incoming!

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
Spuffynism: Good question, lemme check with the team :loading:

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
Spuffynism: Actually, by not validating the options we receive, we delay the validation to the `setProductListingUrl` action, which only validates the payload and doesn't throw if the url's invalid.

Don't we want to throw if an invalid url is fed to the controller?

Spuffynism: The function execution does not stop if the url is empty. This will rightfully log a `SchemaValidationError` error, but the controller will still be created, and be in an unusable state, as the `url` will be the one from `getProductListingInitialState().url` :
```js
const engine = buildCommerceEngine({
  configuration: {
    ...getSampleSearchEngineConfiguration(),
  },
});

const productListing = buildProductListing(engine, {
  options: {
    url: ''
  }
})
console.log(productListing);
```

To prevent this, I think we should have validation on the controller options, and not only in the action. This would completely prevent an invalid controller from being created. 

Am I misunderstanding the use case?

olamothe: My bad, you are right !

olamothe (approved):

packages/headless/src/controllers/commerce/product-listing/headless-product-listing.ts
Spuffynism: We opted for only requiring the url to be set at all, as the format the API supports is quite lax!

fbeaudoincoveo (approved):

PR #3149, opened 2023-08-31
feat(headless): add excludeFacetSearchResult search action

y-lakhdar:
Support exclusion of facet values from performing a facet search.

I also refactored the tests by removing duplicated code and by using `describe.each` and `it.each` instead.

(Added 16 new tests)

https://coveord.atlassian.net/browse/KIT-2706


btaillon (approved):

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
```diff
@@ -230,6 +233,29 @@ export const facetSetReducer = createReducer(
         facetRequest.freezeCurrentValues = true;
         facetRequest.preventAutoSelect = true;
       })
+      // TODO: add tests
```
olamothe, line 236: This TODO should be removed since you've added tests ?

olamothe (approved):

packages/headless/src/features/facets/facet-set/facet-set-slice.ts
y-lakhdar: yes. forgot to remove it

PR #3153, opened 2023-09-05
ci: downgraded nx version

btaillon:
https://coveord.atlassian.net/browse/KIT-2716

Upgrading the version of Nx caused pre-releases to fail (example: https://github.com/coveo/ui-kit/actions/runs/6053065974/job/16429178336).


louis-bompart:
All unknowns licenses are MIT badly refed by package owner, owel

louis-bompart (approved):

PR #3154, opened 2023-09-06
ci: fix typos in atomic-angular task definitions

btaillon:
https://coveord.atlassian.net/browse/KIT-2718

Somehow didn't pay enough attention to my changes to the task definitions of Angular and made some bad copy-and-pastes, which are breaking the release.

### DIFF
--- a/packages/atomic-angular/project.json
+++ b/packages/atomic-angular/project.json
@@ -7,7 +7,7 @@
   "targets": {
     "cached:build": {
       "outputs": ["{projectRoot}/projects/atomic-angular/dist"],
-      "dependsOn": ["^build", "clean"],
+      "dependsOn": ["^build"],
       "executor": "nx:run-commands",
       "options": {
         "commands": ["npm run build:bundles", "npm run build:assets"],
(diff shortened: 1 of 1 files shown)


packages/atomic-angular/project.json
```diff
@@ -7,7 +7,7 @@
   "targets": {
     "cached:build": {
       "outputs": ["{projectRoot}/projects/atomic-angular/dist"],
-      "dependsOn": ["^build", "clean"],
+      "dependsOn": ["^build"],
```
btaillon, line 10: The clean task doesn't exist (and never did AFAIK), so it was just getting skipped.

packages/atomic-angular/project.json
```diff
@@ -34,7 +34,7 @@
           "npm link ../../../../../node_modules/@coveo/release",
           "npx -p=@coveo/release npm-publish"
         ],
-        "parallel": true,
+        "parallel": false,
```
btaillon, line 37: The commands above must be executed in the correct order.

PR #3157, opened 2023-09-07
feat(headless): added most ssr controller `define` functions

btaillon:
https://coveord.atlassian.net/browse/KIT-2714

# What's in this PR
This PR includes the most repetitive additions to Headless's server side rendering utilities: the `define` functions for controllers. In this PR, I also included a usage example in the `samples/headless-ssr-utils` directory (which will be replaced eventually and eventually get unit tests).

To create this PR, I essentially followed these steps:
1. Copy and pasted `src/controllers/index.ts` into a new file (`src/controllers/index.ssr.ts`).
2. Re-exported `src/controllers/index.ssr.ts` from `src/ssr.index.ts`).
3. Replaced all `build*` exports from `src/controllers/index.ssr.ts` with `define*` exports.
   * Also modified file paths to include `.ssr` in them.
4. Fixed all the errors from `src/controllers/index.ssr.ts` by adding the missing exports (e.g., creating `headless-facet.ssr.ts` and exporting all the needed types).

# What's not in this PR
I did not test these define functions, I will test and improve them in separate PRs. Some notable changes missing from this PR which will be included in later PRs:
* Validation / reference docs ([KIT-2663](https://coveord.atlassian.net/browse/KIT-2663))
  * There's no validation that checks whether all controllers have a `define` counterpart.
  * Currently, we have the `@internal` tag on all SSR utility exports, which prevents the doc parser from reading them. Since we'll need some doc parser logic to generate documentation in the future, adding additional logic that looks for `define` functions would be redundant.
* `buildFacetConditionsManager` has no SSR counterpart ([KIT-2723](https://coveord.atlassian.net/browse/KIT-2723))
* `defineUrlManager` is incorrect ([KIT-2724](https://coveord.atlassian.net/browse/KIT-2724)).
  * For now, to simplify this PR, I made it the same as any other `define` function.
* `define` functions which optionally take an ID should require it ([KIT-2725](https://coveord.atlassian.net/browse/KIT-2725))
* `buildGeneratedAnswer` has no SSR counterpart ([KIT-2726](https://coveord.atlassian.net/browse/KIT-2726))
  * This controller is updated multiple times after the search, making it too different to implement here.

[KIT-2663]: https://coveord.atlassian.net/browse/KIT-2663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KIT-2723]: https://coveord.atlassian.net/browse/KIT-2723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


louis-bompart (approved):

packages/headless/src/controllers/search-parameter-manager/headless-search-parameter-manager.ssr.ts
```diff
@@ -10,6 +10,7 @@ export type {
   SearchParameterManagerInitialState,
   SearchParameterManagerState,
   SearchParameterManager,
+  SearchParameters,
```
mrrajamanickam-coveo, line 13: Won't re-exporting using `*` reduce maintenence overhead when new exports are added to the headless controllers?

packages/samples/headless-ssr/src/app/generic/components/facet.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface FacetProps {
+  title: string;
+  ssrState: FacetState;
+  controller?: FacetController;
+}
+
+export const Facet: FunctionComponent<FacetProps> = ({
+  title,
+  ssrState,
```
mrrajamanickam-coveo, line 13: Can we rename these to use the new terminology?

packages/samples/headless-ssr/src/app/react/components/facets.tsx
```diff
@@ -0,0 +1,26 @@
+import {FacetState, Facet as FacetController} from '@coveo/headless';
```
mrrajamanickam-coveo, line 1: Should these import from headless/ssr (although they are the same technically)?

mrrajamanickam-coveo (changes requested):

packages/samples/headless-ssr/src/app/react/components/facets.tsx
btaillon: woops, you're right!

packages/samples/headless-ssr/src/app/generic/components/facet.tsx
btaillon: good catch, I had run a build after updating from `master` and forgot about it when working on something else. the build failed due to some stuff like this

packages/samples/headless-ssr/src/app/react/components/facets.tsx
btaillon: after all, I'm going to copy almost all the exports from `@coveo/headless` in `@coveo/headless/ssr`.

packages/headless/src/controllers/search-parameter-manager/headless-search-parameter-manager.ssr.ts
btaillon: Good point. We already explicitly choose the types to export in the index file (just like we do with the search use case), so there's no need to selectively export stuff here. I'll correct that.

mrrajamanickam-coveo (approved):

PR #3164, opened 2023-09-11
feat(commerce): make commerce team owners of commerce sub-package

Spuffynism:
Since the commerce team works on the headless commerce sub-package, it makes sense for us to own the related files. 

[CAPI-83]

[CAPI-83]: https://coveord.atlassian.net/browse/CAPI-83?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -17,3 +17,4 @@
 /packages/headless/commerce/ @coveo/commerce-routemasters
 /packages/headless/src/**/commerce/ @coveo/commerce-routemasters
 /packages/headless/src/app/commerce-engine/ @coveo/commerce-routemasters
+/packages/headless/src/commerce.index.ts @coveo/commerce-routemasters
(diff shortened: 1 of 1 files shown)


Spuffynism:
I pushed some more commits which make the commerce team have more granular ownership of commerce files. I'm not sure this is the best approach as it feels fragile, as it'll be difficult to keep the CODEOWNERS file up-to-date as changes occur.

I'm not sure of what could be a better solution. Anyone?

louis-bompart:
> I pushed some more commits which make the commerce team have more granular ownership of commerce files. I'm not sure this is the best approach as it feels fragile, as it'll be difficult to keep the CODEOWNERS file up-to-date as changes occur.
> 
> I'm not sure of what could be a better solution. Anyone?

I think this is a symptom of our current file structure.
The issue is that right now we're a bit bound by it (because we did not use explicit entrypoints https://nodejs.org/api/packages.html#package-entry-points).
We do plan to use explicit entrypoints with v3, which will allow us to restructure the files afterward.

So yeah, ain't the best for now, but we have a plan at least.

louis-bompart (approved):

PR #3166, opened 2023-09-11
test(e2e): cypress parallelization

louis-bompart:
## Changes:
- Split E2E of results list
- Leverage `cypress-split` across "just a few" runners. (credits to @mrrajamanickam-coveo for the finds)

## Bikeshedding
Do not resolves the false positive (i.e test being green when they shall not)
Result list specs still need a bunch of love.

## Results:
 back below the 30' mark again.


.github/workflows/prbot.yml
```diff
@@ -46,20 +46,77 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        spec:
+        containers:
           [
```
mrrajamanickam-coveo, line 50: nit: can we use a js expression to generate the range of numbers e.g. `${{ [...Array(PARALLEL_JOBS).keys()].splice(1) }}`

mrrajamanickam-coveo (approved):

.github/workflows/prbot.yml
louis-bompart: Yes could have, although we would have to put it in a runner before and then use the output of this pre runner just to set a static array... I felt this was not really the best, especially because well. We already use heaps and heaps of - runners with that

packages/atomic/cypress/e2e/result-list/folded-result-list.cypress.ts
```diff
@@ -1,219 +1,13 @@
...
-      true
-    );
-  });
-});
+tests(
+  FoldedResultListSelectors,
+  foldedResultListComponent,
+  addResultList,
+  'Folded Result List Component'
+);
```
alexprudhomme, line 13: @louis-bompart Hi, every result list component is hard failing in cypress 13 so I am looking further into them. I think this is a mistake, in the foldedResultList test suite you are using the **atomic-result-list** component. There is a completely different component that is supposed to be tested in this suite called **atomic-folder-result-list**. I might be wrong tho, can you explain why you used **addResultList** ?

PR #3168, opened 2023-09-12
fix(headless): adapted SSR url manager to take parameters at build time

btaillon:
https://coveord.atlassian.net/browse/KIT-2714

# Context

SSR controllers usually receive their props statically at what I call "definition time".

For example, an implementer may decide to create an `engine.ts` file at the root of their project containing:
```js
import {
  defineSearchEngine,
  defineSearchBox,
  defineResultList,
  defineFacet,
} from '@coveo/headless/ssr';

export const engineDefinition = defineSearchEngine({
  configuration: { /* ... */ },
  controllers: {
    searchBox: defineSearchBox(),
    resultList: defineResultList(),
    authorFacet: defineFacet({ facetId: 'author-1', field: 'author' }),
  },
});
```

These controllers are then built using these configurations on both the server side and on the client side.

For example, on the server side:
```js
import { engineDefinition } from '@/engine';

export default async function Page(url) {
  const initialState = await engineDefinition.fetchInitialState();
  return (
    <SearchEngineAndControllersProvider initialState={initialState}>
      {/* ... */}
    </SearchEngineAndControllersProvider>
  );
}
```

And on the client side:
```js
'use client';

import { engineDefinition } from '@/engine';
import { use, useMemo } from 'react';

export default function SearchEngineAndControllersProvider(props) {
  const { engine, controllers } = use(
    useMemo(
      () => engineDefinition.hydrateInitialState(props.initialState.searchFulfilledAction),
      [props.initialState],
    ),
  );

  // ...
}
```

# Problem

The search parameter manager and the URL manager both need to take the current URL as a prop, which isn't something that can be determined statically.

As an escape hatch, it's possible to build them using the regular `buildSearchParameterManager` and `buildUrlManager` functions. However, doing so requires the engine, which isn't returned by `fetchInitialState` (for the sake of predictability / ease of use / fool-proof-ness). If these controllers only had an effect on the client side, this wouldn't be a concern, but unfortunately, they can affect the first query.

* For advanced use cases, implementers can use `engineDefinition.build` instead of `engineDefinition.fetchInitialState`. However, it's more difficult to use, and search parameters feel like a common enough use case that I don't think we should recommend an escape hatch for it.

# What's in this PR

The solution for this was already implemented for the `defineSearchParameters` controller: take props at build time instead of definition time. I'm copying this solution over to `defineUrlManager`.

For example, here's how that would look like in the implementer's engine definition:
```js
import {
  defineSearchEngine,
  defineSearchBox,
  defineResultList,
  defineFacet,
  defineUrlManager
} from '@coveo/headless/ssr';

export const engineDefinition = defineSearchEngine({
  configuration: { /* ... */ },
  controllers: {
    searchBox: defineSearchBox(),
    resultList: defineResultList(),
    authorFacet: defineFacet({ facetId: 'author-1', field: 'author' }),
    urlManager: defineUrlManager(), // No props!
  },
});
```

Here's how that looks like on the server side:
```js
import { engineDefinition } from '@/engine';
import { getFragmentFromUrl } from '@/utils';

export default async function Page(url) {
  const initialState = await engineDefinition.fetchInitialState({
    controllers: {
      urlManager: {
        initialState: {
          fragment: getFragmentFromUrl(url),
        },
      },
    },
  });

  // ...
}
```

And on the client side:
```js
'use client';

import { engineDefinition } from '@/engine';
import { use, useMemo } from 'react';

export default function SearchEngineAndControllersProvider(props) {
  const { engine, controllers } = use(
    useMemo(
      () => engineDefinition.hydrateInitialState(
        props.initialState.searchFulfilledAction,
        {
          controllers: {
            urlManager: {
              initialState: controllers.urlManager.initialState,
            },
          },
        },
      ),
      [props.initialState],
    ),
  );

  // ...
}
```

In this PR, I also updated the tests to use the URL manager, since it composes the search parameter manager, therefore inherently also kinda tests it.

# What's not in this PR

As you can see in the samples I modified, URL management with Next.js and our SSR utilities is still not super simple/intuitive and may require additional utilities. I created a separate JIRA ticket to investigate a solution to those issues ([KIT-2735](https://coveord.atlassian.net/browse/KIT-2735)).

I didn't add unit tests, I only updated end-to-end tests. If I added unit tests now, they would likely be covered anyway by the tests for [KIT-2735](https://coveord.atlassian.net/browse/KIT-2735).

[KIT-2735]: https://coveord.atlassian.net/browse/KIT-2735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
```diff
@@ -31,44 +28,36 @@ export function useSyncSearchParameters({
...
+    if (state.fragment === searchParams.toString()) {
       return;
     }
     const isInitialState = controller === undefined;
     if (isInitialState) {
-      historyRouter.replace(correctedUrl);
+      replace('?' + state.fragment, {shallow: true} as NavigateOptions);
     } else {
-      historyRouter.push(correctedUrl);
+      push('?' + state.fragment, {shallow: true} as NavigateOptions);
```
btaillon, line 59: Discovered this hidden workaround. The `shallow` option allows Next.js to update the parameters without reloading the page, but it isn't mentioned in their docs and isn't accepted by the function's type definition.

louis-bompart: Yeah.... That smells like an internal then? Ain't sure it's a good idea chief.

louis-bompart (changes requested):
Not sure about shallow

packages/samples/headless-ssr/src/app/generic/hooks/search-parameters.ts
btaillon: Yeah, the alternative is to make our own hooks that also use NextJS's hooks underneath and try to hack things into working. We wouldn't need to use the `shallow` option though.

btaillon: For now, I'll close this PR, open a PR to remove `defineUrlManager`, and revisit after we've done a deeper investigation into URL management.

PR #3169, opened 2023-09-12
feat(atomic): add `resultsMustMatch` option to facet components

y-lakhdar:
Allow user to toggle facet value selection logic from `OR` to `AND`.

When not specified, the default value is set to `atLeastOneValue` (`OR`). The user can specified `allValues` which will act as `AND` when multiple facet values are selected.


packages/atomic/src/components/search/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
```diff
@@ -126,6 +127,14 @@ export class AtomicNumericFacet
...
   @Prop({reflect: true}) public sortCriteria: RangeFacetSortCriterion =
     'ascending';
+  /**
+   * Specifies how a result must match the selected facet values.
+   * Allowed values:
+   * - `atLeastOneValue`: A result will match if at least one of the corresponding facet values is selected.
+   * - `allValues`: A result will match if all corresponding facet values are selected.
+   */
+  @Prop({reflect: true}) public resultsMustMatch: FacetResultsMustMatch =
+    'atLeastOneValue';
```
btaillon, line 137: I'm having trouble visualizing a situation where you would want "allValues" for a numeric filter. I think we should have an example here.

packages/atomic/cypress/e2e/facets/numeric-facet/numeric-facet.cypress.ts
```diff
@@ -178,6 +178,58 @@ describe('Numeric Facet V1 Test Suites', () => {
...
+            .init();
+
+          selectIdleCheckboxValueAt(FacetSelectors, 0);
+        }
+
+        beforeEach(() => {
+          setupSelectCheckboxValue();
+        });
+
+        it('should include resultsMustMatch in the request', () => {
```
mrrajamanickam-coveo, line 200: 'should include resultsMustMatch' -> 'should set resultsMustMatch to allValues' ?

packages/atomic/cypress/e2e/facets/numeric-facet/numeric-facet.cypress.ts
```diff
@@ -178,6 +178,58 @@ describe('Numeric Facet V1 Test Suites', () => {
...
```
mrrajamanickam-coveo, line 223: similar to above

packages/atomic/src/components/search/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
y-lakhdar: I agree. Not sure it is relevant to add it to the numeric-facet. 
However, I imagine documents could have a field with multiple numerical values used in a numeric-facet. 
But even with that, I agree that it is not enough to include the option. Let me know what you think

packages/atomic/cypress/e2e/facets/color-facet/color-facet.cypress.ts
```diff
@@ -378,6 +378,55 @@ describe('Color Facet Test Suites', () => {
     });
   });
 
+  describe('with #resultsMustMatch set to default value', () => {
```
mrrajamanickam-coveo, line 381: Notice that most tests only check the request. Would there be value in adding any checks on the UI state (and results etc) or are they taken care of by other existing tests?

y-lakhdar: It's kinda hard to test the UI based on this option since there is no logic in atomic regarding the filtering logic. The `resultsMustMatch` option is tested in headless.

btaillon (approved):

mrrajamanickam-coveo (approved):

louis-bompart (approved):

nathanlb (approved):

PR #3172, opened 2023-09-13
feat(headless): wrap preprocess function that throws to prevent stopping the execution requests

olamothe:
Partial PR to see if we get a consensus on this, as discussed during the scrum:

Preprocess function are exposed as a low level hook implementers can use to either modify outgoing requests, or incoming response from our different APIs.

They are pretty popular, and **should** be used as an escape hatch "when you know what you are doing".

```typescript
//typical example

buildSearchEngine({
   // [...],
   preprocessRequest: (request, origin)=> {
      // request.body is pretty much "anything" as far as the function signature is concerned.
 
      // However, since we typically do `application/json` type of request against most of our API
      // we see this often, without any check/protection/defensive code.

     // We could tomorrow need to talk with a new API that use something else than application/json.
     // All API call will go through that low level function.
     const parsed = JSON.parse(request.body);

     parsed.someParameterNotStillSupportedByHeadless = 'foo';
     request.body = JSON.stringify(parsed);

     return request;
      
   }
})
```

I've noticed however that the preprocessRequests type of function are used more often than they **should** be (ie: there's often a better way to achieve the same result differently by using an actual controller/dispatching an action on the engine).

What this PR does it wrap the function as provided by the implementer in a try catch, log the error to the console, and continue execution of the request as if no hook had been executed in the first place (ie: no modification on the original request).

Pros of doing this:
* Any error in that hook will pretty much "brick" the whole search page. This is banking on the fact that there's more chance that there's less adverse effect on end users (ie: people using the search page) if we just recover and continue on our side, and that there's a good chance they won't even see it threw an error.

Cons:
* The above is an assumption: It's possible we create more problem than we solve with this, and by doing too much magic we negatively impacts deployments because throwing an error was a behaviour the implementer relied upon (to actually cancel or abort a request, maybe ?)

https://coveord.atlassian.net/browse/KIT-2737


olamothe:
An alternative might be to tighten up the `preprocess` function signature in the future, and always enforce stringified JSON for the request body

packages/headless/src/api/platform-client.ts
```diff
@@ -103,6 +97,34 @@ export class PlatformClient {
...
+      requestInfo = {
+        ...defaultRequestOptions,
+        ...processedRequest,
+      };
+    } catch (e) {
+      logger.error(
+        e as Error,
+        'Platform request preprocessing failed. Returning default request options.'
+      );
+      requestInfo = {...defaultRequestOptions};
```
louis-bompart, line 124: ⚠️ preprocessRequest can taint the defaultRequestOptions then throw I think.
We need to ensures we default on the untainted default I think

louis-bompart (changes requested):

btaillon:
I overall don't think this is super impactful, so it's not high up in the list of things I care about. However, weighing the pros vs the cons, I still can't really think of situations where catching the error is preferable, including the error with the docs site. I strongly believe the error with the docs site was because our change was breaking and it happened to pass through our e2e tests.

If the hooks were something that was being added dynamically by external libraries and there wasn't much visibility in which hooks are being used, I would agree with catching errors. However, here I feel like we're just enabling implementations to be badly done.

btaillon:
I think additional e2e tests would be more valuable to prevent this kind of error in the future.

fbeaudoincoveo (commented):
I personally agree with the proposed change.

As @btaillon puts it, this can be seen as enabling implementations to be badly done, but from my perspective, this is more like adding some railings / warning signs around a pitfall that's otherwise easy to overlook and stumble in (with potentially really bad consequences).

One thing that proves how easy it is to fall in that pittrap is that our current documentation on the subject includes examples that parse the request body to JSON without any safeguards: https://docs.coveo.com/en/headless/latest/usage/headless-modify-requests-responses/. We certainly can't blame implementers for following the docs 😅 That being said we should adjust it as well (is there a DOC issue for that?)

Speaking of docs, the reference docs on the preprocessRequest property of all engine configs is quite vague and unhelpful. We should improve that as well, and probably indicate right here and there that Headless will automatically wrap the provided preprocess function in a try-catch, and log an error to the console instead of throwing. We might also point to the more complete topic on modifying requests and responses directly in the reference docs.

packages/headless/src/api/platform-client.ts
olamothe: Fixed and added UT

olamothe:
> One thing that proves how easy it is to fall in that pittrap is that our current documentation on the subject includes examples that parse the request body to JSON without any safeguards: https://docs.coveo.com/en/headless/latest/usage/headless-modify-requests-responses/. We certainly can't blame implementers for following the docs 😅 That being said we should adjust it as well (is there a DOC issue for that?)
> 
> Speaking of docs, the reference docs on the preprocessRequest property of all engine configs is quite vague and unhelpful. We should improve that as well, and probably indicate right here and there that Headless will automatically wrap the provided preprocess function in a try-catch, and log any caught error to the console instead of throwing. We might also point to the more complete topic on modifying requests and responses directly in the reference docs.

Created this ticket here: https://coveord.atlassian.net/browse/KIT-2757

btaillon:
> I personally agree with the proposed change.
> 
> As @btaillon puts it, this can be seen as enabling implementations to be badly done, but from my perspective, this is more like adding some railings / warning signs around a pitfall that's otherwise easy to overlook and stumble in (with potentially really bad consequences).
> 
> One thing that proves how easy it is to fall in that pittrap is that our current documentation on the subject includes examples that parse the request body to JSON without any safeguards: [docs.coveo.com/en/headless/latest/usage/headless-modify-requests-responses](https://docs.coveo.com/en/headless/latest/usage/headless-modify-requests-responses/). We certainly can't blame implementers for following the docs 😅 That being said we should adjust it as well (is there a DOC issue for that?)

I think this is our biggest difference in perspective. From my perspective, it's not a mistake at all for implementers to parse this string as JSON. I believe that a JSON string is always a JSON string unless stated otherwise.

My reasoning is that, say we expect implementers to always write their parameters defensively, and say their hook is crucial for the functionality of their application, implementers would have to start making guesses as to what else than JSON it could be. YAML? XML? A blob? Javascript? If their hook is important, we're putting the responsibility on them to handle **every** possible case.

Also, the way I see it, it's not significantly easier to make a broken hook than any other function. I believe we're the ones who broke the hooks by breaking an implicit "contract", which is that the string is JSON.

packages/headless/src/api/platform-client.test.ts
```diff
@@ -157,6 +158,28 @@
...
+        shouldNotExistsOnTheOutgoingRequest: 'foo',
+      };
+      throw 'boom';
+    };
+    await platformCall({preprocessRequest});
+    expect(mockFetch).toHaveBeenCalledWith(
+      platformUrl(),
+      expect.objectContaining({
+        headers: {
+          Authorization: 'Bearer accessToken1',
```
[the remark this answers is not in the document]
louis-bompart: shush.

louis-bompart (approved):

dmbrooke (approved):
Although I was initially a bit concerned about "smothering" errors, the benefit of not crashing the entire search page certainly outweighs the negatives in my opinion. So long as there's still visibility for the error (i.e., we log an error that devs can easily catch while implementing), this should fix far more problems than it could potentially introduce.

I'm assuming there's already a docs ticket open for the API writers to update their code samples with the best practices (try/catch)? Even if we start catching these errors on our side, it's probably still a good idea to ensure that implementers on older versions don't run into the same problem the docs site did 🙂

PR #3173, opened 2023-09-13
unify cypress version

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2747

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "cross-fetch": "3.1.5",
     "cspell": "6.31.1",
     "cypress-repeat": "2.3.3",
+    "cypress": "10.8.0",
     "cz-conventional-changelog": "3.3.0",
     "detect-indent": "7.0.1",
     "esbuild": "0.14.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,6 +42,7 @@
         "concurrently": "7.6.0",
         "cross-fetch": "3.1.5",
         "cspell": "6.31.1",
+        "cypress": "10.8.0",
         "cypress-repeat": "2.3.3",
         "cz-conventional-changelog": "3.3.0",
         "detect-indent": "7.0.1",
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "cross-fetch": "3.1.5",
     "cspell": "6.31.1",
     "cypress-repeat": "2.3.3",
+    "cypress": "10.8.0",
```
louis-bompart, line 49: We should prolly not have cypress on the root?

alexprudhomme: Why not ? It is a common dependency for multiple projects. It's also in devDependencies so it should not affect any bundle. Thats how it works right ?

alexprudhomme: Also @louis-bompart did you see the license issue ? It is caused by qrcode-terminal which is a dependency of sfdx-cli or more. I think it was caused by your TS5 PR.

louis-bompart: > Why not ? It is a common dependency for multiple packages. I thought that was the point of having a mono repo.

So the main goal of a mono repo is to allow tighter collaboration/integration between different parts of a project (such as Headless and Atomic), but it is true that dependency deduplication is part of the boon of mono repo.
However, this is handled by the package manager. (indeed, it will dedupe the install :))

But let's say for example that Cypress version X is compatible with our Angular output, but not Vue. We'd have to wait to update for a Cypress version being compatible with everything. This complexifies updates, as you cannot divide and conquer the update anymore, so: we should strives to uses the same version everywhere yes, but we should not imposes it (IMO).

louis-bompart: > Also @louis-bompart did you see the license issue ? It is caused by qrcode-terminal which is a dependency of sfdx-cli or more. I think it was caused by your TS5 PR.
> 
> How do I solve these things ?

You don't 🤡
We apply our own discretion to this kind of issue, and here we can see that's an Apache-2.0 package, so no issue really
https://github.com/gtanner/qrcode-terminal/blob/master/LICENSE

PR #3162, opened 2023-09-09
chore: update to TypeScript 5, Stencil 4 (and more :finnadie: )

louis-bompart:
- [x] Rebase/clean the commit history

Mostly upgrade Stencil to v4 and TypeScript to v5.

Bonuses:
 - Fix Angular E2E unstable
 - Jest 29 in most of the project (exclusion on Atomic)

The riskiest/biggest change is around the patch in the Stencil compiler (good old AST :finnadie:) See:

 - https://github.com/ionic-team/stencil/pull/4771
 - https://github.com/ionic-team/stencil/pull/4772
 
Will scope out the type fix of the code-editor component of Storybook into an improvement: Unknown amount of effort, very low reward.

___

https://coveord.atlassian.net/browse/KIT-2312

PR #3176, opened 2023-09-14
ci: false positives for cypress tests

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2751

This PR solves the false positives for cypress tests by adding the cypress-repeat dependency inside every package. It also the missed bugs added (because of false positives) during https://github.com/coveo/ui-kit/pull/3162. 

These bugs are :
- Tests not working for angular, atomic-react, atomic-hosted-page and stencil sample
- Tests not working for foldedResultList

cypress-repeat uses the latest version of cypress. cypress-repeat is broken in cypress 13.1.0. Because of our mono repo and because we dont have a general cypress dependency at root, it uses the latest cypress which is 13.1.0 in master right now. It causes [false positives](https://github.com/coveo/ui-kit/actions/runs/6185845019/job/16794254134#step:4:978). I dont know how but cypress 13.1.0 was introduced in the package-lock.json during the typescript 5 PR. ⁉️ 

The solution is to put the cypress-repeat dependency explicitly in every project.

Proof of it working : https://github.com/coveo/ui-kit/actions/runs/6188918271/job/16801963797#step:4:486


louis-bompart:
Did we check if cypress-repeat was still completing with ✅ after all attempts ends up :X: ?

SimonMilord (approved):
Looks good, but it seems like some tests are still failing?

PR #3177, opened 2023-09-14
chore(renovate): isolate storybook

louis-bompart:
- Add a name to existing groups
- Create a new group for Storybook (divide & conquer)


renovate.json
```diff
@@ -21,8 +22,17 @@
         "react-dom",
         "@types/react"
       ],
+      "groupName": "React",
       "groupSlug": "react",
       "rangeStrategy": "replace"
+    },
+    {
+      "matchPackagePatterns": [
```
mrrajamanickam-coveo, line 30: Notice a number of storybook deps in the ignored list. Is that expected?

louis-bompart: Uh missed that good catch

btaillon (approved):

PR #3179, opened 2023-09-15
test: Add jest/jsdom unit tests for the ssr hooks

mrrajamanickam-coveo:
Adds unit tests for 
- Static, Hydrated state providers
- useSyncMemoizedStore
  - Fixes an issue with useSyncMemoizedStore

https://coveord.atlassian.net/browse/KIT-2694


packages/headless-react/src/client-utils.ts
```diff
@@ -1,23 +1,31 @@
...
  */
 export function useSyncMemoizedStore<T>(
-  subscribe: (listener: () => void) => () => void,
-  getSnapshot: () => T
+  subscribe: Subscriber,
+  getSnapshot: SnapshotGetter<T>
 ) {
-  const stateRef = useRef(getSnapshot());
+  let latestSnapshot = getSnapshot();
+  const getLatestSnapshot = () => latestSnapshot;
```
btaillon, line 21: What was the bug that prompted these changes?

As far as I thought, we shouldn't use variables in a hook since they would be reinitialized on every render (in this case, calling getSnapshot every single time that anything in the parent component causes any re-render). I admittedly haven't tested that though.

When I originally tied to make the `useSyncMemoizedStore` hook, my implementation was significantly more complex (and probably not as efficient), maybe there's something missing from it that would solve your issue. Here it is:
<details>
<summary>use-sync-memoized-store.ts</summary>

```ts
import {DependencyList} from 'react';
import {useEffect, useReducer, useRef} from './wrapped';

function useHasChanged(deps: DependencyList) {
  const ref = useRef<DependencyList | null>(null);
  if (ref.current === null) {
    ref.current = deps;
    return false;
  }
  if (
    ref.current.length === deps.length &&
    !deps.some((dep, i) => !Object.is(ref.current![i], dep))
  ) {
    return false;
  }
  ref.current = deps;
  return true;
}

function useRefWithCallbackInitialState<T>(getInitialState: () => T) {
  const initialValueIsSet = useRef(false);
  const ref = useRef(null as T);
  !initialValueIsSet.current && (ref.current = getInitialState());
  initialValueIsSet.current = true;
  return ref;
}

/**
 * Same as [useSyncExternalStore](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore), but doesn't infinitely re-render with a memoized store.
 */
export function useSyncMemoizedStore<T>(
  subscribe: (listener: () => void) => () => void,
  getSnapshot: () => T
) {
  const state = useRefWithCallbackInitialState<T>(getSnapshot);
  const [, rerender] = useReducer((renderCount: number) => renderCount + 1, 0);
  useEffect(
    () => {
      let canRerender = false;
      const unsubscribe = subscribe(() => {
        if (!canRerender) {
          return;
        }
        state.current = getSnapshot();
        rerender();
      });
      canRerender = true;
      return unsubscribe;
    },
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [subscribe]
  );
  if (useHasChanged([getSnapshot])) {
    state.current = getSnapshot();
  }
  return state.current as T;
}
```
</details>

btaillon (commented):
fantastic tests!

packages/headless-react/src/client-utils.ts
mrrajamanickam-coveo: Without the fixes in 48d0af43f06573733eacd405e0f32d790b4ea338 the tests fail as follows
```
 FAIL  src/client-utils.test.tsx
  ● useSyncMemoizedStore › should update the state when the store changes

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Object {
    -   "count": 1,
    +   "count": 0,
      }

      32 |     expect(result.current).toEqual(snapshot1);
      33 |     rerender();
    > 34 |     expect(result.current).toEqual(snapshot2);
         |                            ^
      35 |   });
      36 |
      37 |   test('should unsubscribe and re-subscribe to new function when subscribe function is changed', () => {

      at Object.<anonymous> (src/client-utils.test.tsx:34:28)

  ● useSyncMemoizedStore › should replace current snapshot when getSnapshot function is changed

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Object {
    -   "count": 1,
    +   "count": 0,
      }

      72 |     expect(result.current).toEqual(snapshot1);
      73 |     rerender({getSnapshot: getSnapshot2});
    > 74 |     expect(result.current).toEqual(snapshot2);
         |                            ^
      75 |   });
      76 |
      77 |   it('should re-render only when the subscribe listener is called', () => {

      at Object.<anonymous> (src/client-utils.test.tsx:74:28)

 PASS  src/search-engine.test.tsx (5.087 s)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       2 failed, 14 passed, 16 total
```

With the code you have provided 1 test fails
```
 FAIL  src/client-utils.test.tsx
  ● useSyncMemoizedStore › should update the state when the store changes

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Object {
    -   "count": 1,
    +   "count": 0,
      }

      32 |     expect(result.current).toEqual(snapshot1);
      33 |     rerender();
    > 34 |     expect(result.current).toEqual(snapshot2);
         |                            ^
      35 |   });
      36 |
      37 |   test('should unsubscribe and re-subscribe to new function when subscribe function is changed', () => {

      at Object.<anonymous> (src/client-utils.test.tsx:34:28)

 PASS  src/search-engine.test.tsx (5.009 s)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 15 passed, 16 total
```

As the current code passes both the unit and e2e tests and is simpler, tempted to leave it as is for now.

btaillon: I'm hesitant, since in our case, `getSnapshot` calls a getter (someController.state). For some controllers, this getter is fairly simple, but either way it'll always generate an object.

I tested, and was able to confirm that calling the method outside of a hook like `useMemo` causes it to be called on every re-render of the parent component (see https://7wvpvj-5173.csb.app/, which logs "Called getSnapshot1" whenever the state is updated every 5 milliseconds, but doesn't log "Called getSnapshot2" repeatedly since it's wrapped in a useMemo hook).

Concerning the first test failure ("should update the state when the store changes"), I suspect it may be failing due to how the test is setup, specifically this line:
```ts
const getSnapshot = jest
  .fn()
  .mockReturnValueOnce(snapshot1)
  .mockReturnValueOnce(snapshot2);
```

As far as I know, the reference to `getSnapshot` never changes, only its returned value, which IMO we don't need to verify (and shouldn't). Instead, I think this test should verify whether replacing the `getSnapshot` function with a different one updates the state.

mrrajamanickam-coveo: Thanks for testing out with an example. But in this case the `getStaticState` from `useCallback` is passed into `useSyncMemoizedStore` - so maybe the memoized return value is used instead?
I have tried out `useRef`, `useMemo` and even `useState` to wrap `getSnapshot()` - but all of them fail tests as indicated above. Is there a specific reason you believe the code is problematic when the tests are passing? Is there any issue with the tests?

mrrajamanickam-coveo: you were right! added test to test getSnapshot is not called on rerender, updated to use useMemo, fixed another test that was not testing the right thing.

packages/headless-react/src/ssr/client-utils.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const effect = jest.fn();
+
+    const {result} = renderHook(() => {
+      const state = useSyncMemoizedStore(subscribe, getSnapshot);
+      useEffect(effect, [state]);
+      return state;
+    });
+
+    expect(result.current).toEqual(snapshot);
+    subscribe();
```
btaillon, line 104: If you remove this `subscribe`, does the test still pass?

I just realized that the test may be passing due to `useEffect` always getting called at-least once. This call doesn't seem to be doing anything?

mrrajamanickam-coveo: That is correct - not sure how I missed that 🤦. Thanks for catching it - Fixed the test.

packages/headless-react/src/ssr/client-utils.ts
```diff
@@ -1,23 +1,30 @@
...
-  const stateRef = useRef(getSnapshot());
+  let latestSnapshot = useMemo(() => getSnapshot(), [getSnapshot, subscribe]);
   const [, forceRender] = useReducer((s) => s + 1, 0);
 
   useEffect(() => {
     let isMounted = true;
     const unsubscribe = subscribe(() => {
       if (isMounted) {
-        stateRef.current = getSnapshot();
+        latestSnapshot = getSnapshot();
```
btaillon, line 27: Is this update needed? I would expect this line of code to do nothing, since `latestSnapshot` will be reinitialized on the next render.

I think a test may be missing, something like "when the listener is called, re-renders and updates the snapshot". I suspect it would fail on this test. It's probably the reason why this fails in the e2e tests:
image

# Solutions

There's probably better solutions than those two, but they are the only ones I can really think of...

## Solution A: `useState`
One potential fix I thought of for this issue would be to replace `latestSnapshot`'s `useMemo` with a `useState`. 

However, unlike memos, states don't have a dependencies array. This means that, when something causes a re-render where `subscribe` and/or `getSnapshot` are updated, the state wouldn't get updated.

There may be ways to work around this issue. For example, you could update the state inside `useEffect`. However, this would cause a second re-render, which is pretty inefficient.

This is why I would fall back to solution B or C.

## Solution B: `useRef`

You could replace `latestSnapshot`'s `useMemo` with a [`useRef`](https://react.dev/reference/react/useRef).

However, similarly to `useState`, `useRef` doesn't have a dependencies array, it would need to be manually updated. Unlike `useState`, updating it wouldn't cause a re-renders and would be effective immediately. Beware though, updating it inside `useEffect` wouldn't affect the current render, since `useEffect` is called after rendering.

You could use the custom [`useHasChanged`](https://github.com/coveo/ui-kit/blob/KIT-1456-btaillon-app-router/packages/samples/headless-next/src/hooks/use-controller.ts#L122) hook I made last time to update it in the current rendering phase, but that's one more hook on the pile of hooks 😅 .

mrrajamanickam-coveo: Thanks! Switched to `useRef` with useHasChanged.
Missed the e2e test failure as I was re-triggering to get rid of some other flaky tests 🤦

btaillon (approved):

packages/headless-react/src/ssr/client-utils.ts
```diff
@@ -1,23 +1,49 @@
...
  * https://github.com/facebook/react/issues/24529
  */
 export function useSyncMemoizedStore<T>(
-  subscribe: (listener: () => void) => () => void,
-  getSnapshot: () => T
+  subscribe: Subscriber,
+  getSnapshot: SnapshotGetter<T>
 ) {
-  const stateRef = useRef(getSnapshot());
+  const snapshot = useRef(getSnapshot());
```
btaillon, line 39: to avoid `getSnapshot` getting called on every re-render, you should set the initial value here to `null`. Your condition at line 59 (`snapshot.current === null`) should be sufficient to call `getSnapshot` the first time.

mrrajamanickam-coveo: Great, thanks! Updated. Couldn't use `null` as it was raising type errors when getStaticState is passed in from buildControllerHook. Used `{}` to match getStaticState.

packages/headless-react/src/ssr/client-utils.ts
```diff
@@ -1,23 +1,49 @@
...
  * Alternate for `useSyncExternalStore` which runs into infinite loops when hooks are used in `getSnapshot`
  * https://github.com/facebook/react/issues/24529
  */
-export function useSyncMemoizedStore<T>(
-  subscribe: (listener: () => void) => () => void,
-  getSnapshot: () => T
+export function useSyncMemoizedStore<T extends {}>(
+  subscribe: Subscriber,
+  getSnapshot: SnapshotGetter<T>
 ) {
```
btaillon, line 38: Does Typescript infer that the return type is T?

packages/headless-react/src/ssr/client-utils.ts
```diff
@@ -1,23 +1,49 @@
...
-  const stateRef = useRef(getSnapshot());
+  const snapshot = useRef({});
```
btaillon, line 39: Concerning your comment about this being null, it may be possible to make this null by passing it as a generic like so:
```ts
const snapshot = useRef<T | null>(null);
```

If the problem is that `T` may be nullable and we don't want to identify `null` as being "call `getSnapshot` immediately", then the ref could be split into two like so:
```ts
const snapshot = useRef<T | null>(null);
const snapshotIsSet = useRef<boolean>(false);
// ...
if (!snapshotIsSet.current) {
  snapshot.current = getSnapshot();
  snapshotIsSet.current = true;
}
// ...
return snapshot.current as T;
```

btaillon (approved):

packages/headless/src/app/commerce-engine/commerce-engine.ts
```diff
@@ -22,7 +22,11 @@ import {cartReducer} from '../../features/commerce/context/cart/cart-slice';
 
 export type {CommerceEngineConfiguration};
 
-const commerceEngineReducers = {productListing: productListingV2Reducer, commerceContext: contextReducer, cart: cartReducer};
+const commerceEngineReducers = {
```
Spuffynism, line 25: Looks like I had formatting issues when merging the context & cart PR. Thanks for fixing it here! 😁  

I assumed the pre-commit hook would format this, did I miss something? A ` npm run lint:fix packages/headless` I assume?

Spuffynism (approved):

packages/headless/src/app/commerce-engine/commerce-engine.ts
mrrajamanickam-coveo: `"lint:fix": "npx eslint --fix .",` has embedded path in it - so you would have to run `npx eslint --fix` directly with the path to lint a specific package. Wish it could be done in a separate PR and merged so it doesn't show up in all open PRs that merge from master - it shows up on 2 of my PRs already 😅

mrrajamanickam-coveo:
Unrelated test failures in quantic and folded-result-list-children. Proceeding to merge ..

PR #3180, opened 2023-09-15
chore: upgrade prettier

btaillon:
https://coveord.atlassian.net/browse/KIT-2755


louis-bompart:
Took the liberty of doing it in #3185 because it was causing issue due to the update of `gts`

PR #3182, opened 2023-09-15
fix(headless): updating GenAI state only when stream id corresponds to the generativeQuestionAnsweringId of the latest search query

mmitiche:
[SFINT-5168](https://coveord.atlassian.net/browse/SFINT-5168)

### Explanantion:
The following bug only has an impact in Salesforce (with Quantic components), in a context where the Locker service is enabled which means that the [Abort Controller ](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)can't  be used.

In that specific context, we can't abort the stream requests, so we previously  added a logic in headless that prevents modifying the headless state if a stream request is received and contains a `streamId` value different from `generativeQuestionAnsweringId` value   of  the latest search query performed in the search interface. 
a bug has been found with this logic, since instead of comparing  the `streamId` value with  the `generativeQuestionAnsweringId` value   of  the latest search query, we were comparing  the `streamId` value with the `generativeQuestionAnsweringId` value   of  a previous search query other than the latest one.

This PR comes to fix this.

### Before
PS: notice that the search button was clicked twice and that generative answer and the citations are written twice.

https://github.com/coveo/ui-kit/assets/86681870/675d646d-d7e3-4fb4-bb96-0ff907c94cb0

### After 
PS: the search button was clicked twice as well but now the generative answer is displayed as expected.

https://github.com/coveo/ui-kit/assets/86681870/6c1f2863-b8fe-445b-80fc-505732598705

[SFINT-5168]: https://coveord.atlassian.net/browse/SFINT-5168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


btaillon (approved):

packages/headless/src/features/generated-answer/generated-answer-actions.ts
```diff
@@ -147,7 +147,7 @@ export const streamAnswer = createAsyncThunk<
       write: (data: GeneratedAnswerStreamEventData) => {
         if (
           request.streamId ===
-          state.search.extendedResults.generativeQuestionAnsweringId
+          config.getState().search.extendedResults.generativeQuestionAnsweringId
```
olamothe, line 150: I guess a simple naming changes could make this more obvious to a future reader, to not fall into the same trap (what the different closure are in this `streamAnswer` asynchronous function)

Suggestion: an inline utility function

```typescript
if(currentStreamMatchesOriginalStream(originalState, currentState) {
     // ...
}
```

olamothe (approved):

packages/headless/src/features/generated-answer/generated-answer-actions.ts
erocheleau: That's a good point, especially since this is repeated 4 times, I would make a function with a clear name indeed.

erocheleau (approved):
Other than the function suggestion, everything looks good

PR #3186, opened 2023-09-16
chore(bueno): ensure bueno types are safe-guarded

louis-bompart:
Bueno build is more thorough while checking the libs and types.
Bueno does not depend on any `@types` packages, thus it does not need `types`.

### DIFF
--- a/packages/bueno/jest.config.js
+++ b/packages/bueno/jest.config.js
@@ -1,3 +1,10 @@
 module.exports = {
-  preset: 'ts-jest',
+  transform: {
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        tsconfig: 'tsconfig.test.json',
+      },
+    ],
+  },
 };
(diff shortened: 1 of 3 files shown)


olamothe:
Was there a specific event that prompted this change ? 
I see it's part of a `prettier-3` upgrade branch. 

Can you expand a bit more on why prettier 3 is related ?

louis-bompart:
> Was there a specific event that prompted this change ? I see it's part of a `prettier-3` upgrade branch.

Packages update allowed the package manager to dedupe packages (and thus hoist'em' up to the root node_modules directory).
Also, new packages are hoisted too, so if we want to add a `@types` package (let's say `@types/testing-library__jest-dom`, that was req. for headless-react), it means bueno typechecking will be impacted.

> Can you expand a bit more on why prettier 3 is related ?

Not related to prettier-3, but I started on top of this branch to ensures everything was going smoothly.

btaillon (approved):

PR #3188, opened 2023-09-18
chore(angular): fix release:phase2

louis-bompart:
image

The link seems to fail, replaced by a dumber thing, should work tm (untested)


btaillon (approved):
YOLO

PR #3190, opened 2023-09-18
feat(commerce): create context & cart controllers

Spuffynism:
Create the cart and context controllers and actions. These controllers, along with their actions, allow populating state necessary for querying the PLP v2 endpoint.

The PR has 3 pieces:
- 🛒 cart actions, controllers, slices + their tests
- 🌎 context actions, controllers, slices + their tests
- 🕸️ Moving the product listing URL handling to the context

[CAPI-96]

[CAPI-96]: https://coveord.atlassian.net/browse/CAPI-96?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export interface ContextOptions {
+  trackingId?: string;
+  language?: string;
+  currency?: string;
+  // TODO: Does it make sense for the clientId to be part of the context state?
+  clientId?: string;
+  user?: User;
+  view?: Partial<View>;
```
fbeaudoincoveo, line 39: I think `view` should be set on the ProductListing controller, not on the "global" commerce context

fbeaudoincoveo: Same thing for `labels` (although I believe we decided not to expose this in Headless initially)

packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export interface ContextProps {
+  /**
+   * The initial options that should be applied to this `Context` controller.
+   */
+  options?: ContextOptions;
+}
+
+// TODO: Does this controller do too much? Should it expose sub-controllers which allow mutating
```
fbeaudoincoveo, line 51: I would probably create a distinct controller for the cart, because it's less "static" than the rest of the "global" commerce context. Otherwise, as I said in another comment, I believe `view` and `labels` (if we choose to expose it) should both be on the ProductListing controller.

packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,212 @@
+import { Schema } from "@coveo/bueno";
+import {requiredNonEmptyString} from '../../../utils/validate-payload';
+import {CommerceEngine} from '../../../app/commerce-engine/commerce-engine';
+import {buildController, Controller} from '../../controller/headless-controller';
+
+const optionsSchema = new Schema({
+  url: requiredNonEmptyString,
```
fbeaudoincoveo, line 7: Granted that we remove the `cart`, `view`, and `labels` attribute from this controller's state, I think we should expose all of the remaining attributes in its options schema.

Here my reasoning:

- Once set, `trackingId` is likely to remain static during the entire browsing session. In fact, can there realistically even be multiple tracking IDs within the same commerce site? I suppose if it's theoretically possible, then it's OK to expose an option to mutate that attribute, but otherwise we should only allow it to be set when initializing the controller.

- The `clientId` and the `user` context are kind of similar. In most sessions, those attributes shouldn't change at all after initialization, unless the user logs out and back in as another user or something. Exposing methods to mutate those attributes is still legitimate, though.

- `language` and `currency` are perhaps more likely to change, but still in most sessions, they'll probably be entirely static. Again, exposing methods to mutate those is legit.

In conclusion, I see the global commerce context as a thing that would typically be set only once in the context of a browsing session and not really mutated thereafter, so I think it makes sense to allow all of its initial state to be specified through its config options.

packages/headless/src/features/commerce/context/context-state.ts
```diff
@@ -0,0 +1,22 @@
...
+  clientId: string;
+  context: ContextParams; // TODO: Should this be spread in its parent instead?
+}
+
+export const getContextInitialState = (): CommerceContextState => ({
+  trackingId: '',
+  language: '',
+  currency: '',
+  clientId: '',
+  context: {
```
fbeaudoincoveo, line 17: I wonder why we even need the context > view part anymore if there's just url under there + it's mandatory for PLP 😁 

I think I made a comment to that effect in the excel file where the plp v2 API request body is described

packages/headless/src/features/commerce/context/context-state.ts
```diff
@@ -0,0 +1,22 @@
+import {ContextParams} from '../../../api/commerce/commerce-api-params';
+
+export interface CommerceContextState {
+  trackingId: string;
+  language: string;
+  currency: string;
+  // TODO: Does it make sense for the clientId to be part of the context state?
+  clientId: string;
+  context: ContextParams; // TODO: Should this be spread in its parent instead?
```
fbeaudoincoveo, line 9: See my comment below, I think we should just have `url` instead. That said, this should be on the PLP controller I think.

packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,212 @@
...
+export interface CartItem {
+  product: Product;
+  quantity: number;
+}
+
+export interface ContextOptions {
+  trackingId?: string;
+  language?: string;
+  currency?: string;
+  // TODO: Does it make sense for the clientId to be part of the context state?
```
fbeaudoincoveo, line 36: I'm personally OK with that for now, yes.

packages/headless/src/controllers/commerce/context/headless-context.ts
Spuffynism: Agreed! Left them exposed on this controller.

packages/headless/src/controllers/commerce/context/headless-context.ts
Spuffynism: I removed `labels` but left `view` for now since we're still figuring out which controller should handle the `url`.

packages/headless/src/controllers/commerce/context/headless-context.ts
Spuffynism: Created a cart controller! Great idea!

Left this comment unresolved while we figure out by which controller we handle the `view` (https://github.com/coveo/ui-kit/pull/3190/commits/9382bfa695c12e5b51aa4f16ec4f1b3e2acd1a80#r1330518824)

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -0,0 +1,151 @@
...
+export interface CartOptions {
+  cart?: CartItem[];
+}
+
+export interface CartItem {
+  product: Product;
+  quantity: number;
+}
+
+export interface Product {
```
olamothe, line 39: Are these properties really all optional ? Even `sku` ?

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -0,0 +1,151 @@
...
+   * Removes a cart item.
+   * @param item - The cart item to remove
+   */
+  removeCartItem(item: CartItem): void;
+
+  /**
+   * Updates a cart item's quantity.
+   * @param item - The cart item to update.
+   */
+  updateCartItemQuantity(item: CartItem): void;
```
olamothe, line 79: That's a kinda weird function signature for what it needs to do.

I would see that more as an `updateCartItem` ?

Nothing specific about quantity.

packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,187 @@
...
+export interface ContextOptions {
+  trackingId?: string;
+  language?: string;
+  currency?: string;
+  clientId?: string;
+  user?: User;
+  view: View;
+}
+
+export interface User {
```
olamothe, line 46: Kind of similar as my above comment: In general, interface where all properties are optional makes me suspicious 😉 

Wouldn't a `userId` always be required ?

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -0,0 +1,151 @@
...
+    ...controller,
+
+    get state() {
+      const {cart, cartItems} = getState().cart
+      return {
+        cart: cartItems.map((id: string) => cart[id]) as CartItem[]
+      };
+    },
+
+    setCart: (cart: CartItem[]) => dispatch(setCart({cart})),
```
fbeaudoincoveo, line 136: In general, the word "card" in the method names feels redundant to me.

I presume implementers would typically store the Cart controller instance in a variable named "cart", so I think it would feel more natural to do `cart.setItems`, `cart.addItem`, etc. than `cart.setCartItems`, `cart.addCartItem`, etc.

packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,187 @@
...
+    setContext({
+      ...options,
+    })
+  );
+
+  return {
+    ...controller,
+
+    get state() {
+      const {trackingId, language, currency, clientId, user, view} = getState().commerceContext;
```
olamothe, line 156: Are there properties we would not want to be returned here from the context ?

Or that we would like to modify/do some processing here before returning them ?

If not, could we simply do:

```suggestion
      get state() {
          return getState().commerceContext;
      }
```

packages/headless/src/features/commerce/context/cart/cart-slice.ts
```diff
@@ -0,0 +1,49 @@
+import {createReducer} from '@reduxjs/toolkit';
+import {CartState, getCartInitialState} from './cart-state';
+import {addCartItem, removeCartItem, setCart, updateCartItemQuantity} from './cart-actions';
+import {ProductParam} from '../../../../api/commerce/commerce-api-params';
+
+const generateId = (product: ProductParam) => [
```
olamothe, line 6: Related to my comment above about all properties being optional on product, this means we can generate empty string for an ID.

packages/headless/src/features/commerce/context/cart/cart-actions.ts
```diff
@@ -0,0 +1,58 @@
...
+import { ArrayValue, NumberValue, RecordValue } from '@coveo/bueno';
+
+export interface SetCartPayload {
+  cart: CartItemParam[]
+}
+export const setCart = createAction(
+  'commerce/cart/set',
+  (payload: SetCartPayload) =>
+    validatePayload(payload, {
+      cart: new ArrayValue({
```
olamothe, line 13: This schema seems a copy of this of what's already in headless-cart.ts.

Since it's a complex one, it would be worth it to define it here as a public module variable and reuse in both files.

packages/headless/src/features/commerce/context/cart/cart-slice.ts
```diff
@@ -0,0 +1,49 @@
...
+].filter(Boolean).join('_')
+
+export const cartReducer = createReducer(
+  getCartInitialState(),
+
+  (builder) => {
+    builder
+      .addCase(setCart, (state, {payload}) => {
+        const {cart, cartItems} = payload.cart.reduce((acc, item) => {
+          const id = generateId(item.product)
```
olamothe, line 19: Should there be any verification on clashing generated IDs ?

packages/headless/src/features/commerce/context/cart/cart-slice.ts
```diff
@@ -0,0 +1,49 @@
...
+              [id]: item.product
+            }
+          }) as CartState;
+        }, getCartInitialState());
+
+        state.cartItems = cartItems;
+        state.cart = cart;
+      })
+      .addCase(addCartItem, (state, {payload}) => {
+        const id = generateId(payload.product)
```
olamothe, line 33: Similar to above: Should there be any verification that we are not overwriting an existing item (since the action is `addCartItem`) ?

packages/headless/src/features/commerce/context/cart/cart-slice.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        state.cartItems = [...state.cartItems, id]
+        state.cart[id] = payload
+      })
+      .addCase(removeCartItem, (state, {payload}) => {
+        const id = generateId(payload.product)
```
olamothe, line 39: Similar to above: Should there be any verification that this is not a no-op and we are really deleting an item ?

packages/headless/src/features/commerce/context/cart/cart-slice.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        state.cartItems = state.cartItems.filter((itemId) => itemId !== id)
+        delete state.cart[id]
+      })
+      .addCase(updateCartItemQuantity, (state, {payload}) => {
+        const id = generateId(payload.product)
```
olamothe, line 45: Similar to above: Should there be any verification that the ID point to something currently in the cart ?

packages/headless/src/features/commerce/context/cart/cart-slice.ts
fbeaudoincoveo, line 6: Do we really need to do this?

The API expects the groupId, productId, and sku to be passed as separate arguments, so we'd have to convert them back anyways before making the request. If, for instance, groupId contained underscore characters before we joined all three parts, we'll have a hard time splitting it back as it must be.

packages/headless/src/features/commerce/context/context-actions.ts
```diff
@@ -0,0 +1,71 @@
...
+  currency?: string;
+  clientId?: string;
+  user?: UserParams;
+  view: ViewParams;
+}
+
+export const setContext = createAction(
+  'commerce/setContext',
+  (payload: SetContextPayload) =>
+    validatePayload(payload, {
```
olamothe, line 18: Similar to my above comment: This complex schema seems a copy of this of what's already in headless-context.ts.

packages/headless/src/features/commerce/context/context-actions.ts
```diff
@@ -0,0 +1,71 @@
+import { createAction } from '@reduxjs/toolkit';
+import {nonEmptyString, requiredNonEmptyString, validatePayload} from '../../../utils/validate-payload';
+import {UserParams, ViewParams} from '../../../api/commerce/commerce-api-params';
+import { RecordValue } from '@coveo/bueno';
+
+export interface SetContextPayload {
+  trackingId?: string;
```
fbeaudoincoveo, line 7: Should probably not be optional

packages/headless/src/integration-tests/commerce.test.ts
```diff
@@ -0,0 +1,49 @@
+import {buildCart, buildCommerceEngine, buildProductListing} from '../commerce.index';
+import {getOrganizationEndpoints} from '../insight.index';
+import {buildContext} from '../controllers/commerce/context/headless-context';
+
+let accessToken = 'no'
+const sleep = (seconds: number) =>
+  new Promise((resolve) => setTimeout(resolve, seconds * 1000));
+
+describe('commerce', () => {
+  it("'s working great",
```
fbeaudoincoveo, line 10: 😂

packages/headless/src/features/commerce/context/context-slice.ts
```diff
@@ -0,0 +1,47 @@
...
+import {getContextInitialState} from './context-state';
+import {setClientId, setContext, setCurrency, setLanguage, setTrackingId, setUser, setView} from './context-actions';
+
+export const contextReducer = createReducer(
+  getContextInitialState(),
+
+  (builder) => {
+    builder
+      .addCase(setContext, (state, {payload}) => {
+        if (payload.trackingId) {
```
olamothe, line 11: This reducer does not really behave like I would expect a `setFoo` to behave, ie: overwrite the full context, even undefined properties.

It seems it would be impossible, with the actions that are defined here, to remove any values.

Should it just be :

```typescript
 .addCase(setContext, (state, {payload}) => {
       state = payload;
}

File: packages/headless/src/features/commerce/product-listing/product-listing-actions.ts [hunk 22]
```diff
@@ -90,20 +74,24 @@ export const fetchProductListing = createAsyncThunk<
...
-    context: state.productListing.context,
+    trackingId: state.commerceContext.trackingId,
+    language: state.commerceContext.language,
+    currency: state.commerceContext.currency,
+    clientId: state.commerceContext.clientId,
+    context: {
+      view: state.commerceContext.view,
+      user: state.commerceContext.user,
+      cart: state.cart.cartItems.map((id) => state.cart.cart[id])
+    },
```
olamothe, line 94:
```suggestion
    ...state.commerceContext,
    context: {
        ...state.commerceContext,
        cart: state.cart.cartItems.map((id) => state.cart.cart[id])
    }
```

olamothe (changes requested):

packages/headless/src/integration-tests/commerce.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        product: {
+          sku: 'small green boots'
+        },
+        quantity: 3
+      })
+
+      const productListing = buildProductListing(engine)
+      productListing.refresh()
+      await sleep(2);
+      console.log('final state!!!', productListing.state)
```
fbeaudoincoveo, line 47: We could turn this into an actual expect hehe

fbeaudoincoveo (commented):
Starting to look good 😄 

I'll do another pass of review, but here's a first volley of comments!

packages/headless/src/features/commerce/context/cart/cart-slice.ts
fbeaudoincoveo: oh nm, I understand why you're doing this here hehe

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
Spuffynism: Yes, according to the API spec, they're all optional: https://docs.google.com/spreadsheets/d/1-WFpnipA3fgwCX3GhvnmJvPSwO_jfiQY-lZytsapfKs/edit?pli=1#gid=0&range=D20:D22

packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,187 @@
+import {RecordValue, Schema } from "@coveo/bueno";
```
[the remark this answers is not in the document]
Spuffynism: Yup, completing the controller tests is my next step!

packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -0,0 +1,187 @@
...
+  view: new RecordValue({
+    options: { required: true },
+    values: {
+      url: requiredNonEmptyString,
+    }
+  })
+});
+
+export interface ContextOptions {
+  trackingId?: string;
```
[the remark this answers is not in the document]
Spuffynism: Agreed, made it required. Also made the `language`, `currency`, and `clientId` params required since they're required by the API!

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -0,0 +1,151 @@
...
+  setCart,
+  updateCartItemQuantity
+} from '../../../../features/commerce/context/cart/cart-actions';
+
+const optionsSchema = new Schema({
+  cart: new ArrayValue({
+    each: new RecordValue({
+      values: {
+        product: new RecordValue({
+          options: {required: true}
```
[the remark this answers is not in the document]
Spuffynism: You're right! it's missing some product properties! Will add them.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export interface CartItem {
+  product: Product;
+  quantity: number;
+}
+
+export interface Product {
+  groupId?: string;
+  productId?: string;
```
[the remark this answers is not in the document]
Spuffynism: Yeah, I agree. I'll change the type to make it require at least one property.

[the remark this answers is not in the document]
Spuffynism: See https://github.com/coveo/ui-kit/pull/3190#discussion_r1337359224

packages/headless/src/features/commerce/context/cart/cart-slice.ts
Spuffynism: What would you think of exposing a `setItem` and `removeItem` actions instead of actions for adding, removing and updating the quantity of an item? This would front the clashing to clients. 

Plus, cart items are now indexed by their productId.

packages/headless/src/features/commerce/context/context-slice.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const contextReducer = createReducer(
+  getContextInitialState(),
+
+  (builder) => {
+    builder
+      .addCase(setContext, (_, {payload}) => {
+        return payload
+      })
+      .addCase(setTrackingId, (state, {payload}) => {
```
olamothe, line 13: We could have a single `setContext` action.

And in `headless-context.ts`: 

```typescript
setTrackingId: (trackingId: string) => dispatch(setContext({...getState().commerceContext, trackingId})),
```

packages/headless/src/integration-tests/commerce.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      let productListing = buildProductListing(engine);
+      await waitForNextStateChange(engine, {
+        action: () => {
+          productListing.refresh()
+        },
+        expectedSubscriberCalls: 2,
+      });
+
+      expect(productListing.state.products).toEqual([
```
olamothe, line 53: Instead of that kind of `strict-equal` on every single property, you can leverage jest to do partial object match.

Also, seems like a lot of these values are ... temporary ? 🤔 ie: will it start to break builds if the underlying org commerce org changes ?

olamothe (approved):

packages/headless/src/features/commerce/context/cart/cart-slice.ts
```diff
@@ -0,0 +1,36 @@
...
+              ...acc.cart,
+              [item.productId]: item
+            }
+          }) as CartState;
+        }, getCartInitialState());
+
+        state.cartItems = cartItems;
+        state.cart = cart;
+      })
+      .addCase(addItem, (state, {payload}) => {
```
olamothe, line 24: I see these are now marked as outdated comments from my previous review, but they would still apply:

There need to be defensive code against productId being invalid/absent before we do any operation in the reducers.

packages/headless/src/integration-tests/commerce.test.ts
Spuffynism: Agreed! Just wanted a quick and dirty way of validating that the end-to-end commerce service call worked. Will adjust the test in consequence to use a valid token, use a different org, and better match against expected results.

packages/headless/src/features/commerce/context/context-slice.ts
Spuffynism: Are there best practices, or guidelines for when to create an action, and when to reuse an existing one?

I'll make the change to use `setContext` for all 4 top-level `string` properties (`trackingId`, `language`, `currency`, `userId`) in the meantime.

packages/headless/src/features/commerce/context/cart/cart-slice.ts
Spuffynism: I've asked about possibly changing the methods we expose on the controller, to limit how defensive we need to be here: https://github.com/coveo/ui-kit/pull/3190#discussion_r1337658071

packages/headless/src/integration-tests/commerce.test.ts
olamothe: You can `.skip(...)` it in the meantime, if it's not ready.

packages/headless/src/features/commerce/context/context-slice.ts
olamothe: > Are there best practices, or guidelines for when to create an action, and when to reuse an existing one?

It's generally better to limit the amount of symbols that gets exported publicly from headless (just to limit the amount of potential breaking change to maintain on our side down the line). And all actions are exported publicly.

In this case I don't think there's much value added to have one action for every single property.
It would be different if we needed to do manipulation/verification/validation etc.. before setting a value in the store.

packages/headless/src/features/commerce/context/cart/cart-slice.ts
olamothe: > This would front the clashing to clients.

What do you mean by this ? I'm not sure I understand

Spuffynism: The `setItem` and `removeItem` method names expose that the controller only sets and removes items.

The cart is effectively a map of product id -> quantity. By exposing only `setItem` and `removeItem` methods on the controller, the cart controller becomes dumber. This puts the responsibility of dealing with potential item clashing on the caller instead of on the callee (cart controller).

---
A second option is to deal with clashes:
- Expose `addItem(item: CartItem)`, `removeItem(item: CartItem)` and `updateItemQuantity(productId: string, quantity: number)` methods
- In `addItem`, if a clash is found, make the total quantity the current quantity + new item quantity. Otherwise, add the item normally.
- In `removeItem`, remove the item if it's present, otherwise be silent
- In `updateItemQuantity`, update the item quantity if it's present, otherwise be silent

packages/headless/src/features/commerce/context/cart/cart-slice.ts
Spuffynism: See https://github.com/coveo/ui-kit/pull/3190#discussion_r1337208140

packages/headless/src/features/commerce/context/cart/cart-slice.ts
Spuffynism: See https://github.com/coveo/ui-kit/pull/3190#discussion_r1337208140

packages/headless/src/controllers/commerce/context/headless-context.ts
Spuffynism: Great point. I've relayed this question to the team channel, as it's not something the API makes explicit: https://coveo.slack.com/archives/C05FR3KKBHB/p1695913390115019

Spuffynism: We'll require either a userId or an email. I'll adjust the PR accordingly!

Spuffynism: Done https://github.com/coveo/ui-kit/pull/3190/commits/eff279c4b77d2f02990b2ae2d5848240b24beb8a

packages/headless/src/features/commerce/context/cart/cart-slice.ts
olamothe: > This puts the responsibility of dealing with potential item clashing on the caller instead of on the callee (cart controller).

If by the caller you mean the implementer, this is not a good idea. 

If there is a clash, we just need to inform implementers, and we don't need to do anything fancy.


> * In addItem, if a clash is found, make the total quantity the current quantity + new item quantity. Otherwise, add > > the item normally.
> * In removeItem, remove the item if it's present, otherwise be silent
> * In updateItemQuantity, update the item quantity if it's present, otherwise be silent

For all 3 scenarios, just a no-op/return early + log an informative warning message (not silent) should do the trick.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -0,0 +1,131 @@
...
+  options?: CartOptions;
+}
+
+/**
+ * The `Cart` controller exposes methods for managing the cart in a commerce interface.
+ */
+export interface Cart extends Controller {
+  /**
+   * Sets the cart items.
+   * @param cart - The new cart.
```
dmbrooke, line 39: Based on this parameter description/name, I'd expect `cart` to be a Cart instance rather than an array of CartItems...

```suggestion
   * @param cartItems - The new cart items.
```

We could make it more wordy to clarify that this _**replaces any existing items in the cart**_ (at least I assume that's what it would do), but I think changing the param name/description should be enough.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -0,0 +1,131 @@
...
+   * Adds a cart item.
+   * @param item - The new cart item.
+   */
+  addItem(item: CartItem): void;
+
+  /**
+   * Removes a cart item.
+   * @param productId - The cart item's product id to remove
+   */
+  removeItem(productId: string): void;
```
dmbrooke, line 53: If a user has added, say, 3 of a single item to their card, will it remove _**all**_ of that item? Or will it only remove 1 at a time?

Based on the fact that `quantity` is a part of the `CartItem` interface, and there's a `updateItemQuantity` method below, I'll assume it's the former. That being said, it may be worthwhile to clarify in the description.

btaillon (approved):

dmbrooke (approved):
A few questions/suggestions for docs strings, but looks good!

packages/headless/src/features/commerce/context/cart/cart-slice.ts
Spuffynism: I've applied your suggestions, without the logging. I didn't log since other slices do not log upon edge cases.

I could throw an Error, but it doesn't seem quite right since these are edge cases and not problematic inputs.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
Spuffynism: It will remove **all** of that item! Since the method exposes only a `productId`, it seems clear enough to me that all of that item's quantity will be removed. Should that be made more explicit through a docstring? Something like this?
```
* Remove all of an item's quantity
```

See: https://github.com/coveo/ui-kit/pull/3190/files#diff-c81cbfcfb6410f7305a9ccc75ecf5f40ff139616b69386fde2ff97ee1d532ec6R38-R43

packages/headless/src/api/commerce/commerce-api-params.ts
```diff
@@ -26,25 +26,36 @@ export interface ContextParam {
...
-export interface CartParams {
-  groupId?: string;
-  productId?: string;
-  sku?: string;
+interface UserIdAndEmail {
+  userId: string;
+  email: string;
+}
+
+export type UserParams = (UserIdRequired | EmailRequired | UserIdAndEmail) & {
```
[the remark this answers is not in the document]
dmbrooke: > See https://github.com/coveo/ui-kit/pull/3190#discussion_r1337184673
> We want to require at least a user id, or at least an email, or both.

I understand that part, but won't we need to have an if-condition within the class/method logic itself to ensure that at least one of the fields is present in the object in order to appease the type check? If it's possible for _either field to be undefined_ without us knowing which field _will be defined_, we'll have to rely on an if-condition to ensure at least one is defined to appease the type check anyways. And if we need to do that, it's functionally the same as using a Partial of the `UserIdAndEmail` interface, right?

Alternatively, if we know that one of the fields will always be populated within a certain context based on its type (e.g., email is populated by a method call that returns a string), then such an if-condition isn't necessary and it still passes the type-check.

This will be my last comment arguing this, I promise 😅 If I'm overlooking something that you're already accounting for, ignore this comment and carry on. I just wanted to elaborate on my thought-process and why I'm uncertain about this being necessary.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
dmbrooke: Thanks for clarifying! Honestly, now that you mention it, it's not confusing once you realize that both the `productId` _**and**_ `quantity` are bound to the same `CartItem` object. The more I reviewed this PR, the more obvious it became, so the original description is likely enough for implementers to understand what's going on 😅

packages/headless/src/api/commerce/commerce-api-params.ts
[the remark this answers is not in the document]
Spuffynism: The type definition is enough to ensure we do not need to do an if/else to ensure it's valid! Plus, the user is sent pretty much as-is to the API; we don't need to manipulate it much.

However, you are right that we'll need an `if` condition to do validation (in the slice): https://github.com/coveo/ui-kit/pull/3190/files#diff-21dff18bbe200c24fe0d214536833d65408d08b0f38b3e26b1d1546da94132e3R36

> And if we need to do that, it's functionally the same as using a Partial of the UserIdAndEmail interface, right?

Not sure I get this. By having the type defined as-is, for typescript users of headless, we can enforce the requirement. For javascript users, we have the condition as defined above.

fbeaudoincoveo (approved):

PR #3200, opened 2023-09-20
fix(atomic): css part and css variable for facet-search-icon / facet-search-clear-icon

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2752

Both css part and css variable was not working for neither facet-search-icon and facet-search-clear-icon.

This PR solves it by putting the part inside the atomic-icon directly and by changing the css rules for the variables.

### Results

Now these two methods work for affecting the size of the icon.

```
atomic-facet::part(search-icon) {
        width: 30rem;
        height: 30rem;
      }
```

```
 :root {
        --atomic-facet-search-icon-size: 30rem;
      }
```


olamothe (approved):

dmbrooke (approved):
LGTM

PR #3202, opened 2023-09-21
fix(atomic): make quickview modal responsive

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2767

This PR makes the quickview modal more responsive with these changes:
- Make the atomic-modal full screen when mobile mode
- Truncate the header title text
- Center the footer "Result of text"
- 'minimizeSidebar' state default value is true if app in mobile mode

### Problem
I want to use the function in the state isMobile() to check if the app is in mobile mode. This works fine in the render function for the full screen thing but it doesn't work for a state since the Bindings are not initialized whenever it is time for the states. 
My solution was just to directly query the atomic-search-layout to find its mobile-breakpoint attribute. I don't really like it but it works. Logically i would use this [function](https://github.com/coveo/ui-kit/blob/0774bf4d03d8ffa11e17d41197e880ca2f1b4889/packages/atomic/src/components/search/atomic-search-interface/store.ts#L74) but since it doesn't work for all of my use cases...

### Before

image

### After

image


packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -82,6 +83,16 @@ export class AtomicQuickviewModal implements InitializableComponent {
 
   private interactiveResult?: InteractiveResult;
 
+  private get isMobile() {
```
olamothe, line 86: This is all available from the stencil store.

```typescript
this.bindings.store.isMobile()
```

olamothe (changes requested):

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
alexprudhomme: My problem with this is that it doesn't work for initializing a State variable since the store is not initialized yet.

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
```diff
@@ -82,6 +82,10 @@ export class AtomicQuickviewModal implements InitializableComponent {
 
   private interactiveResult?: InteractiveResult;
 
+  public componentWillRender(): void {
+    this.minimizeSidebar = this.bindings.store.isMobile();
```
olamothe, line 86: This behaviour should be documented, in JSdoc, on the property

olamothe (approved):

packages/atomic/src/components/search/result-template-components/atomic-quickview-modal/atomic-quickview-modal.tsx
alexprudhomme: Does your comment still apply if `minimizeSidebar` is a state ?

packages/atomic/cypress/e2e/result-list/result-components/quickview.cypress.ts
```diff
@@ -24,8 +25,7 @@ const addQuickviewInResultList = (props: TagProps = {}) =>
   );
 
 const openModal = () => {
-  cy.wait(5000);
-  QuickviewSelectors.button().click();
+  QuickviewSelectors.button().click({timeout: 8000});
```
alexprudhomme, line 28: increase the timout from default value 4000 to 8000 to wait without using cy.wait()

alexprudhomme:
@louis-bompart adressed your comments. Waiting on your review to merge.

louis-bompart (approved):

PR #3203, opened 2023-09-21
refactor(headless,headless-react): prepared SSR-related code for new use cases to be added

btaillon:
https://coveord.atlassian.net/browse/KIT-2769

When working on adding support for other use cases with SSR (such as product recommendation or product listing), I found myself doing these changes no matter what. Since it's a lot of small changes, and they may be needed for multiple PRs, I decided to extract them into their own issue & PR.

In a nutshell, the main changes are:
* Refactored `defineSearchEngine` and extracted some functions out of it so they can be re-utilized by `defineProductRecommendationEngine` and `defineProductListingEngine`.
  * This also resulted in slightly fewer lines of code. Most importantly, it made the code easier to read IMO (especially the middleware stuff).
  * Utilities inside `defineSearchEngine` no longer return a `new Promise`, they are simple `async` functions, thanks to a new utility I added called `createDeferredPromise`.
* Renaming the `searchFulfilledAction` parameter of `hydrateStaticState` to `searchAction`.
  * I believe that at some point I mistakenly made rejected search actions throw an error on the server side. In this PR, I changed it so that we now pass rejected search actions to the client just like we do for fulfilled search actions.
* Extracting search-specific interface(s) out of "common" files.


packages/headless-react/src/ssr/index.ts
```diff
@@ -1,2 +1,6 @@
-export {defineSearchEngine} from './search-engine.js';
+export {
+  defineSearchEngine,
+  MissingEngineProviderError,
+} from './search-engine.js';
+export type {ReactSearchEngineDefinition} from './search-engine.js';
```
btaillon, line 5: These two exports seemed to be missing.

packages/headless/src/app/ssr-engine/types/common.ts
```diff
@@ -122,16 +122,22 @@ export type InferControllersMapFromDefinition<
   TControllers extends ControllerDefinitionsMap<CoreEngine, Controller>,
 > = {[K in keyof TControllers]: InferControllerFromDefinition<TControllers[K]>};
 
-export type InferControllerStaticStateFromDefinition<
-  TDefinition extends ControllerDefinition<CoreEngine, Controller>,
-> = TDefinition extends ControllerDefinition<infer _, infer TController>
-  ? ControllerStaticState<TController['state']>
-  : never;
```
btaillon, line 129: Basically, we went from:
* Infer static state map from definitions map = Infer static state from definition

To:
* Infer static state map from controllers map = Infer static state from controller
* Infer static state map from definitions map = Infer controllers map from definitions map = Infer static state map from controllers map

packages/headless/src/app/search-engine/search-engine.ssr.ts
```diff
@@ -37,6 +41,12 @@ export type SearchEngineDefinitionOptions<
   TControllers extends ControllerDefinitionsMap<SearchEngine, Controller>,
 > = EngineDefinitionOptions<SearchEngineOptions, TControllers>;
 
+function isSearchCompletedAction(
```
mrrajamanickam-coveo, line 44: isSearchCompletedAction -> isSearchActionCompleted ?

btaillon: oh this is meant as a typeguard, checking whether the parameter is of type "SearchCompletedAction" essentially.

To make it more readable, I'll add a type alias called `SearchCompletedAction` which is a union of the fulfilled and rejected actions.

mrrajamanickam-coveo (approved):

mrrajamanickam-coveo (approved):

dmbrooke (approved):
Looks good! Is there anything we should/can test locally before merging?

btaillon:
> Looks good! Is there anything we should/can test locally before merging?

Could test the `packages/samples/headless-ssr` package, but there's no need.

PR #3204, opened 2023-09-21
docs(headless): facet value state: enlists the possible values of filter value state

npushkarskii:
JIRA: https://coveord.atlassian.net/browse/DOC-13275 
Discuss: https://discuss.coveo.com/t/possible-values-for-staticfiltervalue/10107


alexprudhomme (approved):
The title of this PR when merging should be **doc**(headless) : FacetValueState: enlists the possible values of FilterValueState. Using **feat** will trigger a minor version bump and it is not needed for this kind of change.

npushkarskii:
> The title of this PR when merging should be **docs**(headless) : FacetValueState: enlists the possible values of FilterValueState. Using **feat** will trigger a minor version bump and it is not needed for this kind of change.

Thank you for clarifying 🙏 Initially, it was exactly like you said, `docs(headless)...`, but it caused a linter error :( 
@alexprudhomme 

image

npushkarskii:
@alexprudhomme forgot to mention, but I cannot merge the PR myself. If you can help with that, I'd be very grateful 😇

olamothe (approved):
The types of `StaticFilterValueState` should be self explanatory for people in TypeScript 🤔 

So, said differently, i's a nice doc improvement indeed, but there are hundreds of places in the doc where we rely on the types linked to a symbol to explain "what it is" 😅 

So it's a bit weird that the original dev here: https://coveord.atlassian.net/browse/DOC-13275

would complain only about this specific property.

npushkarskii:
> So it's a bit weird

Agreed, I found it a bit strange as well 😯 Hope they would be happy after that small update, hehe.

---

@olamothe Btw, could you plz help me with merging of this PR? I don't have permissions for doing so 😇

btaillon (approved):

olamothe:
> > So it's a bit weird
> 
> Agreed, I found it a bit strange as well 😯 Hope they would be happy after that small update, hehe.

I think the larger comment here would be to think back on what gets generated on the doc website: The fact that symbols are not properly auto documented is... meh 🤷 

So, band-aid kinda thing. 
But, changing this would be a much larger endeavour.

> @olamothe Btw, could you plz help me with merging of this PR? I don't have permissions for doing so 😇

Sure !

PR #3205, opened 2023-09-21
feat(atomic): add pathLimit prop to atomic-breadbox

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2758

This PR adds a new prop to atomic-breadbox. It's purpose is as follows

  ```
/**
   * This prop allows you to control the display of a path by
   * limiting the number of visible values in the path.
   *
   * If the path size is equal to or less than the pathLimit, all values in
   * the path will be displayed without truncation.
   *
   * If the path size exceeds the pathLimit, it will truncate the path by
   * replacing the middle values with ellipses ('...').
   *
   * Minimum: `3`
   * Maximum: `10`
   * @defaultValue `3`
   */
```
First came from this discuss : https://discuss.coveo.com/t/support-00103676-expose-limitpath-option-in-atomic-breadbox/10067/2 

### Examples

path size of 4, path limit of 3, it gets truncated
image

path size of 4, path limit of 4, it doesnt get truncated

image

_**NEW EDGE CASE OF 1**_
path size of 3, path limit of 1 (only last value gets shown)

image

### CODE REVIEW

File: packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx [hunk 1]
```diff
@@ -79,7 +80,24 @@ export class AtomicBreadbox implements InitializableComponent {
...
+   * limiting the number of visible values in the path.
+   *
+   * If the path size is equal to or less than the pathLimit, all values in
+   * the path will be displayed without truncation.
+   *
+   * If the path size exceeds the pathLimit, it will truncate the path by
+   * replacing the middle values with ellipses ('...').
+   *
+   * Minimum: `3`
+   * Maximum: `10`
```
y-lakhdar, line 94: Is there a reason for that value? Can't we have a higher maximum?

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
```diff
@@ -79,7 +80,24 @@ export class AtomicBreadbox implements InitializableComponent {
 
   private breadcrumbShowLessFocus?: FocusTargetController;
 
+  /**
+   * This prop allows you to control the display of a path by
+   * limiting the number of visible values in the path.
```
y-lakhdar, line 85: Maybe it's me but, `values` does not seem like the right term here. Can we say `directory` or `level`?

y-lakhdar (approved):

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
olamothe: Minimum should be `1` because that's just a logical/easily understandable limit.

For maximum, I'd say we should support infinite length if an implementer specifically configure it that way.

Yeah, it would break the rendering/would look silly with hundreds of values, but 🤷

The upper limit of 10 here is too arbitrary. 
We can add a warning I guess, that any big value can cause weird stuff to happen in the template.

And, I'd like if we did some manual test to check what happens if we mock a value with hundreds of values, and try to display them all.

Just to get an idea about how does our standard result templates sections behave.

packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx
louis-bompart: Mh yeah this could be phrased a bit better maybe.
I think I would go with something like:
>  This prop allows you to control how many ancestor/parent values of the selected value to display.

Maybe @spiracha-coveo could chime in and give us some API Writer magic 🗡️

spiracha-coveo: 👋 
Is this more clearer:
> This prop allows you to control the display depth of the path by specifying the number of parent or ancestor breadcrumb links relative to the currently selected value.

alexprudhomme:
@olamothe about this PR. Last week, you said that we should maybe change the behaviour if there was breadbox with very big paths. I think it is fine as it is. The client need was to be able to see the whole path and they now can by using this prop. Even if the path is too long so that the value truncates like this :
image


The client is still able to make the breadbox bigger and allow for more space by using this part :
```
     atomic-breadbox::part(breadcrumb-value) {
        max-width: 95ch;
      }
```
result : 
image

What do you think ?

olamothe:
> @olamothe about this PR. Last week, you said that we should maybe change the behaviour if there was breadbox with very big paths. I think it is fine as it is. The client need was to be able to see the whole path and they now can by using this prop. Even if the path is too long so that the value truncates like this : image
> 
> The client is still able to make the breadbox bigger and allow for more space by using this part :
> 
> ```
>      atomic-breadbox::part(breadcrumb-value) {
>         max-width: 95ch;
>       }
> ```
> 
> result : image
> 
> What do you think ?

Sure !

olamothe (approved):

PR #3206, opened 2023-09-21
fix(quantic): spacing on expandable searchbox

erocheleau:
Before:
image

After:
image

Just a small spacing fix. The spacing is the same as the spacing on the other side of the x


packages/quantic/force-app/main/default/lwc/quanticSearchBox/templates/expandableSearchBox.css
```diff
@@ -18,6 +18,7 @@ textarea.searchbox__input {
   border: none;
   box-shadow: none;
   overflow-x: clip;
+  margin-right: 0.8rem;
```
SimonMilord, line 21: Looks fine to me, but is the 0.8rem an arbitrary value or is there a reason behind it?

SimonMilord (approved):

mmitiche (approved):

packages/quantic/force-app/main/default/lwc/quanticSearchBox/templates/expandableSearchBox.css
erocheleau: It's the same value as the lightning design system value that is on the sides of the icon next to it 🤷‍♂️

PR #3208, opened 2023-09-22
feat(headless,headless-react): Added product recommendations SSR utilities

btaillon:
https://coveord.atlassian.net/browse/KIT-2769


btaillon:
Closing this. It was useful for testing purposes and knowing what code to extract, making it easier to add cases in the future. However, I can't justify spending the time to add tests in this since we're planning to eventually replace product recommendations.

We can reopen and merge this in the future if we really need the product recommendations case. Otherwise, this can serve as an example.

btaillon:
Closing this. It was useful for testing purposes and knowing what code to extract, making it easier to add cases in the future. However, I can't justify spending the time to add tests in this since we're planning to eventually replace product recommendations.

We can reopen and merge this in the future if we really need the product recommendations case. Otherwise, this can serve as an example.

PR #3216, opened 2023-09-25
ci: make smoke tests assert a number of results above 1 instead of exactly 10

btaillon:
https://coveord.atlassian.net/browse/KIT-2775

The regexp only checks that the first digit is 1-9. 0 would be unexpected (and usually should be a "no results" page anyway). The digit before the dash should always be 1, since we didn't press anything on the pager.


louis-bompart (approved):

louis-bompart:
The double contains is still concerning tho 🤔

PR #3217, opened 2023-09-25
ci: remove dependencies of "prerelease" job

btaillon:
https://coveord.atlassian.net/browse/KIT-2776

Prereleases don't need e2e tests IMO, since they aren't "official". This should ensure we can have NPM packages sooner, even when e2e tests are taking a while or are flaky for some reason.

I didn't add "build" as a dependency, since the prerelease job doesn't use the cache from the "build" job.


olamothe (approved):
Good idea 💯

PR #3218, opened 2023-09-26
chore(quantic): Update deprecated cypress plugin in Quantic

SimonMilord:
[SFINT-5173](https://coveord.atlassian.net/browse/SFINT-5173)

**In this PR:**

- Updated the Quantic Package to use the [Cypress real events plugin](https://github.com/dmtrKovalenko/cypress-real-events/tree/main) instead of the [cypress plugin tab](https://github.com/kuceb/cypress-plugin-tab)
- Fixed the e2e tests that were failing

We were stuck on Cypress 10 because Quantic relies on this dependency plugin: [cypress plugin tab](https://github.com/kuceb/cypress-plugin-tab) to trigger keyboard tab events in cypress.

Note: Certain tests affected by the change were skipped previously, so I made the syntax change, but left them skipped for now as they will eventually be fixed as part of this [issue](https://coveord.atlassian.net/browse/SFINT-4875)

### DIFF
--- a/packages/quantic/cypress/e2e/default-1/tab/tab-actions.ts
+++ b/packages/quantic/cypress/e2e/default-1/tab/tab-actions.ts
@@ -10,12 +10,15 @@ function tabActions(selector: TabSelector) {
       }
     },
     findTabPressTabPressSpace: (value: string) => {
-      // .tab() is equivalent to pressing the Tab; .type(' ') is equivalent to pressing the space bar;
-      selector.tab().contains(value).tab().type(' ');
+      // This is equivalent to pressing the Tab key twice and then pressing the Space bar.
+      selector.tab().contains(value).realPress(['Tab', 'Tab', 'Space']);
     },
     findTabPressShiftTabPressSpace: (value: string) => {
-      // tab({shift: true}) is equivalent to pressing the Shift and the Tab; .type(' ') is equivalent to pressing the space bar;
-      selector.tab().contains(value).tab({shift: true}).type(' ');
+      // This is equivalent to holding the Shift key while pressing the Tab key twice and then pressing the Space bar.
+      selector
+        .tab()
+        .contains(value)
+        .realPress(['Shift', 'Tab', 'Tab', 'Space']);
     },
   };
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -14731,16 +14731,6 @@
         "ajv": "^8.8.2"
       }
     },
-    "node_modules/ally.js": {
-      "version": "1.4.1",
-      "resolved": "https://registry.npmjs.org/ally.js/-/ally.js-1.4.1.tgz",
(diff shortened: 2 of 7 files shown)


packages/quantic/cypress/e2e/default-1/tab/tab-actions.ts
```diff
@@ -10,12 +10,15 @@ function tabActions(selector: TabSelector) {
       }
     },
     findTabPressTabPressSpace: (value: string) => {
-      // .tab() is equivalent to pressing the Tab; .type(' ') is equivalent to pressing the space bar;
-      selector.tab().contains(value).tab().type(' ');
+      // This is equivalent to pressing the Tab key twice and then pressing the Space bar.
+      selector.tab().contains(value).realPress(['Tab', 'Tab', 'Space']);
```
mmitiche, line 14: why are we keeping `selector.tab()` here? isn't replaced by `.realPress(['Tab', 'Tab', 'Space'])`

packages/quantic/cypress/e2e/default-1/tab/tab-actions.ts
```diff
@@ -10,12 +10,15 @@ function tabActions(selector: TabSelector) {
...
     },
     findTabPressShiftTabPressSpace: (value: string) => {
-      // tab({shift: true}) is equivalent to pressing the Shift and the Tab; .type(' ') is equivalent to pressing the space bar;
-      selector.tab().contains(value).tab({shift: true}).type(' ');
+      // This is equivalent to holding the Shift key while pressing the Tab key twice and then pressing the Space bar.
+      selector
+        .tab()
```
mmitiche, line 19: Same here, why are we keeping the `.tab()`?

olamothe (approved):

packages/quantic/cypress/e2e/default-1/tab/tab-actions.ts
mmitiche: same here selector.tab() is necessary.

packages/quantic/cypress/e2e/default-1/tab/tab-actions.ts
mmitiche: Ah my bad never mind my comment you can discard it, we need `selector.tab()` cause this not from the plugging, this is just the name of the selector defined in `tab-selectors.ts`

mmitiche (approved):
Please discard the last commit, LGTM!

erocheleau (approved):
All looks good to me 👍

packages/quantic/cypress/e2e/default-1/tab/tab-actions.ts
SimonMilord: yeah strangely the tests were still passing but I think I had a typo. Anyways yeah you are right it is quite confusing since they have the same name. Now, hopefully it will be less so with the `tab()` and `.realPress('Tab')`

PR #3221, opened 2023-09-26
feat(genAI): add parts

dguerinCoveo:
[SVCINT-2527](https://coveord.atlassian.net/browse/SVCINT-2527)

Adds parts to some genAI components. 

In the Search page builder, we need to override the loading bullets and the citations index colors. We need to add parts for that since its inside a shadow dom. 

[SVCINT-2527]: https://coveord.atlassian.net/browse/SVCINT-2527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
+++ b/packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
@@ -9,20 +9,20 @@ import {
   GeneratedAnswerCitation,
 } from '@coveo/headless';
 import {Component, h, State, Element} from '@stencil/core';
+import {buildCustomEvent} from '../../../utils/event-utils';
 import {
   BindStateToController,
   InitializableComponent,
   InitializeBindings,
 } from '../../../utils/initialization-utils';
 import {Heading} from '../../common/heading';
+import {LinkWithResultAnalytics} from '../../common/result-link/result-link';
 import {Bindings} from '../atomic-search-interface/atomic-search-interface';
 import {FeedbackButton} from './feedback-button';
-import {RetryPrompt} from './retry-prompt';
-import {TypingLoader} from './typing-loader';
-import {SourceCitations} from './source-citations';
 import {GeneratedContentContainer} from './generated-content-container';
-import {LinkWithResultAnalytics} from '../../common/result-link/result-link';
-import {buildCustomEvent} from '../../../utils/event-utils';
+import {RetryPrompt} from './retry-prompt';
+import {SourceCitations} from './source-citations';
+import {TypingLoader} from './typing-loader';
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/search/atomic-generated-answer/typing-loader.tsx
```diff
@@ -1,9 +1,9 @@
 import {FunctionalComponent, h} from '@stencil/core';
 
 export const TypingLoader: FunctionalComponent = () => (
-  <div class="typing-indicator" aria-hidden="true">
-    <span></span>
-    <span></span>
-    <span></span>
+  <div part="container-background" class="typing-indicator" aria-hidden="true">
```
nathanlb, line 4: "container" doesn't mean much here.
"typing-indicator" and "typing-indicator-dot"?
"typing-animation" and "typing-animation-dot"?

dguerinCoveo: Agree. Lets go for typing-animation

packages/atomic/src/components/search/atomic-generated-answer/typing-loader.tsx
nathanlb, line 1: Are you sure we need to add parts to this component?
[Here](https://github.com/coveo/ui-kit/blob/3145d738b6100cbfa237ad57b4f0f5b5a27bd018/packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.pcss#L101) I made a mistake and the color should be changed to
```css
  background-color: var(--atomic-primary);
```

If that's in place then the `--atomic-primary-background` and `--atomic-primary` variables control all the colors in this animation and I think that can be controlled in the builder templating.

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
nathanlb, line 1: Same here, normally the citation colors should all be controlled by atomic variables which I think can be overridden in a SIS stylesheet. Have we tried? 🤔

packages/atomic/src/components/search/atomic-generated-answer/typing-loader.tsx
dguerinCoveo: It stems back from that discussion here: https://coveord.atlassian.net/browse/SVCINT-2507. We'd like to override these components to NOT follow `--atomic-primary` and `--atomic-primary-background`.

Its a temp fix though in the SP builder until we find a better way to handle --atomic-primary-background in the SP builder.

dguerinCoveo: Maybe having to add parts justifies finding another solution for the problem identified in SVCINT-2507 though 🤔

dguerinCoveo:
Closing until we are sure we want to override `--atomic` variables in the SP builder

nathanlb:
> Closing until we are sure we want to override `--atomic` variables in the SP builder

TBH that's what i thought we were doing to theme the interfaces 😅

dguerinCoveo:
> TBH that's what i thought we were doing to theme the interfaces 😅

We are already redefining `--atomic-variables` with other colors, yes. But the reason this PR was openened was to modify a component's color to change it from `--atomic-primary-background` to a specific color. Don't know if thats clear 😅

PR #3224, opened 2023-09-27
feat(atomic,headless): ga release for automatic-facets feature

olamothe:
* Remove `@beta` tag
* Add reference link to docs.coveo.com for the important pieces (atomic ui components, headless controller)

https://coveord.atlassian.net/browse/KIT-2784


louis-bompart (approved):

y-lakhdar (approved):

packages/atomic/src/components.d.ts
```diff
@@ -60,12 +60,12 @@ export namespace Components {
         "updateMessage": (region: string, message: string, assertive: boolean) => Promise<void>;
     }
     /**
-     * @beta - This component is part of the automatic facets feature.
-     * Automatic facets are currently in beta testing and should be available soon.
-     * An automatic facet is a special type of facet generated by the automatic facets feature. Unlike traditional facets that
-     * need to be explicitly defined and requested in the search query, automatic facets are dynamically generated by the index
-     * in response to the search query. It's important to note that this component should not be used directly. It is internally used
-     * by the `atomic-automatic-facet-generator` component to automatically render updated facets.
+     * An automatic facet is a special type of facet generated by the automatic facets feature.
```
dmbrooke, line 63: Are there any docs we could link to for automatic facets?

packages/atomic/src/components.d.ts
```diff
@@ -60,12 +60,12 @@ export namespace Components {
...
+     * Unlike traditional facets that need to be explicitly defined and requested in the search query, automatic facets are dynamically generated by the index
+     * in response to the search query.
```
dmbrooke, line 65: IIRC, the API writers tend to avoid `search query` since term `query` doesn't really exist outside of the context of the search.
```suggestion
     * in response to the query.
```

packages/atomic/src/components.d.ts
```diff
@@ -60,12 +60,12 @@ export namespace Components {
...
+     * It's important to note that this component is not typically used directly.
```
dmbrooke, line 66: I personally find this more clear. I was uncertain as to what it meant for a component to be used directly vs indirectly. Moreover, I think this should go _**after**_ the the next sentence that explains that it's used internally by the `atomic-automatic-facet-generator` component.
```suggestion
     * It's important to note that this component is not typically used on its own.
```

packages/atomic/src/components.d.ts
```diff
@@ -60,12 +60,12 @@ export namespace Components {
...
+     * It is internally used by the `atomic-automatic-facet-generator` component to automatically render updated facets.
+     * To read more about the automatic facet generator feature, see: https://docs.coveo.com/en/3383#about-the-facet-generator-feature
```
dmbrooke, line 68: Ah, here's the docs I was asking about 🙂 I don't feel too strongly, but perhaps this would be best to be referenced in-line where I suggested above? It's fine either way though.

packages/atomic/src/components.d.ts
```diff
@@ -75,25 +75,26 @@ export namespace Components {
         "searchStatus": SearchStatus;
     }
     /**
-     * @beta - This component is part of the automatic facets feature.
-     * Automatic facets are currently in beta testing and should be available soon.
      * The `atomic-automatic-facet-generator` is a component that renders the facets from
-     * the automatic facets feature. It dynamically creates multiple atomic-automatic-facet
-     * components based on the desiredCount prop. Automatic facets are generated by the index
+     * the automatic facets feature.
+     * Unlike traditional facets that need to be explicitly defined and requested in the search query, automatic facets are dynamically generated by the index
```
dmbrooke, line 80: If my suspicion was correct about `search query` -> `query`, be sure to apply throughout ✅

packages/atomic/src/components.d.ts
```diff
@@ -75,25 +75,26 @@ export namespace Components {
...
      * in response to the search query.
+     * It dynamically creates multiple atomic-automatic-facet
```
dmbrooke, line 82:
```suggestion
     * It dynamically creates multiple `atomic-automatic-facet`
```

packages/atomic/src/components.d.ts
```diff
@@ -2535,12 +2536,12 @@ declare global {
         new (): HTMLAtomicAriaLiveElement;
     };
     /**
-     * @beta - This component is part of the automatic facets feature.
-     * Automatic facets are currently in beta testing and should be available soon.
-     * An automatic facet is a special type of facet generated by the automatic facets feature. Unlike traditional facets that
-     * need to be explicitly defined and requested in the search query, automatic facets are dynamically generated by the index
-     * in response to the search query. It's important to note that this component should not be used directly. It is internally used
-     * by the `atomic-automatic-facet-generator` component to automatically render updated facets.
+     * An automatic facet is a special type of facet generated by the automatic facets feature.
```
dmbrooke, line 2539: Whoops, just realized now that I'm commenting in the generated code 🤦 Either way, be sure to apply my suggestions (if you agree) universally in their respective places as it appears that it was re-used throughout 👍

dmbrooke (approved):
Didn't realize until near the end that I was commenting on the wrong file, but I have some docs suggestions. Looks go though!

PR #3226, opened 2023-09-27
WIP

olamothe:
https://coveord.atlassian.net/browse/KIT-2728


louis-bompart:
beware, github actions resolves the refs to a commit when the workflow starts.
So if you have jobs in series, and that job A do a commit, job B won't pick it up.

Moreover, the checks won't be applied to the commit created by job A, even if job B do pick it up.

PR #3227, opened 2023-09-27
test: rewrite smart-snippets.cypress with best practices

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2780

This PR improves the smart-snippets tests while doing 3 things :
- Combine many it() statements together for best practices (https://docs.cypress.io/guides/references/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion)
- Put the validateProps inside of the wrapper component (The component was not crashing properly)
- rounded precise pixel assertations
- Also made the wrapper component 'internal', it was not supposed to be exposed in the beginning.


packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.tsx
```diff
@@ -25,6 +25,21 @@ export class AtomicSmartSnippetCollapseWrapper {
 
   private shouldRenderButton = !!this.maximumHeight;
 
+  private validateProps() {
```
y-lakhdar, line 28: Sorry, I'm not sure to understand why this validation is now on the atomic-smart-snippet-collapse-wrapper. There is not `snippetMaximumHeight` nor `snippetCollapsedHeight` property on that clas

alexprudhomme: The `snippetMaximumHeight` is added by the user in the atomic-smart-snippet component. Then inside of it it is sent to the atomic-smart-snippet-collapse-wrapper. [link](https://github.com/coveo/ui-kit/blob/KIT-2780/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet/smart-snippet-common.tsx#L157-L159)

Inside of the collapse wrapper component the prop was simply called maximumHeight instead of snippetMaximumHeight. I just changed the prop name for less confusion and also I think this component is supposed to be @internal . I asked the author about it.

I changed the location of the validation since the error was not functioning correctly (probably because of the atomicSmartSnippetCommon pattern. I followed the same pattern used to validate maximumHeight and collapsedHeight. [link](https://github.com/coveo/ui-kit/blob/KIT-2780/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet/smart-snippet-common.tsx#L95)

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
```diff
@@ -254,33 +251,28 @@ describe('Smart Snippet Test Suites', () => {
...
 
-    describe('then pressing show more', () => {
-      beforeEach(() => {
-        SmartSnippetSelectors.showMoreButton().click();
-      });
+    CommonAssertions.assertAccessibility(smartSnippetComponent);
 
+    it('should work correctly when pressing show more and show less', () => {
+      SmartSnippetSelectors.showMoreButton().click();
+      cy.wait(200);
```
btaillon, line 264: We should avoid using `cy.wait` unless we're asserting that something doesn't happen IMO. `cy.wait` can cause flakiness and slows down tests.

Instead, we should leverage Cypress' retry-ability mechanism ([docs](https://docs.cypress.io/guides/core-concepts/retry-ability)). In other words, the next assertion should be automatically repeated until it passes up to a certain timeout. This may already be the case.

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
```diff
@@ -34,24 +34,18 @@ describe('Smart Snippet Test Suites', () => {
...
       );
-    });
-
-    it('should fallback to a div for the question', () => {
+      // should fallback to a div for the question
       SmartSnippetSelectors.question().should('have.prop', 'tagName', 'DIV');
-    });
-
-    it('render the correct question', () => {
+      //render the correct question
```
btaillon, line 46:
```suggestion
      // render the correct question
```

packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.tsx
```diff
@@ -13,17 +16,33 @@ export class AtomicSmartSnippetCollapseWrapper {
   public error!: Error;
   @Element() public host!: HTMLElement;
 
-  @Prop({reflect: true}) public maximumHeight?: number;
+  @Prop({reflect: true}) public snippetMaximumHeight?: number;
 
-  @Prop({reflect: true}) public collapsedHeight?: number;
+  @Prop({reflect: true}) public snippetCollapsedHeight?: number;
```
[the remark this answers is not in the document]
alexprudhomme: Right, because of yassine's comment. I thought it would be best to keep the same name across the two components but you are right. This doesnt make much sense. I will change it back. thanks.

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
alexprudhomme: The reason, I put cy.wait here is that after the showMoreButton().click() there is an animation and the assert directly after asserts a precise pixel amount (`SmartSnippetAssertions.assertAnswerHeight(height);`)

louis-bompart: You should be able to wait for the exact pixel amount. If not we should relax or remove the test.

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
```diff
@@ -34,24 +34,18 @@ describe('Smart Snippet Test Suites', () => {
     beforeEach(() => {
       new TestFixture().with(addSmartSnippet()).init();
     });
-
-    it('should fallback to a div for the accessibility heading', () => {
+    it('should work correctly', () => {
+      // should fallback to a div for the accessibility heading
```
louis-bompart, line 38: Use cy.log

packages/atomic/cypress/e2e/smart-snippet-assertions.ts
```diff
@@ -4,84 +4,79 @@ import {should} from './common-assertions';
...
 export function assertCollapseWrapperHeight(expectedHeight: number) {
-  it(`the collapse wrapper should have a displayed height of ${expectedHeight}px`, () => {
-    SmartSnippetSelectors.collapseWrapper()
-      .invoke('height')
-      .should('equal', expectedHeight);
-  });
+  SmartSnippetSelectors.collapseWrapper()
+    .invoke('height')
+    .should('equal', expectedHeight);
 }
```
louis-bompart, line 31: Replace should('equal') by should('closeTo', expectedHeight, someDelta)
This will make more sense IMO.
We can go with a delta of .1 IMO

louis-bompart (changes requested):

packages/atomic/cypress/e2e/smart-snippet-assertions.ts
alexprudhomme: nice find.

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
btaillon: The problem here seems to be that `SmartSnippetAssertions.assertAnswerHeight(height);` isn't leveraging Cypress' retry-ability mechanism.

When I originally built this assertion, I recall making sure that it works with the retry-ability mechanism, meaning that, if the height didn't match, it would wait and try again up until Cypress' timeout.

However, it looks like something must've broken it somehow...

packages/atomic/cypress/e2e/common-assertions.ts
```diff
@@ -72,6 +72,16 @@ export function assertContainsComponentError(
   });
 }
 
+export function assertContainsComponentErrorWithoutIt(
```
dmbrooke, line 75: I'm not sure I see the value of having these checks without the `it`... Would you mind elaborating on why this is necessary?

packages/atomic/cypress/e2e/smart-snippet-assertions.ts
```diff
@@ -56,32 +49,27 @@ export function assertAnswerBottomMargin(
     SmartSnippetSelectors.answer()
       .find(`.${lastElementClass}`)
       .distanceTo(SmartSnippetSelectors.footer)
-      .should('have.property', 'vertical', margin);
+      .should('have.property', 'vertical')
+      .and('be.closeTo', margin, 0.1);
   });
 }
 
 export function assertLikeButtonChecked(checked: boolean) {
```
dmbrooke, line 57: I feel like this is more appropriate:
```suggestion
export function assertLikeButtonCheck(checked: boolean) {
```

`Checked` implies that it's already checked (aka active), but this can test both checked and unchecked 🙂

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
```diff
@@ -114,8 +107,13 @@ describe('Smart Snippet Test Suites', () => {
         .init();
     });
 
-    CommonAssertions.assertConsoleError(true);
-    CommonAssertions.assertContainsComponentError(SmartSnippetSelectors, true);
+    it('should display errors', () => {
+      CommonAssertions.assertConsoleErrorWithoutIt(true);
```
dmbrooke, line 111: Don't see why these need to be within an `it` and have the assertions be without the `it`. I don't see anything about this pattern in the best practices.

dmbrooke: Same comment goes for a bunch of different places in this file hehe

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
```diff
@@ -254,33 +251,34 @@ describe('Smart Snippet Test Suites', () => {
...
-        beforeEach(() => {
-          SmartSnippetSelectors.showLessButton().click();
-        });
-
-        SmartSnippetAssertions.assertShowMore(true);
-        SmartSnippetAssertions.assertShowLess(false);
-        SmartSnippetAssertions.assertCollapseWrapperHeight(heightWhenCollapsed);
-        CommonAssertions.assertAccessibility(smartSnippetComponent);
-      });
+      SmartSnippetSelectors.showLessButton().click();
```
dmbrooke, line 271: Even though I see why we don't want to nest a bunch of `describe`s, but perhaps we should still make it clear as to where each "section" is with by adding some comments.

dmbrooke (approved):
Have a few questions/suggestions, but it looks like there are just a few more tests to fix up and it will be ready to go :shipit:

PR #3228, opened 2023-09-27
facet(commerce): implement product-listing-pagination controller

fbeaudoincoveo:


packages/headless/src/controllers/commerce/product-listing/pagination/headless-product-listing-pagination.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    const mockState = buildMockCommerceState();
+    engine = buildMockCommerceEngine();
+    engine.state = {
+      ...mockState,
+      productListing: {
+        ...mockState.productListing,
+        pagination: {
+          ...mockState.productListing.pagination,
+          page: page,
+          totalPages: totalPages,
```
Spuffynism, line 31:
```suggestion
          page,
          totalPages,
```

packages/headless/src/controllers/commerce/product-listing/pagination/headless-product-listing-pagination.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  ProductListingPagination,
+} from './headless-product-listing-pagination';
+
+describe('ProductListingPagination', () => {
+  let engine: MockCommerceEngine;
+  let productListingPagination: ProductListingPagination;
+
+  function initProductListingPagination(page = 0, totalPages = 1) {
+    const mockState = buildMockCommerceState();
+    engine = buildMockCommerceEngine();
```
Spuffynism, line 23: We can pass a pre-populated state to the engine like this:
```suggestion
    engine = buildMockCommerceEngine({ state: {...} });
```
(instead of setting it as engine property)

packages/headless/src/commerce.index.ts
```diff
@@ -41,14 +41,8 @@ export type {
...
 
-export type {
-  PagerInitialState,
-  PagerOptions,
-  PagerProps,
-  Pager,
-  PagerState,
-} from './controllers/commerce/product-listing/pager/headless-product-listing-pager';
-export {buildPager} from './controllers/commerce/product-listing/pager/headless-product-listing-pager';
+export type {ProductListingPagination} from './controllers/commerce/product-listing/pagination/headless-product-listing-pagination';
```
Spuffynism, line 44: Why the rename from pager to pagination? Other pagination controllers use the term "pager", from my understanding.

packages/headless/src/controllers/commerce/product-listing/pagination/headless-product-listing-pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+        );
+      }
+    },
+  };
+
+  function loadProductListingPaginationReducers(
+    engine: CommerceEngine
+  ): engine is CommerceEngine {
+    engine.addReducers({
+      configuration,
```
Spuffynism, line 121: Why do we need the configuration reducer? :eyes:

My understanding is that the pagination reducer requires only the product listing one, since we don't depend on state set through the configuration reducer.

packages/headless/src/features/commerce/pagination/pagination-slice.ts
```diff
@@ -0,0 +1,19 @@
+import {createReducer} from '@reduxjs/toolkit';
+import {nextPage, previousPage, selectPage} from './pagination-actions';
+import {getPaginationInitialState} from './pagination-state';
+
+export const paginationReducer = createReducer(
+  getPaginationInitialState(),
+  (builder) => {
+    builder
+      .addCase(nextPage, (state) => {
+        state.page += 1;
```
Spuffynism, line 10:
```suggestion
        state.page++;
```

packages/headless/src/features/commerce/pagination/pagination-slice.ts
```diff
@@ -0,0 +1,19 @@
...
+      })
+      .addCase(previousPage, (state) => {
+        state.page -= 1;
```
Spuffynism, line 13:
```suggestion
        state.page--;
```

packages/headless/src/state/commerce-app-state.ts
```diff
@@ -1,23 +1,9 @@
...
-  NumericFacetSection,
-  PaginationSection,
   ProductListingV2Section,
-  StructuredSortSection,
   VersionSection,
 } from './state-sections';
 
 export type CommerceAppState = ConfigurationSection &
   ProductListingV2Section &
-  FacetSection &
```
Spuffynism, line 16: 💯 Thanks for cleaning up!

packages/headless/src/controllers/commerce/product-listing/pagination/headless-product-listing-pagination.ts
```diff
@@ -0,0 +1,127 @@
...
+    return getState().page > 0;
+  };
+
+  return {
+    ...controller,
+
+    get state() {
+      return {
+        ...getState(),
+      };
```
Spuffynism, line 82:
```suggestion
      return getState();
```

Spuffynism (approved):

packages/headless/src/commerce.index.ts
fbeaudoincoveo: I'm still hesitating on the controller name, TBH, but here's my reasoning:

I decided to go with "pagination" because that's the name of the property in the product listing API.

Also, I made the name more specific (i.e., productListingPagination) essentially to help with search relevance. Since this pagination controller works completely differently from the pager controllers we have on other engines, I don't want people to be confused if it surfaces in search requests on docs.coveo.com, for instance.

packages/headless/src/features/commerce/pagination/pagination-slice.ts
```diff
@@ -0,0 +1,19 @@
+import {createReducer} from '@reduxjs/toolkit';
```
fbeaudoincoveo, line 1: Oops I realize I forgot to write the pagination-slice tests.

I'll do that

fbeaudoincoveo: (done)

Spuffynism (approved):

packages/headless/src/controllers/commerce/product-listing/pagination/headless-product-listing-pagination.ts
```diff
@@ -0,0 +1,132 @@
...
+    get state() {
+      return getState();
+    },
+
+    selectPage(page: number) {
+      if (pageExists(page)) {
+        dispatch(selectPage(page));
+        dispatch(fetchProductListing());
+      } else {
+        engine.logger.warn(
```
olamothe, line 96: These checks in `selectPage/nextPage`/previousPage` should be done in the reducer.

olamothe (changes requested):

packages/headless/src/controllers/commerce/product-listing/pagination/headless-product-listing-pagination.ts
fbeaudoincoveo: Fixed!

packages/headless/src/api/commerce/product-listings/v2/pagination.ts
```diff
@@ -0,0 +1,6 @@
+export interface Pagination {
+  page: number;
```
dmbrooke, line 2: Is it really necessary to specify a page in the response (e.g., returning the same information back from the request)? Same question for `perPage`.

Unless this type is used for both the request _**and**_ response (which I'm not sure I agree with), I'm not convinced these properties need to be here 🙂

dmbrooke (approved):
LGTM

packages/headless/src/api/commerce/product-listings/v2/pagination.ts
fbeaudoincoveo: Because of the way the index works, It's possible that the page that was requested from the API is actually not available. In such a case, the API would presumably return a fallback page instead (the last available page before the requested page, maybe), and we would update the state with this.

olamothe (approved):

PR #3237, opened 2023-10-02
test(atomic): more simplification for smart snippet tests

olamothe:
In general: Remove lots of `describe`/`beforeEach` block (perform the setup directly in the it block when possible) and perform more assertion as part of a single `it` block.

https://coveord.atlassian.net/browse/KIT-2790

### DIFF
--- a/packages/atomic/cypress/e2e/smart-snippet.cypress.ts
+++ b/packages/atomic/cypress/e2e/smart-snippet.cypress.ts
@@ -30,458 +30,386 @@ function buildAnswerWithHeight(height: number) {
 }
 
 describe('Smart Snippet Test Suites', () => {
-  describe('with no heading level', () => {
-    beforeEach(() => {
-      new TestFixture().with(addSmartSnippet()).init();
-    });
-    it('should work correctly', () => {
-      cy.log('should fallback to a div for the accessibility heading');
-      SmartSnippetSelectors.accessibilityHeading().should(
-        'have.prop',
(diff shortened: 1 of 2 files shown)


packages/atomic/cypress/e2e/smart-snippet.cypress.ts
```diff
@@ -30,458 +30,386 @@ function buildAnswerWithHeight(height: number) {
...
+    SmartSnippetAssertions.assertShowLess(true);
+    SmartSnippetAssertions.assertAnswerHeight(height);
+    SmartSnippetSelectors.showLessButton().click();
+
+    // after collapse
+    SmartSnippetSelectors.body().should('not.have.attr', 'expanded');
+    SmartSnippetAssertions.assertShowMore(true);
+    SmartSnippetAssertions.assertShowLess(false);
+    SmartSnippetAssertions.assertAnswerHeight(heightWhenCollapsed);
   });
```
louis-bompart, line 201: I'm not sure about this part: I think there's value to test interaction 'separately'

louis-bompart (approved):
lgtm, tho might be too much "grouping"
I guess we'll live to see!

packages/atomic/cypress/e2e/smart-snippet.cypress.ts
olamothe: It execute much faster ! 🏃‍♂️

PR #3238, opened 2023-10-03
fix(headless): fix structuredClone for locker service

olamothe:
`structuredClone` is not supported by the "old locker service":" the one I had tested for was actually locker web security.

https://coveord.atlassian.net/browse/SFINT-5193


louis-bompart:
Locker Service and the likes: why we can't have nice things :')

louis-bompart (approved):

PR #3240, opened 2023-10-03
feat(commerce): support sorting v2 product listings

Spuffynism:
[CAPI-92]

[CAPI-92]: https://coveord.atlassian.net/browse/CAPI-92?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/product-listing/sort/headless-product-listing-sort.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  const schema = new Schema<SortInitialState>({
+    criterion: sortCriterionDefinition,
+  });
+
+  validateInitialState(engine, schema, state, 'buildSort');
+}
+
+export interface Sort extends Controller {
```
olamothe, line 65: So, at a quick glance, my initial reaction for this whole file, was that "there must be some overlap/more code reuse we could do with `headless-core-sort.ts`".

But when looking into it in more depth, I'd say in general what we have here is generally good 👍 

One thing we could want to make sure is that this interface here (or the interface for `SortProps`) does not differ too much than what we have for classic search use case for no good reason.

There could be a way to use generics to make sure that for both "search API sort" and "commerce service sort" controller, we have the same `sortBy`, or `isSortedBy` method, but that what they accept as parameter are different.

olamothe (approved):

pvanool (approved):

PR #3241, opened 2023-10-03
refactor(headless): combined SSR *WithProps and *WithoutProps interfaces

btaillon-coveo:
https://coveord.atlassian.net/browse/KIT-2729

These interfaces were difficult to read and modify. While working on KIT-2729, I happened to need to merge them, so I decided to open this as its own PR.

Essentially, I removed some interfaces like `HydrateInitialStateWithProps`/`HydrateInitialStateWithoutProps` and added instead `HydrateInitialState`.

I introduced a new internal utility type called `OptionTuple<T extends {}>` which converts an object to an option parameter, and makes the parameter optional if all keys are optional, or removes the parameter when there are no keys.


packages/headless/src/app/ssr-engine/types/build.ts
```diff
@@ -26,23 +21,10 @@ export interface BuildWithProps<
   /**
    * Initializes an engine and controllers from the definition.
    */
-  build(
```
btaillon, line 29: I extracted method names out of their objects, which will be useful in the next PR. I think it also looks kinda neat.

mrrajamanickam-coveo (approved):
Nice!

alexprudhomme (approved):
Looks good

PR #3243, opened 2023-10-03
chore: upgrade i18n compatibility json version

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2383

I did some investigation as to whether we could have i18next support both json compatibility v3 _**at the same time as**_ v4 (i.e., use whichever version their respective keys are present for), but this doesn't seem to be possible. Thus, I'm making the transition to the new version by making the following changes:

1. Removed the `compatibilityJSON` param from the i18n init call such that it defaults to v4
2. [Pollyfill](https://github.com/eemeli/intl-pluralrules) for [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) so we can guarantee that the `Intl.PluralRules` namespace will exist and i18n _**won't fallback to v3**_ in certain environments.
3. Mass find/replace `locales.json` to use the [v4 syntax](https://www.i18next.com/translation-function/plurals#singular-plural). v4 allows you to specify many different versions of plurals (using `_one`, `_two`, etc.,), but we only specify a single plural, so I simply replaced all instances of `_plural` with `_other` to achieve the same result.

I will mark this as a breaking change because [custom plural localization](https://docs.coveo.com/en/atomic/latest/usage/atomic-localization/) will no longer with the old syntax.

BREAKING CHANGE: plural locales are now in v4 json compatibility format

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26777,6 +26777,11 @@
         "node": ">= 0.10"
       }
     },
+    "node_modules/intl-pluralrules": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/intl-pluralrules/-/intl-pluralrules-2.0.1.tgz",
+      "integrity": "sha512-astxTLzIdXPeN0K9Rumi6LfMpm3rvNO0iJE+h/k8Kr/is+wPbRe4ikyDjlLr6VTh/mEfNv8RjN+gu3KwDiuhqg=="
+    },
     "node_modules/invariant": {
       "version": "2.2.4",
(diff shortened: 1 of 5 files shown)


olamothe (changes requested):
> I will mark this as a breaking change because [custom plural localization](https://docs.coveo.com/en/atomic/latest/usage/atomic-localization/) will no longer with the old syntax.

> BREAKING CHANGE: plural locales are now in v4 json compatibility format

 We cant merge this... Unless I misunderstand the intention here.

PR #3244, opened 2023-10-03
chore: run npm pkg fix

louis-bompart:
Source: https://github.com/coveo/ui-kit/actions/runs/6398635946/job/17369187911

Docs: https://docs.npmjs.com/cli/v10/commands/npm-pkg

https://coveord.atlassian.net/browse/KIT-2794

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,5 +116,5 @@
   "dependencies": {
     "upgrade": "1.1.0"
   },
-  "version": ""
+  "version": "0.0.0"
 }
--- a/package.json
+++ b/package.json
@@ -115,5 +115,6 @@
   },
   "dependencies": {
     "upgrade": "1.1.0"
-  }
+  },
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -115,5 +115,6 @@
   },
   "dependencies": {
     "upgrade": "1.1.0"
-  }
+  },
+  "version": ""
```
louis-bompart, line 119: whatever meh

mrrajamanickam-coveo: should we make it 0.0.0 for consistency and convention as internal pkg?

mrrajamanickam-coveo (approved):

package.json
louis-bompart: Mh yeah, why not.
```suggestion
  "version": "0.0.0"
```

PR #3248, opened 2023-10-04
test(atomic): simplify result-multi-value-text tests

olamothe:
The tests in this file were simplified in this manner: 

* Previously, was calling a `testWithValidFieldValue` function in a loop, which was itself creating multiple describe block, which were themselves containing more embedded describe blocks. This PR flatten all of that to max 1 sub level of describe block

* Lots of utility function to `prepare` the tests, which made it hard to see what we were trying to actually test/what options combination we are actually trying to setup on the component. Those utility function were often time called only in one test: Removed them, and simply create the fixture "verbosely/repetitively" in each individual test.

* The edge case for this component are: 
1- Nothing gets rendered (because of errors, or invalid field values)
2- All values gets rendered (because there's less or equal field values than `max-values-to-display`
3- Values gets truncated (because there's more field values than `max-values-to-display`

We were testing edge case 2 and 3 multiple time in an un-necessary manner. ie: there's not much value to test `fieldValues.length=3` and then (`max-value-to-display=1 || 2 || 3`). We can just test one of them.

https://coveord.atlassian.net/browse/KIT-2797


packages/atomic/cypress/e2e/common-assertions.ts
```diff
@@ -86,7 +86,7 @@ export function assertAccessibilityWithoutIt<T extends HTMLElement>(
   if (typeof component === 'string') {
     cy.checkA11y(component, axeOptions, logAxeIssues);
   } else if (typeof component === 'function') {
-    component().should(([el]) => {
```
olamothe, line 89: Was erroring

packages/atomic/cypress/fixtures/test-fixture.ts
```diff
@@ -243,24 +243,26 @@ export class TestFixture {
         return;
       }
 
-      searchInterfaceComponent.initialize(sampleConfig).then(() => {
```
olamothe, line 246: Semi related to the failures in multi-value-text: 

When there's retry, we can end up trying to initialize the search-interface multiple times (which Atomic will prevent, by throwing warnings/errors in the console).

We can prevent that by just checking the `hydrated` class (ie: what stencil generate when a component is rendered)

packages/atomic/cypress/e2e/result-list/result-components/result-multi-value-text.cypress.ts
```diff
@@ -49,260 +47,150 @@ export interface FacetWithResponseProps {
...
+        .with(addFieldValueInResponse(field, originalValues))
+        .withFieldCaptions(field, values)
+        .withTranslation({'n-more': '{{value}} more'})
+        .with((fixture: TestFixture) => {
+          fixture
+            .with(addFacet({'facet-id': 'blah', field}))
+            .withCustomResponse((response) => {
+              response.facets = [
+                {
+                  facetId,
```
alexprudhomme, line 174: I think you forgot to intialize this facetId variable. I am getting errors in cypress and vscode

olamothe:
Whelp, there was a test failure still on this file, more work to do...

packages/atomic/cypress/e2e/result-list/result-components/result-multi-value-text.cypress.ts
```diff
@@ -49,260 +47,151 @@ export interface FacetWithResponseProps {
...
 
-          assertShouldRenderValues(localizedValues.slice(0, 1));
-          assertDisplaysXMoreLabel(2);
-        });
+    it('when there is a slot it should replace the correct values', () => {
+      const slotElement = generateComponentHTML('span', {
+        slot: `result-multi-value-text-value-${originalValues[1]}`,
       });
+      slotElement.innerText =
+        'The field value in at position 1 should be replaced by this';
```
mrrajamanickam-coveo, line 131:
```suggestion
        'The field value at position 1 should be replaced by this';
```

packages/atomic/cypress/e2e/result-list/result-components/result-multi-value-text.cypress.ts
```diff
@@ -49,260 +47,151 @@ export interface FacetWithResponseProps {
...
 
-            beforeEach(() => {
-              prepareValuesWithMaximum(values, 4, [
-                buildSlotElement(rawValues[1], 'A'),
-                buildSlotElement(rawValues[3], 'B'),
-              ]).init();
-            });
+      assertShouldRenderValues([
+        localizedValues[0],
+        'The field value in at position 1 should be replaced by this',
```
mrrajamanickam-coveo, line 150:
```suggestion
        'The field value at position 1 should be replaced by this',
```

mrrajamanickam-coveo (approved):

alexprudhomme (approved):

PR #3249, opened 2023-10-04
chore(release): fix uv_os_get_passwd in jenkins

louis-bompart:
Similar to https://github.com/microsoft/playwright/pull/22826.
Didn't fish out the exact line where SFDX goes haywire, but this is not the first time docker & this command clashes.
It's unlikely for the CLI to fix it, as this is working as intended probably (they can't assume basic system API won't work..)

So it's our job to make it work ™️. 
For that, we need the docker container to have `/etc/passwd` defined.
The easiest way is to map it to the host.
Another approach would be to investigate what's needed to spoof it enough in the Docker.

[SFINT-5221](https://coveord.atlassian.net/browse/SFINT-5221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

[SFINT-5221]: https://coveord.atlassian.net/browse/SFINT-5221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau (approved):

dmbrooke (approved):

erocheleau:
This should fix SFINT-5221 as well

btaillon-coveo (approved):
I wonder what's in the `/etc/passwd` of the runner 🤔 I would feel a tad safer if we gave it an empty file instead or something of that sort.

If you feel confident that there's nothing exploitable in this passwd file, then feel free to merge.

PR #3250, opened 2023-10-05
test: simplify search box instant result

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2759

I had already worked on this file 1 month ago, so this did not take a long time.

 I was able to simplify these tests because the main part was looking if you were able to navigate through it using the keyboard and the mouse. These 2 do not need a reinitialization of the component. You can keep the same one and go through every use case. This way, it creates less reloading resulting in less flakiness.  
 
 I also took the time to merge the tests in 'with a custom aria label' for the same reasons. The component did not need to be reinitialized.

### Before
image

### After 
image

### DIFF
--- a/packages/atomic/cypress/e2e/common-assertions.ts
+++ b/packages/atomic/cypress/e2e/common-assertions.ts
@@ -172,6 +172,9 @@ export function assertRemovesComponent() {
   );
 }
 
+/**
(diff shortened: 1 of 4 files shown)


olamothe (approved):
💯 Nice job

PR #3251, opened 2023-10-05
feat: double publish the NPM packages into GitHub packages for use by internal repositories also using the @coveo scope

malaporte:
This PR depends on this other one to be merged and published first: https://github.com/coveo/semantic-monorepo-tools/pull/189

Some of our repos use internal NPM packages published on GitHub Packages under the @coveo scope. For these we need to override the registry for that scope to the GitHub one, but this breaks when loading @coveo-scoped public packages from NPM. Since we can't change the scope for GitHub (it has to be the same as the org), there is not much we can do easily ... besides double publishing our few public packages. So here goes.

https://coveord.atlassian.net/browse/KIT-2806


louis-bompart:
lgtm, tho don't forget to update semantic-monorepo-tools

btaillon-coveo:
Unfortunately, I'm not sure this would publish successfully, since our workflows set `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}`.

It looks like something more may be necessary to allow authentication to GitHub Packages for packages published to GitHub Packages. Haven't looked too deep into it, but appending the `secrets.GITHUB_TOKEN` to `.npmrc` only for the GitHub registry (similar to what they do in https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-npm-registry#authenticating-with-a-personal-access-token) may be a good solution.

louis-bompart:
Good catch @btaillon-coveo, but I think we should follow https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm
So instead, in release.yml, let's add a setup-node w/ the git registry and github token. just have to run phase 2 then and gg .

malaporte:
> Good catch @btaillon-coveo, but I think we should follow https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm So instead, in release.yml, let's add a setup-node w/ the git registry and github token. just have to run phase 2 then and gg .
> 
> Won't even need to modify the JS files.

Sounds pretty neat! I've made some changes around this, but tbh I'm not sure what is the kosher way to test it. Should I try to publish a dummy release?

btaillon-coveo:
> > Good catch @btaillon-coveo, but I think we should follow [actions/setup-node@`main`/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md?rgh-link-date=2023-10-06T12%3A32%3A54Z#publish-to-npmjs-and-gpr-with-npm) So instead, in release.yml, let's add a setup-node w/ the git registry and github token. just have to run phase 2 then and gg .
> > Won't even need to modify the JS files.
> 
> Sounds pretty neat! I've made some changes around this, but tbh I'm not sure what is the kosher way to test it. Should I try to publish a dummy release?

You could temporarily bring back the test setup that I removed here: https://github.com/coveo/ui-kit/pull/2924/

This setup was running Verdaccio as a local registry for testing, and was updating `.npmrc` to set the registry.

You could run yours in `prbot` instead of `masterbot`, and you could move the registry config elsewhere than `npmrc`.

.github/workflows/release.yml
```diff
@@ -45,3 +43,13 @@ jobs:
...
           RELEASER_INSTALLATION_ID: ${{ secrets.RELEASER_INSTALLATION_ID }}
           DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - uses: actions/setup-node@v3
+        with:
+          registry-url: 'https://npm.pkg.github.com'
+          node-version-file: '.nvmrc'
+      - name: Publish to GitHub Packages
+        run: npm run release:phase2
```
btaillon-coveo, line 53: I'm not sure whether this will work, since `release:phase2` also depends on `release:phase1` which bumps packages and `release:build` which rebuilds the package.

This *might* be fine because the bump script has this check at the beginning:
```json
const lastTag = await getLastTag(versionPrefix);
const commits = await getCommits(PATH, lastTag);
if (commits.length === 0 && !hasPackageJsonChanged(PATH)) {
  return;
}
```

`commits.length` should already be zero since we just did a release, therefore the new commit should be tagged, and `hasPackageJsonChanged` should return `false` since we're still in the same job as the publish, therefore our Git history should be on the new commit.

However, a safer approach may be to add a "phase5" which re-publishes. I opened a PR here that you can integrate into your PR: https://github.com/coveo/ui-kit/pull/3263

btaillon-coveo (approved):

louis-bompart (approved):

dmbrooke (approved):

malaporte:
Yolo'ing this to test in production :brace:

PR #3252, opened 2023-10-05
chore(atomic): support i18n json compatibility v4

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2383

Change is a lot smaller than it appears, but a majority of the changes come from the duplicate entries in locales.json to appease the new json plural format.

Changes:
* ~Add [polyfill](https://github.com/eemeli/intl-pluralrules) for [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) to ensure v4 json compatibility works should they chose to upgrade.~ Never mind, Intl.PluralRules is supported across all of the same browsers we support. This polyfill won't be necessary.
* Add atomic-search-interface prop for implementers to specify which version to use (v3 by default)
* Log warning when v3 is detected upon changing languages
* UT to ensure correct translations are used for each version.


packages/atomic/src/components/common/interface/interface-common.tsx
```diff
@@ -26,6 +26,7 @@ export interface BaseAtomicInterface<EngineType extends AnyEngineType>
   host: HTMLStencilElement;
   bindings: AnyBindings;
   error?: Error;
+  compatibilityJson: 'v1' | 'v2' | 'v3' | 'v4';
```
alexprudhomme, line 29: Why is v1 and v2 here ? Should this be `i18nCompatibilityJSONVersion` ?

alexprudhomme (approved):
LGTM !

olamothe (changes requested):
I wonder if there's something we could do when we detect an entry override/entry getting added to the i18n dictionary that follows the old format (ie: warn) ?

packages/atomic/src/components/common/interface/interface-common.tsx
dmbrooke: Whoops, could've sworn I removed 'v1' | 'v2' at one point, thanks for pointing this out! 

Also I agree, that name is more clear. I'll use that instead.

dmbrooke:
> I wonder if there's something we could do when we detect an entry override/entry getting added to the i18n dictionary that follows the old format (ie: warn) ?

After doing some research, I'm not convinced that this could be done cleanly without either having to introduce a breaking change to `interface-common` or updating the localization docs...

Ideally, we would create our own class that extends the i18n library class, and then overload the one method we need (`addResourceBundle`) to add a warning before calling the original method. _**However**_, because i18n is actually an interface, I believe we would also need to duplicate all of the other methods that are called within `addResourceBundle`, creating a whole bunch of code duplication.

So for now, I think I'll keep it simple and log a warning whenever v3 is detected on the execution of the first search 🙂

olamothe (approved):

louis-bompart (changes requested):
Why polyfill?
API seems supported on all our supported browser, so I'd say nay to a polyfill?

dmbrooke:
> Why polyfill? API seems supported on all our supported browser, so I'd say nay to a polyfill?

Oh interesting, do we have an official list of supported browsers somewhere for Atomic?

I added the polyfill as a safety net (which is suggested in the [i18next migration guide](https://www.i18next.com/misc/migration-guide#v20.x.x-to-v21.0.0)) in case there was a context in which `PluralRules` didn't exist. Although now that I think about it, this is Atomic (not Headless), so we have a much clearer idea as to where this code will be executing, so you're probably right about not needing it 🙂

louis-bompart (approved):

jelmedini (approved):

PR #3253, opened 2023-10-05
test: show console errors inside of cypress

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2807

Whenever a component would crash inside of cypress, it would not print in the cypress console. I think this is a feature that broke when updating cypress. I had this problem and felt like it wasn't there before. Then I remembered that @mrrajamanickam-coveo had done this change in his old PR to update to cypress 12. 

image


louis-bompart (approved):
Good idea!
I'm thinking we might want to add special task (like what we did for axe-core recently) to log those kind of warning & error, be it the one in the console or the axe-core stuff.
But that could be tackled later

PR #3254, opened 2023-10-05
fix: screw it, max-height transition !

pathiery:
Basically doing the same thing as before, except
- use `--full-height` to calculate a `max-height` instead of `height`
- set the `transition` on `max-height` instead of `height`
- set `height` to `auto`

No visual changes except making the expanded height more consistent.

TBH I think that at this point we could just not calculate `--full-height` and use a static value of like 1 Billion pixels, but I kinda want to get my feature working (it has been a while 😬)

https://github.com/coveo/ui-kit/assets/71142397/5574c9ba-2e04-4170-9214-6821785c3fda

Discussed with @btaillon-coveo and we could probably remove code that use this `--full-height` related logic in both smart-snippet components (`collapse-wrapper` and `expandable-answer`) and use `height: auto` everywhere... TBD

### DIFF
--- a/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
+++ b/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
@@ -8,7 +8,8 @@
   &::part(smart-snippet-collapse-wrapper) {
     @mixin set-font-size var(--atomic-text-lg);
 
-    height: var(--collapsed-size);
+    height: auto;
+    max-height: var(--collapsed-size);
 
     --gradient-start: var(
       --atomic-smart-snippet-gradient-start,
@@ -17,7 +18,7 @@
     color: var(--atomic-on-background);
     mask-image: linear-gradient(black, black var(--gradient-start), transparent 100%);
 
-    transition: height ease-in-out 0.25s;
+    transition: max-height ease-in-out 0.25s;
   }
 
   button atomic-icon {
@@ -27,7 +28,8 @@
 
 :host(.expanded) {
   &::part(smart-snippet-collapse-wrapper) {
-    height: var(--full-height);
+    height: auto;
+    max-height: calc(var(--full-height) + 200px);
     mask-image: none;
   }
 
--- a/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
+++ b/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
@@ -18,7 +18,7 @@
     color: var(--atomic-on-background);
     mask-image: linear-gradient(black, black var(--gradient-start), transparent 100%);
 
-    transition: max-height cubic-bezier(0, 1, 0.22, 1) calc(var(--full-height-value) * 0.001s);
+    transition: max-height 2s cubic-bezier(0, 1, 0.16, 1) -1.82s;
   }
 
   button atomic-icon {
@@ -29,9 +29,9 @@
 :host(.expanded) {
   &::part(smart-snippet-collapse-wrapper) {
     height: auto;
-    max-height: calc(var(--full-height) + 200px);
+    max-height: 9999999px;
+    transition: max-height 2s cubic-bezier(1, 0, 1, 0) 0s;
     mask-image: none;
-    transition: max-height ease-in-out calc(var(--full-height-value) * 0.001s);
   }
 
   button atomic-icon {
--- a/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
+++ b/packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
@@ -18,7 +18,7 @@
     color: var(--atomic-on-background);
     mask-image: linear-gradient(black, black var(--gradient-start), transparent 100%);
 
-    transition: max-height ease-in-out 0.25s;
+    transition: max-height cubic-bezier(0, 1, 0.22, 1) calc(var(--full-height-value) * 0.001s);
   }
 
   button atomic-icon {
@@ -31,6 +31,7 @@
     height: auto;
     max-height: calc(var(--full-height) + 200px);
     mask-image: none;
+    transition: max-height ease-in-out calc(var(--full-height-value) * 0.001s);
   }
 
   button atomic-icon {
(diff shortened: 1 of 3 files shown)


packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
```diff
@@ -27,7 +28,8 @@
 
 :host(.expanded) {
   &::part(smart-snippet-collapse-wrapper) {
-    height: var(--full-height);
+    height: auto;
+    max-height: calc(var(--full-height) + 200px);
```
btaillon-coveo, line 32: what happens when `max-height` isn't present here? Does it break the transition or does it still work?

AFAIK, setting a huge value like you said would break the transition, since it would lerp between the smaller value and the huge value, causing a delay before the animation starts showing visualy.

Also, would it be possible to remove this `+ 200px`? same reason as I mentioned above with the huge value.

mikegron (approved):
I was there and saw it working perfectly 🙆‍♂️

packages/atomic/src/components/common/smart-snippets/atomic-smart-snippet-collapse-wrapper/atomic-smart-snippet-collapse-wrapper.pcss
pathiery: Hey Benjamin ! Did a little update that mostly makes the animation more consistent when collapsing the smart snippet. Also added a gif so that you can see it in action.
Tested with a static value of `--full-height` set to 1000 and it seemed to work really well as well.
I'm keeping the `+200px` as it alleviates cases where `.getBoundingClientRect()` doesn't give the correct size. 
Let me know what you think of the animation in the gif above !

btaillon-coveo: > mostly makes the animation more consistent when collapsing the smart snippet.

What was the inconsistency before?

> Tested with a static value of `--full-height` set to 1000 and it seemed to work really well as well.

Did you try with a huge value like you said, like `999999999px`? What happens?

> I'm keeping the `+200px` as it alleviates cases where `.getBoundingClientRect()` doesn't give the correct size.

What's the reason why it doesn't give the correct size? Are we certain its error will never exceed 200px? Otherwise, this will cut off the snippet and will most likely end up as a bug report.

btaillon-coveo: also, is there a delay between clicking to expand/collapse and the animation starting, if you wait like 2 seconds between pressing "show more", then "show less", then "collapse"?

pathiery: >What was the inconsistency before?

When the max height was larger, the animation was jumpy

> Did you try with a huge value like you said, like 999999999px? What happens?

I've tried with a value of `999999px` and I tweaked the transitions a bit to be 
```
:host(.expanded) {
  &::part(smart-snippet-collapse-wrapper) {
   /* ... */
    transition: max-height 2s cubic-bezier(0, 1, 0.22, 1) -1.75s;
  }
}

:host(.expanded) {
  &::part(smart-snippet-collapse-wrapper) {
    /* ... */
    transition: max-height 2s cubic-bezier(1, 0, 1, 0) 0.01s;
    /* ... */
  }
```
and I got the following animation 

https://github.com/coveo/ui-kit/assets/71142397/451959e8-cfc3-4c40-8cdf-dfa099265f2a

In my humble opinion this is probably good enough for 99% of cases ? I'll let you be the judge of that.

> What's the reason why it doesn't give the correct size? Are we certain its error will never exceed 200px? Otherwise, this will cut off the snippet and will most likely end up as a bug report.

As we discussed previously, the `getBoundingRect` acts up sometimes, I'm guessing that the inner slot is not playing nice with it and about half of the time I get a size that is slightly smaller that the actual height.

> also, is there a delay between clicking to expand/collapse and the animation starting, if you wait like 2 seconds between pressing "show more", then "show less", then "collapse"?

As you can see in my gif above, no delay ! 👍

olamothe (approved):
I'll approve but:

* I would go with a really huge value instead of `+200` (ie: 9999999px). It at least has the advantage of being obvious that it's "hackish", and so anyone coming to change this in X years should have an easier time understanding why its there. In any case, I would suggest adding some comment that explains the problem/decision.

* I played with the IPX locally with smart snippet in this PR, I must say it does not look very good UX wise. The root problem seems to come from the fact that IPX is ~700px high by default, and we try to cram a lot of feature in there. I would really propose to revisit that, because I don't think simply having a smart snippet collapse feature solves anything/makes the interface more usable at the end of the day, when you compare to the mockups. I had a small discussion with @johnnyCoveo about this. Making it something like 80-90% of available screen size in height would alleviate a lot of issues I think. If it's only 700px on localhost then disregard that comment.

* The current setting of putting `snippet-maximum-height="50" snippet-collapsed-height="50"` makes it so we only see the title of the snippet, and then we see some "shadow" of the top of each letter for the next line. Which is pretty weird. I guess it's an option/it's configurable, but I don't know if we want to make this the default or not. It's super important that the default are sane/good. There's a risk that this `50px` setting gets copy pasted everywhere and re-used for a lot of implementation. I think the intention behind the collapse is that we still want to show "the first few lines" so that the user knows there's more to read. From what I've seen in local, I don't think end user would know there's something meaningful to read in the snippet when you only see the title.

pathiery:
Thanks for the review !

> I'll approve but:
> 
> * I would go with a really huge value instead of `+200` (ie: 9999999px). It at least has the advantage of being obvious that it's "hackish", and so anyone coming to change this in X years should have an easier time understanding why its there. In any case, I would suggest adding some comment that explains the problem/decision.

Gotchu, I'll make it work with `9999999px` and remove the calculation part altogether.

> * I played with the IPX locally with smart snippet in this PR, I must say it does not look very good UX wise. The root problem seems to come from the fact that IPX is ~700px high by default, and we try to cram a lot of feature in there. I would really propose to revisit that, because I don't think simply having a smart snippet collapse feature solves anything/makes the interface more usable at the end of the day, when you compare to the mockups. I had a small discussion with @johnnyCoveo about this. Making it something like 80-90% of available screen size in height would alleviate a lot of issues I think. If it's only 700px on localhost then disregard that comment.

I have to agree, it's not perfect at all. It is currently 700px in all our test versions. However, this PR is part of a larger redesign feature for the IPX, which aims to move us closer to @johnnyCoveo 's vision of the IPX as a hub.
TLDR; this is part of a larger feature to make the IPX look nicer !

> * The current setting of putting `snippet-maximum-height="50" snippet-collapsed-height="50"` makes it so we only see the title of the snippet, and then we see some "shadow" of the top of each letter for the next line. Which is pretty weird. I guess it's an option/it's configurable, but I don't know if we want to make this the default or not. It's super important that the default are sane/good. There's a risk that this `50px` setting gets copy pasted everywhere and re-used for a lot of implementation. I think the intention behind the collapse is that we still want to show "the first few lines" so that the user knows there's more to read. From what I've seen in local, I don't think end user would know there's something meaningful to read in the snippet when you only see the title.

Gotchu, I'll make the default larger !

PR #3256, opened 2023-10-06
test: fix ssr-search-parameter-manager e2e tests fails

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2782

The tests were failing both in dev and prod locally.

### DIFF
--- a/packages/samples/headless-ssr/app-router/cypress/e2e/ssr-search-parameter-manager.cy.ts
+++ b/packages/samples/headless-ssr/app-router/cypress/e2e/ssr-search-parameter-manager.cy.ts
@@ -93,7 +93,7 @@ describe('headless ssr with search parameter manager example', () => {
     }
 
     it('renders page in SSR as expected', () => {
-      cy.intercept('/**', (req) => {
+      cy.intercept('/generic/**', (req) => {
         req.continue((resp) => {
           const dom = new DOMParser().parseFromString(resp.body, 'text/html');
           expect(dom.querySelector('.search-box input')).to.have.value(query);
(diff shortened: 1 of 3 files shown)


packages/samples/headless-ssr/app-router/cypress/e2e/ssr-search-parameter-manager.cy.ts
```diff
@@ -96,7 +96,10 @@ describe('headless ssr with search parameter manager example', () => {
       cy.intercept('/**', (req) => {
         req.continue((resp) => {
           const dom = new DOMParser().parseFromString(resp.body, 'text/html');
-          expect(dom.querySelector('.search-box input')).to.have.value(query);
+          const searchBoxInput = dom.querySelector('.search-box input');
+          if (searchBoxInput) {
+            expect(searchBoxInput).to.have.value('def');
+          }
         });
       });
```
alexprudhomme, line 104: Simply adding a null check solved the flakiness issue. I tested with values that should fail and it works fine.

olamothe (commented):
If the test is now correctly working/no flakiness, then we should mark it as blocking for a valid build.

This is done in `.github/worflows/prbot.yml` , making it a "required-job"

mrrajamanickam-coveo:
> If the test is now correctly working/no flakiness, then we should mark it as blocking for a valid build.
> 
> This is done in `.github/worflows/prbot.yml` , making it a "required-job"

The test works without flakiness in CI as it runs against next.js in prod mode. It fails only when next.js is run in dev mode.

mrrajamanickam-coveo (approved):

PR #3263, opened 2023-10-09
add release:phase5

btaillon-coveo:
Opened for comment: https://github.com/coveo/ui-kit/pull/3251/files#r1350550715


packages/atomic-angular/project.json
```diff
@@ -25,14 +25,6 @@
         "command": "npx -p=@coveo/release bump",
         "cwd": "packages/atomic-angular/projects/atomic-angular"
       }
-    },
-    "release:phase2": {
```
btaillon-coveo, line 29: redundant

louis-bompart:
todo add desc to release doc thingy

btaillon-coveo:
> Thanks, that seems to make sense.
> 
> > todo add desc to release doc thingy
> 
> you mean `release-process.md`? It's already in there...?

We have a Confluence page which describes the kinds of releases we do and where they can be downloaded from, this may be what Louis means, I think it'd make sense to update it.

louis-bompart:
> > Thanks, that seems to make sense.
> > > todo add desc to release doc thingy
> > 
> > 
> > you mean `release-process.md`? It's already in there...?
> 
> We have a Confluence page which describes the kinds of releases we do and where they can be downloaded from, this may be what Louis means, I think it'd make sense to update it.

Yes I was a bit too lazy to look it up, but wanted to put a note at least so we don't forget. I meant the md first, but confluence makes sense too, that we know it's there in GitHub packages.

louis-bompart:
Let's skip this for tmrw release. I want to have Quantic released and good, then we can introduce more trouble.

malaporte:
This is a PR onto another branch so it won't break the main build @louis-bompart :p

PR #3265, opened 2023-10-10
fix(headless): prioritize root result when filtering out duplicates in folded collection

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-2785


packages/headless/src/features/folding/folding-utils.ts
```diff
@@ -9,11 +9,18 @@ export function getAllIncludedResultsFrom(relevantResult: ResultWithFolding) {
     .map((result) => result.parentResult!);
 
   const resultsInCollection = removeDuplicates(
-    [relevantResult, ...foldedResults, ...parentResults],
+    [...foldedResults, ...parentResults],
```
olamothe, line 12: This is kind of like calling `removingDuplicates` two times in a row, but just by enforcing order.

I'm fairly certain this would have also worked (without the new code for `includedResults`):

```typescript
const resultsInCollection = removeDuplicates(
    [...foldedResults, ...parentResults, relevantResult],
    (result) => result.uniqueId
)
```

So either we can just change the order here (and rely on the new UT to make sure the implementation of `removeDuplicates` stays with a `reduce(Left)`, or we could change `removeDuplicates` to support a parameter to control the order of the array we want prioritized (reduce left or right)

olamothe (changes requested):

packages/headless/src/utils/utils.ts
```diff
@@ -17,16 +17,22 @@ export function isEmptyString(str: string) {
 
 export function removeDuplicates<T>(
```
olamothe, line 18: I don't think we already have a UT for this file, but we should add one.

And we should test the function for both left/right

packages/headless/src/features/folding/folding-utils.ts
```diff
@@ -9,8 +9,13 @@ export function getAllIncludedResultsFrom(relevantResult: ResultWithFolding) {
     .map((result) => result.parentResult!);
 
   const resultsInCollection = removeDuplicates(
-    [relevantResult, ...foldedResults, ...parentResults],
-    (result) => result.uniqueId
+    [
+      ...foldedResults,
+      ...parentResults,
+      {...relevantResult, parentResult: null},
```
olamothe, line 15: As discussed live, some questions remains regarding setting `parentResult` to null. Might be hiding some other bugs...

fbeaudoincoveo: I made the adjustement to return the "complete" relevantResult (i.e., not remove the parentResult from it), and also fixed the failing test (by specifying distinct uniqueId values for the parent and child in the mock test result).

Looking deeper into this, I can't find any reason for this to cause any issues.

y-lakhdar (approved):

alexprudhomme (approved):
new function looks good to me

packages/quantic/cypress/fixtures/folded-result-list/exampleResultWithMoreChildResults.json
```diff
@@ -67,11 +70,11 @@
       "title": "Feather Stars",
       "children": [
         {"title": "Antarctic Feather Star"},
+        {"title": "Klunzinger's Feather Star"},
```
fbeaudoincoveo, line 73: The reason why the results were returned in that order before is that the array returned by the previous version of the `removeDuplicates` function was generated by running [`Object.values`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values) on an object.

`Object.values` iterates over non-negative integer keys in ascending order by value (in that regard, it behaves as a `for..in` loop, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#description). Since we're filtering out duplicates based on the `uniqueId`, and the `uniqueId` values in the example results are numbers, those numbers became keys in the `removeDuplicates` reduce function, and were then returned in ascending order by `Object.values`.

The new `removeDuplicates` method uses a `Map` (instead of `Record`) when calling `reduce`, and relies on [`Map.prototype.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values) for generating the return array, so that the initial order is preserved even if the keys are numbers.

olamothe (approved):

packages/quantic/cypress/fixtures/folded-result-list/exampleResultWithMoreChildResults.json
```diff
@@ -31,14 +31,15 @@
...
-              "foldingparent": "parent1",
+              "foldingchild": ["node2"],
+              "foldingparent": "node1",
               "urihash": "#4"
             }
           },
           "raw": {
             "foldingcollection": "Example Collection",
-            "foldingparent": "parent2",
+            "foldingchild": ["leaf2"],
```
erocheleau, line 41: @fbeaudoincoveo why do we need to add child to test results that didn't have any before?

Other than this, if the tests are green, go ahead :)

packages/quantic/README.md
```diff
@@ -31,6 +31,7 @@ Optionally install the [VSCode Salesforce Extension Pack](https://marketplace.vi
 ### Enable Dev Hub in Your Salesforce Organization
 
 - Connect to your salesforce developer organization [login](http://login.salesforce.com/)
+- Switch to Salesforces Classic.
```
erocheleau, line 34:
```suggestion

```

fbeaudoincoveo: I couldn't find the option to enable it in the setup in SF Lightning, though

packages/quantic/cypress/fixtures/folded-result-list/exampleResultWithMoreChildResults.json
fbeaudoincoveo: It changes absolutely nothing on the test results, but I feel like it doesn't really make sense to have folded results that don't have a foldingchild value. Also, identifying the folded results with "child" / "leaf" makes it clearer what the intended hierarchy of the example result is IMO.

One possible drawback of this is that we no longer have test cases (at least as far as I'm aware) where the `getChildField` function in `folding-slice.ts` would return `undefined`.

I can revert this change if you prefer, it will have no other impact.

packages/quantic/README.md
fbeaudoincoveo: Nevermind, I can see it now... I must have made a typo or something hehe

PR #3268, opened 2023-10-10
chore: isolate nvm, second attempt

louis-bompart:
#3266 didn't work because `nvm` is a packageManager for renovate, not a package.
Same same but different...

Was not able to validate locally with validator as it didn't flag the current config as ❌ 

KIT-282

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -81,17 +81,11 @@
       "groupSlug": "nx",
       "rangeStrategy": "replace",
       "allowedVersions": "15.x"
-    },
-    {
-      "groupName": "nvm",
-      "groupSlug": "NVM",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -81,17 +81,11 @@
...
-      "groupSlug": "NVM",
-      "fileMatch": [
-        "(^|/)\\.nvmrc$"
-      ],
-      "versioning": "node",
-      "pinDigests": false
     }
   ],
+  "nvm": {
+    "groupName": "NVM"
```
louis-bompart, line 87:
```suggestion
    "groupName": "NVM",
    "description": "18.18.0 breaks see KIT-2787"
```

PR #3270, opened 2023-10-11
refactor(analytics): makeAnalyticsAction accepts structured param

louis-bompart:
First PR for relay support!

Essentially, I want to leverage the `makeAnalyticsAction` central position in the analytics system as the branching point between the three analytics modes (legacy, duplex, next).

To support legacy (cajs) properly without ravaging the codebase (there are 100+ uses of this function), I modified makeAnalyticsAction so that it could be extended in the future (for example with props relative to Relay).
But, I also made its former signature compatible, so that the old usage still works.
Eventually, when all methods have migrated to the new structured object, we could remove the old signature.

Motivations:
 - Honestly, the main one is that I'm too lazy to change these 100 references in one go.
 - Too many changes in too many places at the same time. Does not fit with the 'smooth' transition I'd like to put in place.
 
 https://coveord.atlassian.net/browse/KIT-2818


packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -265,21 +265,101 @@ function makePreparableAnalyticsAction<
...
+  ) => SearchPageClientProvider;
+}
+
+export function makeAnalyticsAction<
+  EventType extends AnalyticsType,
+  StateNeeded extends
+    StateNeededBySearchAnalyticsProvider = StateNeededBySearchAnalyticsProvider,
+>(
+  prefix: string,
+  analyticsType: LegacyAnalyticsOptions<
```
olamothe, line 299: For some small readability gains, can we have named type for all this ?
For example:
```typescript
type __legacy__getBuilder = LegacyAnalyticsOptions<
    EventType,
    StateNeeded
  >['__legacy__getBuilder']

[...]


__legacy__get_builder: __legacy__getBuilder,
__legacy__provider: __legacy__provider,
```

packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -265,21 +265,101 @@ function makePreparableAnalyticsAction<
...
+  StateNeeded
+>;
+export function makeAnalyticsAction<
+  EventType extends AnalyticsType,
+  StateNeeded extends
+    StateNeededBySearchAnalyticsProvider = StateNeededBySearchAnalyticsProvider,
+>(
+  ...params:
+    | [
+        LegacyAnalyticsOptions<EventType, StateNeeded>['prefix'],
```
olamothe, line 332: Going with my above comment about small readability gains, we'd could reuse the temporary types here.

packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -265,21 +265,101 @@ function makePreparableAnalyticsAction<
...
+        LegacyAnalyticsOptions<EventType, StateNeeded>['analyticsType'],
+        LegacyAnalyticsOptions<EventType, StateNeeded>['__legacy__getBuilder'],
+        LegacyAnalyticsOptions<EventType, StateNeeded>['__legacy__provider']?,
+      ]
+    | [AnalyticsActionOptions<EventType, StateNeeded>]
+): PreparableAnalyticsAction<WrappedAnalyticsType<EventType>, StateNeeded> {
+  return params.length === 1
```
olamothe, line 339: As you said, this does make for some ugly and hard to read code, which I'm honestly fine with since it's temporary (famous last word).

Could we add some small comment that explains why this looks like this/that this is temporary ?
So that if someone not aware of your changes stumbles upon that, there's less "wtf" ?

Should there be some UT to ensure things don't get broken for the old system during the transition ? Or I guess we can just rely on atomic end to end tests ?

olamothe (approved):

olamothe:
Another alternative could have been to have two versions of the `CoveoSearchPageClient` (in : One coming from `coveo.analytics`, and one that's just a fake we temporarily implement for relay ?

ie: Do the branching in `src > api > analytics > search-analytics.ts > configureAnalytics` instead ?

Would that be simpler ?

louis-bompart:
> Another alternative could have been to have two versions of the `CoveoSearchPageClient` : One coming from `coveo.analytics`, and one that's just a fake we temporarily implement for relay ?
> 
> ie: Do the branching in `src > api > analytics > search-analytics.ts > configureAnalytics` instead ?
> 
> Would that be simpler ?

I thought about that but decided not to.
The goal of this PR is to avoid at all cost shared interfaces between CAJS and Relay, so that CAJS can be wiped out easily in v3.
That's why there's a lot of `|` and `&` interfaces types

packages/headless/src/features/analytics/analytics-utils.ts
louis-bompart: a5e9cd9a53cb3a546fbdfae6bea37e7799a4eedf

PR #3271, opened 2023-10-11
feat(commerce): make sort comply with api

Spuffynism:
I (wrongly) assumed we could reuse much of the existing sort interfaces to describe the PLP v2 sorting. This PR makes sorting be in line with the actual PLP v2 api response and request structure, which differs from the current sort in headless.

I've also:
- Added an integration test for sorting
- Removed sorting code which was accidentally added to PLP v1 parts of the code
- Extracted the `sort` feature into its own folder, to follow the structure we have for `pagination`
image

[CAPI-92]

[CAPI-92]: https://coveord.atlassian.net/browse/CAPI-92?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olamothe (commented):
Hum, there might be a better way to extend what's actually in `src > feature > sort-criteria > criteria.ts`, as opposed to the old product listing v1.

If you look at how it works for search, we have a different object `sort-criteria` that then gets built into a string in `sort-criteria-slice.ts` with `buildCriterionExpression`

So potentially, we could just reuse what's in `sort-criteria.ts` entirely, but in `product-listing-sort-slice`, we have a new function to store it to it's "proper shape".

We could also just do that transformation in `product-listing-v2-request`: So, we store in the state the object shape as it appear in `sort-criteria.ts`, and only when building the request does the modification happen.

packages/headless/src/controllers/commerce/product-listing/sort/headless-product-listing-sort.test.ts
```diff
@@ -78,8 +78,8 @@ describe('headless product-listing-sort', () => {
 
     it('calling #isSortedBy with a criterion different to the one in state returns false', () => {
```
fbeaudoincoveo, line 79:
```suggestion
    it('calling #isSortedBy with a criterion different from the one in state returns false', () => {
```

fbeaudoincoveo (approved):
LGTM!

Spuffynism:
@olamothe It seems to me that re-using the sort types/interfaces that already exist won't quite be possible. The transformation from the sort we receive from the API to the sort as described in the sort-criteria would be lossy, as the sort-criteria types do not contain a `displayName` property:
https://github.com/coveo/ui-kit/blob/64ff6d8394b05bc03a50fe813a174ed40a8b7ff1/packages/headless/src/features/sort-criteria/criteria.ts#L52

https://github.com/coveo/ui-kit/blob/5e4bb11c42e9b36f885dc8c1d7c88aae687ff5fe/packages/headless/src/features/commerce/product-listing/sort/product-listing-sort.ts#L10-L14

However, we could reuse some types from `criteria.ts`, even if they are not interchangeable. Would that be a better option? Should we favor uniformity with headless, or uniformity with the underlying API?

Spuffynism:
Re-requesting your reviews @olamothe and @fredboutin since I've applied @olamothe 's suggestion, and moved `sort` files from `product-listing/sort` to `sort`, in order to match the structure we have with the pagination.

image

I've also integrated these controllers in barca to confirm that they work with the current API:
image

packages/headless/src/features/commerce/sort/sort-slice.ts
```diff
@@ -0,0 +1,42 @@
...
+import {getCommerceSortInitialState} from './sort-state';
+import {fetchProductListing} from '../product-listing/product-listing-actions';
+import {SortOption} from '../../../api/commerce/product-listings/v2/sort';
+import {
+  buildRelevanceSortCriterion,
+  SortBy,
+  SortCriterion,
+} from '../../sort/sort';
+
+const sortResponseToSort = (sort: SortOption): SortCriterion => {
```
fbeaudoincoveo, line 12: This name is a bit confusing; it can be read as "sort a response that needs to be sorted" or "convert a sort response to sort"

I'd name it `convertSortOptionToStateCriterion` or `convertResponseSortToStateSort`

packages/headless/src/integration-tests/commerce.test.ts
```diff
@@ -69,4 +83,16 @@ describe.skip('commerce', () => {
       ])
     );
   });
+
+  it('applies sort to product listing', async () => {
```
fbeaudoincoveo, line 87: Oh good call, I'll add integration tests for pagination too

fbeaudoincoveo (approved):

packages/headless/src/integration-tests/commerce.test.ts
Spuffynism: They're skipped for now, since we don't have an org with stable-enough fixtures. But they're definitely helpful as a sanity check before doing an actual implementation 😅

packages/headless/src/features/commerce/sort/sort-slice.ts
Spuffynism: Renamed it to `mapResponseSortToStateSort`. Let me know if that's good!

olamothe (approved):

PR #3273, opened 2023-10-11
feat(headless): support did you mean with query suggestion fallback 

olamothe:
Add supports for the new "did you mean" system, based on query suggestions.
If you look at the JIRA ticket, there is a link to the design doc that describes why we went with that approach.

# New feature

* Add support for the search API parameter, called `enableFallbackSearchOnEmptyQueryResults`.
* When set to true, Search API will use the new system under the hood, and automatically correct the query in the backend. 
* We need to adapt the logic in Headless when we process the response in `search-actions-thunk-processor` to detect "the query was already corrected, we should update client side state accordingly". This is so that, for example, the search box content update itself, or that the did you mean widget display the proper message.
* Add new option on `buildDidYouMean` headless controller to opt in the new system. By default, it is set to false. This is because there's backward compatibility concerns when the new system is used. This is also described in the design doc.
* Add new properties on `atomic-did-you-mean` component to easily use the new system.

# Automatic query
* `automaticallyCorrectQuery=false` and `enableFallbackSearchOnEmptyQueryResults=true` are inherently incompatible. The reason why is also described in the design doc. When this is configured, warn the implementers about the contradiction + force to use the old index based system. 
* Added "better support" for `automaticallyCorrectQuery` option. Expose new options for the `didYouMean` headless controller + `atomic-did-you-mean` component.

# V3

In v3, we can change the default value for `enableFallbackSearchOnEmptyQueryResults` to true.

https://coveord.atlassian.net/browse/KIT-2814


olamothe:
@erocheleau : Something you might want to know, for quantic.

olamothe:
@srosenberg-coveo : I'll start working on a draft for docs.coveo.com. We will need to adapt this page here: https://docs.coveo.com/en/1810/searching-with-coveo/query-correction-feature

packages/headless/src/features/did-you-mean/did-you-mean-actions.ts
```diff
@@ -20,3 +20,11 @@ export const applyDidYouMeanCorrection = createAction(
   'didYouMean/correction',
   (payload: string) => validatePayload(payload, requiredNonEmptyString)
 );
+
+export const enableFallbackSearchOnEmptyQueryResults = createAction(
```
acote-coveo, line 24: If we use that name `fallbackSearchOnEmptyQueryResults` and we eventually change the name in our API, are we gonna be able to change the name in the headless state or it should remain the same?

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.ts
```diff
@@ -31,6 +37,19 @@ export interface DidYouMeanProps {
...
+   * This system requires a working and properly configured query suggestions model in the Coveo platform.
+   *
+   * This option is off by default. As such, the Coveo platform will use an older query correction system, powered solely by the index.
+   * By opting in this new system, the Coveo Search API will stop returning the `queryCorrections` field in the response.
+   * Instead, it will start returning a `changedQuery` field.
+   * This implies that the usage of this option introduce a breaking change in the way query corrections are handled, both at the Search API and Headless level.
+   *
+   * The default value is `false`.
+   */
+  enableFallbackSearchOnEmptyQueryResults?: boolean;
```
acote-coveo, line 52: Should we name this `automaticallyCorrectQueryServerSide` or something along those lines? 

These two parameters are kind of dependant and could have been replaced by an enum:

```typescript
type automaticallyCorrectQuery: 'disabled' | 'client-side' | 'server-side'
```

packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -172,29 +173,29 @@ export class AsyncSearchThunkProcessor<RejectionType> {
...
-    const {results, queryCorrections} = successResponse;
+    const {results, queryCorrections, changedQuery} = successResponse;
 
-    const shouldBeAutomaticallyRetried =
+    const shouldExecuteClassicDidYouMeanAutoCorrection =
+      enableDidYouMean === true &&
+      automaticallyCorrectQuery === true &&
       results.length === 0 &&
-      queryCorrections.length !== 0 &&
-      enableDidYouMean &&
```
acote-coveo, line 180: Why do we check for `enableDidYouMean` here? If someone has `enableDidYouMean` set in their query param overrides, does it mean it wouldn't auto correct here?

packages/headless/src/features/did-you-mean/did-you-mean-actions.ts
olamothe: > If we use that name fallbackSearchOnEmptyQueryResults and we eventually change the name in our API, are we gonna be able to change the name in the headless state or it should remain the same?

We prefer to stick to the Search API name. Now is the time to change it on your end, or "wait for a major version" kinda thing.

I will wait to merge/publish this until it's decided.

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.ts
olamothe: The type automaticallyCorrectQuery above (boolean) is already part of public Headless API (through global state/actions to control it).

And so we can't change it now to become an enum.

We could mark it as deprecated, however, and create a new one called automaticQueryCorrection: 'disabled' | 'client-side' | 'server-side'.

I kinda like that 🤔

packages/headless/src/features/search/search-actions-thunk-processor.ts
olamothe: If there's a query override for enableDidYouMean, we will probably have a bug, yes. 

That's a general problem with some query param which the UI * must * know about to do it's job properly, but that people can just decide to override in the backend without the front end knowing about it.

There's a couple of those query parameter actually (pagination comes to mind) that we know about, but the problems don't come too often...

packages/headless/src/features/did-you-mean/did-you-mean-actions.ts
olamothe: The headless option. Which means there's some divergence from Search API, which is not ideal. 

The problem is related to documentation/glossary. 

ie: if someone searches `automaticQueryCorrection` they might not find Search API doc for it, or vice versa if they search for `enableFallbackSearchOnEmptyQueryResults`.

But, when we can offer a "better" abstraction over what the search API parameter actually, it's worth it.

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.ts
louis-bompart: I like the enum too

louis-bompart (approved):
lgtm, enum tho would be nice

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.ts
louis-bompart: - [ ] enum

PR #3277, opened 2023-10-11
feat: add analyticsMode & relay

louis-bompart:
- Add the config option analyticMode to chose between:
  - Legacy: Use CAJS only
  - Next: Use Relay only
  - Duplex: Use both

The option default to legacy of course, and we don't support it for customers (for now)

And of course, add relay into the mix to make things work


packages/headless/src/features/configuration/configuration-state.ts
```diff
@@ -57,12 +57,20 @@ export interface AnalyticsState {
    * Specifies if analytics tracking should be enabled. By default analytics events are tracked.
    */
   enabled: boolean;
+
   /**
+   * @internal
    * The Analytics API base URL to use.
-   * By default, will append /rest/ua to the platformUrl value.
+   * By default, will append /rest/organizations/${organizationId}/events/v1 to the platformUrl value.
```
louis-bompart, line 64:
```suggestion
   * By default, will append `/rest/ua` to the platformUrl value.
```

packages/headless/src/features/configuration/configuration-state.ts
```diff
@@ -57,12 +57,20 @@ export interface AnalyticsState {
...
    */
   legacyApiBaseUrl: string;
 
+  /**
+   * The Analytics API base URL to use.
+   * By default, will append /rest/ua to the platformUrl value.
```
louis-bompart, line 70:
```suggestion
   * By default, will append `/rest/organizations/${organizationId}/events/v1` to the platformUrl value.
```

packages/headless/src/features/configuration/configuration-actions.ts
```diff
@@ -118,6 +118,11 @@ export interface UpdateAnalyticsConfigurationActionCreatorPayload {
    */
   legacyApiBaseUrl?: string;
 
+  /**
+   * The Usage Analytics API base URL to use (e.g., `https://platform.cloud.coveo.com/rest/ua`).
```
louis-bompart, line 122:
```suggestion
   * The Usage Analytics API base URL to use (e.g., `https://platform.cloud.coveo.com/rest/organizations/${organizationId}/events/v1`).
```

packages/headless/src/app/engine-configuration.ts
```diff
@@ -150,6 +150,12 @@ export interface AnalyticsConfiguration {
    * Specifies the URL of the current page or component.
    */
   documentLocation?: string;
+  /**
+   * TBD
+   * @internal
+   */
+  trackingId?: string;
```
samisayegh, line 157: We can start optional, but we'll eventually want `trackingId` to be required. Atm one can pass any string. In Q4, Data Platform will make it possible to create `trackingId` values via the AdminUI, at which point only values registered with the Admin will be valid,

packages/headless/src/app/engine-configuration.ts
```diff
@@ -150,6 +150,12 @@ export interface AnalyticsConfiguration {
...
+  analyticsMode?: 'legacy' | 'next' | 'duplex';
```
samisayegh, line 158: Wondering if it would be clearer if the modes we named according to the protocol? e.g.

```suggestion
  analyticsMode?: 'usage analytics' | 'event protocol' | 'duplex';
```

packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -745,3 +823,14 @@ export const makeCommerceAnalyticsAction = <
     }
   );
 };
+
+function logNextEvent<PayloadType extends RelayPayload = RelayPayload>(
+  emitEvent: (
+    type: string,
+    payload: RelayPayload
+  ) => Promise<void | ValidationResponse>,
```
samisayegh, line 831: Would it be possible to remove `ValidationResponse` from the possible return types? While it's supported now, this is something we plan on changing once the Explorer debugging tool is available. Validation will be sent to the tool rather than returned by the `emit` function.

Since `emit` will soon use beacon api under the hood, it will become fire-and-forget removing the need to await a promise. Maybe the return type here can simply be `void`? If `tsc` complain though, we can do this in a future PR,

samisayegh (approved):

packages/headless/src/features/analytics/analytics-utils.ts
louis-bompart: - [ ] Todo: Coerce to void

packages/headless/src/app/engine-configuration.ts
louis-bompart: Yeah, for now, no need to over fret about this IMO.
Maybe we should coerce backend wise the trackingID tho

packages/headless/src/features/configuration/configuration-slice.ts
```diff
@@ -105,6 +105,15 @@ export const configurationReducer = createReducer(
         if (!isNullOrUndefined(action.payload.legacyApiBaseUrl)) {
           state.analytics.legacyApiBaseUrl = action.payload.legacyApiBaseUrl;
```
olamothe, line 106: Was discussed live, but for current v2, we cannot do breaking change on the state itself (ie: it's part of the public contract).

So moving the "old" to `legacyApiBaseUrl` cannot be done for v2. We can do it when moving to v3 though.

We can add `nextApiBaseUrl` property and store it there instead

packages/headless/src/features/configuration/configuration-state.ts
```diff
@@ -114,6 +122,12 @@ export interface AnalyticsState {
    * Specifies the URL of the current page or component.
    */
   documentLocation: string;
+  /**
+   * TBD trackingId
```
olamothe, line 126: Should probably mark it `@internal` here also.

packages/headless/src/features/configuration/configuration-slice.ts
olamothe: Furthermore.... Since there's no "duplex" mode anymore, do we need to store both analytics endpoints in 2 different properties ?  We could just have one I guess ?

olamothe: Do we need to modify anything related to `getOrganizationEndpoints` utility method ?

In other words, if we look at it from the perspective of Atomic (in `index.html` for example) where we try to do this:
```js
      accessToken: '123',
      organizationId: '456',
      organizationEndpoints: await searchInterface.getOrganizationEndpoints('456'),
      analytics: {
             'analyticsMode' : 'next'
       }
});
```

Do we end up with the right properties stored in the state after the configuration reducer has done it's job ?

There's some code that gets executed in `updateBasicConfiguration` switch case which might need to get modified.

olamothe (changes requested):

packages/headless/src/features/configuration/configuration-slice.ts
louis-bompart: > Furthermore.... Since there's no "duplex" mode anymore, do we need to store both analytics endpoints in 2 different properties ? We could just have one I guess ?

In case we need to have some cohabitation, and because it's the simplest, I'll go with the nextApiBaseUrl instead of adding conditions on the value of apiBaseUrl.

olamothe (approved):

PR #3278, opened 2023-10-12
Headless ESM (yep)

louis-bompart:
This makes Headless work with NextJS.
It probably breaks something somewhere.


olamothe:
Closed to clean up the PR tab, re-open if necessary !

PR #3281, opened 2023-10-13
feat(atomic): add toggle to disable atomic-generated-answer component

lbergeron:
[SVCC-3224](https://coveord.atlassian.net/browse/SVCC-3224)

This PR updates `atomic-generated-answer` to include a toggle switch allowing the user to show/hide the generated answer.

Expected behavior:

* `atomic-generated-answer` is shown only when a stream can be read.
* the toggle switch is checked by default.
* when the user unchecks the toggle switch, the `atomic-generated-answer` is collapsed, hiding the generated answer.
* when the user checks the toggle switch, the `atomic-generated-answer` is expanded, showing the generated answer.
* the state of the toggle switch is persisted across sessions.

https://github.com/coveo/ui-kit/assets/10946843/c1c636a1-2ed1-430f-8d8a-7801abf00364

[SVCC-3224]: https://coveord.atlassian.net/browse/SVCC-3224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/common/switch.tsx
nathanlb, line 1: Cool! All the classes and attributes are very neat.
from the video i think the switch might be a little too chunky. In the mockups it has a height of 24px. Is it just my eyes?

lbergeron: Thanks for pointing that out! At this moment, the atomic version is 64 x 32px. I'll double check the mockups and adjust accordingly.

lbergeron: The switch is now smaller (48x24). Also, the switch handle looks smaller in the Off position but, it is only an illusion. I suppose it is caused by the low contrast over the background. Now the handle is set to 16px regardless of the switch position.

toggle-switch-updated

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -54,9 +62,16 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
   @Element() private host!: HTMLElement;
 
   private stopPropagation?: boolean;
+  private storage: SafeStorage = new SafeStorage();
+  private data?: GeneratedAnswerData;
 
   public initialize() {
-    this.generatedAnswer = buildGeneratedAnswer(this.bindings.engine);
+    this.data = this.readStoredData();
+    this.generatedAnswer = buildGeneratedAnswer(this.bindings.engine, {
```
fbeaudoincoveo, line 70: If `initialState.isVisible` is initially `false`, will an answer still be generated under the hood, and simply not displayed?

If that is the case, I feel like this is a fairly costly operation to perform for users who don't care for generated answers 😅 

Ideally, I'd expect that if the switch is initially off, then no generated answer would be requested, and if I turn it on, a generated answer would be requested at that point.

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -143,30 +184,45 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
+
+            <Switch
+              part="toggle"
+              checked={this.isAnswerVisible}
+              onToggle={(checked) => {
+                checked
+                  ? this.generatedAnswer.show()
+                  : this.generatedAnswer.hide();
+              }}
+              ariaLabel={this.bindings.i18n.t('generated-answer-title')}
```
fbeaudoincoveo, line 213: Should the ariaLabel change based on the state of the switch? E.g., if checked is true, it would be "Hide generated answer" and if it's false it would be "Show generated answer"?

lbergeron: According to [these guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/switch/#:~:text=Important%3A%20it%20is%20critical%20the%20label%20on%20a%20switch%20does%20not%20change%20when%20its%20state%20changes.), the label should remain the same regardless of the switch state.

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
lbergeron: It's funny, I asked the exact same question when implementing the feature :smile: .
Long story short, the switch must be seen as a way to expand and collapse the answer more than a way to enable or disable the answer generation. In fact, Search API callers cannot control whether answers are going to be generated. Answers are generated based on the query pipeline configuration.

fbeaudoincoveo: OK, wasteful, but makes sense 😄

fbeaudoincoveo (approved):

mmitiche (approved):
Looks good to me 👌

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
lbergeron: I went with this as the switch label because it is the only piece of text that is visible when the generated answer is collapsed. Considering it acts like a section title, it made sense to me to also use it as the switch label.

packages/atomic/cypress/e2e/generated-answer-assertions.ts
```diff
@@ -13,3 +14,39 @@ export function assertLogOpenGeneratedAnswerSource(log: boolean) {
...
+    );
+  });
+}
+
+export function assertLocalStorageData(data: {isVisible: boolean}) {
+  it(`should have value in local storage: ${JSON.stringify(data)}`, () => {
+    cy.window().then((win) => {
+      const stored = JSON.parse(
+        win.localStorage.getItem('coveo-generated-answer-data') ?? ''
+      );
```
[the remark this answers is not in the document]
lbergeron: Good point, will fix that.

jelmedini (approved):
Looks really good!

nathanlb (approved):

PR #3286, opened 2023-10-16
feat(quantic): toggle button to show and hide generated answer added

mmitiche:
[SFINT-5213](https://coveord.atlassian.net/browse/SFINT-5213)

This PR updates Quantic Generated Answer component to include a toggle switch allowing the user to show/hide the generated answer.

Expected behavior:

- the toggle switch is checked by default.

- when the user unchecks the toggle switch the event `generatedAnswerHideAnswers` is logged and the Quantic Generated Answer component is collapsed, hiding the generated answer.
- when the user checks the toggle switch the event `generatedAnswerShowAnswers` is logged and the Quantic Generated Answer component is expanded, showing the generated answer.
- the state of the toggle switch is persisted in the session storage.

PS: this feature has been tested with locker service enabled ✅

#### Demo:

https://github.com/coveo/ui-kit/assets/86681870/87a72cab-6b90-4ced-a63c-2daa3ac13065


packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -81,6 +88,39 @@ describe('quantic-generated-answer', () => {
           Expect.dislikeButtonIsChecked(true);
         });
       });
+
+      it('should display the toggle generated answer button', () => {
+        Expect.displayToggleGeneratedAnswerButton(true);
+        Expect.toggleGeneratedAnswerButtonIsChecked(true);
+
+        scope('when toggeling off the generated answer', () => {
```
lbergeron, line 96:
```suggestion
        scope('when toggling off the generated answer', () => {
```

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -12,6 +12,8 @@ import {scope} from '../../../reporters/detailed-collector';
 import {GeneratedAnswerActions as Actions} from './generated-answer-actions';
 import {GeneratedAnswerExpectations as Expect} from './generated-answer-expectations';
 
+const GENERATED_ANSWER_VISIBILITY_KEY = 'generatedAnswerIsVisible';
```
lbergeron, line 15: In Atomic, local storage keys are formatted like this: `coveo-{component name}-data`. I'd suggest to use the same here. Also, in Atomic, I opted to store the visibility value in a data object so we can add additional information to the same storage if needed. I'd suggest to align the components to work the same way.

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -81,6 +88,39 @@ describe('quantic-generated-answer', () => {
...
+          Expect.displayDislikeButton(false);
+          Expect.logHideGeneratedAnswer(streamId);
+          Expect.sessionStorageExists(GENERATED_ANSWER_VISIBILITY_KEY, true);
+          Expect.sessionStorageContains(
+            GENERATED_ANSWER_VISIBILITY_KEY,
+            'false'
+          );
+        });
+
+        scope('when toggeling on the generated answer', () => {
```
lbergeron, line 110:
```suggestion
        scope('when toggling on the generated answer', () => {
```

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -87,6 +94,20 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
+      GENERATED_ANSWER_VISIBILITY_KEY
+    );
+    if (['true', 'false'].includes(storedGeneratedAnswerVisibility)) {
+      return this.headless.buildGeneratedAnswer(engine, {
+        initialState: {
+          isVisible: storedGeneratedAnswerVisibility === 'true' ? true : false,
+        },
+      });
+    }
+    return this.headless.buildGeneratedAnswer(engine);
```
lbergeron, line 108: IMO storing a data object would made this code a little simpler.

Also, why not always pass the `initialState` when building the controller?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
```diff
@@ -1,41 +1,56 @@
...
+          <div class="slds-var-m-left_medium">
+            <lightning-input
+              data-cy="generated-answer__toggle-button"
+              type="toggle"
+              onchange={toggleGeneratedAnswer}
+              checked={isVisible}
+              variant="label-hidden"
+              message-toggle-active=""
+              message-toggle-inactive=""
+              label={toggleGeneratedAnswerLabel}
```
lbergeron, line 35: According to [these guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/switch/#:~:text=Important%3A%20it%20is%20critical%20the%20label%20on%20a%20switch%20does%20not%20change%20when%20its%20state%20changes.), the switch label should always be the same.

lbergeron (commented):
Very nice!

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
mmitiche: Very good point! I will update accordingly

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
mmitiche: Today I learned! thank you

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
mmitiche: I'm adding this logic to pass the `initiallState` only when valid values are being passed to the initial state to avoid having errors with the `setVisible` action in headless.

The scenario I'm trying to prevent with this is a scenario where the client is  manually updating the value stored in the session storage to an invalid value (any value other than `true` or `false`).
In this scenario instead of having a bunch of errors in the console coming from headless saying that it got an invalid value for the `setVisible` action, I'm, from the Quantic component, completely ignoring this invalid values, the Quantic component is taking the session storage in consideration only when it contains valid data. That's my logic here.

lbergeron: Fair enough

lvu285:
Nice job!
- I wonder if the button has tooltip as I don't see it when you clicked ?

SimonMilord (approved):
Do we want to add a tooltip to the toggle?
Other than that looks nice ! Good job !

* just saw lucille's comment about the tooltip

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -160,6 +191,14 @@ export default class QuanticGeneratedAnswer extends LightningElement {
     return this?.state?.isStreaming;
   }
 
+  get shouldDisplayFeedback() {
+    return this.isVisible && !this.isStreaming;
+  }
+
+  get isVisible() {
+    return this?.state?.isVisible;
```
erocheleau, line 199: I don't like this as this will return `undefined` if the state .isVisible is missing, which means the genQA will be hidden.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -160,6 +191,14 @@ export default class QuanticGeneratedAnswer extends LightningElement {
     return this?.state?.isStreaming;
   }
 
+  get shouldDisplayFeedback() {
+    return this.isVisible && !this.isStreaming;
```
erocheleau, line 195:
```suggestion
    return this.isVisible !== false && !this.isStreaming;
```
Maybe if we are specific to expect false?

erocheleau (commented):
Just a question around the state of isVisibile.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
mmitiche: Practically this will never be undefined, as the state in Headless is always defined with the `isVisible` property that has an initial value of `true`, I can remove the `?` to remove the ambiguity.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -160,6 +200,14 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
+  }
+
+  get isVisible() {
+    return this?.state?.isVisible;
```
mmitiche, line 208:
```suggestion
    return this.state.isVisible;
```

mmitiche:
> Nice job!
> 
> * I wonder if the button has tooltip as I don't see it when you clicked ?
> * And also, please help me to check and make sure the accessibility of button

No the button do not have a tooltip that appears on hover, we are here using the Salesforce [Lightning Input component](https://developer.salesforce.com/docs/component-library/bundle/lightning-input/example) that doesn't  come with a tooltip.
However it comes with all accessibility logic built-in.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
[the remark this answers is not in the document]
mmitiche: Actually any value other than `true` or `false` will cause a Headless validation error, that's why I'm having this logic, to  only accept `true` or `false`.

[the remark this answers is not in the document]
erocheleau: Yea I know, but I was trying to enforce true or false only, and basically default to true in the initialState if the session contained anything other than `false`

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: I would say yes to that or default to specific true or false. Don't let it run with a potentially ambiguous "undefined"

erocheleau (changes requested):
Some small changes in logic

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -94,14 +91,22 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
-    );
-    if (['true', 'false'].includes(storedGeneratedAnswerVisibility)) {
+    const storedGeneratedAnswerData = this.readStoredData();
+    const storedGeneratedAnswerVisibility =
+      storedGeneratedAnswerData?.isVisible;
+    if (typeof storedGeneratedAnswerVisibility == 'boolean') {
       return this.headless.buildGeneratedAnswer(engine, {
         initialState: {
-          isVisible: storedGeneratedAnswerVisibility === 'true' ? true : false,
+          isVisible: storedGeneratedAnswerVisibility,
```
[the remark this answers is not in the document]
mmitiche: Good idea more simple, will update 👌
Tho in my other it would still build a generated answer if something weird is in the Session Storage, basically the code was following this logic: if there is a weird value in the local storage then create the controller without an initial state. When creating the controller that way Headless will automatically use the default value which is to by default show the generated answer.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -94,14 +91,22 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
+  }
+
   buildHeadlessGeneratedAnswerController(engine) {
-    const storedGeneratedAnswerVisibility = sessionStorage?.getItem(
-      GENERATED_ANSWER_VISIBILITY_KEY
-    );
-    if (['true', 'false'].includes(storedGeneratedAnswerVisibility)) {
+    const storedGeneratedAnswerData = this.readStoredData();
+    const storedGeneratedAnswerVisibility =
+      storedGeneratedAnswerData?.isVisible;
```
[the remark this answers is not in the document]
mmitiche: yeah it was just a security net x)

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
mmitiche: I think it's clear enough to have a simple condition as `if (this.isVisible`)

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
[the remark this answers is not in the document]
erocheleau: ah ok i think this was my view because I was checking only what was modified in last commit, I didn't see the rest of the if

erocheleau (approved):
Looks good after the changes

PR #3289, opened 2023-10-16
fix(quantic): fixed case classification cypress test

mmitiche:
[SFINT-5240](https://coveord.atlassian.net/browse/SFINT-5240)

Apparently Salesforce changed the structure of the objects returned when using the `getPicklistValuesByRecordType` wire adapter used to get a all picklist values of a field of a given object.

This change cause an issue in the mock function we used to mock this wire adapter call causing the failure of the cypress tests.

Adjusting the mock function to return exactly what's expected solved the issue.

[SFINT-5240]: https://coveord.atlassian.net/browse/SFINT-5240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
+++ b/packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
@@ -47,11 +47,11 @@ describe('quantic-case-classification', () => {
   const coveoDefaultField = 'sfpriority';
   const nonCorrespondingCoveoField = 'sforigin';
   const allOptions = [
-    {id: '0', value: 'Very low', label: 'Very low'},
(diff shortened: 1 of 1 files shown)


SimonMilord (approved):
I would add the link of the source where you found the issue in the PR for future reference

erocheleau (approved):
Maybe we should change our matcher instead to look for a partial match instead of an exact match? :)

mmitiche:
> I would add the link of the source where you found the issue in the PR for future reference

We don't really have a source, the issue was discovered when the tests started failing in the CI.

PR #3290, opened 2023-10-16
feat(headless): add Copy generated answer controller action for GenQA…

jelmedini:
[SVCC-3229](https://coveord.atlassian.net/browse/SVCC-3229)

- Add action to copy directly to the clipboard the content of the generated answer when I click on copy button
- Dispatch the proper analytic event when clicking on copy to the clipboard button.

[SVCC-3229]: https://coveord.atlassian.net/browse/SVCC-3229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fbeaudoincoveo (commented):
The new method has no impact on the state: it only logs an event and attempts to perform an action in the browser. I personally find it awkward to have it on a Headless controller. 

I feel like this should be done in a UI component instead. E.g., something similar to `quanticResultCopyToClipboard.js`

fbeaudoincoveo:
> The new method has no impact on the state: it only logs an event and attempts to perform an action in the browser. I personally find it awkward to have it on a Headless controller.
> 
> I feel like this should be done in a UI component instead. E.g., something similar to `quanticResultCopyToClipboard.js`

You could have a method to log the analytics action from the controller, but the copy to clipboard part itself should be done in the UI component, not in the Headless controller. Otherwise it won't work with SSR.

PR #3295, opened 2023-10-17
feat: add trackingId if analyticsMode=next

louis-bompart:
https://coveord.atlassian.net/browse/KIT-2829

### DIFF
--- a/packages/headless/src/features/configuration/analytics-params.ts
+++ b/packages/headless/src/features/configuration/analytics-params.ts
@@ -22,6 +22,8 @@ export const fromAnalyticsStateToAnalyticsParams = async (
       ...(s.documentLocation && {documentLocation: s.documentLocation}),
       ...(s.deviceId && {deviceId: s.deviceId}),
       ...(getPageID() && {pageId: getPageID()}),
+      ...(s.analyticsMode && s.trackingId && {trackingId: s.trackingId}),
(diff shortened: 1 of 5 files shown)


packages/headless/src/features/configuration/analytics-params.ts
```diff
@@ -22,6 +22,8 @@ export const fromAnalyticsStateToAnalyticsParams = async (
       ...(s.documentLocation && {documentLocation: s.documentLocation}),
       ...(s.deviceId && {deviceId: s.deviceId}),
       ...(getPageID() && {pageId: getPageID()}),
+      ...(s.analyticsMode && s.trackingId && {trackingId: s.trackingId}),
+      ...{enabled: s.analyticsMode === 'next' ?? false},
```
olamothe, line 26: As discussed live this should be `capture=true`.

We will also want to not send any analytics payload to Search API if the end user has DNT=true.
But this can (should?) be part of a different PR.

olamothe (approved):

packages/headless/src/features/configuration/analytics-params.ts
louis-bompart: Should, indeed.
Honouring the DNT/GPC flags should be handled at those two levels:
 - For SAPI calls, in Headless
 - For /events calls, in Relay (I reckon that's already the case)

dmbrooke (approved):

PR #3298, opened 2023-10-17
feat(atomic,headless): support new facet sort model

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2779

This issue got out of scope a smidgen, but I think this is the smoothest way to introduce these changes to implementers.

To summarize, we now support both string (old) and explicit/object (new) facet sort criterion. How I decided to support this in Atomic is by adding a new `sortOrder` prop that _always get leveraged when updating the corresponding Headless controller_. This means that for specific facets, Atomic will now always send explicit sort criterion (since `ascending` is the default behaviour anyways). 

For Headless, I mostly just added onto the existing "switcheroo" that I originally did for `alphanumericDescending`, only this time, it was to handle invalid combinations (namely the `automatic` `type` within an explicit facet sort object). If the `automatic` type is found within an explicit sort criteria, it overwrites the request sortCriteria accordingly. As with my previous change, I decided to try and keep things as close to the actual request as possible as to not have too many conditions scattered throughout the codebase.


packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -75,6 +76,11 @@ export class AtomicInsightFacet
    */
   @Prop({reflect: true}) public sortCriteria: InsightFacetSortCriterion =
     'automatic';
+  /**
+   * The sort criterion to apply to the returned facet values.
+   * Possible values are 'score', 'alphanumeric', 'occurrences', and 'automatic'.
+   */
```
louis-bompart, line 82:
```suggestion
  /**
   * The sort order to apply to the returned facet values.
   * Possible values are 'ascending' and 'descending'.
   */
```

packages/headless/src/features/search/search-request.ts
```diff
@@ -116,19 +116,28 @@ function getAllFacets(state: StateNeededBySearchRequest) {
...
-    At the moment, this is only supported for alphanumeric sorting, but will likely transition to this pattern for other types in the future. */
-    if (request.sortCriteria === 'alphanumericDescending') {
-      return {
-        ...request,
-        sortCriteria: {
-          type: 'alphanumeric',
-          order: 'descending',
-        },
+    let newSortCriteria = request.sortCriteria;
+    if (
```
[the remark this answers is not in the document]
louis-bompart: Uh funky.
Nothing came to mind tho this looks like a fitting hack to me 👍

louis-bompart (approved):
LGTM

packages/headless/src/features/search/search-request.ts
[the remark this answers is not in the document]
dmbrooke: > Uh funky.

Yup, agreed 🙃 I'll reach out to the Search API team and ask if there's anyway they could support `automatic` within the new type so we wouldn't need this hack. Surely it would be simple enough to handle `automatic` the same as it currently works for the new type...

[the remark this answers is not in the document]
dmbrooke: UPDATE: I spent some time this morning trying to implement a more "concrete" solution where it's enforced by type (e.g., `automatic` and `alphanumericDescending` aren't valid for `type` within `SpecificSortCriteriaExplicit`), but it introduced more edge-case handling throughout... 

That being said, I'm still trying to get a hold of someone from the Search API team to figure out exactly what their future plans are. This works now, but if it turns out that we will need to support both string (`automatic`) and object values simultaneously for the foreseeable future, I think it's worth investing the time to implement something that's less of a hack 🤔

louis-bompart (approved):

packages/headless/src/features/facets/facet-set/interfaces/request.ts
```diff
@@ -8,23 +8,22 @@ import {
...
+  SpecificSortCriteriaExplicit,
 } from '../../facet-api/request';
 
 export const facetSortCriteria: FacetSortCriterion[] = [
   'score',
   'alphanumeric',
-  'alphanumericDescending',
   'occurrences',
   'automatic',
 ];
```
alexprudhomme, line 19: While looking at the facetSortCriteria const , I see that it is only used in headless-insight-facet-options.ts .Why did you remove the alphanumericDescending from this const but not from the FacetSortCriterion ?

packages/headless/src/features/facets/facet-api/request.ts
```diff
@@ -87,23 +87,29 @@ export type FacetType =
...
+  | 'descending'
+  | 'occurrences'
+  | 'automatic'
+  | 'alphanumericDescending';
+
+export type SpecificSortCriteriaType = Exclude<
+  SortCriteriaType,
+  'ascending' | 'descending' | 'automatic' | 'alphanumericDescending'
+>;
+
```
alexprudhomme, line 103: If i understood correctly, the SortCriteriaType will be removed in v3. Would it be possible to have a entire different type instead of relying on Exclude the soon to be removed type ?

alexprudhomme (commented):
What is the plan for V3 with this ? Will we remove the string sortCriterion ?

packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
```diff
@@ -112,9 +114,14 @@ export class AtomicFacet implements InitializableComponent, BaseFacet<Facet> {
   @Prop({reflect: true}) public withSearch = true;
   /**
    * The sort criterion to apply to the returned facet values.
-   * Possible values are 'score', 'alphanumeric', 'alphanumericDescending', 'occurrences', and 'automatic'.
+   * Possible values are 'score', 'alphanumeric', 'occurrences', and 'automatic'.
    */
   @Prop({reflect: true}) public sortCriteria: FacetSortCriterion = 'automatic';
```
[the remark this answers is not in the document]
dmbrooke: Whoops! I temporarily considered removing `alphanumericDescending`, but decided against it since we can't have any breaking changes until V3. I'll scrub through this to make sure it's still there in both the `FacetSortCriterion` types as well as the documentation ✅

packages/headless/src/features/facets/facet-api/request.ts
dmbrooke: This lead me down a bit of a rabbit hole honestly. In short, we won't be removing the `SortCriteriaType` type... At least not _all of it_. Certain things, such as `automatic` and the duplicated range facet sort criterion (`ascending` and `descending`) will be used for the foreseeable future, so I think I'll take some time to strip these models down to be more obviously scoped so that when the time comes to pull the plug on the old type(s), it will be much more clear and simple 🙂

packages/headless/src/features/facets/facet-set/interfaces/request.ts
dmbrooke: > Why did you remove the alphanumericDescending from this const but not from the FacetSortCriterion ?

Because I'm a goofball that likes to cause chaos from time-to-time.

Joking aside, thanks for pointing this out! It was a mistake (see previous comment about `alphanumericDescending`) and I'll fix it now 🙂

dmbrooke:
> What is the plan for V3 with this ? Will we remove the string sortCriterion ?

I briefly outlined what needs to be done for V3 as a part of the next issue (which is linked in JIRA):
https://coveord.atlassian.net/browse/KIT-2578

TLDR; I believe that with my most recent type changes, we should only _**really**_ need to remove the `alphanumericDescending` logic for this to be ready to go for V3 🙂

dmbrooke:
Just an update, it turns out that all sort criterion are supported by the new explicit model _**except for `occurrences`**_, so there isn't much that can be done until that's fixed 😅 

This will be done here:
https://coveord.atlassian.net/browse/SEARCHAPI-9096

wmannard:
@louis-bompart is this still useful or should we close it?

PR #3299, opened 2023-10-17
feat: add ze magic cookie

louis-bompart:
The goal is to allow us to use a relay in the site without impacting their end-users.

Why Coveo-Pendragon? --> Why not?

https://coveord.atlassian.net/browse/KIT-2831


dmbrooke (approved):

olamothe (approved):

louis-bompart:
- [ ] rebase

PR #3300, opened 2023-10-18
chore: revert cypress-repeat to 1.5.0

louis-bompart:
See https://github.com/bahmutov/cypress-repeat/issues/52 for the RCA.

https://coveord.atlassian.net/browse/KIT-2833


louis-bompart:
All the 'vuln' packages have been studied, we should try to fix babel-traverse sooner than later, but no risk for our customers, so let's not fret.

olamothe (approved):

PR #3301, opened 2023-10-18
test: actually downgrade cypress split

louis-bompart:
#3300 but that actually work


mrrajamanickam-coveo (approved):
Would be good to mention the rollup renovate config addition in the PR desc.

PR #3306, opened 2023-10-19
fix(atomic): improve screen reader support

lbergeron:
[SVCC-3245](https://coveord.atlassian.net/browse/SVCC-3245)

There was an issue with the `Switch` component regarding screen reader support. Screen readers would announce the switch state as OFF regardless of the actual switch state. For example, it would lead to confusing announcements like:
```
Generated answer ON switch off
```

The solution is to set the `aria-checked` attribute using `true`/`false` as strings so the attribute is always defined on the element.

The screen reader announcement is now:
```
Generated answer ON switch on
```

[SVCC-3245]: https://coveord.atlassian.net/browse/SVCC-3245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathanlb (approved):

packages/atomic/src/components/common/switch.tsx
```diff
@@ -13,7 +13,7 @@ export const Switch: FunctionalComponent<SwitchProps> = (props) => {
   const attributes = {
     onClick: () => props.onToggle?.(!props.checked),
     'arial-label': props.ariaLabel,
-    'aria-checked': props.checked,
+    'aria-checked': props.checked ? 'true' : 'false',
```
dmbrooke, line 16: Perhaps a falsey check would achieve the same thing (didn't test)?

```suggestion
    'aria-checked': !!props.checked,
```

dmbrooke (approved):
LGTM

packages/atomic/src/components/common/switch.tsx
nathanlb:
```suggestion
    'aria-checked': String(props.checked),
```

nathanlb: I believe the point is it has to be a string

alexprudhomme (approved):

packages/atomic/src/components/common/switch.tsx
nathanlb:
```suggestion
    'aria-checked': String(!!props.checked),
```
This then 😛
jk

lvu285 (approved):

PR #3307, opened 2023-10-19
test(headless-ssr): add tests for the defineSearchParameterManager function

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2715

I tried to make sure only the defineSearchParameterManager function was tested and the rest was properly mocked.

### DIFF
--- a/packages/headless/src/controllers/search-parameter-manager/headless-search-parameter-manager.ssr.test.ts
+++ b/packages/headless/src/controllers/search-parameter-manager/headless-search-parameter-manager.ssr.test.ts
@@ -20,6 +20,7 @@ type SearchParameterManagerDefinitionType = ControllerDefinitionWithProps<
   SearchParameterManager,
   SearchParameterManagerBuildProps
 >;
+
 describe('define search parameter manager', () => {
   let searchParameterManagerDefinition: SearchParameterManagerDefinitionType;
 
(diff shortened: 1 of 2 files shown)


packages/headless/src/test/mock-engine.ts
```diff
@@ -72,6 +73,24 @@ interface MockEngine {
...
 type DispatchExts = ThunkDispatch<AppState, void, AnyAction>;
 
+export interface MockSSRSearchEngine extends MockSearchEngine {
+  waitForSearchCompletedAction(): Promise<SearchCompletedAction>;
+}
+
+/**
+ * For internal use only.
+ *
+ * Returns a non-functional `SSRSearchEngine`.
```
olamothe, line 83: Btw, this kind of JSDoc for internal mock function is not necessary/a bit overkill 😄

olamothe (approved):

packages/headless/src/controllers/search-parameter-manager/headless-search-parameter-manager.ssr.test.ts
```diff
@@ -0,0 +1,50 @@
...
+const buildSearchParameterManagerMock = jest.mocked(
+  buildSearchParameterManager
+);
+
+type SearchParameterManagerDefinitionType = ControllerDefinitionWithProps<
+  SSRSearchEngine,
+  SearchParameterManager,
+  SearchParameterManagerBuildProps
+>;
+describe('define search parameter manager', () => {
```
mrrajamanickam-coveo, line 23: prettier ?

packages/headless/src/test/mock-engine.ts
alexprudhomme: Yea I just matched the style of the other mockEngine functions. They all have a doc description !

mrrajamanickam-coveo (approved):
Since the define functions are currently a thin layer guess there is nothing much more to test for? Wish we had code coverage % to look at.
Anyways, can revisit later when re-doing URL management I guess.

packages/headless/src/controllers/search-parameter-manager/headless-search-parameter-manager.ssr.test.ts
alexprudhomme: unfortunately prettier doesn't look at spacing like this.

PR #3308, opened 2023-10-19
chore(headless): replace deprecated platform-client functions

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2379

The scope is to remove all exported deprecated functions, and I only found 2; `platformUrl` and `analyticsUrl` within `platform-client`. I was also able to remove `coveoCloudURL` and `URLOptions` since they were exclusively used within these deprecated functions.

Seeing as we're moving to a URL scheme that depends on an orgId, but the place that still uses these deprecated functions (`configuration-state.ts`) doesn't have access to any orgId, I decided to have it fallback on the defaults provided by `getSampleEngineConfiguration`.

### DIFF
--- a/packages/headless/src/api/platform-client.ts
+++ b/packages/headless/src/api/platform-client.ts
@@ -4,7 +4,7 @@ import fetch from '@coveo/please-give-me-fetch';
 import {backOff} from 'exponential-backoff';
 import {Logger} from 'pino';
 import {DisconnectedError, ExpiredTokenError} from '../utils/errors';
-import {PlatformCombination, PlatformEnvironment} from '../utils/url-utils';
+import {PlatformEnvironment} from '../utils/url-utils';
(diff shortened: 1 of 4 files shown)


olamothe:
Those were exported publicly, right ?

I wanna emphasize, we * cannot * merge anything breaking in master right now. It will need to be done only in a couple months (in a specific V3 branch), not yet.

dmbrooke:
> I wanna emphasize, we * cannot * merge anything breaking in master right now. It will need to be done only in a couple months (in a specific V3 branch), not yet.

Woof, my bad. I made the bad assumption, thinking that these were safe to remove since they are minimally used in the code base and have been marked as deprecated for over 3 years (according to git lense) 🤦 I'll leave the branch as is and we can come back to it in a couple of months.

PR #3309, opened 2023-10-19
test(headless-ssr): split search parameter tests in both 'generic' and 'react' route

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2835

Could also make two separate files but I think it is fine this way. The tests are pretty fast.

image

### DIFF
--- a/packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
+++ b/packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
@@ -6,7 +6,7 @@ import {
 } from './utils';
 
 const searchStateKey = 'search-state';
-const routes = ['generic', 'react'];
+const routes = ['generic', 'react'] as const;
 
 routes.forEach((route) => {
   describe(`${route} Headless ssr with search parameter manager example`, () => {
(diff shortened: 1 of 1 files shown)


packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
```diff
@@ -8,7 +8,17 @@ import {
 const searchStateKey = 'search-state';
 
 describe('headless ssr with search parameter manager example', () => {
-  const route = '/generic';
+  const generic = '/generic';
+  const react = '/react';
+  describe(`when using the ${generic} route`, () => {
```
mrrajamanickam-coveo, line 13: Would be better to use `.each()` on the routes similar to the smoke tests.

mrrajamanickam-coveo (changes requested):

packages/samples/headless-ssr/cypress/e2e/ssr-search-parameter-manager.cy.ts
alexprudhomme: Yea, I saw the smoke tests were different just after this. Ill adjust !

mrrajamanickam-coveo (approved):

PR #3311, opened 2023-10-19
refactor: remove AnalyticsType

louis-bompart:


louis-bompart:
Need to do some manual tests, but looks like we can trim some fat

Spuffynism (approved):

olamothe (approved):
For info: 

This type was not exported, and was "only" serving as a way to do compile time check for implementers using TypeScript.

ie: It was there to verify that when an implementer calls `executeSearch(someAnalyticsActions())`, `someAnalyticsActions()` must be a `search` analytics event, and not a `click`, or `custom` event.

This will no longer be necessary in the future, and was not particularly useful overall, for the amount of complexity it brought.

Moreover, it was not leveraged properly by other use cases.

PR #3313, opened 2023-10-20
feat(atomic): change genai loading animation

olamothe:
https://github.com/coveo/ui-kit/assets/1591893/b5c3fc66-7e81-4010-b303-029f73f36378


packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -276,20 +273,48 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
+      return this.renderContent();
+    }
+  }
+
   public render() {
     const {isLoading} = this.generatedAnswerState;
-    if (this.shouldBeHidden) {
-      return null;
-    }
+    isLoading
```
olamothe, line 304: This "synchronize" removing the result list placeholder (and thus rendering results) when we know there's an answer.

This means we can remove some initial flicker by making sure we display or hide skeletons for both genai + results at the same time.

packages/atomic/src/components/search/atomic-generated-answer/placeholders.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <div class={`${classesAnim} w-full`}></div>
+      <div class={`${classesAnim} w-3/4`}></div>
+      <div class={`${classesAnim} w-3/4`}></div>
+      <div class={`${classesAnim} w-full`}></div>
+      <div class={`${classesAnim} w-1/3`}></div>
+      <div class={`${classesAnim} w-full`}></div>
+      <div class={`${classesAnim} w-1/3`}></div>
+      <div class={`${classesAnim} w-full`}></div>
+      <div class={`${classesAnim} w-3/4`}></div>
+    </div>
```
alexprudhomme, line 41: The animation for the facet values placeholders use random values instead of being hardcoded. Should we do the same here ? https://github.com/coveo/ui-kit/blob/master/packages/atomic/src/components/common/facets/facet-placeholder/facet-placeholder.tsx

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -276,20 +273,48 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
+
     return (
-      <aside
-        class={`mx-auto ${
-          isLoading ? this.loadingClasses : this.contentClasses
-        }`}
-        part="container"
+      <div
+        slot="body"
+        part="wrap_er"
```
alexprudhomme, line 311: Is 'wrap_er' a typo ?

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -276,20 +273,48 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
+          </div>
+          <div class="typing-loader-wrapper absolute inset-x-0 bottom-0 w-full z-1 pt-40">
+            <TypingLoader />
+          </div>
+        </div>
+      );
+    } else {
+      return this.renderContent();
+    }
+  }
```
alexprudhomme, line 300: Could we also add a check in this function for `this.isAnswerVisible` ? In my opinion, the placeholders should not show when toggling off.

https://github.com/coveo/ui-kit/assets/78121423/8219d0c2-a005-43d2-8384-bad991328b29

packages/atomic/src/components/search/atomic-generated-answer/placeholders.tsx
olamothe: Ehhh, not a bad idea, but I'm not sure it would be ... "better" ? I'm not sure, it's kind of subjective I guess :D

packages/atomic/src/pages/examples/genqa.html
```diff
@@ -15,10 +15,11 @@
...
         // Initialization
         await searchInterface.initialize({
-          accessToken: 'xx564559b1-0045-48e1-953c-3addd1ee4457',
-          organizationId: 'searchuisamples',
+          accessToken: '[...]',
+          organizationId: 'docsgenaidevvbmdzqdr',
+          platformUrl: 'https://platformdev.cloud.coveo.com',
           search: {
-            pipeline: 'genqatest',
+            pipeline: 'nlafrance-genqa-testing',
```
nathanlb, line 22: I'm famous!

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.pcss
```diff
@@ -17,6 +17,14 @@
   background: var(--atomic-primary-background);
 }
 
+.placeholder-wrapper {
+  height: 400px;
```
nathanlb, line 21: Maybe we'd want this to be overridable by an implementer. 400px might be too much for a widget like IPX.

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
louis-bompart: @alexprudhomme please takeover this PR and do this :)

packages/atomic/src/components/search/atomic-generated-answer/placeholders.tsx
louis-bompart: @alexprudhomme, scope-out this suggestion in your takeover. (If we're to do it I'd like @johnnyCoveo to review the result, but we don't have time to spare for that, tho @nathanlb if you think @alexprudhomme suggestion is nice, please feel free to pick it up ;) )

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.pcss
louis-bompart: The answer tho might still be 400px tall, and so the issue might repeat itself I reckon.
@nathanlb, I propose that we see if this is actually a problem on the field before adding an option.
Thought?

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
alexprudhomme: this has been added

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.pcss
johnnyCoveo: Looking at the demo, when the answer goes beyond the defined height, it still waits to generate completely before pushing everything at once. I tend to agree with @nathanlb and want it to be overridable by implementer because 400px is very big

nathanlb: exposed it so it can be overridable via css

louis-bompart (approved):

nathanlb:
This suggestion brought up some interesting discussions but we've decided not to incorporate the change at this time.

PR #3314, opened 2023-10-20
test(headless-ssr): add specific web-vitals thresholds for pages router in dev mode

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2834


packages/samples/headless-ssr/cypress/e2e/smoke.cy.ts
```diff
@@ -14,39 +14,60 @@ const timestampSelector = '#timestamp';
...
+    ? {
+        generic: {
+          thresholds: {
+            fcp: 2000,
+            lcp: 2000,
+            cls: 0,
+            ttfb: 60,
+            // TODO: Ensure validity of following input based vitals with interactive elements
+            fid: 400,
+            inp: 400,
```
alexprudhomme, line 32: @mrrajamanickam-coveo  What exactly is this todo ? Do we need to create another issue where we add interactive elements ?

mrrajamanickam-coveo: I think that has been addressed because I see we use `firstInputSelector` in web-vitals test. The TODO can be removed.

packages/samples/headless-ssr/app-router/package.json
```diff
@@ -10,7 +10,7 @@
     "prod": "next start",
     "lint": "next lint",
     "e2e": "cypress run --browser chrome --e2e --project ../",
-    "e2e:watch": "cypress open --browser chrome --e2e --project ../"
+    "e2e:watch": "cypress open --browser chrome --e2e --project ../ --env NEXTJS_ROUTER=app"
```
mrrajamanickam-coveo, line 13: guess this is not used in the tests - but can leave it for consistency and any future requirement

packages/samples/headless-ssr/cypress/e2e/smoke.cy.ts
```diff
@@ -14,39 +14,60 @@ const timestampSelector = '#timestamp';
...
-      // TODO: Ensure validity of following input based vitals with interactive elements
-      fid: 800,
-      inp: 800,
-    },
-  },
-};
+const vitals: Record<(typeof routes)[number], Cypress.ReportWebVitalsConfig> =
+  nodeEnv === 'development' && nextJSRouter === 'pages'
+    ? {
+        generic: {
```
mrrajamanickam-coveo, line 24: What do you think about extending/spreading from a common thresholds config and overriding just the values that differ? Guess it is simpler this way?

packages/samples/headless-ssr/cypress/e2e/smoke.cy.ts
```diff
@@ -14,39 +14,60 @@ const timestampSelector = '#timestamp';
...
-      cls: 0,
-      ttfb: 120,
-      // TODO: Ensure validity of following input based vitals with interactive elements
-      fid: 800,
-      inp: 800,
-    },
-  },
-};
+const vitals: Record<(typeof routes)[number], Cypress.ReportWebVitalsConfig> =
+  nodeEnv === 'development' && nextJSRouter === 'pages'
```
mrrajamanickam-coveo, line 22: From the test failures (in CI) looks like this is not working as expected?

alexprudhomme: yes looks like there is a problem,. Im pretty sure it was working locally !

mrrajamanickam-coveo: Wonder if process.env is not working as expected in CI or if next js or something else is setting NODE_ENV to a different value based on the CI env.

alexprudhomme: Just a silly mistake on my part forgetting to put it in e2e instead of e2e:watch !

packages/samples/headless-ssr/cypress/e2e/smoke.cy.ts
alexprudhomme: I just updated it ! Way cleaner

fbeaudoincoveo (approved):

PR #3317, opened 2023-10-20
feat(commerce): add basic commerce category facets

fbeaudoincoveo:


packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
```diff
@@ -359,7 +360,9 @@ export function buildCoreCategoryFacet(
 
     isSortedBy(criterion: CategoryFacetSortCriterion) {
       const request = getRequest();
-      return request!.sortCriteria === criterion;
+      return 'sortCriteria' in request && request.sortCriteria === criterion
```
fbeaudoincoveo, line 363: There are several things that make me very unhappy with using the core category facet controller as it is:

1. Those little "exceptions" with the `sortCriteria`
2. The fact that the core controller dispatches a registerCategoryFacet action, which is unnecessary in the commerce use case as facets will be pre-configured (i.e., we're not making facet "request", we're just sending the current facet state).
3. The fact that intializing the core controller in the commerce context means that we're loading up quite a bit of dead code: we don't need anything that relates to sorting, enabling, or disabling the facet, and we also don't need the search reducer (and probably not the facetOptions reducer either)...
4. The core controller generates "full" facet requests, so we're sending a bunch of extra params to the listing API.

I see two potential solutions:

a) Create a new core category facet controller that contains the bare minimum code which applies to all use cases (search, insight, old product listing, and commerce).

b) Avoid using a core category facet controller in commerce altogether, and instead create a brand new controller from scratch.

Solution A implies some code refactoring, and possibly an extra layer of abstraction if we want to avoid code duplication among non-commerce category facets.

E.g., we could end up with something like:

Core category facet controller
  Mantle category facet controller
    Category facet controller
    Insight category facet controller
  Commerce category facet controller

(I'm using "core" and "mantle" as an analogy just to make it clear that "mantle" is one layer higher than "core", but still abstract)

I personally don't like the idea of adding an extra layer of abstraction, though; this feels like overengineering. On the other hand, if we don't add that extra layer, we'll end up with code duplication between 3 controllers which I'm even less a fan of.

Solution B implies a fair bit of code duplication. However, I'm not totally against it, because: 1) the alternative [solution A] is probably worse, and 2) it feels slightly awkward to use a common core for features that are powered by distinct APIs, that could potentially evolve in different directions.

@Spuffynism , @olamothe , thoughts?

fbeaudoincoveo: (btw I've stashed a change set where I'm actually implementing solution B)

packages/headless/src/features/facets/facet-set/facet-set-selectors.ts
```diff
@@ -1,21 +1,31 @@
...
 
 export const baseFacetResponseSelector = (
   state: Partial<FacetResponseSection>,
   id: string
 ) => {
   if ('productListing' in state && state.productListing) {
-    return state.productListing.facets.results.find(
-      (response) => response.facetId === id
-    );
+    if ('results' in state.productListing.facets) {
```
fbeaudoincoveo, line 20: This is another example of something that bugs me with using the core controller

packages/headless/src/controllers/core/facets/category-facet/headless-core-category-facet.ts
Spuffynism: My feeling is that we should do B until it becomes annoying/painful, and then do A.

PR #3318, opened 2023-10-20
docs(headless): improve automatic facet doc to reflect what should be used

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2838

The docs are not great and even my solution is not the best. The best would be to actually remove the documentation around the `buildAutomaticFacet` and the `AutomaticFacetProps`. The reason why is because they are not supposed to be used by clients, only internally by the generator. Unfortunately, we cant remove them from the docs because they are needed for the rest of the docs to generate... 

To properly solve this, we would need to refactor the doc-parser to enable _ignoring_ certain function/interfaces.


packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.test.ts
```diff
@@ -1,9 +1,13 @@
...
+} from '../../../controllers/facets/automatic-facet/headless-automatic-facet';
 import {
   deselectAllAutomaticFacetValues,
   toggleSelectAutomaticFacetValue,
 } from '../../../features/facets/automatic-facet-set/automatic-facet-set-actions';
 import {executeSearch} from '../../../features/search/search-actions';
+import {SearchAppState} from '../../../state/search-app-state';
 import {
   MockSearchEngine,
   buildMockSearchAppEngine,
```
alexprudhomme, line 13: Just something random I came across where we were importing from the index.ts

olamothe (approved):

packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
```diff
@@ -40,14 +40,14 @@ export interface AutomaticFacetProps {
 }
 /**
  * The `AutomaticFacet` controller allows you to create a search interface component that the end user
- * can use to refine a query by selecting filters based on item metadata (i.e., field values). Unlike traditional facets that
+ * can use to refine a query by selecting filters based on item metadata (i.e., field values). Unlike regular facets that
  * need to be explicitly defined and requested in the query, automatic facets are dynamically generated by the index
  * in response to the query.
```
fbeaudoincoveo, line 45: The `AutomaticFacet` controller is used internally by the `AutomaticFacetGenerator` controller to manage the individual state of each dynamically generated facet.

To learn more about the automatic facet generator feature, see [About the Facet Generator](https://docs.coveo.com/en/n9sd0159/).

**Important:** This controller is meant for internal use by Headless only. As an implementer, you should never import or use this controller directly in your code.

packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
```diff
@@ -66,10 +66,10 @@ export interface AutomaticFacet extends Controller {
   state: AutomaticFacetState;
 }
 /**
- * Creates a `AutomaticFacet` controller instance.
+ * This initializer is used internally by the AutomaticFacetGenerator controller.
  *
  * @param engine - The headless engine.
- * @param props - The configurable `AutomaticFacet` properties.
+ * @param props - The configurable `AutomaticFacet` properties used internally.
  * @returns A `AutomaticFacet` controller instance.
```
fbeaudoincoveo, line 73:
```suggestion
 * @returns An `AutomaticFacet` controller instance.
```

packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
```diff
@@ -66,10 +66,10 @@ export interface AutomaticFacet extends Controller {
   state: AutomaticFacetState;
 }
 /**
- * Creates a `AutomaticFacet` controller instance.
+ * This initializer is used internally by the AutomaticFacetGenerator controller.
```
fbeaudoincoveo, line 69:
```suggestion
 * This initializer is used internally by the `AutomaticFacetGenerator` controller.
 *
 * **Important:** This initializer is meant for internal use by Headless only. As an implementer, you should never import or use this initializer directly in your code.
```

packages/samples/headless-react/src/components/automatic-facet/automatic-facet.fn.tsx
```diff
@@ -28,3 +28,9 @@ export const AutomaticFacet: FunctionComponent<AutomaticFacetProps> = (
     </ul>
   );
 };
+
```
fbeaudoincoveo, line 31: Might as well just remove the sample if this is meant for internal use only.

fbeaudoincoveo (changes requested):
I left some suggestions to rephrase the documentation, but why do you even need to export AtomaticFacet, AutomaticFacetProps, and AutomaticFacetState in index.ts if this controller is 100% internal?

If you don't export it, you could just remove the docs.

If the controller was ALWAYS intended to be internal and should realisitically have been used in no implementations, then removing the export would not be a breaking change IMO.

alexprudhomme:
@fbeaudoincoveo I made some changes, we now only export the AutomaticFacet and AutomaticFacetState interfaces that are important for implementers. They are now exported inside of the Generator page directly. There wont be a doc page for AutomaticFacet, only one for the Generator !

srosenberg-coveo (approved):

packages/headless/src/controllers/facets/automatic-facet-generator/headless-automatic-facet-generator.ts
```diff
@@ -53,6 +52,48 @@ export interface AutomaticFacetGeneratorProps {
...
+   */
+  label: string;
+  /**
+   * The automatic facet values.
+   */
+  values: FacetValue[];
+}
+/**
+ * The `AutomaticFacet` controller allows you to create a search interface component that the end user
+ * can use to refine a query by selecting filters based on item metadata (i.e., field values). Unlike regular facets that
```
fbeaudoincoveo, line 74:
```suggestion
 * can interact with to refine a query by selecting filters based on item metadata (i.e., field values). Unlike regular facets that
```

pet peeve, but I just don't like "that the end _user_ can _use_" 😅

packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
```diff
@@ -7,70 +7,29 @@ import {logFacetClearAll} from '../../../features/facets/facet-set/facet-set-ana
...
-   */
-  label: string;
-  /**
-   * The automatic facet values.
-   */
-  values: FacetValue[];
-}
-/**
- * The props for the `AutomaticFacet` controller.
+ * @internal
```
fbeaudoincoveo, line 14: If the props / builder aren't exported, do we even need to document them?

fbeaudoincoveo (approved):

packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts
alexprudhomme: I think its fine to leave the comments.

PR #3326, opened 2023-10-23
feat(atomic): mark searchbox as input for deprecation in favour of textarea

olamothe:
For next major version, we would like to only support one mode (textarea, which allows input to expand).

Add warnings for future deprecation notice.

https://coveord.atlassian.net/browse/KIT-2688


louis-bompart (approved):

dmbrooke (approved):
Looks good!

PR #3328, opened 2023-10-23
fix(github): make the Create Quantic Package workflow not fail when there are no changes to the code

SimonMilord:
[SFINT-5245](https://coveord.atlassian.net/browse/SFINT-5245)

We want the Create Quantic Package workflow to send a success message in Slack when the workflow is cancelled due to no changes in the code.

[SFINT-5245]: https://coveord.atlassian.net/browse/SFINT-5245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

The update of the github-script action created a breaking change.
https://github.com/actions/github-script#breaking-changes-in-v5

The actions are no longer directly exposed on the github object but now on `github.rest`

image


.github/workflows/create-quantic-package.yml
```diff
@@ -60,9 +60,15 @@ jobs:
...
+      - name: Check SHA
+        id: check_sha
+        uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3
+        with:
+          path: check-SHA
+          key: check-SHA-${{ github.sha }}
       - run: |
           url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
-          if [[ ${{ contains(needs.*.result, 'failure') }} == "false" ]]; then
+          if [[ ${{ contains(needs.*.result, 'failure') }} == "false" || ${{ steps.check_sha.outputs.cache-hit }} == "true" ]]; then
```
SimonMilord, line 71: Here we are checking if the result of the run is not a failure OR if there are no changes to the code.

erocheleau:
I reverted the proposed changes by Simon and updated instead to the correct call now to the `github.rest.actions`

mmitiche (approved):

PR #3330, opened 2023-10-23
docs(headless-react): add note about alpha release

mrrajamanickam-coveo:
- add note about alpha release for the new headless-react SSR package
Also
- add some build dirs to prettier ignore which it was spending time going through
- remove some redundant `npx` from pjson commands

https://coveord.atlassian.net/browse/KIT-2843


louis-bompart (approved):

packages/headless-react/README.md
```diff
@@ -1,5 +1,8 @@
 # Headless React Utils for SSR
 
+**:warning: Alpha release**
+This is an alpha release with several features such as analytics, URL management (search params) etc yet to be implemented.
```
dmbrooke, line 4: Don't feel too strongly, but I'm not a fan of using `etc` in a sentence. Moreover, I find it a bit odd to use `URL management` just to clarify it as `search params` immediately after, so I think we should use on or the other.
```suggestion
This is an alpha release with several features yet to be implemented (e.g., analytics, URL management, etc).
```

dmbrooke (approved):

packages/headless-react/README.md
mrrajamanickam-coveo: Sounds good 👍 
Yes was not sure about using both URL management and search params either - but wasn't sure if it would be clear for most people reading this (esp those with less coveo experience) that one is synonymous with the other. Earlier I found it hard to understand what either of them meant on their own - but found that together they made more sense.

mrrajamanickam-coveo: @dmbrooke How about just saying `[Synchronize search parameters with the URL](https://docs.coveo.com/en/headless/latest/usage/synchronize-search-parameters-with-the-url/)` reusing the title of from the official docs? It is clear what is being referred and we can even link to the official doc?

PR #3331, opened 2023-10-24
chore: switch all dependencies to use fixed versions 

mrrajamanickam-coveo:
https://coveord.atlassian.net/browse/KIT-2430


mrrajamanickam-coveo:
Even just switching to fixed versions results in build failures in multiple packages 🤷 
```
Failed tasks:
   
   - atomic-angular-builder:cached:build
   - atomic-storybook:build:storybook:assets-list
   - atomic-vuejs-samples:cached:build
   - atomic-react:cached:build
```
Needs further investigation - closing for now.

PR #3332, opened 2023-10-25
feat(commerce): add facet generator and basic facet controller for plp v2

Spuffynism:
This PR creates 2 controllers,1 new reducer with its state interface:
- A controller for creating facet controllers from the list of received facets: `FacetGenerator`
- A controller for creating  the facet controller: `Facet`
- A reducer + state to populate the state from a product listings response.

I've isolated the state in its own reducer to allow for interoperability with other solution types (aka Search), similarly to how we've torn the pagination and sorting states apart from product listings.

[CAPI-87]

[CAPI-87]: https://coveord.atlassian.net/browse/CAPI-87?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/product-listing/facets/headless-product-listing-facet-generator.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export interface FacetGenerator extends Controller {
+  state: FacetGeneratorState;
+}
+
+export interface FacetGeneratorState {
+  facets: Facet[];
+}
+
+export function buildFacetGenerator(engine: CommerceEngine): FacetGenerator {
```
Spuffynism, line 25: This controller is very very similar to the automatic facets controller: https://github.com/coveo/ui-kit/blob/4decff2d91f415d392177b94dfd10e4c3f8566e8/packages/headless/src/controllers/facets/automatic-facet/headless-automatic-facet.ts#L75

packages/headless/src/features/commerce/facets/facets-state.ts
```diff
@@ -0,0 +1,9 @@
+import {AnyFacetResponse} from '../../facets/generic/interfaces/generic-facet-response';
+
+export interface CommerceFacetsState {
```
Spuffynism, line 3: This is meant to be compatible with the commerce search solution type, when we add it. We'll just need to add a case for handling the action that fulfills a commerce search in the reducer above.

packages/headless/src/features/facets/facet-set/facet-set-selectors.ts
```diff
@@ -77,5 +87,9 @@ export const isFacetLoadingResponseSelector = (state: FacetResponseSection) => {
     return state.productListing.isLoading;
   }
 
+  if ('commerceFacets' in state) {
+    return false;
+  }
```
Spuffynism, line 92: I'm not too sure how we should handle this case. We don't have access to the `isLoading` state on the `CommerceFacetSection`. Should we change the state to something like this?
```typescript
export type FacetResponseSection =
  | SearchSection
  | ProductListingSection
  | (CommerceFacetSection & ProductListingV2Section);
  ```
  which would eventually become:
  ```typescript
export type FacetResponseSection =
  | SearchSection
  | ProductListingSection
  | (CommerceFacetSection & (ProductListingV2Section | CommerceSearchSection));
  ```

**Review by olamothe — APPROVED** [review]
My comment about rethinking core-controller pattern is something we can postpone (not absolutely necessary short term), but for the long term health of the project, it will be necessary !

File: packages/headless/src/features/facets/facet-set/facet-set-selectors.ts (hunk 3 above)
Comment (fbeaudoincoveo) [reply packages/headless/src/features/facets/facet-set/facet-set-selectors.ts:92]: Perhaps for now, yes, but this makes me reconsider the way we're structuring the commerce app state.

I'll come up with something and show it to you later today 😄

File: packages/headless/src/controllers/commerce/product-listing/facets/headless-product-listing-facet-generator.ts [hunk 4]
```diff
@@ -0,0 +1,57 @@
...
+        ),
+      };
+    },
+  };
+}
+
+function loadFacetGeneratorReducers(
+  engine: CommerceEngine
+): engine is CommerceEngine<
+  FacetSection & FacetOptionsSection & CommerceFacetSection
```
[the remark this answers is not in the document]
Spuffynism: The core facet controller (which is used by the commerce facet controller) checks that facets are enabled:
https://github.com/coveo/ui-kit/blob/03432f47dd7e3b5d601b3d14341e9eb3efafb9e0/packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts#L365
This check is done by looking at the facet options slice: 
https://github.com/coveo/ui-kit/blob/03432f47dd7e3b5d601b3d14341e9eb3efafb9e0/packages/headless/src/features/facet-options/facet-options-selectors.ts#L7

Spuffynism:
Superseded by https://github.com/coveo/ui-kit/pull/3390 and https://github.com/coveo/ui-kit/pull/3389

PR #3334, opened 2023-10-25
chore: use release branch(es) as tags

louis-bompart:
So, some fix/simplification to the Jenkins shenanigans.
To help Jenkins not get crazy with our thousand tags (indeed, in the long run, Jenkins might choke), I created a little special something: `release/*` branch(es).
This branch is protected and can only be touched by our robot and so are meant to be the absolute source of truth of what should be released.
Pushing on this branch triggers the release in Jenkins (no more logic in the Jenkinsfile, if your branch ain't release/*, Jenkins will not even start a build on it.)

Essentially, now, with that, Jenkins will only check the latest commit of our release branch (yep, no history, the rollout certifier just needs the sha to 👁️ , it can handle itself afterward)

A few boons:
- For v2, we will indeed only have to do a v2 branch on the last v2 commit and the release process will be able to continue smoothly.
- We will have a pointer on the 'latest release v2' even after we moved to v3.
- KISS. IMO, this is simpler.

Jenkins part is tested through replay.
GHA, will be next release.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,6 +44,7 @@ node('heavy && linux && docker') {
         --resolve ATOMIC_HOSTED_PAGE_MAJOR_VERSION=${atomicHostedPageMajor} \
         --resolve ATOMIC_HOSTED_PAGE_MINOR_VERSION=${atomicHostedPageMinor} \
         --resolve ATOMIC_HOSTED_PAGE_PATCH_VERSION=${atomicHostedPagePatch} \
+        --changeset ${env.GIT_COMMIT}
         || true"
       }
     }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -6,17 +6,7 @@ def parseSemanticVersion(String version) {
...
-  def isOnReleaseBranch = env.BRANCH_NAME == 'master'
-
-  if (!isOnReleaseBranch) {
-    return
-  }
-
-  if (!isBump) {
-    return
-  }
+  def releaseCommit
```
mrrajamanickam-coveo, line 9: can't the sh git command be run here after checkout?

mrrajamanickam-coveo (approved):

Jenkinsfile
louis-bompart: honestly I don't know.
I think we need to run it in a stage.
I kinda duplicated what worked in the cli repo until a build went through to be honest 😅

dmbrooke (approved):

PR #3336, opened 2023-10-26
test(cypress): wait for resources in cypress tests

dmbrooke:
https://coveord.atlassian.net/browse/KIT-2848

Couldn't recreate this issue locally. Researched many best practices for this type of problem and pretty much everywhere I looked suggested that this was a matter of sequencing (e.g., we should be checking the component at every level rather than checking most of the component, and then jumping straight to an icon that's nested a couple of levels down).

I wanted to avoid rewriting a bunch of tests, so I instead decided to instead setup an intercept for any resources requests and make cypress wait for the requests to resolve before searching for the shadow element.

Note that I have to rely on a regex for because resources are loaded through webpack as JS files with hashes appended to the file name. Unfortunately, this means we end up waiting for some extra non-icon-related stuff to load before executing **_some_** tests. Thankfully, the amount of time it takes to wait for everything to load is quite minor and I suspect it will have an impact on the execution time of only a few seconds across our entire test-suite. Will closely compare these test executions with main and other branches to get an idea as to whether the impact is noticeable.

### DIFF
--- a/packages/atomic/cypress/fixtures/fixture-common.ts
+++ b/packages/atomic/cypress/fixtures/fixture-common.ts
@@ -29,6 +29,7 @@ export const RouteAlias = {
   Quickview: '@coveoQuickview',
   Locale: '@locale',
   GenQAStream: '@genQAStream',
+  Build: '@build',
 };
 
 export const ConsoleAliases = {
@@ -80,8 +81,8 @@ export function setupIntercept() {
 
   cy.intercept({
     method: 'GET',
-    path: '/build/lang/**.json',
-  }).as(RouteAlias.Locale.substring(1));
+    url: /.*\/build\/[\w|-]+-[\d|\w]+\.js/,
+  }).as(RouteAlias.Build.substring(1));
 }
 
 export function getUABody() {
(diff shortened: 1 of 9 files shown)


packages/atomic/cypress/fixtures/fixture-common.ts
```diff
@@ -80,8 +81,8 @@ export function setupIntercept() {
 
   cy.intercept({
     method: 'GET',
-    path: '/build/lang/**.json',
-  }).as(RouteAlias.Locale.substring(1));
```
louis-bompart, line 84: Is this unused?

louis-bompart (commented):
Solution seems fine to me, the intercept is smart.
Did you check tho that the RouteAlias.Locale was unused before removing its intercept?

packages/atomic/cypress/fixtures/fixture-common.ts
louis-bompart: Ain't

louis-bompart (approved):

packages/atomic/cypress/fixtures/fixture-common.ts
louis-bompart: Well no it is

PR #3337, opened 2023-10-26
feat: support initialState in context controller

y-lakhdar:
I have introduced an optional parameter in the `buildContext()` function, enabling the specification of an initial state. Instead of configuring the state in this manner:

```javascript
context = buildContext(engine);
context.set(someState)
```

Users can now conveniently employ the following format:

```javascript
context = buildContext(engine, {initialState: someState});
```

This feature simplifies the process of defining the initial context when utilizing headless SSR, as demonstrated in the provided samples.

KIT-2764

### DIFF
--- a/packages/headless/src/controllers/core/context/headless-core-context.ts
+++ b/packages/headless/src/controllers/core/context/headless-core-context.ts
@@ -21,10 +21,17 @@ import {
 export type {ContextPayload, ContextValue};
 
 export interface ContextProps {
+  /**
+   * Represents the initial state of the context.
+   * If provided, it should adhere to the structure defined by the `ContextInitialState` interface.
+   */
   initialState?: ContextInitialState;
 }
 
 export interface ContextInitialState {
+  /**
+   * Represents the initial key/value pair of the context.
+   */
   values: ContextPayload;
 }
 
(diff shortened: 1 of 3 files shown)


packages/headless/src/controllers/core/context/headless-core-context.ts
```diff
@@ -11,13 +12,28 @@ import {
...
 } from '../../controller/headless-controller';
 
 export type {ContextPayload, ContextValue};
 
+export interface ContextProps {
+  initialState?: ContextInitialState;
+}
+
+export interface ContextInitialState {
+  values: ContextPayload;
```
olamothe, line 28: `values` is pretty generic for a name :D 

But then, I dont have any good suggestion for an alternative name, so 🤷

packages/headless/src/controllers/core/context/headless-core-context.ts
```diff
@@ -66,6 +85,17 @@ export function buildCoreContext(engine: CoreEngine): Context {
   const {dispatch} = engine;
   const getState = () => engine.state;
 
+  const initialState = validateInitialState(
```
olamothe, line 88: We should add some tests for this behaviour in `headless-core-context-test`: That is: checking that creating the controller with an initial state correctly dispatch the `setContext `action

olamothe (approved):
You will need to make sure all new types and interface that get resurfaced in the public API are documented (JSDoc): This is what the build is complaining about.

PR #3339, opened 2023-10-26
feat(atomic): added negative feedback modal for rga

drosenfeld-coveo:
[SVCC-3225](https://coveord.atlassian.net/browse/SVCC-3225)

Added modal to give feedback if user is unhappy with the rga result.

[Design](https://www.figma.com/file/l63i3ZXWrVxldOR4WUBKS1/RGA-Component?type=design&node-id=2512-12223&mode=design&t=dMdAsw2NUQbzjyAN-0)

Screenshot 2023-10-26 at 2 26 16 PM

Screenshot 2023-10-26 at 2 26 59 PM

Screenshot 2023-10-26 at 2 27 33 PM

[SVCC-3225]: https://coveord.atlassian.net/browse/SVCC-3225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmartel-coveo:
@drosenfeld-coveo Is there a reason why the freetext box doesn't have a label ("Details" [in the copydeck](https://coveord.atlassian.net/wiki/spaces/UXC/pages/3605168284/Search+page#User-feedback-form-done))?

drosenfeld-coveo:
> @drosenfeld-coveo Is there a reason why the freetext box doesn't have a label ("Details" [in the copydeck](https://coveord.atlassian.net/wiki/spaces/UXC/pages/3605168284/Search+page#User-feedback-form-done))?

ahh didnt know about the copydeck, I was going off the designs. Ill update it

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.pcss
```diff
@@ -0,0 +1,26 @@
...
+  @screen desktop-only {
+    @apply mt-6;
+  }
+
+  @screen mobile-only {
+    @apply mt-4;
+  }
+}
+
+.feedbackHeader {
```
olamothe, line 15: Is this class used anywhere ?

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.pcss
```diff
@@ -0,0 +1,26 @@
...
+  @apply p-8;
+}
+
+[part='form'] {
+  @apply text-base leading-4 text-neutral-dark p-2 text-xl;
```
olamothe, line 20: What's the reason to use `@apply` here, as opposed to just setting them as class directly in JSX ?

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.pcss
```diff
@@ -0,0 +1,26 @@
...
+}
+
+[part='cancel-button'],
+[part='submit-button'] {
+  @apply flex justify-center text-sm leading-4 p-2;
```
olamothe, line 25: Same comment as above regarding using class directly

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
+import {GeneratedAnswer} from '@coveo/headless';
+import {GeneratedAnswerFeedback} from '@coveo/headless/dist/definitions/features/generated-answer/generated-answer-analytics-actions';
```
olamothe, line 2: This type of import is not correct, and will break.

The interface needs to be exported properly from Headless entry point if it's necessary for implementers to reference it.

packages/atomic/src/locales.json
nathanlb, line 1: You're going to have to run the [translation generation script](https://github.com/coveo/ui-kit/tree/master/scripts/translation-gpt) for these additions.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  private renderHeader() {
+    return (
+      <div
+        slot="header"
+        part="modalHeader"
+        class="w-full p-2 flex justify-between items-center"
+      >
+        <h1>{this.bindings.i18n.t('generated-answer-feedback-header')}</h1>
+        <Button
```
olamothe, line 91: You could use the `IconButton` functional component here.

As well, this is missing some aria property that would make this accessible (aria label that describes the action of the button)

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+            'generated-answer-feedback-textarea-add-details'
+          )}
+          required
+        ></textarea>
+      </fieldset>
+    );
+  }
+
+  private renderBody() {
+    console.log('feedbackSubmitted', this.feedbackSubmitted);
```
olamothe, line 191: Remove

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -282,14 +290,23 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
-      <aside
-        class={`mx-auto ${
-          isLoading ? this.loadingClasses : this.contentClasses
-        }`}
-        part="container"
-      >
-        <article>{isLoading ? <TypingLoader /> : this.renderContent()}</article>
-      </aside>
+      <Host>
+        <atomic-generated-answer-feedback-modal
```
olamothe, line 294: By rendering the modal like this, inside the generated-answer component, I'm wondering if the modal-trap aspect is working correctly.

That is: When the modal is opened, and you use the keyboard to tab through all the elements in the modal, it should be "trapped" and not focus on elements outside of the modal, in the rest of the page.

Can you confirm this is working correctly ?

packages/atomic/src/locales.json
```diff
@@ -6535,6 +6535,36 @@
     "zh-CN": "哎呀！在生成答案时出错了。",
     "zh-TW": "糟糕！在生成答案時出了些問題。"
   },
+  "generated-answer-feedback-header": {
```
olamothe, line 6538: These keys should be about the text content as much as possible, and not necessarily where or why they are used.

ie: 

```suggestion
  "feedback": {
```

I know there are some other keys in this file which do not follow this rule, but 🤷

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.pcss
```diff
@@ -1,11 +1,26 @@
...
-@import '../../../global/mixins.pcss';
+@import '../../../../global/global.pcss';
+@import '../../../../global/mixins.pcss';
 @import './styles/generated-text.pcss';
 @import './styles/feedback-buttons.pcss';
 @import './styles/rephrase-buttons.pcss';
 @import './styles/typing-indicator.pcss';
 @import './styles/source-citations.pcss';
 
+[part='generated-text'] {
```
nathanlb, line 9: this is already included in the import `@import './styles/generated-text.pcss';`

packages/atomic/src/pages/examples/genqa.html
```diff
@@ -24,6 +24,11 @@
 
         // Trigger a first search
         searchInterface.executeFirstSearch();
+
+        const shadowroot = app - element.shadowRoot;
+
+        const modal = shadowroot.getElementById('modal');
+        modal.setAttribute('is-open', 'true');
```
olamothe, line 31: Is this an artefact of your testing, and should it be removed ?

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.pcss
```diff
@@ -17,6 +32,16 @@
   background: var(--atomic-primary-background);
 }
 
+atomic-modal {
+  width: 100%;
+}
+
+.source-citations {
```
nathanlb, line 39: Should also remove this. It's included in the imported `@import './styles/source-citations.pcss';`

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.pcss
```diff
@@ -17,6 +32,16 @@
   background: var(--atomic-primary-background);
 }
 
+atomic-modal {
```
nathanlb, line 35: IIRC the modal is injected outside of this component's shadow dom, no? Does this CSS actually have any effect?

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+      );
+    }
+    this.feedbackSent.emit();
+  }
+
+  private renderHeader() {
+    return (
+      <div
+        slot="header"
+        part="modalHeader"
```
nathanlb, line 87: I think we usually use snake-case for parts and classes

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.pcss
```diff
@@ -0,0 +1,26 @@
...
+    @apply mt-6;
+  }
+
+  @screen mobile-only {
+    @apply mt-4;
+  }
+}
+
+.feedbackHeader {
+  @apply p-8;
```
nathanlb, line 16: I don't see this class anywhere else in the repo. Are we sure this css does something? If so then snake-case is probably more appropriate and check if the `p-8` can't be applied directly instead of this.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.pcss
nathanlb, line 20: why can't these classes be applied directly in the markup?

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.pcss
nathanlb, line 25: same here. the point of tailwind is to keep the styling in the markup.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.pcss
```diff
@@ -0,0 +1,26 @@
...
+
+[part='reason'] {
+  @apply text-base leading-4 text-neutral-dark;
+
+  @screen desktop-only {
+    @apply mt-6;
+  }
+
+  @screen mobile-only {
+    @apply mt-4;
```
nathanlb, line 11: Things like this can also be applied in the markup directly using `mobile-only:mt-4` FYI

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+              part="submit-button"
+              style="primary"
+              type="submit"
+              form={this.formId}
+              disabled={!this.currentAnswer ? true : false}
+              onClick={() => {
+                this.feedbackSubmitted = true;
+              }}
+            >
+              {this.bindings.i18n.t('smart-snippet-feedback-send')}
```
nathanlb, line 239: smart-snippets

nathanlb: tbh I'm not sure if the ui-kit team would prefer we reuse smart-snippets strings where possible or if we duplicate them to keep things distinct.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+    );
+  }
+
+  public render() {
+    this.updateBreakpoints();
+
+    return (
+      <atomic-modal
+        fullscreen={false}
+        id="rga-feedback-modal"
```
nathanlb, line 263: I don't think we use the `rga` accronym anywhere. Do we even need an `id`?

packages/atomic/src/pages/examples/genqa.html
nathanlb, line 31: what's up with this? suspicious...

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.pcss
olamothe, line 39: Setting those variables like so seems like a test artefact ? Or is this really wanted ?

Shouldn't they be following other variables in use in Atomic, or rem at least ?

olamothe (changes requested):

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+import {
+  InitializableComponent,
+  InitializeBindings,
+} from '../../../../utils/initialization-utils';
+import {updateBreakpoints} from '../../../../utils/replace-breakpoint';
+import {once, randomID} from '../../../../utils/utils';
+import {Button} from '../../../common/button';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+
+@Component({
```
nathanlb, line 24: Absolutely needs to be marked as internal so it doesn't get documented before the GA. It's the case for all our RGA components.

```suggestion
/**
 * @internal
 */
@Component({
```

packages/atomic/cypress/e2e/generated-answer.cypress.ts
```diff
@@ -102,6 +105,45 @@ describe('Generated Answer Test Suites', () => {
       });
     });
 
+    describe('feedback modal', () => {
```
nathanlb, line 108: Good start but IMO we should also test that the proper analytics events are sent.

nathanlb (commented):
Good stuff so far 👍

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+              onClick={() => this.close()}
+            >
+              {this.bindings.i18n.t('cancel')}
+            </Button>
+            <Button
+              part="submit-button"
+              style="primary"
+              type="submit"
+              form={this.formId}
+              disabled={!this.currentAnswer ? true : false}
```
lbergeron, line 234:
```suggestion
              disabled={!this.currentAnswer}
```

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  private setIsOpen(isOpen: boolean) {
+    this.isOpen = isOpen;
+  }
+
+  private close() {
+    this.feedbackSubmitted = false;
+    this.setIsOpen(false);
+    this.generatedAnswer.closeFeedbackModal();
+  }
```
lbergeron, line 64: I guess closing the modal should clear `currentAnswer` and the details input box.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+              {this.bindings.i18n.t('cancel')}
+            </Button>
+            <Button
+              part="submit-button"
+              style="primary"
+              type="submit"
+              form={this.formId}
+              disabled={!this.currentAnswer ? true : false}
+              onClick={() => {
+                this.feedbackSubmitted = true;
```
lbergeron, line 236: Is the feedback modal intended to be single-use? In other words, it seems that once a user submits feedback, opening the modal again would bring back the "success" page, and not the feedback form.

I'm not sure how likely it is for a user to submit more than one feedback for the same generated answer, but I feel it would make more sense to reset the modal once feedback is submitted.

Thoughts?

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
drosenfeld-coveo: yep you can tab through the elements in the modal!

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
drosenfeld-coveo: Right now if the user submits, then re opens the modal, it starts over, letting them submit more feedback. I'm not exactly if it was supposed to be a single time only or not, @johnnyCoveo @nathanlb what do you think?

lbergeron: In my opinion, users are expecting the modal to start clean every time it opens. If that is the actual behavior, then it's all good. I probably missed something when reviewing, will check again.

lbergeron: I've done some testing and the modal state is currently not cleaned when the modal is closed. In other words, if you open the modal a second time, it remembers the option selected previously (even if feedback was submitted).

In my opinion, the modal should start clean every time it opens.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
lbergeron: Yes, one can tab through the modal elements, but the focus trap does not work as focus moves from the "Send feedback" to the "Like" button under the backdrop. When the modal is opened, the focus should cycle only on the modal elements.

lbergeron (changes requested):
Please fix the focus trap, keyboard navigation, accessibility, and component state clean up.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,275 @@
...
+        <legend part="reason-title" class="text-base">
+          {this.bindings.i18n.t('generated-answer-feedback-instructions')}
+        </legend>
+        {options.map(({id, localeKey, correspondingAnswer}) => (
+          <div
+            class={`flex items-center ${correspondingAnswer}`}
+            key={id}
+            part="reason"
+          >
+            <input
```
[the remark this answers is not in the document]
lbergeron: My bad, this one is working.

nathanlb (approved):
Nice work

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -221,7 +229,10 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
                     title={this.bindings.i18n.t('this-answer-was-not-helpful')}
                     variant="dislike"
                     active={this.generatedAnswerState.disliked}
-                    onClick={this.generatedAnswer.dislike}
+                    onClick={() => {
```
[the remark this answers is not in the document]
lbergeron: Out of curiosity, can you elaborate on why the suggested approach is better?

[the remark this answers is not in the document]
nathanlb: Yeah basically it and that it could maybe cause unnecessary rerenders if it is for a prop given the function instance is a new one on every render. I honestly don't know the real impact this has. I found it because in a personal project using stencil I'm using the recommended eslint rules and this one came up https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#es6-classes

packages/atomic/src/locales.json
```diff
@@ -6535,6 +6535,276 @@
...
+    "sv": "Berätta för oss varför det genererade svaret inte var till hjälp.",
+    "th": "บอกเราว่าทำไมคำตอบที่สร้างขึ้นไม่เป็นประโยชน์",
+    "tr": "Oluşturulan yanıtın neden yardımcı olmadığını bize söyleyin.",
+    "zh": "告诉我们生成的答案为什么没有帮助。",
+    "zh-CN": "告诉我们生成的答案为什么没有帮助。",
+    "zh-TW": "告訴我們生成的答案為什麼沒有幫助。"
+  },
+  "irrelevant": {
+    "en": "Irrelevant",
+    "fr": "Non pertinent",
```
lbergeron, line 6594:
```suggestion
    "fr": "Non pertinente",
```

packages/atomic/src/locales.json
```diff
@@ -6535,6 +6535,276 @@
...
+    "sv": "Irrelevant",
+    "th": "ไม่เกี่ยวข้อง",
+    "tr": "İlgisiz",
+    "zh": "不相關",
+    "zh-CN": "不相关",
+    "zh-TW": "不相關"
+  },
+  "not-accurate": {
+    "en": "Not accurate",
+    "fr": "Inexact",
```
lbergeron, line 6621:
```suggestion
    "fr": "Inexacte",
```

packages/atomic/src/locales.json
```diff
@@ -6535,6 +6535,276 @@
...
+    "sv": "Inaktuellt",
+    "th": "ล้าสมัย",
+    "tr": "Modası geçmiş",
+    "zh": "过时",
+    "zh-CN": "过时",
+    "zh-TW": "過時"
+  },
+  "harmful": {
+    "en": "Harmful",
+    "fr": "Nocif",
```
lbergeron, line 6675:
```suggestion
    "fr": "Nocive",
```

lbergeron (approved):
A few minor changes are required, but the rest looks good to me.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer-feedback/atomic-generated-answer-feedback-modal.tsx
```diff
@@ -0,0 +1,291 @@
...
+        class="w-full p-2 flex justify-between items-center"
+      >
+        <h1>{this.bindings.i18n.t('feedback')}</h1>
+        <IconButton
+          style="text-transparent"
+          class="search-clear-button"
+          onClick={() => this.close()}
+          icon={CloseIcon}
+          partPrefix={'close'}
+          ariaLabel={this.bindings.i18n.t('modal-done')}
```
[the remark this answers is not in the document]
drosenfeld-coveo: the text should be "done" after sending feedback like in the copy

PR #3343, opened 2023-10-27
docs: add every missing headless controller in doc-parsed

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2855

Multiple controllers are missing in the doc-parsed config files from many use case. We need to add them all for the validation of SSR to work correctly.

This PR adds them all. I counted each of them using their respective index.ts file.

Should maybe add some sort of ci in the future to remember people that whenever they add a public controller to a index.ts file, they need to add it in the respective doc parser config.


packages/headless/src/case-assist.index.ts
```diff
@@ -89,6 +89,9 @@ export {buildCaseAssistInteractiveResult as buildInteractiveResult} from './cont
 export {buildDocumentSuggestionList as buildDocumentSuggestion} from './controllers/document-suggestion-list/headless-document-suggestion-list';
 
 export type {
+  InteractiveResultCore,
```
olamothe, line 92: 🤔 Why is this necessary to export ? 

This should be internal only.

olamothe (changes requested):

packages/headless/src/case-assist.index.ts
alexprudhomme: This interface is necessary to export because the InteractiveResult extends from it. It is already exported by default, this addition just makes it explicit for the doc-parser. It will result in a doc page looking like the product listing facet one. Here in the index file, the Core interfaces are being exported.

image

And in the documentation, they are not showed, the doc-parser simply builds the documentation using information from the Facet controller and the FacetCore one.

https://docs.coveo.com/en/headless/latest/reference/product-listing/controllers/facet/

louis-bompart (approved):
LGTM, please do check the doc later ™️ to ensures that nothing is exposed.

Please log a tech debt Jira in the v3 epic to fixup/cleanup the types.
Ideally, the doc processor should be able to do its thing, and we should be able to not expose the guts.

PR #3349, opened 2023-10-30
feat(atomic): add copy button for generated answer

jelmedini:
[SVCC-3226](https://coveord.atlassian.net/browse/SVCC-3226)

This PR adds a copy button allowing the user to copy the generated answer to the clipboard.

Expected behaviour:

- The copy button is shown once the answer is fully streamed.
- When we click on copy button the generated answer is copied to the clipboard.
- On hover  the button goes to the blue state.
- When copied the button goes to the green state.
-  When we click on copy button, after a timeout the button should go back to it's default state (dark gray).

This PR also includes adding button state colors for feedback buttons as it wasn't consistent with the mockups.

Screenshot 2023-10-30 at 1 22 28 PM

https://github.com/coveo/ui-kit/assets/119955059/d73fd3a1-d882-441f-aaaa-b21a3ae8b2b0

[SVCC-3226]: https://coveord.atlassian.net/browse/SVCC-3226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/search/atomic-generated-answer/styles/copy-button.pcss
```diff
@@ -0,0 +1,16 @@
+[part='copy-button'] {
+  .icon-container {
+    atomic-icon {
+      &:hover {
```
fbeaudoincoveo, line 4: In the mockups, I see that the icon should go blue on hover.

Is the change voluntary?

packages/atomic/src/components/search/atomic-generated-answer/copy-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <Button
+      title={props.title}
+      part="copy-button"
+      style={'text-neutral'}
+      class={` p-2 rounded-md ${props.isCopied ? 'copied' : ''}`}
+      onClick={props.onClick}
+    >
+      <div class="icon-container text-neutral-dark">
+        <atomic-icon class="w-5" icon={CopyIcon}></atomic-icon>
```
fbeaudoincoveo, line 21: I think we should expose a part to style the icon

```suggestion
        <atomic-icon part="copy-button-icon" class="w-5" icon={CopyIcon}></atomic-icon>
```

fbeaudoincoveo: Out of scope, but this makes me realize we're currently not exposing parts on the FeedbackButton component.

packages/atomic/cypress/e2e/generated-answer.cypress.ts
```diff
@@ -166,6 +168,10 @@ describe('Generated Answer Test Suites', () => {
           GeneratedAnswerSelectors.dislikeButton().should('exist');
         });
 
+        it('should display copy button', () => {
```
fbeaudoincoveo, line 171: Beyond testing visibility, I believe we should test the functionality itself. I.e., that when the button is clicked, the current generated answer gets copied to the clipboard.

fbeaudoincoveo (changes requested):

packages/atomic/src/components/search/atomic-generated-answer/styles/copy-button.pcss
jelmedini: @fbeaudoincoveooh your are right I got confused when it goes green when it's copied, I will fix this :)

packages/atomic/src/components/search/atomic-generated-answer/copy-button.tsx
jelmedini: @fbeaudoincoveo I will add it and i will also add button state on hover, currently it's always dark gray i will change it to green and red accordingly.

packages/atomic/src/locales.json
```diff
@@ -6723,5 +6723,59 @@
...
+    "sv": "Kopiera svar",
+    "th": "คัดลอกคำตอบ",
+    "tr": "Yanıtı kopyala",
+    "zh": "复制答案",
+    "zh-CN": "复制答案",
+    "zh-TW": "複製答案"
+  },
+  "generated-answer-copied": {
+    "en": "Copied!",
+    "fr": "Copié !",
```
lbergeron, line 6756: There should be no whitespace before the exclamation mark.

```suggestion
    "fr": "Copié!",
```

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -154,6 +158,14 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
     return 'mt-0 mb-4 border border-neutral shadow-lg p-6 bg-background rounded-lg p-6 text-on-background';
   }
 
+  private async copyToClipboard(answer: string) {
+    await navigator.clipboard.writeText(answer);
+    this.copied = true;
+    setTimeout(() => {
+      this.copied = false;
+    }, 2000);
```
lbergeron, line 166: In my opinion, this "rendering" bit of logic should be inside the `CopyButton` component. It would make sense to me to have the same behavior when using `CopyButton` elsewhere.

I realize that implementing it using a function component may be somewhat tricky considering the component has no state. I think you could work around this limitation by toggling the CSS class directly from the inner button event handlers.

Another way would be to use an internal class component, but I think using function component is preferred for this repository.

packages/atomic/src/components/search/atomic-generated-answer/copy-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onClick: () => void;
+}
+
+export const CopyButton: FunctionalComponent<CopyButtonProps> = (props) => {
+  return (
+    <Button
+      title={props.title}
+      part="copy-button"
+      style={'text-neutral'}
+      class={` p-2 rounded-md ${props.isCopied ? 'copied' : ''}`}
```
lbergeron, line 17:
```suggestion
      class={`p-2 rounded-md ${props.isCopied ? 'copied' : ''}`}
```

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
jelmedini: @lbergeron I will need the state anyway because I need o change the button label from `copy the answer` to `copied!` when the copy is fulfilled. also, if I understood right,  if I try to do this part of the logic in the copybutton it assumes that the copy is always fulfilled, so if  `await navigator.clipboard.writeText(answer);` wasn't fullfilled the button will still turn to green and I don't think we do want that.

lbergeron: I was thinking that `CopyButton` could use the inner button `click` handler to toggle the CSS class and update the text. I would probably bring the `navigator.clipboard.writeText(answer)` logic in `CopyButton` as well to abstract most of the logic.

By the way I'm fine with the way you implemented the component. I am simply suggesting something slightly different.

lbergeron (approved):

nathanlb (approved):
Nicely done!

fbeaudoincoveo (approved):
💯

PR #3359, opened 2023-11-02
feat(quantic): refactored QuanticResultAction to make it use the QuanticTooltip component

mmitiche:
[SFINT-5249](https://coveord.atlassian.net/browse/SFINT-5249)

### Explanation: 
The QuanticResultAction and the QuanticResuiltQuickView were both implementing the same logic in order to display the tooltip. Now, after the creation of the new Quantic Tooltip component, we can remove the duplicate code and start using the newly created component in order to have the same tooltip across all the Quantic components.

### In this PR:

- Started using the Quantic Tooltip in the Quantic Result Action component.
- Updated the unit test of the Quantic Result Action component.
- Started using the Quantic Tooltip in the Quantic Result Quickview component.
- Updated the E2E tests of the Quantic Result Quickview component.

#### Before:

https://github.com/coveo/ui-kit/assets/86681870/e0da5117-2cf3-40ea-8458-c1fdbe940da6

#### After

https://github.com/coveo/ui-kit/assets/86681870/c28d18e1-26bd-4129-ae9c-a830e2202955

[SFINT-5249]: https://coveord.atlassian.net/browse/SFINT-5249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
```diff
@@ -1,7 +1,7 @@
 .tooltip__content {
   left: 50%;
   transform: translate(-50%, -100%);
-  top: -0.7rem;
+  top: -0.4rem;
```
erocheleau, line 4: I noticed in your videos that the tooltip arrow was now aligned with the icon _inside_ the button now instead of being above the entire button.

Is that expected?

erocheleau (approved):
Just the question of the alignement of the tooltip arrow that now appears "inside" the button instead of above it.
Is this what you wanted? I don't really mind either way

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
mmitiche: Yes I found a bit weird that the tooltip is kinda disconnected from the button and completely above  it.

PR #3360, opened 2023-11-02
fix: add missing reducers to `SearchParameterManager` controller

y-lakhdar:
Some reducers where never executed even if the expected action was dispatched.
That's because the reducers where never loaded in the first place.

KIT-2866

### DIFF
--- a/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
+++ b/packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
@@ -1,19 +1,5 @@
-import {advancedSearchQueriesReducer} from '../../../features/advanced-search-queries/advanced-search-queries-slice';
-import {configurationReducer} from '../../../features/configuration/configuration-slice';
-import {debugReducer} from '../../../features/debug/debug-slice';
-import {facetOptionsReducer} from '../../../features/facet-options/facet-options-slice';
-import {automaticFacetSetReducer} from '../../../features/facets/automatic-facet-set/automatic-facet-set-slice';
-import {categoryFacetSetReducer} from '../../../features/facets/category-facet-set/category-facet-set-slice';
-import {facetSetReducer} from '../../../features/facets/facet-set/facet-set-slice';
-import {dateFacetSetReducer} from '../../../features/facets/range-facets/date-facet-set/date-facet-set-slice';
-import {numericFacetSetReducer} from '../../../features/facets/range-facets/numeric-facet-set/numeric-facet-set-slice';
-import {paginationReducer} from '../../../features/pagination/pagination-slice';
-import {queryReducer} from '../../../features/query/query-slice';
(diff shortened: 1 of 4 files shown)


packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.test.ts
```diff
@@ -47,6 +61,25 @@ describe('search parameter manager', () => {
     initSearchParameterManager();
   });
 
+  it('it adds the correct reducers to engine', () => {
```
fbeaudoincoveo, line 64: 👍 

Out of curiosity, do you know if there's a reason why we usually phrase this `it` description as "it adds ..." rather than just "adds ..."? Total pet peeve, but it does annoy me a tiny bit 😂

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
```diff
@@ -282,3 +301,25 @@ function getAutomaticFacets(state: Partial<SearchParametersState>) {
 
   return Object.keys(af).length ? {af} : {};
 }
+
+function loadSearchParameterManagerReducers(
+  engine: CoreEngine
+): engine is CoreEngine<SearchParameterManager> {
+  engine.addReducers({
```
fbeaudoincoveo, line 308: Reminder: this change will render the code sample in this documentation page kind of pointless https://docs.coveo.com/en/atomic/latest/usage/headless-through-atomic/, so it should be updated.

(I know you're aware of this documentation, but it's easy to forget about those things 😅)

fbeaudoincoveo (approved):

packages/headless/src/controllers/core/search-parameter-manager/headless-core-search-parameter-manager.ts
louis-bompart: Should add smthg smwhere to not forget ;)

louis-bompart (approved):

alexprudhomme (changes requested):
I dont think this is right. If you dont have category facets, you dont need the `restoreSearchParameter` action in the category facet set to work.

louis-bompart:
Option 2 smells, I agree, what are the pros/cons of the two others?

fbeaudoincoveo:
> The reason I added those was for use cases where the user wanted to initialize the SearchParameterManager controller on its own. [...]
> This can be an issue for implementers who use the SearchParameterManager in SSR. Unless they manually load the needed reducers, they will never be able to provide an initialState to that controller.

Can you think of any legitimate use case for initializing the SearchParameterManager controller aside from SSR?

If not, option 3 (load all the reducers in the defineSearchParameterManager function) does seem like the best one to me.

alexprudhomme:
> Load all the reducers in the defineSearchParameterManager function.

I like option 3 !

alexprudhomme (approved):

PR #3361, opened 2023-11-03
feat(quantic): added the ability to have text highlighting in a quantic search interface

mmitiche:
[SFINT-4882](https://coveord.atlassian.net/browse/SFINT-4882)

### Explanation

The fields that can be  highlighted are the following:  `title`, `excerpt`, `printable URI`, `first sentences` and `summary`, for these fields, when highlights are available, the Search API returns additional field in each results. 

These additional  fields have the following pattern `${exampleField}Highlights` and they contain the position of all the highlights present in the value of the field `exampleField`.

I'm creating this new Quantic component, the `QuanticResultHighlightedTextField` to benefit from this feature and use some Headless utilities to properly display the highlights.

### Note
We already have a component called `Quantic Result Text`, this component displays the text field value of a given field in a result, I was able to added the highlighting logic directly to this component instead of creating a new component.

I decided to not go with this alternative for this important reason: The Quantic Result Text component exposes a public property called `formattingFunction`, it's basically a function that be provided to format the result before displaying it and since the highlighting logic plays with `innerHTML` and `lwc:dom:manual` I didn't want to open the door to being able to customize the HTML directly from outside the component with the help of this public property.

Instead I created the `QuanticResultHighlightedTextField` which doesn't expose any property that allow modifying the text that will be displayed.

#### Without text highlighting:
testBefore

#### With text highlighting:
TestAfter2

[SFINT-4882]: https://coveord.atlassian.net/browse/SFINT-4882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/__tests__/quanticResultHighlightedTextField.js
```diff
@@ -0,0 +1,211 @@
...
+      );
+
+      expect(errorMessage).not.toBeNull();
+      expect(console.error).toHaveBeenCalledWith(
+        `The ${componentName} requires a result and a field to be specified.`
+      );
+    });
+  });
+
+  describe('when field is given as am empty string', () => {
```
erocheleau, line 133:
```suggestion
  describe('when and empty string is given as a field', () => {
```

packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/__tests__/quanticResultHighlightedTextField.js
```diff
@@ -0,0 +1,211 @@
...
+          (value) => Object.prototype.toString.call(value) === '[object String]'
+        ),
+    };
+  });
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  describe('when no result is given', () => {
```
erocheleau, line 99: How about we structure it like this?

```
describe('missing properties error handling', () => {
    it('should handle when the result is missing', async () => {});
    it('should handle when the field is missing', async () => {});
    it('should handle when the field is an empty string', async () => {});
});
```

packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/__tests__/quanticResultHighlightedTextField.js
```diff
@@ -0,0 +1,211 @@
...
+      );
+
+      expect(errorMessage).not.toBeNull();
+      expect(console.error).toHaveBeenCalledWith(
+        `The ${componentName} requires a result and a field to be specified.`
+      );
+    });
+  });
+
+  describe('when required props are given', () => {
```
erocheleau, line 149:
```
describe('displaying highlights', () => {
    it('should render the highlights from the result', async () => {});
    it('should not render highlights when the result has no highlights', async () => {});
});

describe('displaying with a label on the field', () => {
    it('should render the field value with a label', async () => {});
});
```

packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/__tests__/quanticResultHighlightedTextField.js
```diff
@@ -0,0 +1,211 @@
...
+      const element = createTestComponent({result: exampleResult});
+      await flushPromises();
+
+      const errorMessage = element.shadowRoot.querySelector(
+        selectors.errorSelector
+      );
+
+      expect(errorMessage).not.toBeNull();
+      expect(console.error).toHaveBeenCalledWith(
+        `The ${componentName} requires a result and a field to be specified.`
```
erocheleau, line 128: You could put that string as a const declaration at the beginning of your describe since you re-use three times

```const expectedConsoleErrorMessage = `The ${componentName} requires a result and a field to be specified.`;```

File: packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/quanticResultHighlightedTextField.css [hunk 5]
```diff
@@ -0,0 +1,7 @@
+.result-text__label {
+    font-weight: var(--lwc-fontWeightBold, 700);
+}
+
+b {
+  font-weight: bolder;
```
erocheleau, line 6: why bolder?

packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/quanticResultHighlightedTextField.js
```diff
@@ -0,0 +1,128 @@
...
+    );
+    /** @type {HighlightKeyword[]} */
+    const highlights = this.headless.ResultTemplatesHelpers.getResultProperty(
+      this.result,
+      `${this.field}Highlights`
+    );
+
+    if (highlights) {
+      const openingDelimiter = '<b>';
+      const closingDelimiter = '</b>';
```
erocheleau, line 102: I think it would be good to add a class to our <b> html, even if the class doesn't do anything, it would be easy to recognize and to use selectors on.

And even eventually to override via perhaps a CSS hook we would offer?

erocheleau (approved):
Some small suggestions to improve test clarity, but otherwise great job!

packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/quanticResultHighlightedTextField.css
mmitiche: So it always takes a value relatively bolder relatively to the parent element: https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#bolder

packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/quanticResultHighlightedTextField.js
mmitiche: sure!

packages/quantic/force-app/main/default/lwc/quanticResultHighlightedTextField/__tests__/quanticResultHighlightedTextField.js
mmitiche: Better 👍

PR #3363, opened 2023-11-03
fix(headless): dispatch action when fetchMoreResults() is called from a folded result list controller

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-2865

The issue was introduced through the progressive refactoring of the result list and folded result list controllers to extract core controllers to reuse with the insight engine.

In a nutshell, before the refactoring, we had:

`buildFoldedResultList` calls `buildResultList`

...and `buildResultList` was directly dispatching the `fetchMoreResultsAction` from the search actions when the `fetchMoreResults` method was called.

After the first pass of refactoring (https://github.com/coveo/ui-kit/pull/2897), we still had:

`buildFoldedResultList` calls `buildResultList`

...but a second level of abstraction got added:

`buildResultList` calls `buildCoreResultList`

...and `buildResultList` was passing the `fetchMoreResultsAction` from the search action as a prop to `buildCoreResultList`, which then dispatched it when the `fetchMoreResults` was called. So everything was still working fine at that point.

Finally, following the second pass of refactoring (https://github.com/coveo/ui-kit/pull/2065), we ended up with:

`insightFoldedResultList` calls `buildCoreFoldedResultList` calls `buildCoreResultList`
and
`foldedResultList` calls `buildCoreFoldedResultList` calls `buildCoreResultList`

...and no `fetchMoreResultsAction` was passed down to `buildCoreResultList`, so it dispatched nothing when the `fetchMoreResults` method was called from a folded result list controller (insight or not), essentially causing infinite scrolling (or load more) to cease working for result lists powered by one of these controllers.

I'm frankly surprised that this issue only surfaced 5 months later!


fbeaudoincoveo:
@lbergeron I'm adding you as a reviewer on this one since I'm modifying the insight folded result list controller.

fbeaudoincoveo:
This makes me realize that passing down action creators like this makes testing those controllers really awkward.

In this specific instance, I believe we should somehow test that when the `showMoreResults` method is called from the folded result list controller, it dispatches the `fetchMoreResults` **search** action, whereas when the same method is called from the insight folded result list controller, it should dispatch the `fetchMoreResults` **insight** action.

However that method is defined on the core result list controller (two abstraction levels down from the folded result list controllers). The `fetchMoreResults` method's behavior is tested there, but the test assumes that 1) `fetcMoreResultsActionLoader` will be defined (it's an optional prop), and 2) it's going to be the fetchMoreResult search action. In short, the current test is flawed.

I didn't check, but we may have similar flaws in our tests elsewhere in Headless when we're using this pattern (i.e., passing action creators to core controllers).

packages/headless/src/controllers/core/folded-result-list/headless-core-folded-result-list.ts
```diff
@@ -249,7 +255,9 @@ export function buildCoreFoldedResultList(
 
 function loadFoldingReducer(
   engine: CoreEngine
-): engine is CoreEngine<SearchSection & ConfigurationSection & FoldingSection> {
+): engine is CoreEngine<
```
fbeaudoincoveo, line 258: Should not change anything, but there was a missing section in the type predicate.

packages/headless/src/controllers/core/result-list/headless-core-result-list.ts
```diff
@@ -50,7 +50,7 @@ export interface ResultListProps {
    * The action creator to build the `fetchMoreResults` action.
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  fetchMoreResultsActionCreator?: () => AsyncThunkAction<unknown, void, any>;
+  fetchMoreResultsActionCreator: () => AsyncThunkAction<unknown, void, any>;
```
fbeaudoincoveo, line 53: I'm making this (and by extension, the controller's `props` parameter) mandatory: we should always pass a fetchMoreResultsActionCreator when building this controller, otherwise the `fetchMoreResults` method won't do anything.

fbeaudoincoveo: I ended up rollings this back; the ResultListProps interface is used in atomic too so it broke stuff.

packages/headless/src/controllers/insight/folded-result-list/headless-insight-folded-result-list.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    foldedResultList.loadCollection({
+      children: [],
+      isLoadingMoreResults: false,
+      moreResultsAvailable: false,
+      result: buildMockResult(),
+    });
+
+    expect(engine.findAsyncAction(loadCollection.pending)).toBeTruthy();
+  });
+  it('#fetchMoreResults dispatches the insight search #fetchMoreResults action', () => {
```
louis-bompart, line 41:
```suggestion
  });

  it('#fetchMoreResults dispatches the insight search #fetchMoreResults action', () => {
```

louis-bompart (approved):
Looks like a nice & precise fix!

lbergeron (approved):
It looks good to me!

PR #3370, opened 2023-11-06
fix(headless): rga controller

jelmedini:
[SVCC-3247](https://coveord.atlassian.net/browse/SVCC-3247)

Currently when we have multiple RGA controller instances, in the  `subscribeToSearchRequest` the calls interfere with each other, which leads to streams cancelled or restarted.
Fix: implementing subscribeToSearchRequest logic like a custom state manager outside the function.

[SVCC-3247]: https://coveord.atlassian.net/browse/SVCC-3247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


louis-bompart:
If the issue was occurring in Quantic, it's likely to keep on occurring in Quantic even with this fix because the abort controller is 'poly filled' to null.

jelmedini:
> If the issue was occurring in Quantic, it's likely to keep on occurring in Quantic even with this fix because the abort controller is 'poly filled' to null.

@louis-bompart Do you have a specific use case where it's not gonna work? because what i've tested so far works fine, we get also rid of a lot unnecessary calls

mmitiche (commented):
Very nice fix, good job 👍 and I confirmed that it works fine with Quantic too.

I would just suggest adding a new unit test for this new logic, a test that would be kinda similar to this one: 
https://github.com/coveo/ui-kit/blob/6e9c411e2e649ebcc81a907468bca985b233d723/packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts#L267
But instead we wanna create two Generated Answer Controllers and then we need to check that the action `streamAnswer` was dispatched just once and not twice.

Let me know if you need any help with setting up the test.

louis-bompart (approved):
Validate in Quantic, LGTM for Atomic

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
```diff
@@ -275,6 +275,26 @@ describe('generated answer', () => {
...
       expect(action).toBeTruthy();
     });
+
+    describe('when we have multiple controllers', () => {
+      let secondEngine: MockSearchEngine;
+      beforeEach(() => {
+        secondEngine = buildMockSearchAppEngine();
+        buildGeneratedAnswer(secondEngine);
+      });
+      it('should dispatch the stream action only once', () => {
```
nathanlb, line 285:
```suggestion
      let secondEngine: MockSearchEngine;
      
      beforeEach(() => {
        secondEngine = buildMockSearchAppEngine();
        buildGeneratedAnswer(secondEngine);
      });
      
      it('should dispatch the stream action only once', () => {
```

nathanlb (approved):
Looking good! 👍

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
```diff
@@ -275,6 +275,26 @@ describe('generated answer', () => {
...
+        engine.state.search.extendedResults = {
+          generativeQuestionAnsweringId: 'another-fake-test-id',
+        };
+
+        callListener();
```
mmitiche, line 290: I think this test would successfully pass even with the previous implementation where we had the bug cause here in the `callListener` we are executing the listener that was added only by the first controller.

We need to make sure we are executing the listeners that were added by both controllers than confirm that the action was dispatched just once.

jelmedini: héhé yes i was actually adding it right now

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
```diff
@@ -275,6 +275,34 @@ describe('generated answer', () => {
...
       expect(action).toBeTruthy();
     });
+
+    describe('when we have multiple controllers', () => {
+      let secondEngine: MockSearchEngine;
+      function callSecondListener() {
+        return (secondEngine.subscribe as jest.Mock).mock.calls.map(
+          (args) => args[0]
+        )[0]();
+      }
```
mmitiche, line 285: To execute the second listener that was added we need to target the index `1` instead of `0` to get something like this:
```
      function callSecondListener() {
        return (secondEngine.subscribe as jest.Mock).mock.calls.map(
          (args) => args[0]
        )[1]();
      }
```
But for a more generic solution I would only update the `callListener` function and make it execute every listener that was added: 
```
    function callListener() {
      return (engine.subscribe as jest.Mock).mock.calls
        .map((args) => args[0])
        .forEach((listener) => {
          listener();
        });
    }
```

jelmedini: nice catch! 👀 Thank youu

PR #3374, opened 2023-11-07
[CAPI-223] POC commerce search

Spuffynism:
🔔 This PR will **NOT** be merged. Some parts may be reused in work done as part of CAPI-243.

--- 

The goal of this POC was to figure out the work needed to support the Search solution type in the headless commerce sub-package.

This PR:
- Creates search & search box controllers
- Creates a core sort controller from the current product listing sort controller, and then implements it for both search and product listings. I've created 2 different controllers, but we might want to have a single controller that knows about both Search and Listings. TBD
- Copies over search actions

I've also added two integration tests that show:
- The reuse of a controller that **does not** need to be adapted for search (the context controller)
- The use of a controller that **does** need to be adapted for search (sort controller)

Furthermore, contrary to the search and insight engines which expose search methods directly on the Engine, I've chosen to expose search methods through a controller. This fits better with the structure of having the commerce engine support three solutions types (search, listings, recs). ❗ **I'm also open to exposing these search methods directly on the engines, I just haven't found any reason to do so yet!** ❗

Finally, this PR makes it evident that we need a query suggest route on the API to fully support search as a solution type in commerce headless.

[CAPI-223]

[CAPI-223]: https://coveord.atlassian.net/browse/CAPI-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CAPI-243]: https://coveord.atlassian.net/browse/CAPI-243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/app/commerce-engine/commerce-engine.ts
+++ b/packages/headless/src/app/commerce-engine/commerce-engine.ts
@@ -21,6 +21,7 @@ import {
   CommerceEngineConfiguration,
   commerceEngineConfigurationSchema,
 } from './commerce-engine-configuration';
+import {searchReducer} from '../../features/commerce/search/search-slice';
 
 export type {CommerceEngineConfiguration};
 
@@ -29,6 +30,7 @@ const commerceEngineReducers = {
   commercePagination: paginationReducer,
   commerceSort: sortReducer,
   commerceContext: contextReducer,
+  commerceSearch: searchReducer,
   cart: cartReducer,
 };
 type CommerceEngineReducers = typeof commerceEngineReducers;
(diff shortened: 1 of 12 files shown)


Spuffynism:
Work to productize this POC has begun in https://github.com/coveo/ui-kit/pull/3400

PR #3376, opened 2023-11-07
fix(headless): update extendedResults on all fulfilled searches

nathanlb:
## The Issue ##
Fetching more results using the pager or the resultsPerPage component returns a new `generativeQuestionAnsweringId` but no new stream is initiated.

## The Fix ##
Update the redux state with the new `generativeQuestionAnsweringId` causing a new stream to be triggered.

### DIFF
--- a/packages/headless/src/features/search/search-slice.test.ts
+++ b/packages/headless/src/features/search/search-slice.test.ts
@@ -226,6 +226,27 @@ describe('search-slice', () => {
 
       expect(finalState.questionAnswer).toEqual(originalQuestionAnswers);
     });
+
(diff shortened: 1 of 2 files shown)


nathanlb:
Back to draft. Still unsure about this behaviour.

fbeaudoincoveo (approved):

nathanlb:
Wasn't sure about this "fixed" behaviour. What was there before seems preferred and more in line with our question answering design as it already matched smart snippets. Going to reject this. If we get reports of this being an issue and contradicting use expectations we can revisit. The change is pretty minor in terms of code.

PR #3381, opened 2023-11-09
refactor(analytics): allow search events to go through a new clean path

louis-bompart:
Review guide:
 - everything in `packages/headless/src/features/search/`
 - Where I sparkled a few comments

### DIFF
--- a/packages/headless/src/app/analytics-middleware.test.ts
+++ b/packages/headless/src/app/analytics-middleware.test.ts
@@ -2,9 +2,9 @@ import {logSearchboxSubmit} from '../features/query/query-analytics-actions';
 import {
   executeSearch,
   ExecuteSearchThunkReturn,
-} from '../features/search/search-actions';
+} from '../features/search/legacy/search-actions';
 import {buildMockSearchAppEngine} from '../test/mock-engine';
-import {buildMockSearch} from '../test/mock-search';
+import {buildMockLegacySearch} from '../test/mock-search';
 
 describe('analytics middleware', () => {
   beforeEach(() => {
@@ -22,7 +22,8 @@ describe('analytics middleware', () => {
 
   it('correctly pass through a search action with no analytics payload', () => {
     const e = buildMockSearchAppEngine();
-    const {analyticsAction, ...mockSearchWithoutAnalytics} = buildMockSearch();
+    const {analyticsAction, ...mockSearchWithoutAnalytics} =
+      buildMockLegacySearch();
 
     const action = executeSearch.fulfilled(
       mockSearchWithoutAnalytics as ExecuteSearchThunkReturn,
@@ -35,7 +36,7 @@ describe('analytics middleware', () => {
 
   it('correctly pass through a search action with an analytics payload', () => {
     const e = buildMockSearchAppEngine();
-    const mockSearch = buildMockSearch();
+    const mockSearch = buildMockLegacySearch();
 
     const action = executeSearch.fulfilled(
       mockSearch,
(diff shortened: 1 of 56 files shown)


packages/headless/src/features/insight-search/insight-search-actions.ts
```diff
@@ -43,7 +43,7 @@ import {
 } from '../query-suggest/query-suggest-actions';
 import {updateQuery} from '../query/query-actions';
 import {getQueryInitialState} from '../query/query-state';
-import {ExecuteSearchThunkReturn} from '../search/search-actions';
+import {ExecuteSearchThunkReturn} from '../search/legacy/search-actions';
```
louis-bompart, line 46: Insight panel out of scope of this migration

packages/headless/src/controllers/insight/search-box/headless-insight-search-box.ts
```diff
@@ -45,6 +45,7 @@ export function buildSearchBox(
     ...props,
     executeSearchActionCreator: executeSearch,
     fetchQuerySuggestionsActionCreator: fetchQuerySuggestions,
+    isNextAnalyticsReady: false,
```
louis-bompart, line 48: This allow the branching to work in the headlessCoreSearchBox.
When Insight migrate will have done the as this PR, they will be able to turn it to true (and we'll remove false and clean up)

packages/headless/src/controllers/core/search-box/headless-core-search-box.ts
```diff
@@ -201,8 +238,11 @@ export function buildCoreSearchBox(
         clearFilters,
       })
     );
-
-    await dispatch(props.executeSearchActionCreator(analytics));
+    if (props.isNextAnalyticsReady) {
+      dispatch(props.executeSearchActionCreator(analytics));
+    } else {
+      dispatch(props.executeSearchActionCreator(analytics.legacy));
+    }
```
louis-bompart, line 245: Allow support of both search and insight sb until insight is migrated

packages/headless/src/app/analytics-middleware.test.ts
```diff
@@ -2,9 +2,9 @@ import {logSearchboxSubmit} from '../features/query/query-analytics-actions';
 import {
   executeSearch,
   ExecuteSearchThunkReturn,
-} from '../features/search/search-actions';
+} from '../features/search/legacy/search-actions';
 import {buildMockSearchAppEngine} from '../test/mock-engine';
-import {buildMockSearch} from '../test/mock-search';
+import {buildMockLegacySearch} from '../test/mock-search';
```
louis-bompart, line 7: This is the only test file that is likely to stay and die on the buildMock, as we're not porting the analyticsMiddleware to v3

packages/headless/src/features/search/legacy/search-actions-thunk-processor.ts
```diff
@@ -0,0 +1,364 @@
+import {AnyAction} from '@reduxjs/toolkit';
```
alexprudhomme, line 1: Should we add legacy in this file's name ? This could get confusing with two files with same name so close to each other.

alexprudhomme (approved):
lgtm ,seems simple to use for the migration.

packages/headless/src/features/search/legacy/search-actions-thunk-processor.ts
louis-bompart: That's why I put them in a /legacy folder, every import of it requires to go 'thru' /legacy because of this
(So, tldr, nope, already in the path so is ok)

louis-bompart:
@alexprudhomme did some more thorough testing and spotted a few issues:
 - Redux issue number 8, no type with the action
 - No customData at all are passed
 This is fixed with 23f5aef
 
 Remaining:
  - All the other customData that are defined by the Atomic-defined analyticsMiddleware https://github.com/coveo/ui-kit/blob/master/packages/atomic/src/components/search/atomic-search-interface/analytics-config.ts
  
 Out of scope:
  - Custom/External middlewares: https://github.com/coveo/ui-kit/blob/master/packages/atomic/src/components/search/atomic-search-interface/analytics-config.ts

PR #3384, opened 2023-11-10
chore(quantic): Replace sfdx-cli with @salesforce/cli

erocheleau:
[SFINT-5114]

sfdx-cli is in maintenance only mode, all the updates are now done to the @salesforce/cli

Pretty much all the commands of `sfdx` still work as-is with @salesforce/cli

But still we want to update all those commands to the new `sf commands format` 

[SFINT-5114]: https://coveord.atlassian.net/browse/SFINT-5114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


louis-bompart (approved):
I see green tests and a dep update, megusta.
🆗 for bypass of the @babel/traverse report.

packages/quantic/README.md
```diff
@@ -15,16 +15,18 @@
...
 
-[SFDX cli](https://developer.salesforce.com/tools/sfdxcli)
+[Salesforce cli](https://developer.salesforce.com/docs/atlas.en-us.sfdx_setup.meta/sfdx_setup/sfdx_setup_install_cli.htm#sfdx_setup_install_cli_npm)
+
+- Run `npm install --global @salesforce/cli`
+- Optional: [Setup CLI autocomplete](https://developer.salesforce.com/docs/atlas.en-us.sfdx_setup.meta/sfdx_setup/sfdx_dev_cli_autocomplete.htm)
 
 ### Update SFDX CLI Plugins
 
-- Run `sfdx plugins:install @salesforce/lwc-dev-server`
```
mmitiche, line 22: Are we giving up on the lwc dev server?

packages/quantic/scripts/build/util/sfdx.ts
```diff
@@ -12,14 +12,14 @@ export interface SfdxResponse {
 
 /**
  * Executes the given sfdx function and returns the result as an object.
- * @param command The command to be executed, starting with force:.
+ * @param command The command to be executed.
  * @returns {SfdxResponse} The result of the command, if successful.
  * @throws {Error} The error thrown by the sfdx command if unsuccessful.
  */
 export function sfdx<T = SfdxResponse>(command: string): Promise<T> {
```
mmitiche, line 19: I think it would be better to update the name of this function to a more accurate name now: 
```suggestion
export function sf<T = SfResponse>(command: string): Promise<T> {
```

packages/quantic/scripts/build/util/sfdx.ts
mmitiche, line 1: The file could also be renamed to `sf.ts`

packages/quantic/scripts/build/util/sfdx-commands.ts
mmitiche, line 1: I suggest renaming to `sf-commands.ts` for a proper cleaning

mmitiche (approved):
Looks good 👌
I only think renaming some stuff from `sfdx` to `sf` would be nice.

packages/quantic/README.md
erocheleau: yes

packages/quantic/scripts/build/util/sfdx.ts
erocheleau: I think it's more confusing personally to just have `sf` instead of `sfdx`.

erocheleau:
> Actually, I checked again why [GHSA-67hx-6x53-jw92](https://github.com/advisories/GHSA-67hx-6x53-jw92) was reported again, and indeed, `@salesforce/cli` was still using (and requiring) a vulnerable version of `@babel/traverse`, until yesterday or so.
> 
> I bumped SFDX again in [bf2de78](https://github.com/coveo/ui-kit/commit/bf2de787961809576ad85981d630287f198804a7) to fix that.

Does this update fix the babel/travers vuln?

I updated first to that version of sf yes but they had a bug on the .8 version, they since released the .10 that you used.
So if the .10 works, then all good to update to that yes.

packages/quantic/scripts/build/util/sfdx.ts
erocheleau: The problem is that there are multiple other types and stuff in the code to rename.
I don't think it's too confusing to keep using SFDX as the prefix, after all this means "Salesforce Developer Experience"

PR #3388, opened 2023-11-13
docs(atomic): load default component story when none is specified in URL or specified story isn't found

fbeaudoincoveo:
🧟‍♂️

### DIFF
--- a/packages/atomic/src/components/common/atomic-icon/atomic-icon.stories.tsx
+++ b/packages/atomic/src/components/common/atomic-icon/atomic-icon.stories.tsx
@@ -73,5 +73,5 @@ const {defaultModuleExport, exportedStory} = defaultStory(
   }
 );
 
-export const DefaultIcon = exportedStory;
 export default {...defaultModuleExport, title: 'Atomic/Icon'};
+export const Default = exportedStory;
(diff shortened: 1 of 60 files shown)


louis-bompart:
it works, is there drawbacks?

fbeaudoincoveo:
@srosenberg-coveo / @npushkarskii , the path to the default story will now always be:

`?path=/story/{atomic-component-tag}--default` (where `{atomic-component-tag}` must be substituted).

E.g., 

`?path=/story/atomic-folded-result-list--default`

fbeaudoincoveo:
> it works, any drawbacks? (other than 'gosh darn it's silly we have to do that)

I did notice that if you don't specify a story at all in the `path` query parameter (e.g., if you do `?path=atomic-search-box` instead of `?path=atomic-search-box--whatever`, the Storybook error message saying that the story can't be found will show up for an instant, and then the router will redirect to the default story.

This is still infinitely better than the current state of things, though (which is that it basically never loads the default example, unless, for some obsure reason, the component's PascalCase name only contains a single word, such as Icon or Breadbox).

louis-bompart (approved):

npushkarskii:
> @srosenberg-coveo / @npushkarskii , the path to the default story will now always be:
> 
> `?path=/story/{atomic-component-tag}--default` (where `{atomic-component-tag}` must be substituted).
> 
> E.g.,
> 
> `?path=/story/atomic-folded-result-list--default`
> 
> You can update the templates in the docs site to avoid flickering, but it should work out of the box once the fix gets deployed.

you're the best, Fred! Thx a lot. @fbeaudoincoveo 

A question: do all components have `atomic-` at the beginning of their tags(names) ? I'm just concerned whether the substitution part is unified on Storybook's end 🤔

\+ @srosenberg-coveo

PR #3390, opened 2023-11-13
[CAPI-87] feat(headless): allow skipping facet registration

Spuffynism:
To support generating facets, we must be able to register facets **outside** of the core facet controller.

Here's an example use case which depends on the mechanism proposed in this PR (https://github.com/coveo/ui-kit/pull/3389):
https://github.com/coveo/ui-kit/blob/5080e7855478c96321ccd33bf5061b1556fc96c1/packages/headless/src/controllers/commerce/facets/headless-facet-generator.ts#L34-L67

Without conditionally registering the facet, the facet generator's state seems to continuously `dispatch` the `registerFacet` action. Only by extracting the registration outside of the core facet, we are able to register facets properly.

[CAPI-87]

### DIFF
--- a/packages/headless/src/controllers/core/facets/facet/headless-core-facet.test.ts
+++ b/packages/headless/src/controllers/core/facets/facet/headless-core-facet.test.ts
@@ -113,6 +113,16 @@ describe('facet', () => {
     expect(() => initFacet()).toThrow('Check the options of buildFacet');
   });
 
+  it('skips registering the facet if explicitly told to', () => {
+    engine = buildMockSearchAppEngine({state});
+    facet = buildCoreFacet(engine, {
+      options,
+      registerFacet: false,
+    });
+    const action = engine.actions.find((a) => a.type === registerFacet.type);
+    expect(action).toBeUndefined();
+  });
+
   it('#state.facetId exposes the facet id', () => {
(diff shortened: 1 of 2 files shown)


fbeaudoincoveo:
Does it 

I see that @alexprudhomme didn't use the core facet controller at all for automatic facets.

Commerce API V2 facets are very similar to automatic facets (except they offer less options for the implementer, i.e., you can't specify how many you want to request, and they're actually pre-configured in the CMH).

I've been hesitant / uncomfortable with using the core controllers for various reasons since we started working on the new commerce sub-package. This adds to my discomfort 😅 

In short, the way I see it, the implementer probably won't even build a facet generator controller: rather, this would be done for them under the hood when they build a product listing controller and fetch the initial response. The way I see it, the implementer has to assume that every product listing interface is paginated, and may or may not include facets and sorts as well. They have no control over this at all: the product listing configuration can change at any time in the CMH, and the UI has to respond to that.

The main reason why I think we still a facet generator controller (as opposed to integrating this directly in the product listing controller) is to avoid code duplication when we get to the commerce search use case, because it's probably going to be very similar.

Spuffynism:
Will use the approach proposed by @fbeaudoincoveo in https://github.com/coveo/ui-kit/pull/3392!

PR #3392, opened 2023-11-14
[WIP] commerce core facet controller

Spuffynism:
This PR is a work in progress of creating a core facet controller for commerce.

Some interfaces might not be in the right places, many tests are missing, and I have yet to replace the usage of facet values by the commerce facet value types.

In more details, I've created 4 controllers:
- a core facet controller
- a core facet generator controller
- a plp facet controller (that uses the core commerce controller)
- a plp facet generator controller (that uses the core commerce controller)

I've also created a state section (with its state, actions, and reducers), `CommerceFacetSetSection`.

We'll be able to re-use the core facet controllers for future commerce solution types (aka search).


packages/headless/src/controllers/commerce/facets/core/headless-core-facet-generator.ts
```diff
@@ -0,0 +1,73 @@
...
+        });
+    }
+  };
+
+  return {
+    ...controller,
+
+    get state() {
+      return {
+        facets: engine.state.facetOrder.map(createFacet) ?? [],
```
fbeaudoincoveo, line 60: A thought:

As a data structure, the JavaScript `Map` object remembers the order of insertion (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

Could we define the `CommercefacetSetState` type as a `Map<string, CommerceFacetSlice>` instead of a `Record<string, CommerceFacetSlice>` to avoid having to duplicate facets in a `facetOrder` array?

packages/headless/src/controllers/commerce/facets/core/headless-core-facet.test.ts
```diff
@@ -0,0 +1,184 @@
...
+import {CommerceFacetRequest} from '../../../../features/commerce/facets/facet-set/interfaces/request';
+import {buildMockCommerceFacetSlice} from '../../../../test/mock-commerce-facet-slice';
+import {buildMockCommerceFacetRequest} from '../../../../test/mock-commerce-facet-request';
+import {buildMockCommerceState} from '../../../../test/mock-commerce-state';
+import {buildMockFacetValue} from '../../../../test/mock-facet-value';
+import {fetchProductListing} from '../../../../features/commerce/product-listing/product-listing-actions';
+import {logFacetDeselect} from '../../../../features/facets/facet-set/facet-set-product-listing-v2-analytics-actions';
+
+
+describe('CoreCommerceFacet', () => {
```
fbeaudoincoveo, line 14: It seems like these are the tests for the headless-product-listing-facet controller, because the core controller doesn't dispatch analytics actions.

fbeaudoincoveo (commented):
I have a few questions / comments, but this looks like a good start!

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export const commerceFacetSetReducer = createReducer(
+  getCommerceFacetSetInitialState(),
+
+  (builder) => {
+    builder
+      .addCase(fetchProductListing.fulfilled, (state, action) => {
+        const facets = action.payload.response.facets;
+        facets.forEach((facetResponse) =>
+          mutateStateFromFacetResponse(
```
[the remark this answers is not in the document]
Spuffynism: Done in https://github.com/coveo/ui-kit/pull/3389

packages/headless/src/controllers/commerce/facets/core/headless-core-facet.test.ts
Spuffynism: Done in https://github.com/coveo/ui-kit/pull/3389

PR #3393, opened 2023-11-14
fix(genai): dispatch like and dislike properly

jelmedini:
[SVCC-3262](https://coveord.atlassian.net/browse/SVCC-3262)

Currently If user clicks Like/Dislike button twice or more (continuously), we will dispatch and log 2 or more UA event.

**FIX:**  
Dispatch like/dislike , and log UA events only when state change.

[SVCC-3262]: https://coveord.atlassian.net/browse/SVCC-3262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathanlb (approved):
LGTM

louis-bompart (approved):

lbergeron (approved):

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
```diff
@@ -113,6 +124,18 @@ describe('generated answer', () => {
     expect(analyticsAction).toBeTruthy();
   });
 
+  it('#dislike dispatches no analytics action when #disliked is set to true', () => {
+    engine.state.generatedAnswer.disliked = true;
+    generatedAnswer.dislike();
+    const action = findAction(likeGeneratedAnswer.type);
+    const analyticsAction = engine.findAsyncAction(
+      logLikeGeneratedAnswer().pending
+    );
```
fbeaudoincoveo, line 133: I don't think this test is passing for the correct reason 😁 

```suggestion
    const action = findAction(dislikeGeneratedAnswer.type);
    const analyticsAction = engine.findAsyncAction(
      logDislikeGeneratedAnswer().pending
    );
```

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -215,13 +215,17 @@ export function buildGeneratedAnswer(
     },
 
     like() {
-      dispatch(likeGeneratedAnswer());
-      dispatch(logLikeGeneratedAnswer());
+      if (!this.state.liked) {
```
fbeaudoincoveo, line 218: Headless controller methods are usually pretty "dumb", in the sense that they don't check conditions before dispatching their actions (and hence their test suites are very straightforward). If any state-based logic should be executed after an action has been dispatched, it should normally be done in the reducer.

Consequently, I believe those "if"'s should be moved to the `generated-answer-analytics-actions.ts` and `generated-answer-slice.ts`, and tested in the corresponding files.

@lbergeron: Out of scope for this PR, but I realize that I missed a similar occurrence when I reviewed your code for the show / hide feature of generated answers.

fbeaudoincoveo (changes requested):

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
lbergeron: @fbeaudoincoveo Thanks for pointing that out. Regarding this statement:

> If any state-based logic should be executed after an action has been dispatched, it should normally be done in the reducer.

Maybe I misunderstand your point but the controller isn't updating the state. It is reading the state to decide whether actions should be dispatched, which looks legit to me. I don't think moving the condition into `likeGeneratedAnswer` and `logLikeGeneratedAnswer` would make sense here.

fbeaudoincoveo: I meant state-based logic including evaluating conditions that are based on the state 😄 

In other words when a controller method is called, it should just "dumbly" dispatch the action, and then the reducer decides what to do with it.

jelmedini: @fbeaudoincoveo in this case the state is already changed when we want to do the check in `generated-answer-analytics-actions.ts` , I mean, the two actions `likeGeneratedAnswer` and `logLikeGeneratedAnswer ` in like/dislike will interfere with each other in this case,  and evaluating the state in different places seems more handy.

lbergeron: @fbeaudoincoveo Regarding this comment:

> In other words when a controller method is called, it should just "dumbly" dispatch the action...

I think the controller should not be limited to only dispatching actions. The controller also has a role to play in making sure the dispatched actions make sense. In this case, `likeGeneratedAnswer` and `logLikeGeneratedAnswer` have to be dispatched together. Dispatching only one of the two actions at a time would not be semantically correct. In my opinion, it  is the best place to implement this check.

> ...and then the reducer decides what to do with it.

In our case, the actions are triggering network requests. A reducer is not the place to have this kind of logic.

nathanlb: I agree with Luc here

fbeaudoincoveo: Alright, I'm convinced 😁

fbeaudoincoveo (approved):

mmitiche (approved):
all good 👌

PR #3394, opened 2023-11-15
fix(quantic): fix citation pop over glitch

mmitiche:
[SFINT-5291](https://coveord.atlassian.net/browse/SFINT-5291)

#### Before

https://github.com/coveo/ui-kit/assets/86681870/bc2436de-1960-470b-86d1-199fb690f88d

#### After

https://github.com/coveo/ui-kit/assets/86681870/43a905b1-13be-4e10-83ee-d46593979bb6

#### The issue
The small space between the citation and the tooltip was the cause of the issue: 
Citation
 
When the mouse is in this highlighted space, the mouse thecnically become outside the citation so the tooltip is no longer displayed.

To solve this issue, I removed the ability to hover over a citation, this would allow us to respect the mockups and avoid the glitchy behaviour, more discussion can be found here: https://coveord.atlassian.net/browse/SFINT-5292?focusedCommentId=706476

[SFINT-5291]: https://coveord.atlassian.net/browse/SFINT-5291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
The fix looks good to me. I just wonder what happens in the case of a keyboard user? Does the tooltip still shows up for screen readers if a user is not using a mouse to hover/focus on the citation? Am I overthinking this? 😅

mmitiche:
> The fix looks good to me. I just wonder what happens in the case of a keyboard user? Does the tooltip still shows up for screen readers if a user is not using a mouse to hover/focus on the citation? Am I overthinking this? 😅

No no it's a good question, for now we are showing the tooltip only on `mouseenter` events but I think it wouldn't be that complicated to support showing the tooltip on focus as well if necessary.
But anyways the screen reader always announces the labels of these buttons which is the relevant information that's displayed inside the tooltip.

erocheleau (approved):

fbeaudoincoveo (approved):

PR #3397, opened 2023-11-15
feat(doc): generate documentation for generated answer in headless, atomic and quantic

lbergeron:
[SVCC-3261](https://coveord.atlassian.net/browse/SVCC-3261)

This PR enables documentation for all controllers and components related to the generated answer feature. It covers the:

* headless controllers
* atomic components
* quantic components

[SVCC-3261]: https://coveord.atlassian.net/browse/SVCC-3261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticCitation/quanticCitation.js
```diff
@@ -7,15 +7,15 @@ const debounceDurationBeforeHoverMs = 200;
...
+ * @category Search
  * @example
  * <c-quantic-citation citation={citation} onclick={handleClick} onhover={handleHover}></c-quantic-citation>
  */
 export default class QuanticCitation extends LightningElement {
   /**
    * @api
    * @type {{title: string, index: number, text: string, clickUri: string}}
-   * The id of the citation.
+   * The citation item information.
```
mmitiche, line 18: thank you @lbergeron 👍

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerCopyToClipboard/quanticGeneratedAnswerCopyToClipboard.js
```diff
@@ -5,7 +5,7 @@ import {LightningElement, api} from 'lwc';
 
 /**
  * The `QuanticGeneratedAnswerCopyToClipboard` component displays a button responsible for copying the generated answer to clipboard.
- * @category Internal
+ * @category Search
```
mmitiche, line 8: I think I would keep this component internal cause the main purpose behind creating it separately was to reduce the complexity  of the main component, the `QuanticGeneratedAnswer` component. 
Thus I think we shouldn't overload our documentation with a component that is unlikely to be handled directly by the end users.

What do you think @lbergeron  @erocheleau  @SimonMilord ?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerRephraseButtons/quanticGeneratedAnswerRephraseButtons.js
```diff
@@ -8,13 +8,13 @@ import {LightningElement, api} from 'lwc';
 
 /**
  * The `QuanticGeneratedAnswerRephraseButtons` component displays a set of rephrase buttons allowing the modification of the answer style of the generated answer.
- * @category Internal
+ * @category Search
```
mmitiche, line 11: Same thing for this component, it's more a component to reduce the complexity of the Quantic Generated Answer and it's not meant to be directly handled by the end users.

mmitiche (commented):
On the Quantic side I would only include the two component: the QuanticGeneratedAnswer and the QuanticCitation components in the documentation.

The Quantic Generated Answer because it's the main component that the end users will add to their search pages and the QuanticCitation component because it's more like a generic component that can be used not in only with RGA but in any situation where displaying a citation would be necessary so it's likely to be useful for the end users.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -48,11 +45,11 @@ export interface GeneratedAnswer extends Controller {
    */
   retry(): void;
   /**
-   * Determines if the generated answer was liked, or upvoted by the end user.
+   * Marks the generated answer as relevant to the end user.
    */
   like(): void;
   /**
-   * Determines if the generated answer was disliked, or downvoted by the end user.
+   * Marks the generated answer as not relevant to the end user.
```
nathanlb, line 52:
```suggestion
   * Marks the generated answer as irrelevant to the end user.
```
I'm not sure about this modification given the dislike action will open the modal and the user can specify what they dislike. Irrelevant is one of the choices but there are others.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -33,7 +33,18 @@ import {SourceCitations} from './source-citations';
...
+ * @part container - The container displaying the generated answer.
+ * @part header-label - The header of the generated answer container.
+ * @part feedback-button - The "like" and "dislike" buttons.
+ * @part toggle - The switch to toggle the visibility of the generated answer.
+ * @part copy-button - The "Copy answer" button.
+ * @part retry-container - The container for the "retry" section.
+ * @part generated-text - The text of the generated answer.
+ * @part citations-label - The header of the citations list.
+ * @part rephrase-label - The header of the rephrase options.
+ * @part rephrase-button - The button for each of the rephrase options (e.g., step-by-step instructions, bullet-point list, summary).
```
srosenberg-coveo, line 47:
```suggestion
 * @part rephrase-button - The button for each of the rephrase options (step-by-step instructions, bulleted list, and summary).
```

srosenberg-coveo (approved):
I only commented suggestions for the first occurrence of each item in the Atomic files, but they should be applied throughout if accepted.

packages/atomic/src/components/search/generated-answer/atomic-citation/atomic-citation.tsx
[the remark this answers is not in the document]
nathanlb: Plus if they're internal we have more freedom to modify them as the feature evolves.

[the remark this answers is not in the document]
lbergeron: One problem I see if we mark the citation and feedback modal components as internal is that implementers will have no documentation for styling the components. They will have to check the DOM and/or the atomic source code to get the information. If this limitation is acceptable, then yes I can keep the child components internal.

As a side note, Smart Snippets is using the same pattern where its feedback modal component is publicly documented.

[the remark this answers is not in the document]
nathanlb: > As a side note, Smart Snippets is using the same pattern where its feedback modal component is publicly documented.
Yes you're right I saw that but I'm not sure why it is 😅 

If it's for parts visibility maybe we could use the [exportParts](https://stenciljs.com/docs/styling#exportparts) attribute as a solution?

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
lbergeron: I tried rephrasing the description to better convey the idea that invoking `like()` does modify the generated answer state to mark it as relevant. The previous description was more like a state retrieval function.

At first, I was using `helpful/not helpful` but went with `relevant/irrelevant` because it was already being used in other descriptions. The idea was to remain consistent.

packages/atomic/src/components/search/generated-answer/atomic-citation/atomic-citation.tsx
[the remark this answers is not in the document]
lbergeron: Thanks for the link. I realize that one has to use `exportparts` to allow styling child components.

> If you have a Stencil component nested within another component, any part specified on elements of the child component will not be exposed through the parent component.

I'll check again with that in mind to make sure the component documentation matches what's exposed.

packages/atomic/src/components.d.ts
```diff
@@ -487,14 +502,26 @@ export namespace Components {
...
+     * The `atomic-generated-answer` component displays a generated answer to a particular query.
+     */
     interface AtomicGeneratedAnswer {
         /**
           * The answer style to apply when the component first loads. Options:   - `default`: Generates the answer without additional formatting instructions.   - `bullet`: Requests the answer to be generated in bullet-points.   - `step`: Requests the answer to be generated in step-by-step instructions.   - `concise`: Requests the answer to be generated as concisely as possible.
          */
         "answerStyle": GeneratedAnswerStyle;
     }
+    /**
+     * The `atomic-generated-answer-feedback-modal` component displays a modal allowing the user to send feedback regarding an irrelevant generated answer.
```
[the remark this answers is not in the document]
lbergeron: It turns out the component will remain internal, but thanks a lot for your suggestion, it is much better indeed.

packages/atomic/src/components/search/generated-answer/atomic-citation/atomic-citation.tsx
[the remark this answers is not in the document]
nathanlb: yeah what the attribute allows is to expose parts of child components as parts of the parent. Not sure how the parts doc is generated but if they're all accessible via the atomic-generated-answer then that might solve that issue for the citations. Wouldn't solve it for the modal i think because we inject that one as a sibling IIRC.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
lbergeron: I love @srosenberg-coveo suggestion to use "the generated answer does not meet the user expectations" to convey the idea of an irrelevant answer. It also solves the conflict between the broad meaning of "irrelevant" versus the specific option in the list.

What would you think of something like this?

```typescript
/**
 * Indicate that the generated answer meets the user expectations.
 */
like(): void;
/**
 * Indicate that the generated answer does not meet the user expectations.
 */
dislike(): void;
```

packages/atomic/src/components/search/generated-answer/atomic-citation/atomic-citation.tsx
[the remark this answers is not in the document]
lbergeron: Very good point, I set the `exportparts` attribute only for the `atomic-citation`.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
nathanlb: love it

nathanlb (approved):
GG

mmitiche (approved):
perfecto!

louis-bompart (approved):

PR #3398, opened 2023-11-15
fix(RGA): no longer showing citation popover on small screens

nathanlb:
It was decided that the citation details popover should not be enabled for mobile devices. This change hides them based on media queries.


packages/quantic/force-app/main/default/lwc/quanticCitation/quanticCitation.html
```diff
@@ -17,7 +17,7 @@
         {citationTitle}
       </p>
     </a>
-    <c-quantic-tooltip light-theme>
+    <c-quantic-tooltip light-theme class="slds-show_medium">
```
mmitiche, line 20: Easy 👌

mmitiche (approved):
👍

packages/quantic/force-app/main/default/lwc/quanticCitation/quanticCitation.html
nathanlb: peasy

fbeaudoincoveo (approved):

PR #3404, opened 2023-11-16
fix(quantic): fix issue with expandable searchbox hiding tabs

SimonMilord:
[SFINT-5288](https://coveord.atlassian.net/browse/SFINT-5288)

**ISSUE:**

The Quantic Expandable Search box currently has a CSS position of absolute making it invisible to the other elements in the page. For example in the Quantic Example Search Box this behaviour is making the expandable search box hide the tabs:

https://github.com/coveo/ui-kit/assets/73316533/688d5125-f9ca-4cba-ab7b-320a162e4205

**FIX:**

Set a `min-height: 3.125rem;` to the `searchbox_floating-container`parent element, which prevents the searchbox's height to collapse when in position absolute.

When expanded (nevermind the clear button, that has been fixed in another PR):
Screenshot 2023-11-27 at 10 20 56 AM

When not expanded (doesn't hide tabs no more):
Screenshot 2023-11-27 at 10 20 38 AM

Standalone:

image

[SFINT-5288]: https://coveord.atlassian.net/browse/SFINT-5288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/templates/expandableStandaloneSearchBox.css
```diff
@@ -26,7 +26,6 @@ textarea.searchbox__input {
 }
 
 .searchbox_floating-container {
-  position: absolute;
```
SimonMilord, line 29: Here I removed the `position: absolute` as well despite being 'standalone' just in the event where some content or element happens to be close to the searchbox and it hides the content/element like it does for the `expandableSearchBox`

mmitiche (commented):
Actually that's not really the expected behaviour that we wanna achieve, we do not want all the page to move down when multiple lines are being written in the search box.

That's the expected behaviour we wanna have:

https://github.com/coveo/ui-kit/assets/86681870/48bb8d23-4d9c-4876-96b4-5a4b87e28382

Only the search box expands without moving any other part in the page.

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/templates/expandableStandaloneSearchBox.css
erocheleau: Did you try it in the header of the community though?
You'll need to make an aura component

```
<aura:component implements="forceCommunity:searchInterface" access="global">
    <c-quantic-standalone-search-box
            engine-id="something"
            textarea>
        </c-quantic-standalone-search-box>
</aura:component>
```

And that should work to have a draggable component in the community header

packages/quantic/force-app/main/default/lwc/quanticSearchBox/templates/expandableSearchBox.css
```diff
@@ -27,7 +27,6 @@ textarea.searchbox__input {
 }
 
 .searchbox_floating-container {
-  position: absolute;
```
erocheleau, line 30: I'm pretty sure I needed this for a good reason though :D 

I'm thinking it might be to work correctly in the header of the SF community as explained in the other comment.

To make it so the expansion of the searchbox doesn't expand also the header.

erocheleau (commented):
Need to have proof it works in the community header.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/templates/expandableSearchBox.css
SimonMilord: yeah that would make a lot of sense

SimonMilord: image
it seems to work

mmitiche (commented):
Can you please elaborate on the logic of your solution, I don't quite understand it

mmitiche (commented):
Now that you are applying a min height to`.searchbox__container-min-height`  which is the parent of `.searchbox_floating-container`
I think you wouldn't even need to play with the positions absolute/relative.

I didn't try it, but I think just adding the CSS to set the min height will be enough to solve this without touching any other thing.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/templates/expandableSearchBox.html
```diff
@@ -1,6 +1,6 @@
 <!-- sldsValidatorIgnore -->
 <template>
-  <div class="slds-form-element z-index-high">
+  <div class="slds-form-element z-index-high searchbox__container-min-height">
```
mmitiche, line 3: I would also call this class simply `searchbox__container` for simplicity and scalability.
imagine in the future we would need to add additional CSS rules to this class, the name will no longer be very relevant.

SimonMilord: I can't since there is already another element with that class name, I could change it to something more generic however or change the name of that other class `searchbox__container`

SimonMilord: changed it to `searchbox__container-wrapper` is that better?

mmitiche: For now yeah that's better, maybe when refactoring the search box it would be worth rechecking these multiple containers wrapping the search box, maybe they can be simplified.

cause having multiple containers around elements is also something a bit fishy, sometimes it is indeed necessary but it's worth rechecking for a potential simplification.

mmitiche (commented):
What about the Quantic Standalone Search Box, I think it will also need this new CSS class in order to have a min height and avoid hiding what's below it.

mmitiche (approved):
👍

PR #3282, opened 2023-10-13
feat(relay): add support for analytics/result/open

louis-bompart:
https://github.com/coveo/analytics-migration-project/issues/54

PR #3405, opened 2023-11-16
feat(relay): add type safety for event types

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2892

All changes in analytics-util are needed for type safety.

Example of it working -> https://github.com/coveo/ui-kit/pull/3282


packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -775,11 +775,12 @@ export const validateProductRecommendationPayload = (
   productRec: ProductRecommendation
 ) => new Schema(productRecommendationPartialDefinition).validate(productRec);
 
-async function logNextEvent<PayloadType extends RelayPayload>(
+async function logNextEvent<PayloadType>(
   emitEvent: ReturnType<typeof createRelay>['emit'],
   type: string,
   payload: PayloadType
 ): Promise<void> {
+  //@ts-ignore
```
alexprudhomme, line 783: This ts-ignore will be removed once Relay adjust the emitEvent function. It currently uses the RelayPayload type internally

samisayegh: Hi Alex, I've created [this jira](https://coveord.atlassian.net/browse/LENS-1349). Could you add more details about the error on the jira? Could it be because the interface is using `unknown` rather than `any`?

louis-bompart, line 783:
```suggestion
  //@ts-except-error
```

louis-bompart: Good work've been done on Relay side.
Morphed the ignore into an except-error, the flag will get rise when no error occurs like this.
We can revisit then when the flag is raised/relay is updated. Merging.

louis-bompart (approved):

PR #3406, opened 2023-11-16
feat(IPX): make button load recs on click

dguerinCoveo:
Sets the onClick behaviour directly in the atomic-ipx-button. This allows us to have much cleaner code in the SIS in the loader script.

### DIFF
--- a/packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
+++ b/packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
@@ -1,3 +1,4 @@
+import {loadRecommendationActions} from '@coveo/headless/recommendation';
 import {Component, Element, Fragment, h, Prop, State} from '@stencil/core';
 import CloseIcon from '../../../images/close.svg';
 import SearchIcon from '../../../images/search.svg';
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/ipx/atomic-ipx-button/atomic-ipx-button.tsx
```diff
@@ -45,7 +46,24 @@ export class AtomicIPXButton implements InitializableComponent {
...
+  private async loadRecommendations() {
+    const recsEngine = this.bindings.interfaceElement.querySelector(
+      'atomic-recs-interface'
+    )?.engine;
+    if (recsEngine) {
+      this.recommendationsLoaded = true;
+      recsEngine.dispatch(
+        loadRecommendationActions(recsEngine).getRecommendations()
+      );
+    }
```
dguerinCoveo, line 60: @louis-bompart do you think this is legit?

louis-bompart: Confused, why would the engine be not there?

dguerinCoveo: It should always be there. We always put the recs engine in the IPX in the SIS and we are the only users of these components

louis-bompart: so why the if?

dguerinCoveo: Just because the addition of the recs engine kinda requires a manual step since it has to be in the html. 

But my question was more: is it okay to put a dependency to the `recs-engine` inside the `ipx-button` component. I think it is, since the IPX components are more of an internal lib of which the SIS is the only user, so we can make it work the way we like. From your comments I understand you don't have any issues with it, so I'll open the PR.

louis-bompart: It will increase the size of the file that you need to import to instantiate the button. But if, as owner of the IPX, you're fine with this, eh, your call :D

pathiery (approved):

mikegron (approved):

fbeaudoincoveo (approved):

PR #3411, opened 2023-11-17
chore: new release lock process+tweak release branch update

louis-bompart:
Leveraging the new [Ruleset](https://github.blog/2023-10-11-enforcing-code-reliability-by-requiring-workflows-with-github-repository-rules/) feature of GitHub, we have a few new tools in the shed to create the "main trunk bouncer" to protect master during releases.

The lifecycle of the release remains unchanged, however, instead of relying on the unreliable git-lock file which created pollution and confusion, we can now rely on this:
image

This comes on top of the branch protection we have for normal operations (and both as to be ✅ to merge)

Also include the fix for the release/v2 'update' that was not working.


utils/release/git-publish-all.mjs
```diff
@@ -73,12 +73,8 @@ process.chdir(process.env.INIT_CWD);
...
   // TODO v3: Bump to release/v3
   const currentReleaseBranch = 'release/v2';
-  await gitCheckoutBranch('release/v2');
-  await gitSetRefOnCommit(
-    'origin',
-    `refs/heads/${currentReleaseBranch}`,
-    commit
-  );
+  await gitCheckoutBranch(currentReleaseBranch);
+  await gitReset({resetMode: 'hard', ref: commit});
```
louis-bompart, line 77: This is the fix for why the release is red (and requires manual ops to complete)

louis-bompart:
What it looks like when the branch is locked out:
image
When clicking 'View rules'
image

fbeaudoincoveo (approved):

PR #3419, opened 2023-11-20
refactor(context): one context to bind them all

louis-bompart:
It was all the same darn things.


fbeaudoincoveo (commented):
> It was all the same darn thing.

Is it possible that those "duplicates" exist to ensure that the context controller appears in the documentation of each engine?

E.g., https://docs.coveo.com/en/headless/latest/reference/recommendation/controllers/

I wonder if removing those will also remove the docs at the same time.

alexprudhomme (approved):
Concerning the fred comment. IMO, this is fine. I looked into what the documentation would look like by checking the `packages/headless/dist/parsed_doc.json`. The buildContext controller is still in every documentation use case. It only requires a CoreEngine instead of its original engine. This is fine since it doesn't need the specific arguments of those engine anyway.

image

alexprudhomme:
Could do the same for the FacetManager 😄

louis-bompart:
> > It was all the same darn thing.
> 
> Is it possible that those "duplicates" exist to ensure that the context controller appears in the documentation of each engine?
> 
> E.g., https://docs.coveo.com/en/headless/latest/reference/recommendation/controllers/
> 
> I wonder if removing those will also remove the docs at the same time.

I think it should still be available because we still export stuff through the various indexes.

PR #3422, opened 2023-11-21
feat(commerce): add search analytics

Spuffynism:
Add commerce search analytics events.

~⚠️ I'm not sure what I should be adding for the `analyticsType` and `analyticsPayloadBuilder` properties of the search actions. 🤔~ Edit: We'll tackle this in CAPI-325 and CAPI-326

[CAPI-244]

[CAPI-244]: https://coveord.atlassian.net/browse/CAPI-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/commerce/search/search-analytics-actions.ts
```diff
@@ -0,0 +1,37 @@
...
+import {OmniboxSuggestionMetadata} from '../../query-suggest/query-suggest-analytics-actions';
+
+export const logInterfaceLoad = (): CommerceSearchAction =>
+  makeCommerceAnalyticsAction({
+    prefix: 'analytics/commerce/interface/load',
+    __legacy__getBuilder: (client) => client.makeInterfaceLoad(),
+    analyticsType: '',
+    analyticsPayloadBuilder: (_state) => {
+      return {};
+    },
```
Spuffynism, line 14: What goes in these? Looking at the [analytics schema](https://github.com/coveo/analytics_schema/tree/master/schemas/client), I don't see a "commerce search" event.

fbeaudoincoveo: I'm not sure: is there a plan to define one, or are we using the [regular search schemas](https://github.com/coveo/analytics_schema/tree/master/schemas/internal/api.search.search/2)?

@pvanool , thoughts?

pvanool: It's currently on pause. We definitely want to go forward with it in the future. Is this something that we can easily add when we have a schema ready?

Spuffynism: As stated in the PR description, we'll tackle this in CAPI-325 and CAPI-326

louis-bompart (changes requested):
The events are sent through the old stack, that ain't gonna fly, even for the purpose of the demo.
We can build subsequent PRs on top, but for Search CMH, I don't think we should merge that, as, afaik, search reported this way won't wwork down the line (i.e in the CMH UI).

Spuffynism:
@louis-bompart Ok, closing this until we figure out how analytics should work for headless commerce. https://coveord.atlassian.net/browse/CAPI-325 and https://coveord.atlassian.net/browse/CAPI-326 will cover this follow-up work.

PR #3426, opened 2023-11-22
Search migration/facet select

alexprudhomme:


alexprudhomme:
I have decided to implement both facetSelect & facetDeselect in this PR because of cases like this.

image

There are multiple places in the code where we use a function like this that calls both logFacetSelect and logFacetDeselect.

alexprudhomme:
I decided to also migrate facetExclude at the same time, since it is intertwined with facetDeselect in cases like this :

image

packages/headless/src/features/facets/range-facets/generic/range-facet-insight-utils.ts
```diff
@@ -21,15 +17,3 @@ export const getInsightAnalyticsActionForToggleRangeFacetSelect = (
...
-  facetId: string,
-  selection: RangeFacetValue
-) => {
-  const facetValue = `${selection.start}..${selection.end}`;
-  const payload: FacetSelectionChangeMetadata = {facetId, facetValue};
-
-  return isRangeFacetValueExcluded(selection)
-    ? logFacetDeselect(payload)
-    : logFacetExclude(payload);
-};
```
alexprudhomme, line 35: This function was added by mistake during the implementation of facet exclusion. It is never actually used.

louis-bompart (approved):

louis-bompart:
We might need to revisit depending on the upcoming discussion regarding how we send the facet_data thingies, but right now, lgtm

PR #3427, opened 2023-11-22
feat(quantic): Better handling of Salesforce Result links

erocheleau:
SFINT-5261

- Set proper href on QuanticResultLink when it's a SF result. 
- Updated labels because they didn't make sense.
- Updated test and added a test

**Before:**
The QuanticResultLink, when you had a Salesforce result on right click > open would navigate to the Salesforce internal url still.

**After**
If the result is a Salesforce record (has a raw.sfid field value).
If the result is part of your current Salesforce org:
image
The href will be properly set, and the navigation will still use the correct Salesforce NavigateTo API.
Same if you are in the console:
image

If the result is NOT part of your current Salesforce org:
There is no `href` set on the result, on click we will still try to NavigateTo with Salesforce.
But there's no more option to right click > open
image

The other option is to have `javascript:void(0)` as the href. I think this may be better for accessibility purposes, but it still lets you right click > open, but it will open a void page.


packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1092,6 +1092,13 @@
       <protected>false</protected>
       <shortDescription>No related items</shortDescription>
     </labels>
+    <labels>
+      <fullName>quantic_NavigateToRecord</fullName>
```
SimonMilord, line 1096: do we not want a translation for this label also or is that an english only label?

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -1,5 +1,5 @@
+import navigateToRecord from '@salesforce/label/c.quantic_NavigateToRecord';
 import opensInBrowserTab from '@salesforce/label/c.quantic_OpensInBrowserTab';
-import opensInSalesforceSubTab from '@salesforce/label/c.quantic_OpensInSalesforceSubTab';
```
SimonMilord, line 2: we should also remove the `quantic_OpensInSalesforceSubTab` label and its translation since they will now not be used anywhere in the package

mmitiche: Removing a label that was previously part of the package could create a breaking change. This I wouldn't remove it now. it would be safe to remove in the next v3.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
mmitiche: translation added.

SimonMilord (approved):

PR #3428, opened 2023-11-23
feat(quantic): quantic tooltip made responsive

mmitiche:
[SFINT-5287](https://coveord.atlassian.net/browse/SFINT-5287)

- The Quantic Tooltip component is now made more intelligent, it automatically reacts to the size and the scroll position of the screen.

## Three main features are now supported:

1. #### Showing the tooltip above or below the target according to the scroll position.
2. #### Adjusting the horizontal position of the tooltip to make it always appear and visible inside the screen.
3. #### Adjusting the max width of the tooltip to the width of the screen in smaller screens, to avoid having a tooltip larger than the actual window.

## Demo

- ### Before:

https://github.com/coveo/ui-kit/assets/86681870/b9ddafb8-f948-44ad-ae75-fd77a572f5b4

Screenshot 2023-11-13 at 10 13 27 AM

- ### After:

https://github.com/coveo/ui-kit/assets/86681870/d1386a93-9353-4972-af02-a6b0cf5cfa09

Screenshot 2023-11-23 at 7 42 15 AM

[SFINT-5287]: https://coveord.atlassian.net/browse/SFINT-5287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
```diff
@@ -1,24 +1,57 @@
+:host {
+  --x-translation: var(--adapted-x-translation, 0px);
+  --max-width: var(--adapted-max-width, 20rem);
+}
+
 .tooltip__content {
   left: 50%;
-  transform: translate(-50%, -100%);
+  transition: transform 0.05s linear 0.05s, opacity 0.1s linear 0s, visibility 0.1s linear 0s;
```
erocheleau, line 8: I think this is causing the <> under the tooltip to be visible for a fraction of a second when the tooltip is first rendered.
Can you apply the same transition to the <> as well?

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
```diff
@@ -1,24 +1,57 @@
...
+  z-index: 2;
+}
+
+.tooltip__content--positioned-above {
+  transform: translate(calc(-50% + var(--x-translation)), -100%);
```
erocheleau, line 13: Can we split the translate in to translateX and translateY just for clarity?

You could have one single class 

```
.tooltip__content {
  transform: translateX(calc(-50% + var(--x-translation)));
}

.tooltip__content--positioned-above {
  transform: translateY(-100%);
}

.tooltip__content--positioned-below {
  transform: translateY(0.7rem);
}
```

Not sure that works, but it feels more readable to me

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
```diff
@@ -1,24 +1,57 @@
...
   width: max-content;
-  max-width: var(--slds-c-tooltip-sizing-max-width, var(--sds-c-tooltip-sizing-max-width, var(--lwc-sizeMedium, 20rem)));
 }
 
-.slds-nubbin_bottom-left:before,
-.slds-nubbin_bottom-left:after {
+.tooltip__arrow {
   left: 50%;
+  transform: translate(-50%, -100%);
+  transition: opacity 0.08s linear 0s, visibility 0.06s linear 0s;
```
erocheleau, line 34:
```suggestion
  transition: opacity opacity 0.1s linear 0s, visibility 0.1s linear 0s;
```
I think your two transitions should be the same delay so the tooltip arrow isn't visible before the content?

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
```diff
@@ -1,24 +1,57 @@
...
+.tooltip__arrow--light {
+  background-color: var(--lwc-colorBackgroundAlt, rgb(255, 255, 255));
+}
+
+.tooltip__arrow--positioned-above {
+  top: -0.7rem;
+}
+
+.tooltip__arrow--positioned-below {
+  transform: translate(-50%, 0.7rem);
```
erocheleau, line 51: Why is the arrow below translated -50% in X?

packages/quantic/force-app/main/default/lwc/quanticCitation/quanticCitation.html
```diff
@@ -17,7 +17,7 @@
         {citationTitle}
       </p>
     </a>
-    <c-quantic-tooltip light-theme class="slds-show_medium">
+    <c-quantic-tooltip light-theme class="slds-show_medium" target={template.host}>
```
erocheleau, line 20: Does it really need to be the host? can't it be like the `<li> element`?

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
```diff
@@ -30,23 +38,127 @@ export default class QuanticTooltip extends LightningElement {
...
+    this.resizeTimer = setTimeout(() => {
+      this.updateTooltipMaxWidth();
+      this.updateTooltipVerticalPosition();
+    }, 300);
+  };
+
+  updateTooltipMaxWidth() {
+    const windowWidth = window.innerWidth;
+    const styles = this.template.host?.style;
+    styles.setProperty('--adapted-max-width', `${windowWidth}px`);
```
erocheleau, line 84: Have you tried it on the hosted insight panel when the tab isn't active?
It feels like we're going to have the same issues than with the tab bar, the tooltip won't have it's proper width set because the innerWidth will be 0 for the non-active tab

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
```diff
@@ -30,23 +38,127 @@ export default class QuanticTooltip extends LightningElement {
...
+  /** @type {NodeJS.Timeout} */
+  resizeTimer;
+  /** @type {boolean} */
+  displayTooltipAboveTarget = true;
   /** @type {boolean} */
   isVisible = false;
 
+  connectedCallback() {
+    window.addEventListener('resize', this.handleResizeAndScroll);
+    window.addEventListener('scroll', this.handleResizeAndScroll);
```
erocheleau, line 56: I'm curious why aren't all the calculations actually done only when we show the tooltip?

Like before showing the tooltip, calculate all the things, then add the class to make it visible?

erocheleau (commented):
Some questions about the innerworkings of when we do the calculations, but other than that it looks great.

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
mmitiche: Yea I can adjust the transition of the small arrow

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
mmitiche: Unfortunately this will not work cause the transform attributes will be overwritten between these three classes, 
True that each class modify a given axis but they all are under the CSS attribute `transform` so for example  if you add `tooltip__content--positioned-above` you will lose the translateX of `tooltip__content`.

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
mmitiche: so it stays in the middle of the tooltip

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
mmitiche: will give it a try and see 👍

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
mmitiche: That's a good question actually cause this was one of the main problems I encountered. 
My initial idea was like you described, do the calculation when we are about to show the tooltip but then got a serious issue kinda hard to imagine but I will do my best x)

basically this is about moving the tooltip  in the y axis, to show it above the target or below it.
Now imagine we are doing the calculations  when we are about to show the tooltip and we are in a situation where the tooltip needs to be displayed `below`.

That's what will happen:

- Initially the tooltip is hidden and is above the target.
- Step01: The user hovers over the target.
- We execute the `showTooltip` method which is responsible for doing the calculation and displaying the tooltip.
- We do the calculation, we found that the tooltip needs to be displayed below so we apply a translate the tooltip in the y axis.
- while this transition is happening we display the tooltip.
- at a certain fraction of second while doing this y axis translation and since the tooltip will be transition from above to below, at a certain fraction of second the tooltip will cover the target, this will yield a `mouseleave` event on the target cause technically the mouse is no longer on the target now that it is covered.
- on `mouseleaver` the target calls `hideTooltip` method.
- `hideTooltip` method resets the position of the tooltip
- we come back to Step 01...

This was the infinite loop that was occuring, that's why I switched to doing the calculation related to the y axis on resize and scroll and not on hover, so the tooltip is already on its correct y position when the user hovers.

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
SimonMilord: I agree here, it would prevent this glitchy behaviour:
image

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
SimonMilord: just setting bottom: 0.7rem; doesnt keep it centered like when positioned above?

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
erocheleau: gotcha yea I wasn't sure, all good then

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
erocheleau: but why isn't it also applied to`tooltip__arrow--positioned-below` that translate -50% in X?

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
```diff
@@ -16,12 +17,19 @@ export default class QuanticTooltip extends LightningElement {
...
   @api lightTheme = false;
+  /**
+   * The element targeted by the tooltip.
+   * @api
+   * @type {Element}
+   */
+  @api target;
   /**
    * @api
    * Method tha shows the tooltip.
```
SimonMilord, line 28:
```suggestion
   * Method that shows the tooltip.
```

packages/quantic/force-app/main/default/lwc/quanticCitation/quanticCitation.html
erocheleau: Yea I just was thinking maybe we instead rename that property to something like `tooltip-for` and we document it so that people can actually use it without having to know what the template.host is :)

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
erocheleau: Lol so basically you're telling me we would only need to add a delay before displaying the tooltip corresponding to the transition time to support the case where we would only do the calculations at the moment where we want to display it?

I think I would prefer that than having something that constantly calculates ALL tooltips positions while scrolling etc. no?

Because right now for each instance of a tooltip in the page you're adding a scroll and resize listener, IMO that's pretty costly.

I would just update the showTooltip to add a timer to only show after the transition is finished.
Or is it possible to simply not have a transition on the translation? The transition of the tooltip imo only should be on the opacity right?

mmitiche: Of course it would be better to re-calculate on each resize/scroll, let me just try it again  cause I'm kinda sure this wasn't the only issue that made change this idea.

I will it try it again and intensively test and if it works as expected it would be better indeed.

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
mmitiche: css value adjusted.

packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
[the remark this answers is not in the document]
mmitiche: nah they are not, just a typo.

PR #3432, opened 2023-11-24
fix(headless): import correct analytics in product listing

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2908

This will simplify the migration process of search events.


louis-bompart (approved):
This makes sense to me, I think we should verify with whoever own the product-listing for a stamp of approval.

alexprudhomme:
> This makes sense to me, I think we should verify with whoever own the product-listing for a stamp of approval.

Yea I can verify, but I really thing this was just an error in the implementation. In certain files, they were importing from both web file and product listing file.

alexprudhomme:
@thudon-coveo, hi I see in the commit history that you worked on product listing. Can you confirm that what I am doing here doesn't break anything ?

thudon-coveo (approved):

PR #3434, opened 2023-11-27
doc(headless): update description for atomic field condition

eaguilar-coveo:
[JIRA LINK](https://coveord.atlassian.net/browse/DOC-13438)

* Update to the `atomic-field-condition` description
* Update the `condition` property description.

For the `condition` property example, I looked further into it and the use case seem to be valid because the existing markups `must-match` `must-not-match` looks for the whole string value. In the `condition` example, it looks for a substring ("Singapore") within the value.

The changes in this repo are only the parts where the docs site uses to generate the documentation page for the component.

The full changes including the other examples mentionned in the JIRA ticket were made on the documentation site. Seen in this [PR](https://github.com/coveo/doc_jekyll-public-site/pull/10115).


packages/atomic/src/components.d.ts
```diff
@@ -374,12 +374,13 @@ export namespace Components {
     }
```
fbeaudoincoveo, line 374: This file is automatically generated by the Stencil compiler, so you can revert those changes 😄

fbeaudoincoveo (approved):
Just revert the changes in components.d.ts and it's good to go!

packages/atomic/src/components.d.ts
eaguilar-coveo: Got it, thanks 😃

For reverting, do you mean to let the Stencil compiler automatically update this file or have the PR just have one file changed being: `atomic-field-condition/atomic-field-condition.tsx`?

I noticed too that it made changes to the `components.d.ts ` file under the "Add generated files" commit and when I reverted those changes, that commit showed up again after pushing the changes.  😅

I did see for a brief moment the number of changed files in the PR go from 2 to 1, but then the `components.d.ts` got updated again.

image
image

fbeaudoincoveo: oh nevermind me again, it's all good 😅

fbeaudoincoveo (approved):

alexprudhomme (approved):

PR #3436, opened 2023-11-28
fix: decode HTML content

louis-bompart:
Revert most of #3134.
I don't re-add the dependencies because TextDecoder/Encoder is a common web feature.
Yes, there's the question of React Native:
 - I confirmed with @olamothe in October: we don't support React Native
 - The documentation agrees: No trace of React Native (except some Qubit stuff ;) )
 
So, if anyone wants to use React-Native comes up about this, our answer will be:
1. We don't support React Native
2. If you still want to try, our best advice is: to try to polyfill all the missing libraries.

### DIFF
--- a/packages/headless/src/api/search/encoding-finder.test.ts
+++ b/packages/headless/src/api/search/encoding-finder.test.ts
@@ -1,0 +1,22 @@
+import {findEncoding} from './encoding-finder';
+
+describe('#findEncoding', () => {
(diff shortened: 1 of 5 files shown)


fbeaudoincoveo (approved):
I gather this information was incorrect or misunderstood: https://github.com/coveo/ui-kit/pull/3134#discussion_r1307799448 😕

PR #3439, opened 2023-11-29
feat(headless): create article insight analytics action created

SimonMilord:
[SFINT-5277](https://coveord.atlassian.net/browse/SFINT-5277)

- Created the headless Create Article analytics action required to support the Create Article Button for the insight use case.
- Added unit tests to cover this new action
- Updated the version of `coveo.analytics.js` from 2.28.21 to 2.28.23 in the headless package to be able to access the `logCreateArticle` event added last week.

image

[SFINT-5277]: https://coveord.atlassian.net/browse/SFINT-5277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/create-article/create-article-insight-analytics-actions.ts
mmitiche, line 1: I don't think we need a whole new file to add this action, we already have some files that serves for defining the Headless actions specific to the insight panel: [insight-analytics-actions.ts](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/features/analytics/insight-analytics-actions.ts) you can add the new action directly here.

packages/headless/src/features/create-article/create-article-insight-analytics-actions.ts
```diff
@@ -0,0 +1,16 @@
+import {CreateArticleMetadata} from 'coveo.analytics';
+import {
+  InsightAction,
+  makeInsightAnalyticsAction,
+} from '../analytics/analytics-utils';
+import {getCaseContextAnalyticsMetadata} from '../case-context/case-context-state';
+
+export const logCreateArticle = (
```
mmitiche, line 8: In order to have the ability to dispatch this action from an LWC, this action needs to be included in the Insight Analytics Actions Loader: [insight-analytics-actions-loader.ts](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/features/analytics/insight-analytics-actions-loader.ts)

packages/headless/src/features/create-article/create-article-insight-analytics-actions.ts
SimonMilord: alright, moved the action to the insight-analytics-actions.ts :)

packages/headless/src/features/create-article/create-article-insight-analytics-actions.ts
SimonMilord: added it, will try it in a quantic component

packages/headless/src/features/analytics/insight-analytics-actions.test.ts
```diff
@@ -101,3 +108,39 @@ describe('logInterfaceChange', () => {
...
+    );
+    const expectedCaseContext = {
+      caseContext: {
+        Case_Subject: exampleSubject,
+        Case_Description: exampleDescription,
+      },
+      caseId: exampleCaseId,
+      caseNumber: exampleCaseNumber,
+    };
+    expect(mockLogCreateArticle).toHaveBeenCalledTimes(1);
```
mmitiche, line 138: In the other two tests of this file we are still using `toBeCalledTimes` which is I think is deprecated, could you please replace `toBeCalledTimes` by `toHaveBeenCalledTimes` since we are already here modifying this file.

packages/headless/src/features/analytics/insight-analytics-actions.ts
```diff
@@ -15,3 +20,13 @@ export const logInsightInterfaceChange = (): InsightAction =>
       interfaceChangeTo: state.configuration.analytics.originLevel2,
     });
   });
+
+export const logInsightCreateArticle = (
+  createArticleMetadata: CreateArticleMetadata
+): InsightAction =>
+  makeInsightAnalyticsAction('analytics/createArticle', (client, state) => {
```
mmitiche, line 27:
```suggestion
  makeInsightAnalyticsAction('analytics/insight/createArticle', (client, state) => {
```

mmitiche (approved):
LGTM 👍

louis-bompart (commented):
Are we sure we want to do this?
Ideally at this stage, I'd like to not see new legacy/cajs events being added.
But let's assume we want to do it, what I'd ask for sure is for the  event to be assessed by @felixpelletier team, to confirm that the event will survive.
If we're adding it to kill it off with v3, let's not add it.

felixpelletier:
> Are we sure we want to do this? Ideally at this stage, I'd like to not see new legacy/cajs events being added. But let's assume we want to do it, what I'd ask for sure is for the event to be assessed by @felixpelletier team, to confirm that the event will survive. If we're adding it to kill it off with v3, let's not add it.

We're fine with it if it's specific to the insight panel (and not a duplicate of another event just with more context).
But you're right, we will have to migrate this to v3. If it's just for internal use, it's not too bad. We just won't backport the custom ua event if no customer use it.

louis-bompart:
> > Are we sure we want to do this? Ideally at this stage, I'd like to not see new legacy/cajs events being added. But let's assume we want to do it, what I'd ask for sure is for the event to be assessed by @felixpelletier team, to confirm that the event will survive. If we're adding it to kill it off with v3, let's not add it.
> 
> We're fine with it if it's specific to the insight panel (and not a duplicate of another event just with more context). But you're right, we will have to migrate this to v3. If it's just for internal use, it's not too bad. We just won't backport the custom ua event if no customer use it.

Eh... I'm not fond of the idea of pushing back the decision to keep or remove to later.
We're on the eve of v3, so if we add it, we will migrate it. If we don't want to migrate it, we don't add it.
Otherwise, you'd adding a contract/api for the public to use just to remove it later... That'd be a bad DevExp, so I'd have to veto it :)

louis-bompart:
> > > Are we sure we want to do this? Ideally at this stage, I'd like to not see new legacy/cajs events being added. But let's assume we want to do it, what I'd ask for sure is for the event to be assessed by @felixpelletier team, to confirm that the event will survive. If we're adding it to kill it off with v3, let's not add it.
> > 
> > We're fine with it if it's specific to the insight panel (and not a duplicate of another event just with more context). But you're right, we will have to migrate this to v3. If it's just for internal use, it's not too bad. We just won't backport the custom ua event if no customer use it.
> 
> Eh... I'm not fond of the idea of pushing back the decision to keep or remove to later.
> We're on the eve of v3, so if we add it, we will migrate it. If we don't want to migrate it, we don't add it.
> Otherwise, you'd adding a contract/api for the public to use just to remove it soon after... That'd be a bad DevExp, so I'd have to veto it :) 

Talked with @tedre191 & @felixpelletier , event will be used for customers reporting & migrate to v3. Lgtm

PR #3449, opened 2023-12-01
facet(commerce): add commerce numeric facets

fbeaudoincoveo:
https://coveord.atlassian.net/browse/CAPI-91

This also involved quite a bit of refactoring to minimize code duplication.


packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.test.ts
```diff
@@ -33,6 +34,8 @@ describe('facet-set slice', () => {
     expect(finalState).toEqual({});
   });
 
+  // TODO: #toggleSelectNumericFacetValue + #toggleExcludeNumericFacetValue tests
```
fbeaudoincoveo, line 37: I'll add these tests next week.

packages/headless/src/commerce.index.ts
```diff
@@ -92,20 +92,14 @@ export {
   SortDirection,
```
fbeaudoincoveo, line 92: I haven't yet worked on the Barca Sports integration, so I might end up exporting more (but probably not less). I'm starting with what I believe will be the bare minimum.

packages/headless/src/controllers/commerce/facets/core/generator/headless-commerce-facet-generator.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/**
+ * @internal
+ *
+ * The `CommerceFacetGenerator` options used internally.
+ */
+export interface CommerceFacetGeneratorOptions {
+  buildRegularFacet: CommerceRegularFacetBuilder;
+  buildNumericFacet: CommerceNumericFacetBuilder;
```
fbeaudoincoveo, line 49: The addition of this option is the main change in this controller.

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -27,66 +31,93 @@ import {buildController} from '../../../controller/headless-controller';
...
+  options: CoreCommerceFacetOptions;
 }
 
-export interface FacetOptions {
-  /**
-   * A unique identifier for the controller.
-   * */
+export interface CoreCommerceFacetOptions {
   facetId: string;
+  toggleSelectActionCreator?: ActionCreatorWithPreparedPayload<
```
fbeaudoincoveo, line 62: This is the big change here: the core facet controller accepts action creators.

I made them optional because I'm not sure how to make them pass schema validation... but since this will be 100% handled internally, I think we probably don't even need to validate (or publicly expose) the props of this controller at all.

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -125,14 +156,34 @@ export function buildCoreFacet(
   return {
     ...controller,
 
-    toggleSelect: (selection: FacetValue) =>
-      dispatch(toggleSelectFacetValue({facetId, selection})),
+    toggleSelect: (selection: AnyFacetValue) => {
+      if (props.options.toggleSelectActionCreator) {
```
fbeaudoincoveo, line 160: If I make the action creator props mandatory, that'll get rid of this awkwardness.

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -125,14 +156,34 @@ export function buildCoreFacet(
...
+        dispatch(props.options.toggleSelectActionCreator({selection, facetId}));
+        if (props.options.fetchResultsActionCreator) {
+          dispatch(props.options.fetchResultsActionCreator());
+          // TODO: analytics
```
fbeaudoincoveo, line 164: I got rid of analytics for facets altogether for the moment.

I think it's a waste of time to dispatch the old client-side events if we're never going to support them in the new commerce sub-package, and we need the new stack for the NRF demo anyways.

packages/headless/src/commerce.index.ts
```diff
@@ -92,20 +92,14 @@ export {
...
 
+export type {CommerceRegularFacet} from './controllers/commerce/facets/core/regular/headless-commerce-regular-facet';
+export type {CommerceNumericFacet} from './controllers/commerce/facets/core/numeric/headless-commerce-numeric-facet';
 export type {
-  FacetGenerator,
-  FacetGeneratorState,
-} from './controllers/commerce/facets/core/headless-core-facet-generator';
-export type {
-  FacetValue,
+  RegularFacetValue,
```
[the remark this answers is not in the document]
louis-bompart: TextFacetValue?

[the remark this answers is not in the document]
louis-bompart: Or maybe even just FacetValue :)
Avoid duplication of terms is important for the DX I think, so maybe let's pull in someone from the @coveo/api-writers ?

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
louis-bompart: Is that for analytics mainly? or something else?

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
```diff
@@ -113,24 +180,62 @@ function updateStateFromFacetResponse(
...
+        ...(valuesAfter as NumericRangeRequest[]),
+      ];
+      break;
+    case 'dateRange':
+      // TODO
+      break;
+    case 'hierarchical':
+      // TODO
+      break;
+  }
```
louis-bompart, line 238: If this switch job is to define the values, to simplify the code and make it easier to read, might be good to put it in another func

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.test.ts
```diff
@@ -1,6 +1,7 @@
 import {commerceFacetSetReducer as commerceFacetSet} from '../../../../features/commerce/facets/facet-set/facet-set-slice';
```
fbeaudoincoveo, line 1: @Spuffynism I didn't really change anything in this test suite (yet).

I'm having a look at it this morning (Monday, Dec. 4) to see if the test coverage is still OK given the refactoring I've done.

packages/headless/src/controllers/commerce/product-listing/facets/headless-product-listing-facet-generator.ts
```diff
@@ -1,20 +1,34 @@
...
+ * The `ProductListingFacetGenerator` headless controller creates commerce facet controllers from Commerce API
+ * product listing responses.
  *
- * @param engine - The headless engine.
- * @returns A `FacetGenerator` controller instance.
+ * Commerce facets are not requested by the implementer, but rather pre-configured through the Coveo Merchandising Hub
+ * (CMH). The implementer is only responsible for leveraging the facet controllers created by this controller to
+ * properly render facets in their application.
+ */
+export interface ProductListingFacetGenerator extends CommerceFacetGenerator {}
```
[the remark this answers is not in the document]
fbeaudoincoveo: That being said, this is a change we can do a bit later (but before we make the commerce sub-package publicly available) 😄

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
```diff
@@ -35,14 +49,39 @@ export const commerceFacetSetReducer = createReducer(
...
+
+        facetRequest.preventAutoSelect = true;
+
+        const existingValue = (
+          facetRequest.values as WritableDraft<NumericRangeRequest>[]
+        ).find(
+          (req) =>
+            req.start === selection.start &&
+            req.end === selection.end &&
+            req.endInclusive === selection.endInclusive
```
fbeaudoincoveo, line 84: In the findRange method in range-facet-reducers.ts, we're not considering the endInclusive value when finding the range value. I added this check because 0-3 end inclusive is completely different from 0-3 not end inclusive.

packages/headless/src/commerce.index.ts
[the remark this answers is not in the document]
npushkarskii: I personally think that Louis' suggestion of `TextFacetValue` is on point 👍 @fbeaudoincoveo

Spuffynism (commented):
Nice, I really like having a core facet controller with action creators :+1:.

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
```diff
@@ -35,14 +49,39 @@ export const commerceFacetSetReducer = createReducer(
         const {facetId, selection} = action.payload;
         const facetRequest = state[facetId]?.request;
 
-        if (!facetRequest) {
+        if (!facetRequest || facetRequest.type !== 'regular') {
```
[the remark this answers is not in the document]
fbeaudoincoveo: Otherwise we're implicitly trusting that the correct toggle select action was dispatched for that facet.

If actions could only be dispatched from our controllers I suppose that would be OK. However, let's suppose the action is dispatched from custom code, nothing would prevent an implementer from dispatching a toggleSelectFacetValue on, say, a numeric facet value. If we don't return here, we might end up doing some non-sensical state updates.

Thoughts?

packages/headless/src/commerce.index.ts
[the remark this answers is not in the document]
fbeaudoincoveo: @Spuffynism , @y-lakhdar , any strong opinions on this?

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -0,0 +1,286 @@
...
+    never,
+    never
+  >;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  fetchResultsActionCreator: () => AsyncThunkAction<unknown, void, any>;
+}
+
+export type CoreCommerceFacet<
+  T1 extends AnyFacetValueRequest,
+  T2 extends AnyFacetValueResponse,
```
Spuffynism, line 77:
```suggestion
  ValueRequest extends AnyFacetValueRequest,
  ValueResponse extends AnyFacetValueResponse,
```

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -0,0 +1,286 @@
...
+   * The facet values
+   */
+  values: T[];
+};
+
+export type CoreCommerceFacetBuilder = typeof buildCoreCommerceFacet;
+
+export function buildCoreCommerceFacet<
+  T1 extends AnyFacetValueRequest,
+  T2 extends AnyFacetValueResponse,
```
Spuffynism, line 153:
```suggestion
  ValueRequest extends AnyFacetValueRequest,
  ValueResponse extends AnyFacetValueResponse,
```

packages/headless/src/controllers/commerce/facets/core/numeric/headless-commerce-numeric-facet.ts
```diff
@@ -0,0 +1,56 @@
...
+  CoreCommerceFacetOptions,
+  'toggleSelectActionCreator' | 'toggleExcludeActionCreator'
+>;
+
+/**
+ * The `CommerceNumericFacet` controller offers a high-level programming interface for implementing numeric commerce
+ * facet UI component.
+ */
+export interface CommerceNumericFacet
+  extends CoreCommerceFacet<NumericRangeRequest, NumericFacetValue> {}
```
Spuffynism, line 24: The controller interface is now just a type alias. We could simplify its definition like this:
```suggestion
export type CommerceNumericFacet = CoreCommerceFacet<NumericRangeRequest, NumericFacetValue>;
```

packages/headless/src/controllers/commerce/facets/core/numeric/headless-commerce-numeric-facet.ts
```diff
@@ -0,0 +1,56 @@
...
+    options: {
+      ...options,
+      toggleSelectActionCreator: toggleSelectNumericFacetValue,
+      toggleExcludeActionCreator: toggleExcludeNumericFacetValue,
+    },
+  });
+
+  return {
+    ...coreController,
+  };
```
Spuffynism, line 55: No need for the intermediary `const`. We can could return the controller right away:
```suggestion
  return buildCoreCommerceFacet<
    NumericRangeRequest,
    NumericFacetValue
  >(engine, {
    options: {
      ...options,
      toggleSelectActionCreator: toggleSelectNumericFacetValue,
      toggleExcludeActionCreator: toggleExcludeNumericFacetValue,
    },
  });
```
Same comment applies for the `buildCommerceRegularFacet`!

packages/headless/src/controllers/commerce/product-listing/facets/headless-product-listing-numeric-facet.ts
```diff
@@ -0,0 +1,37 @@
...
+    fetchResultsActionCreator: fetchProductListing,
+  });
+
+  if (!loadProductListingReducer(engine)) {
+    throw loadReducerError;
+  }
+
+  return {
+    ...coreController,
+  };
```
Spuffynism, line 36: Is there a reason we load the reducers after creating the controller? From what I can tell, we usually load the reducers first.

Doing so also allows us to return the controller right away. 👀 
```suggestion
  if (!loadProductListingReducer(engine)) {
    throw loadReducerError;
  }
  return buildCommerceNumericFacet(engine, {
    ...options,
    fetchResultsActionCreator: fetchProductListing,
  });
```
Same comment applies to `buildProductListingRegularFacet`

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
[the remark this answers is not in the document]
Spuffynism: No strong opinions. Shouldn't the check be made in the `validatePayload` part of the action though? 🤔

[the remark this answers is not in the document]
fbeaudoincoveo: We can't: what we want to validate is in the state, not in the payload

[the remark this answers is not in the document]
Spuffynism: Hah! Ok I'm fine with leaving it as-is.

packages/headless/src/controllers/commerce/product-listing/facets/headless-product-listing-numeric-facet.ts
fbeaudoincoveo: No reason, no! It's an artefact of multiple rounds of refactoring 😁

Spuffynism (approved):

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -0,0 +1,285 @@
...
+      dispatch(deselectAllFacetValues(facetId));
+      dispatch(props.options.fetchResultsActionCreator());
+    },
+
+    showMoreValues() {
+      const numberInState = getRequest().numberOfValues;
+      const initialNumberOfValues = getRequest().initialNumberOfValues;
+      const numberToNextMultipleOfConfigured =
+        initialNumberOfValues - (numberInState % initialNumberOfValues);
+      const numberOfValues = numberInState + numberToNextMultipleOfConfigured;
```
louis-bompart, line 235: seems complicated

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
```diff
@@ -113,24 +197,62 @@ function updateStateFromFacetResponse(
...
+  switch (facetResponse.type) {
+    case 'numericalRange':
+      return convertToRangeRequests(
+        (facetResponse as NumericFacetResponse).values
+      );
+    case 'dateRange': // TODO
+    case 'hierarchical': // TODO
+    case 'regular':
+    default:
+      return (facetResponse as RegularFacetResponse).values.map(
```
louis-bompart, line 216: why the cast?

louis-bompart (approved):

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
fbeaudoincoveo: Right now it's because I'm defaulting to treating the facet response as a "regular" facet response if its type doesn't match any of the currently handled facetResponse.type values (i.e., numericalRange and regular).

However now that I think of it, this is not a good default behavior.  The method should instead not even try to do a conversion and return nothing when the type is not yet supported.

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
fbeaudoincoveo: Indeed, this comes from the original core facet so I don't have full context.

I believe the reason why we're not simply doing `numberOfValues = numberInState + initialNumberOfValues` is that the number of values could go down in the response if a value is selected and some values that were previously being displayed no longer have matching results. The % allows us to "reset" the slate to request a nice multiple of the initial number of values.

PR #3452, opened 2023-12-04
feat(quantic): added relevant generative answering (rga) example in the quantic examples org

SimonMilord:
[SFINT-5301](https://coveord.atlassian.net/browse/SFINT-5301)

**IN THIS PR:**
- Added a separate Relevant Generative Answering example in the LWC examples org. As of right now we only have full search and insight panel for the use cases.

https://github.com/coveo/ui-kit/assets/73316533/524e1a0b-837d-4982-a174-624d697dc335

[SFINT-5301]: https://coveord.atlassian.net/browse/SFINT-5301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/examples/main/lwc/exampleRelevantGenerativeAnswering/exampleRelevantGenerativeAnswering.js
```diff
@@ -0,0 +1,82 @@
...
+    const isThread = CoveoHeadless.ResultTemplatesHelpers.fieldMustMatch(
+      'source',
+      ['iNaturalistTaxons']
+    );
+    const isChild = CoveoHeadless.ResultTemplatesHelpers.fieldMustMatch(
+      'quantic__templateId',
+      ['myChildTemplate']
+    );
+
+    resultTemplatesManager.registerTemplates(
```
SimonMilord, line 55: I added the results templates here also instead of importing them from `exampleSearch` since if something ever changes with the example search, at least it will not break this example. (there is code duplication as a result) 

If this is not something we want, I can change the import or refactor the `exampleSearch` to add properties and render in the HTML of the RGA example the `exampleSearch` with extra properties to make it "RGA". 

I am not a fan of this option however, I believe each example should be independant.

packages/quantic/force-app/examples/main/lwc/exampleRelevantGenerativeAnswering/exampleRelevantGenerativeAnswering.js
```diff
@@ -0,0 +1,82 @@
...
+// @ts-ignore
+import youtubeTemplate from './resultTemplates/youtubeResultTemplate.html';
+
+export default class ExampleRelevantGenerativeAnswering extends LightningElement {
+  /** @type {string} */
+  @api engineId = 'example-relevant-generative-answering';
+  /** @type {string} */
+  @api searchHub = 'default';
+  /** @type {string} */
+  @api pipeline = 'genqatest';
```
SimonMilord, line 19: set the pipeline to `genqatest` so that it could generate answers. However, I was wondering if we should not add a note on the page to try to use : `How to resolve netflix connection with tivo` as a query? Otherwise someone could make queries that does not generate an answer, which defeats the purpose of this example...

packages/quantic/force-app/examples/main/lwc/exampleRelevantGenerativeAnswering/exampleRelevantGenerativeAnswering.js
mmitiche: In this example component, the goal is to show case how the generated answer component can be used, so I think, for simplicity sake and to keep the focus on the generated answer component, we could completely avoid adding this result templates logic. The default templates will be then be used and I think this should be fine.

packages/quantic/force-app/examples/main/lwc/exampleRelevantGenerativeAnswering/exampleRelevantGenerativeAnswering.html
```diff
@@ -0,0 +1,205 @@
...
+    >
+      <c-quantic-aria-live></c-quantic-aria-live>
+      <div class="slds-grid slds-grid_vertical slds-grid_align-center">
+        <div
+          class="slds-col slds-size_1-of-1 slds-large-size_6-of-12 slds-align-middle"
+        >
+          <div class="slds-m-vertical_small">
+            <c-quantic-search-box
+              engine-id={engineId}
+              textarea="true"
```
mmitiche, line 21:
```suggestion
              textarea
```

packages/quantic/force-app/examples/main/lwc/exampleRelevantGenerativeAnswering/exampleRelevantGenerativeAnswering.js
mmitiche: This component could serve as code sample that we can share with the clients, so it will not be necessarily deployed, people would just be curious to see how the `c-quantic-generated-answer` can be added to their HTML in order to appear in the search page.
But we could add a comment mentioning the query that could yield a generated answer with this pipeline.

packages/quantic/force-app/examples/main/lwc/exampleRelevantGenerativeAnswering/exampleRelevantGenerativeAnswering.js
SimonMilord: sounds good! will remove the logic for templates

mmitiche (approved):
Nice 👌

PR #3454, opened 2023-12-04
fix(quantic): Fix few issues related to overflow in mobile view 

SimonMilord:
[SFINT-5303](https://coveord.atlassian.net/browse/SFINT-5303)

**IN THIS PR:**
- On mobile view, there are 2 overflow issues right now for citations and printableURI

**ISSUES:**

image-20231122-152922

1) The rephrase button overflows in mobile
2) The bottom metadata overflows in mobile view

**SOLUTIONS:**

1) Setting the multilineFooter property to true in the generated answer component fixes that issue. 

DEMO:

Screenshot 2023-12-04 at 12 39 49 PM

2) The result templates had a `slds-grid` which was wrapping the bottom metadata and made the link overflow even though it had `slds-truncate`. I removed the class which did not affect the display even when not in mobile view. 

DEMO:

Screenshot 2023-12-04 at 12 35 25 PM

[SFINT-5303]: https://coveord.atlassian.net/browse/SFINT-5303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -387,7 +387,7 @@ export default class QuanticGeneratedAnswer extends LightningElement {
   }
 
   get generatedAnswerFooterCssClass() {
-    return `slds-grid slds-grid_align-spread generated-answer__footer--${
+    return `slds-grid slds-grid_align-spread slds-wrap generated-answer__footer--${
```
mmitiche, line 390: This unfortunately would break how we kinda want to display the citations. we want the citations to be wrapped in multiple lines when necessary, however if we give 'slds-wrap' to container holding the citations and the rephrase buttons it means that the citations will not  really need to be wrapped since the rephrase buttons will be them wrapped and moved to the next line.

My solution for this is just to use the public property 'multilineFoolter'.
If a user wants to use the generated answer component in a mobile interface, he just needs to set `multilineFooter` to `true` that's the purpose of this property, to solve this overflow issues.

mmitiche (commented):
The result templates fix looks good, the generate answer fix however needs to be discarded I think.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
SimonMilord: ok I will scratch that fix, but I tried and setting the multilineFooter property to true creates overflow on the rephrase buttons for some of the mobile resolutions (ie : iphone SE and other smaller mobile screen sizes)
See:
Screenshot 2023-12-04 at 4 30 13 PM

To fix this, I suggest adding `slds-wrap` class to the rephrase buttons container itself.

See:
Screenshot 2023-12-04 at 4 31 47 PM

mmitiche: Unfortunately this also  has an impact that we do not really want. For example:

In the following case we do not really want to display the rephrase buttons in multiple lines, we always want to display them in the same line: 

https://github.com/coveo/ui-kit/assets/86681870/ad7b09b6-0db7-4f2b-89d1-23e09f4248ac



While very small phones are not really our main concern the ideal solution to fix that case I think is to make the rephrase buttons display only the icons and not the labels, this would liberate more space and would allow to display the rephrase buttons in one line.

SimonMilord: will leave it as is then since the multilineFooter option is good enough for now

mmitiche (approved):
LGTM 👌

PR #3455, opened 2023-12-04
fix(quantic): Fix issue where tooltip is displayed under elements above the generated answer component

SimonMilord:
[SFINT-5321](https://coveord.atlassian.net/browse/SFINT-5321)

ISSUE:

image (14)

FIX: 
- Increased the z-index of the tooltip so that it is never hidden by any other elements

DEMO:

Screenshot 2023-12-04 at 1 43 51 PM

[SFINT-5321]: https://coveord.atlassian.net/browse/SFINT-5321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
+++ b/packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
@@ -6,7 +6,7 @@
 .tooltip__content {
   left: 50%;
   transition: transform 0s linear 0s, opacity 0.1s linear 0s, visibility 0.1s linear 0s;
-  z-index: 2;
+  z-index: 9001;
 }
 
 .tooltip__content--positioned-above {
@@ -32,7 +32,7 @@
   left: 50%;
   transform: translate(-50%, -100%);
   transition: opacity 0.1s linear 0s, visibility 0.1s linear 0s;
-  z-index: 2;
+  z-index: 9001;
 }
 
 .tooltip__arrow--dark {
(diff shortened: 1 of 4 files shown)


packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.css
```diff
@@ -6,7 +6,7 @@
 .tooltip__content {
   left: 50%;
   transition: transform 0s linear 0s, opacity 0.1s linear 0s, visibility 0.1s linear 0s;
-  z-index: 2;
+  z-index: 9001;
```
mmitiche, line 9: Why does it need to be 9001?
I wouldn't put such high value cause this could cause other issues.
What's the z-index value of the search box?

SimonMilord: I can put whatever value, but if I put lets say 31 and a user has another element with z-index 50 it still doesnt fix the issue

SimonMilord: makes sense, the z-index value of the expandable search box is 30 currently so I could set it to 40 instead and do increments of 10 (I think its unlikely a users needs more than 10 levels of z-index for his components). Or we can also set by increments of 30 and set it to 60. What do you think is best? Personally option 1 seems fine to me.

mmitiche: If we set the expandable search box to 40 than we do increments of 10 we are kinda breaking this equal increments rule, cause we went from 0 to 40 to 50,60...

I would find it best if we got 0, 20, 40, 60 or 0, 10, 20, 30 ...

SimonMilord: I tried with 10 for the searchbox and it seems to work, but I think I would be more comfortable doing 20 increments just in case so I will set the searchbox z-index to 20 and the tooltip to 40

mmitiche (approved):
👍

PR #3462, opened 2023-12-07
fix(headless): 404 error in query stream

jelmedini:
[SVCC-3275](https://coveord.atlassian.net/browse/SVCC-3275)

-Currently when we perform different searches continuously without letting the stream finish, we encounter 404 error because the stream is not found.

-The Problem: the retry function is called after timeout even if it was already cancelled, so the stream return 404 error as the stream does not exist anymore.
-The solution: Manage timeouts during the stream of the generated answer, by Keeping track of used timeouts to control the retry mechanism.

Before:
Screenshot 2023-12-01 at 9 42 32 AM

After in atomic:

https://github.com/coveo/ui-kit/assets/119955059/f647d5cf-4d04-4498-bdc6-83b50a3618e8

After in Quantic:

https://github.com/coveo/ui-kit/assets/119955059/de4f7cb5-7213-4db9-b819-94eb6b07fc4b

### DIFF
--- a/packages/headless/src/api/generated-answer/generated-answer-client.ts
+++ b/packages/headless/src/api/generated-answer/generated-answer-client.ts
@@ -46,6 +46,18 @@ interface StreamCallbacks {
   resetAnswer: () => void;
 }
 
+class TimeoutStateManager {
+  private usedTimeouts: Set<ReturnType<typeof setTimeout>> = new Set();
+
+  public recordUsedTimeout(timeout: ReturnType<typeof setTimeout>) {
+    this.usedTimeouts.add(timeout);
+  }
+
+  public isTimeoutUsed(timeout: ReturnType<typeof setTimeout>): boolean {
+    return this.usedTimeouts.has(timeout);
+  }
+}
+
 export class GeneratedAnswerAPIClient {
   private logger: Logger;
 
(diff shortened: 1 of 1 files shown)


packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -46,6 +46,18 @@ interface StreamCallbacks {
   resetAnswer: () => void;
 }
 
+class TimeoutStateManager {
```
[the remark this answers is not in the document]
jelmedini: good point !

packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -46,6 +46,22 @@ interface StreamCallbacks {
...
+  private timeouts: Set<ReturnType<typeof setTimeout>> = new Set();
+
+  public add(timeout: ReturnType<typeof setTimeout>) {
+    this.timeouts.add(timeout);
+  }
+
+  public remove(timeout: ReturnType<typeof setTimeout>) {
+    this.timeouts.delete(timeout);
+  }
```
nathanlb, line 58: L135
L146
L153
We run `clearTimeout` in these places but we don't remove them from this state manager. Would it be safer to execute `timeouStateManager.remove()` and have remove defined like this:
```suggestion
  public remove(timeout: ReturnType<typeof setTimeout>) {
    clearTimeout(timeout);
    this.timeouts.delete(timeout);
  }
```

jelmedini: yes, It make sense!

louis-bompart (approved):

PR #3463, opened 2023-12-07
refactor(headless): migrate searchbox submit event for search case

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-2933

The migration includes:

Marking the old log functions with `//TODO: KIT-2859`.
Adjusting every reference of every particular function to use the new 'next' analytics pattern.
Add tests for each deprecated functions to see if legacy analytics === next analytics in the search request.

Migrations:

- https://github.com/coveo/analytics-migration-project/issues/23
- https://github.com/coveo/analytics-migration-project/issues/49


packages/headless/src/controllers/insight/search-box/headless-insight-search-box.ts
```diff
@@ -30,6 +29,42 @@ export interface SearchBoxProps {
...
+   * Deselects all facets and triggers a search query.
+   */
+  submit(): void;
+
+  /**
+   * The state of the `SearchBox` controller.
+   */
+  state: SearchBoxState;
+}
+
```
alexprudhomme, line 67: The insight use case was entirely exporting the core search box interface that has a different submit function. This is technically a breaking change but can be considered a bug. The other stuff was never supposed to be exposed...

olamothe: Headless insight panel is also "used by us" (Coveo R&D), not clients/partners directly.

So this change is completely fine as long as everything still works in e2e tests.

olamothe (approved):

PR #3465, opened 2023-12-07
fix(atomic): remove unnecessary code in AtomicSearchBox

y-lakhdar:
There is a bug where if the user hovers over a query suggestion and then presses the "Enter" key, the hovered suggestion is applied to the search instead of the actual query in the text box. This behavior should be changed so that query suggestions are only executed when the user specifically clicks on them or uses the up/down arrows to navigate to them before pressing "Enter".

The new paradigm now considers a suggestion "active" only when the user uses the up/down arrows. As a consequence, a suggestion that has been moused over will no longer have its part selector set to `active-suggestion`. Feedback is welcome to determine if this change is considered breaking or a fix.

### DIFF
--- a/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
+++ b/packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
@@ -386,7 +386,7 @@ describe('Search Box Test Suites', () => {
           SearchBoxSelectors.inputBox().type('{enter}', {force: true});
         });
 
-        it('has recent query', () => {
+        it('has same query as before', () => {
           SearchBoxAssertions.assertHasText('not a suggestion');
         });
       });


packages/atomic/cypress/e2e/search-box/search-box-assertions.ts
```diff
@@ -78,6 +78,13 @@ export function assertSuggestionIsSelectedWithoutIt(index: number) {
     .should('contain', 'active-suggestion');
 }
 
+export function assertSuggestionIsNotSelectedWithoutIt(index: number) {
+  SearchBoxSelectors.querySuggestions()
+    .eq(index)
+    .should('have.attr', 'part')
+    .and('contain', 'active-suggestion');
```
fbeaudoincoveo, line 85: I'm not super used to Cypress, not having worked with it much myself, but the way this reads makes it seem like a suggestion that is not selected should contain an `active-suggestion` part, which feels pretty counter-intuitive to me 🤔 

Can you explain how this assertion works?

packages/atomic/cypress/e2e/search-box/search-box.cypress.ts
```diff
@@ -374,6 +374,23 @@ describe('Search Box Test Suites', () => {
...
+          SearchBoxSelectors.inputBox().focus();
+          SearchBoxSelectors.inputBox().type('not a suggestion', {
+            delay: 200,
+            force: true,
+          });
+          SearchBoxSelectors.querySuggestions().eq(1).trigger('mouseover');
+          SearchBoxSelectors.inputBox().type('{enter}', {force: true});
+        });
+
+        it('has recent query', () => {
```
fbeaudoincoveo, line 389:
```suggestion
        it('has same query as before', () => {
```

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -633,17 +633,7 @@ export class AtomicSearchBox {
     this.searchBoxAriaMessage = '';
   }
 
-  private onSuggestionMouseOver(
-    item: SearchBoxSuggestionElement,
-    side: 'left' | 'right',
-    id: string
-  ) {
-    const thisPanel = side === 'left' ? this.leftPanelRef : this.rightPanelRef;
```
fbeaudoincoveo, line 641: Does this imply that if instant results are being displayed in the right panel, then hovering a query suggestion in the left panel would not update the instant results? I.e., you'd have to use the arrows to set focus on those?

If so, I'm not convince this is the behavior we want in this specific case

PR #3466, opened 2023-12-07
fix(quantic): color styling issue fixed with the citations inside salesforce

mmitiche:
[SFINT-5323](https://coveord.atlassian.net/browse/SFINT-5323)

The SLDS design token we were using returns the color black in a community and the color white in the service console, so I stopped using this design token.

### Before
Screenshot 2023-12-07 at 11 28 24 AM

### After
Screenshot 2023-12-07 at 11 20 04 AM

[SFINT-5323]: https://coveord.atlassian.net/browse/SFINT-5323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mmitiche:
> Ugh....
> 
> Isn't there one that would be better suited for just regular text color?

I will double check

SimonMilord (approved):

PR #3467, opened 2023-12-07
feat(quantic): new public property fieldsToIncludeInCitations added to the QuanticGeneratedAnswer component

mmitiche:
[SFINT-5324](https://coveord.atlassian.net/browse/SFINT-5324)

- Exposed a new property that allows specifying the fields that need to be fetched and included in the citations used to generate the answer.

#### Usage:
```
 <c-quantic-generated-answer 
   engine-id={engineId} 
   fields-to-include-in-citations="date,author,source,language,filetype,parents">
</c-quantic-generated-answer>

```

#### Results:

https://github.com/coveo/ui-kit/assets/86681870/e0fdcda3-4ffa-4a1d-9760-79111e5c87c6

[SFINT-5324]: https://coveord.atlassian.net/browse/SFINT-5324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -100,6 +102,13 @@ describe('quantic-generated-answer', () => {
         Expect.searchQueryContainsCorrectRephraseOption(defaultRephraseOption);
       });
 
+      it('should perform a search query with the default fields to include iin citations', () => {
```
erocheleau, line 105:
```suggestion
      it('should perform a search query with the default fields to include in citations', () => {
```

erocheleau (approved):
Looks good to me, the only thing I was wondering is if we adopt this, then if you provide your own list of fieldsToIncludeInCitations, it will override the default value, meaning you'll break the Salesforce auto fields.

I was wondering if it was better to add our defaults no matter what to the list of fields to include provided by the user?

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -359,6 +368,27 @@ describe('quantic-generated-answer', () => {
...
+      const streamId = crypto.randomUUID();
+      const customFields = 'foo,bar';
+
+      beforeEach(() => {
+        mockSearchWithGeneratedAnswer(streamId);
+        mockStreamResponse(streamId, genQaMessageTypePayload);
+        visitGeneratedAnswer({fieldsToIncludeInCitations: customFields});
+      });
+
+      it('should send a search query with the right fields to include in citations option as a parameter', () => {
```
SimonMilord, line 381:
```suggestion
      it('should perform a search query with the right fields to include in citations option as a parameter', () => {
```

Just a suggestion for the test suite to be more consistent with the previous test (default) description

SimonMilord (approved):

mmitiche:
> Looks good to me, the only thing I was wondering is if we adopt this, then if you provide your own list of fieldsToIncludeInCitations, it will override the default value, meaning you'll break the Salesforce auto fields.
> 
> I was wondering if it was better to add our defaults no matter what to the list of fields to include provided by the user?

Indeed, a user could break the salesforce auto fields if they use custom fields and forget to include the salesforce fields.

That's the same logic used with the fields to include in the results in the QuanticResultList and QuanticFoldedResultList, so if we change this here it would be best to change it everywhere.
However so far we have never heard of a similar issue where a client used custom fields to include and missed the salesforce fields. 

So both solutions make sense, the actual solution favours total freedom and control on the fields to include and the solution you suggested favours safety.

We can adapt to the suggested solution but I think this should be added in a separate PR, first because we need to adapt the changes to the QuanticResultList and QuanticFoldedResultList components and we need to add specific tests that make sure this logic is being performed correctly.

PR #3475, opened 2023-12-09
feat(commerce): extract core pagination for plp and search

Spuffynism:
Extract a core pagination controller for commerce and reuse it for PLPs and for Search.

[CAPI-246]

[CAPI-246]: https://coveord.atlassian.net/browse/CAPI-246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/pagination/core/headless-core-commerce-pagination.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(engine.addReducers).toBeCalledWith({
+      commercePagination,
+    });
+  });
+
+  it('exposes #subscribe method', () => {
+    expect(pagination.subscribe).toBeTruthy();
+  });
+
+  it('#selectPage dispatches #selectPage & #fetchResultsActionCreator', () => {
```
fbeaudoincoveo, line 39: I know I did it this way originally, but now that I look back at it, I think in general it would be a cleaner if we grouped tests by method (unless there's only one thing to test for a method, in which case a single 'it' would suffice) and do one test per expected dispatched action. Also, we should test that the correct payload is being passed with the dispatched action when applicable.

E.g., here we could do:

```typescript
describe('#selectPage', () => {
  let page: number;

  beforeEach(() => {
    page = 0;
    pagination.selectPage(pageToSelect);
  });

  it('dispatches #selectPage with correct payload', () => {
    expect(engine.actions).toContainEqual(selectPage(page));
  });

  it('dispatches #fetchResultsActionCreator' () => {
    const action = engine.findAsyncAction(fetchResultsActionCreator.pending);
    expect(action).toBeTruthy();
  });
});
```

packages/headless/src/controllers/commerce/pagination/core/headless-core-commerce-pagination.ts
```diff
@@ -0,0 +1,108 @@
...
+} from '../../../controller/headless-controller';
+
+/**
+ * The `Pagination` controller is responsible for navigating between pages of results in a commerce interface.
+ */
+export interface Pagination extends Controller {
+  /**
+   * Navigates to a specific page.
+   *
+   * @param page The page to navigate to.
```
fbeaudoincoveo, line 22:
```suggestion
   * @param page - The page to navigate to.
```

packages/headless/src/controllers/commerce/search/pagination/headless-search-pagination.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!loadSearchPagination(engine)) {
+    throw loadReducerError;
+  }
+
+  return buildCorePagination(engine, {
+    fetchResultsActionCreator: executeSearch,
+  });
+}
+
+function loadSearchPagination(
```
fbeaudoincoveo, line 26: If you agree with my idea of using a helper to load the product listing reducer, we could use a similar pattern here.

fbeaudoincoveo (approved):

packages/headless/src/features/commerce/pagination/pagination-slice.ts
```diff
@@ -22,12 +23,13 @@ export const paginationReducer = createReducer(
...
-      .addCase(fetchProductListing.fulfilled, (state, action) => {
-        const responsePagination = action.payload.response.pagination;
-        state.perPage = responsePagination.perPage;
-        state.totalCount = responsePagination.totalCount;
-        state.totalPages = responsePagination.totalPages;
-        // TODO: replace with state = responsePagination once the API response pagination returns the current page.
-      });
+      .addCase(
+        fetchProductListing.fulfilled,
+        (_, action) => action.payload.response.pagination
```
pvanool, line 28: Please forgive me my noobness with redux, but how does this knows the correct context when both a search and listing pagination is on the same page?

packages/headless/src/controllers/commerce/search/pagination/headless-search-pagination.ts
```diff
@@ -0,0 +1,31 @@
...
+  engine: CommerceEngine
+): engine is CommerceEngine {
```
pvanool, line 28: This is a bit sneaky when you're not fluent in typescript. Not returning anything will give a falsy value 👀

packages/headless/src/features/commerce/pagination/pagination-slice.test.ts
```diff
@@ -70,4 +74,36 @@ describe('pagination slice', () => {
 
     expect(finalState.page).toBe(0);
   });
+
+  it('sets the pagination on #fetchProductListing.fulfilled', () => {
+    const pagination = {
```
pvanool, line 79: nit, is this convention to inline these definition? Could see a use case to define a constant on top of the file.

pvanool (approved):

packages/headless/src/controllers/commerce/search/pagination/headless-search-pagination.ts
BenKandelaars: Why return anything at all. This function will always return `true`

Spuffynism: This is a common pattern throughout the repository. I'd raise this in the #searchheadlessviplounge slack channel.

packages/headless/src/features/commerce/pagination/pagination-slice.ts
Spuffynism: It doesn't! :) This is why we'll eventually want to adopt one of the propositions in [this RFC](https://docs.google.com/document/d/1shl4QnnDnQVZat00QT4PqcLAsvKUbGo_zTqwX3iZJWg/edit#heading=h.jgfle3phkvb2), which will allow us to have multiple commerce solution types active on the same page.

packages/headless/src/features/commerce/pagination/pagination-slice.test.ts
Spuffynism: Done in https://github.com/coveo/ui-kit/pull/3475/commits/77429ee7d8c8f87aa6322cbbea77ff06cf3d1743

packages/headless/src/controllers/commerce/search/pagination/headless-search-pagination.ts
Spuffynism: Done in https://github.com/coveo/ui-kit/pull/3475/commits/d1a791410e8384bf85f31c99633036b56aac3f3d

PR #3484, opened 2023-12-14
feat(headless): removed the triggeredBy from the insight analytics actions for logCreateArticle

SimonMilord:
[SFINT-5327](https://coveord.atlassian.net/browse/SFINT-5327)

**IN THIS PR:**

- Removed the triggeredBy from the insight analytics actions as it is no longer needed/ deemed useful
- updated coveo.analytics.js to 2.28.24

[SFINT-5328]: https://coveord.atlassian.net/browse/SFINT-5328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SFINT-5327]: https://coveord.atlassian.net/browse/SFINT-5327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/features/analytics/insight-analytics-actions.ts
+++ b/packages/headless/src/features/analytics/insight-analytics-actions.ts
@@ -7,7 +7,6 @@ import {InsightAction, makeInsightAnalyticsAction} from './analytics-utils';
 
 export interface CreateArticleMetadata {
   articleType: string;
-  triggeredBy: string;
 }
 
 export const logInsightInterfaceLoad = (): InsightAction =>
@@ -33,7 +32,6 @@ export const logInsightCreateArticle = (
     (client, state) => {
       validatePayload(createArticleMetadata, {
         articleType: requiredNonEmptyString,
-        triggeredBy: requiredNonEmptyString,
       });
       return client.logCreateArticle(
         createArticleMetadata,
--- a/package-lock.json
+++ b/package-lock.json
@@ -54628,9 +54628,9 @@
       }
     },
     "packages/headless/node_modules/coveo.analytics": {
-      "version": "2.28.23",
-      "resolved": "https://registry.npmjs.org/coveo.analytics/-/coveo.analytics-2.28.23.tgz",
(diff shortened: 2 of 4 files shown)


alexprudhomme (approved):

mmitiche (approved):

mmitiche (commented):
There is a build error related to this change that needs to be figured out.

packages/headless/src/features/analytics/insight-analytics-actions.ts
```diff
@@ -33,7 +32,6 @@ export const logInsightCreateArticle = (
     (client, state) => {
       validatePayload(createArticleMetadata, {
         articleType: requiredNonEmptyString,
-        triggeredBy: requiredNonEmptyString,
```
fbeaudoincoveo, line 36: I see that `triggeredBy` is still included in the `analytics/expandToFullUI` action meta. Do we want to remove it from there as well?

More fundamentally, though, this change will essentially "break" customer reports using `triggeredBy` as a custom dimension. Should we instead postpone this change to UI Kit V3?

SimonMilord: That is a good point, I think we could leave the `triggeredBy` in the expandToFullUI until V3. Personnally, I dont see much value in this action meta.... @mmitiche @erocheleau what do you guys think? I believe the `triggeredBy` for logExpandToFullUI was added essentially for the same reasons

erocheleau: Yea leave it as-is in the expandtofullui, and no this won't break customers reporting on it since this is just us creating the first time this would be logged with hosted insight panel.

erocheleau (approved):

fbeaudoincoveo (approved):

PR #3486, opened 2023-12-14
fix(error): added margin at the buttom of the error component

drosenfeld-coveo:
[SVCC-3252](https://coveord.atlassian.net/browse/SVCC-3252)

added padding the bottom of the error component
before: 
image

after:
Screenshot 2023-12-14 at 3 50 00 PM

blue line added to show margin

[SVCC-3252]: https://coveord.atlassian.net/browse/SVCC-3252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olamothe (changes requested):
It would be better to add that margin at a "higher" (more specific) level in the food chain than that low level component (that is used in different context). Because otherwise you make it so that component has some magic margin that might not make sense in all context.

Alternative are: 

* In the .html page, in the layout section itself that holds the atomic-insight-query-error
* In the `atomic-insight-query-error.tsx` file itself.

nathanlb:
> It would be better to add that margin at a "higher" (more specific) level in the food chain than that low level component (that is used in different context). Because otherwise you make it so that component has some magic margin that might not make sense in all context.
> 
> Alternative are:
> 
> * In the .html page, in the layout section itself that holds the atomic-insight-query-error
> * In the `atomic-insight-query-error.tsx` file itself.

I'm the one that suggested the spacing could be added to the base component. Is there a scenario where we would not want proper spacing with whatever is around this error component?

nathanlb (approved):

louis-bompart (approved):

PR #3489, opened 2023-12-15
feat(headless): refactored generated answer controllers to support working in multiple use cases

mmitiche:
[SFINT-5199](https://coveord.atlassian.net/browse/SFINT-5199)

In this PR I refactored the Generated Answer Headless controller in order to have the following:

- A core Generated Answer Headless controller that contains the majority of the logic that will be used by both the search use case and insight use case.
- A Search Generated Answer Headless controller, the controller that will be used in the search use case.
- An Insight Generated Answer Headless controller, the controller that will be used in the insight use case.
- Added the generated answer pipeline rule parameter to the insight query sent to the Customer Service API.
- Added a step to create the Generated Answer API client in the Insight engine.

### The RGA component inside the Hosted insight panel

https://github.com/coveo/ui-kit/assets/86681870/5062454d-5f13-44a3-bafa-d66e4b50a16f

[SFINT-5199]: https://coveord.atlassian.net/browse/SFINT-5199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/insight-search/insight-search-request.ts
```diff
@@ -45,6 +45,15 @@ export const buildInsightBaseRequest = (
       parentField: state.folding.fields.child,
       filterFieldRange: state.folding.filterFieldRange,
     }),
+    ...(state.generatedAnswer && {
+      pipelineRuleParameters: {
+        mlGenerativeQuestionAnswering: {
+          responseFormat: state.generatedAnswer.responseFormat,
+          citationsFieldToInclude:
```
SimonMilord, line 52: Is this something we want to send in the request of the search useCase as well? I don't currently see `citationsFieldToInclude` included in the current search-request

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -0,0 +1,327 @@
...
+  logCopyToClipboard(): void;
+  /**
+   * Logs a custom event indicating a cited source link was hovered.
+   * @param citationId - The ID of the clicked citation.
+   * @param citationHoverTimeMs - The number of milliseconds spent hovering over the citation.
+   */
+  logCitationHover(citationId: string, citationHoverTimeMs: number): void;
+}
+
+export interface GeneratedAnswerProps {
```
SimonMilord, line 108:
```suggestion
export interface GeneratedAnswerCoreProps {
```
or 

```suggestion
export interface CoreGeneratedAnswerProps {
```
I would consider renaming the core props to avoid confusion

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  return {
+    ...controller,
+
+    get state() {
+      return controller.state;
+    },
+
+    retry() {
```
SimonMilord, line 47: we are leaving the `retry()` & the `rephrase()` methods in the individual controllers and not the core one because of the different analyticsClients right?

packages/headless/src/controllers/core/generated-answer/headless-core-interactive-citation.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export interface InteractiveCitationOptions
+  extends InteractiveResultCoreOptions {
+  /**
+   * The generated answer citation.
+   */
+  citation: GeneratedAnswerCitation;
+}
+
+export interface InteractiveCitationProps extends InteractiveResultCoreProps {
```
SimonMilord, line 19:
```suggestion
export interface InteractiveCitationCoreProps extends InteractiveResultCoreProps {
```

?

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
nathanlb: It's because of the 'executeSearch' function which is specific to the use case here.

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
olamothe: Since those are currently re-exported as is in `headless-generated-answer` (and in turn, gets re-exported globally in index.ts), renaming those types would be a breaking change.

olamothe (approved):

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
nathanlb: Should we give it the executeSearch action too and it could define all the functions?

packages/headless/src/controllers/core/generated-answer/headless-core-interactive-citation.ts
olamothe: Similar as above, since those types are ultimately re-exported from index globally, the external name must not change or that would be considered a breaking change.

It's possible to rename here, but it means we have to re-create the old type name (which can then be just a copy of core options) before re-exported it globally.

Since there's no benefit, I would suggest just leaving it named `InteractiveCitationProps` here

nathanlb (commented):
Looks great so far! Too bad we need to wait for the deployment freeze to be over so the API change can be deployed.

packages/headless/src/controllers/core/generated-answer/headless-core-interactive-citation.ts
SimonMilord: ok makes sense! no need for change then!

SimonMilord (approved):

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -0,0 +1,327 @@
...
+
+interface SubscribeStateManager {
+  abortController: AbortController | undefined;
+  lastRequestId: string;
+  lastStreamId: string;
+  getIsStreamInProgress: () => boolean;
+  setAbortControllerRef: (ref: AbortController) => void;
+  subscribeToSearchRequests: (
+    engine: CoreEngine<
+      GeneratedAnswerSection & SearchSection & DebugSection,
```
erocheleau, line 133: Why is this different than what we had before in the SearchEngine? Where is that DebugSection coming from in this new CoreController?

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -0,0 +1,327 @@
...
+    },
+
+    logCitationHover(citationId: string, citationHoverTimeMs: number) {
+      dispatch(
+        analyticsClient.logHoverCitation(citationId, citationHoverTimeMs)
+      );
+    },
+
+    rephrase(responseFormat: GeneratedResponseFormat) {
+      dispatch(updateResponseFormat(responseFormat));
```
erocheleau, line 297: Why aren't we dispatching an executeSearch here? 
Because the search is different in insight than with search?

Should we have like a "SearchClient" just like the analytics client that would provide the methods to call to execute search etc.?

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -196,110 +30,38 @@ export function buildGeneratedAnswer(
...
-      dispatch(logHoverCitation(citationId, citationHoverTimeMs));
+      dispatch(
+        executeSearch({
+          legacy: generatedAnswerAnalyticsClient.logRetryGeneratedAnswer(),
+        })
+      );
     },
 
     rephrase(responseFormat: GeneratedResponseFormat) {
       dispatch(updateResponseFormat(responseFormat));
```
erocheleau, line 56: Wouldn't you end up doing 
`controller.rephrase(responseFormat);`
instead of redefining here the same thing as in the CoreController?

Right now this action exists on the Core Controller, but if we want to add behavior to it, that would impact all controllers dependent on the Core, right now it would have no effect?

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
```diff
@@ -0,0 +1,66 @@
...
+      dispatch(
+        executeSearch(
+          generatedAnswerInsightAnalyticsClient.logRetryGeneratedAnswer()
+        )
+      );
+    },
+
+    rephrase(responseFormat: GeneratedResponseFormat) {
+      dispatch(updateResponseFormat(responseFormat));
```
erocheleau, line 56: Same as in the search generated-answer
`controller.rephrase(responseFormat);` no?

packages/headless/src/features/insight-search/insight-search-request.ts
erocheleau: They are there: https://github.com/coveo/ui-kit/blob/bd95c9fbad8a9f33b14cc6448607069146a20c4c/packages/headless/src/features/search/search-request.ts#L77

erocheleau (approved):
Very nice work!
All seems clean after double checking from the original controller to the new core controller, everything seems to have followed.

erocheleau (commented):
Very nice work!
All seems clean after double checking from the original controller to the new core controller, everything seems to have followed.

erocheleau (approved):

erocheleau (approved):

erocheleau:
Weird spam from the .dev github code review screen haha just check the first batch of comments.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
mmitiche: Sure thing!

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
mmitiche: It's because the `streamAnswer` action used by this controller woks with the `StateNeededByGeneratedAnswerStream` which is define like the following:
```
type StateNeededByGeneratedAnswerStream = ConfigurationSection &
  SearchSection &
  GeneratedAnswerSection &
  DebugSection;
```
We need the debug section from the state in the logic to execute this `streamAnswer` action.

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
mmitiche: Exactly we are not dispatching an executeSearch here because the search is different in insight than with search.

I'm not having a SearchClient  because currently for example for the rephrase method we do:
#### For search: 
```
        executeSearch({
          legacy:
            generatedAnswerAnalyticsClient.logRephraseGeneratedAnswer(
              responseFormat
            ),
        })
```
#### For insight:
````
        executeSearch(
          generatedAnswerInsightAnalyticsClient.logRetryGeneratedAnswer()
        )
````
So even the analytics action passed to the executeSearch is different, one is using the key `legacy` and one is not.
Same thing with the `Retry` method in this controlle.

That's why I thought that it would be better and more simple to avoid using a search client and define the methods separately in the two controllers.

Probably when we standardize how the analytics should be logged with the new protocol, it would be much easier to use a SearchClient exactly like you mentioned.

File: packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.test.ts [hunk 9]
```diff
@@ -0,0 +1,59 @@
...
+
+    expect(action).toBeTruthy();
+  });
+
+  describe('#rephrase', () => {
+    const responseFormat: GeneratedResponseFormat = {
+      answerStyle: 'concise',
+    };
+
+    it('dispatches the update action', () => {
```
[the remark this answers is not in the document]
mmitiche: Indeed that's the strategy, but this test is not included in the tests of the core controller that's  why I'm doing it separately for both search and insight.

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
mmitiche: I tried that and I found that it's a bit tricky to get that to work with passing the `executeSearch` action, mainly because of the following: https://github.com/coveo/ui-kit/pull/3489#discussion_r1434947111

packages/headless/src/insight.index.ts
```diff
@@ -307,6 +307,19 @@ export type {
...
+  GeneratedAnswerProps,
+} from './controllers/insight/generated-answer/headless-insight-generated-answer';
+export {buildGeneratedAnswer} from './controllers/insight/generated-answer/headless-insight-generated-answer';
+
+export type {
+  InteractiveCitation,
+  InteractiveCitationProps,
+  InteractiveCitationOptions,
+} from './controllers/insight/generated-answer/headless-insight-interactive-citation';
+export {buildInteractiveCitation} from './controllers/insight/generated-answer/headless-insight-interactive-citation';
```
nathanlb, line 322: Gonna need this little guy too
```suggestion
export {buildInteractiveCitation} from './controllers/insight/generated-answer/headless-insight-interactive-citation';
export type {GeneratedAnswerStyle} from './features/generated-answer/generated-response-format';
```

nathanlb (approved):
Nice refactor! And thanks for the detailed answers

PR #3494, opened 2023-12-18
docs(quantic): new information added to better document result templates components dealing with fields

mmitiche:
[SFINT-4910](https://coveord.atlassian.net/browse/SFINT-4910)

The following new information has been added to the Quantic result templates component dealing with fields:
```
Make sure the field specified in this component is also included in the field array for the relevant template. See the this example: [Quantic usage](https://docs.coveo.com/en/quantic/latest/usage/#javascript)
```

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.js
@@ -5,6 +5,7 @@ import {LightningElement, api} from 'lwc';
 
 /**
  * The `QuanticResultText` component displays a given result field value.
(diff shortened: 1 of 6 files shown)


packages/quantic/force-app/main/default/lwc/quanticResultText/quanticResultText.js
```diff
@@ -27,6 +28,7 @@ export default class QuanticResultText extends LightningElement {
   @api label;
   /**
    * The field whose values you want to display.
+   *
```
mmitiche, line 31:
```suggestion
```

erocheleau (approved):

SimonMilord (approved):

PR #3496, opened 2023-12-19
fix(commerce): fix commerce search facet selectors and facet order

Spuffynism:
This PR contains two fixes:
- Upon search query response, we set the facet order
- In the core facet controller, we use a solution-type specific selector. This fixes a bug where we'd always check the `productListing` state for facets, even when trying to select facet values. By doing so, the core facet controller is now fully solution-type agnostic.
 
[CAPI-393]

### DIFF
--- a/packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
+++ b/packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
@@ -3,12 +3,9 @@ import {
   AsyncThunkAction,
 } from '@reduxjs/toolkit';
 import {CommerceEngine} from '../../../../app/commerce-engine/commerce-engine';
-import {
-  commerceFacetResponseSelector,
-  isCommerceFacetLoadingResponseSelector,
-} from '../../../../features/commerce/facets/facet-set/facet-set-selector';
 import {commerceFacetSetReducer as commerceFacetSet} from '../../../../features/commerce/facets/facet-set/facet-set-slice';
 import {
+  AnyFacetResponse,
   AnyFacetValueResponse,
   DateFacetValue,
   FacetType,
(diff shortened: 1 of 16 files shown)


packages/headless/src/features/commerce/facets/facet-set/facet-set-selector.ts
```diff
@@ -1,58 +1,9 @@
...
-  if ('productListing' in state) {
-    return state.productListing.isLoading;
-  }
-
-  if ('commerceSearch' in state) {
-    return state.commerceSearch.isLoading;
-  }
-
-  return undefined;
-};
```
Spuffynism, line 58: These three selectors have all been split into the solution-type-specific selectors

packages/headless/src/controllers/commerce/search/facets/headless-search-facet-options.ts
```diff
@@ -0,0 +1,35 @@
...
+export const commonOptions: Pick<
+  CoreCommerceFacetOptions,
+  | 'fetchResultsActionCreator'
+  | 'facetResponseSelector'
+  | 'isFacetLoadingResponseSelector'
+> = {
+  fetchResultsActionCreator: executeSearch,
+  facetResponseSelector,
+  isFacetLoadingResponseSelector,
+};
```
Spuffynism, line 35: This logic is very similar to the one for `ehadless-product-listing-facet-options.ts`. Maybe in the future, instead of having two selectors, we could have a single `getResponseState` selector which results `{ isLoading: boolean, facets: AnyFacetResponse[] }`

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -74,13 +71,20 @@ export interface CoreCommerceFacetOptions {
   >;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   fetchResultsActionCreator: () => AsyncThunkAction<unknown, void, any>;
+  facetResponseSelector: (
```
fbeaudoincoveo, line 74: We should test that the use case-specific facet controllers are passing the correct facet selector

We should also test that the core commerce facet controller is using the facet selectors that got passed to it (probably in the #state tests).

Spuffynism: Done in https://github.com/coveo/ui-kit/pull/3496/commits/e30f4997ba1fc9b669ccea882e4129f81b663d61 and https://github.com/coveo/ui-kit/pull/3496/commits/06d1f67ec53b85539486a41899ea821dd1505c45

fbeaudoincoveo (approved):
💯

PR #3498, opened 2023-12-20
feat(headless): getGeneratedAnswerMetadata method added to the InsightAnalytics provider

mmitiche:
[SFINT-5248](https://coveord.atlassian.net/browse/SFINT-5248)

### DIFF
--- a/packages/headless/src/api/analytics/insight-analytics.ts
+++ b/packages/headless/src/api/analytics/insight-analytics.ts
@@ -72,6 +72,15 @@ export class InsightAnalyticsProvider
     return buildFacetStateMetadata(getStateNeededForFacetMetadata(this.state));
   }
 
+  public getGeneratedAnswerMetadata() {
+    const state = this.getState();
+    const formattedObject: Record<string, string | boolean> = {};
+    if (state.generatedAnswer?.isVisible !== undefined) {
+      formattedObject['showGeneratedAnswer'] = state.generatedAnswer.isVisible;
+    }
+    return formattedObject;
+  }
+
   private get queryText() {
     return this.state.query?.q || getQueryInitialState().q;
   }
--- a/packages/headless/src/api/analytics/insight-analytics.test.ts
+++ b/packages/headless/src/api/analytics/insight-analytics.test.ts
@@ -1,6 +1,7 @@
 import {CoveoAnalyticsClient} from 'coveo.analytics';
 import pino from 'pino';
 import {getConfigurationInitialState} from '../../features/configuration/configuration-state';
+import {getGeneratedAnswerInitialState} from '../../features/generated-answer/generated-answer-state';
 import {buildMockFacetRequest} from '../../test/mock-facet-request';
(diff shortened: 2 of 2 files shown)


packages/headless/src/api/analytics/insight-analytics.ts
```diff
@@ -72,6 +72,15 @@ export class InsightAnalyticsProvider
...
   }
 
+  public getGeneratedAnswerMetadata() {
+    const state = this.getState();
+    const formattedObject: Record<string, string | boolean> = {};
+    if (state.generatedAnswer?.isVisible !== undefined) {
+      formattedObject['showGeneratedAnswer'] = state.generatedAnswer.isVisible;
+    }
+    return formattedObject;
+  }
```
olamothe, line 82: I think this would also do ?

```suggestion
public getGeneratedAnswerMetadata() {
    const state = this.getState();
    return {
      ...(state.generatedAnswer?.isVisible !== undefined && {
        showGeneratedAnswer: state.generatedAnswer.isVisible,
      }),
    };
}
```

olamothe (approved):
As far as the code is concerned its okay, but I'm sure you are aware that does not really fit with new event protocol (as was discussed in the design doc)

I guess @felixpelletier  / @jstpierref are currently working on defining those in new analytics schema ?

felixpelletier:
> As previously discussed in the following [design doc](https://docs.google.com/document/d/1vv9hKHZJ4tjSFRNoxG88DKqeTxvBAG-jBYMp-mAg6oE/edit?usp=sharing) about the UA events for RGA, it was decided that we wanna include the information about whether the generated answer is shown or not in the Search UA events of search interfaces and insight interfaces.

Yes we are.
We've defined some of them already. See [api.event.qnaAnswerAction](https://github.com/coveo/analytics_schema/tree/master/schemas/client/api.event.qnaAnswerAction) and other api.event.qna* events.

For the new protocol, there are still blockers that we've been trying to remove for a while now surrounding how we want to send info through (or maybe besides) Search Events and the number one thing we're trying to avoid is to add data to the `customData` field. I'm hoping to have that solved and have an answer for you in early January (kind of late to promise a resolution this week). For the crunchy details if you're really interested: [GDoc: Logging additional event data for Service products calling the Search API](https://docs.google.com/document/d/1ZBeQEg5t5HHtiCUjgKZgq0hTK96pfzxy6jEJ7n6bQ_Y/edit?usp=sharing)

Plus, in the new protocol, search events should not be used, because we already have server-side search events that log the same things, only more reliably.

SimonMilord (approved):

erocheleau (approved):
From my understanding this is indeed purely for the current analytics, and the new schemas we'll rethink that entire thing :)

packages/headless/src/api/analytics/insight-analytics.ts
mmitiche: Exactly 👌

PR #3499, opened 2023-12-20
feat(rga): hide copy to clipboard button when Clipboard API is not defined

lbergeron:
[SVCC-3362](https://coveord.atlassian.net/browse/SVCC-3362)

The Copy to clipboard button relies on the Clipboard API to do its thing. However, there are situations where this API is not available. This PR attempts to improve the behavior when the API is not available (or fails). If the API is not available, the Copy to clipboard button is hidden. On the other hand, if the API is available but the permission is not granted, the button color/tooltip is updated to show the issue, and an error is logged in the console.

I checked the possibility of using `navigator.permissions.query()` to find out whether permission to write to the clipboard is granted. Unfortunately support is uneven across browsers which makes it unreliable.

[SVCC-3362]: https://coveord.atlassian.net/browse/SVCC-3362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


louis-bompart (approved):

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -275,18 +298,25 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
-                      this.generatedAnswer.logCopyToClipboard();
-                    }}
-                  />
+                  {this.showCopyToClipboard && (
+                    <CopyButton
+                      title={this.copyToClipboardTooltip}
+                      isCopied={this.copied}
+                      error={this.copyError}
+                      onClick={async () => {
+                        try {
```
nathanlb, line 307: Is it possible that this try/catch block is redundant given this.copyToClipboard already has one?
If we add `this.generatedAnswer.logCopyToClipboard();` in there instead I think we can simplify this callback. No?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
```diff
@@ -22,11 +22,13 @@
               hide-labels
               data-cy="generated-answer__feedback"
             ></c-quantic-feedback>
-            <c-quantic-generated-answer-copy-to-clipboard
-              data-cy="generated-answer__copy-to-clipboard"
-              answer={answer}
-              class="slds-var-m-horizontal_xx-small"
-            ></c-quantic-generated-answer-copy-to-clipboard>
+            <template lwc:if={showCopyToClipboard}>
+              <c-quantic-generated-answer-copy-to-clipboard
```
nathanlb, line 26: No console error or tooltip text change in quantic?

louis-bompart (approved):

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
lbergeron: You are absolutely right.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
lbergeron: The quantic button fallbacks on the deprecated `document.execCommand('copy')` when the clipboard API fails.  I opted to leave it as is to avoid side effects but, at some point, the fallback will stop working.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -354,6 +354,10 @@ export default class QuanticGeneratedAnswer extends LightningElement {
     );
   }
 
+  get showCopyToClipboard() {
+    return !!navigator?.clipboard?.writeText;
+  }
+
```
mmitiche, line 360: It's just a question of preference, but I would see this condition and the logic in the HTML better inside the `QuanticGeneratedAnswerCopyToClipboard` component. The main point behind this is keeping the QuanticGeneratedAnswer component as simple as possible.

mmitiche (approved):
Looks good to me 👌 thanks for addressing the issue in Quantic as well!

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
lbergeron: My rule of thumb is to have simple child components that always render themselves, and delegate the resposibility of showing/hiding those simple components to a container component. When the visibility logic becomes too much for the top container, then I usually add an intermediate container component to deal with that part of the logic.

I don't mind moving the condition into the child component if you prefer that.

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -96,6 +99,7 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
 
   private storage: SafeStorage = new SafeStorage();
   private data?: GeneratedAnswerData;
+  private showCopyToClipboard: boolean = false;
```
olamothe, line 102: We could do just a getter function
```suggestion
   private get hasClipboard() {
     return !!navigator?.clipboard?.writeText;
   }
```

packages/atomic/src/components/search/generated-answer/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -201,11 +207,29 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
   private async copyToClipboard(answer: string) {
-    await navigator.clipboard.writeText(answer);
-    this.copied = true;
+    try {
+      await navigator.clipboard.writeText(answer);
+      this.copied = true;
+      this.generatedAnswer.logCopyToClipboard();
+    } catch (error) {
+      this.copyError = true;
+      console.error(`Failed to copy to clipboard: ${error}`);
```
olamothe, line 227: We should not use `console` directly anywhere in Atomic/Headless as much as possible .

Instead try to use the logger (accessible with `this.bindings.engine.logger` in this case).  

That way the logging stays consistent everywhere/can be formatted by user if they want depending on their environment/can control logging level etc.

olamothe (approved):

nathanlb (approved):

PR #3500, opened 2023-12-20
fix(commerce): handle facet, sort and pagination reset on context change

Spuffynism:
On context change, clear the facets, the sorting and the pagination.

I expect that once we adopt the proposal of routing/grouping, to allow multiple commerce solution types content on the same page, we'll have to alter this logic.

Additionally, we make the facet controller tolerate having no backing state. This prevents React from erroring when it accesses a controller's state which hasn't been destroyed yet, but which is about to.

See accompanying PR in barca sports which demos this fix: https://github.com/coveo/barca-sports/pull/144

[CAPI-426]

[CAPI-426]: https://coveord.atlassian.net/browse/CAPI-426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -178,18 +179,24 @@ export function buildCoreCommerceFacet<
...
   const getResponse = () =>
-    props.options.facetResponseSelector(engine.state, facetId)!;
+    props.options.facetResponseSelector(engine.state, facetId);
   const getIsLoading = () =>
     props.options.isFacetLoadingResponseSelector(engine.state);
 
   const getNumberOfActiveValues = () => {
-    return getRequest().values.filter((v) => v.state !== 'idle').length;
+    return (getRequest()?.values ?? []).filter((v) => v.state !== 'idle')
+      .length;
```
louis-bompart, line 191:
```suggestion
    return getRequest()?.values?.filter((v) => v.state !== 'idle').length ?? 0;
```

louis-bompart (approved):
lgtm

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
```diff
@@ -271,17 +277,17 @@ export function buildCoreCommerceFacet<
...
       );
-      const canShowMoreValues = response ? response.moreValuesAvailable : false;
+      const canShowMoreValues = response?.moreValuesAvailable ?? false;
 
       return {
         facetId,
-        type: response.type,
-        field: response.field,
-        displayName: response.displayName,
+        type: response?.type ?? 'regular',
```
fbeaudoincoveo, line 288: It feels a bit awkward to return a facet state at all when there's no facet response.

Maybe get state() should just return an empty object in such a case... What do you think?

fbeaudoincoveo (approved):

packages/headless/src/controllers/commerce/facets/core/headless-core-commerce-facet.ts
Spuffynism: Returning an empty object would force us to change the state's type. Since this is kind of a hack-ish fix, I'd rather keep things as-is.

When we introduce the concept of routes, we should be able to revert some of the changes in this PR. Facets, sorting and pagination will be set on a PLP's (or search's) state, and not on a global state on the engine, which means we won't need to handle this interference we're fixing here.

PR #3509, opened 2024-01-03
test(core-pager): rework with new practices

louis-bompart:
This is an example PR for how I think we may want to modify our tests to try to 'fit' with some Redux good practices (and adapt some others to our specificity).

My first attempt was to try to use integration testing with our controllers, but was not able to. The main reasons are that it's cumbersome/inadequate to test all reducers using the actions triggered by a controller, and vice-versa.

So I went the opposite route, where I isolated everything

I reason that we can test everything that does consume an engine in isolation, and we mostly do actually.
The only piece that hasn't been black boxed like this is the controllers, and essentially this is what this PR do with the pager controller.

KIT-2954

### DIFF
--- a/packages/headless/src/controllers/core/pager/headless-core-pager.test.ts
+++ b/packages/headless/src/controllers/core/pager/headless-core-pager.test.ts
@@ -1,9 +1,20 @@
 import {configuration} from '../../../app/common-reducers';
 import {
-  SearchEngine,
-  buildSearchEngine,
-} from '../../../app/search-engine/search-engine';
+  nextPage,
+  previousPage,
+  registerPage,
+  updatePage,
+} from '../../../features/pagination/pagination-actions';
+import {
+  currentPageSelector,
+  currentPagesSelector,
+  maxPageSelector,
+} from '../../../features/pagination/pagination-selectors';
 import {paginationReducer as pagination} from '../../../features/pagination/pagination-slice';
+import {
+  MockedSearchEngine,
+  buildMockSearchEngine,
+} from '../../../test/mock-engine-v2';
 import {createMockState} from '../../../test/mock-state';
 import {
   Pager,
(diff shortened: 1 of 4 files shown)


packages/headless/src/tsconfig.build.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "isolatedModules": true
+    "isolatedModules": true,
+    "types": ["node", "jest"]
   },
```
louis-bompart, line 6:
```suggestion
    "isolatedModules": true
  },
```

packages/headless/src/controllers/core/pager/headless-core-pager.test.ts
```diff
@@ -18,19 +13,26 @@ import {
...
+    preloadedState.pagination.totalCountFiltered =
+      preloadedState.pagination.numberOfResults * maxPageNumber;
+    engine = buildSearchEngine({
+      configuration: {
+        accessToken: 'token',
+        organizationId: 'organizationId',
+      },
+      loggerOptions: {level: 'silent'},
+      preloadedState,
+    });
```
louis-bompart, line 34: This would probably be extracted into an internal util shared across tests.

packages/headless/src/controllers/core/pager/headless-core-pager.test.ts
```diff
@@ -49,7 +51,7 @@ describe('Pager', () => {
   });
 
   it('it adds the correct reducers to engine', () => {
-    expect(engine.addReducers).toHaveBeenCalledWith({
+    expect(addReducersSpy).toHaveBeenCalledWith({
```
fbeaudoincoveo, line 54: I like this

packages/headless/src/controllers/core/pager/headless-core-pager.test.ts
```diff
@@ -91,70 +91,70 @@ describe('Pager', () => {
     initialState.page = 2;
     initPager();
 
-    expect(engine.actions).toContainEqual(registerPage(2));
+    expect(engine.state.pagination?.firstResult).toEqual(10);
```
fbeaudoincoveo, line 94: Mmm... I'm not sure I agree with this approach.

In general, what I believe we should do is:

1. Test that controller methods affecting the Redux state are dispatching the correct actions with the correct payloads
3. Test that the reducers are handling dispatched actions as expected (i.e., that the state is modified correctly).

It seems to me like what you're doing here is testing the reducer behavior directly in the controller tests. I don't think this is right.

louis-bompart:
Back to draft after longer thought.

packages/headless/src/controllers/core/pager/headless-core-pager.test.ts
louis-bompart: Simplified/absorbed with mock-engine-v2

packages/headless/src/controllers/core/pager/headless-core-pager.test.ts
louis-bompart: As discussed, I adapted my draft to represent the conclusion of our rubberducking session

olamothe (approved):
Nice

PR #3510, opened 2024-01-04
feat(atomic): add suggestion delay option for atomic search box

olamothe:
Add a new option on the search box that helps optimizing input lag on the search box. 

The rendering of complex UI elements bound to keypress on the search box can sometimes cause input lag.

The variable that cause lag to happen or not are:

* End user machine spec
* How complex/rich the instant result template are. ie: which component (potentially custom) are used in the template.
* How often they change on each keypress. ie: type 10 letters, get 10 new query suggestion on each letter -> 10 rerender loop for instant result. VS the opposite extreme scenario: type 10 letters, same top query suggestion is always returned -> 1 rerender loop for instant result.

By delaying the render loop after the end user stops typing, we can offer implementers some control over the last variable.

https://coveord.atlassian.net/browse/KIT-2947


louis-bompart (approved):
LGTM, seems like a balanced solution.

louis-bompart (approved):

PR #3511, opened 2024-01-04
fix(quantic): fix genQA copy-to-clipboard e2e

erocheleau:
SFINT-5341

Updated the e2e copy to clipboard cypress test to not check for copy to clipboard on browsers other than electron.

The problem was that the browser is requesting clipboard access through a popup which is not manageable with cypress.

So this specific test will only execute on electron.


SimonMilord (approved):
👌

PR #3517, opened 2024-01-09
fix(atomic): set end of range to end of day when applying date input

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-2953


louis-bompart (approved):
LGTM

PR #3519, opened 2024-01-12
test(case-assist): rework case-assist test with new practices

louis-bompart:
same as #3509, for case-assist.

This PR introduces the first rework instances of controllers depending on core controllers. I elected to isolate the core controller from the tests and only test the ingress/egress points between the core controller and the parent controller to avoid duplicating tests of the core controller into the parent controller.


packages/headless/src/app/analytics-middleware.test.ts
louis-bompart, line 1: Deleted, as this concept will not survive v3, I ain't enclined to spend time migrating it.

packages/headless/src/controllers/quickview/case-assist-headless-quickview.test.ts
```diff
@@ -1,170 +1,133 @@
...
-    initQuickview();
-
-    expect(quickview.state.content).toEqual('');
-  });
 
-  [true, false].forEach((testValue) => {
-    it(`when the result #hasHtmlVersion is ${testValue} #state.resultHasPreview should be ${testValue}`, () => {
-      options.result = buildMockResult({hasHtmlVersion: testValue});
-      initQuickview();
+    it('exposes `currentDocument` in the state ', () => {
```
fbeaudoincoveo, line 95: In the case of those "derived" controller state properties, I think we should fully test the intended behavior, because the implementation is in the controller code.

In this specific case, we expect state.currentDocument to return the 1-based position of the current document suggestion.

This is already what we're testing, but I think the test description should be more specific, e.g.:

'#state.currentDocument returns the 1-based position of the document suggestion being quick viewed'

fbeaudoincoveo: The fact that we're testing potentially complex behaviors in controller state getter properties makes me think that this kind of stuff "smells" a little bit. Maybe it would be preferable, in general, to test those selectors separately, and then test that a derived state property is calling the correct selector...

On the other hand, this might be introducing a new layer of abstraction just for the sake of it, so it's probably not worth the effort. It's just ever so slightly annoying me 😅

packages/headless/src/controllers/quickview/case-assist-headless-quickview.test.ts
```diff
@@ -1,170 +1,133 @@
...
     });
-  });
 
-  [true, false].forEach((testValue) => {
-    it(`when the resultPreview state #isLoading is ${testValue} #state.isLoading should be ${testValue}`, () => {
-      engine.state.resultPreview = buildMockResultPreviewState({
-        isLoading: testValue,
-      });
-      initQuickview();
+    it('exposes `totalDocuments` in the state ', () => {
```
fbeaudoincoveo, line 107:
```suggestion
    it('#state.totalDocuments returns the total number of document suggestions', () => {
```

packages/headless/src/controllers/quickview/case-assist-headless-quickview.test.ts
```diff
@@ -1,170 +1,133 @@
...
-    expect(options.maximumPreviewSize).toBe(0);
-  });
-
-  it('#preparePreviewPagination on initialization', () => {
-    initQuickview();
-    expect(engine.actions).toContainEqual(
-      preparePreviewPagination({
-        results: engine.state.documentSuggestion.documents,
-      })
+  it('when the core quickview calls fetchResultContentCallback, it dispatch a logQuickviewDocumentSuggestionClick with the proper uniqueId', () => {
```
fbeaudoincoveo, line 119:
```suggestion
  it('#fetchResultContentCallback dispatches a #logQuickviewDocumentSuggestionClick with the proper uniqueId', () => {
```

fbeaudoincoveo (approved):
Nice

packages/headless/src/controllers/case-field/headless-case-field.test.ts
```diff
@@ -51,27 +64,31 @@ describe('Case Field', () => {
   });
 
   it('building a case field registers the case field in the state', () => {
```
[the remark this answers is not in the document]
louis-bompart: > . In the controller, we should only test that the action is getting dispatched (which is what we do), so I believe we should update the test description accordingly.

👍 

> I'm also wondering about conditionally dispatching the registerCaseField action based on what's currently in the state. It seems to me like a better approach would be to always dispatch the action from the controller, and then let the reducer determine whether to modify the state or not.

I decided to not change anything in the logic, but I flagged most files of the case-assist as they infringe on https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially.

TL;DR: `case-assist` might need some love regarding how they do actions, but at the same time: if it works, why touch it?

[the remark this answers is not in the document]
louis-bompart:
```suggestion
  it('dispatch #registerCaseField on init if it was not already registered', () => {
```

packages/headless/src/controllers/case-field/headless-case-field.test.ts
```diff
@@ -99,108 +116,124 @@ describe('Case Field', () => {
...
-        expect.objectContaining({
-          type: fetchDocumentSuggestions.pending.type,
-        })
+      expect(mockedFetchDocumentSuggestions).toHaveBeenCalled();
+      expect(engine.dispatch).toHaveBeenCalledWith(
+        mockedFetchDocumentSuggestions.mock.results[0].value
       );
     });
 
     it('dispatches both #fetchCaseClassifications and #fetchDocumentSuggestions when required', () => {
```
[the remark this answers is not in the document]
louis-bompart: Probably, but I decided to not change the coverage/tests too much, so those PRs can be used as 'examples' of how to migrate.

I think also this is a good way to shine some glaring light on the https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially infringements

packages/headless/src/controllers/case-input/headless-case-input.test.ts
```diff
@@ -47,26 +58,32 @@ describe('Case Input', () => {
   });
 
   it('building a case input registers the input field in the state', () => {
```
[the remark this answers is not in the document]
louis-bompart:
```suggestion
  it('dispatch #updateCaseInput on init if it was not already registered', () => {
```

packages/headless/src/controllers/case-input/headless-case-input.test.ts
```diff
@@ -47,26 +58,32 @@ describe('Case Input', () => {
...
+    expect(mockedUpdateCaseInput).toHaveBeenCalledWith({
+      fieldName: testFieldName,
+      fieldValue: '',
+    });
+    expect(engine.dispatch).toHaveBeenCalledWith(
+      mockedUpdateCaseInput.mock.results[0].value
     );
   });
 
   it('building a case input that was already registered does not register the input field again', () => {
```
louis-bompart, line 72:
```suggestion
  it('do not dispatch #updateCaseInput on init if the field was already registered', () => {
```

packages/headless/src/controllers/case-field/headless-case-field.test.ts
```diff
@@ -51,27 +64,31 @@ describe('Case Field', () => {
...
+    expect(mockedRegisterCaseField).toHaveBeenCalledWith({
+      fieldName: testFieldName,
+      fieldValue: '',
+    });
+    expect(engine.dispatch).toHaveBeenCalledWith(
+      mockedRegisterCaseField.mock.results[0].value
     );
   });
 
   it('building a case field that was already registered does not register the case field again', () => {
```
louis-bompart, line 80:
```suggestion
  it('does not dispatch #registerCaseField on init if the field is already registered', () => {
```

packages/headless/src/controllers/document-suggestion-list/case-assist-headless-interactive-result.test.ts
```diff
@@ -1,64 +1,74 @@
...
-  it('when calling select(), logs documentOpen', () => {
-    interactiveResult.select();
-    expectLogDocumentActionPending();
+  it('calls logDocumentSuggestionOpen when the action is triggered for the first time', () => {
+    const mockedLogDocumentSuggestionOpen = jest.mocked(
+      logDocumentSuggestionOpen
+    );
+
+    mockedSelect();
+    expect(mockedLogDocumentSuggestionOpen).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
louis-bompart: I agree that we should:
 - [ ] test that the action creator was called with the proper argument (the result's unique ID)
 
I disagree with "we're not interested in the number of times the action has been dispatched".

https://github.com/coveo/ui-kit/blob/ce858c6499a41c47b1ead38e24e6ae744f111542/packages/headless/src/controllers/instant-results/headless-interactive-instant-result.ts#L40-L48

We can see that `wasOpened` plays the role of 'fuse', and that the `logAnalyticsIfNeverOpened` should only 'do something' once.

packages/headless/src/controllers/quickview/case-assist-headless-quickview.test.ts
```diff
@@ -1,170 +1,133 @@
...
+      {options},
+      buildResultPreviewRequest,
+      '/html',
+      expect.any(Function)
+    );
   });
 
-  it('it adds the correct reducers to engine', () => {
+  it('it adds the case-assist specific reducers to engine', () => {
     expect(engine.addReducers).toHaveBeenNthCalledWith(1, {
```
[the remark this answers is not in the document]
louis-bompart: It's not 'it's been called once' that's tested here, but that the 1st call has occurred with those params

https://jestjs.io/docs/expect#tohavebeennthcalledwithnthcall-arg1-arg2-

packages/headless/src/controllers/quickview/case-assist-headless-quickview.test.ts
louis-bompart: > The fact that we're testing potentially complex behaviors in controller state getter properties makes me think that this kind of stuff "smells" a little bit. Maybe it would be preferable, in general, to test those selectors separately, and then test that a derived state property is calling the correct selector...
>
>On the other hand, this might be introducing a new layer of abstraction just for the sake of it, so it's probably not worth the effort. It's just ever so slightly annoying me 😅

You are correct on both accounts I think:
 - Yes, complex 'state driller' should probably be isolated in selectors, and tested on their own.
 - Yes, doing so is probably ain't worth the hassle, especially on 'old static' code.
 
 However, we should I think keep that in mind for the future ;)
 
 In the meantime, for there, let's test it in the controller.

louis-bompart, line 95:
```suggestion
    it('#state.currentDocument returns the 1-based position of the document suggestion being quick viewed', () => {
```

packages/headless/src/controllers/document-suggestion-list/case-assist-headless-interactive-result.test.ts
```diff
@@ -1,64 +1,77 @@
...
+    expect(mockedLogDocumentSuggestionOpen).toHaveBeenCalledWith(
+      resultStringParams.uniqueId
+    );
   });
 
-  it('when calling select(), logs documentOpen', () => {
-    interactiveResult.select();
-    expectLogDocumentActionPending();
+  it('does not call logDocumentSuggestionOpen when the action is triggered for the second time', () => {
```
louis-bompart, line 68:
```suggestion
  it('does not dispatch logDocumentSuggestionOpen when the action is triggered for the second time', () => {
```

PR #3521, opened 2024-01-13
test(product-recommendation): test with new practices 

louis-bompart:


packages/headless/src/features/product-recommendations/product-recommendations.test.ts
louis-bompart, line 1: imo overkill/useless: retest reducers mostly

fbeaudoincoveo (approved):
I only did a cursory review since all of this stuff will eventually be replaced with the new commerce sub-package, but the changes seem fine to me.

PR #3526, opened 2024-01-15
chore(headless/ssr): bundle SSR utils into dedicated package

y-lakhdar:
Restructured the serialization logic by relocating it to `@coveo/headless/ssr` and eliminate any Next.js references within the ssr package. The updated logic has been incorporated into the sample applications.

Additionally, the serialization interface has been simplified. Instead of using the previous approach:

```ts
NextJsSearchParameterSerializer.fromSearchParameters(
    state.parameters
).applyToUrlSearchParams(newURL.searchParams);
```

Users can now utilize the following function for serialization:

```ts
const {serialize} = buildSSRSearchParameterSerializer();
serialize(newURL.searchParams, newUrl);
```

[KIT-2952](https://github.com/coveo/ui-kit/tree/KIT-2952)

[KIT-2952]: https://coveord.atlassian.net/browse/KIT-2952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/features/search-parameters/search-parameter-utils.test.ts
+++ b/packages/headless/src/features/search-parameters/search-parameter-utils.test.ts
@@ -190,27 +190,36 @@ describe('search-parameter-utils', () => {
     it('should add numeric ranges to searchParams if key is "nf"', () => {
       const searchParams: Record<string, unknown> = {};
       const key = 'nf-color';
-      const value = ['1', '100'];
+      const value = '1..100';
 
       extendSearchParameters(searchParams, key, value);
 
       expect(searchParams).toEqual({
         nf: {
-          color: [{start: 1, end: 100}],
+          color: [{start: 1, end: 100, endInclusive: false, state: 'selected'}],
         },
       });
     });
 
     it('should add date ranges to searchParams if key is "df"', () => {
+      const date1 = '2010/01/01@05:00:00';
+      const date2 = '2011/01/01@05:00:00';
       const searchParams: Record<string, unknown> = {};
       const key = 'df-date';
-      const value = ['2022-01-01', '2022-12-31'];
+      const value = `${date1}..${date2}`;
 
       extendSearchParameters(searchParams, key, value);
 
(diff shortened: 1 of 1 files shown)


packages/headless/src/features/search-parameters/search-parameter-serializer.ssr.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    const initialUrl = new URL('https://example.com/custom/route');
+
+    const result = serialize(searchParameters, initialUrl);
+
+    expect(result).toBe(
+      'https://example.com/custom/route?q=query&f-color=red&f-color=green&f-shape=square'
+    );
+  });
+
+  it('should not overwrite existing (non-coveo) search parameters in URL', () => {
```
fbeaudoincoveo, line 102: This makes me think: we should ensure that the list of "reserved" Coveo query parameters is documented somewhere so that implementers can take action on their end if needed to avoid conflicts.

packages/headless/src/features/search-parameters/search-parameter-serializer.test.ts
```diff
@@ -17,497 +22,565 @@ const someSpecialCharactersThatNeedsEncoding = [
...
-        const result = serialize({nf});
+      it('when the #f parameter contains invalid values, it does not include it', () => {
+        const f = {author: [1]} as never;
+        const result = serialize({f});
         expect(result).toEqual('');
       });
 
-      it('if a range does not have an #end key, it does not include it', () => {
-        const nf = {size: [{start: 10}]} as never;
+      it('serializes the #nf parameter correctly', () => {
```
fbeaudoincoveo, line 82: Should we also test that end inclusive ranges are serialized with the proper separator?

fbeaudoincoveo (approved):
Nice!

louis-bompart (approved):

PR #3529, opened 2024-01-17
feat(commerce): allow setting referrer

Spuffynism:
Allow setting the referrer through the context controller.

[CAPI-368]

[CAPI-368]: https://coveord.atlassian.net/browse/CAPI-368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/context/headless-context.ts
```diff
@@ -134,6 +135,14 @@ export function buildContext(
     ...props.options,
   };
 
+  if (
+    typeof window !== 'undefined' &&
+    typeof window.document !== 'undefined' &&
```
Spuffynism, line 140: Would this be the way to check for the controller running in a browser? I couldn't find a standardized way in headless.

If we _don't_ want to check for the controller being ran in a browser, I'd simply remove this guessing logic, but it'd be nice to automatically populate the referrer.

cc @fbeaudoincoveo since you might have opinions on this as well

fbeaudoincoveo: Mmm

I'm not sure I'd even try to automatically set the referrer in Headless. It seems to me like this is something we should do in Atomic.

Spuffynism: Removed the logic that automatically sets the referrer.

fbeaudoincoveo (approved):

PR #3530, opened 2024-01-17
feat(headless,atomic): support new query correction system

olamothe:
Add support for new query correction system, which leverage ML powered query suggestions.

The way this is fronted in Headless (and atomic) is through the addition of a `queryCorrectionMode` option/parameter.

The new system also properly support `automaticallyCorrectQuery` option, which is a nice improvement over the initial draft for this feature.

The feature is potentially breaking due to changes in the API response: a different API property is used to provide correction. 

In v2, the default value for `queryCorrectionMode` is `legacy`, with the option to opt in with the new system.
 
In v3, we can modify the default for `queryCorrectionMode` to `next` (or simply remove the possibility to choose, always defaulting to `next`).


louis-bompart:
> In v3, we can modify the default for queryCorrectionMode to next (or simply remove the possibility to choose, always defaulting to next).

Is there value to keep the legacy option?

packages/headless/src/controllers/core/did-you-mean/headless-core-did-you-mean.test.ts
```diff
@@ -15,6 +15,17 @@ import {
...
+jest.mock('pino', () => ({
+  ...jest.requireActual('pino'),
+  __esModule: true,
+  default: () => ({
+    info: jest.fn(),
+    error: jest.fn(),
+    warn: jest.fn(),
+    debug: jest.fn(),
+  }),
+}));
```
louis-bompart, line 27: future note, maybe we should put that in our root mock or something.
Pino is a PITA to isolate sometimes.

louis-bompart (approved):
lgtm, nice one!

PR #3531, opened 2024-01-18
feat(headless): Support the context search param with the insight engine

erocheleau:
SFINT-5359

- Added the context actions to the insight bundle.
- Added the context parameter mapping when building the insight query.
- Added tests.

It's on purpose that I didn't add the context to the querySuggest insight api client piece. This custom context is not used for personalization really, and if it becomes necessary we can add it later.


louis-bompart (approved):

mmitiche (approved):
LGTM!

SimonMilord (approved):

PR #3537, opened 2024-01-22
fix(quantic): Pressing ENTER key while targeting the clear icon submits search request in standalone searchbox

SimonMilord:
[SFINT-5340](https://coveord.atlassian.net/browse/SFINT-5340)

**Issue:**

When I focus on the clear icon using my keyboard (tab) and press enter, it clears the input and then triggers a search with an empty query payload instead of clearing the input and displaying suggestions while having focus on the input, which is the expected behaviour of a clear button and what happens when clicking on it.

Following discussions, we concluded that clearing by pressing ENTER should have the same behaviour as clicking on the icon. (Clearing, focusing on input and showing suggestions)
Example:

https://github.com/coveo/ui-kit/assets/73316533/c5973e49-d99b-43bb-8803-32995c900c3a

**Expected behaviour:**

We want to clear the input and leave focus on the input to display suggestions when focusing on the icon with a keyboard and then pressing ENTER, not trigger a search after clearing.

**Solution:**

Re-worked the solution to instead add only 1 `onkeydown` listener on the clearButton which calls the `handleKeyDownOnClearButton` method. This method checks that the key pressed is ENTER, and if that is the case, sets the `targetClearButton` boolean variable to true. After that, the keyup is triggered, calls `handleEnter` and checks if ENTER is pressed that the `targetClearButton` is false before calling `sendSubmitSearchEvent()` which was triggering the search request.

**DEMO:**

https://github.com/coveo/ui-kit/assets/73316533/5ecd1694-1599-486a-860b-506b0bbff64d

S/O to @erocheleau for the idea.

[SFINT-5340]: https://coveord.atlassian.net/browse/SFINT-5340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/defaultSearchBoxInput.html
```diff
@@ -37,6 +37,8 @@
               <template if:false={isQueryEmpty}>
                 <lightning-button-icon
                   onclick={clearInput}
+                  onkeyup={onKeyup}
```
SimonMilord, line 40: this supports the case where a key is pressed and triggers the `handleEnter()` function which will see that the clear icon is indeed focused and therefore call `clearInput()`

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/defaultSearchBoxInput.html
```diff
@@ -37,6 +37,8 @@
...
+                  onfocus={onClearButtonFocus}
```
SimonMilord, line 41: this makes sure it adds the `is-focused` attribute on the clear icon for the check in the `handleEnter()` function

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/defaultSearchBoxInput.html
mmitiche: How was the bug happening before the existence of this listener? how and why  was clicking enter on this icon triggering a search?

SimonMilord: 1- For your question about whether it was a standalone searchbox only issue: no it was a behaviour that was happening in both searchboxes (where tab and then pressing ENTER on clear icon would clear + trigger a search with empty query) whereas we want : to simply clear the input without triggering a search.

2- This listener was indeed useless. I thought I needed it because it was triggering the `onkeyup` listener of the input. What I found was that when we were calling `clearInput()` we were setting focus on the input and therefore I believe the browser was clearing and then triggering a `keyup` event on the input which had focus at that time. 

I removed the listener and remove the focus on input while clearing the input. 

3- BONUS: I noticed that if I tab to focus on clearIcon and then shift+tab to focus back on the input and then pressed ENTER, it would no longer trigger a search. This is why I added the `onblur()` listener to remove focus from the `clearIcon`

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -240,6 +242,14 @@ export default class QuanticSearchBoxInput extends LightningElement {
     this.adjustTextAreaHeight();
   }
 
+  onClearButtonFocus() {
+    this.inputClearButton.setAttribute('is-focused', 'true');
+  }
+
+  onClearButtonBlur() {
```
SimonMilord, line 249: this method prevents the issue where the focus is still on the `inputClearButton` and where:
```
tab on clear Icon
shift + tab back on input
press ENTER 
```

was no longer triggering a search

erocheleau: I don't see how manually controlling the focus is the solution to the issue...

You have the problem of the ENTER no longer triggering a search because you're manually checking this:

```
} else if (this.inputClearButton?.getAttribute('is-focused') === 'true') {
      this.clearInput();
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/defaultSearchBoxInput.html
```diff
@@ -37,6 +37,8 @@
               <template if:false={isQueryEmpty}>
                 <lightning-button-icon
                   onclick={clearInput}
```
erocheleau, line 39: We shouldn't have to manually handle a ENTER key event, pressing ENTER or for the matter SPACE is the same as the `onclick` on this button.

You can see that the issue isn't reproduced on an older version of quantic for example here: https://connect.coveo.com/s/global-search/%40uri#q=index%20ms%20teams&tab=All

Pressing ENTER or SPACE after tabbing to the focus on the clear button clears the inputs, closes the suggestions and does NOT trigger a new query.

So I think the issue is with the refactor of the searchbox that we did recently. Something isn't right after that?

erocheleau (changes requested):
Manually handling the focus on the X element shouldn't be necessary.
I don't think this is the right solution, I would look more closely towards the interaction between the searchbox, the searchboxinput and the events it sends.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/defaultSearchBoxInput.html
SimonMilord: The reason why we are manually handling the ENTER key event is that on the `clearInput` method, we are setting the focus back into the input in order to display the suggestions after clearing.

image

However, this results in the `onClick` of the clear button to be triggered, calling `clearInput()` which will change focus on the `this.input`, will will trigger the keyup event on the input and thus calling `handleEnter()`, which will finally trigger the search after clearing. 

I believe that by manually handling the ENTER key event we can both 1) not trigger a search after clearing and 2) still getting to display the suggestions and keeping the focus on the input after clearing.

I tried a much simpler solution where I keep only the `onClick` listener on the clear button which clears the input, lose focus, does not trigger a search but does not display suggestions either...

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -93,6 +93,9 @@ export default class QuanticSearchBoxInput extends LightningElement {
     this.suggestionListElement?.resetSelection();
   }
 
+  /** @type {boolean} */
+  targetClearButton = false;
```
mmitiche, line 97: For better readability: 
```suggestion
  isClearButtonTargeted = false;
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -222,7 +231,9 @@ export default class QuanticSearchBoxInput extends LightningElement {
   onKeyup(event) {
     switch (event.key) {
       case keys.ENTER:
-        this.handleEnter();
+        if (!this.targetClearButton) {
+          this.handleEnter();
+        }
```
mmitiche, line 236: Instead of checking the condition here, I would add the logic to check the condition inside the `handleEnter` method, in order to keep the `onKeyup` method clear and very easy to read.

mmitiche (commented):
Good fix!
Minor improvements suggested.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
SimonMilord: Moved it!

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -185,8 +188,11 @@ export default class QuanticSearchBoxInput extends LightningElement {
     if (this.areSuggestionsOpen && selectedSuggestion) {
       this.sendSelectSuggestionEvent(selectedSuggestion.rawValue);
     } else {
-      this.sendSubmitSearchEvent();
+      if (!this.isClearButtonTargeted) {
+        this.sendSubmitSearchEvent();
+      }
```
mmitiche, line 193: According to my previous comments I didn't mean  to say to change anything with the logic, cause it was fine and it was giving the right behaviour. 

This last commit is changing the logic.

With the current logic, the suggestions list  will not be opened after clicking on the enter ket on the clear input icon because of line 212 that will be executed:
```
this.input.blur();
```

mmitiche: I would suggest coming back to your previous solution with just a minor restructuring:

```
  handleEnter() {
    if (!this.isClearButtonTargeted) {
      const selectedSuggestion =
        this.suggestionListElement?.getCurrentSelectedValue();
      if (this.areSuggestionsOpen && selectedSuggestion) {
        this.sendSelectSuggestionEvent(selectedSuggestion.rawValue);
      } else {
        this.sendSubmitSearchEvent();
      }
      this.input.blur();
    }
  }
```
And
```
  onKeyup(event) {
    switch (event.key) {
      case keys.ENTER:
        this.handleEnter();
        break;
      case keys.ARROWUP:
        this.suggestionListElement?.selectionUp();
        break;
      case keys.ARROWDOWN:
        this.suggestionListElement?.selectionDown();
        break;
      default:
        this.handleKeyValues();
    }
    this.isClearButtonTargeted = false;
  }
```

SimonMilord: yeah makes sense, I saw it was still working then after the current version, but it might have not deployed properly or something

mmitiche (approved):

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -206,6 +211,12 @@ export default class QuanticSearchBoxInput extends LightningElement {
     this.input.blur();
   }
 
+  handleKeyDownOnClearButton(event) {
+    if (event.key === keys.ENTER) {
+      this.isClearButtonTargeted = true;
```
erocheleau, line 216:
```suggestion
      // Ignore next enter in searchbox input, to not trigger a search when we press enter on the clear button.
      this.ignoreNextEnter = true;
```
This needs to be explicit what it does, otherwise we'll never remember what or why we did this.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -180,14 +183,16 @@ export default class QuanticSearchBoxInput extends LightningElement {
   }
 
   handleEnter() {
-    const selectedSuggestion =
-      this.suggestionListElement?.getCurrentSelectedValue();
-    if (this.areSuggestionsOpen && selectedSuggestion) {
-      this.sendSelectSuggestionEvent(selectedSuggestion.rawValue);
-    } else {
-      this.sendSubmitSearchEvent();
+    if (!this.isClearButtonTargeted) {
```
erocheleau, line 186:
```suggestion
    if (!this.ignoreNextEnter) {
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -233,6 +244,7 @@ export default class QuanticSearchBoxInput extends LightningElement {
       default:
         this.handleKeyValues();
     }
+    this.isClearButtonTargeted = false;
```
erocheleau, line 247:
```suggestion
    this.ignoreNextEnter = false;
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/expandableSearchBoxInput.html
```diff
@@ -46,9 +46,10 @@
                 >
                   <lightning-button-icon
                     onclick={clearInput}
+                    onkeydown={handleKeyDownOnClearButton}
                     icon-name="utility:close"
-                    class="slds-grid searchbox__clear-button"
                     variant="bare"
+                    class="slds-grid searchbox__clear-button"
```
erocheleau, line 52: What is the effect of adding this class?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -93,6 +93,9 @@ export default class QuanticSearchBoxInput extends LightningElement {
     this.suggestionListElement?.resetSelection();
   }
 
+  /** @type {boolean} */
+  isClearButtonTargeted = false;
```
erocheleau, line 97:
```suggestion
  ignoreNextEnter = false;
```

erocheleau (commented):
The name needs to change. And I would suggest adding a comment to remember why we're doing this.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/expandableSearchBoxInput.html
SimonMilord: it was added in a separate PR mostly for the unit tests as a selector, but the change here is just the order in which it appears, I changed it so that the attributes are in the same order in both default and expandable searchBoxInput templates.

erocheleau (approved):
Looks good

PR #3538, opened 2024-01-23
fix(quantic): adjust citation tooltip width in specific situations inside Salesforce

mmitiche:
[SFINT-5350](https://coveord.atlassian.net/browse/SFINT-5350)

### The issue:
In some specific situations is Salesforce, like in the Hosted Insight Panel for example, the citations tooltip were being cropped because they were displayed inside a Salesforce container that had some stying rules that hide any thing that overflows the container:

Issue

### The solution

In order to avoid such situations, I added a new logic to the Quantic Tooltip component that limits the width of the tooltip relatively to the width of its target.
This way the tooltip will never be too big to be cropped.

Additionally, since we reduced the width, I alsso made the tooltip of the citations display the first 6 lines of the descriptions instead of 4 and this keep always approximately  displaying the same amount  of information. In other words I reduced the width of the citation tooltip but increased its height.

Screenshot 2024-01-23 at 6 43 51 AM

[SFINT-5350]: https://coveord.atlassian.net/browse/SFINT-5350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
@@ -1,6 +1,8 @@
 import {getAbsoluteWidth, getAbsoluteHeight} from 'c/quanticUtils';
 import {LightningElement, api} from 'lwc';
 
+const targetWidthToTooltipWidthRatio = 1.2;
+
 /**
  * The `QuanticTooltip` displays a tooltip containing a small amount of text that can be displayed when hovering over certain elements.
  * This component should be used inside a container with a CSS position attribute set to the value `relative`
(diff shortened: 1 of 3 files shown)


packages/quantic/force-app/main/default/lwc/quanticTooltip/quanticTooltip.js
```diff
@@ -55,8 +57,15 @@ export default class QuanticTooltip extends LightningElement {
 
   updateTooltipMaxWidth() {
     const windowWidth = window.innerWidth;
+    const targetWidth = getAbsoluteWidth(this.target);
+
+    const adaptedWidth =
+      targetWidth > 0
+        ? Math.min(windowWidth, targetWidth * targetWidthToTooltipWidthRatio)
```
SimonMilord, line 64: if I understand properly, we will set the width of the tooltip to 1.2 * the width of the citation if the width of the citation is > 0?

SimonMilord (approved):
nice solution! cleaner than using `getBoundingClientRect()` shenanigans  thats for sure :D

erocheleau (approved):
Everything looks good, please share to @johnnyCoveo to make sure everything is fine in this design :)

PR #3540, opened 2024-01-23
feat(commerce): isolate core commerce controllers

Spuffynism:
The current structure of having `{feature}/core` folders doesn't really make sense. Let's use a `core/{feature}` structure instead.
Before:
image
After:
image


fbeaudoincoveo (approved):
❤️

PR #3542, opened 2024-01-23
fix(quantic): fixed issue of copy to clipboard button not being  displayed

mmitiche:
[SFINT-5364](https://coveord.atlassian.net/browse/SFINT-5364)

Removed the logic that was making the issue of making the copy to clipboard button not appear in the generated answer component.

`navigator.clipboard.writeText()` is not supported yet in Salesforce service console, that's what was causing the issue.
 but the thing is that in Quantic, we have a fallback method that does the copy to clip board job without relying on `navigator.clipboard` so we don't really need this logic.

This condition has been initially added by the SVCC team cause they had this problem in Atomic, but in our side it's not needed cause we have a fallback method allowing us to copy to clipboard properly.

[SFINT-5364]: https://coveord.atlassian.net/browse/SFINT-5364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau (approved):
Wow sneaky thing

SimonMilord (approved):

PR #3543, opened 2024-01-23
feat(commerce): create commerce url and parameter managers

Spuffynism:
Create a few new controllers:
- A core url manager for commerce
- A core parameter manager for commerce
- Product listings & search url manager controllers
- Product listings & search parameter manager controllers

There are still a few TODOs to resolve (listed below), but I wanted to open the PR up for review. I've scoped the PR down to handling only the `q` query parameter, to reduce the size and complexity of the PR. In follow-up tasks, we'll handle sorting, pagination, facets, etc.

TODO:
- [x] Url managers tests
- [x] Documentation
- [x] TODOs in comments

Companion PR in barca: https://github.com/coveo/barca-sports/pull/154

[CAPI-339]

[CAPI-339]: https://coveord.atlassian.net/browse/CAPI-339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/commerce/search-parameters/search-parameter-serializer.ts
```diff
@@ -0,0 +1,69 @@
+import {CommerceSearchParameters, Parameters} from './search-parameter-actions';
```
fbeaudoincoveo, line 1: Mmm... in the end when we support all query params, that'll be a lot of duplication from the original search-parameter-serializer. Why do we need to duplicate this? Could we instead make the current search-parameter-serializer a tad more generic?

fbeaudoincoveo (commented):
Good start!

olamothe:
It seems like a lot of code to do something that Headless already has a bunch of function to handle already.

Did you look into the possibility to do (small) refactor to what's already available in `headless-core-search-parameter-manager` (and the surrounding files), making it a bit more generic if necessary ?

Spuffynism:
@olamothe Here's what I tried:
- Reusing the [search parameter serializer](https://github.com/coveo/ui-kit/blob/6be4d6df728a793c23461267368a840b264f8f87/packages/headless/src/features/search-parameters/search-parameter-serializer.ts). 
Since it's intimately tied to the [SearchParameters](https://github.com/coveo/ui-kit/blob/338e740bfa108cead0785f34490192ed3bd50fd6/packages/headless/src/features/search-parameters/search-parameter-actions.ts#L10), it's difficult to reuse code without making it extremely generic. By using the approach in this PR, we share the logic that we _can_ share with the current parameter serializer, without completely re-writing the parameter manager.
- Turning the current [url manager](https://github.com/coveo/ui-kit/blob/6be4d6df728a793c23461267368a840b264f8f87/packages/headless/src/controllers/url-manager/headless-url-manager.ts) into a core.
Same thing as my previous point: the controller is tightly bound to the engine and its state. If we want to reuse it, we'd need to create a core controller, reverse its dependencies (a bit like I did on the [commerce core url manager controller](https://github.com/coveo/ui-kit/blob/326480eea544ad561fd02f34431b154338eebc6b/packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.ts#L102) in this PR) by passing it state selectors and action creators.

  If we want consistency between the current search controller interfaces and the commerce ones, this would be an option. However, I don't think that's the case. Therefore, I preferred going with the approach of creating core controllers for commerce only.
  
  
Are there assumptions I'm making that I shouldn't be?

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.test.ts
louis-bompart, line 1: Should use the mock-engine-v2

packages/headless/src/controllers/commerce/core/url-manager/headless-core-url-manager.test.ts
louis-bompart, line 1: Should use the mock-engine-v2

packages/headless/src/controllers/commerce/product-listing/parameter-manager/headless-product-listing-parameter-manager.test.ts
louis-bompart, line 1: Should use the mock-engine-v2

packages/headless/src/controllers/commerce/product-listing/url-manager/headless-product-listing-url-manager.test.ts
louis-bompart, line 1: Should use the mock-engine-v2

packages/headless/src/controllers/commerce/search/parameter-manager/headless-search-parameter-manager.test.ts
louis-bompart, line 1: Should use the mock-engine-v2

packages/headless/src/controllers/commerce/search/url-manager/headless-search-url-manager.test.ts
louis-bompart, line 1: Should use the mock-engine-v2

louis-bompart (changes requested):
We're now mandating that every PR that includes or modify tests to use the new testing methodology.

See #3509 and #3519 for examples on how to do so
And https://coveo.slack.com/archives/G016TA2G485/p1705095628101109 for more context

Spuffynism:
@louis-bompart @fredboutin Re-requesting your reviews since I've applied the new testing approach in https://github.com/coveo/ui-kit/pull/3543/commits/ee71e48ab55f976154f6836c530b6907b75434de! 😁

louis-bompart (approved):

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    initEngine();
+    initParameterManager();
+  });
+
+  it('exposes #subscribe method', () => {
+    expect(parameterManager.subscribe).toBeTruthy();
+  });
+
+  describe('initialization', () => {
+    it('validates initial state against schema', () => {});
```
fbeaudoincoveo, line 55: This test currently does nothing hehe

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        const mockedExecuteSearch = jest.mocked(executeSearch);
+        const parameters = {};
+        parameterManager.synchronize(parameters);
+        expect(mockedExecuteSearch).toHaveBeenCalled();
+      });
+    });
+  });
+
+  describe('#state', () => {
+    it('contains #parameters', () => {
```
fbeaudoincoveo, line 104: Should we test that #parameters invokes the #activeParametersSelector?

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    jest.resetAllMocks();
+    initEngine();
+    initParameterManager();
+  });
+
+  it('exposes #subscribe method', () => {
+    expect(parameterManager.subscribe).toBeTruthy();
+  });
+
+  describe('initialization', () => {
```
fbeaudoincoveo, line 54: Maybe add the (kind of trivial) test "initializes".

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    it('validates initial state against schema', () => {});
+
+    it('dispatches #restoreActionCreator', () => {
```
fbeaudoincoveo, line 57: Feel free to apply or not, but you could also test with a non empty initial state.

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  parametersDefinition: SchemaDefinition<Required<T>>;
+
+  /**
+   * The selector to retrieve the active parameters from the state.
+   */
+  activeParametersSelector: (state: CommerceEngine['state']) => T;
+
+  /**
+   * The action to dispatch to update the parameters in state.
```
fbeaudoincoveo, line 32:
```suggestion
   * The action to dispatch to update the parameters in the state.
```

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+) =>
+  new Schema<Required<ParameterManagerInitialState<T>>>({
+    parameters: new RecordValue({
+      options: {required: true},
+      values: parametersDefinition as SchemaDefinition<Record<string, Object>>,
+    }),
+  });
+
+/**
+ * The `ParameterManager` controller allows restoring parameters that affect the results (e.g., from the a URL).
```
fbeaudoincoveo, line 71:
```suggestion
 * The `ParameterManager` controller allows restoring parameters that affect the results (e.g., from the URL).
```

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * The state relevant to the `SearchParameterManager` controller.
+   */
+  state: ParameterManagerState<T>;
+}
+
+export interface ParameterManagerState<T extends Parameters> {
+  /**
+   * The parameters affecting the search response.
```
fbeaudoincoveo, line 89:
```suggestion
   * The parameters affecting the response.
```
since it's the core manager we might as well keep the description more generic.

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  parameters: T;
+}
+
+/**
+ * @internal
+ * Creates a `ParameterManager` controller instance.
+ *
+ * @param engine - The headless engine.
```
fbeaudoincoveo, line 98:
```suggestion
 * @param engine - The headless commerce engine.
```

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.test.ts
fbeaudoincoveo: (same comment elsewhere when applicable)

packages/headless/src/controllers/commerce/core/url-manager/headless-core-url-manager.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    it('initial #restoreActionCreator should parse the "active" fragment', () => {
+      initUrlManager('q=windmill');
+      testLatestRestoreSearchParameters({
+        q: 'windmill',
+      });
+    });
+  });
+
+  describe('state', () => {
+    it('contains the serialized fragment of the search parameters state', () => {
```
fbeaudoincoveo, line 79: Since we have tests for the serializer itself, an alternative would be to test that state.fragment calls the serializer.

packages/headless/src/controllers/commerce/core/parameter-manager/headless-core-parameter-manager.test.ts
fbeaudoincoveo: Similar question/comment elewhere when applicable

packages/headless/src/features/commerce/search-parameters/search-parameter-serializer.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    it('deserializes a string with special characters', () => {
+      someSpecialCharactersThatNeedsEncoding.forEach((char) => {
+        const result = deserialize(
+          `q=${encodeURIComponent(char)}&enableQuerySyntax=true`
+        );
+        expect(result).toEqual({q: char});
+      });
+    });
+
+    it('deserializes a string where the value contains an equals sign (e.g. q)', () => {
```
fbeaudoincoveo, line 72:
```suggestion
    it('deserializes a string where the value contains an equals sign (e.g., q)', () => {
```

fbeaudoincoveo (approved):

packages/headless/src/controllers/commerce/core/url-manager/headless-core-url-manager.test.ts
Spuffynism: In line with [Redux's recommendations on testing](https://redux.js.org/usage/writing-tests#selectors), I'd prefer leaving the test as-is. This test ensures the output is as expected, given a certain state.

> Selectors are also generally pure functions, and thus can be tested using the same basic approach as reducers: set up an initial value, call the selector function with those inputs, and **assert that the result matches the expected output**.

PR #3544, opened 2024-01-23
feat(headless,atomic): migrate framework versions from customData to source w/ next analytics

louis-bompart:
Context, the new schema defined a new place for Coveo frameworks to report their versions: https://github.com/coveo/analytics_schema/pull/292.

This PR:
 - Add a new config parameter in headless: `configuration.analytics.atomicVersion`. This will replace the analyticsMiddleware that atomic was using to set its version in the payloads.
 - When `analyticsMode=next`, set the Headless version & Atomic Version (if any) to:
    - The instantiation parameters of Relay
    - The analytics section of the request body of the call made to the Search API
 - When `analyticsMode=next`, do not set  `coveoHeadlessVersion` anymore in the customData. 
 
 
 KIT-2930

### DIFF
--- a/packages/atomic/src/components/common/interface/analytics-config.ts
+++ b/packages/atomic/src/components/common/interface/analytics-config.ts
@@ -37,8 +37,15 @@ export function augmentAnalyticsConfigWithDocument(): AnalyticsConfiguration {
   };
 }
 
+const versionMatcher = /^(\d+\.\d+\.\d+)/;
+
 export function augmentAnalyticsConfigWithAtomicVersion(): Required<
-  Pick<AnalyticsConfiguration, 'frameworkVersions'>
+  Pick<AnalyticsConfiguration, 'source'>
 > {
-  return {frameworkVersions: {'@coveo/atomic': getAtomicEnvironment().version}};
+  return {
+    source: {
+      '@coveo/atomic':
+        versionMatcher.exec(getAtomicEnvironment().version)?.[0] || '0.0.0',
+    },
+  };
 }
--- a/packages/atomic/src/components/common/interface/analytics-config.ts
+++ b/packages/atomic/src/components/common/interface/analytics-config.ts
@@ -38,7 +38,7 @@ export function augmentAnalyticsConfigWithDocument(): AnalyticsConfiguration {
 }
 
 export function augmentAnalyticsConfigWithAtomicVersion(): Required<
-  Pick<AnalyticsConfiguration, 'atomicVersion'>
+  Pick<AnalyticsConfiguration, 'frameworkVersions'>
 > {
-  return {atomicVersion: getAtomicEnvironment().version};
+  return {frameworkVersions: {'@coveo/atomic': getAtomicEnvironment().version}};
 }
--- a/packages/atomic/src/components/common/interface/analytics-config.ts
+++ b/packages/atomic/src/components/common/interface/analytics-config.ts
@@ -36,3 +36,9 @@ export function augmentAnalyticsConfigWithDocument(): AnalyticsConfiguration {
     ...(document.referrer && {originLevel3: document.referrer}),
   };
 }
+
+export function augmentAnalyticsConfigWithAtomicVersion(): Required<
+  Pick<AnalyticsConfiguration, 'atomicVersion'>
+> {
+  return {atomicVersion: getAtomicEnvironment().version};
+}
--- a/packages/headless/src/api/analytics/search-analytics.test.ts
+++ b/packages/headless/src/api/analytics/search-analytics.test.ts
@@ -508,15 +508,17 @@ describe('#configureAnalytics', () => {
 });
 
 describe('#getAnalyticsSources', () => {
-  it('without atomicVersion, returns an array only with `@coveo/headless`', () => {
+  it('without a frameworkVersions, returns an array only with `@coveo/headless`', () => {
     const state = createMockState();
     expect(getAnalyticsSource(state.configuration.analytics)).toEqual([
       `@coveo/headless@${VERSION}`,
     ]);
   });
-  it('with atomicVersion, returns an array only with `@coveo/headless`', () => {
+  it('with a frameworkVersions, returns an array only with `@coveo/headless`', () => {
     const state = createMockState();
-    state.configuration.analytics.atomicVersion = '1.2.3';
+    state.configuration.analytics.frameworkVersions = {
+      '@coveo/atomic': '1.2.3',
+    };
     expect(getAnalyticsSource(state.configuration.analytics)).toEqual([
       `@coveo/headless@${VERSION}`,
       '@coveo/atomic@1.2.3',
--- a/packages/headless/src/api/analytics/search-analytics.test.ts
+++ b/packages/headless/src/api/analytics/search-analytics.test.ts
@@ -27,6 +27,7 @@ import {QuerySuggestCompletion} from '../search/query-suggest/query-suggest-resp
 import {
   configureAnalytics,
   configureLegacyAnalytics,
+  getAnalyticsSource,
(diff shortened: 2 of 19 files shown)


packages/headless/src/app/engine-configuration.ts
```diff
@@ -155,7 +155,17 @@ export interface AnalyticsConfiguration {
...
+   * Specifies the analytics mode to use.
+   * By default, `legacy`.
+   * @internal
+   */
   analyticsMode?: 'legacy' | 'next';
+  /**
+   * The version of `@coveo/atomic` used.
+   * @internal
+   */
+  atomicVersion?: string;
```
olamothe, line 168: This feels like cheating, by adding a key that's named specifically "for us"...

What about Quantic, or a partner building/investing into a UI layer with Headless ?

How about naming it generically ? 
```js
uiVersion: {name: string; version: string; owner: string; }

// eg:
// uiVersion: {name: 'atomic', version: '1.2345', owner: '@coveo'}
// uiVersion:{name: 'foo', version: '4.5678', owner: '@acme.inc'}
```

And then we can adapt `getAnalyticsSource` to take that generic object and format it properly

louis-bompart: So the 'for us' here is by design, but I scoped it too narrowly, you're right.
By that, I mean we want to allow reporting only on our framework, to ensure a certain purity of the `source` field for the analytics.
However, you're right that it should not be Atomic-Exclusive.
There'll be an allow-list of frameworks (for now, Quantic & Atomic).

However, I'll sync up with DP soon because it's been 2 months ish since I had those discussion and my memory might be a bit fuzzy

louis-bompart: So, thought about it while doing the change in the implementation and verifying the background work done in the backend.
I see nothing blocking preventing people from indicating their custom framework in the source parameter.
However, I do prefer to take the cautious route and keep the API narrow and extend it if the need arises, a
So, I'd keep things as they currently are as of 3a646c9 for the time being

If we need to add singular keys, we can adopt a punch-hole strategy and allow them one by one.
If we need to open the gate and allow more, the format is extensible and should allow us to allow custom frameworks to hook themselves. 

(moreover, going from loose to strict is hard to do without breaking changes/degradation of Headless API)

olamothe (approved):
This is better, with the possibility of generalizing this to be more than just our UI if necessary without breaking changes.

`frameworkVersions`, I'm not a fan of as a name, I won't die on this hill though :) 

I propose simply `source`, since `analytics.source` seems descriptive enough, and it match with the name accepted by the schemas/what you'd see going over the network browser call through relay.

olamothe:
Also, do we send this type of info to Search API call ? I don't think so.... 🤔 Haven't checked.

ie: In analytics payload of search API request, is there a place where we tell it that "this was logged by coveo/atomic@1.1234" ?

So that it gets tracked for server side events also ?

louis-bompart:
> Also, do we send this type of info to Search API call ? I don't think so.... 🤔 Haven't checked.
> 
> ie: In analytics payload of search API request, is there a place where we tell it that "this was requested by coveo/atomic@1.1234" ?
> 
> So that it gets tracked for server-side events also?

That's done in this PR, 84d482c.

I'll open a thread in Slack to triple-check tho that the ingestion will be done.

packages/headless/src/api/analytics/search-analytics.test.ts
```diff
@@ -472,3 +478,50 @@ describe('search analytics', () => {
...
+});
+
+describe('#getAnalyticsSources', () => {
+  it('without a frameworkVersions, returns an array only with `@coveo/headless`', () => {
+    const state = createMockState();
+    expect(getAnalyticsSource(state.configuration.analytics)).toEqual([
+      `@coveo/headless@${VERSION}`,
+    ]);
+  });
+  it('with a frameworkVersions, returns an array only with `@coveo/headless`', () => {
```
fbeaudoincoveo, line 517:
```suggestion
  it('with a frameworkVersions, returns an array with `@coveo/headless` and `@coveo/atomic`', () => {
```

packages/headless/src/features/configuration/configuration-actions.ts
```diff
@@ -148,37 +157,60 @@ export interface UpdateAnalyticsConfigurationActionCreatorPayload {
...
   trackingId?: string;
-
+  /**
+   * Specifies the analytics mode to use.
+   * By default, `legacy`.
+   * @internal
+   */
   analyticsMode?: 'legacy' | 'next';
+  /**
+   * The version of `@coveo/atomic` used.
```
fbeaudoincoveo, line 167:
```suggestion
   * The version of `@coveo/atomic` or `@coveo/quantic` used.
```

packages/headless/src/app/engine-configuration.ts
```diff
@@ -155,7 +156,17 @@ export interface AnalyticsConfiguration {
...
    */
   trackingId?: string;
+  /**
+   * Specifies the analytics mode to use.
+   * By default, `legacy`.
+   * @internal
+   */
   analyticsMode?: 'legacy' | 'next';
+  /**
+   * The version of `@coveo/atomic` used.
```
fbeaudoincoveo, line 166:
```suggestion
   * The version of `@coveo/atomic` or `@coveo/quantic` used.
```

packages/headless/src/features/configuration/configuration-state.ts
```diff
@@ -127,8 +128,17 @@ export interface AnalyticsState {
...
   trackingId: string;
-
+  /**
+   * Specifies the analytics mode to use.
+   * By default, `legacy`.
+   * @internal
+   */
   analyticsMode: 'legacy' | 'next';
+  /**
+   * The version of `@coveo/atomic` used.
```
fbeaudoincoveo, line 138: [See previous comment](https://github.com/coveo/ui-kit/pull/3544/files#diff-21cfc0c7d2b75f9d016dfb8c09b9322767bd002b34634cca40fdd8058252058fR167)

packages/headless/src/utils/validate-payload.ts
```diff
@@ -35,6 +35,12 @@ export const nonEmptyStringArray = new ArrayValue({
   required: true,
 });
 
+export const optionalNonEmptyVersionString = new StringValue({
+  required: false,
+  emptyAllowed: false,
+  regex: /^\d+\.\d+\.\d+$/,
```
fbeaudoincoveo, line 41: This won't match "custom" versions such as `2.42.1-query-suggestion-fallback-poc` (or any pre-release version).

fbeaudoincoveo (approved):

packages/headless/src/utils/validate-payload.ts
louis-bompart: Good point 👍 Although, this is fixed/coming from the analytics schema.
So instead, I'll change the logic in Atomic so that it only send the matching bit, not the rest (i.e. `/^\d+\.\d+\.\d+/`, no `$`

packages/atomic/src/components/common/interface/analytics-config.ts
```diff
@@ -36,3 +36,9 @@ export function augmentAnalyticsConfigWithDocument(): AnalyticsConfiguration {
     ...(document.referrer && {originLevel3: document.referrer}),
   };
 }
+
+export function augmentAnalyticsConfigWithAtomicVersion(): Required<
+  Pick<AnalyticsConfiguration, 'frameworkVersions'>
```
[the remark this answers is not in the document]
fbeaudoincoveo: I realize that this is kind of the opposite of what @olamothe was suggesting haha

I agree with Oli that "source" would be a better name if we want to keep it generic.

PR #3545, opened 2024-01-24
chore(headless): refactor Type into BaseFacetRequest

louis-bompart:
I noticed that every interfaces that extends `Type` also extends BaseFacetRequest (make sense).
I think that refactoring `Type` into `BaseFacetRequest` is a good thing to do because it makes the latter more representative of what is sent to the SearchAPI: i.e. If I want to describe what's sent to describe a facet, regardless of its type, I think BaseFacetRequest is the one to use. It was missing the `type` field, now it doesn't.

Trade-off tho, there was a single place where we decided to extract FacetId from the BaseFacetRequest, I'm assuming to DRY the JSDOC associated with this field.
To solve that there were no pretty solutions I think, here's a few I explored;
 - Extract FieldId into another interface: It recreates the same issue I'm trying to solve here, and is tedious :x:
 - Duplicate: Meh, not a fan, especially because the uses is mostly around documentation, we're very likely to forget I reckon, :x:
 - Set a default value to TFacetType: I think that setting 'specific' by default would have been a source of error, so I chose not to :x:
 - Use a specific facetType, even if that's wrong: That's what I chose, because it isolate the problem in a corner 😂

### DIFF
--- a/packages/headless/src/features/facets/category-facet-set/interfaces/request.ts
+++ b/packages/headless/src/features/facets/category-facet-set/interfaces/request.ts
@@ -2,7 +2,6 @@ import {
   BaseFacetRequest,
   CurrentValues,
   Delimitable,
-  Type,
   SortCriteria,
   BaseFacetValueRequest,
 } from '../../facet-api/request';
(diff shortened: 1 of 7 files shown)


fbeaudoincoveo (commented):
While I don't strongly disagree with this refactoring, I personally find it slightly more awkward than what we currently have.

The way I see it, the `type` facet request parameter acts as a "discriminator" to determine which parameters are supported/required by the request in addition to the "base" ones. In that sense, it feels fitting to me that for the facet request `type` to be defined in each specific facet request interface.

PR #3548, opened 2024-01-25
fix(quantic): fixed issue with the Quantic Search Box Input

mmitiche:
[SFINT-5374](https://coveord.atlassian.net/browse/SFINT-5374)

#### Work made in this PR:

- Issue fixed.
- New test added to make sure this bug will not happen again.
- Test improved to execute the whole tests for both the default search box and the standalone searchbox.

#### The issue

The issue appears after executing the following actions:

- Enter a query in the search box, for example: “abc“.
- Click the down arrow key.
- Click the del key to delete the last character in the search box.
- The query transforms to a whole other word.

https://github.com/coveo/ui-kit/assets/86681870/1464d98a-1716-40ac-9441-92225f069856

#### The solution

The issue was occurring because the clicking the Down arrow key was triggering the selection of the first suggestion of the suggestions list, however at the moment of the click on the Down Arrow the suggestions list was empty causing this unexpected behaviour.

To fix this I added a condition that displays the suggestions list only when it's not empty.

That was the original behaviour in the Quantic Search Box component, but we have missed that in the last refactoring.

https://github.com/coveo/ui-kit/assets/86681870/deaafe03-0b7c-4c2b-8e12-7dfcc8e4cfdb

[SFINT-5374]: https://coveord.atlassian.net/browse/SFINT-5374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
@@ -185,7 +185,7 @@ describe('c-quantic-search-box-input', () => {
     });
   });
 
-  describe('when suggestions are found', () => {
+  describe('when the suggestions list is not empty', () => {
     it('should display the suggestions in the suggestions list', async () => {
       const element = createTestComponent({
         ...defaultOptions,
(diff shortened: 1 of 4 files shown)


packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -86,234 +86,266 @@ describe('c-quantic-search-box-input', () => {
...
-      it('should dispatch a #quantic__selectsuggestion event with the selected suggestion as payload', async () => {
+  [true, false].forEach((textareaValue) => {
+    describe(`when the textarea property is set to ${textareaValue}`, () => {
+      it(`should display the ${
+        textareaValue ? 'expandable' : 'default'
+      } search box properly`, async () => {
         const element = createTestComponent({
           ...defaultOptions,
-          suggestions: mockSuggestions,
+          textarea: true,
```
SimonMilord, line 96:
```suggestion
          textarea: textareaValue,
```

shouldnt this be set to `textareaValue` ?

SimonMilord (approved):

erocheleau (approved):
Just the comment from Simon to fix in your test suite I think, otherwise good fix!

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
mmitiche: Indeed, fixed!

PR #3551, opened 2024-01-25
fix(headless): remove un-necessary client side events for new query correction

olamothe:
After testing against Barca sport, this client side UA event is ultimately un-necessary. 

We just need one to be logged (returned from this function):

```
analyticsAction: logDidYouMeanAutomatic(),
```

Importantly, `logDidYouMeanAutomatic` will log `q={corrected query}`, which is what ML will learn against, and not the typo.

https://coveord.atlassian.net/browse/KIT-2965


louis-bompart (approved):
LGTM, did you check if the SAPI w/ SSSE, for consistency sake?

fbeaudoincoveo (approved):

PR #3552, opened 2024-01-26
fix(headless): Improving how to handle the subscribeStateManager in the rga controller

mmitiche:
[SFINT-5372](https://coveord.atlassian.net/browse/SFINT-5372)

### The issue

The Generated Answer Core controller has a dedicated logic built to trigger new generated answer streamings whenever a new stream ID is received after new search queries. 
This logic is handled by an object called the `SubscribeStateManager` it's a global object that globally stores the `last stream ID` used.
 
The following is a small explanation of this logic:

1.  When a new generated answer controller is created, it subscribes to engine state changes using the function `subscribeToSearchRequests` that comes with the globally declared object the `SubscribeStateManager`. 

2. The function `subscribeToSearchRequests` will make the controller execute a listener whenever a change in the state of the engine occurs.

3. The listener executed will compare the `last stream ID` stored globally in  the object `SubscribeStateManager` to the `stream ID` present in the state, if they are different then it initiates a new streaming and it sets a new value for the `last stream ID` in the `SubscribeStateManager` global object.

### Example Scenario:

For example imagine we have `Engine A` and `Engine B`, both are implementing RGA, so they have respectively two controllers: `controller A`  and controller `controller B`:

Initially let's say we have the following:
```
SubscribeStateManager = {lastStreamID: 'foo'}
```

- A new search query that returns a stream ID : `bar`,  is being made by `Engine A`. `bar` is compared with `foo`, since they are different a new streaming is initiated for Engine A and `bar` is saved globally in the object the `SubscribeStateManager`.
- A new search query that returns a stream ID: `qux`,   is being made by `Engine B`. `qux` is compared with `bar,` since they are different a new streaming is initiated for Engine B and `qux` is saved globally in the object the `SubscribeStateManager`.
- Now a state change in `Engine A` occurs without  having a new stream id, the listener get executed and this time it compares `bar` which is stored in the state of `Engine A` to `qux` which is the last stream id stored in the object `SubscribeStateManager`, since these values are different a new streaming is initiated while it should NOT. 

### Visualization of the issue:
The RGA component suddenly disappears:

https://github.com/coveo/ui-kit/assets/86681870/d7e7f4f8-519c-4021-ba88-2d7d063b416e

# The solution:

Instead of having one global object that all the engines speaks  to, now we are kinda having a global object for each engine, so the information about the `last stream ID` is stored separately for each engine.

To achieve this I'm saving an ID for each engine that allows to differentiate the logic of each engine and initiate the streaming of each engine properly.

New unit test have been added to cover this bug.  

### Visualization of the solution:

https://github.com/coveo/ui-kit/assets/86681870/7ee033d1-e4be-4519-b5f0-531587a95a91

[SFINT-5372]: https://coveord.atlassian.net/browse/SFINT-5372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -241,7 +245,7 @@ export function buildCoreGeneratedAnswer(
   if (fieldsToIncludeInCitations) {
     dispatch(registerFieldsToIncludeInCitations(fieldsToIncludeInCitations));
   }
-
+  const subscribeStateManager = new SubscribeStateManager();
```
nathanlb, line 248: Doesn't this create one SubscribeStateManager per controlller and restore the issue we had that led to it's creation intially?
If we have multiple generatedAnswer controllers in the same search interface are they in sync like they should be?

mmitiche: Indeed this restores the issues we had when multiple generated answer controllers with the same engine are working in the same time.

Even though the unit tests are passing successfully. But I just found that the unit tests to test this case were not setup properly and thus were passing all the time.

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.test.ts
```diff
@@ -368,33 +378,66 @@ describe('generated answer', () => {
...
+
+        callListener(engine);
+        callListener(secondEngine);
+        callListener(engine);
+        callListener(secondEngine);
+
+        const streamAnswerCountInFirstEngine = engine.actions.filter(
+          (a) => a.type === streamAnswer.pending.type
+        ).length;
+        const streamAnswerCountInSecondEngine = engine.actions.filter(
```
mmitiche, line 435:
```suggestion
        const streamAnswerCountInSecondEngine = secondEngine.actions.filter(
```

olamothe (commented):
> This logic is handled by an object called the SubscribeStateManager it's a global object that globally stores the last stream ID used.

Why does this need to be global ? Would it be solved by just making `subscribeStateManager` a local variable to `buildGeneratedAnswer` function ?

mmitiche:
> > This logic is handled by an object called the SubscribeStateManager it's a global object that globally stores the last stream ID used.
> 
> Why does this need to be global ? Would it be solved by just making `subscribeStateManager` a local variable to `buildGeneratedAnswer` function ?
@olamothe 
If we make `subscribeStateManager` it means that every controller will have it's own variable and this was causing another issue:

Multiple controllers tied to the same engine will all try to initiate a new streaming whenever a new stream ID is received, and this was causing several problems like the interference of the streamings lunched from the different controllers and the unexpected cancelation of other streams.

erocheleau:
> > > This logic is handled by an object called the SubscribeStateManager it's a global object that globally stores the last stream ID used.
> > 
> > 
> > Why does this need to be global ? Would it be solved by just making `subscribeStateManager` a local variable to `buildGeneratedAnswer` function ?
> 
> @olamothe If we make `subscribeStateManager` it means that every controller will have it's own variable and this was causing another issue:
> 
> Multiple controllers tied to the same engine will all try to initiate a new streaming whenever a new stream ID is received, and this was causing several problems like the interference of the streamings lunched from the different controllers and the unexpected cancelation of other streams.
> 
> Initially when we first created this controller it was using `subscribeStateManager` as a local variable but as we noticed the issues mentioned above we fixed this in this PR: [coveo/salesforceintegration#876](https://github.com/coveo/salesforceintegration/pull/876)

In other words, we need one and only one stream per engine, so we made it a global thing, but we didn't anticipate having multiple engine on the same "page".

erocheleau (approved):
I think this all makes sense with what we discussed together 👍 
Let's see if someone has a brilliant idea :)

louis-bompart:
I think the issue is fundamentally that the subscription manager should not be module-scoped (or as you said global).
Actually, IMHO, apart from functions `buildXYZ` and constants, anything that is JS and module-scoped in Headless is likely to be a smell.
Do you have technical design documents around this piece of the RGA code?

louis-bompart (approved):

olamothe (approved):

PR #3558, opened 2024-01-29
fix(headless): limit memory usage and property comparison for history manager

olamothe:
As discussed on slack, this is a performance fix related to history management in Headless.

* Remove instant result from history management. It is completely un-necessary to keep it in memory to restore history on navigation. This change on it's own would probably fix (most) of the problem for the original maintenance ticket.

* Limit the amount of data kept in memory for `past` history entry, by limiting it to 10 entry. This change on it's own would probably fix (most) of the problem for the original maintenance ticket.

* Do a local optimization in history-manager controller to only compare `current-snapshot` vs `previous-snapshot` with a JSON.stringify compare. This override the "base controller" generic comparison function, which does a full JSON.stringify compare on all properties of it's local state (ie: would stringify and compare all `past` entries). This change is not strictly necessary to fix the original ticket.

* Mark future improvements for v3 as TODO (as explained on slack, those changes would be breaking in current minor version). JIRA which explains the future improvement we can do in v3: https://coveord.atlassian.net/browse/KIT-2969

https://coveord.atlassian.net/browse/KIT-2970

### DIFF
--- a/packages/headless/src/app/undoable.test.ts
+++ b/packages/headless/src/app/undoable.test.ts
@@ -87,5 +87,36 @@ describe('undoable', () => {
         future: [],
       });
     });
+
+    it('supports a maximum of 10 past history entries', () => {
+      const mockUndoableReducer = undoable({
+        reducer,
+        actionTypes: {
+          redo: redo().type,
+          undo: undo().type,
+          snapshot: snapshot().type,
+        },
+      });
+      const initialOverflow = 50;
+
+      const initialState = {
+        past: Array.from(
+          {length: initialOverflow},
+          (_, i) => `entry-overflow-${i}`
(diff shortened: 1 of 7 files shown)


fbeaudoincoveo (approved):

jfallaire (approved):
Should we add an option on historyManager for max past history count ? Default to 10 i'm fine with it but what if 10 is not enough or too much for xyz reason

olamothe:
> Should we add an option on historyManager for max past history count ? Default to 10 i'm fine with it but what if 10 is not enough or too much for xyz reasons

I prefer to wait until this is a real need before adding an option. It's easier to add than to remove in the case of configuration option.

PR #3559, opened 2024-01-29
ASM proposal

louis-bompart:
Reducer based-approach, where the bit of code relative to the abort controller

### DIFF
--- a/packages/headless/src/controllers/core/generated-answer/headless-answer-stream-manager.ts
+++ b/packages/headless/src/controllers/core/generated-answer/headless-answer-stream-manager.ts
@@ -1,0 +1,53 @@
+import {GeneratedAnswerAPIClient} from '../../../api/generated-answer/generated-answer-client';
+import {CoreEngine} from '../../../app/engine';
+import {ClientThunkExtraArguments} from '../../../app/thunk-extra-arguments';
+import {
+  notifyStreamAborted,
+  registerAnswerStreamManager,
+  streamAnswer,
+} from '../../../features/generated-answer/generated-answer-actions';
+import {
+  GeneratedAnswerSection,
+  SearchSection,
+  DebugSection,
+} from '../../../state/state-sections';
+
+type AnswerStreamManagerEngine = CoreEngine<
+  GeneratedAnswerSection & SearchSection & DebugSection,
+  ClientThunkExtraArguments<GeneratedAnswerAPIClient>
+>;
+
+export function buildAnswerStreamManager(engine: AnswerStreamManagerEngine) {
+  if (engineHasAnswerStreamManager(engine)) {
+    return;
+  }
+
+  let abortController: AbortController | undefined;
+  const setAbortControllerRef = (ref: AbortController) => {
+    abortController = ref;
+  };
+
+  function abortStream() {
+    abortController?.abort();
+    abortController = undefined;
+  }
+
+  engine.dispatch(registerAnswerStreamManager);
+  engine.subscribe(() => {
+    if (engine.state.generatedAnswer.shouldStartStream) {
+      abortStream();
+      engine.dispatch(
+        streamAnswer({
+          setAbortControllerRef,
+        })
+      );
+    } else if (engine.state.generatedAnswer.shouldAbortStream) {
+      abortStream();
+      engine.dispatch(notifyStreamAborted);
+    }
+  });
+}
+
+function engineHasAnswerStreamManager(engine: AnswerStreamManagerEngine) {
+  return engine.state.generatedAnswer.hasAnswerStreamManager;
+}
--- a/packages/headless/src/features/generated-answer/generated-answer-slice.ts
+++ b/packages/headless/src/features/generated-answer/generated-answer-slice.ts
@@ -1,5 +1,7 @@
 import {createReducer} from '@reduxjs/toolkit';
+import type {WritableDraft} from 'immer/dist/internal';
 import {RETRYABLE_STREAM_ERROR_CODE} from '../../api/generated-answer/generated-answer-client';
+import {executeSearch} from '../search/search-actions';
 import {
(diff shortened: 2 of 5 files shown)


packages/headless/src/features/generated-answer/generated-answer-slice.ts
```diff
@@ -68,12 +75,19 @@ export const generatedAnswerReducer = createReducer(
...
-          fieldsToIncludeInCitations: state.fieldsToIncludeInCitations,
-          isVisible: state.isVisible,
-        };
+        return handleResetAnswer(state);
+      })
+      .addCase(executeSearch.fulfilled, (state, payload) => {
+        const streamId =
+          payload.payload.response.extendedResults
+            .generativeQuestionAnsweringId;
+        return state.currentStreamId === streamId
```
louis-bompart, line 84: This is the important line:
We moved the 'current-to-next' comparison to the reducer, which allows us to remove storage of state in in the ASM.

packages/headless/src/controllers/core/generated-answer/headless-answer-stream-manager.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type AnswerStreamManagerEngine = CoreEngine<
+  GeneratedAnswerSection & SearchSection & DebugSection,
+  ClientThunkExtraArguments<GeneratedAnswerAPIClient>
+>;
+
+export function buildAnswerStreamManager(engine: AnswerStreamManagerEngine) {
+  if (engineHasAnswerStreamManager(engine)) {
+    return;
+  }
```
louis-bompart, line 23: This + L35 of the same file ensures we only have a single instance per engine.

packages/headless/src/controllers/core/generated-answer/headless-answer-stream-manager.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  engine.dispatch(registerAnswerStreamManager);
+  engine.subscribe(() => {
+    if (engine.state.generatedAnswer.shouldStartStream) {
+      abortStream();
+      engine.dispatch(
+        streamAnswer({
+          setAbortControllerRef,
+        })
+      );
```
mmitiche, line 43: I think `shouldStartStream`  should be reset to `false` in the state after dispatching the `streamAnswer` action.

Otherwise we would just keep streaming answers infinitely for every change that occurs on the state.

packages/headless/src/features/generated-answer/generated-answer-slice.ts
```diff
@@ -68,12 +75,19 @@ export const generatedAnswerReducer = createReducer(
...
       .addCase(resetAnswer, (state) => {
-        return {
-          ...getGeneratedAnswerInitialState(),
-          responseFormat: state.responseFormat,
-          fieldsToIncludeInCitations: state.fieldsToIncludeInCitations,
-          isVisible: state.isVisible,
-        };
+        return handleResetAnswer(state);
+      })
+      .addCase(executeSearch.fulfilled, (state, payload) => {
```
mmitiche, line 80: @nathanlb I remember you told me Friday about a situation where triggering live streams on   `executeSearch.fulfilled` was not enough, can you remember why?

packages/headless/src/features/generated-answer/generated-answer-slice.ts
```diff
@@ -68,12 +76,22 @@ export const generatedAnswerReducer = createReducer(
...
+        const streamId =
+          payload.payload.response.extendedResults
+            .generativeQuestionAnsweringId;
+        return state.currentStreamId === streamId
+          ? state
+          : {
+              ...handleResetAnswer(state),
+              shouldStartStream: !!streamId,
+              currentStreamId: streamId,
```
mmitiche, line 90: When the stream ID is truthy, I think we should set `shouldAbortStream` to false.
Since by setting `shouldStartStream ` to true the ASM listener will already abort the previous stream, no?
```suggestion
              ...handleResetAnswer(state),
              shouldStartStream: !!streamId,
              shouldAbortStream: !streamId,
              currentStreamId: streamId,
```

mmitiche (commented):
Very neat solution!

packages/headless/src/features/generated-answer/generated-answer-slice.ts
nathanlb: I remember considering it but unfortunately I can't quite remember if there was a specific situation this omitted. It's possible that I had moved on because it still required the abortController to be handled somewhere else and I just didn't get back to it and consider this particular setup involving "should" flags.
We should do some QA testing but in theory I think this works.

packages/headless/src/features/generated-answer/generated-answer-state.ts
```diff
@@ -54,10 +58,23 @@ export interface GeneratedAnswerState {
    * A list of indexed fields to include in the citations returned with the generated answer.
    */
   fieldsToIncludeInCitations: string[];
+  /**
+   * Whether the current answer stream should be aborted.
+   */
+  shouldAbortStream: boolean;
```
nathanlb, line 64: We have a few state flags here and I'm wondering if we should group them together into one attribute with multiple states.
What we have:
- isLoading
- isStreaming
- shouldStartStream
- shouldAbortStream
ex:
```ts
    streamStatus: 'LOADING' | 'STREAM_PENDING' | 'STREAMING' | 'ABORT' | 'INACTIVE';
```

More of a state machine type of setup.
Just throwing this out there for your consideration.
I think it cleans up the state and potentially makes things simpler as we can't have weird combinations of boolean flags born of implementation errors.

nathanlb: Ah crap this is breaking though. Maybe we can't change all of them yet but maybe for v3?

PR #3561, opened 2024-01-30
chore(headless): burns the barrels

louis-bompart:
https://flaming.codes/en/posts/barrel-files-in-javascript/#:~:text=In%20JavaScript%20(or%20TypeScript)%2C,using%20a%20single%20import%20statement.

So, we have an issue where it's quite easy to import stuff from the wrong place because of barrels import, which makes it difficult to ensure that the points only bundle what's needed (see bundle size reporter as proof).

So, let's be radical, let's burn all barrels, and force the imports at the roots ;)

https://coveord.atlassian.net/browse/KIT-2977

This is one of maybe multiple PR that'll come to untangle some mess.


packages/atomic/cypress/e2e/search-interface.cypress.ts
```diff
@@ -90,6 +90,7 @@ describe('Search Interface Component', () => {
               searchHub: options.engine?.searchHub,
             },
           },
+          middlewares,
```
louis-bompart, line 93:
```suggestion
```

packages/headless/.eslintrc.js
```diff
@@ -0,0 +1,30 @@
...
+module.exports = {
+  ignorePatterns: ['dist', 'temp', 'ponyfills'],
+  plugins: ['canonical'],
+  settings: {
+    'import/parsers': {
+      '@typescript-eslint/parser': ['.ts', '.tsx'],
+    },
+    'import/resolver': {
+      typescript: {
+        project: resolve(__dirname, 'tsconfig.json'),
```
louis-bompart, line 12: To ensures proper resolve, needs to go js

olamothe (approved):

fbeaudoincoveo (approved):

PR #3563, opened 2024-01-30
refactor(headless): refactor generated-answer controller to remove abort logic

olamothe:
Refactor proposition for generated answer controller. 

* Tries to remove pretty much all logic about aborting request from the controller, which is problematic when there are multiple controller all trying to cancel the same request. It is now only handled inside `generated-answer-client`, which there is only one by engine, and the client instances are not shared between engine.

* Remove all logic about requesting a generated answer in the controller, which is problematic when there are multiple controller trying to request the same answer. It is now only handled in `search-action-thunk-processor` (ie: the file in charge of processing that needs to happen when we receive a search response, such as handling errors, or applying query correction).

* Exports missing generative answering redux actions + add some JSDoc which was missing.

https://coveord.atlassian.net/browse/KIT-2978


packages/headless/src/features/generated-answer/generated-answer-slice.ts
```diff
@@ -94,4 +96,9 @@ export const generatedAnswerReducer = createReducer(
           ...new Set(state.fieldsToIncludeInCitations.concat(action.payload)),
         ];
       })
+      .addCase(executeSearch.pending, (state) => handleResetAnswer(state))
+      .addCase(streamAnswer.pending, (state) => {
```
fbeaudoincoveo, line 100: I think we should test those new cases in generated-answer-slice.test.ts

packages/headless/src/features/search/legacy/search-actions-thunk-processor.test.ts
```diff
@@ -248,4 +251,45 @@ describe('AsyncSearchThunkProcessor', () => {
...
+      duration: 123,
+      queryExecuted: 'foo',
+      requestExecuted: buildMockSearchRequest(),
+    };
+
+    await processor.process(fetched);
+    expect(streamAnswer).not.toHaveBeenCalled();
+  });
+
+  it('does call streaming API if there is a generative answering id', async () => {
```
fbeaudoincoveo, line 275:
```suggestion
  it('calls streaming API if there is a generative answering id', async () => {
```

packages/headless/src/features/search/legacy/search-actions-thunk-processor.test.ts
```diff
@@ -248,4 +251,45 @@ describe('AsyncSearchThunkProcessor', () => {
...
+      response: {
+        success: searchResponse,
+      },
+      duration: 123,
+      queryExecuted: 'foo',
+      requestExecuted: buildMockSearchRequest(),
+    };
+
+    await processor.process(fetched);
+    expect(streamAnswer).toHaveBeenCalledWith({streamId: '123'});
```
fbeaudoincoveo, line 293: Since you're not calling `jest.clearAllMocks()` between each test, the order of execution of those two tests is important. Not necessarily a big deal right now, but it could cause problems later on if we ever need to add / move tests that rely on this mock.

packages/headless/src/features/search/search-actions-thunk-processor.ts
```diff
@@ -295,6 +297,22 @@ export class AsyncSearchThunkProcessor<RejectionType> {
     };
   }
 
+  private processGenerativeAnswer(
+    fetched: FetchedResponse
+  ): ValidReturnTypeFromProcessingStep<RejectionType> | null {
```
fbeaudoincoveo, line 302: From what I can see this method always returns null

fbeaudoincoveo (approved):

packages/headless/src/features/generated-answer/generated-answer-slice.ts
```diff
@@ -94,4 +96,9 @@ export const generatedAnswerReducer = createReducer(
...
+        state.isLoading = true;
+        state.isStreaming = false;
+      })
```
mmitiche, line 103: I think we need a new case here for when the stream answer is completed that sets `isLoading` to `false`.

For now we are reseting `isLoading` to false only  in the cases `updateMessage`, `updateCitations` and `updateError`.
If for example streamId is returned but no streaming is received, `isLoading`  will remain `true` causing an issue.

```suggestion
      .addCase(streamAnswer.pending, (state) => {
        state.isLoading = true;
        state.isStreaming = false;
      })
      .addCase(streamAnswer.fulfilled, (state) => {
        state.isLoading = false;
      })
```

mmitiche (commented):
Very neat solution indeed 👍

 The thing I see missing is a similar logic that initiates the streaming  from the [search-actions-thunk-processor.ts](https://github.com/coveo/ui-kit/pull/3563/files#diff-88b8a775f2b7149fccfa8805b686516ed2ca70dec78dc2170cee272319d25acf) but for the insight use case.

Cause the mechanism of sending search queries is a bit different between these two use cases, for insight we would probably initiate the streaming from [insight-search-actions.ts](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/features/insight-search/insight-search-actions.ts)

packages/headless/src/features/generated-answer/generated-answer-slice.ts
olamothe: Unfortunately cannot do this, as `streamAnswer` is not truly "async": it exit/return in a synchronous manner when you dispatch it, and actually writes/stream the answer by dispatching other actions.

So, net net, is that the `loading` state would just become un-necessary from a UX standpoint (you'd never see the loading animation).

I think all scenarios are covered with `write/abort/close` on the stream, and for each of those the loading state is properly handled with a dispatch.

PR #3566, opened 2024-01-31
feat(headless,atomic): expose and leverage user defined label in administration tool for facets

olamothe:
As the ticket explains: Administration tool allows you to define labels for facetable field. Which we can leverage in Headless and Atomic.

https://coveord.atlassian.net/browse/KIT-2639

### DIFF
--- a/packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts
+++ b/packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts
@@ -209,6 +209,10 @@ export interface CoreFacetState {
 
   /** Whether the facet is enabled and its values are used to filter search results. */
   enabled: boolean;
+  /**
+   * The name to display if this field is used by the Facet Generator in your interface.
+   */
+  label?: string;
 }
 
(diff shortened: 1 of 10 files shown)


louis-bompart (approved):

packages/headless/src/controllers/core/facets/facet/headless-core-facet.ts
```diff
@@ -209,6 +209,10 @@ export interface CoreFacetState {
 
   /** Whether the facet is enabled and its values are used to filter search results. */
   enabled: boolean;
+  /**
+   * The name to display if this field is used by the Facet Generator in your interface.
```
fbeaudoincoveo, line 213:
```suggestion
   * The name to display if this field is used by the Facet Generator in your interface. See [Change Facet Generator options](https://docs.coveo.com/en/n9sd0159#change-facet-generator-options).
```

Cause I don't think it's gonna be super obvious to most people where this label would be coming from and how they can configure it 😅

fbeaudoincoveo (approved):

PR #3569, opened 2024-02-01
feat(quantic): added support to entering line breaks in expandable search box

mmitiche:
[SFINT-5289](https://coveord.atlassian.net/browse/SFINT-5289)

- Added the ability to enter line breaks in the expandable search box.

### Demo:

https://github.com/coveo/ui-kit/assets/86681870/719a3175-62d8-43c9-af0e-c2033d63368e

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
@@ -211,7 +211,7 @@ export default class QuanticSearchBoxInput extends LightningElement {
    * @param {KeyboardEvent} event
    */
   onKeydown(event) {
-    if (event.key === keys.ENTER) {
+    if (event.key === keys.ENTER && !event.shiftKey) {
       event.preventDefault();
     }
   }
@@ -222,7 +222,9 @@ export default class QuanticSearchBoxInput extends LightningElement {
   onKeyup(event) {
     switch (event.key) {
       case keys.ENTER:
-        this.handleEnter();
+        if (!event.shiftKey) {
+          this.handleEnter();
+        }
         break;
       case keys.ARROWUP:
         this.suggestionListElement?.selectionUp();
@@ -254,10 +256,11 @@ export default class QuanticSearchBoxInput extends LightningElement {
     if (!this.textarea) {
       return;
     }
-    this.input.value = this.input.value.replace(/\n/g, '');
+    // this.input.value = this.input.value.replace(/\n/g, '');
     this.input.style.height = '';
     this.input.style.whiteSpace = 'pre-wrap';
     this.input.style.height = this.input.scrollHeight + 'px';
+    this.input.style.overflow = 'auto';
   }
 
   collapseTextArea() {
@@ -265,7 +268,9 @@ export default class QuanticSearchBoxInput extends LightningElement {
       return;
     }
     this.input.style.height = '';
+    this.input.scrollTop = 0;
     this.input.style.whiteSpace = 'nowrap';
+    this.input.style.overflow = 'hidden';
   }
 
   clearInput() {
@@ -291,6 +296,7 @@ export default class QuanticSearchBoxInput extends LightningElement {
 
   handleHighlightChange(event) {
     this.input.value = event.detail?.rawValue;
+    this.adjustTextAreaHeight();
   }
 
   handleSuggestionSelection(event) {
--- a/packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
@@ -256,7 +256,6 @@ export default class QuanticSearchBoxInput extends LightningElement {
     if (!this.textarea) {
       return;
     }
-    // this.input.value = this.input.value.replace(/\n/g, '');
     this.input.style.height = '';
     this.input.style.whiteSpace = 'pre-wrap';
     this.input.style.height = this.input.scrollHeight + 'px';


packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -353,6 +353,33 @@ describe('c-quantic-search-box-input', () => {
               functionsMocks.exampleHandleSubmitSearch
             ).toHaveBeenCalledTimes(1);
           });
+
+          it(`should ${
+            textareaValue ? 'not ' : ''
+          }dispatch a #quantic__submitsearch custom event when the shift key is pressed`, async () => {
```
SimonMilord, line 359: wouldn't it be more like `shift+enter key is pressed`?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -179,15 +179,18 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
-  handleEnter() {
-    const selectedSuggestion =
-      this.suggestionListElement?.getCurrentSelectedValue();
-    if (this.areSuggestionsOpen && selectedSuggestion) {
-      this.sendSelectSuggestionEvent(selectedSuggestion.rawValue);
-    } else {
-      this.sendSubmitSearchEvent();
+  handleEnter(event) {
+    const isLineBreak = this.textarea && event.shiftKey;
+    if (!isLineBreak) {
```
SimonMilord, line 184: Small question, do we want to be able to trigger the submit if we somehow press `shift+enter` to select a suggestion? My understanding is that this fix would prevent that and nothing would happen in that case?

SimonMilord (approved):
A small question regarding desired behaviour here, but otherwise looks fine to me

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
mmitiche: No not really, `shift+enter` should not trigger the selection of a suggestions cause it's different from just clicking `enter`.

`shift+enter` is supposed to solely introduce a line break.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
mmitiche: Yes, pressing enter is mentioned the parent `describe` block of this test but we can make more simple yeah.

mmitiche, line 359:
```suggestion
          }dispatch a #quantic__submitsearch custom event when the shift key is pressed along with the enter key`, async () => {
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
SimonMilord: all good then!

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -246,26 +249,28 @@ export default class QuanticSearchBoxInput extends LightningElement {
   }
 
   onTextAreaInput() {
-    this.handleValueChange();
+    this.sendInputValueChangeEvent(this.input.value, true);
```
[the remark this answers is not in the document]
mmitiche: This has changed because  `handleValueChange` was doing: 
```
this.sendInputValueChangeEvent(this.input.value, false);
```
The second parameter is named `resetSelection`, when it's true the hover on the suggestion list items will be reset.

When the value of the input changes, we need to reset the suggestions and that's why I'm updating this second parameter to `true`, it was causing an issue before.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -353,6 +353,33 @@ describe('c-quantic-search-box-input', () => {
               functionsMocks.exampleHandleSubmitSearch
             ).toHaveBeenCalledTimes(1);
           });
+
+          it(`should ${
+            textareaValue ? 'not ' : ''
+          }dispatch a #quantic__submitsearch custom event when the shift key is pressed along with the enter key`, async () => {
```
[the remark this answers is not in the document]
mmitiche: sure!

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -246,26 +249,28 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
 
   adjustTextAreaHeight() {
     if (!this.textarea) {
       return;
     }
-    this.input.value = this.input.value.replace(/\n/g, '');
     this.input.style.height = '';
     this.input.style.whiteSpace = 'pre-wrap';
     this.input.style.height = this.input.scrollHeight + 'px';
+    this.input.style.overflow = 'auto';
```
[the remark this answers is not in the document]
mmitiche: Sure, actually `overflow = 'auto'` is the natural behaviour so it shouldn't change anything.
This is added just because I'm adding `this.input.style.overflow = 'hidden';` when collapsing the expandable search box.


https://github.com/coveo/ui-kit/assets/86681870/858365d0-19f4-4aab-9ab8-2d5602bae672

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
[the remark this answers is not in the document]
mmitiche: the thing is that `handleValueChange ` is an event listener in the default search box, so we can't really add a parameter to it.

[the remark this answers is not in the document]
mmitiche: Also did a small refactor by removing `handleKeyValues ` to avoid duplicating the same logic twice.

PR #3570, opened 2024-02-01
feat(analytics): migrate quickview to EP

louis-bompart:
https://coveord.atlassian.net/browse/KIT-2984

- Rescope tests relative to quickviews.
- Do the migration

Closes https://github.com/coveo/analytics-migration-project/issues/28


packages/headless/src/controllers/insight/quickview/headless-insight-quickview.test.ts
```diff
@@ -46,90 +57,44 @@ describe('Insight Quickview', () => {
...
+    const someHtmlRequestOptions: HtmlRequestOptions = {
+      uniqueId: 'some-id',
+      requestedOutputSize: 0,
+    };
 
-  [true, false].forEach((testValue) => {
-    it(`when the result #hasHtmlVersion is ${testValue} #state.resultHasPreview should be ${testValue}`, () => {
-      options.result = buildMockResult({hasHtmlVersion: testValue});
-      initQuickview();
+    mockedBuildCoreQuickview.mock.calls[0][2](state, someHtmlRequestOptions);
```
fbeaudoincoveo, line 81: At a glance, this is pretty cryptic to me and makes it dififcult to quickly grasp how the test actually works.

Looking at the Jest documentation and at the `buildCoreQuickview` function signature, I gather that we're calling the third argument of the first call made to the mocked `buildCoreQuickview`.

How about we do something like:

```suggestion
    const mockedBuildResultPreviewRequest = mockedBuildCoreQuickview.mock.calls[0][2];
    mockedBuildResultPreviewRequest(state, someHtmlRequestOptions);
```

in these kinds of tests?

This comments applies to all similar instances.

packages/headless/src/features/result-preview/result-preview-analytics-actions.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      engine = buildSearchEngine({
+        configuration: {
+          ...getSampleSearchEngineConfiguration(),
+        },
+      });
+    });
+
+    it('should call coveo.analytics.makeRecommendationOpen properly', async () => {
+      engine.dispatch(logDocumentQuickview(testResult));
+      await wait();
```
fbeaudoincoveo, line 50: Mmm.. why is this necessary?

fbeaudoincoveo (approved):

packages/headless/src/controllers/quickview/headless-quickview.test.ts
```diff
@@ -1,32 +1,43 @@
...
+    engine = buildMockSearchEngine(preloadedState);
+  }
+
   function initQuickview() {
     quickview = buildQuickview(engine, {options});
   }
 
   beforeEach(() => {
-    engine = buildMockSearchAppEngine();
+    initEngine();
```
fbeaudoincoveo, line 40: I think you should also call `jest.resetAllMocks();` here

packages/headless/src/features/result-preview/result-preview-analytics-actions.test.ts
louis-bompart: There are 'unbound' promises (i.e. out-of-reach promises) that we need to wait for.
This is a trick that allows us to flush the microtask queue to an extent.

packages/headless/src/controllers/insight/quickview/headless-insight-quickview.test.ts
louis-bompart: > Looking at the Jest documentation and at the buildCoreQuickview function signature, I gather that we're calling the third argument of the first call made to the mocked buildCoreQuickview.

Yes

I'll work something out based on your suggestions (I don't want to use mock when it's not a mock, and the 3rd param is not a mock)

packages/headless/src/features/result-preview/result-preview-analytics-actions.test.ts
fbeaudoincoveo: In the meantime, might be worth adding a little comment (maybe where we define the wait() function) to explain this cause it's not super obvious 😅

olamothe (approved):

PR #3572, opened 2024-02-02
fix(atomic): fix facet order in modal refine

y-lakhdar:
Make sure the order in modal respects facet order in the DOM.

I added some E2E to ensure this does not break in the future

KIT-2981

### DIFF
--- a/packages/atomic/cypress/e2e/refine-toggle.cypress.ts
+++ b/packages/atomic/cypress/e2e/refine-toggle.cypress.ts
@@ -149,25 +149,6 @@ describe('Refine Toggle Test Suites', () => {
         .should('have.length', allFacets.length);
     });
 
-    it('should display the facets in the correct order', () => {
-      const expectedFacetOrder = [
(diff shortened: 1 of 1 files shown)


fbeaudoincoveo (approved):

packages/atomic/cypress/e2e/refine-toggle.cypress.ts
```diff
@@ -111,6 +149,25 @@ describe('Refine Toggle Test Suites', () => {
         .should('have.length', allFacets.length);
     });
 
+    it('should display the facets in the correct order', () => {
```
fbeaudoincoveo, line 152: We are expecting the original order to be preserved even if the facets are wrapped in a facet manager?

olamothe: I think its more that when you resize the page from desktop to mobile, the order is kept the same.

olamothe (approved):

PR #3581, opened 2024-02-05
test: change query in the E2E test

y-lakhdar:
The E2E standalone search box test involves entering a keyword into the search box and anticipating a suggestion from the Search API. However, currently, there are no suggestions being provided for this particular keyword (“awards”).

To address this issue, I propose utilizing a query that is more likely to yield suggestions from the API.

KIT-2989


mmitiche (approved):

louis-bompart (approved):
how

PR #3582, opened 2024-02-06
fix(atomic): improve query suggestions selection

y-lakhdar:
## Issues Addressed
The search box suggestions had 2 issues:

#### Problem
Hovering on a Query Suggestion while pressing on Enter was applying the suggestion hovered by the mouse instead of submitting what was in the search box.
#### Solution
Only click on the active element when on the right panel (instant results). Otherwise, always submit what is inside the searchbox.

#### Problem
Multiple suggestions could be active at the same time when using both the keyboard (up/down) arrows and mouse.
#### Solution
Remove the `hover:bg-neutral-light`.

KIT-2941

### DIFF
--- a/packages/atomic/cypress/e2e/search-box/instant-results/search-box-instant-results.cypress.ts
+++ b/packages/atomic/cypress/e2e/search-box/instant-results/search-box-instant-results.cypress.ts
@@ -149,7 +149,7 @@ describe('Instant Results Test Suites', () => {
     );
     InstantResultsAssertions.assertResultIsSelected(0);
 
-    SearchBoxAssertions.assertSuggestionIsHighlighted(1);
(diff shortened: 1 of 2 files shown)


louis-bompart (approved):

fbeaudoincoveo (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -534,7 +534,10 @@ export class AtomicSearchBox {
   }
 
   private onSubmit() {
-    if (this.activeDescendantElement) {
+    if (
+      this.activeDescendantElement &&
```
fbeaudoincoveo, line 538: Should we test that behavior in our e2e tests?

PR #3584, opened 2024-02-07
refactor(atomic): refactor "PagerCommon" into distincts functional components

olamothe:
This is a showcase of the type of refactor I'd like to propose for the `Common` component pattern we have throughout the project. This is to pre-emptively prepare what we might need for atomic vs commerce CMHV2.

I chose pager because it's very simple, but the philosophy should stay the same.

Essentially, I want to avoid having too much deeply nested logic. A ⚠️ sign is when the common components receives reference to "complex" and complicated objects, which they then just end up passing down multiple levels down.

In more egregious case (like `FacetCommon`), these components are stateful, and the spaghetti 🍝  to understand which property is wired where becomes very hard to follow, and brittle/easy to break.

I'd like to "flatten" and explode everything into multiple simple block, which are then handled by the top level `render()` function used by each component.

For example, if we ever wanted to have the pager next button behave or look a bit differently for insight-panel, we'd have to add more `if/else` into the big `PagerCommon` code-block. And risk introducing regression for the pager used in the full search page.

With this pattern, we could instead just remove the generic `PagerNextButton` from the `render()` function of `atomic-insight-pager` and replace it with a precisely scoped one. And have pretty much 0 risk of negatively affected the full search page use case.

On the negative side, each top level `render()` function (ie: `atomic-pager` vs `atomic-insight-pager`) are more complex, and there's a more code duplication.

https://coveord.atlassian.net/browse/KIT-2992


packages/atomic/src/components/common/pager/pager-button.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const PagerPreviousButton: FunctionalComponent<
+  PagerNavigationButtonProps
+> = (props) => {
+  return (
+    <Button
+      {...props}
+      style="outline-primary"
+      part="previous-button"
+      class="p-1 min-w-[2.5rem] min-h-[2.5rem] flex justify-center items-center"
```
Spuffynism, line 24: In the functional components defined here, should we spread `props` after having declared the default props, to allow overriding them?
```suggestion
      style="outline-primary"
      part="previous-button"
      class="p-1 min-w-[2.5rem] min-h-[2.5rem] flex justify-center items-center"
      {...props}
```

Spuffynism (approved):
I like this approach! It's quite similar to a [compound components](https://www.smashingmagazine.com/2021/08/compound-components-react) approach ([mantine does this for modals](https://mantine.dev/core/modal/#compound-components))!

I don't know if this is possible with stencil, or if it's really appropriate in this case, but did you consider using a context to hold the state for the functional components? It would in theory make the declaration of pagers for insight/search/commerce simpler, as we wouldn't need to initialize the functional components with as many props, in the `render` function.

packages/atomic/src/components/common/pager/pager-button.tsx
```diff
@@ -0,0 +1,77 @@
+import {FunctionalComponent, h} from '@stencil/core';
```
fbeaudoincoveo, line 1: Since the file contains all pager button definitions, I'd rename it to "pager-buttons.tsx"

packages/atomic/src/components/common/pager/pager-button.tsx
```diff
@@ -0,0 +1,77 @@
...
+      style="outline-neutral"
+      checked={props.isSelected}
+      ariaCurrent={props.isSelected ? 'page' : 'false'}
+      class="btn-page focus-visible:bg-neutral-light p-1 min-w-[2.5rem] min-h-[2.5rem]"
+      part={`page-button${props.isSelected ? ' active-page-button' : ''}`}
+    ></RadioButton>
+  );
+};
+
+export const PagerPages: FunctionalComponent = (_, children) => {
```
fbeaudoincoveo, line 71: I'd rename this component to "PagerPageButtons"

packages/atomic/src/components/common/pager/pager-nav.tsx
```diff
@@ -0,0 +1,18 @@
+import {FunctionalComponent, h} from '@stencil/core';
```
fbeaudoincoveo, line 1: For the sake of consistency I'd rename the file to "pager-navigation.tsx"

fbeaudoincoveo (approved):
Lovely

packages/atomic/src/components/common/pager/pager-button.tsx
olamothe: Since `PagerNavigationButtonProps` is an interface that `Omit<ButtonProps, 'style' | 'part' | 'class'>`, there's no override that should enter the equation.

So putting the prop spread before or after should have no effect !

olamothe: And moreover, we don't want to override the style or class from upstream.

PR #3585, opened 2024-02-07
fix(atomic): ensure redirection on instant result click

y-lakhdar:
The redirection fails to occur based on the user's click location within the instant result.

This problem arises when the mouse click event target is 'atomic-result,' preventing the 'atomic-search-box-instant-results' from extracting the document URI.

KIT-2993


louis-bompart (changes requested):
I think the getLink should be agnostic of where the user click on a result.
I'll open a counter-proposal 🔜

fbeaudoincoveo (approved):

louis-bompart:
https://github.com/coveo/ui-kit/commit/6a8641a0864618598fc221d1c11423affefeb0a9

`getLink` is called when we declare the Node/Element of the `atomic-result` we want to 'bind' ourselves to.
So let's just grab a ref and use it instead of depending on what exact element caught the click initially.

louis-bompart (approved):

PR #3588, opened 2024-02-08
fix(quantic): fixed issue with timeframe facet not communicating properly with the refine toggle

mmitiche:
[SFINT-5367](https://coveord.atlassian.net/browse/SFINT-5367)

- When a timeframe range that doesn't return any results is selected, the expected behaviour is that the refine toggle should be enabled and should allow opening the refine modal to unselect the timeframe range, however a bug has been observed where the refine toggle button is being disabled in this situation.

- The problem has been identified in the communication between the timeframe facets and the refine toggle component. 

- The facets components communicate the information about whether they are visible or not to the Quantic Refine Toggle component via custom events. A logic in  Refine Toggle component then, based on this information, will enable/disable the refine toggle.

- The issue was that the Quantic Timeframe Facet wasn't communicating its visibility information to the Refine Toggle when the state of the timeframe ranges changes.  

### The issue:

https://github.com/coveo/ui-kit/assets/86681870/538e656a-daf2-4b93-bb8a-2203c9279b15

### The solution:

- The solution is to make the Quantic Timeframe facet communicate properly with the Refine Toggle component by sending the right events whenever it should be sent.

https://github.com/coveo/ui-kit/assets/86681870/5b7de803-e38e-484d-982a-dfc994136e19

[SFINT-5367]: https://coveord.atlassian.net/browse/SFINT-5367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
```diff
@@ -169,7 +169,7 @@ export default class QuanticRefineToggle extends LightningElement {
   }
 
   get isContentEmpty() {
-    return this.hideSort && !this.someFacetsRendered;
+    return !!this.hideSort && !this.someFacetsRendered;
```
erocheleau, line 172: hummmm?

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
```diff
@@ -322,7 +322,7 @@ export default class QuanticTimeframeFacet extends LightningElement {
   get hasActiveValues() {
     return (
       this.formattedValues.some((v) => v.selected) ||
-      this.dateFilterState?.range
+      !!this.dateFilterState?.range
```
erocheleau, line 325: You're going to have to explain this one lol

erocheleau (commented):
I have some questions lol

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
mmitiche: That's nothing really, just a better way to see the return value of this getter, and this is not the fix fo the issue.

Before this change, `hasActiveValues` is truthy when `this.dateFilterState?.range` is  truthy, but since this latter is not a boolean, it means that the return value  of this getter will not be a boolean, it can be truthy or falsy but not a boolean, in other words not a simple value `true` or `false`. 
Which is something that can be a bit misleading while debugging the code.

So instead, now I'm making the return value of this getter a proper boolean, it can only be `true` or `false` no truthy/falsy  confusion could happen.

`!!this.dateFilterState?.range`  returns a boolean based on whether `this.dateFilterState?.range` is truthy or falsy.
for example:
```
this.dateFilterState?.range = {start: '...', end: '...'}   ==>  (!!this.dateFilterState?.range) equals true
```

```
this.dateFilterState?.range = null  ==>  (!!this.dateFilterState?.range) equals false
```

for reference: https://developer.mozilla.org/en-US/docs/Glossary/Truthy

SimonMilord (approved):

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
```diff
@@ -467,6 +459,22 @@ export default class QuanticTimeframeFacet extends LightningElement {
...
+    const renderFacetEvent = new CustomEvent('renderfacet', {
+      detail: {
+        id: this.facetId ?? this.field,
+        shouldRenderFacet: this.showFacet,
+      },
+      bubbles: true,
+      composed: true,
+    });
+    this.dispatchEvent(renderFacetEvent);
   }
```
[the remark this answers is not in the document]
mmitiche: This will need to be done in the v3 with all the other events , cause it can be a breaking change.

packages/quantic/cypress/e2e/default-1/refine-toggle/refine-toggle.cypress.ts
```diff
@@ -26,6 +28,31 @@ const viewResultsLabel = (value: number) => {
...
+  response.body.totalCountFiltered = 0;
+  response.body.facets = [
+    {
+      facetId: 'date_input',
+      field: 'date',
+      moreValuesAvailable: false,
+      values: [
+        {
+          start: '2024/02/09@00:00:00',
+          end: '2024/02/10@23:59:59',
```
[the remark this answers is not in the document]
mmitiche: No not really, cause it's not really about the date used, it's about having a range that returns no results, and in this mock function I'm forcing the response to contain no results.

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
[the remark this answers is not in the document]
SimonMilord: ok ! 👌

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
erocheleau: Yes, but why was it working before without that :P

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
erocheleau: `hideSort` IS already a boolean no?

packages/quantic/force-app/main/default/lwc/quanticTimeframeFacet/quanticTimeframeFacet.js
mmitiche: It's not something that would break anything. 

It just that we were relying on truthy/falsy values  and now we are relying on true/false values just for more clarity but the logic is exactly the same it's totally equivalent.

packages/quantic/force-app/main/default/lwc/quanticRefineToggle/quanticRefineToggle.js
mmitiche: when it's not provided to the component it's actually `undefined`.

so I'm adding the `!!` to end up with an expression like `false && false` instead of `undefined && false` for example.

erocheleau (approved):

PR #3589, opened 2024-02-08
chore: use octokit for release branch update

louis-bompart:
I think the ref update of `release/v2` is failing because GH does not recognize the deploy key as our bot and that maketh tables flips.

### DIFF
--- a/utils/release/git-publish-all.mjs
+++ b/utils/release/git-publish-all.mjs
@@ -13,8 +13,6 @@ import {
   gitUpdateRef,
   gitPublishBranch,
   gitPull,
-  gitPush,
-  gitReset,
 } from '@coveo/semantic-monorepo-tools';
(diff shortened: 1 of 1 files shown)


olamothe (approved):
🤞 I guess this should be patched in `@coveo/semantic-monorepo-tools` if that works ?

louis-bompart:
> 🤞 I guess this should be patched in `@coveo/semantic-monorepo-tools` if that works ?

👎 No, semantic-monorepo-tools right now is trying to SDKify CLI based operations used for CI/CD operations, think `git`, `pnpm` and `npm`.
Here, we're shifting from `git` to `GitHub` to do the operation (because it will be properly signed off thanks to that 😄, famouslastwords), and octokit does a good job (IMO).

Did you have a specific idea of what should be added to semantic-monorepo-tools?

PR #3607, opened 2024-02-12
feat(IPX): add PageView in IPX actions history

dguerinCoveo:
Allows the tracking of pageViews in the actions history for IPX. 

image

My first plan was to use coveo.analytics package to modify the actions.history array in the web local storage directly from the atomic component. However, coveo.analytics is imported in browser mode in the atomic package when I install it, and this mode [does not contain the history utils](https://github.com/coveo/coveo.analytics.js/blob/master/src/coveoua/browser.ts#L29). 

Therefore, I need to have a headless action to update the actions history array in local storage, so here we are.


packages/headless/src/features/analytics/generic-analytics-actions-loader.ts
```diff
@@ -44,6 +46,15 @@ export interface GenericAnalyticsActionCreators {
...
   logCustomEvent(payload: LogCustomEventActionCreatorPayload): CustomAction;
+
+  /**
+   * Adds an entry in the action history.
+   * @param itemPermanentId - The unique identifier of the item.
+   * @returns A dispatchable action.
+   */
+  addPageViewEntryInActionsHistory(
+    itemPermanentId: string
+  ): AsyncThunkAction<void, string, {}>;
```
dguerinCoveo, line 57: How do you feel about adding that there @louis-bompart ? Is there another place which would make more sense?

packages/headless/src/index.ts
```diff
@@ -550,6 +550,7 @@ export {buildResultTemplatesManager} from './features/result-templates/result-te
 export * from './features/analytics/search-analytics-actions-loader';
 export * from './features/analytics/click-analytics-actions-loader';
 export * from './features/analytics/generic-analytics-actions-loader';
+export * from './features/actions-history/ipx-actions-history-actions-loader';
```
dguerinCoveo, line 553: I feel like this goes here and not in the region grouped section, but I'm guessing more than anything

olamothe: The problem is the IPX naming: There's no "IPX" package in Headless like there is for commerce, case assist, insight panel. It's confusing.

I would export it as part of classic/generic analytics functionality.

ie: 
* Delete `features/actions-history/ipx-actions-history-actions-loader.ts`
* Delete `features/actions-history/ipx-actions-history-actions.ts`
* Add the action in `features/analytics/analytics-action.ts`
* Export it in `features/analytics/generic-analytics-action-loader.ts`

packages/headless/src/features/actions-history/ipx-actions-history-actions.ts
```diff
@@ -0,0 +1,17 @@
...
+    const state = getState() as SearchAppState;
+    if (state.configuration.analytics.enabled) {
+      historyStore.addElement({
+        name: 'PageView',
+        value: itemPermanentId,
+        time: JSON.stringify(new Date()),
+      });
+    }
+  }
+);
```
olamothe, line 17: Coupled with with other comment below, I would make this just a generic function for all entry in action history, not something just specific for pageview VS IPX.

```ts
export const addEntryInActionsHistory = createAsyncThunk(
  'analytics/actionsHistory/add',
  async (entry: HistoryElement, {getState}) => {
    const state = getState() as SearchAppState;
    if (state.configuration.analytics.enabled) {
      historyStore.addElement(entry);
    }
  }
);
```

packages/atomic/src/components/ipx/atomic-ipx-result-link/atomic-ipx-result-link.tsx
```diff
@@ -0,0 +1,119 @@
...
+  public connectedCallback() {
+    const slotName = 'attributes';
+    this.hasDefaultSlot = !!getDefaultSlotFromHost(this.host);
+    this.linkAttributes = getAttributesFromLinkSlot(this.host, slotName);
+  }
+
+  public async onSelect() {
+    const resultPermanentId = this.result.raw.permanentid;
+    if (resultPermanentId) {
+      const action = loadIPXActionsHistoryActions(
```
olamothe, line 80: I would load the action on `initialize`.

olamothe (changes requested):
It seems we don't have any IPX package in Headless -- Which is a bummer.

packages/atomic/src/components/ipx/atomic-ipx-result-link/atomic-ipx-result-link.tsx
dguerinCoveo: good point

packages/atomic/src/components.d.ts
```diff
@@ -5229,6 +5248,15 @@ declare namespace LocalJSX {
...
         "collapseFacetsAfter"?: number;
     }
+    /**
+     * The `atomic-ipx-result-link` component automatically transforms a search result title into a clickable link that points to the original item. It is an experimental internal component not intended for general use.
+     */
+    interface AtomicIpxResultLink {
+        /**
+          * Specifies a template literal from which to generate the `href` attribute value (see [Template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals)).  The template literal can reference any number of result properties from the parent result. It can also reference the window object.  For example, the following markup generates an `href` value such as `http://uri.com?id=itemTitle`, using the result's `clickUri` and `itemtitle` fields. ```html <atomic-ipx-result-link href-template='${clickUri}?id=${raw.itemtitle}'></atomic-ipx-result-link> ```
+         */
+        "hrefTemplate"?: string;
```
mikegron, line 5258: tu a combiné les modifs pour le href template?

packages/atomic/src/components/ipx/atomic-ipx-result-link/atomic-ipx-result-link.tsx
```diff
@@ -0,0 +1,127 @@
...
+ * @slot default - Lets you display alternative content inside the link
+ * @slot attributes - Lets you pass [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes) down to the link element, overriding other attributes, to be used exclusively with an "a" tag such as `<a slot="attributes" target="_blank" download></a>`.
+ * @internal
+ */
+@Component({
+  tag: 'atomic-ipx-result-link',
+  styleUrl: 'atomic-ipx-result-link.pcss',
+  shadow: false,
+})
+export class AtomicIPXResultLink implements InitializableComponent {
```
mikegron, line 37: Ont peux tu hérité de result link direct?

mikegron (approved):

packages/atomic/src/components.d.ts
dguerinCoveo: Non, AtomicResultLink definit deja une prop pour hrefTemplate. C'est juste du copy pasta

packages/atomic/src/components/ipx/atomic-ipx-result-link/atomic-ipx-result-link.tsx
louis-bompart: No Inheritance for components is our rule of thumb. Do prefer composition :)

dguerinCoveo:
@olamothe, @louis-bompart gentle reminder to review the PR please :)

louis-bompart (changes requested):
I have to concur with Olivier, please create an IPX package to contain and isolate the code unique to IPX.

dguerinCoveo:
> I have to concur with Olivier, please create an IPX package to contain and isolate the code unique to IPX.

In headless you mean? I think Olivier's point was the opposite. When he reviewed, I had the headless actions inside IPX specific packages. I moved my headless code to the generic package following his comment.

olamothe (approved):

olamothe:
> I have to concur with Olivier, please create an IPX package to contain and isolate the code unique to IPX.

For Headless specifically (and for the analytics actions added in this PR), I don't see the need to create an IPX package, or IPX specific files. Pushing entries in actions-history (for legacy events) is a "generic" event that can be used by anyone, even for classic search pages.

louis-bompart (changes requested):
After discussion with @olamothe:
 - Indeed, an IPX entrypoint at this stage is overkill, let's not.
 - Instead, let's go back to the solution originally proposed:
   - Dedicated action-loader for IPX
 The rationale is that PageView in ActionHistory is an out-of-standard feature, so we do not want that to be included by our users (and we won't support it either).
 Moreover, all this will be wiped in V3.
 
 TL;DR: `git reset --hard HEAD~1` and it should be alright.

dguerinCoveo:
Changed the package to the reached consensus. I've re-asked only for your approval @louis-bompart , but tell me if we need oli's eyes on that 🙏

louis-bompart (approved):

pathiery (approved):

PR #3615, opened 2024-02-14
fix(Quantic): Should not return placeholder instead of field value

erocheleau:
SFINT-5383

If the field was missing from the result, we were displaying the placeholder instead of the value.

This is not the desired behavior. This would mean we were sending stuff like `${Excerpt}` in the body of the Post To Feed result action for example.

Also added some tests for the `buildTemplateTextFromResult` quanticUtils method.

### DIFF
--- a/packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
+++ b/packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
@@ -1,4 +1,8 @@
-import {I18nUtils, copyToClipboard} from 'c/quanticUtils';
+import {
+  I18nUtils,
+  buildTemplateTextFromResult,
+  copyToClipboard,
+} from 'c/quanticUtils';
 
 describe('c/quanticUtils', () => {
   describe('I18nUtils', () => {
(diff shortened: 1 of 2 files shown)


packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
```diff
@@ -160,4 +164,40 @@ describe('c/quanticUtils', () => {
...
+      expect(buildTemplateTextFromResult(undefined, testResult)).toBe('');
+    });
+
+    it('should not return the template field name as placeholder if the result does not contain the field', () => {
+      const resultWithoutExcerpt = {
+        clickUri: 'https://www.example.com',
+        title: 'Example Title',
+      };
+
+      // @ts-ignore
```
mmitiche, line 195: why do we need the `@ts-ignore`

mmitiche (approved):
LGTM!

packages/quantic/force-app/main/default/lwc/quanticUtils/__tests__/quanticUtils.test.js
erocheleau: Because the `buildTemplateTextFromResult` is expecting a full `coveo.Result` type complete with all the required fields. I didn't want to spend the time to make a full Result mock for this unit test.

If we already have one somewhere I will gladly use it instead :)

erocheleau: All that's really relevant for the test are the fields in the object I pass to the utils method. I'm wondering if the method shouldn't be changed to just expect an Object instead of a `coveo.Result`

SimonMilord (approved):
👌

PR #3619, opened 2024-02-16
fix(refine-modal): Sort facets values based on their DOM position

louis-bompart:
> [!WARNING]
> This PR uses esoteric JavaScript APIs.
>Caffeine consumption/injection is advised before attempting a review.

## What went wrong?
During #3572, we changed the clone node algorithm:
https://github.com/coveo/ui-kit/blob/12140a1d91ec018a1bae05a94d87b42d70378452/packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx#L168-L190

This algorithm fails as it queries the DOM again for the facet, but only considering the tags of the components, and not taking into account the prefiltering that happened before in the execution stack:
https://github.com/coveo/ui-kit/blob/12140a1d91ec018a1bae05a94d87b42d70378452/packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx#L143-L149

This causes facets to be duplicated when some facets are invisible, such as when a  `depends-on` condition is unmet.

> [!NOTE]
> Decision 1: Revert the clone algorithm to the previous one.

## What really caused the issue of #3572?
This line:
https://github.com/coveo/ui-kit/blob/12140a1d91ec018a1bae05a94d87b42d70378452/packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx#L141
We sort facets using a facet manager, regardless of if the facets were originally managed, which leads to issue when some facets (or all) are not managed.

This leads of course to

> [!NOTE]
> Decision 2: the refine modal must not use a `FacetManager`

## Proposed solution

This is where the _fun_ begins. The gig is to 'sort facets' in the same order as they appear to the user.
Let's assume the HTML DOM tree does represent the order in which the information will be displayed.
We can then use [`compareDocumentPosition`](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition) to figure out if an Element precedes or follows another Element.
This should solve all our problems (said no one ever)

> [!NOTE]
> Decision 3: We'll use the DOM tree as the source of truth for facet order.

However, we know that most of the time facets are put in `atomic-section`, which morphs where DOM goes, so we have to make some opinionated decisions:

> [!NOTE]
> Decision 4: Facets first criterion of sorting is "Which section is your parent". The priority order is as such:
> 1. `facets` section
> 2. `horizontal-facet` section
> 3. all the other facets
>
> Decision 3 criterion is still relevant as the second sort criterion

To reduce the complexity of the algorithm to use to sort all facets mixed up together, I decided to first split them up into the 3 categories defined in Decision 3, and then sort them out in each of their set.

### DIFF
--- a/packages/atomic/src/components/common/facets/facet-common.tsx
+++ b/packages/atomic/src/components/common/facets/facet-common.tsx
@@ -313,6 +313,39 @@ export function getAutomaticFacetGenerator(
   );
 }
 
+/**
+ * Triage elements by their parents.
+ * @param facets Facet Elements
+ * @param parents Elements that may contains the facets
+ * @returns an array in the same order as the parents, containing the facets that are contained by the corresponding parent.
+ * The last element of the array contains the facets that are not contained by any of the parents.
+ */
+export function triageFacetByParents(
+  facets: BaseFacetElement[],
+  ...parents: (HTMLElement | null)[]
+) {
+  const sortedFacets: BaseFacetElement[][] = new Array(parents.length + 1)
+    .fill(null)
+    .map(() => []);
+  facets: for (const facet of facets) {
+    parents: for (
+      let parentIndex = 0;
+      parentIndex < parents.length;
+      parentIndex++
+    ) {
+      const parent = parents[parentIndex];
+      if (parent === null) {
+        continue parents;
+      } else if (parent.contains(facet)) {
+        sortedFacets[parentIndex].push(facet);
+        continue facets;
+      }
+    }
+    sortedFacets[parents.length].push(facet);
+  }
+  return sortedFacets;
+}
+
 export function sortFacetsUsingManager(
   facets: BaseFacetElement[],
   facetManager: FacetManager
--- a/packages/atomic/src/components/common/facets/facet-common.tsx
+++ b/packages/atomic/src/components/common/facets/facet-common.tsx
@@ -313,6 +313,21 @@ export function getAutomaticFacetGenerator(
   );
 }
 
+const get2DMatrix = (xSize: number, ySize: number = 0) =>
+  new Array(xSize).fill(null).map(() => new Array(ySize));
+
+function findIndiceOfParent(
+  facet: BaseFacetElement,
+  parents: (HTMLElement | null)[]
+) {
+  for (let i = 0; i < parents.length; i++) {
+    if (parents[i]?.contains(facet)) {
+      return i;
+    }
+  }
+  return parents.length;
+}
+
 /**
  * Triage elements by their parents.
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -166,27 +189,16 @@ export class AtomicRefineModal implements InitializableComponent {
...
         this.collapseFacetsAfter === -1
           ? false
-          : index + 1 > this.collapseFacetsAfter;
-      clones.push(clone);
+          : i + 1 > this.collapseFacetsAfter;
+      return clone;
     });
-
-    return clones;
   }
```
louis-bompart, line 202: Revert to the old clone algo

packages/atomic/src/components/search/atomic-refine-modal/atomic-refine-modal.tsx
```diff
@@ -137,8 +138,30 @@ export class AtomicRefineModal implements InitializableComponent {
     this.addFacetColumnStyling(divSlot);
 
     const facets = this.bindings.store.getFacetElements() as BaseFacetElement[];
-
-    const sortedFacets = sortFacetsUsingManager(facets, this.facetManager);
+    const atomicSearchInterface = this.host.closest('atomic-search-interface')!;
```
louis-bompart, line 141: We don't want to explore the whole document, because we don't know if there are other interfaces in the page.
We limit the facet triage to the current interface.

fbeaudoincoveo (approved):

packages/atomic/src/components/common/facets/facet-common.tsx
```diff
@@ -313,6 +313,39 @@ export function getAutomaticFacetGenerator(
...
+ * @returns an array in the same order as the parents, containing the facets that are contained by the corresponding parent.
+ * The last element of the array contains the facets that are not contained by any of the parents.
+ */
+export function triageFacetByParents(
+  facets: BaseFacetElement[],
+  ...parents: (HTMLElement | null)[]
+) {
+  const sortedFacets: BaseFacetElement[][] = new Array(parents.length + 1)
+    .fill(null)
+    .map(() => []);
```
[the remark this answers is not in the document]
louis-bompart: That was my first attempt 😄 
The issue is that fill will use the same array for each element
```js
const myArrayOfArrays = (new Array(2)).fill([]);
myArrayOfArrays[0].push('hi');
console.log(myArrayOfArrays[1].length) // logs 1 but should be 0
```

PR #3626, opened 2024-02-19
fix(quantic): adding support to opening case comments in Salesforce sub tabs

mmitiche:
# [SFINT-5343](https://coveord.atlassian.net/browse/SFINT-5343)

## The issue:

In the Hosted Insight Panel, when you have a Case Comment indexed and you  try to open this Case Comment  result, we try to open a primary tab with the Case Comment as the record which does not work in Salesforce and gives the following error:
 

https://github.com/coveo/ui-kit/assets/86681870/68f43654-40ee-4c10-988d-167c400e555e

## The solution:

Instead of opening the Case Comment, we adapted our logic in Quantic to open the parent of that Case Comment which is the Case it self. 
This will match the behaviour we had with the standard insight panel.

In order to do this, I needed to added the `sfparentid` to the list of fields to include by default when we fetch results
Having this field will allow us to open the parent record of the Case Comment.
I also updated the logic of the Quantic Result Link component in order to support the case where the parent of the record is the one that needs to be opened:

https://github.com/coveo/ui-kit/assets/86681870/f9c93d99-8451-4f93-be65-dbcbaef05c21

[SFINT-5343]: https://coveord.atlassian.net/browse/SFINT-5343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -8,6 +8,11 @@ import {ResultUtils} from 'c/quanticUtils';
 import {NavigationMixin} from 'lightning/navigation';
 import {LightningElement, api} from 'lwc';
 
+/**
+ * Some document types cannot be opened directly in Salesforce, but we need to open their parent record, as is the case for the Case Comment document type.
+ */
+const documentTypesRequiringParentRecord = ['CaseComment'];
```
erocheleau, line 14: Have you tried if it's possible to have the comment of a comment?

Aka try to figure out in which situation this is **not** actually working.

I don't exactly know anymore, but there's two types of objects that can be considered "CaseComments" too, the first one is an old feature called Case Comment, and the 2nd one are what we call specific FeedItems records where the parent is a Case.

But there's also FeedComment where the parent can be a FeedItem on a case, etc.

all in all make sure it works with grand-children too etc.

erocheleau (approved):
Just make sure to test to see if you can actually break that.

SimonMilord (approved):
Looks good to me, just not 100% sure if there are other document types requiring the parent record as etienne brought it up

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
mmitiche: The logic I added was meant to be easily scalable to adapt to any other similar situation/issue that we would face. 
But for now that's the only case where I managed to reproduce the issue with.

PR #3628, opened 2024-02-22
feat(headless-commerce): use getVisitorID to set clientId

sallainCoveo:
Jira: https://coveord.atlassian.net/browse/LENS-1525

First part to potentially improve attribution. Previously, the `clientId` was set in the state, using an action to update the commerceContext. This branch is to instead fetch the `clientId` for the commerce API's request. To do so, I'm using the `getVisitorID` function from `coveoua`. Although we want to focus on `Relay` in the future, we will be using the stable functionality used across the package for now. Beside, the `clientId` value should be the same for both analytics' package.

To test it, I've linked the local build to barca sport to test this change. The first part of the following demo is from the current state of Barca sports in DEV, then the second part is what is done is in a local environment:

https://github.com/coveo/ui-kit/assets/58052881/9ef50d7e-e6c9-4794-aa22-612003443d9f

When it's merge, we can update Barca sports.


packages/headless/src/api/analytics/event-protocol-utils.ts
```diff
@@ -0,0 +1,18 @@
...
+  const relayOptions = {
+    url: state.analytics.nextApiBaseUrl,
+    token: state.accessToken,
+    trackingId: state.analytics.trackingId,
+    source: getAnalyticsSource(state.analytics),
+  };
+  const relay = createRelay(relayOptions);
+  const {getMeta} = relay;
+  return getMeta(type).clientId;
+};
```
samisayegh, line 18: I feel it would be better to use the `getVisitorID` function since it's used by the rest of the code base.
In the future we can switch that function to use relay, but for now, I feel having one way to retrieve the id is better than two,

sallainCoveo: It's a fair take (and I'm comfortable with making that change as it would be the same value regardless of the library used). 

On the flip side, if we know we will be using Relay on the commerce-engine, I think it would be nice to add it right now (especially as I'm working on emitting events with the controllers).

Spuffynism (approved):
Looks great!

packages/headless/src/api/analytics/event-protocol-utils.ts
samisayegh: I feel we need to look at this from the perspective of headless rather than the commerce engine. Headless has multiple engines, and server events are at the headless/api level. All engines need to retrieve the client id in a consistent way. 

Commerce engine was approaching this differently, and this led to a bug. While `relay` and `coveoua` _should_ behave the same regarding client id, there's still a risk that that's not the case. The easiest way to remove the risk is by using the same approach.

For the client events sent by headless, using relay makes sense. This is the case with the other engines too, which sent `itemClick` using relay.

Eventually we'll have relay powering everything, we just need to take small steps to get there, and keep all of headless consistent at all times.

sallainCoveo: You convinced me! I'll test it out with coveoa and if everything works well, will remove the getClientId function (and the `event-protocol-utils` file at the same time)

samisayegh (approved):
Looks great ✅

fbeaudoincoveo (approved):
💯

PR #3629, opened 2024-02-22
fix(quantic): Fix production release script

erocheleau:
https://coveord.atlassian.net/browse/KIT-3008

There was a confusion when we updated the env variables, the Jenkinsfile was not updated with the changes to the env variables.


louis-bompart (approved):
oops

mmitiche (approved):
👌

SimonMilord (approved):

PR #3632, opened 2024-02-23
spike(commerce): demonstrate controller-compatible routes usage

Spuffynism:
The central idea to this approach is that I'm extending the engine type to create a route type. This makes current controllers immediately route-compatible. This also means slices don't need to be route-aware.

Routes are registered at runtime, when we inject the necessary reducers for it.

[CAPI-496]

[CAPI-496]: https://coveord.atlassian.net/browse/CAPI-496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/app/reducer-manager.ts
+++ b/packages/headless/src/app/reducer-manager.ts
@@ -10,6 +10,7 @@ import {fromEntries} from '../utils/utils';
 export interface ReducerManager {
   combinedReducer: Reducer;
   add(newReducers: ReducersMapObject): void;
+  remove(oldReducers: ReducersMapObject): void;
   containsAll(newReducers: ReducersMapObject): boolean;
   addCrossReducer(reducer: Reducer): void;
 }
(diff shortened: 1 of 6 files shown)


packages/headless/src/app/reducer-manager.ts
```diff
@@ -56,6 +57,12 @@ export function createReducerManager(
         .forEach((key) => (reducers[key] = newReducers[key]));
     },
 
+    remove(oldReducers: ReducersMapObject) {
+      Object.keys(oldReducers)
+        .filter((key) => key in reducers)
+        .forEach((key) => (reducers[key] = <Reducer>(() => {})));
```
louis-bompart, line 63: - Double iterations :x:
- Consider a `for..of` loop over a `forEach` to reduce the stack size.

louis-bompart (commented):
So right now, what would Route do?

Spuffynism:
Closing this since we've decided on a different approach as part of https://coveord.atlassian.net/browse/CAPI-496

PR #3636, opened 2024-02-26
feat(quantic): logic to open links in salesforce refactored

mmitiche:
## [SFINT-5391](https://coveord.atlassian.net/browse/SFINT-5391)

- The logic of properly opening links directly inside Salesforce is now completely isolated in the `Quantic Result Link` component.

- Any component that want to benefit from this feature will just need to simply use the Quantic Result Link component exactly like the Quantic Result Printable Uri component does. 

- New unit tests added to the `Quantic Result Link` component.

#### Demo of the `Quantic Result Printable Uri` component:

https://github.com/coveo/ui-kit/assets/86681870/0d4d6f2e-529f-43a2-b9b0-9aa397fc40aa

[SFINT-5391]: https://coveord.atlassian.net/browse/SFINT-5391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -71,6 +71,11 @@ export default class QuanticResultLink extends NavigationMixin(
       focusTarget.focus();
     }
   }
+  /**
+   * @api
+   * @type {string}
+   */
```
mmitiche, line 77:
```suggestion
  /**
   * Indicates the result field to display as the link text.
   * @api
   * @type {string}
   * @defaultValue `'title'`
   */
```

packages/quantic/force-app/main/default/lwc/quanticResultPrintableUri/quanticResultPrintableUri.html
```diff
@@ -16,9 +16,7 @@
         </div>
       </template>
       <template lwc:if={shouldDisplayPrintableUriLink}>
-        <a class="slds-truncate" href={hrefValue} aria-label={ariaLabelValue} target={target} onclick={handleClick} tabindex="0" role="link" data-test="link">
-          <c-quantic-result-text result={result} field="printableUri"></c-quantic-result-text>
-        </a>
+        <c-quantic-result-link engine-id={engineId} result={result} target={target} displayed-field="printableUri"></c-quantic-result-link>
```
erocheleau, line 19: Didn't we have a case where instead of the clickUri the printableUri was supposed to display the "path" to the document for example in sub-folders?

packages/quantic/cypress/e2e/default-2/recommendation-list/recommendation-list-selectors.ts
```diff
@@ -30,7 +30,7 @@ export const RecommendationListSelectors: RecommendationListSelector &
     ),
   recommendationLinks: () =>
     RecommendationListSelectors.get().find(
-      '[data-cy="recommendations__item"] c-quantic-result-link a'
+      '[data-cy="recommendations__item"] [data-cy="result-link"] a'
```
erocheleau, line 33: Why exactly are we changing this?

packages/quantic/force-app/main/default/lwc/quanticResultPrintableUri/quanticResultPrintableUri.html
mmitiche: yeah we still have this case, it's just a bit higher in the code of this file
Screenshot 2024-02-26 at 2 39 41 PM

packages/quantic/cypress/e2e/default-2/recommendation-list/recommendation-list-selectors.ts
mmitiche: just because the selectors where not precise enough,  previously we  were selecting  `c-quantic-result-link`  and our goal was only to select the links that are used to display the title of the result.

However now, since `c-quantic-result-link` is now used by the `QuanticResultPrintableUri` it means that we were selecting some links that we weren't supposed to select.

So updated the selectors to be more precise, yielding the selection only of the result titles.

SimonMilord: Yeah according to the cypress documentation, `data-cy` seems like the best option

packages/quantic/force-app/main/default/lwc/quanticResultPrintableUri/quanticResultPrintableUri.html
erocheleau: Oh gotcha it's purely in the html I missed it.

packages/quantic/cypress/e2e/default-2/recommendation-list/recommendation-list-selectors.ts
erocheleau: makes sense 👍

erocheleau (approved):

PR #3637, opened 2024-02-26
feat(headless): insight analytics actions migrated to new event protocol

mmitiche:
## [SFINT-5397](https://coveord.atlassian.net/browse/SFINT-5397)

In this PR, the following insight analytics events have been migrated to the new event protocol:

- logCaseAttach.
- logCaseDetach.
- logCopyToClipboard.
- logCaseSendEmail.
- logFeedItemTextPost.
- logDocumentOpen.
- logExpandToFullUI.
- logCreateArticle.

These mentioned actions will, according to the value of `analyticsMode` in the engine state, log analytics either with `coveo.analytics.js` when `analyticsMode` is set to `legacy` or with the new `relay` library if `analyticsMode` is set to `next`.

- Unit tests have been added to cover this new logic.

- More insight events migration will come soon after this PR.

### Validity example:

Screenshot 2024-02-26 at 11 41 09 AM

### Demo in the Hosted Insight Panel:

https://github.com/coveo/ui-kit/assets/86681870/7f1f50fe-0e8d-4185-8f55-1b21ae285240

[SFINT-5397]: https://coveord.atlassian.net/browse/SFINT-5397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/analytics/insight-analytics-actions.test.ts
mmitiche, line 1: This file was moved under `src/features/insight-search` for a better organization.

packages/headless/src/features/analytics/insight-analytics-actions.ts
mmitiche, line 1: This file was moved under `src/features/insight-search` for a better organization.

packages/headless/src/features/insight-search/insight-analytics-actions.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+const exampleSubject = 'example subject';
+const exampleDescription = 'example description';
+const exampleCaseId = '1234';
+const exampleCaseNumber = '5678';
+const exampleCreateArticleMetadata = {
+  articleType: 'Knowledge__kav',
+};
+
+describe('result actions insight analytics actions', () => {
```
mmitiche, line 51:
```suggestion
describe('insight analytics actions', () => {
```

packages/headless/src/features/insight-search/insight-search-analytics-actions.test.ts
mmitiche, line 1: This file was only refactored for a better organization.

packages/headless/src/features/result/result-insight-analytics-actions.test.ts
```diff
@@ -42,45 +63,116 @@ const expectedDocumentInfo = {
...
+    urihash: 'example documentUriHash',
+    source: 'example sourceName',
+    collection: 'example collectionName',
+    permanentid: 'example contentIDValue',
+  }),
+};
+
+const testResult = buildMockResult(resultParams);
+
+describe('result actions insight analytics actions', () => {
```
mmitiche, line 95:
```suggestion
describe('result insight analytics actions', () => {
```

packages/headless/src/features/attached-results/attached-results-analytics-actions.test.ts
```diff
@@ -71,61 +91,117 @@ const resultParams = {
...
+      });
+    });
+
+    describe('logCaseAttach', () => {
+      it('should call relay.emit properly', async () => {
+        engine.dispatch(logCaseAttach(testResult));
+        await clearMicrotaskQueue();
+
+        expect(emit).toHaveBeenCalledTimes(1);
+        expect(emit.mock.calls[0]).toMatchSnapshot();
```
erocheleau, line 193: That's cool the ideas of snapshots!

packages/headless/src/features/attached-results/attached-results-analytics-actions.ts
```diff
@@ -8,20 +9,75 @@ import {
...
+      const information = partialDocumentInformation(result, state);
+      return {
+        itemMetadata: {
+          uniqueFieldName: identifier.contentIDKey,
+          uniqueFieldValue: identifier.contentIDValue,
+          title: information.documentTitle,
+          author: information.documentAuthor,
+          url: information.documentUri,
+        },
+        context: {
```
erocheleau, line 76: So we don't have the Case Subject?

packages/headless/src/features/insight-search/insight-analytics-actions.ts
```diff
@@ -0,0 +1,74 @@
...
+      };
+      return client.logExpandToFullUI(meta);
+    },
+    analyticsType: 'InsightPanel.ExpandToFullUI',
+    analyticsPayloadBuilder: (_state): InsightPanel.ExpandToFullUI => {
+      return {
+        context: {
+          targetId: caseId,
+          targetType: 'Case',
+          caseNumber: caseNumber,
```
erocheleau, line 40: Why are we getting this from the parameter and not from the state? Looks like this is also available in the state caseContext no?

packages/headless/src/features/insight-search/insight-analytics-actions.test.ts
```diff
@@ -0,0 +1,175 @@
...
+          },
+          caseId: exampleCaseId,
+          caseNumber: exampleCaseNumber,
+        };
+
+        expect(mockLogCreateArticle).toHaveBeenCalledTimes(1);
+        expect(mockLogCreateArticle.mock.calls[0][0]).toStrictEqual(
+          exampleCreateArticleMetadata
+        );
+        expect(mockLogCreateArticle.mock.calls[0][1]).toStrictEqual(
```
erocheleau, line 101: Didn't want to create snapshots for the old analytics?

packages/headless/src/features/insight-search/insight-search-analytics-actions.ts
```diff
@@ -41,19 +41,17 @@ export const logContextChanged = (
...
+    client.logInterfaceLoad(
+      getCaseContextAnalyticsMetadata(state.insightCaseContext)
+    )
+  );
+
+export const logInsightInterfaceChange = (): InsightAction =>
+  makeInsightAnalyticsAction('analytics/interface/change', (client, state) => {
+    client.logInterfaceChange({
+      ...getCaseContextAnalyticsMetadata(state.insightCaseContext),
+      interfaceChangeTo: state.configuration.analytics.originLevel2,
```
erocheleau, line 55: I'm assuming there's no breaking changes to any of this?

packages/headless/src/features/result-actions/result-actions-insight-analytics-actions.ts
```diff
@@ -21,13 +22,38 @@ export const logCopyToClipboard = (result: Result): InsightAction =>
...
+        },
+        position: information.documentPosition,
+        searchUid: state.search?.response.searchUid || '',
+        action: 'copyToClipboard',
+        context: {
+          targetId: metadata.caseId || '',
+          targetType: 'Case',
+          caseNumber: metadata.caseNumber,
+        } as InsightPanel.Context,
+      };
```
erocheleau, line 49: I think everything here could be placed into some function to generate 
`itemMetadata, position, searchUid` <-- in 1 method
`context` <-- in another method
And we would only be left setting the `action` in the specific InsightPanel.ItemAction.

What do you think?
I wouldn't mind having a `features/analytics/insight-analytics-utils.ts` file with some generic functions in it.

You have this code in

`result-insight-analytics-actions`
`result-actions-insight-analytics-actions`
`attached-results-analytics-actions.ts`

erocheleau (commented):
Some improvements I think we can make, but overall I like the new file structure too!

packages/headless/src/features/attached-results/attached-results-analytics-actions.ts
```diff
@@ -8,20 +9,75 @@ import {
...
+          targetId: metadata.caseId || '',
+          targetType: 'Case',
```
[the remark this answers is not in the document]
mmitiche: Indeed that's good point we will need to improve!
Do we currently have plans to support the HIP for other types of records?

packages/headless/src/features/insight-search/insight-analytics-actions.test.ts
mmitiche: No really I guess since starting from the v3 this code will be deleted.

packages/headless/src/features/insight-search/insight-search-analytics-actions.ts
mmitiche: Yeah just changed the where the actions is defined but everything is exported and exposed like it was before so no breaking changes, just a slightly better organization.

packages/headless/src/features/attached-results/attached-results-analytics-actions.ts
mmitiche: Exactly, InsightPanel.Context is supposed to have the `CaseSubject`.

However, currently we don't have this information in the state. 

That's what we have in the state:
```
{
    caseContext: {...}
    caseId: '...',
    caseNumber: '...',
  };
``` 

So only the `caseId` and the `caseNumber` are the the information that we sure have.

The case subject could be added in the `caseContext` object, but we have no way of knowing under which key it is.

Additionally we can't add all the `CaseContext` in the metadata of the event, cause any key  sent un the metadata and which is not defined in the schema will strictly make the event invalid.

Unless there is a very powerful reason behind why we need the subject here, I think we should just update the schema by removing the `CaseSubject`

erocheleau: @felixpelletier ?

packages/headless/src/features/analytics/analytics-utils.ts
```diff
@@ -242,7 +242,7 @@ export type AnalyticsActionOptions<
   LegacyProvider,
   Client,
   PayloadType,
-> = Exclude<
+> = Omit<
```
olamothe, line 245: What is this change for 🤔 ?

packages/headless/src/features/attached-results/attached-results-analytics-actions.test.ts
```diff
@@ -71,61 +91,117 @@ const resultParams = {
...
-    const engine = buildMockInsightEngine({
-      state: buildMockInsightState({
-        search: buildMockSearchState({
-          results: [testResult],
-        }),
-        insightCaseContext: {
-          caseContext: {
-            Case_Subject: exampleSubject,
-            Case_Description: exampleDescription,
+describe('attached results analytics actions', () => {
```
olamothe, line 100: We are in the process of upgrading one core dependency in Headless (redux toolkit) and this way of doing tests unfortunately won't work anymore with the new version.

I'm currently in the process of migrating all tests, and you could already do the change in this PR while adding new tests.

Here is an example of the type of change that's needed: https://github.com/coveo/ui-kit/pull/3644/files#diff-152307c304f6b731b4f0bb4e7591cec659599180ab305f7d24bee53d0b5e5f81R1

* Change import from `'../../test/mock-engine'` to `'../../test/mock-engine-v2'`
* Test calling the action directly: `logCaseAttach(testResult)()(engine.dispatch, ()=> engine.state, {} as ThunkExtraArguments)` as opposed to going through the engine dispatch. You can mock anything in `ThunkExtraArguments` instead of an empty object if needed, but for most analytics actions and this type of test, this should not be necessary🤞.

This comment would apply for all the new test files added in this PR ! Thanks !

packages/headless/src/features/insight-search/insight-analytics-actions.test.ts
olamothe: Yeah

olamothe (approved):

packages/headless/src/features/result/result-insight-analytics-actions.ts
```diff
@@ -20,5 +21,29 @@ export const logDocumentOpen = (result: Result) =>
...
+          author: information.documentAuthor,
+          url: information.documentUri,
+        },
+        position: information.documentPosition,
+        searchUid: state.search?.response.searchUid || '',
+        action: 'open',
+        context: {
+          targetId: metadata.caseId || '',
+          targetType: 'Case',
+        },
```
SimonMilord, line 46: question do we not want to pass `caseNumber: metadata.caseNumber` in the context for logDocumentOpen ?

SimonMilord (approved):
I agree with Etienne's comment I think we could create a `features/analytics/insight-analytics-utils.ts` that would generate the stuff we return in analytics, it would be more readable IMO, but the rest looks 🔥 GG 🎉

packages/headless/src/features/analytics/analytics-utils.ts
mmitiche: That was a fix to a problem we had with the `makeAnalyticsActionFactory`, @louis-bompart  identified and fixed this issue by making this change.

louis-bompart: Yes. oopsie daisy.

louis-bompart (changes requested):
Overall, looks good to me.
_However_, the `context` here is not set "the right way":
Indeed, context information should be sent to both search requests (even tho it goes thru the Customer Service API) and EP events.
To handle that, we uses contextController.
Check: ./packages/headless/src/controllers/core/context/example-deleteme-headless-case-context.ts and feel free to adapt it ;D

mmitiche:
> Overall, looks good to me. _However_, the `context` here is not set "the right way": Indeed, context information should be sent to both search requests (even tho it goes thru the Customer Service API) and EP events. To handle that, we uses contextController. Check: ./packages/headless/src/controllers/core/context/example-deleteme-headless-case-context.ts and feel free to adapt it ;D

Thanks for pointing this out @louis-bompart, using a context controller seems indeed like more organized, clean  and general way to handle the case context we are dealing with in the insight use case.

However I think this PR is not really the right place to introduce this new logic, I can create a Jira ticket and work on it in a separate PR that will all include:

- The creation of the Case Context Controller using the logic you described.
- Updating all the occurrences where we had to grab and use the context for the insight use case and make it grab the context from `ContextSection`  instead of the `InsightCaseContextSection`  that we are currently using with the current logic.

What do you think?

louis-bompart:
> > Overall, looks good to me. _However_, the `context` here is not set "the right way": Indeed, context information should be sent to both search requests (even tho it goes thru the Customer Service API) and EP events. To handle that, we uses contextController. Check: ./packages/headless/src/controllers/core/context/example-deleteme-headless-case-context.ts and feel free to adapt it ;D
> 
> Thanks for pointing this out @louis-bompart, using a context controller seems indeed like more organized, clean and general way to handle the case context we are dealing with in the insight use case.
> 
> However I think this PR is not really the right place to introduce this new logic, I can create a Jira ticket and work on it in a separate PR that will all include:
> 
> * The creation of the Case Context Controller using the logic you described.
> * Updating all the occurrences where we had to grab and use the context for the insight use case and make it grab the context from `ContextSection`  instead of the `InsightCaseContextSection`  that we are currently using with the current logic.
> 
> What do you think?

👍, please create the Jira and flag up with `//TODO JIRA-123` the places of impact in the changes introduced in this PR, and you have a deal ;)

mmitiche:
> > > Overall, looks good to me. _However_, the `context` here is not set "the right way": Indeed, context information should be sent to both search requests (even tho it goes thru the Customer Service API) and EP events. To handle that, we uses contextController. Check: ./packages/headless/src/controllers/core/context/example-deleteme-headless-case-context.ts and feel free to adapt it ;D
> > 
> > 
> > Thanks for pointing this out @louis-bompart, using a context controller seems indeed like more organized, clean and general way to handle the case context we are dealing with in the insight use case.
> > However I think this PR is not really the right place to introduce this new logic, I can create a Jira ticket and work on it in a separate PR that will all include:
> > 
> > * The creation of the Case Context Controller using the logic you described.
> > * Updating all the occurrences where we had to grab and use the context for the insight use case and make it grab the context from `ContextSection`  instead of the `InsightCaseContextSection`  that we are currently using with the current logic.
> > 
> > What do you think?
> 
> 👍, please create the Jira and flag up with `//TODO JIRA-123` the places of impact in the changes introduced in this PR, and you have a deal ;)

Jira created: https://coveord.atlassian.net/browse/SFINT-5420 and TODO added 👌

louis-bompart (approved):

packages/headless/src/features/result/result-insight-analytics-actions.ts
mmitiche: Case number has been added!

PR #3642, opened 2024-02-26
feat(auto-facet): bump max count to 20

louis-bompart:
https://coveord.atlassian.net/browse/KIT-3013

### DIFF
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -88,7 +88,7 @@ export namespace Components {
      */
     interface AtomicAutomaticFacetGenerator {
         /**
-          * The desired count of automatic facets.  Minimum: `1` Maximum: `10`
+          * The desired count of automatic facets.  Minimum: `1` Maximum: `20`
           * @defaultValue `5`
          */
         "desiredCount": number;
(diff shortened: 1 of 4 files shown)


olamothe (approved):
I was first wondering where this request was coming from. 

I see it's a change in search API, but do you know if this originate from a client or implementation request initially ? Curious about the usage of this feature.

louis-bompart:
> I was first wondering where this request was coming from.
> 
> I see it's a change in search API, but do you know if this originate from a client or implementation request initially ? Curious about the usage of this feature.

@jeremierobert-coveo?

fbeaudoincoveo (approved):

jeremierobert-coveo:
> I was first wondering where this request was coming from.
> 
> I see it's a change in search API, but do you know if this originate from a client or implementation request initially ? Curious about the usage of this feature.

It was coming from customers. The maximum is dynamically controlled on our side.
See https://coveo.slack.com/archives/C03596QE3DE/p1708532518949489

Do you think it makes sense to keep a max in the UI if the limit is validated on the backend side? (Could yield some 4XX though if used badly)

louis-bompart:
> > I was first wondering where this request was coming from.
> > I see it's a change in search API, but do you know if this originates from a client or implementation request initially? Curious about the usage of this feature.
> 
> It was coming from customers. The maximum is dynamically controlled on our side. See https://coveo.slack.com/archives/C03596QE3DE/p1708532518949489
> 
> Do you think it makes sense to keep a max in the UI if the limit is validated on the backend side? (Could yield some 4XX though if used badly)

Yes, two reasons essentially:
 1. People won't check the DOC of the SearchAPI for a Headless/Atomic implementation, so having the values here helps by preventing them to shoot'em self in the foot, by having it in Intellisense and the doc.
 2. We prevent doomed queries to be sent, which is nice cost-wise.

PR #3645, opened 2024-02-26
refactor(ci): refactor quantic jobs into reusable workflow

louis-bompart:
- Refactor Quantic CI into a reusable workflow
- Skip Quantic on `pull_request` and `merge_group` if no changes on headless, quantic, or the pjson has been detected

### DIFF
--- a/.github/workflows/e2e-quantic.yml
+++ b/.github/workflows/e2e-quantic.yml
@@ -5,8 +5,6 @@ on:
         required: true
       SFDX_AUTH_JWT_KEY:
         required: true
-      GITHUB_TOKEN:
-        required: true
 jobs:
   e2e-quantic-setup:
     name: 'Setup e2e tests on Quantic'
--- a/.github/workflows/e2e-quantic.yml
+++ b/.github/workflows/e2e-quantic.yml
@@ -1,0 +1,53 @@
+on:
+  workflow_call:
+    secrets:
+      SFDX_AUTH_CLIENT_ID:
+        required: true
+      SFDX_AUTH_JWT_KEY:
+        required: true
+      GITHUB_TOKEN:
+        required: true
+jobs:
+  e2e-quantic-setup:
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-quantic.yml
```diff
@@ -0,0 +1,50 @@
...
+            'cypress/e2e/default-2/**/*',
+            'cypress/e2e/facets-1/**/*',
+            'cypress/e2e/facets-2/**/*',
+          ]
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+      - uses: ./.github/actions/setup
+      - uses: ./.github/actions/e2e-quantic
+      - uses: ./.github/actions/setup-sfdx
+        with:
```
louis-bompart, line 39:
```suggestion
```

scripts/ci/hasFileChanged.mjs
```diff
@@ -0,0 +1,53 @@
...
+    stdio: 'pipe',
+    encoding: 'utf-8',
+  });
+}
+
+function getPatterns() {
+  return process.argv.slice(3);
+}
+
+function checkPatterns(files, patterns) {
```
fbeaudoincoveo, line 33: Not a big deal at all, but you don't even need to check all files against all patterns; as soon as there's a single match you can return true.

fbeaudoincoveo (approved):
I like this idea 👍

scripts/ci/hasFileChanged.mjs
louis-bompart: This is what the code does already tho 😄

mmitiche (approved):
LGTM!

scripts/ci/hasFileChanged.mjs
fbeaudoincoveo: Indeed, I'm silly 🙃

PR #3652, opened 2024-02-28
chore: return upload-artifact to v3

louis-bompart:
https://github.com/actions/upload-artifact/blob/main/README.md#not-uploading-to-the-same-artifact

https://coveord.atlassian.net/browse/KIT-3017


louis-bompart:
We can also go straight with https://github.com/coveo/ui-kit/pull/3653

fbeaudoincoveo (approved):

PR #3654, opened 2024-02-28
feat(headless/commerce): allow purchasing from cart controller

sallainCoveo:
Jira: https://coveord.atlassian.net/browse/LENS-1498

This branch is to add an action to log the `ec.purchase` when the `purchase` function from the CartController is called. 

Demo on Barca sports (linked to this local build)

https://github.com/coveo/ui-kit/assets/58052881/9f713808-f7d0-4a6c-b16f-c117fb06c411

Open question: it would be consistent to emit a `ec.purchase` event wih an empty array of product (you basically buy nothing), it would be a valid event. Do we have some situations that it does make sense, or should we update the schema?


packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -173,8 +178,20 @@ export function buildCart(engine: CommerceEngine, props: CartProps = {}): Cart {
       }
     },
 
-    purchase(_transactionId: string, _transactionRevenue: number) {
-      // TODO LENS-1498: log ec.purchase with all products in cart.
+    purchase(transactionId: string, transactionRevenue: number) {
```
samisayegh, line 181: Is it worth making the arguments an object, since we might ask for more details about the transaction in the future? e.g. shipping and taxes.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -173,8 +178,20 @@ export function buildCart(engine: CommerceEngine, props: CartProps = {}): Cart {
...
+        ({quantity, ...product}) => ({quantity, product})
+      );
+      const transaction = {id: transactionId, revenue: transactionRevenue};
+
+      const payload: Ec.Purchase = {
+        currency,
+        products,
+        transaction,
+      };
+      engine.relay.emit('ec.purchase', payload);
```
samisayegh, line 193: Nice to see this, seems much simpler to dispatching an action. Is the rest of headless moving to this pattern?

samisayegh (approved):

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
olamothe: Problem with this is that there's a significant amount of logic embedded in the controller to emit a proper purchase event.

Versus a more "self enclosed" function which would get access to the state to do this job. And so it becomes less portable.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
sallainCoveo: IMO, yes. The only concern I have is it will cause an error in Barca sports. Should it be identify as a breaking change?

samisayegh: Yep, if we make this adjustment, it would be considered a breaking change. The window to make these freely will soon be closing but I think it's still open,

sallainCoveo: This [commit](https://github.com/coveo/ui-kit/pull/3654/commits/caf3117d46a62caea8cda7c6e3e6550e2ea0cf60) contains this change (I'm open to keep it or revert it).

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
sallainCoveo: I'm not too sure how to interpret this to be honest. Are you suggesting that the logic (and function) shouldn't be part of the controller? Is there a preferred pattern for these kind of actions?

I wanted to mention that I'm not sure we can rely fully on the state (at least for now) for some of the events. For example, if we want to know whether we "add" or "remove" an item from the cart, we have to compare the item quantity from state vs the item we are updating. Maybe it means we could have functions in which we pass some arguments when required, but the rest is from the state or we have to update the state.

sallainCoveo: Ok I think I have a better understanding, and I agree (especially to avoid DRY code). 

I'm not sure there an obvious answer, but let's say that we introduce this flow with this PR, at which level of the package should we implement these functions? It seems to me it would be part of the `CoreEngine` (because Relay is LOBs agnostic), but it seems a little bit too high level (and I'm not sure an implementer expect to have these functions availables either).

sallainCoveo: I just want to confirm that I understand the idea:
We add a Redux action (for example `purchaseItems`) within the `cart-action` file. Then, we add a `.addCase` within the `cart-slices` file to adress this action (clear the items + send the analytics). I had a discussion with @samisayegh to figure out how we can tackle this one (which diverge from the initial suggestion), and we came with four options:

OPTION A:  We set some kind of helper method in which we pass the engine and the transaction as parameters. That way, we have a reusable function, we can use the relay instance to emit the function and we dispatch only one action ('purchaseItems'). The downpoint, the action does not emit the EP event.

OPTION B: We set two actions; one using the `makeAnalyticsActions` to emit the `ec.purchase` event and one to clear the items in the cart `purchaseItems`. It is not aligned with this Redux [recommendation](https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially), but the analytics action should not impact the UI. However, if a reducer could listen to a makeAnalyticsActions' function (within `addCase`) we could avoid two actions :eyes: .

OPTION C: We use the `buildThunkExtraArguments` function to set Relay as the apiClient (in the commerce engine), and we make an `AsyncThunkAction` for `purchaseItems`. This option will allow us to emit an EP event within the action + empty the cart. However, it would mean that the action is async (even if it's not necessary) and we have to make sure that the Relay's instance is dynamic (if we set the apiClient once, it increases the risk to have out-of-sync configurations).

OPTION D: We pass all necessary arguments to construct the event payload and create an instance of relay in the action's payload (so we pass the transaction, the currency and the configuration state). In this [recommendation](https://redux.js.org/style-guide/#put-as-much-logic-as-possible-in-reducers), we should add as much logic as possible within the reducer. However, it seems to me that the arguments we pass in the payload wouldn't have an impact on the cart state, so I'm not sure the recommendation applies in that case.

Obviously, I might have miss some relevant points on how the project is structured and works (there's probably more option), so your input is more than welcome on how we should implement this action. But, if I had to pick an option, I would lean towards option(s) A or B (B especially if a reducer can be set with an analytics' function).

louis-bompart: Mh, this is a tricky matter, I don't want to rush a decision.
As @olamothe mentioned this change can be done without any impact to this controller.
So let's scope this issue out for now.

louis-bompart (approved):

louis-bompart:
I did some experiment #3693 is the result.
IMO the ThunkAction is the way.
I was mistaken when I said "let's send the analytic in the reducer" as it goes against https://redux.js.org/style-guide/#reducers-must-not-have-side-effects. (which is a big nono)

Instead, after reading a bunch on RTKQ and AsyncThunk, I concluded that an AsyncThunk is the lesser evil.

I think it fit the bill too:
 - Users could use it outside of a cartController, it's just that _if_ they have a cartController, it'll do the 'setItems([]) as well.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -90,10 +90,9 @@ export interface Cart extends Controller {
   /**
    * Emits an `ec.purchase` analytics event and then empties the cart without emitting any additional events.
    *
-   * @param transactionId - The transaction ID.
-   * @param transactionRevenue - The total revenue from the transaction, including taxes, shipping, and discounts.
+   * @param transaction - The object with the id and the total revenue from the transaction, including taxes, shipping, and discounts.
    */
-  purchase(transactionId: string, transactionRevenue: number): void;
+  purchase(transaction: {id: string; revenue: number}): void;
```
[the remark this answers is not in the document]
sallainCoveo: Do you think the transaction interface will be used in other circumstances than emitting an event? I'm asking, because I wonder if it should be added to the @coveo/relay-event-types package

PR #3655, opened 2024-02-28
chore(quantic): replace xml2json with xml2js

erocheleau:
SFINT-5396

Replace the obsolete package xml2json with xml2js, a package that seems to have more usage and is up-to-date.

Compare the result of both functions

Above: xml2json (before)
Below: xml2js (now)

_And we only use `apiVersion`, `isExposed` and `targets.target` in the script `publish.js` which creates jsdoc for the components._

image


SimonMilord (approved):

mmitiche (approved):
LGTM 👍

fbeaudoincoveo (approved):

louis-bompart:
Will require https://github.com/coveo/ui-kit/pull/3659

olamothe (approved):

PR #3658, opened 2024-02-29
feat(headless): expose relay on the engine

louis-bompart:
- Refactor the configureAnalytics & getAnalyticsSource
  - Make them into memoized selectors to shave off some computing
- Exposes Relay onto the engine.

KIT-3018

### DIFF
--- a/packages/headless/src/api/analytics/analytics-relay-client.test.ts
+++ b/packages/headless/src/api/analytics/analytics-relay-client.test.ts
@@ -1,8 +1,10 @@
 import {createRelay} from '@coveo/relay';
 import {createMockState} from '../../test/mock-state';
 import {getRelayInstanceFromState} from './analytics-relay-client';
+import {getAnalyticsSource} from './analytics-selectors';
 
 jest.mock('@coveo/relay');
+jest.mock('./analytics-selectors');
 
 describe('#getRelayInstanceFromState', () => {
   const mockedCreateRelay = jest.mocked(createRelay).mockImplementation(() => ({
@@ -14,6 +16,10 @@ describe('#getRelayInstanceFromState', () => {
     updateConfig: jest.fn(),
     version: 'test',
   }));
+
+  beforeEach(() => {
+    jest.mocked(getAnalyticsSource).mockReturnValue(['baguette']);
+  });
 
   afterEach(() => {
     jest.clearAllMocks();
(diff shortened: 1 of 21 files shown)


packages/headless/src/api/analytics/analytics-relay-client.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  it('creates a Relay client properly and returns it', () => {
+    const state = createMockState();
+
+    const relay = getRelayInstanceFromState(state);
+
+    expect(mockedCreateRelay).toHaveBeenCalledWith({
+      url: state.configuration.analytics.nextApiBaseUrl,
+      token: state.configuration.accessToken,
+      trackingId: state.configuration.analytics.trackingId,
+      source: expect.arrayContaining([]),
```
olamothe, line 31: Shouldn't there be something in the array ? Like `@coveo/headless@unit-testing-version` (from createMockState()) ?

packages/headless/src/app/engine.test.ts
```diff
@@ -259,4 +261,23 @@ describe('engine', () => {
...
   });
+
+  it('should exposes a Relay instance', () => {
+    const engine = initEngine();
+    expect(engine.relay).toBeDefined();
+  });
+
+  it('should return the same instance of Relay if the state has not changed', () => {
+    const engine = initEngine();
+    expect(engine.relay).toBe(engine.relay);
```
olamothe, line 272: That seems like a useless test 🤔  ?

louis-bompart: Not really: This test shows that if we call the relay getter twice for a given state, we get the same instance of Relay.

packages/headless/src/api/analytics/analytics-relay-client.test.ts
louis-bompart: 👍 tested through isolation

packages/headless/src/app/engine.test.ts
olamothe: Ok I understand. 
Then maybe just a small comment that explains this would help future readers

olamothe (approved):

packages/headless/src/app/engine.test.ts
```diff
@@ -259,4 +261,23 @@ describe('engine', () => {
       });
     });
   });
+
+  it('should exposes a Relay instance', () => {
+    const engine = initEngine();
+    expect(engine.relay).toBeDefined();
+  });
+
+  it('should return the same instance of Relay if the state has not changed', () => {
```
louis-bompart, line 270:
```suggestion
  it('should return the exact same instance of Relay if the state has not changed between two calls to engine.relay', () => {
```

Spuffynism (approved):

PR #3667, opened 2024-03-05
feat(headless/commerce): set currency type as currency code iso4217 instead of string

sallainCoveo:
Jira: https://coveord.atlassian.net/browse/LENS-1589

This branch is to change the type of the currency set in the commerce state (context-state). It was a `string`, it's now switch to `CurrencyCodeISO4217`. 

By doing this, the value set by the user for the commerce context will be type-checked differently (illustrate by the demo below):

https://github.com/coveo/ui-kit/assets/58052881/472fbfd2-b1c6-415e-9107-2076b997d592

### DIFF
--- a/packages/headless/src/features/commerce/context/context-state.ts
+++ b/packages/headless/src/features/commerce/context/context-state.ts
@@ -1,3 +1,4 @@
+import {CurrencyCodeISO4217} from '@coveo/relay-event-types';
 import {
   UserParams,
   ViewParams,
@@ -7,7 +8,7 @@ export interface CommerceContextState {
   trackingId: string;
   language: string;
   country: string;
-  currency: string;
+  currency: CurrencyCodeISO4217;
   user?: UserParams;
   view: ViewParams;
 }
@@ -16,7 +17,7 @@ export const getContextInitialState = (): CommerceContextState => ({
   trackingId: '',
   language: '',
   country: '',
-  currency: '',
+  currency: '' as CurrencyCodeISO4217,
   view: {
     url: '',
   },
--- a/packages/headless/src/features/commerce/context/context-validation.ts
+++ b/packages/headless/src/features/commerce/context/context-validation.ts
@@ -1,9 +1,15 @@
-import {RecordValue, Schema} from '@coveo/bueno';
+import {RecordValue, Schema, StringValue} from '@coveo/bueno';
+import {CurrencyCodeISO4217} from '@coveo/relay-event-types';
 import {
   nonEmptyString,
   nonRequiredEmptyAllowedString,
   requiredNonEmptyString,
 } from '../../../utils/validate-payload';
+
+export const currencyDefinition = new StringValue<CurrencyCodeISO4217>({
+  required: true,
+  emptyAllowed: false,
+});
 
 export const viewDefinition = {
   url: requiredNonEmptyString,
@@ -21,7 +27,7 @@ export const contextDefinition = {
   trackingId: requiredNonEmptyString,
   language: requiredNonEmptyString,
   country: requiredNonEmptyString,
-  currency: requiredNonEmptyString,
+  currency: currencyDefinition,
   user: new RecordValue({
     values: {
       ...userDefinition,
--- a/packages/headless/src/features/commerce/context/context-validation.ts
+++ b/packages/headless/src/features/commerce/context/context-validation.ts
@@ -6,9 +6,12 @@ import {
   requiredNonEmptyString,
 } from '../../../utils/validate-payload';
 
+const currencies = Intl.supportedValuesOf('currency') as CurrencyCodeISO4217[];
+
 const currencyDefinition = new StringValue<CurrencyCodeISO4217>({
   required: true,
   emptyAllowed: false,
+  constrainTo: currencies,
 });
 
 export const viewDefinition = {
--- a/packages/headless/src/features/commerce/context/context-validation.ts
+++ b/packages/headless/src/features/commerce/context/context-validation.ts
@@ -6,7 +6,7 @@ import {
   requiredNonEmptyString,
 } from '../../../utils/validate-payload';
 
-export const currencyDefinition = new StringValue<CurrencyCodeISO4217>({
+const currencyDefinition = new StringValue<CurrencyCodeISO4217>({
   required: true,
   emptyAllowed: false,
 });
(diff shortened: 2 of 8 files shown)


packages/headless/src/features/commerce/context/context-state.ts
```diff
@@ -16,7 +17,7 @@ export const getContextInitialState = (): CommerceContextState => ({
   trackingId: '',
   language: '',
   country: '',
-  currency: '',
+  currency: '' as CurrencyCodeISO4217,
```
sallainCoveo, line 20: If we want to avoid casting here, should we set a currency (like "USD")?

packages/headless/src/features/commerce/context/context-validation.ts
```diff
@@ -1,10 +1,16 @@
-import {RecordValue, Schema} from '@coveo/bueno';
+import {RecordValue, Schema, StringValue} from '@coveo/bueno';
+import {CurrencyCodeISO4217} from '@coveo/relay-event-types';
 import {
   nonEmptyString,
   nonRequiredEmptyAllowedString,
   requiredNonEmptyString,
 } from '../../../utils/validate-payload';
 
+export const currencyDefinition = new StringValue<CurrencyCodeISO4217>({
```
sallainCoveo, line 9: Important to note that this definition won't evaluate in the runtime if the string set is actually a currency. Currently, we don't have an exposed list of values that could be set in the `constrainTo` parameter. 

The question would be is if it's worth adding a list in this project (which would increase the weight of the project) and if the runtime validation would bring value in that case.

packages/headless/src/features/commerce/context/context-state.ts
louis-bompart: Mh, I think it's OK with the cast

louis-bompart (approved):

Spuffynism (approved):

packages/headless/src/features/commerce/context/context-validation.ts
olamothe: I've done a bit of digging, and I think we could get that list from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf

```javascript
Intl.supportedValuesOf('currency')
```

So, no need to define that list somewhere else and import it, it seems native to all modern browser and node environments

olamothe, line 9: Since it's not used anywhere else, we can remove the `export`

sallainCoveo: Oh, TIL! That's interesting to know :eyes:. Will try to use that.

samisayegh: Would `Intl` be available in nodejs when rendering on the server? Could be something to consider in the cost of validating in headless.

Also worth noting that both Commerce API and and Event API will soon be validating the value at runtime. If the value is invalid, Commerce API will not return results,

packages/headless/src/features/commerce/context/context-state.ts
samisayegh: Out of scope for this PR, but it's worth thinking about when these properties need to be set. They have to be in place before the first request. and they're all required. It feels like they are configuration options on the commerce engine rather than context?

If they were set on init, then we could set default values, avoid casting, and not worry about confusing a developer as to where `USD` came from. If we use context, then I'd avoid setting a default. Casting is ok, but avoiding the problem by making them config options feels like the better DX,

What do you think @Spuffynism?

Spuffynism: Great point!

Is the engine always created _after_ we know all the information in the context? If so, eliminating the context controller and replacing it by configuration options on the engine would make sense to me 🤔

I'm thinking in a SSR context, one might want to create the engine server-side, and then create the context controller client side, but I'm not sure that's an actual use case.

packages/headless/src/features/commerce/context/context-validation.ts
louis-bompart: > Would `Intl` be available in nodejs when rendering on the server? Could be something to consider in the cost of validating in headless.
> 
> Also worth noting that both Commerce API and and Event API will soon be validating the value at runtime. If the value is invalid, Commerce API will not return results,

I had the same question for Oli earlier, short answer yes.
Long answer: Yes, on Node 18+, which is the minimal version of NodeJS we will support for V3 (so essentially that we will support for this engine once GA)
image

sallainCoveo: Just FYI, if I try to add this function in the code, I will get this error: `property 'supportedvaluesof' does not exist on type 'typeof intl'`

From my understanding, we have to use at least Typescript 5.1, Nodejs v18 and use at least "ES2022" lib (in the tsconfig). There is a workaround describe [here](https://github.com/microsoft/TypeScript/issues/49231), but I wonder what would be the best approach to use this method.

louis-bompart: > Just FYI, if I try to add this function in the code, I will get this error: `property 'supportedvaluesof' does not exist on type 'typeof intl'`
> 
> From my understanding, we have to use at least Typescript 5.1, Nodejs v18 and use at least "ES2022" lib (in the tsconfig). There is a workaround describe [here](https://github.com/microsoft/TypeScript/issues/49231), but I wonder what would be the best approach to use this method.

Can you try to add `"ES2022.Intl"` into `lib` array of the tsconfig of headless?
I think it's likely to work.

sallainCoveo: Yep it works that way!

I think I still need to do casting at some point though (Relay expect `CurrencyCodeISO4217`, so even if we constrain to the values from Intl.supportedValuesOf('currency'), the type is considered as string[]). 

On another note, there is some differences between the values returned by `Intl` and `CurrencyCodeISO4217`; the latter contains crypto currencies (such as bitcoin).

packages/headless/src/features/commerce/context/context-state.ts
samisayegh: One might want to update the values after the engine is created, e.g. a customer changes language mid session, so the controller would be useful in that scenario.

I think the values will be known before creating the engine. They have to be there before the engine can make he first request 🤷. This feels like the "primary" setup flow and would guarantee that the params are set before the first request, but the controller will still be helpful for implementations that are multi-language,

PR #3670, opened 2024-03-05
feat(atomic,headless): remove customData & middleware iff analyticsMode=next

louis-bompart:
- Rework all SearchActions (the new ones) to remove the getExtraEventPayload as it was solely used for the customData
- Remove the customData from the SAPI payload iif analyticsMode=next
- Do not register middleware iif analyticsMode=next


olamothe (approved):
:gg: Really clean with all the metadata gone

PR #3672, opened 2024-03-05
feat(commerce): create commerce product view controller shell for analytics

Spuffynism:
Create product view controller shell for product view analytics event.

[CAPI-501]

[CAPI-501]: https://coveord.atlassian.net/browse/CAPI-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/product-view/headless-product-view.test.ts
```diff
@@ -0,0 +1,36 @@
+import {ProductRecommendation} from '../../../api/search/search/product-recommendation';
+import {buildMockCommerceState} from '../../../test/mock-commerce-state';
+import {
+  buildMockCommerceEngine,
+  MockedCommerceEngine,
+} from '../../../test/mock-engine-v2';
+import {buildProductView, ProductView} from './headless-product-view';
+
+
+describe('Productview', () => {
```
samisayegh, line 10: Nit: would camel-casing be more consistent?
```suggestion
describe('ProductView', () => {
```

packages/headless/src/controllers/commerce/product-view/headless-product-view.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  let engine: MockedCommerceEngine;
+  // TODO: What goes here?
+  let product: ProductRecommendation;
+  let productView: ProductView;
+
+  function initializeProductView() {
+    // TODO: What goes here?
+    product = {};
```
samisayegh, line 18: The product has only these [three properties](https://github.com/coveo-platform/analytics_schema/blob/master/schemas/common/ec/product.json#L4). We can fill these in when integrating relay,

samisayegh (approved):
Thanks 🙏

packages/headless/src/controllers/commerce/product-view/headless-product-view.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Creates an `ProductView` controller instance.
+ *
+ * eslint-disable-next-line @cspell/spellchecker
+ * TODO LENS-1500: Make use of the engine
+ * @param _engine - The headless commerce engine.
+ * @returns A `ProductView` controller instance.
+ */
+export function buildProductView(
```
Spuffynism, line 26: I questioned having the controller receive the product as props, but decided against since doing so would prevent it being reused.

Spuffynism (commented):
@sallainCoveo @samisayegh Wanna take over this controller/PR for LENS-1500? Without the events being dispatched, the controller doesn't do very much. I don't think it's worth it to invest efforts into getting the PR merged without analytics.

packages/headless/src/controllers/commerce/product-view/headless-product-view.ts
samisayegh: Why would props prevent reuse?

samisayegh:
Ok, we'll take it from here 👍

packages/headless/src/controllers/commerce/product-view/headless-product-view.ts
Spuffynism: Sorry, that wasn't super clear! What I mean is that requiring a product when creating the controller would prevent reuse across products, without creating a new controller.

By having the controller not require a product as property, we can call `productView.view(product)` for every new `product` that's viewed, and reuse the same controller. Maybe that isn't what we want though? Would a product view event usually only be called when we're on a pdp, with a single product?

louis-bompart: Actually.
I'd like to question (again) the utility of this event, or at the very least its implementation in Headless through a controller.

To start, let's remember the idea that with the new data platform, we only and solely track what allows us to empower our Platform to bring relevance to our customers. (At least, that's my understanding and what we've been working towards).

With that in mind, in which case the view of a product should be signalled to our Data Platform? I see a few cases.

1. When a customer navigates on the website to a PDP.
   1.  The Product View is/should already tracked by the User Interaction that triggered the navigation. If this user interaction occurred on a Coveo-powered element, an associated click should probably be triggered (or the product view event itself)
   1.  Moreover, if there's a Recommendation carousel or any kind of coveo-powered interface on the PDP, the actionCause associated to the initial search event that occurs after/during the navigation should probably be 'marked'  
1. When a customer lands on a PDP
   1.  If we have some kind of interfaces, 1.i applies. If not, why should we track? No Coveo was nor is involved. 
   
In essence, this event reminds me too much of the "PageView" event that we are flushing in the new protocol (afaik).
I think we should be able to capture the data through other means, which would reduce the burden on the implementers.

@fvanham-coveo , I'd like your insight on this maybe?


TL;DR: ProductView should already be tracked through the initial Search Event of an interface associated tightly to a PDP and product Click, but probably nothing else (IMO)

louis-bompart (commented):
Ain't sure chief.

packages/headless/src/controllers/commerce/product-view/headless-product-view.ts
samisayegh: Hey Louis, the `ec.productView` replaces the `detail` collect event, and is indeed similar to the UA view event (aka "page view"). The event signals that a user showed interest in a document that exists in the index. When researching last year, ML told us that the event is important and the use-cases are listed [here](https://docs.google.com/spreadsheets/d/1eegdQ4TYTYjfDmgTiqHdX97aGifCC2PhgG8kDYDJFfA/edit#gid=0&range=H6).

It is different to the commerce "page view" which signals that a page was viewed. This event is now [deprecated](https://docs.coveo.com/en/l3am0254/coveo-for-commerce/capture-a-page-view-event#:~:text=The%20page%20view%20event%20has%20been%20deprecated%20and%20is%20no%20longer%20necessary%20to%20send.%20Unless%20you%20are%20creating%20custom%20reports%20using%20this%20event%2C%20we%20recommend%20that%20you%20stop%20sending%20it%2C%20as%20doing%20so%20will%20simplify%20your%20implementation.) in UA and is not available in EP.

You're right that the event is sent on a page that's usually not powered by Coveo. However, similar to add-to-cart and purchase events, these signals are used to learn and improve the experience, and to power reports.

Can ML use other signals to infer the information? It might be possible to be more efficient in the future, but for launch this would have been too big of an ask for the ML stack.

louis-bompart: It can make sense to indeed send events for now 👍 
I'm still not convinced about the place/role of Headless in this tho.
I don't see the controller either:
 - Having to play with Headless state
 - Having to play with the 'Search' queries
 
 So I wonder why it should be a controller 🤔

olamothe: Indeed, as it stands right now, there's no use for it to be a controller. 

I can definitely see that possibly, in the future, there could be a need for a product view controller because it will need to interact with the state, or read some properties from it. But as it stands, with this PR, it's an empty shell.

And so I'd prefer if we refrain from creating a controller for this right now because "potentially it will need to do more in the future".

Since relay instance is exposed on the engine and created from the state, and so the "configuration" of relay is then done automatically for implementers, maybe we could just recommend this approach for implementers instead: 

```
commerceEngine.relay.emit('ec.productView', {theProduct})
```

samisayegh: Great points hmm, it's definitely not a traditional controller for the reasons you listed.

We could simply ask implementers to use the relay instance on the engine, but two considerations:

1. There would be a DX benefit to providing something. It would handle populating the `ec.productView` string and provide intellisense on the payload type. Asking implementers to use the `relay` instance means they'd need to reference the docs to get it right, and only have runtime validation.
2. While it doesn't have to be a controller, it would be consistent to use such a pattern because it is familiar to implementers? We provide a controller for click, cart and purchase events, but product view would be missing? This would stand out in the docs too. I think it would feel incomplete to readers?

If we don't want to create dedicated controller just yet, could a middle ground be to include it as a method on the cart controller?

Product views will typically be emitted on a PDP, which is also where add-to-carts happen. So having the methods next to one another feels reasonable? These events are closer than the `purchase` event which currently sits on the cart controller.

Another idea could be to generalize the name of the cart controller to reflect that it handles everything related to the journey after the click, from viewing to adding to cart to purchasing. e.g.

```
const j = buildProductJourney()

j.viewProduct(...)
j.addToCart(...)
j.purchase();
```
I'm not a fan of the name, I like "cart" more because it's concrete. Just wanted to point out that the controller is doing more than cart-related functionality.

Spuffynism: Agree about the DX benefit of providing something :+1:

---

> Another idea could be to generalize the name of the cart controller to reflect that it handles everything related to the journey **after** the click, from viewing to adding to cart to purchasing. e.g.

Why only after? How would you feel about having a journey controller handle the product click event as well? Doing so, would allow us to replace the interactive result controller entirely. We could spread its method on the journey controller, or even provide a reference to it:
```ts
interface Journey {
  // either we bring up the methods
  beginSelect(p: Product): void;
  cancelSelect(p: Product): void;
  select(p: Product): void;

  // or we offer a reference to it
  interactiveResult: InteractiveResult;

  // rest of the interface is unchanged
  viewProduct(p: Product): void;
  addToCart(p: Product): void;
  purchase(): void;
}
```

samisayegh: We could include the click too, I like that it groups all the interactions/events in one place 🙌

olamothe (approved):

packages/headless/src/controllers/commerce/product-view/headless-product-view.ts
samisayegh: From the discussion in [slack](https://coveo.slack.com/archives/G016TA2G485/p1710336972900509?thread_ts=1710331201.276499&cid=G016TA2G485), the preference is to use a dedicated controller rather than aggregate all events in a journey controller.

> Picking between a dedicated product view controller that does too little or a journey controller that (seem) to do too much and is more abstract, I vote for product view controller

Will resolve this conversation and build on the work in this PR, thanks all 🙏

packages/headless/src/controllers/commerce/product-view/headless-product-view.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Creates an `ProductView` controller instance.
+ *
+ * @param engine - The headless commerce engine.
+ * @returns A `ProductView` controller instance.
+ */
+export function buildProductView(engine: CommerceEngine): ProductView {
+  return {
+    view: (payload) => {
+      engine.relay.emit('ec.productView', payload);
```
Spuffynism, line 24: Should we take the currency from `state.commerceContext.currency`, instead of from the `payload`? It would allow us to flatten the `payload` param to just the product's properties.

Spuffynism (commented):
I can't approve since I'm the PR author so here's a handmade approval: ✅

packages/headless/src/controllers/commerce/product-view/headless-product-view.ts
louis-bompart: I like the idea.
I'd propose to drop the controller, and just exposes an action in the vein of https://github.com/coveo/ui-kit/blob/3dd096730cd1bc6e64acb2144b892f4da4fd6f57/packages/headless/src/features/commerce/context/cart/cart-actions.ts#L13-L25

You'd just do `engine.dispatch(productView(product))` and job done.

louis-bompart: (note; an asyncAction doesn't need to be reduced to be 'executed')

samisayegh: I like the idea of pulling `currency` from state, will make the change.
I'll adjust to dispatch an action too, but I don't like implementers needing to use a different pattern just for this event. 

I'll stress again that offering a consistent DX is important. If we ask a developer to dispatch an action for product view, we need to ask them to dispatch an action for all other client events.

louis-bompart: Fair enough 👍

louis-bompart (approved):

sallainCoveo (approved):
:raised_hands: Nice @Spuffynism & @samisayegh !

buggyvelarde (approved):

PR #3675, opened 2024-03-07
build(headless): update to redux toolkit 2

olamothe:
The migration/upgrade per say, following all tests migrations.

It did uncover some issues in some tests, but it was mostly a matter of changing/adjusting types.

🤞 All tests should pass...

https://coveord.atlassian.net/browse/KIT-2955


olamothe:
Whelp things are blowing up it seems 🔬

packages/headless/src/app/engine.ts
```diff
@@ -149,7 +151,7 @@ export interface ExternalEngineOptions<State extends object> {
 }
 
 function getUpdateAnalyticsConfigurationPayload(
-  options: EngineOptions<ReducersMapObject>,
+  options: any,
```
louis-bompart, line 154: :suspect: can you detail/explain, this seems wrong.

packages/headless/src/features/commerce/sort/sort-slice.test.ts
```diff
@@ -95,7 +95,9 @@ describe('product-listing-sort-slice', () => {
     it('resets sort', () => {
       state.appliedSort = sort;
       state.availableSorts = [sort];
-      const finalState = sortReducer(state, action);
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const finalState = sortReducer(state, action({} as any));
```
louis-bompart, line 100: I'm not fond of this, as this IMO should not work/pass, as the action creators should :rage1: if you pass an empty obj, and thus the reducer should not be triggered.

But I reckon this is outside the scope of this PR.

Same for the pagination-slice ☝️

louis-bompart (approved):

packages/headless/src/app/engine.ts
olamothe: Fixed. 
Explanation: I was lazy (forgot to fix after temp change) 😉

Issue #192, opened 2020-10-19
[BB pr#192] feat(headless): basic recommendation state + reducers


coveobot:
>  **Pull request** :twisted_rightwards_arrows: created by **@olamothe** on 2020-10-19 19:01
> Last updated on 2020-10-20 17:04
> Original Bitbucket pull request id: 192
>
> Participants:
>
> * **@acote-coveo** (reviewer)
> * **@ThibodeauJF** (reviewer)
> * **@btaillon** (reviewer)
> * bitbucket user **coveo-anti-throttling_02**
> * **@jlblais** (reviewer)
> * **@samisayegh** (reviewer) :heavy_check_mark:
> * **@nrawji** (reviewer)
>
> Source: Commit `93b1b33414a6` on branch `KIT-197`> Destination: https://bitbucket.org/coveord/ui-kit/commits/88f44b24e010 on branch master
> Merge commit: https://github.com/None/commit/51717fd006d1
>
> State: **`MERGED`**

Ground work: Will obviously need to add reducers for recommendation itself.

‌

But hopefully this shows how a new “shape” for the state can be done by picking the sections we need.

https://coveord.atlassian.net/browse/KIT-197

‌

PR #3679, opened 2024-03-08
feat(headless): in development, rebuild modules with source maps 

BenKandelaars:
### Proposed changes

Build ui-kit, with the immediate use case of Headless, in dev on file change for hot reloading when debugging with Barca Sports. Also, remove minification and add source maps to allow debugging alongside an implementation.

Companion PR in Barca Sports [#192](https://github.com/coveo/barca-sports/pull/192)

Hot reloading example
https://github.com/coveo/ui-kit/assets/15649627/825480c4-79c2-45b6-b9c6-e97fd8309e5c

Debugging with source maps example
Uploading Screen Recording 2024-03-08 at 18.03.23.mov…

### How to test

Link with Barca sports, using `yarn link` and add console.logs into the the running headless code. Changes should be reflected in the browser.

### Checklist:

- [x] Service tested locally
 

[CAPI-646]: https://coveord.atlassian.net/browse/CAPI-646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/esbuild.mjs
+++ b/packages/headless/esbuild.mjs
@@ -87,15 +87,23 @@ const browserEsm = Object.entries(useCaseEntries).map((entry) => {
   const [useCase, entryPoint] = entry;
   const outDir = getUseCaseDir('dist/browser', useCase);
   const outfile = `${outDir}/headless.esm.js`;
+  const config = {
+    entryPoints: [entryPoint],
+    outfile,
+    format: 'esm',
+  };
 
-  return buildBrowserConfig(
-    {
-      entryPoints: [entryPoint],
-      outfile,
-      format: 'esm',
-    },
-    outDir
-  );
+  return devMode
+    ? buildBrowserConfig(
+        {
+          ...config,
+          watch: devMode,
+          minify: false,
+          sourcemap: true,
+        },
+        outDir
+      )
+    : buildBrowserConfig(config, outDir);
 });
 
 const browserUmd = Object.entries(useCaseEntries).map((entry) => {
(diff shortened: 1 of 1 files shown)


packages/headless/esbuild.mjs
```diff
@@ -87,15 +87,23 @@ const browserEsm = Object.entries(useCaseEntries).map((entry) => {
...
-      outfile,
-      format: 'esm',
-    },
-    outDir
-  );
+  return devMode
+    ? buildBrowserConfig(
+        {
+          ...config,
+          watch: devMode,
```
louis-bompart, line 100:
```suggestion
          watch: true,
```

louis-bompart (approved):
I think this is fine.

Spuffynism (approved):
😍

packages/headless/esbuild.mjs
```diff
@@ -87,15 +87,23 @@ const browserEsm = Object.entries(useCaseEntries).map((entry) => {
   const [useCase, entryPoint] = entry;
   const outDir = getUseCaseDir('dist/browser', useCase);
   const outfile = `${outDir}/headless.esm.js`;
+  const config = {
```
olamothe, line 90: nitpicky: I would just modify the `config` used based on `devMode` value instead of calling `buildBrowserConfig` differently with override on this config.

olamothe (approved):

olamothe:
Just small heads up about making sure the commit type is followed as closely as possible in the PR title. (ie: `feat vs fix vs {everything else}`)

It controls the tooling around automatic semver version bump.

packages/headless/esbuild.mjs
BenKandelaars: done

PR #3689, opened 2024-03-11
fix(atomic): add missing end label

louis-bompart:
GitHub copilot translation.


olamothe (approved):
Potential future improvement: Generate all valid keys in a `.d.ts` which augment i18n to fail compilation when referencing a missing key "only for us/for out of the box component" (should not pollute implementer definition for i18n)

```
bindings.i18n.t('should-not-compile');
```

PR #3693, opened 2024-03-11
chore(relay): make relay available in Actions as extra

louis-bompart:
Prototype to handle analytics actions related to state changes, like the purchase action for e.g.

https://coveord.atlassian.net/browse/KIT-3038

### DIFF
--- a/packages/headless/src/app/engine.ts
+++ b/packages/headless/src/app/engine.ts
@@ -32,7 +32,7 @@ import {LoggerOptions} from './logger';
 import {logActionErrorMiddleware} from './logger-middlewares';
 import {createReducerManager, ReducerManager} from './reducer-manager';
 import {createRenewAccessTokenMiddleware} from './renew-access-token-middleware';
-import {Store, configureStore} from './store';
+import {ExtraArgumentsWithRelay, Store, configureStore} from './store';
 import {ThunkExtraArguments} from './thunk-extra-arguments';
 
 const coreReducers = {configuration, version};
@@ -182,7 +182,10 @@ export function buildEngine<
 >(
   options: EngineOptions<Reducers>,
   thunkExtraArguments: ExtraArguments
-): CoreEngine<StateFromReducersMapObject<Reducers>, ExtraArguments> {
+): CoreEngine<
+  StateFromReducersMapObject<Reducers>,
+  ExtraArgumentsWithRelay & ExtraArguments
+> {
   const engine = buildCoreEngine(options, thunkExtraArguments);
   const {accessToken, organizationId} = options.configuration;
   const {organizationEndpoints} = options.configuration;
@@ -245,9 +248,20 @@ function buildCoreEngine<
     reducerManager.addCrossReducer(options.crossReducer);
   }
   const logger = thunkExtraArguments.logger;
-  const store = createStore(options, thunkExtraArguments, reducerManager);
+  const thunkExtraArgumentsWithRelay: ExtraArgumentsWithRelay & ExtraArguments =
+    {
+      ...thunkExtraArguments,
+      get relay() {
+        return getRelayInstanceFromState(engine.state);
+      },
+    };
+  const store = createStore(
+    options,
+    thunkExtraArgumentsWithRelay,
+    reducerManager
+  );
 
-  return {
+  const engine = {
     addReducers(reducers: ReducersMapObject) {
       if (reducerManager.containsAll(reducers)) {
         return;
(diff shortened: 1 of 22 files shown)


packages/headless/src/app/engine.ts
```diff
@@ -245,9 +248,20 @@ function buildCoreEngine<
...
   }
   const logger = thunkExtraArguments.logger;
-  const store = createStore(options, thunkExtraArguments, reducerManager);
+  const thunkExtraArgumentsWithRelay: ExtraArgumentsWithRelay & ExtraArguments =
+    {
+      ...thunkExtraArguments,
+      get relay() {
+        return getRelayInstanceFromState(engine.state);
+      },
+    };
```
louis-bompart, line 257: This is the nasty part.

packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
louis-bompart, line 1: Some refactor++:
 - Moved the itemsSelector to `packages/headless/src/features/commerce/context/cart/cart-selector.ts`: It's now a 'context' concern, not a controller concern.
 - Makes all selector memoized.

packages/headless/src/controllers/commerce/context/cart/headless-cart.test.ts
```diff
@@ -117,7 +119,7 @@ describe('headless commerce cart', () => {
       jest.resetAllMocks();
     });
 
-    it('emits the #ec.purchase event with the expected payload', () => {
+    it.skip('emits the #ec.purchase event with the expected payload', () => {
```
Spuffynism, line 122: 🧐

packages/headless/src/features/commerce/context/cart/cart-selector.ts
```diff
@@ -0,0 +1,32 @@
...
+  (cartState: CartState) => cartState.cart,
+  (cartState: CartState) => cartState.cartItems,
+  (cart, cartItems) => cartItems.map((id: string) => cart[id])
+);
+
+const productQuantitySelector = createSelector(itemsSelector, (items) =>
+  items.map(({quantity, ...product}) => ({
+    quantity,
+    product,
+  }))
```
Spuffynism, line 31: Isn't this the same as returning `items` directly? Or are we creating a deep copy?
```suggestion
  items
```

Spuffynism (approved):
lgtm 👍

packages/headless/src/controllers/commerce/context/cart/headless-cart.test.ts
louis-bompart: Yeah ain't complete chief. But here, we ain't gonna emit the ecPurchase as we mock the action.
I need to UT my action instead.

sallainCoveo (approved):
It seems to me that this POC is promising! 

Should we assume that if we adopt this approach, all EP events should be done with an action?

olamothe (approved):
That's neat !

packages/headless/src/app/engine.ts
```diff
@@ -245,9 +248,20 @@ function buildCoreEngine<
...
+  const store = createStore(
+    options,
+    thunkExtraArgumentsWithRelay,
+    reducerManager
+  );
 
-  return {
+  const engine = {
```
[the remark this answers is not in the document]
louis-bompart: That's why it's the nasty bit ahah.
I do keep a ref to the engine instance in the scope of the function to reuse it in the scope of the relay getter given to the extra arguments of every actions/redux thingies.

By doing it this way, I can make Relay available without having to either:

Have a non-reactive Relay (i.e. let's say we change the trackingId or some other stuff)
Require every action that needs to send analytics to have the analytic section in its reducer

packages/headless/src/features/commerce/context/cart/cart-selector.ts
```diff
@@ -0,0 +1,32 @@
+import {Ec} from '@coveo/relay-event-types';
+import {createSelector} from '@reduxjs/toolkit';
+import {CommerceEngineState} from '../../../../app/commerce-engine/commerce-engine';
+import {getCurrency} from '../context-selector';
+import {CartState} from './cart-state';
+
+export interface Transaction {
```
sallainCoveo, line 7: Nit: just to mention that a [Transaction](https://github.com/coveo/ui-kit/pull/3654/files#diff-64e4398703c6034135ce88296c4b4bb23fcb6c55b19eee73655ef0b228a1383cR55) interface is declared in the headless-cart file at the moment (it might be the same purpose of this one).

PR #3695, opened 2024-03-12
chore(IPX+IP): remove filters top title

dguerinCoveo:
Remove double filters title in IPX and Insight since we don't have sorts for IPX and SPs

|   | Before | After |
| ------------- | ------------- | ------------- |
| IPX  | imagel  | image  |
| Insight  | image  | image  |


pathiery (approved):
DELIVERING VALUE ON A DAILY BASIS @dguerinCoveo 🚀🚀🚀

PR #3696, opened 2024-03-12
feat(quantic): bare bones to start testing the new event protocol in quantic added

mmitiche:
## [SFINT-5424](https://coveord.atlassian.net/browse/SFINT-5424)

in this PR:

- Added the logic that will intercept analytics sent using the event protocol.
- Created the `next-analytics-expectations.ts`, this file will be a home to all the next analytics expectations, created the first expectation as an example.
- Created `analytics-utils.ts` which contain the utility that allow us to enable sending the analytics according to the event protocol.

This is the base to all the PRs that will come after where we will update our test to check the analytics with the new event protocol.


packages/quantic/cypress/page-objects/search.ts
```diff
@@ -90,6 +94,11 @@ export const InterceptAliases = {
     UndoQuery: uaAlias('undoQuery'),
     SearchboxSubmit: uaAlias('searchboxSubmit'),
   },
+  NextAnalytics: {
+    Qna: {
+      AnswerAction: nextAnalayticsAlias('Qna.AnswerAction'),
```
SimonMilord, line 99:
```suggestion
      AnswerAction: nextAnalyticsAlias('Qna.AnswerAction'),
```

packages/quantic/cypress/page-objects/search.ts
```diff
@@ -90,6 +94,11 @@ export const InterceptAliases = {
     UndoQuery: uaAlias('undoQuery'),
     SearchboxSubmit: uaAlias('searchboxSubmit'),
   },
+  NextAnalytics: {
```
SimonMilord, line 97: ok this is where we will add `InterceptAliases` for all the new protocol events? should we create its own alias object instead for clarity or do we need to add it to the existing object of `InterceptAliases` ?

mmitiche: We do want in the same object, cause the logic to intercept queries is handled at the same place so it would make sense to have only one object that this logic will need to look at.

packages/quantic/cypress/e2e/next-analytics-expectations.ts
```diff
@@ -0,0 +1,23 @@
+import {InterceptAliases} from '../page-objects/search';
+
+function nextAnalyticsExpectations() {
+  return {
+    emitQnaAnswerActionEvent: (action: string) => {
+      cy.wait(InterceptAliases.NextAnalytics.QNA.AnswerAction)
```
mmitiche, line 6:
```suggestion
      cy.wait(InterceptAliases.NextAnalytics.Qna.AnswerAction)
```

packages/quantic/cypress/e2e/next-analytics-expectations.ts
```diff
@@ -0,0 +1,23 @@
...
+      cy.wait(InterceptAliases.NextAnalytics.Qna.AnswerAction)
+        .then((interception): void => {
+          const eventBody: Record<string, unknown> =
+            interception?.request?.body?.[0];
+          const eventMeta = eventBody.meta;
+          expect(eventBody).to.have.property('action', action);
+          expect(eventMeta).to.have.property('type', 'Qna.AnswerAction');
+        })
+        .logDetail(
+          `should emit the Qna.AnswerStyle event with action "${action}"`
```
mmitiche, line 15:
```suggestion
          `should emit the Qna.AnswerAction event with action "${action}"`
```

SimonMilord (approved):

PR #3697, opened 2024-03-12
fix: remove trackingId from commerce context

samisayegh:
The commerce engine was storing `trackingId` under commerce context, but the value was already available under the `configuration` slice. This was creating two sources of truth for `trackingId` on the commerce engine. If `trackingId` was set on init of the engine, it would not be used.

This PR resolves the duplication and uses the value on the `configuration` slice.
We're making this breaking change now to get ahead of customer adoption.


Spuffynism (approved):
LGTM! Thank you!

sallainCoveo (approved):
:100:

olamothe (approved):

PR #3700, opened 2024-03-12
chore: disable github-pkg

louis-bompart:
Disabling this step for now, we reenable for future release.

In the latest [releases run](https://github.com/coveo/ui-kit/actions/runs/8254691568/job/22579419419) (and a few before), the release log are mangled for this specific action, for unknown reasons.
Manual runs on a 'full clone' of the repo is required/recommended but will likely take a good chunk of time.
We still want to release to the main channels (NPM/S3) in the meantime, so for now, we disable the GH PKG channel publishing


olamothe (approved):

olamothe:
Can you add an explanation in the PR description so future reader will understand why its commented out.

PR #3701, opened 2024-03-12
chore: update esbuild

olamothe:
* Update esbuild to latest version
    * Bring changes to how watch mode works
* Remove our "homebrew" umd wrapper and replace it with a plugin
    * For v3, we can remove umd output completely and just rely on IIFE
* Remove "ignore esbuild" exception from renovate config

https://coveord.atlassian.net/browse/KIT-2676


package.json
```diff
@@ -49,8 +49,9 @@
     "cspell": "8.6.0",
     "cz-conventional-changelog": "3.3.0",
     "detect-indent": "7.0.1",
-    "esbuild": "0.14.2",
+    "esbuild": "0.20.1",
```
fbeaudoincoveo, line 52: 0.20.2 was published a few days ago, so we might as well upgrade to that version

https://github.com/evanw/esbuild/releases/tag/v0.20.2

fbeaudoincoveo (approved):

package.json
olamothe: Good catch !

PR #3702, opened 2024-03-12
feat(atomic): style rephrase buttons and add auto option

jelmedini:
[SVCC-3510](https://coveord.atlassian.net/browse/SVCC-3510)

Change Rephrasing Buttons styles to match the new mockups and Add auto option selected by default.

**Before:**
Generated snippet - v2

**After:**

Screenshot 2024-03-13 at 9 41 26 AM

**Mobile layout:**

Screenshot 2024-03-12 at 5 28 40 PM

**Very small Mobile Layout:**

Screenshot 2024-03-12 at 5 29 04 PM

**Demo:**

https://github.com/coveo/ui-kit/assets/119955059/1b40b5ad-998b-423f-9f12-9503b5aa4193

[SVCC-3510]: https://coveord.atlassian.net/browse/SVCC-3510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


louis-bompart (approved):

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
```diff
@@ -203,6 +203,16 @@ export class GeneratedAnswerCommon {
     this.props.getGeneratedAnswer()?.dislike();
   };
 
+  private onChangeAnswerStyle = (answerStyle: GeneratedAnswerStyle) => {
```
olamothe, line 206:
```suggestion
  private onChangeAnswerStyle(answerStyle: GeneratedAnswerStyle) {
```

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
```diff
@@ -20,7 +20,19 @@
   }
 
   [part='rephrase-button'] {
-    @apply rephrase-with-text;
+    @apply rephrase-with-text flex-auto justify-center;
```
olamothe, line 23: `flex-auto justify-center` can be set as `style` directly in JSX

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
olamothe: And could you modify the other methods in this file so they follow the same format ? Thank you !

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
```diff
@@ -20,7 +20,19 @@
...
+
+    .icon-container {
+      @apply mx-0;
```
olamothe, line 26: Same as above

packages/atomic/src/components/common/generated-answer/styles/rephrase-buttons.pcss
```diff
@@ -36,11 +40,23 @@
...
   &.active {
-    @apply border-primary text-primary;
-
     .icon-container,
     .rephrase-btn-label {
       @apply text-inherit;
     }
   }
+
+  &:hover {
```
olamothe, line 49: All of those can be added directly in JSX as well (I think)

olamothe (changes requested):
Make sure that the styling is applied in JSX by default, and to use the `pcss` file only when it's absolutely impossible to use tailwind classes directly inline (ie: need to apply some mixin).

This is very important for consistency's sake. Thanks !

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
jelmedini: @olamothe this is specific for mobile layout that s why they are declared here

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
jelmedini: same here defined for mixin mobile layout

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
```diff
@@ -193,15 +193,25 @@ export class GeneratedAnswerCommon {
...
+  }
+
+  private onChangeAnswerStyle(answerStyle: GeneratedAnswerStyle) {
+    if (
+      this.props.getGeneratedAnswerState()?.responseFormat.answerStyle ===
+      answerStyle
+    ) {
+      return;
+    }
+    this.props.getGeneratedAnswer()?.rephrase({answerStyle});
```
lbergeron, line 213: Id suggest reversing the condition so we don't need the implicit "else" clause.

```suggestion
    if (
      this.props.getGeneratedAnswerState()?.responseFormat.answerStyle !==
      answerStyle
    ) {
      this.props.getGeneratedAnswer()?.rephrase({answerStyle});
    }
```

packages/atomic/src/components/common/generated-answer/styles/rephrase-buttons.pcss
```diff
@@ -9,10 +9,14 @@
...
   }
+
+  &.active .rephrase-btn-label {
+    @apply text-white;
+  }
 }
 
 [part='rephrase-button'] {
-  --height: 2.2em;
+  --height: 21px;
```
lbergeron, line 19: The previous height was relative to the font size just like the icon height. Did you try increasing the font size in the browser? Does it render properly?

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
```diff
@@ -296,9 +306,7 @@ export class GeneratedAnswerCommon {
                     .answerStyle ?? 'default'
                 }
                 i18n={this.props.getBindings().i18n}
-                onChange={(answerStyle: GeneratedAnswerStyle) =>
-                  this.props.getGeneratedAnswer()?.rephrase({answerStyle})
-                }
+                onChange={this.onChangeAnswerStyle}
```
lbergeron, line 309: I tried running the code locally, and clicking the rephrase buttons fail with this error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'getGeneratedAnswerState')
```

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
```diff
@@ -193,15 +193,25 @@ export class GeneratedAnswerCommon {
       });
   }
 
-  private clickDislike = () => {
+  private clickDislike() {
```
lbergeron, line 196: I tried running the code locally, and clicking the Dislike button fails with this error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'getGeneratedAnswer')
```

lbergeron (changes requested):
I am getting errors when clicking the dislike and rephrase buttons.

jelmedini:
> I am getting errors when clicking the dislike and rephrase buttons.

@lbergeron Ive change method format as suggested in the review but the method calling needed to be also updated in this case, it's fixed now 🙈

packages/atomic/src/components/common/generated-answer/styles/rephrase-buttons.pcss
lbergeron: My point is that `1em` typically equals `16px`. It means an icon size of `1.1em` is almost `18px`. Things change when the end user decides to scale up the font size to 150% for instance. In this case `1em` becomes `24px` and `1.1em` becomes `26px`. At some point, an icon that is `1.1em` tall might not fit in a `21px` (fixed size) container. Usually we try and use relative dimensions everywhere so it all scale properly according to the font size.

jelmedini: hmm I see, It renders properly when tested increasing the font size, let me know if there is something i didn' catch 🙏

lbergeron: I tested by changing the `Font size` browser setting instead of `Page zoom`.

jelmedini: @lbergeron  fixed also for citations as discussed :) Thanks for raising this point! 
Large:
Screenshot 2024-03-14 at 2 09 38 PM

Small:
Screenshot 2024-03-14 at 2 06 54 PM

lbergeron (approved):

PR #3708, opened 2024-03-13
chore: ensure command stands on a singleline

louis-bompart:
Each line was interpreted independently of the other. Oops.

This concat the script, and ensures we have a single whitespace every time


louis-bompart:
Tested thru #3707

olamothe (approved):

PR #3716, opened 2024-03-14
test(quantic): testing the analytics new event protocol with RGA components

mmitiche:
## [SFINT-5399](https://coveord.atlassian.net/browse/SFINT-5399)

In this PR: 

- Adapted the Cypress e2e test for the RGA feature to testing for analytics sent with the new protocol, now the test spec tests for both case: the case when send legacy analytics and the case where we sent next analytics, following this structure:
Screenshot

- Created `next-analytics-expectations.ts` which contains the functions to verify the events sent with new event protocol.
```
Expect.logDislikeGeneratedAnswer(streamId);
```
```
if (analyticsMode === "next") {
  NextAnalyticsExpectations.emitQnaDislikeEvent(
    {
      feedback: {
        liked: false,
      },
      answer: {
        id: streamId,
        type: "CRGA",
      },
    },
    exampleTrackingId
  );
} else {
  Expect.logDislikeGeneratedAnswer(streamId);
}
```

#### PS:
This is just the first PR to introduce these changes that will make the E2E Quantic tests test for the analytics sent with the event protocol, similar PRs will follow soon for other features.

[SFINT-5399]: https://coveord.atlassian.net/browse/SFINT-5399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/cypress/page-objects/search.ts
```diff
@@ -90,6 +94,18 @@ export const InterceptAliases = {
     UndoQuery: uaAlias('undoQuery'),
     SearchboxSubmit: uaAlias('searchboxSubmit'),
   },
+  NextAnalytics: {
+    Qna: {
+      AnswerAction: nextAnalyticsAlias('Qna.AnswerAction'),
```
SimonMilord, line 99: does AnswerAction encompasses everything else for Qna? like RetryGeneratedAnswer, ShowGeneratedAnswer, HideGeneratedAnswer, etc. ?

packages/quantic/cypress/e2e/next-analytics-expectations.ts
```diff
@@ -0,0 +1,144 @@
...
+            expectedEvent,
+            expectedTrackingId
+          );
+        })
+        .logDetail(
+          'should emit the Qna.SubmitFeedback event for the dislike action'
+        );
+    },
+
+    emitQnaSubmitNegativeFeedbackEvent: (
```
SimonMilord, line 123:
```suggestion
    emitQnaSubmitFeedbackEvent: (
```

Even though feedback is mostly submitted when there is a dislike, I think this name would match the name of the event better no?  (ie: Qna.SubmitFeedback)

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -87,582 +97,842 @@ describe('quantic-generated-answer', () => {
...
-                Actions.dislikeGeneratedAnswer();
-                Expect.displayFeedbackModal(false);
-              }
-            );
-          }
-        );
-      });
+              it('should log the stream ID in the search event custom data', () => {
+                if (analyticsMode === 'legacy') {
+                  Expect.logStreamIdInAnalytics(streamId, param.useCase);
```
SimonMilord, line 130: ok this is one of the search events no longer sent correct?

mmitiche: exactly 👍

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -87,582 +97,842 @@ describe('quantic-generated-answer', () => {
...
+                  });
+
+                  scope('when selecting a feedback option', () => {
+                    Actions.dislikeGeneratedAnswer();
+                    Actions.clickFeedbackOption(
+                      feedbackOptions.indexOf(irrelevantOption)
+                    );
+                    Actions.clickFeedbackSubmitButton();
+                    if (analyticsMode === 'next') {
+                      NextAnalyticsExpectations.emitQnaSubmitNegativeFeedbackEvent(
```
SimonMilord, line 233:
```suggestion
                      NextAnalyticsExpectations.emitQnaSubmitFeedbackEvent(
```

same here, everywhere else its `SubmitFeedback` so it would remove the negative, but its up to you

packages/quantic/cypress/e2e/next-analytics-expectations.ts
SimonMilord: I see, then maybe something like: emitQnaSubmitFeedbackDetailsEvent or emitQnaSubmitFeedbackReasonEvent ? anyways I'll let you decide what is best here this is not a game breaking change in the first place.

mmitiche:
> Very nice PR description, good work! 👌
> 
> Small question will the legacy events be supported indefinitely as deprecated or legacy code or is there a plan to remove these once the migration is fully completed?

Thank you!
No they will not be supported indefinitely, we will only support them during the transaction to the v3.
Once the v3 is here we will be able to clean our tests by removing anything related to legacy analytics.

packages/quantic/cypress/e2e/next-analytics-expectations.ts
mmitiche: good idea `emitQnaSubmitFeedbackReasonEvent ` represent better this function 👍

PR #3719, opened 2024-03-14
feat(interactive-result): allow sending ecproduct click event

sallainCoveo:
[LENS-1595](https://coveord.atlassian.net/browse/LENS-1595)

Draft solution: dispatch an action to send the productClick event. The main challenge is to make sure we are getting the right information based on the solution type. 

Current solution is to check if the `responseId` and `products` are defined in the `productListing` state. If that's the case, we get these parameters from that state. Otherwise, we get them from the `commerSearch` state.

[LENS-1595]: https://coveord.atlassian.net/browse/LENS-1595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/features/commerce/interactive-result/interactive-result-selectors.ts
+++ b/packages/headless/src/features/commerce/interactive-result/interactive-result-selectors.ts
@@ -1,0 +1,58 @@
+import {Ec, Product} from '@coveo/relay-event-types';
+import {createSelector} from '@reduxjs/toolkit';
+import {ProductRecommendation} from '../../../api/search/search/product-recommendation';
+import {CommerceEngineState} from '../../../app/commerce-engine/commerce-engine';
+import {getCurrency} from '../context/context-selector';
+import {ProductListingV2State} from '../product-listing/product-listing-state';
+import {CommerceSearchState} from '../search/search-state';
+
+export const getECProductClickPayload = (
+  product: ProductRecommendation,
+  state: CommerceEngineState
+): Ec.ProductClick => ({
+  product: formatProduct(product),
+  responseId: getResponseId({...state.productListing, ...state.commerceSearch}),
+  currency: getCurrency(state.commerceContext),
+  position: getPosition(product, state),
+});
+
+const formatProduct = (product: ProductRecommendation): Product => ({
+  productId: product.permanentid,
+  name: product.ec_name ?? '',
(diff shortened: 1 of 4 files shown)


packages/headless/src/features/commerce/interactive-result/interactive-result-selectors.ts
```diff
@@ -0,0 +1,58 @@
...
+import {getCurrency} from '../context/context-selector';
+import {ProductListingV2State} from '../product-listing/product-listing-state';
+import {CommerceSearchState} from '../search/search-state';
+
+export const getECProductClickPayload = (
+  product: ProductRecommendation,
+  state: CommerceEngineState
+): Ec.ProductClick => ({
+  product: formatProduct(product),
+  responseId: getResponseId({...state.productListing, ...state.commerceSearch}),
```
Spuffynism, line 14: When the RFC around multiple solution types resolves, we should have a way to create an interactive result controller from its parent controller. This means we'll only have access to the state from the solution type in question, which will make finding the `responseId` easy-peasy.

However, if we want a single interactive result controller that can be created without a reference to its parent, I don't think this approach will work. For example, let's take a listing page with a recs carroussel as example:
1. User visits page
2. Listing is received with `responseId` = `listing-response-id`
3. Recs are received with `responseId` = `recs-response-id`
4. User clicks on listing product, which would use `recs-response-id` 🔴

If we really want a single controller, we have to put the "control" in the hands of the implementer:
1. We let them use the state themselves ([as you proposed!](https://coveord.atlassian.net/browse/LENS-1595))
2. Or, we create multiple controllers

samisayegh: Let's put this jira on hold until the RFC is ready? Is there a jira we can link to in the CAPI project?

Spuffynism: Yes! https://coveord.atlassian.net/browse/CAPI-496

sallainCoveo: Thanks for your inputs! I will close the PR for now, and will revisit after the RFC is approved!

PR #3727, opened 2024-03-18
chore(quantic): skip authorize outside CI

louis-bompart:
Following @mmitiche advice, but automatedish

https://coveord.atlassian.net/browse/KIT-3044

### DIFF
--- a/packages/quantic/scripts/build/deploy-community.ts
+++ b/packages/quantic/scripts/build/deploy-community.ts
@@ -312,7 +312,9 @@ async function deleteScratchOrg(
 
   try {
     let communityUrl = '';
-    runner.add(async (log) => await authorizeDevOrg(log, options));
+    if (process.env.CI) {
+      runner.add(async (log) => await authorizeDevOrg(log, options));
+    }
     if (options.deleteOldOrgs) {
       runner.add(async (log) => {
         await deleteOldOrgs(log, options);
--- a/packages/quantic/scripts/build/deploy-community.ts
+++ b/packages/quantic/scripts/build/deploy-community.ts
@@ -320,9 +320,6 @@ async function deleteScratchOrg(
         await deleteOldOrgs(log, options);
       });
     }
-    if (isCi()) {
-      runner.add(async (log) => await authorizeDevOrg(log, options));
-    }
     runner
       .add(async (log) => {
         await ensureScratchOrgExists(log, options);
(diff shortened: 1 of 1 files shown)


packages/quantic/scripts/build/deploy-community.ts
```diff
@@ -312,7 +312,9 @@ async function deleteScratchOrg(
 
   try {
     let communityUrl = '';
-    runner.add(async (log) => await authorizeDevOrg(log, options));
+    if (process.env.CI) {
+      runner.add(async (log) => await authorizeDevOrg(log, options));
+    }
```
mmitiche, line 317: in line `323` we re do the same step:
```
runner.add(async (log) => await authorizeDevOrg(log, options));
```
I don't think it's necessary, is it?

packages/quantic/scripts/build/deploy-community.ts
```diff
@@ -312,7 +312,9 @@ async function deleteScratchOrg(
 
   try {
     let communityUrl = '';
-    runner.add(async (log) => await authorizeDevOrg(log, options));
+    if (process.env.CI) {
```
mmitiche, line 315: Kinda the same thing but  in this file we already have a function called `isCi`.
```suggestion
    if (isCi()) {
```

packages/quantic/scripts/build/deploy-community.ts
louis-bompart: Yeah same you're right.

packages/quantic/scripts/build/deploy-community.ts
```diff
@@ -312,13 +312,17 @@ async function deleteScratchOrg(
...
+    }
     if (options.deleteOldOrgs) {
       runner.add(async (log) => {
         await deleteOldOrgs(log, options);
       });
     }
-    runner.add(async (log) => await authorizeDevOrg(log, options));
+    if (isCi()) {
+      runner.add(async (log) => await authorizeDevOrg(log, options));
+    }
```
mmitiche, line 325: I mean is authorizing the dev org  the for the second time really necessary?
```suggestion
```

mmitiche (approved):

packages/quantic/scripts/build/deploy-community.ts
louis-bompart: I reckon yes because deleteOldOrgs do connect to other org I reckon.

louis-bompart: Could try without tho, yolo.
If it works gg

PR #3729, opened 2024-03-20
fix(commerce): replace barrel import with proper import

Spuffynism:
As part of CAPI-496, we compared bundle sizes for different changes we wanted to make to headless and spotted issues with imports in the commerce package.

Running dependency cruiser to map a graph of dependencies in the `packages/headless` folder showed that `index.ts` was being included ([full dependencies.svg file here](https://github.com/coveo/test-headless-tree-shaking/blob/main/bundler-comparison/dependencies.svg)):
```shell
npx depcruise --no-config \
  --exclude "^node_modules" \
  --output-type dot \
  src/commerce.index.ts > dependencies.out && dot dependencies.out -T svg -o dependencies.svg
```
image

We then found the source of the issue was `product-listing-recent-results` with:
```shell
npx depcruise --no-config \
  --exclude "^node_modules" \
  --output-type dot \
  --ts-config tsconfig.json \
  --do-not-follow "^src/index.ts" \
  --focus "^src/index.ts" \
  src/commerce.index.ts > dependencies.out && dot dependencies.out -T svg -o dependencies.svg
```
image

This PR replace the `index.ts` barrel import with the proper redux import.

### DIFF
--- a/packages/headless/src/features/product-listing/product-listing-recent-results.ts
+++ b/packages/headless/src/features/product-listing/product-listing-recent-results.ts
@@ -1,4 +1,4 @@
-import {createAction} from '../..';
+import {createAction} from '@reduxjs/toolkit';
 import {validateProductRecommendationPayload} from '../analytics/analytics-utils';
 import {ProductRecommendation} from './../../api/search/search/product-recommendation';
 


Spuffynism:
We still depend on a few bits and pieces that we could not depend on (slices, actions, etc.), but I don't think it's worth it to invest some more time in this at this time. Nonetheless, here they are:
- did you mean
- trigger
- folded result list
- breadcrumb
- history
- static-filter-set
- question-answering
- dictionary-field-context
- context
- search-hub
- search-and-folding

If one ever wishes to reduce the bundle size for commerce, these would be good starting points. Using the interactive mode with dependency cruiser can help identify these issues ([interactive.html](https://github.com/coveo/test-headless-tree-shaking/blob/main/bundler-comparison/interactive.html)):
```shell
npx depcruise --no-config \
  --exclude "^node_modules" \
  --output-type dot \
  --ts-config tsconfig.json \
  src/commerce.index.ts | dot -T svg -Grankdir=TD | npx depcruise-wrap-stream-in-html > interactive.html && open interactive.html
  ```

fbeaudoincoveo (approved):

PR #3731, opened 2024-03-20
fix(quantic): avoid pushing to recent results when use case is not search

mmitiche:
### [SFINT-5456](https://coveord.atlassian.net/browse/SFINT-5456)

The recent results feature is not yet supported for the insight use case, I consider this as a temporary fix.
A real improvement to improve this would be the following:

- Make the necessary updates in the Headless library to support this feature in the insight use case, we should also make the feature about the recent queries list available for the insight use case.
- Improve the push recent action in headless v3, for now it takes the whole result as a parameter, while just sending the unique id should be enough to find the result in the Headless state.

#### BEFORE

https://github.com/coveo/ui-kit/assets/86681870/1d9dbd5f-ba2e-496f-b1cb-caee4d416708

#### AFTER

https://github.com/coveo/ui-kit/assets/86681870/cdea0d77-5d92-4739-85e1-9ba396b7e0ce

[SFINT-5456]: https://coveord.atlassian.net/browse/SFINT-5456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticResultQuickview/__tests__/quanticResultQuickview.test.js
```diff
@@ -0,0 +1,171 @@
+import * as mockHeadlessLoader from 'c/quanticHeadlessLoader';
+// @ts-ignore
```
SimonMilord, line 2:
```suggestion
```

mmitiche: We do need it cause createElement is not recognized.

SimonMilord (approved):

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/__tests__/quanticResultQuickview.test.js
SimonMilord: I tought etienne brought it up with salesforce as a bug, it will eventually be fixed but in the meantime (at least on my end it does not seem to prevent the tests from running correctly.

Also dont know about you, but even with the `ts-ignore`, I still get the red line (even restarted the ts server in case) 
image

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/__tests__/quanticResultQuickview.test.js
```diff
@@ -0,0 +1,171 @@
...
+      expect(functionMocks.pushRecentResult).toHaveBeenCalled();
+    });
+  });
+
+  describe('when the component is not used in the search use case', () => {
+    beforeAll(() => {
+      useCase = 'insight';
+    });
+
+    it('should dispatch the pushRecentResult action', async () => {
```
erocheleau, line 158:
```suggestion
    it('should not dispatch the pushRecentResult action', async () => {
```

erocheleau (approved):

PR #3734, opened 2024-03-20
feat(commerce): commerce api recommendations controller

BenKandelaars:
Add a recommendations controller to headless using Commerce API recommendations endpoint. The controller uses the same architecture as the listings controller which also uses the Commerce API.

[CAPI-169](https://coveord.atlassian.net/jira/software/c/projects/CAPI/boards/1103?selectedIssue=CAPI-169)

[CAPI-169]: https://coveord.atlassian.net/browse/CAPI-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/api/commerce/commerce-api-params.ts
```diff
@@ -84,3 +84,7 @@ export interface FacetQueryParam {
 export interface FacetIdParam {
   facetId: string;
 }
+
+export interface SlotId {
```
olamothe, line 88: For uniformity:

```suggestion
export interface SlotIdParam {
```

packages/headless/src/api/commerce/common/response.ts
```diff
@@ -7,11 +7,18 @@ import {ProductRecommendation} from '../../search/search/product-recommendation'
 import {Pagination} from './pagination';
 import {Sort} from './sort';
 
+export interface RecommendationCommerceSuccessResponse {
+  responseId: string;
+  headline: string;
+  products: ProductRecommendation[];
```
olamothe, line 13: Depending on the outcome of this PR / timing when it gets merged, there will be a conflict on this precise type: 

https://github.com/coveo/ui-kit/pull/3738

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export interface RecommendationState {
+  headline: string;
+  products: ProductRecommendation[];
+  error: CommerceAPIErrorStatusResponse | null;
+  isLoading: boolean;
+  responseId: string;
+}
+
+export type RecommendationControllerState = Recommendation['state'];
```
olamothe, line 40: This should not be a necessary type to create nor export globally.

Just `RecommendationState` should be fine.

packages/headless/src/test/mock-recommendation-v2.ts
```diff
@@ -0,0 +1,20 @@
+import {RecommendationCommerceSuccessResponse} from '../api/commerce/common/response';
+import {QueryRecommendationCommerceAPIThunkReturn} from '../features/commerce/common/actions';
+
+export function buildFetchRecommendationV2Response(
```
olamothe, line 4: `Mock` would be a better name to standardize with the other functions in the `/test` folder.

```suggestion
export function buildMockRecommendationV2Response(
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Creates a `ProductListing` controller instance.
+ *
+ * @param engine - The headless commerce engine.
+ * @returns A `ProductListing` controller instance.
+ */
+export function buildRecommendations(
```
olamothe, line 48: All other controller follow the same format, and so this one must absolutely follow the same.

First parameter is always the engine, and the second one is always a `prop` , with an `options` property.

And, on `options`, you can pass parameters.

Something like this:

```typescript
interface RecommendationsOptions {
  slotId: string;
}

interface RecommendationsProps {
    options: RecommendationsOptions;
}

buildRecommendations(engine: CommerceEngine, props: RecommendationsProps) {
   // [...]
}
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,91 @@
...
+  slotId: string,
+  engine: CommerceEngine
+): Recommendation {
+  if (slotId === '') {
```
olamothe, line 52: Since this gets validated when dispatching the `updateRecommendationSlotId`, this is un-necessary, and non standard.

ie: on line 62, if the string is empty, the validation function inside `updateRecommendationSlotId` will throw and a standard error message will appear.

If you truly needed to do props/options validation in the controller that would not be caught downstream, the proper way would be to use `bueno`, that is used throughout the project.

packages/headless/src/utils/errors.ts
```diff
@@ -17,3 +17,7 @@ export class DisconnectedError extends Error {
     this.statusCode = statusCode ?? 0;
   }
 }
+
+export const slotIdError = new Error(
```
olamothe, line 21: Related to above comment about validation, this should be removed.

olamothe (changes requested):

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  if (!loadBaseRecommendationReducers(engine)) {
+    throw loadReducerError;
+  }
+
+  const controller = buildController(engine);
+  const {dispatch} = engine;
+  dispatch(updateRecommendationSlotId({slotId}));
+
+  const getState = () => engine.state;
```
fbeaudoincoveo, line 64: One thing we've started doing recently to take advantage of memoization is to use the `createSelector` function from @redux/toolkit.

Here, you could replace `const getState = () => engine.state` with:

```typescript
const recommendationStateSelector = createSelector(
  (state: CommerceEngineState) => state.recommendation,
  (recommendation) => ({
    headline: recommendation.headline,
    products: recommendation.products,
    error: recommendation.error,
    isLoading: recommendation.isLoading,
    responseId: recommendation.responseId,
  })
);
```

and then define the `get state()`  that you return as follows:

```typescript
get state() {
  return recommendationStateSelector(engine.state);
},
```

packages/headless/src/features/commerce/recommendation/recommendation-state.ts
```diff
@@ -0,0 +1,22 @@
+import {CommerceAPIErrorStatusResponse} from '../../../api/commerce/commerce-api-error-response';
```
fbeaudoincoveo, line 1: Ultimately we will want to support 0-* concurrent recommendation interfaces in the Commerce engine state so that multiple carousels can be displayed on the same page without having to initialize multiple engines (I think we want to avoid that at all cost).

This implies that the recommendation state should probably be defined as a set (e.g., `Record<string, RecommendationSlice>`). The record key could potentially be the recommendation's slot ID (assuming that it is intended as a unique identifier).

When calling the Recommendation controller builder, the recommendation interface would be "registered" in the state, and its state selector would point to `engine.state.recommendation[slotId]` (this is similar to what we do for facets, for instance).

The trickier part is pagination. Right now, the pagination state is "global" in the Commerce engine, which hasn't been an issue so far because we only ever display one product listing or search interface at a time. In order to support multiple concurrent paginated interfaces in the same engine state, however, the Recommendation, ProductListing, and Search controllers should each have their own Pagination controller / state.

The approach proposed in https://coveord.atlassian.net/wiki/spaces/COM/pages/3907256341/Commerce+Headless+Routing is to have those "parent" controllers return builders for the sub-controller that they are likely to require.

Should we do this right now, or in an ulterior phase?

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
BenKandelaars: We export this for listings as well. @Spuffynism?

packages/headless/src/features/commerce/recommendation/recommendation-actions.ts
```diff
@@ -0,0 +1,59 @@
...
+  QueryRecommendationCommerceAPIThunkReturn,
+  StateNeededByQueryCommerceAPI,
+} from '../common/actions';
+
+const buildRecommendationCommerceAPIRequest = async (
+  slotId: string,
+  state: StateNeededByQueryCommerceAPI
+): Promise<CommerceRecommendationRequest> => {
+  const commerceAPIRequest = await buildCommerceAPIRequest(state);
+  delete commerceAPIRequest.facets;
```
BenKandelaars, line 18: @Spuffynism could we refactor that `buildCommerceAPIRequest` so facets are optional, like that case of sort?

packages/headless/src/features/commerce/recommendation/recommendation-actions.ts
```diff
@@ -0,0 +1,46 @@
...
+import {isErrorResponse} from '../../../api/search/search-api-client';
+import {nonEmptyString, validatePayload} from '../../../utils/validate-payload';
+import {logQueryError} from '../../search/search-analytics-actions';
+import {
+  buildRecommendationCommerceAPIRequest,
+  QueryRecommendationCommerceAPIThunkReturn,
+  StateNeededByQueryCommerceAPI,
+} from '../common/actions';
+
+export const fetchRecommendation = createAsyncThunk<
```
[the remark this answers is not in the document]
BenKandelaars: I go back and forth on this issue. I don't think it so important

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @param engine - The headless commerce engine.
+ * @returns A `ProductListing` controller instance.
+ */
+export function buildRecommendations(
+  slotId: string,
+  engine: CommerceEngine
+): Recommendation {
+  if (slotId === '') {
+    throw slotIdError;
+  }
```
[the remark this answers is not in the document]
BenKandelaars: I've had a go at adding this

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * A scoped and simplified part of the headless state that is relevant to the `Recommendation` controller.
+   */
+  state: RecommendationState;
+}
+
+export interface RecommendationState {
+  headline: string;
+  products: Product[];
```
BenKandelaars, line 41: @Spuffynism should I use this type or `ProductRecommendation`?

packages/headless/src/features/commerce/recommendation/recommendation-actions.ts
Spuffynism: Yeah, good idea! I'd prefer that to `delete`ing the facets here.

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
Spuffynism: `Product` is the right one, since it's the type for products as returned from the commerce api.

packages/headless/src/features/commerce/recommendation/recommendation-state.ts
Spuffynism: I would propose we do it now, but in a second PR. Doing this now would not be a wasted effort since we will need to write that logic to support different state parts for a controller anyway.

packages/headless/src/features/commerce/recommendation/recommendation-slice.ts
```diff
@@ -0,0 +1,32 @@
+import {createReducer} from '@reduxjs/toolkit';
+import {
+  fetchRecommendations,
+  updateRecommendationSlotId,
+} from './recommendation-actions';
+import {getRecommendationV2InitialState} from './recommendation-state';
+
+export const recommendationV2Reducer = createReducer(
```
Spuffynism, line 8:
```suggestion
export const recommendationReducer = createReducer(
```

packages/headless/src/api/commerce/common/response.ts
```diff
@@ -7,6 +7,13 @@ import {Pagination} from './pagination';
 import {Product} from './product';
 import {Sort} from './sort';
 
+export interface RecommendationCommerceSuccessResponse {
```
Spuffynism, line 10: Since `responseId`, `products` and `pagination` are common to both success responses defined in this file, we could extract them to a `BaseCommerceSuccessResponse` interface only used in this file, and declare `RecommendationCommerceSuccessResponse` and `CommerceSuccessResponse` from it.

packages/headless/src/api/commerce/recommendation/recommendation-request.ts
```diff
@@ -0,0 +1,52 @@
...
+  CurrencyParam,
+  LanguageParam,
+  PageParam,
+  QueryParam,
+  SlotIdParam,
+  TrackingIdParam,
+} from '../commerce-api-params';
+import {CommerceAPIRequest, baseRequest} from '../common/request';
+
+export type CommerceSearchRequest = CommerceAPIRequest & QueryParam;
```
Spuffynism, line 15: I don't think we need this?
```suggestion
```

packages/headless/src/api/commerce/recommendation/recommendation-request.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export type CommerceRecommendationRequest = BaseParam &
+  SlotIdParam &
+  TrackingIdParam &
+  LanguageParam &
+  CountryParam &
+  CurrencyParam &
+  ClientIdParam &
+  ContextParam &
+  PageParam;
```
Spuffynism, line 25: These parameters are all common to commerce requests. Can we extract them to an interface that can be used by other commerce requests? 👀

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
Spuffynism: I think that's a mistake in the commerce sub-package, because I can't find us doing this outside it. Let's remove it.
image

See: https://github.com/search?q=repo%3Acoveo%2Fui-kit+%22ControllerState%22+path%3A%2F%5Epackages%5C%2Fheadless%5C%2F%2F&type=code

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,118 @@
...
+import {validateInitialState} from '../../../utils/validate-payload';
+import {
+  buildController,
+  Controller,
+} from '../../controller/headless-controller';
+
+/**
+ * The `Recommendations` controller exposes a method for retrieving recomendations content in a commerce interface.
+ */
+export interface Recommendation extends Controller {
```
Spuffynism, line 27: Whatever we settle on, we should be consistent
```suggestion
 * The `Recommendation` controller exposes a method for retrieving recomendations content in a commerce interface.
 */
export interface Recommendation extends Controller {
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,118 @@
...
+  slotId: string;
+}
+
+interface RecommendationsProps {
+  options: RecommendationsOptions;
+}
+
+function validateRecommendationInitialState(
+  engine: CommerceEngine,
+  state: RecommendationsOptions
```
Spuffynism, line 59:
```suggestion
  options: RecommendationsOptions
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,118 @@
...
+) {
+  if (!state) {
+    return;
+  }
```
Spuffynism, line 63: `state` can't be falsy as `state` is `props.options`, which is required upon controller initialization.
```suggestion
```

packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -14,6 +17,7 @@ import {
   FacetSection,
   NumericFacetSection,
   ProductListingV2Section,
+  RecommendationV2Section,
```
Spuffynism, line 20:
```suggestion
  RecommendationSection,
```

packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -41,6 +46,11 @@ export interface QueryCommerceAPIThunkReturn {
   analyticsAction: PreparableAnalyticsAction<StateNeededByQueryCommerceAPI>;
 }
 
+export interface QueryRecommendationCommerceAPIThunkReturn {
+  /** The successful search response. */
```
Spuffynism, line 50: comment + this should be moved to `recommendations-actions.ts`
```suggestion
  /** The successful recommendations response. */
```

packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -22,6 +26,7 @@ import {SortBy, SortCriterion} from '../sort/sort';
 
 export type StateNeededByQueryCommerceAPI = ConfigurationSection &
   ProductListingV2Section &
+  RecommendationV2Section &
```
Spuffynism, line 29: Instead of changing this `StateNeededByQueryCommerceAPI` type, we could extend it specifically for recs. Here's an example of what I mean with facet search:
https://github.com/coveo/ui-kit/blob/2f88117c593eab09c92fc07fa2331c5e49f9a192/packages/headless/src/features/commerce/facets/facet-search-set/commerce-facet-search-state.ts#L7-L9

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  const schema = new Schema<RecommendationsOptions>({
+    slotId: recommendationSlotDefinition,
+  });
```
Spuffynism, line 67: This schema definition can be moved to [features/commerce/recommendation/recommendation.ts](https://github.com/coveo/ui-kit/pull/3734/files#diff-6f834abc4da1172ac499c73f2e9334b33f60943e27dafc68155da2b018362ff5) and reused in the payload validation in `updateRecommendationsSlotId`: [features/commerce/recommendation/recommendation-actions.ts](https://github.com/coveo/ui-kit/pull/3734/files#diff-1a9264e1774a126d8783aa61a25d80bf628f26b77acdcf58dbdd40cb4cf21b5c)

packages/headless/src/features/commerce/recommendation/recommendation-slice.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    state.error = err;
+
+    const response = buildMockRecommendationV2Response();
+
+    const action = fetchRecommendations.fulfilled(response, '');
+    const finalState = recommendationV2Reducer(state, action);
+    expect(finalState.error).toBeNull();
+  });
+
+  it('set the isLoading state to true during fetchProductListing.pending', () => {
```
Spuffynism, line 64:
```suggestion
  it('set the isLoading state to true during fetchRecommendations.pending', () => {
```

packages/headless/src/features/commerce/recommendation/recommendation-slice.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    state = getRecommendationV2InitialState();
+  });
+
+  it('should have an initial state', () => {
+    expect(recommendationV2Reducer(undefined, {type: 'foo'})).toEqual(
+      getRecommendationV2InitialState()
+    );
+  });
+
+  it('when a fetchProductListing fulfilled is received, should set the state to the received payload', () => {
```
Spuffynism, line 22:
```suggestion
  it('when a fetchRecommendations.fulfilled is received, should set the state to the received payload', () => {
```

packages/headless/src/features/commerce/common/actions.ts
BenKandelaars: We are not keeping sync with listings by dropping that

packages/headless/src/features/commerce/recommendation/recommendation-slice.ts
BenKandelaars: This is keeping consistency with listings on naming

packages/headless/src/api/commerce/common/response.ts
```diff
@@ -7,11 +7,19 @@ import {Pagination} from './pagination';
...
   responseId: string;
   products: Product[];
-  facets: AnyFacetResponse[];
   pagination: Pagination;
+}
+
+export interface RecommendationsCommerceSuccessResponse
+  extends BaseCommerceSuccessResponse {
+  headline: string;
+}
```
Spuffynism, line 19: This type should reside somewhere like `api/commerce/recommendations/response.ts`
https://github.com/coveo/ui-kit/tree/c71140b51b118e9516aa8e97316ecf0801ac9827/packages/headless/src/api/commerce/recommendations

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,116 @@
...
+import {recommendationsReducer as recommendations} from '../../../features/commerce/recommendations/recommendations-slice';
+import {loadReducerError} from '../../../utils/errors';
+import {validateInitialState} from '../../../utils/validate-payload';
+import {
+  buildController,
+  Controller,
+} from '../../controller/headless-controller';
+
+/**
+ * The `Recommendation` controller exposes a method for retrieving recomendations content in a commerce interface.
```
Spuffynism, line 24:
```suggestion
 * The `Recommendations` controller exposes a method for retrieving recommendations content in a commerce interface.
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export interface Recommendations extends Controller {
+  /**
+   * Fetches the recommendations.
+   */
+  refresh(): void;
+
+  /**
+   * A scoped and simplified part of the headless state that is relevant to the `Recommendation` controller.
```
Spuffynism, line 33:
```suggestion
   * A scoped and simplified part of the headless state that is relevant to the `Recommendations` controller.
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export interface RecommendationsOptions {
+  slotId: string;
+}
+
+interface RecommendationsProps {
+  options: RecommendationsOptions;
+}
+
+function validateRecommendationInitialState(
```
Spuffynism, line 54:
```suggestion
function validateRecommendationsInitialState(
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,116 @@
...
+  engine: CommerceEngine,
+  options: RecommendationsOptions
+) {
+  if (!options) {
+    return;
+  }
+
```
Spuffynism, line 61: No need for this check since `options` is required in the controller builder
```suggestion
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,116 @@
...
+    recommendationsOptionsSchema,
+    options,
+    'buildRecommendations'
+  );
+}
+/**
+ * Creates a `Recommendations` controller instance.
+ *
+ * @param engine - The headless commerce engine.
+ * @returns A `Recommendation` controller instance.
```
Spuffynism, line 73:
```suggestion
 * @returns A `Recommendations` controller instance.
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,116 @@
...
+    },
+
+    refresh: () => dispatch(fetchRecommendations()),
+  };
+}
+
+function loadBaseRecommendationsReducers(
+  engine: CommerceEngine
+): engine is CommerceEngine {
+  engine.addReducers({recommendations, commerceContext, configuration});
```
Spuffynism, line 114: My understanding is that since these reducers will have been loaded by now, there's no need to re-load them.
```suggestion
  engine.addReducers({recommendations});
```

packages/headless/src/features/commerce/recommendations/recommendations-state.ts
```diff
@@ -0,0 +1,22 @@
+import {CommerceAPIErrorStatusResponse} from '../../../api/commerce/commerce-api-error-response';
+import {Product} from '../../../api/commerce/common/product';
+
+export interface RecommendationsState {
+  slotId: string;
+  headline: string;
+  error: CommerceAPIErrorStatusResponse | null;
+  isLoading: boolean;
+  requestId: string;
```
Spuffynism, line 9: I don't see us using the `requestId` property. I think we can remove it. This will affect the `fetchRecommendations.pending` case in the slice as well.
```suggestion
```

packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -22,6 +23,7 @@ import {SortBy, SortCriterion} from '../sort/sort';
 
 export type StateNeededByQueryCommerceAPI = ConfigurationSection &
   ProductListingV2Section &
+  Partial<RecommendationsSection> &
```
Spuffynism, line 26: We should extend the state like we do for search instead: https://github.com/coveo/ui-kit/blob/8f4d68d7a54d6aa539ed85b7fd0516c123097eb7/packages/headless/src/features/commerce/search/search-actions.ts#L15-L16
```suggestion
```

packages/headless/src/features/commerce/recommendations/recommendations-actions.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export const fetchRecommendations = createAsyncThunk<
+  QueryRecommendationsCommerceAPIThunkReturn,
+  void,
+  AsyncThunkCommerceOptions<StateNeededByQueryCommerceAPI>
+>(
+  'commerce/recommendation/fetch',
+  async (_action, {getState, dispatch, rejectWithValue, extra}) => {
+    const state = getState(); // TODO: Add slot id to the state
```
Spuffynism, line 38: By declaring a type to describe the state needed to query the commerce recs api as I suggest above, we'll have access to the slot id here!

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
BenKandelaars: The test that checks they have been loaded if I remove them, so I don't think they are being loaded

packages/headless/src/features/commerce/recommendations/recommendations-state.ts
BenKandelaars: Good point

packages/headless/src/features/commerce/common/actions.ts
BenKandelaars: I've made the change but not 100% this is the correct solution

packages/headless/src/features/commerce/recommendations/recommendations-actions.ts
BenKandelaars: That was a message to an earlier me, by an even earlier me :>

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,112 @@
...
+function validateRecommendationsInitialState(
+  engine: CommerceEngine,
+  options: RecommendationsOptions
+) {
+  validateInitialState(
+    engine,
+    recommendationsOptionsSchema,
+    options,
+    'buildRecommendations'
+  );
```
Spuffynism, line 63: nit: we could inline this since it doesn't do much

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
Spuffynism: What I mean is that to use a recs controller, one will have loaded a context controller before, which loads the context reducer: https://github.com/coveo/ui-kit/blob/fe2a772d3b61aac90fb407fd494769e9b03f6683/packages/headless/src/controllers/commerce/context/headless-context.ts#L168-L173

I thought the config controller was loaded in the context controller as well but it isn't. Nonetheless, this means we could remove the context reducer (and adjust the assertion in the test as well).
```suggestion
  engine.addReducers({recommendations, commerceContext});
```

packages/headless/src/features/commerce/recommendations/recommendations-actions.ts
```diff
@@ -0,0 +1,66 @@
...
+  /** The successful recommendations response. */
+  response: RecommendationsCommerceSuccessResponse;
+}
+
+const buildRecommendationCommerceAPIRequest = async (
+  slotId: string,
+  state: StateNeededByQueryRecommendationsCommerceAPI
+): Promise<CommerceRecommendationsRequest> => {
+  const commerceAPIRequest = await buildCommerceAPIRequest(state);
+  delete commerceAPIRequest.facets;
```
Spuffynism, line 24: Do we still need this?
```suggestion
```

packages/headless/src/features/commerce/recommendations/recommendations-actions.ts
```diff
@@ -0,0 +1,66 @@
...
+  void,
+  AsyncThunkCommerceOptions<StateNeededByQueryRecommendationsCommerceAPI>
+>(
+  'commerce/recommendation/fetch',
+  async (_action, {getState, dispatch, rejectWithValue, extra}) => {
+    const state = getState();
+    const {apiClient} = extra;
+    const fetched = await apiClient.getRecommendations(
+      await buildRecommendationCommerceAPIRequest(
+        state.recommendations?.slotId || '',
```
Spuffynism, line 42: No need for an optional here, since we now have access to `recommendations` through `RecommendationsSection` in `StateNeededByQueryRecommendationsCommerceAPI`
```suggestion
        state.recommendations.slotId,
```

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  const recommendationStateSelector = createSelector(
+    (state: CommerceEngineState) => state.recommendations,
+    (recommendation) => ({
+      headline: recommendation.headline,
+      products: recommendation.products,
+      error: recommendation.error,
+      isLoading: recommendation.isLoading,
+      responseId: recommendation.responseId,
+    })
```
Spuffynism, line 93: The state exposed by the controller and the state stored for recs is the same. I think we can do this:
```suggestion
    (recommendations) => recommendations
```

Spuffynism (changes requested):
Nice work! Getting there!

olamothe (approved):

packages/headless/src/controllers/commerce/recommendations/headless-recommendations.ts
BenKandelaars: I think I get what you mean

packages/headless/src/features/commerce/recommendations/recommendations-actions.ts
BenKandelaars: Facets are always included in the type, so changing this introduces a bigger change to the `CommerceAPIRequest` object

Spuffynism: Maybe since we have a `BaseCommerceAPIRequest`, we can create a `buildBaseCommerceAPIRequest` and use that inside `buildCommerceAPIRequest` and `buildRecommendationCommerceAPIRequest` to build the shared properties then?

BenKandelaars: Yep, I think I have that working now

Spuffynism (approved):
lgtm! Nice work!! ⭐⭐⭐

packages/headless/src/features/commerce/recommendation/recommendation-state.ts
fbeaudoincoveo: Aye! 👍

fbeaudoincoveo (approved):

PR #3737, opened 2024-03-21
chore(ci): deploy patch&minor at the right place

louis-bompart:
We had an incident where, for version v.X.Y.Z, we deployed `/vX/`, `/vY/` and `/vZ/` to our CDN instead of `/vX/`, `/vX.Y/` and `/vX.Y.Z/`.
faeba9fecf72f55da2151fd7ec977e0e652ae8d0 fixes that, and b0a208d0d219561c3bb8d4c583d79daf6720cecd ensure deployment to a given environment can start only after manual approval.
We want the latter to verify our fix is working.

https://coveord.atlassian.net/browse/CDX-1535

### DIFF
--- a/scripts/deploy/execute-deployment-pipeline.mjs
+++ b/scripts/deploy/execute-deployment-pipeline.mjs
@@ -23,16 +23,16 @@ const atomicHostedPage = getVersionComposants(atomicHostedPageJson.version);
 console.log(execSync(`
   deployment-package package create --with-deploy \
     --resolve HEADLESS_MAJOR_VERSION=${headless.major} \
-    --resolve HEADLESS_MINOR_VERSION=${headless.minor} \
-    --resolve HEADLESS_PATCH_VERSION=${headless.patch} \
+    --resolve HEADLESS_MINOR_VERSION=${headless.major}.${headless.minor} \
+    --resolve HEADLESS_PATCH_VERSION=${headless.major}.${headless.minor}.${headless.patch} \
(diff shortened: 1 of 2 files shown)


louis-bompart:
- [ ] @louis-bompart: unfreeze deployments after approval

fbeaudoincoveo (approved):

scripts/deploy/execute-deployment-pipeline.mjs
```diff
@@ -23,16 +23,16 @@ const atomicHostedPage = getVersionComposants(atomicHostedPageJson.version);
 console.log(execSync(`
   deployment-package package create --with-deploy \
     --resolve HEADLESS_MAJOR_VERSION=${headless.major} \
-    --resolve HEADLESS_MINOR_VERSION=${headless.minor} \
-    --resolve HEADLESS_PATCH_VERSION=${headless.patch} \
+    --resolve HEADLESS_MINOR_VERSION=${headless.major}.${headless.minor} \
```
olamothe, line 26: Should there be some kind of validation in the script that the `parse` function does not return garbage stuff ?

olamothe (approved):

scripts/deploy/execute-deployment-pipeline.mjs
louis-bompart: We could add a layer of validation, but IMHO, this would make the script more awkward than it is, for not so much added value.
The parse function worked as intended here, so I don't see that as a need from the incident perpective

PR #3739, opened 2024-03-21
feat(quantic): disclaimer message for rga

erocheleau:
Added a disclaimer at the bottom of the RGA message.

The disclaimer has a default look and text of : 

`Generated content may contain errors. Verify important information.`

And the section for the disclaimer is also a slot, so if someone wants to completely change what gets displayed as the disclaimer, they can not only edit the text, but also the HTML.

The message is also a label, so it can be translated, etc.

Single line footer:
image

Multi-line footer:
image

In the Insight Panel:
image


packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference types="cypress" />
```
mmitiche, line 1: is this fixing the typing issues?

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -136,6 +140,12 @@ describe('quantic-generated-answer', () => {
             Expect.generatedAnswerIsStreaming(false);
           });
 
+          it('should display the disclaimer', () => {
+            Expect.displayDisclaimer(true);
+            Expect.disclaimerContains(GENERATED_ANSWER_DISCLAIMER);
+            Expect.generatedAnswerIsStreaming(false);
+          });
+
```
mmitiche, line 148: We should add the same test to the describe `describe('when the property multilineFooter is set to true', ...)` which can be found on line 215.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1400,11 +1400,18 @@
...
   </labels>
-    <labels>
+  <labels>
     <fullName>quantic_GeneratedAnswerOff</fullName>
     <value>Generated answer OFF</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Generated answer OFF</shortDescription>
   </labels>
+  <labels>
```
mmitiche, line 1410: Please also include the french translation.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -84,6 +85,13 @@ export default class QuanticGeneratedAnswer extends LightningElement {
    * @default {false}
    */
   @api multilineFooter;
+  /**
+   * Indicates whether the disclaimer should be hidden.
+   * @api
+   * @type {boolean}
+   * @default {false}
+   */
+  @api noDisclaimer;
```
mmitiche, line 94: Just my opinion but I feel that naming the property `withoutDisclaimer` or `hideDisclaimer` feels a bit more natural.

Plus if we keep the public property it will need to be tested with Cypress as well.

mmitiche: Also is really necessary to have this property, like we are already exposing a slot for the disclaimer so we already have a way to get a generated answer component without a disclaimer.

I assume that the majority of the people using this component would want the disclaimer since it's useful and they can customize its content.
for the few/rare users who don't want a disclaimer, they can remove it but specifying an empty slot, I know it would not be a super fancy way  to hide a disclaimer, but considering that only very few people will need to do that I would not add a new public property just for this particular situation. 

The idea is I don't see the point of having a public property that allows you to do something that you can achieve without using this property.

mmitiche (commented):
The disclaimer looks good!

Couple adjustments to the tests, french translations, and a question raised about the necessity of a new public property.

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
erocheleau: woops, no it wasn't working, forgot to remove it.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: Do you know if we can empty a slot? If we can then yes, I wouldn't have this property

mmitiche: This should do it:
```
   <c-quantic-generated-answer
      engine-id={engineId}
      answer-style={answerStyle}
      multiline-footer={multilineFooter}
    >
      <div slot="disclaimer"></div>
   </c-quantic-generated-answer>
```

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
```diff
@@ -65,6 +65,13 @@
                 ></c-quantic-generated-answer-rephrase-buttons>
               </div>
             </template>
+            <template lwc:if={shouldShowDisclaimer}>
+              <slot name="disclaimer">
+                <div class={disclaimerCssClass} data-cy="generated-answer__disclaimer">
```
mmitiche, line 70: Some of the styling classes present in `disclaimerCssClass ` needs to be outside the slot, otherwise all these styles will not be take in consideration when displaying custom disclaimers yielding an unwanted display.

For example:
```
<c-quantic-generated-answer engine-id={engineId} answer-style={answerStyle} multiline-footer={multilineFooter}>
    <div slot="disclaimer">Custom Disclaimer</div>
</c-quantic-generated-answer>
```

will give me this::
 
Screenshot 2024-03-22 at 9 28 27 AM

packages/quantic/quantic-examples-community/experiences/Quantic_Examples1/views/insightPanelExample.json
```diff
@@ -34,7 +34,7 @@
                 {
                   "componentAttributes": {
                     "engineId": "example-insight-panel",
-                    "insightId": "6a333202-b1e0-451e-8664-26a1f93c2faf"
+                    "insightId": "142be676-703c-445f-b2d3-fcc7c0a3ded8"
```
SimonMilord, line 37: why this change? did we change the config or something?

SimonMilord (approved):
Mehdi's comment pretty much cover it imo, otherwise looks good!

packages/quantic/quantic-examples-community/experiences/Quantic_Examples1/views/insightPanelExample.json
mmitiche: Since we are it, let's also remove/update where this configuration id is set in other places, it is set in multiple places and it is not always useful: 

-  https://github.com/coveo/ui-kit/blob/8f4d68d7a54d6aa539ed85b7fd0516c123097eb7/packages/quantic/force-app/examples/main/lwc/exampleInsightPanel/exampleInsightPanel.js-meta.xml#L15
- https://github.com/coveo/ui-kit/blob/8f4d68d7a54d6aa539ed85b7fd0516c123097eb7/packages/quantic/force-app/examples/main/lwc/exampleInsightPanel/exampleInsightPanel.js#L24

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -258,6 +268,7 @@ describe('quantic-generated-answer', () => {
             Expect.displayCopyToClipboardButton(false);
             Expect.displayToggleGeneratedAnswerButton(true);
             Expect.toggleGeneratedAnswerButtonIsChecked(true);
+            Expect.displayDisclaimer(false);
```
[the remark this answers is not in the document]
erocheleau: The test already checks that the `generated-answer__card` isn't being displayed, the disclaimer is inside the answer card, so if the card isn't displayed, then the disclaimer isn't displayed either.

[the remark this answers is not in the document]
erocheleau: But I added it after the answer card is back to visible.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: ok then I'm happy to remove the option :)

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
erocheleau: Yes my goal with not putting the css class outside the slot was to allow the customers to style their own disclaimer without "competing" with our own style.

I know it's easier for them if they just want to add some text, but if they would want to display the disclaimer differently it can become impossible if the classes are outside the slot. What do you think?

packages/quantic/quantic-examples-community/experiences/Quantic_Examples1/views/insightPanelExample.json
erocheleau: Yea sure we can remove it in the default example component.

@SimonMilord This is just a different config for the HIP which has RGA enabled and is using a query we know will generate a RGA response.

packages/quantic/force-app/examples/main/lwc/exampleInsightPanel/exampleInsightPanel.js-meta.xml
```diff
@@ -12,7 +12,7 @@
     <targetConfigs>
         <targetConfig targets="lightningCommunity__Default, lightning__AppPage, lightning__RecordPage, lightning__HomePage">
             <property name="engineId" type="String" default="example-insight-panel" label="Enter Coveo Headless engine ID"/>
-            <property name="insightId" type="String" default="6a333202-b1e0-451e-8664-26a1f93c2faf" label="The insight panel config ID"/>
+            <property name="insightId" type="String" default="" label="The insight panel config ID"/>
```
mmitiche, line 15: ?
```suggestion
            <property name="insightId" type="String" label="The insight panel config ID"/>
```

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
mmitiche: Indeed I get your point, but we can at least set the class that would display the disclaimer  under the citation and the rephrase buttons outside the slot so we know for sure that the disclaimer will be positioned in the right position.
Then, all the other classes can be set inside the slot so they can easily be overwritten by the classes that the user wanna use to style the disclaimer.

mmitiche (approved):
Looks good to me 👍

packages/quantic/force-app/examples/main/lwc/exampleInsightPanel/exampleInsightPanel.js-meta.xml
mmitiche: Sorry I didn't explain well my comment, I meant why not completely remove the `default=""` since it does nothing.

PR #3741, opened 2024-03-22
chore: disable quantic release, split production targets

louis-bompart:
Currently, we're doing the following operation sequentially:
1. Move the latest tag of our npm package to the beta tag we created earlier
2. Promote the SFDX package in production
3. Notify the doc website to do their release

All those operations can and should occur in parallel:
- We want to promote NPM into production as soon as it's available in production through our CDN
- We want our documentation to be updated when we release new versions in production, even if it's only in the CDN (if the NPM promotion were to fail let's say)
- Quantic package does not depend on the two other operations

This PR does that by splitting the three tasks into separate jobs, controlled by different GitHub Environments. Those Environments are just a split of Production into 3, with a separation of concern applied to all.

By that account. GitHub package releases, while currently disabled should follow NPM logic as well, and so I moved the disabled part of the script into its own, commented, job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,22 +40,13 @@ jobs:
           RELEASER_INSTALLATION_ID: ${{ secrets.RELEASER_INSTALLATION_ID }}
           DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


louis-bompart:
- [ ] @louis-bompart after approval, create GitHub Environments. Ensure appropriate credentials are set for enabled targets. (NPM & Doc as we speak)

fbeaudoincoveo (approved):

PR #3742, opened 2024-03-22
feat(headless): case assist analytics migrated to the event protocol

mmitiche:
## [SFINT-5398](https://coveord.atlassian.net/browse/SFINT-5398)

- Case Assist analytics actions migrated to the new event protocol.
- All the analytics are now unit tested.

https://github.com/coveo/ui-kit/assets/86681870/8b68d7d5-599d-406a-ad60-e152feebbc53

[SFINT-5398]: https://coveord.atlassian.net/browse/SFINT-5398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/case-assist/__snapshots__/case-assist-analytics-actions.test.ts.snap
mmitiche, line 1: Hello @jstpierref, @felixpelletier 

I'm adding you for visibility, here I'm migrating the case assist  analytics to the event protocol.
Looking at this *.test.ts.snap should give you quite a good idea of the events being test from the Headless library.
I also confirmed all the events sent here are valid using the `validate` endpoint.

Thank you!

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -31,128 +36,249 @@ export const logCaseNextStage = (
...
-  );
+      });
+    },
+    analyticsType: 'CaseAssist.CreateTicket',
+    analyticsPayloadBuilder: (state): CaseAssist.CreateTicket => {
+      const {id, category, subject, description, productId} =
+        caseAssistCaseSelector(state);
+      return {
+        ticket: {
+          id: id!,
```
louis-bompart, line 52: Why can't caseAssistSelector return with certitude an id?

louis-bompart (approved):
LGTM, just a :suspect:

SimonMilord (approved):
lgtm, just a question about tests

packages/headless/src/features/case-assist/case-assist-analytics-actions.test.ts
```diff
@@ -0,0 +1,520 @@
...
+        engine.dispatch,
+        () => engine.state,
+        {} as ThunkExtraArguments
+      );
+
+      const mockToUse = mockLogEnterInterface;
+
+      expect(mockToUse).toHaveBeenCalledTimes(1);
+      expect(mockToUse).toHaveBeenCalledWith({ticket: expectedTicketPayload});
+    });
```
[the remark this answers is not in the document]
mmitiche: Actually all the legacy tests you see here are temporary, they will be deleted once we pass to the v3.
While the refactoring idea is good one, it is in these circumstances not worth it.

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -10,14 +11,18 @@ import {
...
+  makeCaseAssistAnalyticsAction({
+    prefix: 'analytics/caseAssist/case/start',
+    __legacy__getBuilder: (client, state) => {
+      return client.logEnterInterface({
         ticket: caseAssistCaseSelector(state),
-      })
-  );
+      });
+    },
+    analyticsType: 'CaseAssist.Start',
```
erocheleau, line 21: Apparently this should be `camelCase` instead of `PascalCase`.
The casing doesn't matter for the event protocol, but at least when sending it we should use the right casing. :)

erocheleau: BTW it's okay that the TypeScript Type starts with a capital letter.
But not the `type` in the event.

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
erocheleau: I think we're not sure to always get it back from Salesforce? Might be just because we are too careful xD

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -31,128 +36,249 @@ export const logCaseNextStage = (
...
         ),
         ticket: caseAssistCaseSelector(state),
-      })
-  );
+      });
+    },
+    analyticsType: 'CaseAssist.SelectFieldClassification',
+    analyticsPayloadBuilder: (state): CaseAssist.SelectFieldClassification => {
+      return {
+        autoselected: false,
```
erocheleau, line 156: Do we just omit it? Or do we have to be explicit that it's "not" autoselected?

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -31,128 +36,249 @@ export const logCaseNextStage = (
...
           state,
           suggestionId,
           true
         ),
         ticket: caseAssistCaseSelector(state),
-      })
-  );
+      });
+    },
+    analyticsType: 'ItemClick',
```
erocheleau, line 202:
```suggestion
    analyticsType: 'itemClick',
```

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -31,128 +36,249 @@ export const logCaseNextStage = (
...
+        responseId,
+        suggestionId: uniqueId,
+      } = caseAssistDocumentSuggestionSelector(state, suggestionId, true);
+      return {
+        searchUid: responseId,
+        position: suggestion.documentPosition,
+        actionCause: 'preview',
+        itemMetadata: {
+          uniqueFieldName: 'uniqueId',
+          uniqueFieldValue: uniqueId,
```
erocheleau, line 215: This is the uniqueId of the document suggestion? Is it just the permanentId? Or is there a separate doc suggestion Id returned by the API?

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -31,128 +36,249 @@ export const logCaseNextStage = (
...
           suggestionId,
           false,
           true
         ),
         ticket: caseAssistCaseSelector(state),
-      })
-  );
+      });
+    },
+    analyticsType: 'ItemClick',
```
erocheleau, line 239:
```suggestion
    analyticsType: 'itemClick',
```

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
```diff
@@ -31,128 +36,249 @@ export const logCaseNextStage = (
...
+    __legacy__getBuilder: (client, state) => {
+      return client.logRateDocumentSuggestion({
         rating,
         suggestion: caseAssistDocumentSuggestionSelector(state, suggestionId),
         ticket: caseAssistCaseSelector(state),
-      })
-  );
+      });
+    },
+    analyticsType: 'CaseAssist.DocumentSuggestionFeedback',
```
erocheleau, line 274:
```suggestion
    analyticsType: 'caseAssist.documentSuggestionFeedback',
```

packages/headless/src/features/case-assist/case-assist-analytics-actions.test.ts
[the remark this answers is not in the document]
erocheleau: Is it the same reason you didn't use the snapshots for the legacy events too?

packages/headless/src/features/case-assist/case-assist-analytics-actions.test.ts
```diff
@@ -0,0 +1,520 @@
...
+
+    it('should log #logDocumentSuggestionRating with the right payload', async () => {
+      await logDocumentSuggestionRating(exampleUniqueId, 1)()(
+        engine.dispatch,
+        () => engine.state,
+        {} as ThunkExtraArguments
+      );
+
+      expect(emit).toHaveBeenCalledTimes(1);
+      expect(emit.mock.calls[0]).toMatchSnapshot();
```
erocheleau, line 517: Damn it's so beautiful! :D

erocheleau (approved):
Just some naming convention but otherwise I think it's all there! GG

packages/headless/src/features/case-assist/case-assist-analytics-actions.test.ts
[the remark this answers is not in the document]
SimonMilord: ok yeah makes sense

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
mmitiche: It is kinda weird to write the type in two multiple ways, one `CaseAssist.Start` and `caseAssist.start`.
I would prefer consistency in the naming to avoid any confusions.

mmitiche: updated the names to be `camelCase`.

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
mmitiche: yes we need to be explicit.

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
mmitiche: exactly it's because we are too careful  but realistically there should not really be a situation where the id is not found.

packages/headless/src/features/case-assist/case-assist-analytics-actions.ts
mmitiche: Indeed this the uniqueId and this is the only doc suggestion id returned by the API.

PR #3749, opened 2024-03-26
refactor(atomic): first phase refactor of atomic result list common

olamothe:
First phase to refactor problematic files under `atomic/src/components/common/result-list`.

**There's work left to be done**:  Mainly, the naming is still problematic. You will see there's an ugly mix/match of the keyword `result` and `item` sprinkled everywhere in this PR.

As an end goal, everything under `atomic/src/components/common` should not (or not prominently) contain the keyword `result`, because that's a bias for the `search` or `service` use case, as opposed to `commerce` where were dealing with `products`. A proposition from Fred was to use the keyword `item` to cover both in a general manner.

However, as I started to do the renaming changes, this (already very large PR) started to balloon even more. So I want to push that to a different PR.

This PR primarily modifies `CommonResultList` to remove most of it's responsibilities, and explode them in a bunch of smaller functional components.

I invite you to look at `atomic-result-list` to get a good overview of the changes.

* A lot of the changes is "moving code around": it's not all net new code, quite a bit of cut and pasting/salvaging.

* Rendering the result (ie: `atomic-result`, `atomic-insight-result`), and mapping the proper types to individual functional components is now done at the web component level (`atomic-result-list`, `atomic-folded-result-list`, `atomic-recommendation-list`, `atomic-insight-result-list`.
* There should not be new types added that are dependant on any Headless sub package in `/src/components/common/result-list` in the new code added...
* Added `*Guards` type component that deals with blocking rendering when there's no results, or placeholder when appropriate
* Reworked `Grid/Table-Display` and `Wrapper` type component to be more about letting upstream code deciding the child render function
* Switch/case code to decide which display should be applied (`grid`, `list`, `table`) for the result list is now done at the web component level. If you dug into how it was before, there was un-necessary code for some use case being handled by the (too large) ResultListCommon component. For example, there is no `table` display for insight panel

* As mentioned in the design doc, drawback to this approach is that there's more code duplication in the top level web component If you compare all the `*-result-list` component code.

* `ResultListCommon` is not completely eliminated: It is still used for accessibility related functionality (ie: control keyboard focus after results gets rendered). The code that deals with that is stateful (keep track of last result index or current focused element), so I think some stateful class still makes sense to handle that: Might as well use what we already have and is proven to work.

https://coveord.atlassian.net/browse/KIT-3033


olamothe:
Putting it back to draft due to issue with table display

fbeaudoincoveo (approved):

PR #3754, opened 2024-03-28
test(quantic): adapted Quantic E2E  for case assit with the new event protocol

mmitiche:
## [SFINT-5400](https://coveord.atlassian.net/browse/SFINT-5400)

- Adapted the Cypress e2e test for Case Classification and Document Suggestion features  for testing with analytics sent with the new protocol, now the test spec tests for both cases: the case when send legacy analytics and the case where we sent next analytics.

```
Expect.logClickingSuggestion(clickIndex, allDocuments);
```
```
if (analyticsMode === "next") {
  NextAnalyticsExpectations.emitCaseAssistDocumentSuggestionClick(
    {
      documentSuggestion: {
        id: allDocuments[clickIndex].uniqueId,
        responseId: exampleResponseId,
      },
    },
    exampleTrackingId
  );
} else {
  Expect.logClickingSuggestion(clickIndex, allDocuments);
}

```

[SFINT-5400]: https://coveord.atlassian.net/browse/SFINT-5400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/cypress.config.ts
```diff
@@ -9,6 +9,7 @@ export default defineConfig({
     runMode: 3,
     openMode: 1,
   },
+  numTestsKeptInMemory: 0,
```
SimonMilord, line 12: why do we want the memory to keep 0 tests here? I think by default cypress keeps 50, was that causing the tests to be particularly slow? Won't it cause issues when tests have to re-try?

packages/quantic/cypress/e2e/default-2/document-suggestion/document-suggestion.cypress.ts
```diff
@@ -91,29 +293,13 @@ describe('quantic-document-suggestion', () => {
...
-
-        sendRating(clickIndex);
-        Expect.logRatingSuggestion(clickIndex, allDocuments);
-      });
     });
   });
 
   describe('when using a custom number maxDocuments', () => {
-    const maxDocuments = 3;
+    const maxDocuments = 2;
```
SimonMilord, line 300: ?

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -250,17 +1641,17 @@ describe('quantic-case-classification', () => {
...
-        Expect.correctValue(allOptions[clickedIndex].value);
-      });
+      //   Actions.openSelectInput();
+      //   Actions.clickSelectOption(clickedIndex);
+      //   Expect.logUpdatedClassificationFromSelectOption(
+      //     coveoDefaultField,
+      //     clickedIndex
+      //   );
+      //   Expect.correctValue(allOptions[clickedIndex].value);
+      // });
```
SimonMilord, line 1654: I think you forgot this test here :P

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -61,66 +67,1498 @@ describe('quantic-case-classification', () => {
...
+                },
+                exampleTrackingId
+              );
+              NextAnalyticsExpectations.emitUpdateField(
+                {
+                  fieldName: coveoDefaultField,
+                  fieldValue: allOptions[firstSuggestionIndex].value,
+                },
+                exampleTrackingId
+              );
```
SimonMilord, line 537:
```suggestion
            if (analyticsMode === 'next') {
              NextAnalyticsExpectations.emitCaseAssistSelectFieldClassification(
                {
                  fieldClassification: {
                    id: allOptions[firstSuggestionIndex].id,
                    responseId: exampleResponseId,
                  },
                  autoselected: false,
                },
                exampleTrackingId
              );
              
              NextAnalyticsExpectations.emitUpdateField(
                {
                  fieldName: coveoDefaultField,
                  fieldValue: allOptions[firstSuggestionIndex].value,
                },
                exampleTrackingId
              );
```

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -61,66 +67,1498 @@ describe('quantic-case-classification', () => {
...
+              NextAnalyticsExpectations.emitUpdateField(
+                {
+                  fieldName: coveoDefaultField,
+                  fieldValue: allOptions[clickedIndex].value,
+                },
+                exampleTrackingId
+              );
+            } else {
```
SimonMilord, line 692:
```suggestion
              NextAnalyticsExpectations.emitCaseAssistSelectFieldClassification(
                {
                  fieldClassification: {
                    id: allOptions[clickedIndex].id,
                    responseId: exampleResponseId,
                  },
                  autoselected: false,
                },
                exampleTrackingId
              );
              
              NextAnalyticsExpectations.emitUpdateField(
                {
                  fieldName: coveoDefaultField,
                  fieldValue: allOptions[clickedIndex].value,
                },
                exampleTrackingId
              );
            } else {
```

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -61,66 +67,1498 @@ describe('quantic-case-classification', () => {
...
+                },
+                exampleTrackingId
+              );
+              NextAnalyticsExpectations.emitUpdateField(
+                {
+                  fieldName: coveoDefaultField,
+                  fieldValue: allOptions[clickedIndex].value,
+                },
+                exampleTrackingId
+              );
```
SimonMilord, line 792:
```suggestion
              NextAnalyticsExpectations.emitCaseAssistSelectFieldClassification(
                {
                  fieldClassification: {
                    id: allOptions[clickedIndex].id,
                    responseId: exampleResponseId,
                  },
                  autoselected: false,
                },
                exampleTrackingId
              );
              
              NextAnalyticsExpectations.emitUpdateField(
                {
                  fieldName: coveoDefaultField,
                  fieldValue: allOptions[clickedIndex].value,
                },
                exampleTrackingId
              );
```

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -61,66 +67,1498 @@ describe('quantic-case-classification', () => {
...
+                  exampleTrackingId
+                );
+                NextAnalyticsExpectations.emitUpdateField(
+                  {
+                    fieldName: coveoDefaultField,
+                    fieldValue: allOptions[clickedIndex].value,
+                  },
+                  exampleTrackingId
+                );
+              } else {
```
SimonMilord, line 1115:
```suggestion
              if (analyticsMode === 'next') {
                NextAnalyticsExpectations.emitCaseAssistSelectFieldClassification(
                  {
                    fieldClassification: {
                      id: allOptions[clickedIndex].id,
                      responseId: exampleResponseId,
                    },
                    autoselected: false,
                  },
                  exampleTrackingId
                );
                
                NextAnalyticsExpectations.emitUpdateField(
                  {
                    fieldName: coveoDefaultField,
                    fieldValue: allOptions[clickedIndex].value,
                  },
                  exampleTrackingId
                );
              } else {
```

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -61,66 +67,1498 @@ describe('quantic-case-classification', () => {
...
```
SimonMilord, line 1353:
```suggestion
              if (analyticsMode === 'next') {
                NextAnalyticsExpectations.emitCaseAssistSelectFieldClassification(
                  {
                    fieldClassification: {
                      id: allOptions[clickedIndex].id,
                      responseId: exampleResponseId,
                    },
                    autoselected: false,
                  },
                  exampleTrackingId
                );
                
                NextAnalyticsExpectations.emitUpdateField(
                  {
                    fieldName: coveoDefaultField,
                    fieldValue: allOptions[clickedIndex].value,
                  },
                  exampleTrackingId
                );
              } else {
```

SimonMilord (approved):
Good job! Just wondering whether it would be a good idea to make a seperate file for analytics related tests for the case-classification suite or if we absolutely want to keep it in the same file... It has more than 1700 lines and at times can get a bit confusing. Just food for thought curious to hear what you think?

packages/quantic/cypress.config.ts
mmitiche: This was a suggestion from Cypress, sometimes when the spec files have a lot of tests, the tests can crash running out of memory, one of the factors that should help avoiding this is reducing the number of tests kept in memory to 0, so I'm setting the value and I'm monitoring the tests to see if the running out of memory problem is still happening.

mmitiche:
> Good job! Just wondering whether it would be a good idea to make a seperate file for analytics related tests for the case-classification suite or if we absolutely want to keep it in the same file... It has more than 1700 lines and at times can get a bit confusing. Just food for thought curious to hear what you think?

Indeed it's a good idea and it's a possibility that we could consider.
The thing that also we need to consider is that starting from the v3 we would remove the complexity that comes with testing both the event protocol and and the legacy way of sending analytics, removing this complexity will reduce the size of these tests considerably.

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -61,66 +67,1498 @@ describe('quantic-case-classification', () => {
...
+                    id: allOptions[firstSuggestionIndex].id,
+                    responseId: exampleResponseId,
+                  },
+                  autoselected: true,
+                },
+                exampleTrackingId
+              );
+            } else {
+              Expect.logClickedSuggestion(firstSuggestionIndex, true);
+            }
```
erocheleau, line 128: Would it be better to organize things like this instead:

```
const analyticsExpectation = Expect.clickedSuggestionExpectation(analyticsMode);
if(analyticsMode === 'next') {
    analyticsExpectation({
        fieldClassification: {
            id: allOptions[firstSuggestionIndex].id,
            responseId: exampleResponseId,
        },
        autoselected: true,
    },
    exampleTrackingId);
} else {
    analyticsExpectation(firstSuggestionIndex, true);
}
```

And in the case-classification-expectations file, you could have the function `clickedSuggestionExpectation` return either the previous analytics expectations function, or the next-analytics-expectations function?

I think this way it would be easier later to remove the test in V3 for example if we want to stop testing for previous analytics?

It would be even better if we could be able to pass the same object to the `Expect.clickedSuggestionExpectations` so that the "body" of the expected analytics is built from the same object.
This way we could have perhaps

```
Expect.clickedSuggestionAnalyticsExpectations(dataToBuildTheExpectedBody, analyticsMode);
```

packages/quantic/cypress/e2e/default-1/case-classification/case-classification.cypress.ts
```diff
@@ -61,66 +67,1498 @@ describe('quantic-case-classification', () => {
...
+            } else {
+              Expect.logUpdatedClassificationFromSelectOption(
+                coveoDefaultField,
+                clickedIndex
+              );
+            }
+            Expect.correctValue(allOptions[clickedIndex].value);
+          });
+        });
+      });
```
erocheleau, line 188: Why are those moved to the end?

packages/quantic/cypress/e2e/default-2/document-suggestion/document-suggestion.cypress.ts
mmitiche: 3 was the default number of max documents, so I'm using a real custom value in this test, 2.

PR #3766, opened 2024-04-02
refactor(atomic): rename results keyword to use case agnostic item

olamothe:
Sister PR of https://github.com/coveo/ui-kit/pull/3749 , which deals with mass name change explained in the other PR.

Mostly straight up rename with no behaviour impact.

Minor code change in the decorators function/file (`item-decorators.ts`) , since those were (needlessly) pulling in from `@coveo/headless`, while they are located in the `common` folder of Atomic (so: should be use case agnostic).

Once all refactors are finished for Atomic for Commerce, I'll look to put in place linting rule that prevent those kind of imports from that specific folder in Atomic: They are problematic because they make it hard to genericize components, as explained in the original design doc linked to the epic.

https://coveord.atlassian.net/browse/KIT-3033


packages/atomic/src/components/common/item-list/item-decorators.tsx
louis-bompart, line 1: Quite a mouthful 😅

louis-bompart (approved):
lgtm overall

PR #3671, opened 2024-03-05
feat(commerce): open interactive result controller to all commerce solution types

Spuffynism:
Bring up the interactive result controller, out from the product listing folder, to make it usable for all commerce solution types.

In LENS-1500, the controller will trigger a product click event on result selection.

[CAPI-500]

PR #3768, opened 2024-04-03
feat(commerce): expose interactive result sub-controllers

Spuffynism:
Expose the interactive results controller as a solution-type sub-controller.

Supersedes https://github.com/coveo/ui-kit/pull/3671

[CAPI-500]

[CAPI-500]: https://coveord.atlassian.net/browse/CAPI-500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/commerce/core/result-list/headless-core-interactive-result.ts
```diff
@@ -1,28 +1,41 @@
...
   InteractiveResultCoreOptions,
-  InteractiveResultCoreProps,
+  InteractiveResultCoreProps as InteractiveResultHeadlessCoreProps,
 } from '../../../core/interactive-result/headless-core-interactive-result';
 
 export interface InteractiveResultOptions extends InteractiveResultCoreOptions {
   /**
    * The product.
    */
+  // TODO: Which Product should this be? The headless one or the relay one? The product view controller uses the relay one.
```
Spuffynism, line 15: I personally think it would make more sense for headless controllers to expose the headless `Product` and not the relay `Product`. My understanding is that we can map relay `Product` properties from a headless `Product` (as I've done below).

The TODOs below have related questions as well.

@samisayegh @sallainCoveo thoughts?

packages/headless/src/controllers/commerce/core/sub-controller/headless-sub-controller.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface SolutionTypeSubControllers {
+  interactiveResult: (props: InteractiveResultProps) => InteractiveResult;
+}
+
+interface SubControllerProps {
+  responseIdSelector: () => string;
+}
+
+export function buildSolutionTypeSubControllers(
```
Spuffynism, line 16: Having this helper method allows us to add sub-controller support for solution type controllers without changing them individually. When we'll add support for pagination, sort and facets as sub-controllers in [CAPI-719](https://coveord.atlassian.net/browse/CAPI-719), this is where we'll do changes.

packages/headless/src/controllers/commerce/core/result-list/headless-core-interactive-result.ts
samisayegh: Hmm, since the products one would pass to the interactive result controller would normally have the shape of those returned by commerce api/headless, that suggests the headless `Product` interface.

However, this would hide errors because the product fields need to be present. We know this will sometimes not be the case because e.g. `ec_price` is fetched from a third-party system in real-time.

So, if we used the headless `Product` interface, at best the fields would be optional e.g.

```
HProduct {
  ec_price?: number;
}
```
but `ec_price` is required for the analytics event

```
RProduct {
  ec_price: number;
}
```

Using the relay type would be more strict/accurate. But, it means that the following would error:

```
buildInteractiveResult({
  product: commerceEngine.products[0] // error: ec_price is optional but is required
}));
```
So one would have to do:

```
buildInteractiveResult({
  product: {
    ec_price: commerceEngine.products[0].ec_price!
    ... same for other fields ...
  }
})
```

tl;dr, I think the relay type would provide the more accurate type check, but it means implementers would need to map the fields, which is not super nice

packages/headless/src/controllers/commerce/core/result-list/headless-core-interactive-result.ts
```diff
@@ -44,7 +57,19 @@ export function buildInteractiveResult(
...
     wasOpened = true;
-    // TODO: Log on click
+    engine.dispatch(
+      productClick({
+        // TODO: If we used the relay Product, we would not need to construct the product here.
+        product: {
+          productId: props.options.product.permanentid,
+          name: props.options.product.ec_name ?? '',
+          price: props.options.product.ec_price ?? 0,
+        },
```
samisayegh, line 67: Agreed with the todo. A "benefit" of the current approach is it by-passes the typescript check, which means one would only realize the prices are always 0 at runtime, hmm.

packages/headless/src/controllers/commerce/core/result-list/headless-core-interactive-result.ts
Spuffynism: > We know this will sometimes not be the case because e.g.`ec_price` is fetched from a third-party system in real-time.

For this reason, I'd be comfortable exposing the relay `Product`. We must provide the flexibility because there is no way to consistently extract the product click payload from the headless product. I'll go with this option and then open the PR for review.

olamothe (approved):

fbeaudoincoveo (approved):

packages/headless/src/controllers/commerce/core/result-list/headless-core-interactive-result.ts
```diff
@@ -0,0 +1,74 @@
...
+} from '../../../core/interactive-result/headless-core-interactive-result';
+
+export interface InteractiveResultOptions extends InteractiveResultCoreOptions {
+  /**
+   * The product to log analytics for.
+   */
+  product: Product;
+
+  /**
+   * The product's position, using a 1-based index.
```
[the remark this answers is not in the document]
Spuffynism: I've clarified that the position is across the entire result set, not across a page.

packages/headless/src/controllers/commerce/core/result-list/headless-core-interactive-result.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    name: 'clicked on product',
+    price: 17.99,
+  };
+
+  function initializeInteractiveResult(delay?: number) {
+    buildCoreInteractiveResult(engine, {
+      options: {
+        product,
+        position: 1,
+        selectionDelay: delay,
```
sallainCoveo, line 26: To learn a little bit more about this, why do we want to delay selection sometimes?

sallainCoveo (approved):

packages/headless/src/controllers/commerce/core/result-list/headless-core-interactive-result.test.ts
Spuffynism: With `select`, one can trigger a click immediately. The core controller offers the `delay` option to handle DOM touchstart and DOM touchend events. In other words, it helps handle longpresses:
https://github.com/coveo/ui-kit/blob/a5296ec9640f9453b6b1b661abf04621b523b8fc/packages/headless/src/controllers/core/interactive-result/headless-core-interactive-result.ts#L88-L96
More info in the docs here (I found the example snippet up top helpful!): https://docs.coveo.com/en/headless/latest/reference/search/controllers/interactive-result/#begindelayedselect

PR #3779, opened 2024-04-08
chore(headless): applying abortable fetch shim for quantic headless bundle

mmitiche:
[SFINT-5481](https://coveord.atlassian.net/browse/SFINT-5481)

This PR aims to fix the issue that was occurring in Salesforce when LWS is disabled, this issue occurred because of the absence of the `AbortController`.

In order to fix this, we are building a specific headless bundle made only for the Quantic library.
In this bundle that will only be used inside Salesforce, we are injecting some ponyfills that are providing the features that are normally provided by the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController ).

These ponyfills that are injected using ESbuild, will allow the headless code to run properly by executing abortable fetch calls even when the Abort Controller web API is not available inside Salesforce.

https://github.com/coveo/ui-kit/assets/86681870/06d6e2b5-116b-4071-8c4c-cf7304509ac7

[SFINT-5481]: https://coveord.atlassian.net/browse/SFINT-5481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/api/generated-answer/generated-answer-client.ts
```diff
@@ -104,6 +104,7 @@ export class GeneratedAnswerAPIClient {
     const stream = () =>
       fetchEventSource(buildStreamingUrl(url, organizationId, streamId), {
         method: 'GET',
+        fetch,
```
erocheleau, line 107: What's that parameter being added does exactly?
Assuming this is us injecting what "fetch" client to use for the event source library?

erocheleau (commented):
Looks good, if you can @mmitiche could you please add a demo here of a quantic version with this "patched" headless code that works inside the package and the hosted insight panel?

This is the situation I am most sure we'll have the issue, if you disable the "use LWS" it also affects the console in salesforce, and the community always had it's own "weird" version of locker version the version running in the Salesforce console.

packages/headless/src/api/generated-answer/generated-answer-client.ts
mmitiche: Exactly yeah!

if we don't pass this parmeter, by default it uses `window.fetch`, however the ponyfill we are using is exporting `fetch` so it will not take any effect when you can `window.fetch`

So I'm passing this parameter here to make the library use `fetch` aka our ponyfill aboratable fetch.

mmitiche:
> Looks good, if you can @mmitiche could you please add a demo here of a quantic version with this "patched" headless code that works inside the package and the hosted insight panel?
> 
> This is the situation I am most sure we'll have the issue, if you disable the "use LWS" it also affects the console in salesforce, and the community always had it's own "weird" version of locker version the version running in the Salesforce console.

Sure yeah! I will try that.

packages/headless/esbuild.mjs
```diff
@@ -128,6 +128,28 @@ const browserUmd = Object.entries(useCaseEntries).map((entry) => {
   );
 });
 
+const quanticUmd = Object.entries(useCaseEntries).map((entry) => {
```
fbeaudoincoveo, line 131: Since you're iteraing over each element in `useCaseEntries`, I believe you'll be generating some superfluous bundles.

I don't think you need product-recommendation, product-listing, ssr, and commerce in the context of Quantic so maybe you could skip those.

fbeaudoincoveo (approved):

mmitiche:
@erocheleau  I was able to confirm that this new bundle works correctly in the Salesforce console:

### Before this PR:
Screenshot 2024-04-09 at 4 20 41 PM

### After this PR:
Screenshot 2024-04-09 at 4 36 08 PM

packages/headless/esbuild.mjs
```diff
@@ -128,6 +128,28 @@ const browserUmd = Object.entries(useCaseEntries).map((entry) => {
...
+  const outDir = getUseCaseDir('dist/quantic/', useCase);
+  const outfile = `${outDir}/headless.js`;
+
+  const globalName = getUmdGlobalName(useCase);
+
+  return buildBrowserConfig(
+    {
+      entryPoints: [entryPoint],
+      outfile,
+      format: 'cjs',
```
mmitiche, line 142: Tried the `iife` format, and SF was not able to load the bundle properly. switched to UMD for now and I'm currently investigating the problem of `iife` is SF.

louis-bompart (approved):
LGTM, nice job!

packages/headless/esbuild.mjs
louis-bompart: It's 'good enough' for a fix, don't necessarily drop the investigation w/ iife, but we can move on and merge this if @erocheleau thinks it's good.
If that's the case, we could have a green release of Quantic tomorrow, which would be swell

erocheleau:
> @erocheleau I was able to confirm that this new bundle works correctly in the Salesforce console:
> 
> ### Before this PR:
> Screenshot 2024-04-09 at 4 20 41 PM
> ### After this PR:
> Screenshot 2024-04-09 at 4 36 08 PM

Perfect! Approved!

erocheleau (approved):

PR #3782, opened 2024-04-10
chore: update to storybook v8

louis-bompart:
PoC to demonstrate the migration process and what's required.

1. The main job is to comply with: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#titles-are-statically-computed. We can't opt-out anymore in v8, so we need to do some grunt work to:
  - Remove id from the default export: We relinquish control.
  - Exposes the name 'directly' from the template, not as the result of a function
This PR demonstrates the grunt work on a select set of stories (and 🗑️ the other, so that it builds)

2. Run the migration script https://storybook.js.org/docs/migration-guide. The result is in 2e66499d49f1d84b500c5d92c6daec976d7d8012

For now, I checked:
- that the `path` query param fit with what's present on the CDN
- Shadow Part is reflected on both the HTML preview and code preview
- Control (attribute changes) is reflected on both the HTML preview and code preview


utils/atomic-storybook/.storybook/recommendation/recs-story.tsx
```diff
@@ -1,17 +1,14 @@
...
-import {Args} from '@storybook/api';
 import defaultStory from '../default-story';
 import {DefaultStoryAdvancedConfig} from '../default-story-shared';
 import {initializeInterfaceDebounced} from './recs-init';
 
 export default function recsStory(
-  title: string,
   componentTag: string,
-  defaultArgs: Args,
+  defaultArgs: Record<string, unknown>,
```
louis-bompart, line 8: Args is not exported anymore

utils/atomic-storybook/.storybook/main.js
```diff
@@ -62,10 +61,14 @@ module.exports = {
     };
   },
   framework: {
-    name: '@storybook/html-webpack5',
+    name: getAbsolutePath("@storybook/html-webpack5"),
```
louis-bompart, line 64: 🤷 migration codemod

utils/atomic-storybook/.storybook/default-story-shared.tsx
```diff
@@ -111,8 +109,6 @@ export default function sharedDefaultStory(
   };
 
   const defaultModuleExport = {
-    title,
-    id: componentTag,
```
louis-bompart, line 115: the default story do not reflect the id anymore

packages/atomic/src/components/common/atomic-icon/atomic-icon.stories.tsx
```diff
@@ -73,5 +72,9 @@ const {defaultModuleExport, exportedStory} = defaultStory(
   }
 );
 
-export default {...defaultModuleExport, title: 'Atomic/Icon'};
+export default {
+  ...defaultModuleExport,
+  title: 'Atomic/Icon',
+  id: 'atomic-icon',
+};
```
louis-bompart, line 79: So, yes it may seems that we could refactor the first argument of `defaultStory`and the id exported, but no: it breaks storybook.
We need to export it 'straight up'

fbeaudoincoveo (approved):
I tested locally and aside from the default story names (which really has no impact)  the only "odd" thing I noticed is that the `atomic-format-*` stories now appear both under ResultList/ResultNumber/Format and NumericFacet/Format (they used to only appear under the former).

I believe this will cause no issues on the docs site.

louis-bompart:
> I tested locally and aside from the default story names (which really has no impact) the only "odd" thing I noticed is that the `atomic-format-*` stories now appear both under ResultList/ResultNumber/Format and NumericFacet/Format (they used to only appear under the former).
> 
> I believe this will cause no issues on the docs site.

And they are actually two different stories, so it was most likely a bug even IMO

olamothe (approved):

PR #3784, opened 2024-04-10
feat(genQA): add show more button

jelmedini:
[SVCC-3574](https://coveord.atlassian.net/browse/SVCC-3574)

- An option should be added to the RGA component to enable the collapsible mode.
- By default the show more option is disabled.
- if the response height is  >  250px, we show the show  button, otherwise the show button is hidden.
- When a query is performed, the response is collapsed by default .
- When the answer is still generating, The size is still set to 250px, and the rest of the reponse hidden. We show the show more once the response is generated with the disclaimer.
- Feedback buttons are hidden when the answer is collapsed.
- When we click on show more, we show the full response.
- When we click on show less , the answer is collapsed.

Screenshot 2024-04-19 at 12 13 47 AM

Screenshot 2024-04-10 at 6 17 12 PM

Screenshot 2024-04-10 at 6 20 20 PM

Screenshot 2024-04-10 at 6 16 10 PM

When Answer is generating
Screenshot 2024-04-25 at 4 30 34 PM

https://github.com/coveo/ui-kit/assets/119955059/2ddc310c-af70-4933-a992-69589b1f1c57

Analytics:

Screenshot 2024-04-26 at 2 43 23 PM
Screenshot 2024-04-26 at 2 43 15 PM

[SVCC-3574]: https://coveord.atlassian.net/browse/SVCC-3574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/common/generated-answer/show-button.tsx
```diff
@@ -0,0 +1,31 @@
+import {FunctionalComponent, h} from '@stencil/core';
+import {i18n} from 'i18next';
+import ArrowBottomIcon from '../../../images/arrow-bottom-rounded.svg';
+import ArrowTopIcon from '../../../images/arrow-top-rounded.svg';
+import {Button} from '../button';
+
+interface ShowButtonProps {
+  i18n: i18n;
+  isCollapsed: boolean;
+  onClick(): void;
```
lbergeron, line 10: I'd suggest to write it this way. Both patterns are used in the project, but I see more occurrences of this one.

```suggestion
  onClick: () => void;
```

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
```diff
@@ -66,3 +75,30 @@
...
+  }
+
+  @screen mobile-only {
+    @mixin mobile-generated-answer-footer-layout;
+  }
+}
+
+[part='generated-container'] {
+  &.answer-collapsed {
+    @apply max-h-[250px] overflow-hidden relative content-[''];
```
lbergeron, line 98: There should be an option on the component to modify the collapsible height.

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
```diff
@@ -66,3 +75,30 @@
...
+  }
+  &.answer-collapsed:before {
+    @apply h-full w-full absolute top-0 left-0 content-[''];
+    background: linear-gradient(transparent 180px, white);
```
lbergeron, line 102: Considering the collapsible height can be modified, this gradient has to update accordingly.

packages/atomic/src/components/insight/atomic-insight-generated-answer/atomic-insight-generated-answer.tsx
```diff
@@ -142,6 +170,68 @@ export class AtomicInsightGeneratedAnswer
...
+      element.classList.add(className);
+    } else {
+      element.classList.remove(className);
+    }
+  }
+
+  private adaptAnswerHeight() {
+    if (this.isShowButtonDisplayed()) {
+      this.fullAnswerHeight = this.host?.shadowRoot
+        ?.querySelector('p[part="generated-text"')
```
lbergeron, line 188:
```suggestion
        ?.querySelector('p[part="generated-text"]')
```

packages/atomic/src/components/insight/atomic-insight-generated-answer/atomic-insight-generated-answer.tsx
```diff
@@ -142,6 +170,68 @@ export class AtomicInsightGeneratedAnswer
...
+        ?.getBoundingClientRect().height;
+      this.updateAnswerHeight();
+    }
+  }
+
+  private getShowButton() {
+    return this.host?.shadowRoot?.querySelector('[part="answer-show-button"');
```
lbergeron, line 195:
```suggestion
    return this.host?.shadowRoot?.querySelector('[part="answer-show-button"]');
```

packages/atomic/src/components/insight/atomic-insight-generated-answer/atomic-insight-generated-answer.tsx
```diff
@@ -142,6 +170,68 @@ export class AtomicInsightGeneratedAnswer
...
+  }
+
+  private getAnswerContainer() {
+    return this.host?.shadowRoot?.querySelector('[part="generated-container"');
```
lbergeron, line 199:
```suggestion
    return this.host?.shadowRoot?.querySelector('[part="generated-container"]');
```

packages/atomic/src/components/insight/atomic-insight-generated-answer/atomic-insight-generated-answer.tsx
```diff
@@ -142,6 +170,68 @@ export class AtomicInsightGeneratedAnswer
...
+  }
+
+  private getAnswerFooter() {
+    return this.host?.shadowRoot?.querySelector(
+      '[part="generated-answer-footer"'
```
lbergeron, line 204:
```suggestion
      '[part="generated-answer-footer"]'
```

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -142,6 +171,68 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
+      element.classList.add(className);
+    } else {
+      element.classList.remove(className);
+    }
+  }
+
+  private adaptAnswerHeight() {
+    if (this.isShowButtonDisplayed()) {
+      this.fullAnswerHeight = this.host?.shadowRoot
+        ?.querySelector('p[part="generated-text"')
```
lbergeron, line 189: There are a few CSS selectors with an unclosed square bracket in this file.

lbergeron (commented):
Added a few comments, but it looks good to me so far :+1:

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
jelmedini: I talked with Audrey, the collapsible height is not an option in the end, it will always be 250px

lbergeron: What would you think of making it `16em` tall so it can scale with the font size?

packages/atomic/cypress/fixtures/test-fixture.ts
```diff
@@ -298,3 +298,11 @@ export const generateComponentHTML = (tag: string, props: TagProps = {}) => {
...
   return e;
 };
+
+export const generateLongTextAnswer = () => {
+  let longText = '';
+  for (let i = 0; i < 200; i++) {
+    longText += 'This is my long answer. ';
+  }
+  return longText.trim();
+};
```
dmgauthier, line 308: Have fun and try `new Array(200).fill('This is a long answer').join(' ')`

image

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
```diff
@@ -91,6 +95,10 @@ export class GeneratedAnswerCommon {
     }
 
     if (isGenerating) {
+      if (this.props.collapsible) {
+        this.props.setIsCollapsed(true);
+      }
+
```
dmgauthier, line 101: Is there a reason why you wouldn't want the prop at all (either true or false).

Cuz this would be as straightforward as we could get. `this.props.setIsCollapsed(this.props.collapsible);`

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
```diff
@@ -313,18 +322,28 @@ export class GeneratedAnswerCommon {
...
+              <div class="text-neutral-dark text-xs">
+                <slot name="disclaimer" slot="disclaimer">
+                  {this.props
+                    .getBindings()
+                    .i18n.t('generated-answer-disclaimer')}
+                </slot>
+              </div>
+            )}
+          </div>
+        )}
```
dmgauthier, line 346: [ :rotating_light: nitpicking alert :rotating_light: ]

The short circuiting pattern inside a render (`true && ...render`) is making my soft brain hurt, usually.

When I am about to do this, this is usually the signal for me to split the component in more clear an consise units with each their own rendering condition(s).

so, in short I usually prefer this :point_down: to short circuiting. This is lengthier and dumber. But I guess I just like long dumb stuff.

```
const Patate = () => {
    if (!canRenderPatate) {
        return null;
    }
    return <div>Patate</div>;
};

const Pwell = () => {
    if (!canRenderPwell) {
        return null;
    }
    return <div>Pwell</div>;
};

const PatatePwell = () => (
    <div>
        <Patate />
        <Pwell />
    </div>
);
```

Roger out, I should not bother you again with this point of view of mine. If in most case you make sure the component is readable, its a matter of preferences in the end.

packages/atomic/src/components/common/generated-answer/show-button.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export const ShowButton: FunctionalComponent<ShowButtonProps> = (props) => {
+  const btnClasses = 'hidden items-center hover:bg-transparent';
+
+  return (
+    <Button
+      style="text-primary"
+      part="answer-show-button"
+      class={`${btnClasses} ${props.collapsible ? (props.isStreaming ? 'flex invisible' : 'flex') : 'hidden'}`}
```
dmgauthier, line 22: [clsx](https://www.npmjs.com/package/clsx) weights 8.5 kb.  Might worth a shot here.

```
const btnClasses = 'hidden items-center hover:bg-transparent';
const streamingClasses = props.isStreaming ? 'flex invisible' : 'flex'
[...]
class={clsx([btnClasses, streamingClasses, {hidden: props.collapsible}])}
```
Perfect for soft brain like me.

I get why we wouldn't want to add another lib in a client facing bundle. But I would argue that this is pretty small, and probably already used in most client side app nowaday when dlded 17 millions times a week.

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
```diff
@@ -66,3 +75,31 @@
     @mixin mobile-footer-xs-layout;
   }
 }
+
+[part='generated-answer-footer'] {
+  @container (max-width: 600px) {
```
dmgauthier, line 80: just curious, are hard coded pixel stuff the norm in atomic ? cuz it doesn't sound really responsive to me.
(Seeing myself out !  :running_man: )

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
jelmedini: yes it should be true all the time for each new generated answer

packages/atomic/cypress/e2e/generated-answer.cypress.ts
```diff
@@ -174,6 +174,70 @@ describe('Generated Answer Test Suites', () => {
...
+          setupGeneratedAnswerWithoutFirstIntercept(streamId, {
+            collapsible: true,
+          });
+        });
+
+        GeneratedAnswerAssertions.assertShowButton(true);
+        GeneratedAnswerAssertions.assertAnswerCollapsed(true);
+        GeneratedAnswerAssertions.assertShowMoreLabel(true);
+        GeneratedAnswerAssertions.assertFeedbackButtonsVisibility(false);
+        GeneratedAnswerAssertions.assertCopyButtonVisibility(false);
```
fbeaudoincoveo, line 201: I'm not sure we should hide the feeback and copy buttons...

What if somebody has found the answer they were looking for (or figured that the answer was not relevant to their question) in the first few sentences? Do we still want to force them to expand the answer to provide us with their feedback?

For both the feeback and copy buttons, suppose that the user has expanded (and read) the answer and then decides to collapse it again to have more vertical real estate to look at the result page. I don't think it makes sense to hide the buttons in this context either.

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -80,14 +85,22 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
    */
   @Prop() answerStyle: GeneratedAnswerStyle = 'default';
 
+  /**
+   * The answer is collapsed when the text is longer than 250px
```
fbeaudoincoveo, line 89:
```suggestion
   * Whether to allow the answer to be collapsed when the text is taller than 250px.
```

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -110,6 +125,30 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
     });
     this.searchStatus = buildSearchStatus(this.bindings.engine);
     this.generatedAnswerCommon.insertFeedbackModal();
+
+    if (window.ResizeObserver) {
```
fbeaudoincoveo, line 129: Should we only create the resize observer when this.collapsible is true?

```suggestion
    if (window.ResizeObserver && this.collapsible) {
```

packages/atomic/src/components/common/generated-answer/show-button.tsx
fbeaudoincoveo: I personally disagree; clsx would add 0 value for the end user and would only make our source code slightly more readable. It'd be one more dependency to maintain. I don't think it's worth it.

dmgauthier: This is exactly the code readability issue that I wanna stress here. I would argue that happy devs that are less inclined to do mistakes because they have to maintain nested ternaries are not 0 value for the end user. But I will not fight over it.

Here I would at least try to get rid of the imbedded ternary in the prop.

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
jelmedini: indeed very good point!

packages/atomic/cypress/e2e/generated-answer.cypress.ts
jelmedini: This was UX requirements :/ maybe @adroletCoveo has more insights on this

packages/atomic/src/components/common/generated-answer/show-button.tsx
olamothe: > Here I would at least try to get rid of the imbedded ternary in the prop.

Yes ! Also, `hidden` is there 2 times, but gets overriden (?) if `collapsible` ? 
This certainly smells fishy and not optimal. Please rework into something more manageable 🙏 .

As for adding a lib, the reasons for adding one in the bundler needs to be "bullet proof" and a "no brainer" in the context of Atomic.

I personally think we can very easily achieve what `clsx` does with just proper string building "discipline". I've looked at the doc of clsx and I must say I would not know what this does exactly: 

```
clsx('foo', [1 && 'bar', { baz:false, bat:null }, ['hello', ['world']]], 'cya');
```

VS just a standard string concatenation (without ternary)

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
olamothe: Yes agreed that this file is getting out of hand. It will not be easily re-usable in other context than classic search and insight panel very well as it stands right now.

ie: If/when we ever need it for commerce, we'll need to refactor this file.

packages/atomic/src/components/insight/atomic-insight-generated-answer/atomic-insight-generated-answer.tsx
```diff
@@ -142,6 +181,62 @@ export class AtomicInsightGeneratedAnswer
     this.ariaMessage = message;
   };
 
+  private setIsCollapsed = (isCollapsed: boolean) => {
+    this.isCollapsed = isCollapsed;
+  };
+
+  private toggleClass(element: Element, className: string, condition: boolean) {
```
olamothe, line 188: This whole section of the code is suspicious: Having to manipulate classes/get reference to DOM directly like this instead of just using the render loop as normal...


Why can't we just output the correct classes on every element in the render function depending on `isCollapsed` prop ?

olamothe (changes requested):

packages/atomic/src/components/common/generated-answer/generated-answer-common.tsx
jelmedini: Yes agreed, I added this section to match what we have until now, But I will change the whole file to be more readable

packages/atomic/src/components/common/generated-answer/show-button.tsx
dmgauthier: :+1: Thank you both :+1:

packages/atomic/src/components/insight/atomic-insight-generated-answer/atomic-insight-generated-answer.tsx
olamothe: For this example: `this.toggleClass(container, 'answer-collapsed', this.isCollapsed);`

Is there a reason that, as part of the `render()` loop function, is `this.isCollapsed===true`, we set the class `answer-collapsed` on the `container` ?

I don't understand why it is necessary to do it in this non-standard manner ?

jelmedini: we need to manipulate the DOM regarding if collapsible option is enabled and When we resize the answer! in this case the answer is either partially hidden with show more button displayed or displayed fully if the height of the answer < 250px or when it is not collapsed with show less displayed. 

the whole logic is checked regarding these conditions and different use cases and with that we inject the right class. we have basically the classes: `answer-collapsed` and `show-button-visible` and `is-collapsible` that we add/remove.

I don't get your point, what are you suggesting to make this work ?

dmgauthier: Like @jelmedini, In this situation I can't say how I could handle it better than this. TBH I really dislike how all of this has been built. This common class extension pattern and DOM mutation methods makes all of this way too convoluted for my taste.

I do not want to kick a wasp nest uselessly, but while Coveo RGA capabilities are quite new, I am curious as why we choose to use a kinda already dated React class pattern that could probably have been done way cleaner with a more modern functional approach like compound Components.

That being said... without any solution to propose. I will let you folks resolve this.

packages/atomic/cypress/e2e/generated-answer.cypress.ts
jelmedini: @VykaNguyen, as Audrey is not available for now, what do you think about this use case ? :)

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
lbergeron: I agree. In this case it means `37.5rem`.

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
```diff
@@ -66,3 +75,31 @@
...
+  }
+}
+
+[part='generated-container'] {
+  &.answer-collapsed {
+    @apply max-h-64 overflow-hidden relative content-[''];
+  }
+  &.answer-collapsed:before {
+    @apply h-full w-full absolute top-0 left-0 content-[''];
+    background: linear-gradient(transparent 180px, white);
```
lbergeron, line 95: We should use a relative dimension here.

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -142,6 +181,62 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
+    this.isCollapsed = isCollapsed;
+  };
+
+  private toggleClass(element: Element, className: string, condition: boolean) {
+    if (condition) {
+      element.classList.add(className);
+    } else {
+      element.classList.remove(className);
+    }
+  }
```
lbergeron, line 194: Unless I am mistaken, you can use [classList.toggle](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle) to achieve the same result.

```
element.toggle(className, condition)
```

jelmedini: so much cleaner thanks!

packages/atomic/cypress/e2e/generated-answer.cypress.ts
jelmedini: update: For now we will try it like this, and regarding the client feedbacks we could change this!

packages/atomic/cypress/fixtures/test-fixture.ts
erocheleau: So unoriginal.... 

```
const testLongText = `Call me Ishmael. Some years ago—never mind how long precisely—having little or no money in my purse, and nothing particular to interest me on shore, I thought I would sail about a little and see the watery part of the world. It is a way I have of driving off the spleen and regulating the circulation. Whenever I find myself growing grim about the mouth; whenever it is a damp, drizzly November in my soul; whenever I find myself involuntarily pausing before coffin warehouses, and bringing up the rear of every funeral I meet; and especially whenever my hypos get such an upper hand of me, that it requires a strong moral principle to prevent me from deliberately stepping into the street, and methodically knocking people’s hats off—then, I account it high time to get to sea as soon as I can. This is my substitute for pistol and ball. With a philosophical flourish Cato throws himself upon his sword; I quietly take to the ship. There is nothing surprising in this. If they but knew it, almost all men in their degree, some time or other, cherish very nearly the same feelings towards the ocean with me.
There now is your insular city of the Manhattoes, belted round by wharves as Indian isles by coral reefs—commerce surrounds it with her surf. Right and left, the streets take you waterward. Its extreme downtown is the battery, where that noble mole is washed by waves, and cooled by breezes, which a few hours previous were out of sight of land. Look at the crowds of water-gazers there.`;
```

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -80,14 +85,22 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
...
+   * The answer is collapsed when the text is longer than 250px
+   */
+  @Prop() collapsible?: boolean;
+
   @AriaLiveRegion('generated-answer')
   protected ariaMessage!: string;
 
   private generatedAnswerCommon!: GeneratedAnswerCommon;
+  private fullAnswerHeight?: number;
+  private maxCollapsedHeight = 250;
```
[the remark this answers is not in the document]
erocheleau: This is not so much relative to the height of the viewport, but more related to the height of the content of the answer.
This could be transformed in a max height in `em` for example, but it would make it very hard I think to calculate in js instead of in CSS...

packages/atomic/cypress/e2e/generated-answer.cypress.ts
erocheleau: > What if somebody has found the answer they were looking for (or figured that the answer was not relevant to their question) in the first few sentences? Do we still want to force them to expand the answer to provide us with their feedback?

Yes absolutely, perhaps the first few lines are not relevant but the rest is VERY relevant. Also how valuable is the feedback of someone who didn't even bother to expand the answer to read more than the first lines?

> For both the feeback and copy buttons, suppose that the user has expanded (and read) the answer and then decides to collapse it again to have more vertical real estate to look at the result page. I don't think it makes sense to hide the buttons in this context either.

It would make it much more complex if we had to track if the user has ever expanded the answer, but then decided to collapse it, to still allow them to give us feedback. To gain probably little to no value for the few who would expand > collapse > want to give feedback after this...

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
```diff
@@ -68,13 +73,51 @@
...
     @mixin mobile-footer-layout;
   }
-  @container (max-width: 400px) {
+  @container (max-width: 25rem) {
     @mixin mobile-footer-xs-layout;
   }
 }
+
+[part='generated-answer-footer'] {
+  @container (max-width: 600px) {
```
lbergeron, line 88: Is there a reason why this dimension is still in pixels?

packages/atomic/src/components/search/atomic-generated-answer/atomic-generated-answer.tsx
```diff
@@ -142,6 +184,54 @@ export class AtomicGeneratedAnswer implements InitializableComponent {
     this.ariaMessage = message;
   };
 
+  private toggleClass(element: Element, className: string, condition: boolean) {
+    element.classList.toggle(className, condition);
+  }
+
+  private adaptAnswerHeight() {
+    this.fullAnswerHeight = this.host?.shadowRoot
+      ?.querySelector('p[part="generated-text"]')
```
lbergeron, line 193: I'd suggest to remove the element from the selector. For rich formatting, `generated-text` may either be `<p>` or a `<div>` depending if it is plain text or not.

```suggestion
      ?.querySelector('[part="generated-text"]')
```

lbergeron (approved):
It looks good to me! Well done!

dmgauthier (approved):

packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
jelmedini: good catch! forgot to change this one, thanks!!

olamothe (approved):

PR #3786, opened 2024-04-11
chore(ci): always report failure or success

louis-bompart:
This ensure the status is always reported, and always properly reported

### DIFF
--- a/.github/workflows/prbot.yml
+++ b/.github/workflows/prbot.yml
@@ -245,8 +245,9 @@ jobs:
     secrets:
       SFDX_AUTH_CLIENT_ID: ${{ secrets.SFDX_AUTH_CLIENT_ID }}
       SFDX_AUTH_JWT_KEY: ${{ secrets.SFDX_AUTH_JWT_KEY }}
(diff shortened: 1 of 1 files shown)


fbeaudoincoveo (commented):
As discussed, this will most likely cause us pain when flaky e2e tests fail for no good reason, and an otherwise fine change set gets ejected from the merge queue only to be merged again as is, and then potentially get re-ejected again because of another flaky test and so on.

On the other hand, this is arguably better than the false sense of security we have right now.

Our principal flakiness issue is with the Quantic e2e tests. I think what we should do for now is avoid running those tests on merge group. This is not ideal, but IMO It's better than disabling those tests altogether... or being trapped in a merge groundhog day on almost every Headless PR.

One habit we could possibly take as a team (I wouldn't make this a formal process, really just a habit) would be to at least take note of flaky tests when we encounter one. E.g., we could create a tech debt Jira and indicate the cause of failure from the logs. Over time this could lead us to uncover (and address) systematic issues.

fbeaudoincoveo (approved):

PR #3791, opened 2024-04-12
feat: add nonce support

louis-bompart:
- [x] Test todo. #3802 

Add 1st party support for `nonce.`
Essentially, this is used for CSP _without_ ` unsafe-inline,` which prevents the execution of inline scripts or style.

The issue is two folds:
 - Stencil injects a bunch of styles. This is solved by using `setNonce` from Stencil https://stenciljs.com/docs/csp-nonce
 - We inject a few style tags. To solve that, I did the following:
   - Added a `nonce` parameter on `atomic-*-search-interface` (for the one that needed it/used the bindings).
   - Added a new `createStyleElement` function in the binding that wraps `document.createElement` to create a style tag _with_ the nonce parameter
   - Replace calls to `document.createElement('style')` with this.
   - Bonus, added a `createScript` equivalent.


packages/atomic/src/components/common/interface/bindings.ts
```diff
@@ -47,6 +47,20 @@ export interface CommonBindings<
    * A reference to the atomic interface element.
    */
   interfaceElement: InterfaceElement;
+
+  /**
+   * Creates a Style Information element with the nonce attribute if it exists.
+   * Note: the element is **not** appended to the DOM.
+   * @returns a style element.
+   */
+  createStyleElement: () => HTMLStyleElement;
```
olamothe, line 56: Other alternatives to add these function would be:

* Utility function on the `store`, which can be specific (ie: not the "common" store)
* `@Method()` on the search/insight interface itself

I think these might be better instead of adding them to the "base" interface, in order to limit its scope/narrow the surface a bit.

So, for a given type of `atomic-xyz-interface` component, we can decide to add support for it when necessary, and not force it for all types of `atomic-xyz-interface`.

packages/atomic/src/components/insight/atomic-insight-interface/atomic-insight-interface.tsx
```diff
@@ -112,6 +113,11 @@ export class AtomicInsightInterface
    */
   @Prop({reflect: true}) resultsPerPage = 5;
 
+  /**
+   * Doc TODO
```
olamothe, line 117: Do not merge with this unresolved.

packages/atomic/src/components/common/interface/bindings.ts
olamothe: The JSDoc would also greatly benefit from MDN links and/or wording that explain what a nonce is, what type of error it could help solve, etc.

To give an example, if an implementer sees `CSP` directive error from their web browser console, as was shown in the discuss post, they might be tempted to simply copy paste the error message in our documentation. It would be neat if this documentation would pop up correctly.

olamothe (approved):

packages/atomic/src/components/common/interface/bindings.ts
fpbrault: What are some examples of situations in which an implementer may use this feature?

packages/atomic/src/components/insight/atomic-insight-interface/atomic-insight-interface.tsx
fbeaudoincoveo: Suggestion:

The value to set the [`nonce`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) attribute to on inline `script` and `style` elements generated by this component.

If your application is served with a Content Security Policy (CSP) that doesn't include the `script-src: 'unsafe-inline'` or `style-src: 'unsafe-inline'` directives, you must ensure that your application server generates a new nonce on every page load and uses the generated value to set this prop and serve the corresponding CSP response headers (i.e., `script-src 'nonce-<YOUR_GENERATED_NONCE>'` and `style-src 'nonce-<YOUR_GENERATED_NONCE>'`). Otherwise you may see console errors such as `Refused to execute inline script because it violates the following Content Security Policy directive: [...]` or `Refused to apply inline style because it violates the following Content Security Policy directive: [...]`.
```

fbeaudoincoveo (approved):

packages/atomic/src/components/common/interface/bindings.ts
louis-bompart: > @Method() on the search/insight interface itself

I don't think so, as those methods are potentially required on any component (I think digging up to get the atomic-interface and its method would be a bit crude).

> Utility function on the store, which can be specific (ie: not the "common" store)

I thought about that, but it'd denature the store nature of the store a bit, which is why I went with the bindings.

> We can decide to add support for it when necessary, and not force it for all types of atomic-xyz-interface.

This is a good point, I'll see what I can do to reduce the sprawl and target only the main interface.

louis-bompart: > What are some examples of situations in which an implementer may use this feature?

The most detailed one is on [Discuss](https://discuss.coveo.com/t/support-00111577-using-coveo-atomic-with-content-security-policy-csp-without-unsafe-inline/11103)

olamothe: > I don't think so, as those methods are potentially required on any component (I think digging up to get the atomic-interface and its method would be a bit crude).

There should not be any digging necessary 🤔 The search interface reference is already passed in bindings to all components

louis-bompart:
I created a new branch for automated testing: I can't use Cypress easily because I need something that does not interfere with/inject stuff in the page and a dedicated server configuration.

packages/atomic/src/components/common/interface/bindings.ts
louis-bompart: Split into NonceBindings

louis-bompart:
@coveo/service-core, are you fine with the current changes or do you prefer to opt out of the nonce support?

PR #3796, opened 2024-04-15
test(quantic): smart snippet tests adapted with new event protocol

mmitiche:
## [SFINT-5454](https://coveord.atlassian.net/browse/SFINT-5454)

- Adapted the Cypress e2e test for smart snippet and smart snippet suggestions features  for testing with analytics sent with the new protocol, now the test spec tests for both cases: the case when send legacy analytics and the case where we sent next analytics.

```
                    Expect.logExpandSmartSnippetSuggestion({
                      answerSnippet: suggestion.answerSnippet,
                      question: suggestion.question,
                      documentId: suggestion.documentId,
                    });
```
```
                  if (analyticsMode === 'next') {
                    NextAnalyticsExpectations.emitQnaAnswerActionEvent(
                      {
                        answer: {
                          type: answerType,
                        },
                        action: 'expand',
                      },
                      exampleTrackingId
                    );
                  } else {
                    Expect.logExpandSmartSnippetSuggestion({
                      answerSnippet: suggestion.answerSnippet,
                      question: suggestion.question,
                      documentId: suggestion.documentId,
                    });
                  }

```

[SFINT-5454]: https://coveord.atlassian.net/browse/SFINT-5454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/cypress/e2e/default-2/smart-snippet-suggestions/smart-snippet-suggestions.cypress.ts
```diff
@@ -90,88 +103,165 @@ describe('quantic-smart-snippet-suggestions', () => {
...
-                question: suggestion.question,
-                documentId: suggestion.documentId,
+          describe('when the query returns smart snippet suggestions', () => {
+            it('should properly display the smart snippet suggestions', () => {
+              visitPage({useCase: param.useCase});
+
+              scope('when loading the page', () => {
+                Expect.displaySmartSnippetSuggestionsCard(true);
+                exampleRelatedQuestions.forEach((suggestion, index) => {
+                  Actions.toggleSuggestion(index);
```
erocheleau, line 119: This test is a bit weird no?

Why are we testing expand/collapse in the "when loading the page" test suite?

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet.cypress.ts
```diff
@@ -79,159 +92,291 @@ describe('quantic-smart-snippet', {browser: 'chrome'}, () => {
...
+                scope('when loading the page', () => {
+                  Expect.displaySmartSnippetCard(true);
+                  Expect.displaySmartSnippetQuestion(
+                    exampleSmartSnippetQuestion
+                  );
+                  Expect.displaySmartSnippetAnswer({
+                    text: exampleAnswerText,
+                    link: {
+                      text: exampleInlineLinkText,
+                      href: `${location.origin}/examples/${pageUrl}#`,
```
erocheleau, line 115: Why is this changing?
Seems weird that the location.origin is used to determine the href of the SS source

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet.cypress.ts
```diff
@@ -79,159 +92,291 @@ describe('quantic-smart-snippet', {browser: 'chrome'}, () => {
...
-                title: exampleSmartSnippetSourceTitle,
-                uri: exampleSmartSnippetSourceUri,
-                permanentId: examplePermanentId,
+                    Actions.clickSmartSnippetInlineLink();
+                    if (analyticsMode === 'legacy') {
+                      Expect.logOpenSmartSnippetInlineLink({
+                        title: exampleSmartSnippetSourceTitle,
+                        uri: exampleSmartSnippetSourceUri,
+                        permanentId: examplePermanentId,
+                        linkUrl: `${location.origin}/examples/${pageUrl}${exampleInlineLinkUrl}`,
```
erocheleau, line 193: Same question as above

erocheleau (approved):
Some general questions but everything is good with the PR.

SimonMilord (approved):

packages/quantic/cypress/e2e/default-2/smart-snippet/smart-snippet.cypress.ts
mmitiche: Before this change, we were using a random url as the SS source, it was for example `https://www.coveo.com/en`.

The problem with that, is first I noticed some flakiness when doing that originating from an internal cypress error that sometimes occurs when from a page you navigate to another page and then you come back to your page, this problem was rarely occurring tho, so that's not the main reason of doing this.

The main reason is that I wanted to simplify the tests and avoiding navigating to multiple pages, so I wanted to make the SS source points towards `#`  so when we click on it we will stay on the same page.

when we set the herf of the ss source to `#`, `location.origin` is automatically set as the herf to make the link navigate to the same page and that's why I'm using it here.

PR #3802, opened 2024-04-15
test: atomic w/ strict CSP

louis-bompart:
- Add a server.mjs that serves what stencil dev server was doing, but with CORS (and no inline-script/style)
- Copy-paste index.html to page/csp.html, with nonce everywhere
- Make header.js nonce-compatible.

KIT-3086


packages/atomic/src/pages/examples/csp.html
louis-bompart, line 1: A copy of index.html, but with nonce everywhere, and a twist (see below)

packages/atomic/src/pages/examples/csp.html
```diff
@@ -0,0 +1,388 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />
+    <title>Coveo Atomic</title>
+    <link rel="icon" href="data:," />
+    <link rel="stylesheet" href="/themes/coveo.css" />
+    <script type="module" nonce="1234567890">
+      import {setNonce} from '/build/atomic.esm.js';
+
+      setNonce('1234567890');
```
louis-bompart, line 12: We change how we import Atomic. If we call setOnce after we imported atomic in a different script, we'll still have some CSP error.
From what I experienced, if we do not call `setOnce` in the same microtask as the import of atomic, some style tag will slip through and end up without nonce.

packages/atomic/src/pages/header.js
```diff
@@ -33,22 +33,45 @@ const makeLinks = () => {
...
+        display: inline-block;
+        list-decoration: none;
+      }
+      a {
+        margin-right: 10px;
+        color: var(--atomic-primary);
+        text-decoration: none;
+      }
+    }`;
+styleTag.setAttribute('nonce', '1234567890');
```
louis-bompart, line 66: `style=` attribute ain't allowed, without `unsafe-inline`. So I moved the style somewhere else, with a nonce.

packages/atomic/csp/test.mjs
louis-bompart, line 1: Cypress add quite the overhead and I didn't manage in a reasonable amount of time to make it work with my server.
So let's KISS: Puppeteer for browser instrumentation, and vanilla NodeJS for the rest.

packages/atomic/csp/server.mjs
louis-bompart, line 1: Decided not to go with express/helmet to keep things unobfuscated and with the 'bare minimum' of CSP to make things positively breaks (so that the test can be made)

packages/atomic/cypress/integration-insight-panel/insight-panel-selectors.ts
```diff
@@ -44,7 +44,7 @@ export const InsightPanelsSelectors = {
     InsightPanelsSelectors.tabPopover().find('[part="popover-button"]'),
   layoutStyleTags: () =>
     InsightPanelsSelectors.interface().find('atomic-insight-layout > style'),
-  topLevelStyleTags: () => cy.get('style[data-styles]'),
+  topLevelStyleTags: () => cy.get('head > style'),
```
louis-bompart, line 47: Small tweaks to the e2e to take into account the changes in header.js

packages/atomic-hosted-page/src/pages/index.html
```diff
@@ -6,19 +6,17 @@
...
-        // TODO: switch for prod - searchuisamples
-        await page.initialize({
-          pageId: '88411b30-39fb-4bd3-8663-2d1af2b27659',
-          accessToken: 'xx71d82ce9-f6fe-4be0-9209-c0e43f6fae81',
-          organizationId: 'searchuisamples',
-          organizationEndpoints: await page.getOrganizationEndpoints('searchuisamples'),
-        });
-      })();
+      const page = document.querySelector('atomic-hosted-page');
+      // TODO: switch for prod - searchuisamples
```
fbeaudoincoveo, line 13: I'm not sure I understand what this TODO is for; isn't prod searchuisamples already where we're poiting?

packages/atomic/csp/server.mjs
```diff
@@ -0,0 +1,63 @@
+import {createReadStream} from 'node:fs';
+import {createServer} from 'node:http';
+import {resolve} from 'node:path';
+
+const __dirname = import.meta.dirname;
+
+function tableFlip(res) {
```
fbeaudoincoveo, line 7: ❤️

packages/atomic/csp/server.mjs
```diff
@@ -0,0 +1,63 @@
...
+  res.statusCode = 404;
+  res.end('File not found');
+}
+
+function getContentTypeFromExtension(filePath) {
+  const extension = /\.(\w+)$/.exec(filePath);
```
fbeaudoincoveo, line 13:
```suggestion
  const extension = /\.(html|json|js|css|svg|ico)$/.exec(filePath);
  
  if (!extension) {
    return;
  }
```

packages/atomic/src/components/recommendations/atomic-recs-interface/atomic-recs-interface.tsx
```diff
@@ -140,6 +141,24 @@ export class AtomicRecsInterface
...
 
+  /**
+   * The value to set the [nonce](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) attribute to on inline script and style elements generated by this interface and its child components.
+   * If your application is served with a Content Security Policy (CSP) that doesn't include the `script-src: 'unsafe-inline'` or `style-src: 'unsafe-inline'` directives,
+   * you should ensure that your application server generates a new nonce on every page load and uses the generated value to set this prop and serve the corresponding CSP response headers
+   * (i.e., script-src 'nonce-<YOUR_GENERATED_NONCE>' and style-src 'nonce-<YOUR_GENERATED_NONCE>').
+   * Otherwise you may see console errors such as
+   *  - Refused to execute inline script because it violates the following Content Security Policy directive: [...]
+   *  - Refused to apply inline style because it violates the following Content Security Policy directive: [...].
+   * @example:
```
fbeaudoincoveo, line 152: I think we should add

> When using a nonce, the first import of Atomic should be to import & call the `setNonce` function with the generated nonce value.

In the documentation each time this prop is exposed on a component.

fbeaudoincoveo (approved):

packages/atomic/src/components/recommendations/atomic-recs-interface/atomic-recs-interface.tsx
louis-bompart: recs was not ported, that was some artifacts

packages/atomic/src/pages/examples/csp.html
fpbrault: I guess we don't need to change the title since this page is not included in the header?

olamothe (approved):

packages/atomic-hosted-page/src/pages/index.html
```diff
@@ -10,7 +10,6 @@
       await customElements.whenDefined('atomic-hosted-page');
 
       const page = document.querySelector('atomic-hosted-page');
-      // TODO: switch for prod - searchuisamples
       await page.initialize({
```
fpbrault, line 13: Why is this included in the PR?

PR #3803, opened 2024-04-16
chore: use module script everywhere w/ top-level-await

louis-bompart:
- Uniformize/simplify all HTML sample w/ top-level-awaits & type=module.
- Unformize all usage of `url()` CSS function to use root-based URL.


packages/atomic/src/pages/accessibility/commerce-full.html
```diff
@@ -8,98 +8,96 @@
...
-                    parentsElement.appendChild(parentElement);
-                  });
-                result.raw.parents = '<?xml version="1.0" encoding="utf-16"?>' + parentsElement.outerHTML;
-              });
-              return response;
-            },
+      await searchInterface.initialize({
+        accessToken: 'xxc23ce82a-3733-496e-b37e-9736168c4fd9',
+        organizationId: 'electronicscoveodemocomo0n2fu8v',
+        platformUrl: 'https://platform.cloud.coveo.com',
```
olamothe, line 18: Should modify this sample to also use `organizationEndpoints` instead of `platformUrl` (deprecated)

olamothe: Moreover.... this whole file is dubious, filled with "hacks". IIRC this is an artefact leftover from the accessibility testing with Deque.

I would not be opposed to simply deleting it, since this is not how "commerce" should be done, in the future.

olamothe (approved):

PR #3805, opened 2024-04-16
chore: update setup action to include fail-on-cache-miss input

y-lakhdar:
https://coveord.atlassian.net/browse/KIT-3089

Prevent the following error whenever the release takes too long
image

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -39,7 +39,7 @@ runs:
       with:
         path: node_modules/.cache/nx
         key: nx-${{ github.sha }}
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -35,7 +39,7 @@ runs:
       with:
         path: node_modules/.cache/nx
         key: nx-${{ github.sha }}
-        fail-on-cache-miss: true
+        fail-on-cache-miss: ${{ inputs.fail-on-cache-miss }}
```
louis-bompart, line 42: You should do `== 'true'` I think

louis-bompart (changes requested):

PR #3807, opened 2024-04-17
fix: patch event source to safeguard window.* calls

louis-bompart:
Ensure `window.*`, `document.*` and the like don't blow up in our faces while running on NodeJS

https://coveord.atlassian.net/browse/KIT-3097


fpbrault (approved):
I remember you mentioned these patches before, but I would like to know how they get applied.

louis-bompart:
> I remember you mentioned these patches before, but I would like to know how they get applied.

Good question :)
We use https://www.npmjs.com/package/patch-package
You can see it set up https://github.com/coveo/ui-kit/blob/master/package.json#L5

olamothe:
Was this reported by someone/a customer ?

olamothe (approved):

PR #3808, opened 2024-04-17
feat(commerce): namespace pagination on a per-solution-type basis

Spuffynism:
To make the pagination solution-type specific, I introduce the concept of a `solutionTypeId` to access to pagination slice. We could also name this `sliceName`, `slice`, or `namespace`! Open to suggestions!

I also considered registering reducers on a **slice** instead of on the state section directly, but couldn't easily make it work, which is how I landed with the current approach. Registering reducers for a slice would allow us to leave the reducers relatively unchanged, apart from the fact that their state would be a slice's state, and not a state section. 
Here's pseudocode of how this would look like:
```ts
engine.addReducer({
  [solutionTypeId]: {
    ...currentSolutionTypeReducers,
    paginationReducer
  }
});

// then, we'd have to find a way to let actions be dispatched on a specific *slice*... something like this
engine.dispatch(action, solutionTypeId);
```

Keeping this as draft because I want feedback on the approach before continuing further.

[CAPI-719]

[CAPI-719]: https://coveord.atlassian.net/browse/CAPI-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/features/commerce/common/actions.ts
+++ b/packages/headless/src/features/commerce/common/actions.ts
@@ -1,3 +1,5 @@
+import {SchemaDefinition} from '@coveo/bueno';
+import {createAction} from '@reduxjs/toolkit';
 import {getVisitorID} from '../../../api/analytics/coveo-analytics-utils';
 import {SortParam} from '../../../api/commerce/commerce-api-params';
 import {
@@ -20,6 +22,10 @@ import {
   RecommendationsSection,
   VersionSection,
 } from '../../../state/state-sections';
+import {
+  requiredNonEmptyString,
+  validatePayload,
+} from '../../../utils/validate-payload';
 import {PreparableAnalyticsAction} from '../../analytics/analytics-utils';
 import {StateNeededByFetchProductListingV2} from '../product-listing/product-listing-actions';
 import {SortBy, SortCriterion} from '../sort/sort';
@@ -47,6 +53,7 @@ export interface QueryCommerceAPIThunkReturn {
 }
 
 export const buildBaseCommerceAPIRequest = async (
+  solutionTypeId: string,
   state: StateNeededByQueryCommerceAPI
 ): Promise<BaseCommerceAPIRequest> => {
   const {view, user, ...restOfContext} = state.commerceContext;
@@ -62,20 +69,21 @@ export const buildBaseCommerceAPIRequest = async (
       view,
       cart: state.cart.cartItems.map((id) => state.cart.cart[id]),
     },
-    ...(state.commercePagination && {
-      page: state.commercePagination.page,
-      ...(state.commercePagination.perPage && {
-        perPage: state.commercePagination.perPage,
+    ...(state.commercePagination?.[solutionTypeId] && {
+      page: state.commercePagination[solutionTypeId].page,
+      ...(state.commercePagination[solutionTypeId].perPage && {
+        perPage: state.commercePagination[solutionTypeId].perPage,
       }),
     }),
   };
 };
 
 export const buildCommerceAPIRequest = async (
+  solutionTypeId: string,
   state: StateNeededByQueryCommerceAPI
 ): Promise<CommerceAPIRequest> => {
   return {
-    ...(await buildBaseCommerceAPIRequest(state)),
+    ...(await buildBaseCommerceAPIRequest(solutionTypeId, state)),
     facets: getFacets(state),
     ...(state.commerceSort && {
       sort: getSort(state.commerceSort.appliedSort),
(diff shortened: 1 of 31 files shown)


packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -62,20 +69,21 @@ export const buildBaseCommerceAPIRequest = async (
...
       cart: state.cart.cartItems.map((id) => state.cart.cart[id]),
     },
-    ...(state.commercePagination && {
-      page: state.commercePagination.page,
-      ...(state.commercePagination.perPage && {
-        perPage: state.commercePagination.perPage,
+    ...(state.commercePagination?.[solutionTypeId] && {
+      page: state.commercePagination[solutionTypeId].page,
+      ...(state.commercePagination[solutionTypeId].perPage && {
+        perPage: state.commercePagination[solutionTypeId].perPage,
```
louis-bompart, line 75: If we want to extends the principle to others things, such as facet and the like, should we instead have `state[solutionTypeId].commercePagination.*` ?

louis-bompart: Because we could have several productListing and recs, right?

Spuffynism: Yup!

Spuffynism: Well, we have to support multiple recs slots on the same page.

For listing and search, there should technically only be one per page. However, by supporting multiple recs slots, I think the effort to support multiple listings or multiple search pages would be minimal.

Spuffynism: I had a go at using multiple reducers for multiple slices: https://github.com/coveo/ui-kit/pull/3814

It's in an unfinished state, but I think it demonstrates the approach pretty well. I'm not convinced the added complexity and necessary changes to headless internals are worth it.

fbeaudoincoveo: I understand that allowing multiple listing / search interfaces to be supported in the same engine would require minimal effort, and I appreciate the elegance of the proposed solution.

However, what I'm hesitant about is that it would be making it easy to implement odd things that I believe should only be done exceptionally, if at all. This could confuse people into thinking that they _should_ typically be displaying multiple parallel listing / search interfaces, which is definitely not what we want.

If someone thinks that they need two product listing pages to be displayed simultaneously, I personally believe we should challenge that idea because, well, it's weird and unusual. If they still really want to do it but we still feel it's an edge case, we could always tell them to use two engines. If they do manage to convince us that this is a legitimate and common enough use case (or if a sufficient number of people ask for this feature), then we could consider allowing it for everyone. But I think our core assumption should be that this is weird and should normally not happen.

Now regarding facets and sort, I think we only need them to be at one place in the state. A recommendation interface should never have facets or sorting, and only one product listing or search interface should ever be displayed at any given time.

So, I think the issue we need to solve, really, is pagination, because there can and will be multiple paginated interfaces displayed simultaneously. One very simple and pragmatic way we could deal with this would be to not use the pagination controller on recommendation interfaces at all, and instead make pagination a core feature of the recommendation controller. This way we'd have a set of recommendations, each with its own pagination.

Spuffynism: Excellent point! I 100% agree. I'll have a go at your proposed approach!

fbeaudoincoveo: @louis-bompart having discussed this briefly live with you I know that you don't 100% agree with the idea of having pagination as a core feature of the recommendations controller, can you elaborate on your position?

louis-bompart: I agree with most of https://github.com/coveo/ui-kit/pull/3808#discussion_r1570857058, apart from the solution proposed in the last paragraph.
In my opinion, it's still important to keep things divided because if a user, for example, needs to play with the pagination of its recs; I do think it's best if they can do so with a dedicated controller instead of being able to control the whole recs. 

So for the state, I don't think it changes anything, we're still using `state.commerceRecommendations[solutionTypeId].commercePagination`
However, it's modified with a dedicated controller (and ofc each instance of the controller can only access/modify the pagination of one solution).

Spuffynism: ☝️ Gave this approach a try in https://github.com/coveo/ui-kit/pull/3825

PR #3809, opened 2024-04-17
feat(headless-cart): allow both productid and SKU when updating a cart item with the cart controller

sallainCoveo:
[LENS-1761](https://coveord.atlassian.net/browse/LENS-1761)

It was noticed that the `productId` emitted with the cart events (`ec.cartAction` & `ec.purchase`) in Barca sports was actually the sku of the product. To guarantee the uniqueness of the cart item in the state and that we send the right information to EP, we decided to add the `sku` property to the `updateItem` action.

Demo of linking the solution to Barca sports

https://github.com/coveo/ui-kit/assets/58052881/524973b7-ab52-4072-9b38-ce955f96d174

[LENS-1761]: https://coveord.atlassian.net/browse/LENS-1761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/api/commerce/commerce-api-params.ts
+++ b/packages/headless/src/api/commerce/commerce-api-params.ts
@@ -57,7 +57,7 @@ export type UserParams = (UserIdRequired | EmailRequired | UserIdAndEmail) & {
 };
 
 export interface CartItemParam {
-  productId: string;
+  sku: string;
   quantity: number;
 }
 
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
@@ -38,8 +38,8 @@ export interface CartItem {
   productId: string;
 
   /**
-   * The unique identifier of the product's variant.
-   * If the item is not a variant, set the same value as the productId.
+   * The stock keeping unit for the item being added to cart.
+   * Depending on how your catalog is structured, this may be the same value as the productId.
    */
   sku: string;
 
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
@@ -191,7 +191,7 @@ export function buildCart(engine: CommerceEngine, props: CartProps = {}): Cart {
     currentItem: CartItem,
     prevItem: CartItemWithMetadata | undefined
   ): Ec.CartAction {
-    const {quantity: currentQuantity, ...product} = currentItem;
+    const {quantity: currentQuantity, sku, ...product} = currentItem;
     const action = getCartAction(currentItem, prevItem);
     const quantity = !prevItem
       ? currentQuantity
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
@@ -38,6 +38,11 @@ export interface CartItem {
   productId: string;
 
   /**
+   * The unique identifier of the product variant.
+   */
+  sku: string;
+
+  /**
    * The human-readable name of the product.
    */
   name: string;
@@ -215,7 +220,7 @@ export function buildCart(engine: CommerceEngine, props: CartProps = {}): Cart {
     },
 
     updateItem(item: CartItem) {
-      const prevItem = itemSelector(getState(), item.productId);
+      const prevItem = itemSelector(getState(), item.sku);
       const doesNotNeedUpdate = !prevItem && item.quantity <= 0;
 
       if (doesNotNeedUpdate || isEqual(item, prevItem)) {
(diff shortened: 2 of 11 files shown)


sallainCoveo:
I think this solution would introduce a breaking change, is the commerce-engine still internal?

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -37,6 +37,11 @@ export interface CartItem {
    */
   productId: string;
 
+  /**
+   * The unique identifier of the product variant.
+   */
+  sku: string;
```
Spuffynism, line 43: Should we make this optional? Sometimes product's don't have variants. We could fallback to the `productId` when the `sku` isn't provided.

packages/headless/src/api/commerce/commerce-api-params.ts
```diff
@@ -57,7 +57,7 @@ export type UserParams = (UserIdRequired | EmailRequired | UserIdAndEmail) & {
 };
 
 export interface CartItemParam {
-  productId: string;
+  sku: string;
```
Spuffynism, line 60: Would naming this `variantId` cause more confusion? `variantId` seems more generic than `sku`.
```suggestion
  variantId: string;
```

Spuffynism (commented):
No need to worry about breaking changes, the engine is still internal.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
louis-bompart: Interesting point. The data guru of Commerce should validate that.
My guess/take is that "A product MUST have a variant"

"Any color the customer wants, as long as it's black."

sallainCoveo: I've updated the description to indicates that IF the cart item is not a variant, then set the same value as the product ID. 

Technically, not all products has variants (you could have a catalog that has only products), so it's true that the `sku` might be irrelevant in these cases. 

I'm open to set as optional, but my main concern is that people could forget to set this attribute because of it (resulting of bad attribution). If you set this attribute as required, but specify that it could be the same value as `productId`, I _think_ it's less prone to implementation error.

packages/headless/src/api/commerce/commerce-api-params.ts
sallainCoveo: That's a good question. I think I saw more often `sku`, but I might be tainted with the commerce ecosystem?

fbeaudoincoveo: I'd tend to use `sku` because if I'm not mistaken, that's the field name in the default catalog configuration, and this is the field we use in our documentation, e.g.: https://docs.coveo.com/en/m53g0506/coveo-for-commerce/catalog-variant-data-and-product-groupings

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
fbeaudoincoveo: > My guess/take is that "A product MUST have a variant"

It's possible to configure a catalog without variants, see https://docs.coveo.com/en/3139/coveo-for-commerce/commerce-catalog#add-a-catalog

fbeaudoincoveo (approved):
💯

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -37,6 +37,12 @@ export interface CartItem {
    */
   productId: string;
 
+  /**
+   * The unique identifier of the product's variant.
+   * If the item is not a variant, set the same value as the productId.
```
samisayegh, line 42: Could it be clearer if we rephrased to:

```
The stock keeping unit for the item being added to cart.
Depending on how your catalog is structured, this may be the same value as the productId.
```

Reasoning:
- sku implies a unique identifier, so there isn't a need to specify this explicitly
- By asking for a "stock keeping unit" we are guiding the implementer to pass a value that semantically means "sku" rather than a different unique identifier.
- Depending on how the catalog is structured, this might be the same value as `productId`. However, it is possible for a customer to store a unique identifier in `permanentid` (productId) field, and to also send the sku in `ec_sku`, in a products-only catalog. Not sure if this is happening in practice, but being vague ("it depends on the catalog"), we are guiding the reader to send a sku, while still reassuring them that it's ok if it's the same value as the `productId` field.

samisayegh (approved):
Nice PR 💯

packages/headless/src/features/commerce/context/cart/cart-state.ts
```diff
@@ -1,6 +1,10 @@
 import {CartItemParam} from '../../../../api/commerce/commerce-api-params';
 
 export interface CartItemWithMetadata extends CartItemParam {
+  /**
+   * The product unique identifier of the cart item.
+   */
```
Spuffynism, line 6: We use this in the `CartItem`. Both seem like good description, but we could be consistent.
```suggestion
  /**
   * The unique identifier of the product.
   */
```

Spuffynism (approved):
nice work!

PR #3812, opened 2024-04-18
feat(headless): adapting the analytics sent for the qna features to send the response id

mmitiche:


packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -261,11 +282,12 @@ export const logCollapseSmartSnippetSuggestion = (
       });
     },
     analyticsType: 'Qna.AnswerAction',
-    analyticsPayloadBuilder: (): Qna.AnswerAction => {
+    analyticsPayloadBuilder: (state): Qna.AnswerAction => {
       return {
         action: 'collapse',
         answer: {
-          id: payload.questionAnswerId,
+          // we can't really tell which of the suggestions was clicked.
```
mmitiche, line 289:
```suggestion
```

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -301,53 +322,76 @@ export const logOpenSmartSnippetSuggestionSource = (
...
         payload.questionAnswerId
       );
-      const result = answerSourceSelector(state, relatedQuestion?.documentId)!;
-      const information = partialDocumentInformation(result, state);
       return {
-        searchUid: state.search?.response.searchUid ?? '',
-        position: information.documentPosition,
-        itemMetadata: analyticsEventItemMetadata(result, state),
+        answer: {
+          responseId: state.search?.response.searchUid || '',
```
mmitiche, line 333: @jstpierref 

When a smart snippet suggestion source is clicked, you can see here that we are sending a `Qna.CitationClick` with the `searchUid` as the `responseId` field, so we have the identifier of the search query that was made.

However with one search query multiple smart snippet suggestions could have been returned, so we can't know which smart snippet suggestion  this event is related to. is that okay? what do you think?

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -301,53 +322,76 @@ export const logOpenSmartSnippetSuggestionSource = (
...
+          type: 'SmartSnippetSuggestion',
+        },
+        citation: {
+          // Question what should be the citation id for SS.
+          id: relatedQuestion?.documentId.contentIdValue || '',
```
mmitiche, line 338: @jstpierref, In the case of smart snippets, what's the value expected to be received as the key `id` in the `citation` object, 
is it the permanentId of the result from where the snippet was extracted?

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -301,53 +322,76 @@ export const logOpenSmartSnippetSuggestionSource = (
...
       return {
-        searchUid: state.search?.response.searchUid ?? '',
-        position: information.documentPosition,
-        itemMetadata: analyticsEventItemMetadata(result, state),
+        answer: {
+          responseId: state.search?.response.searchUid || '',
+          type: 'SmartSnippetSuggestion',
+        },
+        citation: {
+          // Question what should be the citation id for SS.
```
mmitiche, line 337:
```suggestion
```

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
jstpierref: Ideally, what uniquely identify the citation would be the `citation.id`. Now, for Smart Snippets specific `permanentId` per citation (which is not unique), but I think this is good enough given the reporting needs.

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
jstpierref: Yes, that would be the `permanentId`, even if it can be shared among many related questions

louis-bompart (approved):

jstpierref (approved):

olamothe (approved):

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
erocheleau: If both events become Qna.CitationClick what would be the way to differentiate the two events?
Differentiate a CitationClick from SmartSnippet from a CitationClick from RGA?

erocheleau (approved):
Small suggestion on how to handle null values, and a question, but other than this all the code looks good.

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -114,35 +113,57 @@ export const logOpenSmartSnippetSource = (): ClickAction =>
         documentIdentifier(result)
       );
     },
-    analyticsType: 'itemClick',
-    analyticsPayloadBuilder: (state): ItemClick => {
+    analyticsType: 'Qna.CitationClick',
```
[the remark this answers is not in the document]
mmitiche: The  difference here is that qna is a name space, so should  namespace always starts with a capital letter?
 I will try to clarify this, and once it's clear I will open a separate PR to adapt all the event types to one singular naming convention.

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
mmitiche: The `answer.type` sent in the payload of the event will do this differentiation, it can take the following values: `RGA`,  `SmartSnippetSuggestion` or `SmartSnippet`

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -114,35 +113,57 @@ export const logOpenSmartSnippetSource = (): ClickAction =>
...
+        }
+      );
+    },
+    analyticsType: 'Qna.CitationClick',
+    analyticsPayloadBuilder: (state): Qna.CitationClick => {
+      const result = answerSourceSelector(state)!;
+      const identifier = documentIdentifier(result);
+      return {
+        answer: {
+          responseId: state.search?.response.searchUid || '',
```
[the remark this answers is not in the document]
mmitiche: we can't use `?` and `!` in the same expression, cause it's contradictory.

PR #3814, opened 2024-04-18
feat(commerce): attempt using multiple reducers for commerce solution types

Spuffynism:
This is an attempt at loading sub-reducers for solution types dynamically. My feeling with this approach is that it adds complexity that's difficult to handle in headless, even though it keeps reducers nice.

As the commerce engine state will stay internal, we could switch to this approach at a later time.

See https://github.com/coveo/ui-kit/pull/3808#discussion_r1569263781
[CAPI-719]

[CAPI-719]: https://coveord.atlassian.net/browse/CAPI-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/app/commerce-engine/commerce-engine.ts
```diff
@@ -103,6 +106,39 @@ export function buildCommerceEngine(
...
+      const sliceReducer = combineReducers(reducers);
+
+      // TODO(nico): We're adding the reducers to the engine as top-level reducers because it's a bit difficult to get
+      //  headless to play nice with nested reducers. We should and CAN revisit this in the future since the top-level
+      //  engine state won't be exposed on the commerce engine.
+      engine.addReducers({
+        [slice]: namespaceReducer(sliceReducer, slice),
+      });
+    },
+
```
Spuffynism, line 141: This shows the trick used to make this approach work.

First, we create a `dispatch` specifically for namespaced actions (pagination actions, sort actions, facet actions, etc.).

Then, we expose `setSliceReducers` to allow creating the reducer for the specific instance of the solution type. This is needed because there is no way to "hook" into the reducer manager and add/remove sub-reducers. This means we must then keep the list of reducers for the solution type in the sub-controller: https://github.com/coveo/ui-kit/pull/3814/files#diff-70fb810df3b4d7646a54a4eb9c51a65a82fe48917db532f2018810b71ea340ec

The `namespaceReducer` function is adapted from https://redux.js.org/usage/structuring-reducers/reusing-reducer-logic#customizing-behavior-with-higher-order-reducers

Spuffynism:
Superseded by https://github.com/coveo/ui-kit/pull/3842

PR #3818, opened 2024-04-18
fix: `enableQuerySyntax` is not passed to recent-query-list controller

y-lakhdar:
https://coveord.atlassian.net/browse/KIT-3096

Whenever a submit action (Enter key pressed) is being recorded on a recent-query suggestion, the `prepareForSearchWithQuery` search action is triggered with the appropriate `enableQuerySyntax` value.

However, when a mouse click action is recorded on a recent-query suggestion, the `prepareForSearchWithQuery` search action called by the headless-recent-queries-list controller does not specify the `enableQuerySyntax` value.

### DIFF
--- a/packages/headless/src/controllers/recent-queries-list/headless-recent-queries-list.test.ts
+++ b/packages/headless/src/controllers/recent-queries-list/headless-recent-queries-list.test.ts
@@ -80,7 +80,7 @@ describe('recent queries list', () => {
     });
 
     afterEach(() => {
-      mockedprepareForSearchWithQuery.mockClear();
+      mockedPrepareForSearchWithQuery.mockClear();
     });
 
     it('should register with props on init', () => {
@@ -110,8 +110,8 @@ describe('recent queries list', () => {
       engine.state.query = buildMockQueryState();
       engine.state.recentQueries = {...testInitialState, ...testOptions};
       recentQueriesList.executeRecentQuery(0);
-      expect(mockedprepareForSearchWithQuery).toHaveBeenCalledTimes(1);
-      expect(mockedprepareForSearchWithQuery).toHaveBeenCalledWith({
+      expect(mockedPrepareForSearchWithQuery).toHaveBeenCalledTimes(1);
+      expect(mockedPrepareForSearchWithQuery).toHaveBeenCalledWith({
         q: testInitialState.queries[0],
         clearFilters: testOptions.clearFilters,
         enableQuerySyntax: false,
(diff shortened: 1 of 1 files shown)


packages/headless/src/controllers/recent-queries-list/headless-recent-queries-list.ts
```diff
@@ -199,6 +205,9 @@ export function buildRecentQueriesList(
         prepareForSearchWithQuery({
           q: this.state.queries[index],
           clearFilters: registrationOptions.clearFilters,
+          ...(isBoolean(engine.state.query?.enableQuerySyntax) && {
```
fbeaudoincoveo, line 208: This prompted me to make a comment in an old TypeScript that unfortunately got rejected a while back (https://github.com/microsoft/TypeScript/issues/32495#issuecomment-2065249303) regarding the fact that TypeScript doesn't check for excess properties in spread object literals, which I believe is problematic.

I don't have an efficient workaround to suggest, though.

louis-bompart (approved):

fbeaudoincoveo (approved):

olamothe (approved):

packages/headless/src/controllers/recent-queries-list/headless-recent-queries-list.ts
fpbrault: I've been reading on this and I think this issue has most of the details:
https://github.com/microsoft/TypeScript/issues/13444
There intended solution to this is to implement Exact types:
https://github.com/microsoft/TypeScript/issues/12936

I know it does not prevent the underlying issue from happening, but what about doing something like this instead? 

```typescript

      const queryOptions: Parameters<typeof prepareForSearchWithQuery>[0] = {
        q: this.state.queries[index],
        clearFilters: registrationOptions.clearFilters,
      };

      if (isBoolean(engine.state.query?.enableQuerySyntax)) {
        queryOptions.enableQuerySyntax = engine.state.query.enableQuerySyntax;
      }

      dispatch(prepareForSearchWithQuery(queryOptions));
```

fpbrault: Forgot to mention that there's also discussions about changing the error message to make it more clear that excess property errors are more of a warning than an actual error 
https://github.com/microsoft/TypeScript/pull/55152#issuecomment-1658831938

fpbrault (approved):

PR #3820, opened 2024-04-18
feat: use navigator relay & context on search-actions

louis-bompart:
This PR leverages #3819 in the various search actions.
Insight Engine is purposefully excluded to keep the scope tight

The SSR engine needs to be 'unbodged' with a more reliable solution, which will be done in a subsequent PR.

https://coveord.atlassian.net/browse/KIT-3024


packages/samples/headless-ssr/common/components/common/search-engine-config.ts
```diff
@@ -23,6 +23,11 @@ export const config = {
     authorFacet: defineFacet({options: {facetId: 'author-1', field: 'author'}}),
     searchParameterManager: defineSearchParameterManager(),
   },
+  navigatorContextProvider: () => ({
+    location: 'todo',
+    referrer: 'todo',
+    userAgent: 'todo',
+  }),
```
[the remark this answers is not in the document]
fpbrault: @y-lakhdar what do you mean by "custom object to store middleware values in context" here?

Something like this? https://clerk.com/blog/nextjs-pass-value-from-middleware-to-api-routes-and-getserversideprops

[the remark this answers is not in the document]
louis-bompart: For now handled in PR upstream, but still a problem for future-me.

Spuffynism (approved):

packages/headless/src/features/query-suggest/query-suggest-actions.ts
```diff
@@ -166,7 +180,15 @@ export const buildQuerySuggestRequest = async (
     ...(s.configuration.analytics.enabled && {
       visitorId: await getVisitorID(s.configuration.analytics),
       ...(s.configuration.analytics.enabled &&
-        (await fromAnalyticsStateToAnalyticsParams(s.configuration.analytics))),
+      s.configuration.analytics.analyticsMode === 'legacy'
+        ? await legacyFromAnalyticsStateToAnalyticsParams(
+            s.configuration.analytics
+          )
+        : await fromAnalyticsStateToAnalyticsParams(
```
fbeaudoincoveo, line 187: Probably doesn't need to be awaited, see my comment here: https://github.com/coveo/ui-kit/pull/3820/files#diff-94c94221a30278160bf3d13828c55f9c85ab3bedb259dbd080231da783a13284L8

packages/headless/src/features/standalone-search-box-set/standalone-search-box-set-actions.ts
```diff
@@ -148,9 +158,15 @@ export const buildPlanRequest = async (
...
     }),
     ...(state.configuration.analytics.enabled &&
-      (await fromAnalyticsStateToAnalyticsParams(
-        state.configuration.analytics
-      ))),
+    state.configuration.analytics.analyticsMode === 'legacy'
+      ? await legacyFromAnalyticsStateToAnalyticsParams(
+          state.configuration.analytics
+        )
+      : await fromAnalyticsStateToAnalyticsParams(
```
fbeaudoincoveo, line 165: remove await

packages/headless/src/features/search-and-folding/search-and-folding-request.ts
```diff
@@ -14,6 +12,8 @@ type StateNeededByExecuteSearchAndFolding = ConfigurationSection &
 
 export const buildSearchAndFoldingLoadCollectionRequest = async (
```
fbeaudoincoveo, line 13: Here also I don't think it needs to be async anymore (which implies you can remove awaits where this function gets called).

packages/headless/src/features/search-and-folding/search-and-folding-request.ts
```diff
@@ -23,11 +23,10 @@ export const buildSearchAndFoldingLoadCollectionRequest = async (
     locale: state.configuration.search.locale,
     debug: state.debug,
     tab: state.configuration.analytics.originLevel2,
-    referrer: state.configuration.analytics.originLevel3,
+    referrer: navigatorContext.referrer,
```
fbeaudoincoveo, line 26: Am I correct to assume that the originLevel3 property in the engine's analytics configuration will become completely unused?

packages/headless/src/features/search/search-request.test.ts
```diff
@@ -397,81 +554,167 @@ describe('search request', () => {
...
+        await buildSearchRequest(
+          state,
+          {location: '', referrer: '', userAgent: ''},
+          mockRelay()
+        )
+      ).request.tab
+    ).toBe(originLevel2);
   });
 
   it('#searchRequest.referrer holds the #originLevel3', async () => {
```
fbeaudoincoveo, line 609: I'm surprised that this test is passing!

Shouldn't it take the referrer from the NavigatorContext passed as a second argument (i.e., '' in this case)?

louis-bompart: Not with analyticsMode: legacy ;)

packages/headless/src/api/search/search-api-client.test.ts
```diff
@@ -105,7 +106,13 @@ describe('search api client', () => {
...
-      const req = (await buildSearchRequest(state)).request;
+      const req = (
+        await buildSearchRequest(
+          state,
+          {location: '', referrer: '', userAgent: ''},
+          mockRelay()
+        )
+      ).request;
       const res = await searchAPIClient.search(req);
       if (!isErrorResponse(res)) {
```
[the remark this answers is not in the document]
louis-bompart: I tried to remove them: They are type-guards, so it makes sense in the test

packages/headless/src/features/search-and-folding/search-and-folding-request.ts
louis-bompart: Yes, with the next analytics

packages/headless/src/features/configuration/analytics-params.ts
louis-bompart, line 1: In the end, this one being used solely iif analyticsMode=next, removed all branching regarding analyticsMode

packages/headless/src/features/configuration/analytics-params.ts
```diff
@@ -1,38 +1,36 @@
...
-  const isDefaultReferrer = s.originLevel3 === 'default';
+): AnalyticsParam => {
   return {
     analytics: {
-      clientId: await getVisitorID(s),
+      clientId: relay.getMeta('').clientId,
       clientTimestamp: new Date().toISOString(),
-      documentReferrer:
-        isNextAnalytics && isDefaultReferrer ? null : s.originLevel3,
+      documentReferrer: navigatorContext.referrer,
```
louis-bompart, line 19: I ensured that we always send null where we sent default, and that's all.

Further improvement could probably be made, but I do prefer to close this PR, it has lived long enough.

FYI @samisayegh probably some potential hiccup with documentLocation similar to #3974
No regression tho, so I will not hold this PR because of that.

samisayegh: Since the initial PR, we've made referrer [less strict](https://github.com/coveo-platform/analytics_schema/pull/435) (i.e., `urlEmpty`), so the allowed types now are:

```
location: uri | null
referrer: uri | "" | null
```
We didn't support an empty string for `location` because unlike `document.referrer`, we didn't see `window.location.href` return empty string, even on a new tab.

When navigatorContext is in nodejs mode, would it be possible to send `null` for both keys?
Also if it's helpful, `relay.getMeta()` has this behavior encoded into it. That said, getting access to the relay instance might not be so easy,

packages/headless/src/app/navigatorContextProvider.ts
```diff
@@ -15,7 +15,7 @@ export const defaultBrowserNavigatorContextProvider: NavigatorContextProvider =
 
 export const defaultNodeJSNavigatorContextProvider: NavigatorContextProvider =
   () => ({
-    referrer: 'default',
+    referrer: null,
     userAgent: 'Node.js',
     location: 'default',
```
samisayegh, line 20: Related to the other comment, could we set `null` here and for `userAgent` too?

packages/headless/src/features/search-and-folding/search-and-folding-request.test.ts
```diff
@@ -128,12 +168,20 @@ describe('buildSearchAndFoldingLoadCollectionRequest', () => {
     });
 
     it('#visitorId is not included in the request', async () => {
-      const request = await buildSearchAndFoldingLoadCollectionRequest(state);
+      const request = buildSearchAndFoldingLoadCollectionRequest(
+        state,
+        {location: '', referrer: '', userAgent: ''},
```
samisayegh, line 173: It could be more flexible to convert this object into a mock (e.g., `buildMockNavigatorContext`) so that if it changes, one only needs to update a single place

olamothe (approved):

PR #3823, opened 2024-04-18
feat(atomic): add atomic tabs with facets support

fpbrault:
This PR adds the atomic-tab and atomic-tab-section components, and support for filtering facets based on the active tab.

### DIFF
--- a/packages/atomic-angular/projects/atomic-angular/src/lib/stencil-generated/components.ts
+++ b/packages/atomic-angular/projects/atomic-angular/src/lib/stencil-generated/components.ts
@@ -95,14 +95,14 @@ export declare interface AtomicBreadbox extends Components.AtomicBreadbox {}
 
 
 @ProxyCmp({
-  inputs: ['basePath', 'delimitingCharacter', 'dependsOn', 'facetId', 'field', 'filterByBasePath', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'sortCriteria', 'tabs', 'withSearch']
+  inputs: ['basePath', 'delimitingCharacter', 'dependsOn', 'facetId', 'field', 'filterByBasePath', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withSearch']
 })
 @Component({
   selector: 'atomic-category-facet',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['basePath', 'delimitingCharacter', 'dependsOn', 'facetId', 'field', 'filterByBasePath', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'sortCriteria', 'tabs', 'withSearch'],
+  inputs: ['basePath', 'delimitingCharacter', 'dependsOn', 'facetId', 'field', 'filterByBasePath', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withSearch'],
 })
 export class AtomicCategoryFacet {
   protected el: HTMLElement;
@@ -117,14 +117,14 @@ export declare interface AtomicCategoryFacet extends Components.AtomicCategoryFa
 
 
 @ProxyCmp({
-  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabs', 'withSearch']
+  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withSearch']
 })
 @Component({
   selector: 'atomic-color-facet',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabs', 'withSearch'],
+  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withSearch'],
 })
 export class AtomicColorFacet {
   protected el: HTMLElement;
@@ -205,14 +205,14 @@ export declare interface AtomicExternal extends Components.AtomicExternal {}
 
 
 @ProxyCmp({
-  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'enableExclusion', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabs', 'withSearch']
+  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'enableExclusion', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withSearch']
 })
 @Component({
   selector: 'atomic-facet',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'enableExclusion', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabs', 'withSearch'],
+  inputs: ['allowedValues', 'customSort', 'dependsOn', 'displayValuesAs', 'enableExclusion', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'resultsMustMatch', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withSearch'],
 })
 export class AtomicFacet {
   protected el: HTMLElement;
@@ -556,14 +556,14 @@ export declare interface AtomicNotifications extends Components.AtomicNotificati
 
 
 @ProxyCmp({
-  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'rangeAlgorithm', 'sortCriteria', 'tabs', 'withInput']
+  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'rangeAlgorithm', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withInput']
 })
 @Component({
   selector: 'atomic-numeric-facet',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'rangeAlgorithm', 'sortCriteria', 'tabs', 'withInput'],
+  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'injectionDepth', 'isCollapsed', 'label', 'numberOfValues', 'rangeAlgorithm', 'sortCriteria', 'tabsExcluded', 'tabsIncluded', 'withInput'],
 })
 export class AtomicNumericFacet {
   protected el: HTMLElement;
@@ -711,14 +711,14 @@ export declare interface AtomicQuickview extends Components.AtomicQuickview {}
 
 
 @ProxyCmp({
-  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabs']
+  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabsExcluded', 'tabsIncluded']
 })
 @Component({
   selector: 'atomic-rating-facet',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabs'],
+  inputs: ['dependsOn', 'displayValuesAs', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabsExcluded', 'tabsIncluded'],
 })
 export class AtomicRatingFacet {
   protected el: HTMLElement;
@@ -733,14 +733,14 @@ export declare interface AtomicRatingFacet extends Components.AtomicRatingFacet 
 
 
 @ProxyCmp({
-  inputs: ['dependsOn', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabs']
+  inputs: ['dependsOn', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabsExcluded', 'tabsIncluded']
 })
 @Component({
   selector: 'atomic-rating-range-facet',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['dependsOn', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabs'],
+  inputs: ['dependsOn', 'facetId', 'field', 'filterFacetCount', 'headingLevel', 'icon', 'injectionDepth', 'isCollapsed', 'label', 'maxValueInIndex', 'minValueInIndex', 'numberOfIntervals', 'tabsExcluded', 'tabsIncluded'],
 })
 export class AtomicRatingRangeFacet {
   protected el: HTMLElement;
@@ -1726,14 +1726,14 @@ export declare interface AtomicSearchLayout extends Components.AtomicSearchLayou
 
 
 @ProxyCmp({
-  inputs: ['allowedValues', 'customSort', 'dependsOn', 'facetId', 'field', 'filterFacetCount', 'injectionDepth', 'label', 'numberOfValues', 'sortCriteria', 'tabs']
+  inputs: ['allowedValues', 'customSort', 'dependsOn', 'facetId', 'field', 'filterFacetCount', 'injectionDepth', 'label', 'numberOfValues', 'sortCriteria', 'tabsExcluded', 'tabsIncluded']
 })
 @Component({
   selector: 'atomic-segmented-facet',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['allowedValues', 'customSort', 'dependsOn', 'facetId', 'field', 'filterFacetCount', 'injectionDepth', 'label', 'numberOfValues', 'sortCriteria', 'tabs'],
+  inputs: ['allowedValues', 'customSort', 'dependsOn', 'facetId', 'field', 'filterFacetCount', 'injectionDepth', 'label', 'numberOfValues', 'sortCriteria', 'tabsExcluded', 'tabsIncluded'],
 })
 export class AtomicSegmentedFacet {
   protected el: HTMLElement;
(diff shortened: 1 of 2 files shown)


packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -133,6 +134,10 @@ export class AtomicCategoryFacet
    * Also determines the number of additional values to request each time more values are shown.
    */
   @Prop({reflect: true}) public numberOfValues = 8;
+  /**
+   * The tabs on which to display the facet.
+   */
+  @Prop({reflect: true}) public tabs: string = '';
```
olamothe, line 140: This should be an `@ArrayProp`, like we do for `atomic-search-interface` -> `fieldsToInclude` for example.

We should not rely on character delimiter for such type of properties, even though it might seem more elegant: They tend to cause other type of issues. 

It's uglier, but "safer" to instead expect implementer to pass in a JSON array... And also, more uniform with the rest.

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+})
+export class AtomicTabSection {
+  @InitializeBindings() public bindings!: Bindings;
+  @Element()
+  private host!: HTMLElement;
+
+  /**
+   * When provided, this tab will be selected by default when the component loads. Otherwise, the first tab is selected automatically.
+   */
+  @Prop() defaultActiveTab?: string;
```
olamothe, line 40: Could we do without this ? Just always have the first tab be selected by default ? It is really necessary to have multiple mode to handle which one is the default tab ?

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+          text={tab.label}
+          onClick={() => {
+            tab.select();
+          }}
+        />
+      </li>
+    ));
+  }
+
+  updateActiveTab() {
```
olamothe, line 109: All of this function should be removed, and we can simply know "what's the label of the current tab" by reading from Headless state.

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+      if (tab.isActive) {
+        const span = dropdownButton?.getElementsByTagName('span')[0];
+        if (span) {
+          span.textContent = tab.label;
+        }
+      }
+    });
+
+    dropdownOptions.forEach((option) => {
```
olamothe, line 128: As explained in my earlier comment, all of this should be done in the render function `buildDropdown` to check if the current button being rendered is the active one in Headless state

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+    if (initialTab && !activeTab) {
+      this.bindings.engine.dispatch(
+        loadTabSetActions(this.bindings.engine).updateActiveTab(initialTab.name)
+      );
+      this.tabInit.emit();
+    }
+    this.updateActiveTab();
+  }
+
+  buildDropdown() {
```
olamothe, line 93: Nit: ```renderDropdown()```

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+    dropdownOptions.forEach((option) => {
+      if ((option as HTMLButtonElement).title === activeTabName) {
+        option.classList.add('font-bold');
+      } else {
+        option.classList.remove('font-bold');
+      }
+    });
+  }
+
+  clearState() {
```
olamothe, line 137: All of this logic should be added to the Headless tab controller itself, as an option when it gets built, which handle `clearing breadcrumbs/query` on `select()`

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  public render() {
+    return (
+      <div class="mb-4 overflow-visible ">
+        <div class="flex flex-row tabs-container">
+          <div class="flex flex-row w-full mb-2 border-b tabs-area ">
+            <slot></slot>
+          </div>
+          <div class="relative flex-row hidden mb-2 border-b cursor-pointer group dropdown-area ">
+            <Button
```
olamothe, line 163: There's quite a few accessibility problem with trying to simulate what is essentially a `<select>` with buttons/dropdown and the likes with JS code.

Then you can add in mobile support, which is much better when you use the correct semantic element, versus a cramped up dropdown.

packages/atomic/src/utils/tab-utils.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      tabSet: {
+        tab1: {id: 'tab1', isActive: false, expression: ''},
+        tab2: {id: 'tab2', isActive: true, expression: ''},
+        tab3: {id: 'tab3', isActive: false, expression: ''},
+      },
+    };
+    expect(getActiveTab(state)).toEqual({tab: 'tab2'});
+  });
+
+  it('returns the first active tab when multiple tabs are active', () => {
```
fbeaudoincoveo, line 31: Mostly musing here but...

If knowing what tab is currently active is important for facets (and other controllers/components), shouldn't we expose the id of the active tab on the state of those controllers? It would either be undefined (if the tab section isn't in the state at all), null (if the tab section is there but no tab is selected) or a string (the id of the currently selected tab).

Maybe the shape of the tab section could also change in the engine state so that instead of being a `Record<string, TabSlice>` it would be something like

```typescript
export type TabSetState = { tabs: Record<string, TabSlice>; activeTab?: string }
```

and then when we dispatch `updateActiveTab(id)`, in the tabSetReducer we'd also set the active tab there... which means we wouldn't have to iterate through all tabs all the time to find the active one.

Now the problem with what I'm suggesting above is that modifying the definition of the `TabSetState` type would definitely be a breaking change.

Another possibility would be to add a new section in the state just to represent the active tab, and perhaps modify the TabSetState type in v3 or v4... but maybe in v4 the engine's state won't be public anymore and then we can do whatever we want ;)

packages/atomic/src/utils/tab-utils.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      tabSet: {
+        tab1: {id: 'tab1', isActive: true, expression: ''},
+        tab2: {id: 'tab2', isActive: false, expression: ''},
+        tab3: {id: 'tab3', isActive: false, expression: ''},
+      },
+    };
+    expect(shouldDisplayOnCurrentTab(tabs, state)).toBe(false);
+  });
+
+  it('returns false when active tab is excluded and included in the excluded tabs', () => {
```
fbeaudoincoveo, line 68: This is not what you're testing, though; this test is identical to the one above

I think you should have

```typescript
const tabs = '!tab1;tab2;!tab2;!tab3';
```

and I'd simplify the description of the test:

`returns false when active tab is both included and excluded`

packages/atomic/src/utils/tab-utils.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      tabSet: {
+        tab1: {id: 'tab1', isActive: false, expression: ''},
+        tab2: {id: 'tab2', isActive: true, expression: ''},
+        tab3: {id: 'tab3', isActive: false, expression: ''},
+      },
+    };
+    expect(shouldDisplayOnCurrentTab(tabs, state)).toBe(false);
+  });
+
+  it('returns true when active tab is included and not excluded', () => {
```
fbeaudoincoveo, line 80: this test becomes redundant if you modify the one above as I suggested

packages/atomic/src/utils/tab-utils.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        tab2: {id: 'tab2', isActive: true, expression: ''},
+        tab3: {id: 'tab3', isActive: false, expression: ''},
+      },
+    };
+    expect(getActiveTab(state)).toEqual({tab: 'tab1'});
+  });
+});
+
+describe('shouldDisplayOnCurrentTab', () => {
+  it('returns true when active tab is included and no tabs are excluded', () => {
```
fbeaudoincoveo, line 44:
```suggestion
  it('returns true when active tab is included and not excluded', () => {
```

packages/atomic/src/utils/tab-utils.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      tabSet: {
+        tab1: {id: 'tab1', isActive: false, expression: ''},
+        tab2: {id: 'tab2', isActive: true, expression: ''},
+        tab3: {id: 'tab3', isActive: false, expression: ''},
+      },
+    };
+    expect(shouldDisplayOnCurrentTab(tabs, state)).toBe(true);
+  });
+
+  it('returns true when a single tab is excluded and the active tab is different', () => {
```
fbeaudoincoveo, line 92: It doesn't matter if it's one excluded tab or many, as long as the active tab is different, so I'd rephrase:

`returns true when tabs are excluded, no tabs are included, and the active tab is different`

with `const tabs = '!tab1;!tab3'`

packages/atomic/src/utils/tab-utils.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      tabSet: {
+        tab1: {id: 'tab1', isActive: true, expression: ''},
+        tab2: {id: 'tab2', isActive: false, expression: ''},
+        tab3: {id: 'tab3', isActive: false, expression: ''},
+      },
+    };
+    expect(shouldDisplayOnCurrentTab(tabs, state)).toBe(true);
+  });
+
+  it('returns false when active tab is excluded and no tabs are included', () => {
```
fbeaudoincoveo, line 56:
```suggestion
  it('returns false when active tab is excluded', () => {
```

packages/atomic/src/utils/tab-utils.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    const state = {
+      tabSet: {
+        tab1: {id: 'tab1', isActive: false, expression: ''},
+        tab2: {id: 'tab2', isActive: true, expression: ''},
+        tab3: {id: 'tab3', isActive: false, expression: ''},
+      },
+    };
+    expect(shouldDisplayOnCurrentTab(tabs, state)).toBe(true);
+  });
+
```
fbeaudoincoveo, line 91: we should also add:

```typescript
it('returns false when tabs are included, no tabs are excluded, and the active tab is different', () => {
    const tabs = 'tab1,tab3';
    const state = {
      tabSet: {
        tab1: {id: 'tab1', isActive: false, expression: ''},
        tab2: {id: 'tab2', isActive: true, expression: ''},
        tab3: {id: 'tab3', isActive: false, expression: ''},
      },
    };
    expect(shouldDisplayOnCurrentTab(tabs, state)).toBe(false);
});
```

`returns false when tabs are included and the active tab is different`

with `const tabs = 'tab1`;

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * The internal name of the atomic tab.
+   */
+  @Prop() name!: string;
+  /**
+   * Indicates whether the tab is currently active.
+   */
+  @Prop({reflect: true}) isActive: boolean = false;
+  /**
+   * The expression that will be passed to the search as a `cq` paramenter upon being selected.
```
fbeaudoincoveo, line 85:
```suggestion
   * The [constant query expression (`cq`)](https://docs.coveo.com/en/2830/searching-with-coveo/about-the-query-expression#constant-query-expression-cq) to apply when the tab is the active one.
```

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * The label displayed on the tab.
+   */
+  @Prop() label!: string;
+  /**
+   * The internal name of the atomic tab.
+   */
+  @Prop() name!: string;
+  /**
+   * Indicates whether the tab is currently active.
```
fbeaudoincoveo, line 81:
```suggestion
   * Whether the tab is the active one.
```

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+  @State()
+  private tabState!: TabState;
+
+  @Event({
+    eventName: 'atomic/tabClick',
+  })
+  tabClick!: EventEmitter;
+
+  /**
+   * Whether this tab is active upon rendering.
```
fbeaudoincoveo, line 68:
```suggestion
   * Whether to make this tab the active one upon rendering.
```

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+   * If multiple tabs are set to active on render, the last one to be rendered will override the others.
```
fbeaudoincoveo, line 69:
```suggestion
   * If this prop is set to `true` on multiple tabs, the last tab to render will be the active one.
```

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+   */
+  @Prop({reflect: true, mutable: true}) public active = false;
+  /**
+   * The label displayed on the tab.
```
fbeaudoincoveo, line 73:
```suggestion
   * The label to display on the tab.
```

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * The expression that will be passed to the search as a `cq` paramenter upon being selected.
+   */
+  @Prop() public expression: string = '';
+
+  private tab!: Tab;
+  private unsubscribe: Unsubscribe = () => {};
+
+  /**
+   * Activates the tab.
```
fbeaudoincoveo, line 93:
```suggestion
   * Makes the tab the active one.
```

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+  select(): void;
+  /**
+   * The state of the `Tab` controller.
+   */
+  state: TabState;
+}
+
+interface TabState {
+  /**
+   * Indicates whether the current tab is selected.
```
fbeaudoincoveo, line 34:
```suggestion
   * Whether the current tab is the active one.
```

packages/atomic/src/components/search/tabs/atomic-tab/atomic-tab.tsx
```diff
@@ -0,0 +1,142 @@
...
+  InitializeBindings,
+} from '../../../../utils/initialization-utils';
+import {Button} from '../../../common/button';
+import {CommonBindings} from '../../../common/interface/bindings';
+import {dispatchTabLoaded} from '../../../common/tabs/tab-common';
+import {AtomicStore} from '../../atomic-search-interface/store';
+
+interface Tab extends Controller {
+  /**
+   * Activates the tab.
```
fbeaudoincoveo, line 23:
```suggestion
   * Makes the tab the active one.
```

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  setInitialTab() {
+    const initialTab =
+      this.tabs.find((tab) => tab.name === this.defaultActiveTab) ||
+      this.tabs[0];
+    const activeTab = getActiveTab(this.bindings.engine.state)?.tab;
+
+    if (initialTab && !activeTab) {
+      this.bindings.engine.dispatch(
```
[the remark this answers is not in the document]
fpbrault: Yep, I did this initially to prevent a double query. The tab controller can now change the tab without triggering a search.

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
fpbrault: Done!

packages/atomic/src/components/common/facets/facet-tabs/facet-tabs-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Updates the visibility of a facet based on the active tab.
+ * @param tabs - The active tab.
+ * @param facet - The facet to update.
+ * @param bindings - The bindings object.
+ * @returns A boolean indicating whether the facet was enabled or disabled.
+ */
+export function updateFacetVisibilityForActiveTab(
+  tabs: string,
```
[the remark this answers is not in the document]
fpbrault: The doc is indeed wrong! 

I can do that, but my thinking was that since the logic is the same on all the facets, it would be simpler (and make it easier to modify with the upcoming facet refactoring) to use a shared function.

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+import {InitializeBindings} from '../../../../utils/initialization-utils';
+import {getActiveTab} from '../../../../utils/tab-utils';
+import {Button} from '../../../common/button';
+import {Bindings} from '../../atomic-search-interface/atomic-search-interface';
+
+/**
+ * @internal
+ */
+@Component({
+  tag: 'atomic-tab-section',
```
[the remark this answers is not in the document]
fpbrault: I think `atomic-tab-manager` and `atomic-tab` are better than `atomic-tab-dropdown`and `atomic-tab-expression` as if fit their roles/appearance better

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
fpbrault: I think there are some scenarios where being able to set a different initial tab, but I'm fine removing this attribute/logic since it is possible to replicate the same behaviour using the tab param in the url... It is just less obvious

packages/atomic/src/components/common/facets/facet-tabs/facet-tabs-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+      disableFacet(facet);
+      return false;
+    }
+  }
+  return true;
+}
+
+function disableFacet(facet: AnyFacetType) {
+  facet.state.enabled = false;
+  facet.disable();
```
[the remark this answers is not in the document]
fpbrault: Indeed! Ty!

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
fpbrault: A different tab controller instance is created for each tab, so I would need to create a new tab manager controller and use that to get the active tab in the atomic-tab-manager component, right?

fpbrault: I added a (very basic) tab-manager controller and refactored atomic-tab-section to use it

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
```diff
@@ -0,0 +1,185 @@
...
+  styleUrl: 'atomic-tab-section.pcss',
+  shadow: true,
+})
+export class AtomicTabSection {
+  @InitializeBindings() public bindings!: Bindings;
+  @Element()
+  private host!: HTMLElement;
+
+  /**
+   * When provided, this tab will be selected by default when the component loads. Otherwise, the first tab is selected automatically.
```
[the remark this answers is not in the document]
fpbrault: @fbeaudoincoveo This had me realize that instead of setting a default tab in atomic-tab-section (and having to trigger a hash update in atomic-search-interface so that the url reflects the current state), we can set a default tab by setting the active prop on the desired tab. 

This seems a lot simpler to me, but what do you think?

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
fpbrault: I checked what would need to be done to make it more accessible. It could be done relatively easily, but it would require adding the aria-haspopup attribute to our button component. However, I decided to just go with a select+option as you suggested :)

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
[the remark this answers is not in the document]
fbeaudoincoveo: I think the simplest and best approach would be to try to get the active tab from the URL state and fallback to the first tab found under the (soon to be renamed?) `atomic-tab-section` component. This should cover most use cases. After all, why would you want to select anything but the first tab by default?

One problem I see with having an `active` prop on each atomic-tab component is that you have to deal with possible misconfigurations. E.g., what happens when someone sets `active` to `true` on multiple tabs? You'd probably want to throw a warning or something that explains what tab was selected due to the misconfiguration... It feels a bit awkward.

Arguably there's a similar concern with having a prop on the `atomic-tab-section`: what if the specified tab ID doesn't exist? Do you throw a runtime error or gracefully select the first tab as a fallback and show a console warning?...

I'd get rid of both props 😁

packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -133,6 +134,10 @@ export class AtomicCategoryFacet
    * Also determines the number of additional values to request each time more values are shown.
    */
   @Prop({reflect: true}) public numberOfValues = 8;
+  /**
+   * The tabs on which to display the facet.
```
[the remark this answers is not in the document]
fpbrault: @louis-bompart since you were the one that suggested this syntax, what are your thoughts on this? 

Also, since it is related what do you think about Oli's suggestion to use `@ArrayProp`?

packages/atomic/src/components/search/tabs/atomic-tab-section/atomic-tab-section.tsx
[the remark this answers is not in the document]
fpbrault: I guess we can do that, if someones wants to set another tab by default, they can just change the link to the search page..!

The only downside I see is that if we want the url to have the correct tab name on initial load without triggering another search we need to update it.

I had added an event listener in atomic-search-interface for an event triggered by atomic-search-manager that would call updateHash(), but do you know a better way I could do this?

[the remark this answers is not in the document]
fpbrault: Actually, it does not matter if we don't show the tab name in the url on first load, it will always be the first one anyway.

packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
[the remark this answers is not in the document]
louis-bompart: I'm not that attached to the syntax, I'd trust the DXUI hivemind opinion :)
ArrayProp in any case make sense

packages/atomic/src/utils/tab-utils.spec.ts
louis-bompart: > Now the problem with what I'm suggesting above is that modifying the definition of the TabSetState type would definitely be a breaking change.

That's the problem ;)

louis-bompart (commented):
answering comment, other did a better review that I did :D

packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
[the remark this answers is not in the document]
fpbrault: What do you think of something like this then? More concise than what you suggested, but seems clear enough to me.

```typescript
  /**
   * The tabs on which to display the facet.
   *
   * Specify the property as an array using a JSON string representation:
   * ```html
   *  <atomic-facet tabs-included='["tabA", "tabB"]'></atomic-facet>
   * ```
   * If a tab is specified in both `tabs-included` and `tabs-excluded`, the exclusion takes precedence and the component will not be displayed on that tab.
   */
  @ArrayProp()
  @Prop({reflect: true, mutable: true})
  public tabsIncluded: string[] | string = '[]';

  /**
   * The tabs on which to exclude the facet.
   *
   * Specify the property as an array using a JSON string representation:
   * ```html
   *  <atomic-facet tabs-excluded='["tabA", "tabB"]'></atomic-facet>
   * ```
   * If a tab is specified in both `tabs-included` and `tabs-excluded`, the exclusion takes precedence and the component will not be displayed on that tab.
   */
  @ArrayProp()
  @Prop({reflect: true, mutable: true})
  public tabsExcluded: string[] | string = '[]';
  ```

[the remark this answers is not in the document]
Comment (fbeaudoincoveo) [reply packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx:138]: I like it! Here's a counter proposal with some nitpicks...

```typescript
  /**
   * The tabs on which the facet can be displayed. This property complements `tabs-excluded`.
   *
   * Set this property as a stringified JSON array, e.g.,
   * ```html
   *  <atomic-facet tabs-included='["tabIDA", "tabIDB"]'></atomic-facet>
   * ```
   * If you don't set this property, or set it to `'[]'`, the facet can be displayed on any tab. Otherwise, the facet can only be displayed on the specified tabs. In either case, the facet won't be displayed on any of the tabs specified in the `tabs-excluded` property (exclusion takes precedence).
   */
  @ArrayProp()
  @Prop({reflect: true, mutable: true})
  public tabsIncluded: string[] | string = '[]';

  /**
   * The tabs on which this facet must not be displayed. This property complements `tabs-included`.
   *
   * Set this property as a stringified JSON array, e.g.,
   * ```html
   *  <atomic-facet tabs-excluded='["tabIDA", "tabIDB"]'></atomic-facet>
   * ```
   * If you don't set this property, or set it to `'[]'`, the facet can be displayed on any tab. Otherwise, the facet won't be displayed on any of the specified tabs. In either case, the `tabs-included` property can further restrict the tabs on which the facet can be displayed.
   */
  @ArrayProp()
  @Prop({reflect: true, mutable: true})
  public tabsExcluded: string[] | string = '[]';
```

packages/atomic/src/components/search/tabs/atomic-tab-manager/atomic-tab-manager.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  async setInitialTab() {
+    const initialTab = this.tabs[0];
+    const activeTab = getActiveTab(
+      this.bindings.engine.state.tabSet || {}
+    )?.tab;
+
+    if (initialTab && !activeTab) {
+      await initialTab.select(false);
```
olamothe, line 56: I think this pattern of having a boolean to control if it's "the first selection do not trigger a query" is because of a code smell.

Instead of having this pattern:

```html
atomic-tab-manager
   atomic-tab expression="foo"
   atomic-tab expression="bar"
atomic-tab-manager
```

Where: Each `atomic-tab` element creates its own tab controller, then "communicate" back to the manager, where the manager then has to copy the list and call a public method on the child `atomic-tab`, here what I would propose instead:

* atomic-tab-manager parse it's list of child element to read the list of expression/list of headless tab controller that should be created.

* atomic-tab-manager is in charge of creating N amount of headless tab controller depending on the above.

* Since it knows the full list of child element, the tab controller that should be active first (the first one in the DOM) gets a special initialization: `buildTabController(engine, {initialState: {isActive: true}`

* The render function of the manager become (pseudo code ):

```typescript
render() {
  <select>
     {this.allTabsController.map(tabController=> {
           <option value={tabController.id} onChange={()=> tabController.select()}>   
   })
   </select>
}
```

louis-bompart:
CLosing for now

PR #3824, opened 2024-04-18
feat(headless): add core support of rich formatting for CRGA

lbergeron:
# [SVCC-3604](https://coveord.atlassian.net/browse/SVCC-3604)

This PR adds the core support of rich formatting for CRGA.

Here is how the protocol with the Coveo Platform works:

1. When sending the search request, Headless sends a list of all `contentFormat`s that are supported by the front-end (e.g., `text/markdown`, `text/plain`). Default is `text/plain` to maintain backward compatibility with actual implementations.
2. When reading the stream, a header message is sent containing the actual answer format (e.g., `text/plain` or `text/markdown`).
3. The answer is concatenated in the Headless state as usual.
4. Knowing the `answerContentFormat`, the UI library can then transform the Markdown to HTML, and then render the answer.

The feature is activated in the Coveo Platform using a model association parameter.

[SVCC-3604]: https://coveord.atlassian.net/browse/SVCC-3604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olamothe:
> 
image

Oh boy...

olamothe:
Was there any attempt and/or plan to have the ML service do the transformation to HTML ?

The added size to the bundle is very significant, and I'm afraid I cannot give my approval on this approach unless were convinced it is the best possible solution to push this transformation to the front end.

Bundle size optimization is extremely important and something we have (semi) frequent complains about from our customers/implementers.

lbergeron:
> Was there any attempt and/or plan to have the ML service do the transformation to HTML ?
> 
> The added size to the bundle is very significant, and I'm afraid I cannot give my approval on this approach unless were convinced it is the best possible solution to push this transformation to the front end.
> 
> Bundle size optimization is extremely important and something we have (semi) frequent complains about from our customers/implementers.

I must admit I am also shocked by the numbers. Let's take it offline.

tedre191:
>Was there any attempt and/or plan to have the ML service do the transformation to HTML ?

There's a lot of advantages to decouple HTML generation and the model itself. With the current stack, that pretty much leaves it to the UI, which isn't a bad thing IMO (if we forget the bundle size issue).

With that being said, the new routes we are building in service could change the way we see this problem. 

Bumping the bundle size by 50% is not acceptable; but I find it hard to believe this is "correct".. surely the lib can't take that much space .. we'll investigate to understand what is going on.

packages/headless/package.json
```diff
@@ -64,8 +64,14 @@
     "node-abort-controller": "^3.0.0",
     "pino": "8.19.0",
     "redux-thunk": "3.1.0",
+    "remark-gfm": "4.0.0",
```
fbeaudoincoveo, line 67: Oof, that's a lot of new dependencies. I'm not a fan of shipping all of this in our package 😕 

That's almost a 50% increase in bundle size, and most of these libs haven't been updated in more than 6 months.

Could we take another approach? Maybe have an endpoint that can convert markdown to sanitized HTML?

louis-bompart (approved):
So this PR now focus exclusively on the support of the REST API parameter and state holding, correct?
If so, LGTM.

jelmedini (approved):
looks very good 👍

olamothe (approved):

fbeaudoincoveo (approved):

lbergeron:
> So this PR now focus exclusively on the support of the REST API parameter and state holding, correct? If so, LGTM.

@louis-bompart yes, exactly

louis-bompart:
Flakyness run noted & logged https://coveord.atlassian.net/browse/KIT-3115
Exercising admin merge.

PR #3829, opened 2024-04-19
docs(quantic): use parseString asynchronously

jpmarceau:
I noticed that the xmlMeta values came out as null in the latest quantic-docs.json
 
Screenshot 2024-04-19 at 12 20 43 PM

It seems to be due to `parseString` which requires to be run async. When I do that, the JSON looks good again:

Screenshot 2024-04-19 at 12 24 21 PM


jpmarceau:
Hello @erocheleau , @SimonMilord , @mmitiche 👋 
Could I get a review on this one? 🥺

PR #3836, opened 2024-04-22
fix(quantic): fixed issue with reseting feedback state after new queries

mmitiche:
### The issue:

The `QuanticSmartSnippet` component was not allowing feedback submission after submitting one for the first time,   

### Before

https://github.com/coveo/ui-kit/assets/86681870/4c4bfe20-3e26-4699-b7f6-8add610f36eb

### After

https://github.com/coveo/ui-kit/assets/86681870/6723425f-b050-4732-8025-ec08254b7eae

[SFINT-5503]: https://coveord.atlassian.net/browse/SFINT-5503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
nice!

packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js
```diff
@@ -110,6 +126,11 @@ export default class QuanticSmartSnippet extends LightningElement {
     }
   }
 
+  updateSearchStatusState() {
+    this.feedbackSubmitted = false;
+    this.shouldBeInNeutralState = true;
+  }
```
erocheleau, line 132:
```suggestion
  updateSearchStatusState() {
    this.feedbackSubmitted = false;
    this.liked = false;
    this.disliked = false;
  }
```

erocheleau: Why isn't this handled at the headless level instead?
Shouldn't the state of the SmartSnippet like/dislike reset on a new query instead? 

IMO we shouldn't have to handle on the Quantic side no? Otherwise we'll have the same issue within Atomic?

Shouldn't there simply be a line somewhere in here:
https://github.dev/coveo/ui-kit/blob/2851cb2a9413badf6999a420516a79b262e61c89/packages/headless/src/features/question-answering/question-answering-slice.ts#L84

that simply sets liked/disliked to false?

erocheleau (changes requested):
I don't think this is the right fix. I think the fix should be in Headless. 

Clearly this means that the liked/disliked state doesn't reset on new query no?

packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js
mmitiche: Yes the state of the Smart Snippet like/dislike should be reset on query, and that's already the case, see [L103](https://github.com/coveo/ui-kit/blob/2851cb2a9413badf6999a420516a79b262e61c89/packages/headless/src/features/question-answering/question-answering-slice.ts#L103-L104)

So at the Headless level everything is setup properly,  the only case where we do not reset in Headless is when doing a search query that returns the same answer, which I think make sense and is like that by design.

So that's why this issue is just a Quantic issue, we do not have it in Atomic mostly because the component re-rendering logic differs slightly. 

I updated the solution in Quantic to make it better  much better.
Let me know what you think.

erocheleau: 🆗

packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js
```diff
@@ -127,8 +125,9 @@ export default class QuanticSmartSnippet extends LightningElement {
   }
 
   updateSearchStatusState() {
-    this.feedbackSubmitted = false;
-    this.shouldBeInNeutralState = true;
+    if (!this.state.disliked && !this.state.liked) {
```
erocheleau, line 128: So this state would be reset correctly already in headless on each new query unless the answer is the same?

erocheleau (approved):
A lot cleaner 👍

PR #3838, opened 2024-04-22
tests: disable flaky tests

louis-bompart:
KIT-3107


fbeaudoincoveo (approved):
I'm OK with disabling those temporarily, but I believe the Jiras to reduce / eliminate their flakiness should be given high priority (as in, they should be done during the upcoming sprint, otherwise we should re-enable the tests). We don't want to have a false sentiment of security for too long ;)

PR #3840, opened 2024-04-22
refactor(atomic): phase 2 refactor facet-common

olamothe:
PR focus on "cleaning up" facet-common by trying to achieve no import from `@coveo/headless` or any `search` specific types/interfaces in what should be a "common"/"generic" file.

* Remove lots of complicated type: `AnyFacetType`, `BaseFacet`, `BaseFacetElement` etc. Those were used pretty  inconsistently throughout components, by declaring (only some) of the facet components as `implements BaseFacet`. So while they could, in theory, be useful to enforce some standardization of what a facet component should contain as property, `BaseFacet` as it stand today is too broad, and not really applicable for commerce. Couple that with the fact that many facet components were not even implementing the interface in the first place. Which mean that the interface was missing the mark. So, instead of trying to do a complex refactor that would *really enforce* all facet components to really follow the same structure, I propose to simply get rid of it. If we see a future need to really make sure all facet components offer standard properties, we can rethink that in a future PR.

* Genericize and isolate the `parseDependsOn` function, which is used for all facets that support the `depends-on` property (parent/child facet relationship).

* Move some function out of `facet-common` closer to where they are actually used (often time, we exported function from facet-common to then only use them in a single module). Example: `sortFacetsUsingManager`

* Finish removal/deletion of `FacetCommon` class by moving it's remaining duties at the top level component instead: 
 ** `initPopover()` (which makes it so a facet can be displayed inside `atomic-popover`
**  `initConditionManager()` (which makes it so a facet can support `depends-on` property as described above
** `registerFacet()` (which register facet to `atomic-store`, in order to make available the state of different facet HTML elements to other components)

https://coveord.atlassian.net/browse/KIT-3031


packages/atomic/src/components/common/facets/depends-on.ts
```diff
@@ -0,0 +1,88 @@
...
+            return value.value === expectedValue;
+          }
+          return false;
+        });
+      },
+    };
+  });
+}
+
+function isCategoryFacet(request: unknown): request is CategoryFacet {
```
fbeaudoincoveo, line 53: isCategoryFacet -> isCategoryFacetValue

packages/atomic/src/components/common/facets/depends-on.ts
```diff
@@ -0,0 +1,88 @@
...
+  for (const child of value.children) {
+    const selectedValue = getSelectedCategoryFacetValueRequest(child);
+    if (selectedValue !== null) {
+      return selectedValue;
+    }
+  }
+  return null;
+}
+
+function isSimpleFacet(value: unknown): value is SimpleFacet {
```
fbeaudoincoveo, line 81: isSimpleFacet -> isSimpleFacetValue

packages/atomic/src/components/common/facets/depends-on.ts
```diff
@@ -0,0 +1,88 @@
+type GenericCondition<AnyRequest> = {
```
fbeaudoincoveo, line 1: AnyRequest -> AnyFacetValueRequest

PR #3845, opened 2024-04-23
feat(service): add action on insight result action click

mikegron:
[SVCINT-2951](https://coveord.atlassian.net/browse/SVCINT-2951)

### Proposition
As we talked about a while back it makes sense to add a prop to the insight-result-action component and dispatch custom events. This way, custom integrations like Genesys, ServiceNow and SAP will be able to hook on those events and execute custom processing without the need for integration-specific components.

I followed this doc: https://stenciljs.com/docs/events

Ex:

https://github.com/coveo/ui-kit/assets/7459000/86e9140c-9cab-443d-8603-cd7d2a5378cf

[SVCINT-2951]: https://coveord.atlassian.net/browse/SVCINT-2951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
```diff
@@ -16,25 +23,44 @@ import {Bindings} from '../../search/atomic-search-interface/atomic-search-inter
...
   /**
-   * The text tooltip to show on the result action icon
+   * The text tooltip to show on the result action icon.
    */
   @Prop({mutable: true}) public tooltip = '';
 
+  /**
+   * The type of action to perform when the result action is clicked. This will be sent along the event fired when the button is clicked.
+   */
+  @Prop({mutable: true}) public action = '';
```
olamothe, line 50: Should this be typed in any way to stricten or guide what type of actions should be available, or do we really want `any string` ?

packages/atomic/src/pages/examples/insights.html
```diff
@@ -116,9 +116,17 @@
...
                 <atomic-insight-result-action-bar>
-                  <atomic-insight-result-action tooltip="test"></atomic-insight-result-action>
-                  <atomic-insight-result-action tooltip="test"></atomic-insight-result-action>
-                  <atomic-insight-result-action tooltip="test"></atomic-insight-result-action>
+                  <atomic-insight-result-action
+                    tooltip="copy to clipboard"
+                    action="copyToClipboard"
+                    icon='<svg width="26" height="26" viewBox="0 0 26 26" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M22 1H9C7.9 1 7 1.9 7 3V4H19C20.1 4 21 4.9 21 6V20H22C23.1 20 24 19.1 24 18V3C24 1.9 23.1 1 22 1ZM19 8C19 6.9 18.1 6 17 6H4C2.9 6 2 6.9 2 8V23C2 24.1 2.9 25 4 25H17C18.1 25 19 24.1 19 23V8ZM10 11.5C10 11.8 9.8 12 9.5 12H5.5C5.2 12 5 11.8 5 11.5V10.5C5 10.2 5.2 10 5.5 10H9.5C9.8 10 10 10.2 10 10.5V11.5ZM14 19.5C14 19.8 13.8 20 13.5 20H5.5C5.2 20 5 19.8 5 19.5V18.5C5 18.2 5.2 18 5.5 18H13.5C13.8 18 14 18.2 14 18.5V19.5ZM16 15.5C16 15.8 15.8 16 15.5 16H5.5C5.2 16 5 15.8 5 15.5V14.5C5 14.2 5.2 14 5.5 14H15.5C15.8 14 16 14.2 16 14.5V15.5Z"/></svg>'
```
olamothe, line 122: That's so so for an example 🤷

If we tighten up the available actions to be some known list instead of any string, the icon could be inferred internally in the component, without requiring implementers to provide those.

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
mikegron: You mean adding an Enum of possible actions? I would leave it to any since it just re-passes the action to events and you do whatever you want with it after anyway

packages/atomic/src/pages/examples/insights.html
mikegron: I don't like the fact that the icon is a straight up svg data nonplus.. :(

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
```diff
@@ -16,25 +23,44 @@ import {Bindings} from '../../search/atomic-search-interface/atomic-search-inter
...
+  }
+
   public render() {
     return (
       <IconButton
         partPrefix="result-action"
         style="outline-neutral"
         icon={this.icon}
         title={this.tooltip}
+        onClick={() => this.onClick()}
```
pathiery, line 63: so 
```suggestion
        onClick={this.onClick}
```
right ?

pathiery (approved):

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
mikegron: absolutely

dguerinCoveo (approved):
Nice. You could keep the event listener with the console log in the demo file maybe. I think I more inclined towards using a enum for actions too with the icons inferred from there

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
```diff
@@ -1,11 +1,30 @@
...
 import {IconButton} from '../../common/iconButton';
 import {Bindings} from '../../search/atomic-search-interface/atomic-search-interface';
+import {ResultContext} from '../../search/result-template-components/result-template-decorators';
+
+export interface InsightResultActionClickedEvent {
+  action: string;
+  result: Result;
+}
+
+enum Actions {
```
olamothe, line 21: I would suggest to refrain from using enums: They are a "weird"/un-necessary TypeScript concept that do not map super well to underlying generated JS code. VS a simple string.

```typescript
type Action = 'copyToClipboard' | 'attachToCase' | 'quickview' | 'postToFeed' | 'sendAsEmail';
```

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
olamothe: The actual accepted type should be reflected here: `public actions: Action`. Can also extend to `Action | string` if it's really the desired behaviour that it can be "anything" (I doubt it, but...)

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
```diff
@@ -1,11 +1,30 @@
-import {Component, Prop, State, h} from '@stencil/core';
-import Quickview from '../../../images/quickview.svg';
+import {Result} from '@coveo/headless';
+import {Component, Event, EventEmitter, Prop, State, h} from '@stencil/core';
+import AttachIcon from '../../../images/attach.svg';
+import CopyIcon from '../../../images/copy-dark.svg';
+import EmailIcon from '../../../images/email.svg';
+import QuickviewIcon from '../../../images/preview.svg';
```
olamothe, line 6: There's already a `Quickview` icon in the project. Here we add a new one called `preview`, but with a named import for `Quickview`. 

I think we should delete `preview.svg` ?

olamothe (approved):

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
```diff
@@ -16,25 +35,65 @@ import {Bindings} from '../../search/atomic-search-interface/atomic-search-inter
...
   /**
-   * The text tooltip to show on the result action icon
+   * The text tooltip to show on the result action icon.
    */
   @Prop({mutable: true}) public tooltip = '';
 
+  /**
+   * The type of action to perform when the result action is clicked. This will be sent along the event fired when the button is clicked.
+   */
+  @Prop({mutable: true}) public action = '';
```
olamothe, line 62: The actual accepted type should be reflected here: public actions: Action. Can also extend to Action | string if it's really the desired behaviour that it can be "anything" (I doubt it, but...)

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
olamothe: They look extremely similar (exact same?) also when you compare them visually.

packages/atomic/src/components/insight/atomic-insight-result-action/atomic-insight-result-action.tsx
mikegron: Yeah we want to let it be anything if you need to do a specific custom action for instance

PR #3846, opened 2024-04-24
fix(atomic): separator should not be consider for ARIA

louis-bompart:
Spec: https://w3c.github.io/aria/#list

## Issue

The important bit is this:
image

Essentially, anything but a role=listitem is forbidden under a `role=list`. `ol` is defacto a list, always (afaik). So `role=separator` is a nogo.

Example here: https://louis-bompart.github.io/automatic-octo-spoon/index.html

## Solution

The obvious: remove the faulty role.

Editorial choice: I don't think the separator has any value whatsoever for a screen reader, so I elected to use `aria-hidden` to make it 'disappear' for all intent and purposes from the standpoint of a screen reader or accessibility tools.

## Tests:

This PR includes flakiness proofing on the accessibility test of this component (the method could be reused on as needed).

### Wait for component to be hydrated

The flakiness of the test was reproducable on the client-side, and was indeed a 'false green'. This was caused by axe core running _before_ the component was hydrated, and on the component element nonetheless. So yeah: cool cucumber, testing no DOM is ARIA compliant :slowclap:

This add a protection on that regard: Do wait for the component to be hydrated before doing anything else.

The wait for hydration is used 'as needed' on the green path (i.e when the component actually tries to render something)

### Properly reset between retries.
Fun stuff: If you are on foo.com/acme.html and navigate to foo.com/acme.html#, you're not doing a DOM unload/load.
Essentially, it's the same stuff with cy.visit. I noted a pollution of components in-between tests, which would skew the result, I reckon.
To fix that, I added a `cy.visit('/')` (last choice: `about:blank` tries to goto `localhost:3333/about:blank` in the CI, and `chrome://dino` redirect to `https://chrome://dino`, which is very disappointing.)

### DIFF
--- a/packages/atomic/cypress/e2e/common-assertions.ts
+++ b/packages/atomic/cypress/e2e/common-assertions.ts
@@ -84,9 +84,15 @@ export function assertAccessibilityWithoutIt<T extends HTMLElement>(
   const axeOptions = {rules, retries: 3};
 
   if (typeof component === 'string') {
+    cy.get(component).then((el) => {
+      cy.log(el.html());
+      cy.log(el[0]?.shadowRoot?.innerHTML ?? '');
+    });
     cy.checkA11y(component, axeOptions, logAxeIssues);
   } else if (typeof component === 'function') {
     component().then(([el]) => {
+      cy.log(el.outerHTML);
+      cy.log(el?.shadowRoot?.innerHTML ?? '');
       cy.checkA11y(el, axeOptions, logAxeIssues);
     });
   } else {
--- a/packages/atomic/cypress/e2e/common-assertions.ts
+++ b/packages/atomic/cypress/e2e/common-assertions.ts
@@ -84,15 +84,9 @@ export function assertAccessibilityWithoutIt<T extends HTMLElement>(
   const axeOptions = {rules, retries: 3};
 
   if (typeof component === 'string') {
-    cy.get(component).then((el) => {
-      cy.log(el.html());
-      cy.log(el[0]?.shadowRoot?.innerHTML ?? '');
-    });
     cy.checkA11y(component, axeOptions, logAxeIssues);
   } else if (typeof component === 'function') {
     component().then(([el]) => {
-      cy.log(el.outerHTML);
-      cy.log(el?.shadowRoot?.innerHTML ?? '');
       cy.checkA11y(el, axeOptions, logAxeIssues);
     });
   } else {
--- a/packages/atomic/cypress/e2e/facets/facet-depends-on/facet-depends-on.cypress.ts
+++ b/packages/atomic/cypress/e2e/facets/facet-depends-on/facet-depends-on.cypress.ts
@@ -115,7 +115,6 @@ describe('Facet DependsOn Test Suites', () => {
 
     describe('when deselecting a parent facet', () => {
       beforeEach(() => {
-        cy.window().then((window) => (window.location.href = 'about:blank'));
         new TestFixture()
           .with(addThreeLevelDependantFacet())
           .withHash(hashToDisplayFacets(true, true))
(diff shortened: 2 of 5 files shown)


packages/atomic/cypress/e2e/common-assertions.ts
```diff
@@ -84,9 +84,15 @@ export function assertAccessibilityWithoutIt<T extends HTMLElement>(
   const axeOptions = {rules, retries: 3};
 
   if (typeof component === 'string') {
+    cy.get(component).then((el) => {
+      cy.log(el.html());
+      cy.log(el[0]?.shadowRoot?.innerHTML ?? '');
+    });
```
louis-bompart, line 90:
```suggestion
```

packages/atomic/cypress/e2e/common-assertions.ts
```diff
@@ -84,9 +84,15 @@ export function assertAccessibilityWithoutIt<T extends HTMLElement>(
...
     cy.checkA11y(component, axeOptions, logAxeIssues);
   } else if (typeof component === 'function') {
     component().then(([el]) => {
+      cy.log(el.outerHTML);
+      cy.log(el?.shadowRoot?.innerHTML ?? '');
```
louis-bompart, line 95:
```suggestion
```

packages/atomic/cypress/e2e/facets/facet-depends-on/facet-depends-on.cypress.ts
```diff
@@ -115,7 +115,6 @@ describe('Facet DependsOn Test Suites', () => {
 
     describe('when deselecting a parent facet', () => {
       beforeEach(() => {
-        cy.window().then((window) => (window.location.href = 'about:blank'));
```
louis-bompart, line 118: We do a navigation to '/' now instead, it's embedded in the testfixture.

packages/atomic/cypress/e2e/facets/facet-depends-on/facet-depends-on.cypress.ts
```diff
@@ -134,7 +133,7 @@ describe('Facet DependsOn Test Suites', () => {
 
         it('should not have a duplicate history state', () => {
           cy.go('back');
-          cy.location('href').should('eq', 'about:blank');
+          cy.location('pathname').should('eq', '/');
```
louis-bompart, line 136: https://github.com/coveo/ui-kit/pull/3846/files/890528bf25d988e017edac2561cd5c29ed98448f#r1577259872 ... and so blank is now pathname=/, it does the same

packages/atomic/cypress/fixtures/test-fixture.ts
```diff
@@ -284,6 +284,11 @@ export class TestFixture {
 
     return this;
   }
+  public waitForComponentHydration(componentTag: string) {
+    cy.document()
+      .find(componentTag, {timeout: 10e3, includeShadowDom: true})
+      .should('have.class', 'hydrated');
```
louis-bompart, line 290: in other words: I ain't moving till the component is cooked and ready, you have 10 seconds.

packages/atomic/src/components/search/result-template-components/atomic-result-multi-value-text/atomic-result-multi-value-text.tsx
```diff
@@ -142,7 +142,7 @@ export class AtomicResultMultiText {
   private renderSeparator(beforeValue: string, afterValue: string) {
     return (
       <li
-        role="separator"
+        aria-hidden="true"
```
louis-bompart, line 145: no mo' separator, and let's hide it for aria

packages/atomic/cypress/fixtures/test-fixture.ts
louis-bompart: Not really: for example, sometimes we'll intercept the tag (nice), but it won't be hydrated: you can see that, for example, when "you didn't put this cmp in a cmp that can hold it, unacceptable."

That was one of my first guess, but we can't do it that easily, but we should probably add it as we go for "green paths" tests

louis-bompart: I guess we could maybe add a bool for that... I'll create a separate task for that.

PR #3848, opened 2024-04-24
fix(headless): generated-answer-state expanded should be false

erocheleau:
SFINT-5520

The reason why it should be false, is because otherwise we have to send an analytics event on load of the component to “make it collapsed”.

For example, if the component is collapsible=true when loading the first time, the answer should be collapsed if it’s too long.

If expanded starts true in the state, this means that in order to have it collapsed, we need to trigger the action generatedAnswer.collapse from the headless controller. This action will also send an analytics event. Which means that on every load of the component, if collapsible=true there would be a “fake” analytics event as if the user had hit the “Show Less” button.


louis-bompart:
🤔 should we UT that?

packages/headless/src/features/generated-answer/generated-answer-state.ts
```diff
@@ -91,6 +91,6 @@ export function getGeneratedAnswerInitialState(): GeneratedAnswerState {
     feedbackSubmitted: false,
     fieldsToIncludeInCitations: [],
     isAnswerGenerated: false,
-    expanded: true,
+    expanded: false,
```
mmitiche, line 94: The only problem I see with this is that when the property `collapsible` will be set to `false` in the UI component we will have a discrepancy between the state in headless (expanded: false) and how the UI component is displaying (displaying the component in the expanded way).

I would suggest instead to keep the default state of `expanded` to `true` to reflect the default behaviour in the UI.

And add a way in the core Headless Generted Controller to set an initial state for the key `expanded` here:  https://github.com/coveo/ui-kit/blob/6a4e14ea9478b7fde2a0eda2e568448b5098a310/packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts#L121

So when in the UI component `collapsible` is set to true, we send  `expanded: false,` as an initial state for the controller which would make the state like it should be without having to log analytics.

erocheleau: The initial state of `expanded` is always false. No matter the `collapsible` option value.

If `collapsible` is false, the value of this state is simply ignored. The answer is neither `expanded` or `collapsed` it's simply not collapsible.

And the other issue I pointed out is that if `expanded` starts true, you HAVE to set it to false to make the feature work correctly. And the only way to do that is you have to trigger a changed of state that will send an analytics event on load. Which is wrong.

This state variable doesn't represent the answer being expanded if `collapsible=false` the answer is not expanded, it's simply that the component doesn't account for the expanded state.

erocheleau:
> 🤔 should we UT that?

Well the actions that have an impact on this state are already covered by UT. It just makes more sense to have it default as "false" than true.

packages/headless/src/features/generated-answer/generated-answer-state.ts
mmitiche: The fact of having a state in headless that will be given a value but that will be ignored in certain situations is something that I don't really like because for me, it makes us error prone, logical error prone. 

Especially that Headless, is not aware at all of when this state needs to be ignored and when it needs to be taken in consideration. So this piece of state initialized to `false` can be miss leading/confusing. Future developers working on Headless or using Headless would just not know that this state should be ignored until `collapsible` is set to true,  `collapsible`  which is not even in headless.

In the other hand, the alternative I'm proposing avoid this  discrepancy and solves your problem. I'm well aware that you do not want to send an analytics when collapsing  the answer when `collapsed ` is set to `true` and this solution that I'm suggesting respects this, no unwanted analytics will be sent.

## Suggested solution:

- We update the interface of the `GeneratedAnswerProps` interface([#L120-L135](https://github.com/coveo/ui-kit/blob/6a4e14ea9478b7fde2a0eda2e568448b5098a310/packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts#L120-L135)) by adding `expanded ` to the initial state:
```
export interface GeneratedAnswerProps {
  initialState?: {
    // other existing keys
    expanded?: boolean;
  };
  // other existing keys
}
```

- In the `buildCoreGeneratedAnswer` function [#L241](https://github.com/coveo/ui-kit/blob/6a4e14ea9478b7fde2a0eda2e568448b5098a310/packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts#L241C17-L241C41) we add the logic that will take in consideration the initial state of `expanded`, it would be something very similar to what we have have here for the initial state of the property `isVisible` ([#L264-L267](https://github.com/coveo/ui-kit/blob/6a4e14ea9478b7fde2a0eda2e568448b5098a310/packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts#L264-L267)) :
```
const expanded = props.initialState?.expanded;
  if (expanded !== undefined) {
    if(!expanded) {
        dispatch(collapseGeneratedAnswer());
    } else {
        dispatch(expandGeneratedAnswer());
    }
  }
```
PS: these actions dispatched here solely update the state, they DO NOT  send analytics.

- Then in Quantic:
When `collapsible` is set to `false`, we do nothing we keep creating the generated answer controller normally.

  When `collapsible` is set to `true`, we create the generated answer controller in the following way:

```
this.headless.buildGeneratedAnswer(engine, {
      initialState: {
        // other existing keys
        expanded: false
      },
      // other existing keys
    });
```

erocheleau: We had an offline chat and I think for now what we'll implement is a compromise between this and the current PR.

`expanded` will default to false, but still I will add the initial state option to allow overriding it if it's necessary in the future.

Basically this:

```
export interface GeneratedAnswerProps { 
  initialState?: { 
    // other existing keys
    expanded?: boolean; 
  }; 
  // other existing keys 
}
```

`const expanded = props.initialState?.expanded ?? false;`

jelmedini (approved):

packages/headless/src/features/generated-answer/generated-answer-state.ts
erocheleau: @mmitiche see the new changes, I added the initial state props

jelmedini (approved):

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -275,6 +279,11 @@ export function buildCoreGeneratedAnswer(
     dispatch(registerFieldsToIncludeInCitations(fieldsToIncludeInCitations));
   }
 
+  const expanded = props.initialState?.expanded;
+  if (expanded) {
+    dispatch(expandGeneratedAnswer());
```
mmitiche, line 284: We can add a small unit tests that confirms that this action has been called when the initial state `expanded` is passed.

mmitiche (approved):
A small unit test could be added, but looks good!

SimonMilord (approved):

PR #3851, opened 2024-04-24
feat(atomic): add support for RGA rich formatting

lbergeron:
# [SVCC-3606](https://coveord.atlassian.net/browse/SVCC-3606)

This PR adds support for RGA rich formatting to the Atomic Generated Answer component. Here is how it works:

1. `atomic-generated-answer` sets its capabilities using the `contentFormat` configuration parameter (e.g., `contentFormat: ['text/markdown', 'text/plain']`.
2. When reading the stream, the actual answer format is communicated to Headless via a header message. The content format is stored in the `generated-answer` state.
3. Upon rendering, the `atomic-generated-answer` component checks the content format. When receiving `text/markdown` a transformation pipeline is invoked to transform the output to HTML to be injected in the DOM.

The transformation is performed by [Marked](https://github.com/markedjs/marked). Some custom renderers have been added to handle the following:

* remove unnecessary paragraph elements when list items are separated by empty lines
* table elements are wrapped in a scrolling container to enforce a maximum height
* some inline elements such as bold, italic, and code are automatically closed to avoid showing Markdown artifacts when streaming the answer

atomic-question-1

atomic-question-2

## How to test

1. You must have a working RGA model with rich formatting enabled
2. Select the Gen Q&A
3. When asking a question, the rich formatting should kick in when using the "Auto" answer style. Rich formatting is not yet supported for other answer styles.

[SVCC-3606]: https://coveord.atlassian.net/browse/SVCC-3606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/atomic/src/themes/coveo.css
```diff
@@ -22,6 +22,13 @@
   --atomic-success-background: #d4fcf0;
   --atomic-error-background: #fcbdc0;
   --atomic-primary-background: #edf6ff;
+  --atomic-inline-code: #cd2113;
+
+  /* Table colors */
+  --atomic-table-outer-border: #bfc4c8;
```
olamothe, line 28: I get why you added those, but it means that other tables used in Atomic should be retrofitted to start using those variable.

Otherwise this is causing theming/inconsistency issues for implementers

There's some mixin that were created for `atomic-result-table`. I wonder if they could be genericized/extracted and reused for this use case... 🤔 

In `components/common/item-list/styles/mixins.pcss`

olamothe (commented):
The JS code itself is neat, and the end result is pretty cool.

However, it's really important that using Tailwind, and all of it's utility classes as much as possible is respected in Atomic. It should be rare exception that we have to "handwrite" CSS rules.

There's quite a few CSS files already for CRGA components which do not follow that rule, unfortunately. But let's try to reverse that trend.

We want to eventually put in a *lot* more effort regarding theming capability, and Tailwind is the main tool we will rely on to achieve this.

So it's not about making CSS easier for us (Coveo contributor) to write, but to make it so Atomic is "chameleon" for our customers when they want to blend it in their website.

packages/atomic/src/components/common/generated-answer/generated-content/markdownUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  if (unclosedItalic.test(text)) {
+    return text.replace(unclosedItalic, '<em>$1</em>');
+  }
+  if (unclosedCode.test(text)) {
+    return text.replace(unclosedCode, '<code>$1</code>');
+  }
+
+  return text;
+};
```
dmgauthier, line 23: I am curious about the usage of the many `return` s.

Looking at the function layout. I assume that the text will never have both a unclosed italic and a unclosed strong at the same time. Since the function will stop its execution at the italic condition.
So I guess my question is more "what does the _text_ parameter means ?", its a bit vague. is it a complete paragraph ? a word ?

I get that you extend marked with this https://marked.js.org/using_pro#use, and use this to close elements for text and list items. But I feel that we might miss something if there is the possibility of many unclosed elements.

dmgauthier:
There is a lot of work in this.  :heart:

dmgauthier (approved):

packages/atomic/src/themes/coveo.css
lbergeron: Thanks for the pointer, I'll look to harmonize that.

packages/atomic/src/components/common/generated-answer/generated-content/markdownUtils.ts
lbergeron: Very good question, let me shed some light on this.

The transformation is a two-stage process. The first pass is done by the "lexer" which takes the markdown and creates a tree of tokens. Any closed inline element is represented using a specific token. Any content that is not a known Markdown structure ends up being "text".

The second pass is done by the "parser" which takes the token tree and renders it to HTML. In this case, I'm relying on a [custom renderer](https://marked.js.org/using_pro#renderer) to try and find unclosed inline elements (that the lexer couldn't detect) and complete them.

To be fair, I tried adding the custom logic to the lexer, but didn't have much success manipulating the token tree.

To circle back to your original question, `text` can represent the content of a paragraph, a list item, a table cell, etc.

I didn't not experience issues with nested inline elements, but you are right that there could be some corner cases to handle there.

packages/atomic/src/components/common/generated-answer/generated-content/generated-markdown-content.pcss
```diff
@@ -0,0 +1,132 @@
...
+
+  blockquote {
+    font-style: italic;
+    margin-left: 3.75em;
+    margin-right: 3.75em;
+  }
+
+  .scrollable-table {
+    display: inline-block;
+    overflow: auto;
```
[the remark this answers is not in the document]
lbergeron: It is for both.

packages/atomic/src/themes/coveo.css
lbergeron: I can't get close enough to the target look and feel using the existing colors. Is it ok with you if I add one more shade of gray between `--atomic-neutral` and `--atomic-neutral-dark`? Not sure which name to use though...

packages/atomic/src/components/common/generated-answer/generated-content/generated-markdown-content.pcss
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  blockquote {
+    font-style: italic;
+    margin-left: 3.75em;
+    margin-right: 3.75em;
+  }
+
+  .scrollable-table {
```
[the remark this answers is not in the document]
lbergeron: Considering Atomic and Quantic both need their own copy of `markdown-utils.ts`, I would prefer not to add tailwind classes to the rendered HTML so it is easier to maintain parity between both.

packages/atomic/src/themes/coveo.css
lbergeron: @olamothe I went for `--atomic-neutral-dim` which, I think can fit between `--atomic-neutral` and `--atomic-neutral-dark`. Let me know if it makes sense to you.

packages/atomic/src/components/common/generated-answer/generated-content/markdownUtils.ts
dmgauthier: Like we discussed outside this thread, maybe the model will never return stuff like bold in the middle of an italic text. Trying to fix those case would be like constructing a bridge over a river that doesn't exist. 

Thanks for the clarification, in all case !

olamothe (approved):

PR #3853, opened 2024-04-25
chore(quantic): octokit rule for quantic removed from renovate

mmitiche:
## [SFINT-5436](https://coveord.atlassian.net/browse/SFINT-5436)

This rule was previously added because a new major version of `@octokit/graphql` broke  our implantation in Quantic script that uses this dependency.
This turned out to be an issue from their side that was fixed shortly after that new major version : https://github.com/octokit/graphql.js/pull/551 

Plus this renovate rules were containing a typo in the library name that prevented the rule from doing its job properly.
we had `"matchPackagePatterns": ["^@otokit/graphql$"],` instead of `"matchPackagePatterns": ["^@octokit/graphql$"],`

thus this dependency has been already updated by renovate to the latest version as you can see here: https://github.com/coveo/ui-kit/blob/1b28010482d0c88daf64aa98dae12f9a94a85c11/packages/quantic/package.json#L61-L62

[SFINT-5436]: https://coveord.atlassian.net/browse/SFINT-5436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau (approved):
So what you're saying is that it was working all along? 😂

olamothe (approved):

PR #3856, opened 2024-04-25
fix(headless/commerce): use numberOfValues from response when setting initialNumberOfValues / numberOfValues in facet request

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3122

### DIFF
--- a/packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
+++ b/packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
@@ -464,14 +464,14 @@ function updateStateFromFacetResponse(
   if (!facetRequest) {
     state[facetId] = {request: {} as AnyFacetRequest};
     facetRequest = state[facetId].request;
-    facetRequest.initialNumberOfValues = facetResponse.values.length;
+    facetRequest.initialNumberOfValues = facetResponse.numberOfValues;
   } else {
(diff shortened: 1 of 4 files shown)


Spuffynism (approved):

packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.ts
```diff
@@ -540,12 +538,18 @@ function insertNewValue(
...
 }
 
-function resetAllFacetValues(state: CommerceFacetSetState) {
+function setAllFacetValuesToIdle(state: CommerceFacetSetState) {
   Object.values(state).forEach((facet) => {
     facet.request.values.forEach((value) => (value.state = 'idle'));
   });
 }
 
+function clearAllFacetValues(state: CommerceFacetSetState) {
```
fbeaudoincoveo, line 547: A bit out of scope, but I also noticed that when the context changes, we were only resetting facet value states, not clearing the values entirely.

This implies, for example, that if a user is on PLP A in which a facet with ID X is configured and they navigate to PLP B in which a facet with ID X is also configured, then the facet would be displayed with the values from PLP A in PLP B, which makes no sense.

I believe the correct approach is to clear facet values on context changes, and set facet value states to idle when breadcrumbs are cleared.

PR #3858, opened 2024-04-25
refactor(atomic): refactor atomic category facet into function components

olamothe:
Simpler refactor than previous story/components:

* Category facet is currently exclusively used for the classic search use case. As such there is way less pollution from `@coveo/headless` in common files related to this one. It also benefits from the previous refactor for `facet-common`. The only one I had to modify was `category-facet/search-value`

* This PR focuses mostly on making sure there is no `"part"` attribute configured at the web component level, and instead extract them into lots of smaller functional components . If you refer to the design doc for  Atomic for commerce epic, one of the goal is to ensure styling consistency through `parts` attribute. So we want to make sure that atomic-commerce-category-facet re-use/output the exact same part name as classic search category facet.

https://coveord.atlassian.net/browse/KIT-3060

### DIFF
--- a/packages/atomic/src/components/common/facets/category-facet/child-value-link.tsx
+++ b/packages/atomic/src/components/common/facets/category-facet/child-value-link.tsx
@@ -1,0 +1,18 @@
+import {FunctionalComponent, h} from '@stencil/core';
+import {
+  CategoryFacetValueLink,
+  CategoryFacetValueLinkProps,
+} from './value-link';
+
+interface CategoryFacetParentValueLinkProps
+  extends Omit<CategoryFacetValueLinkProps, 'isParent'> {}
+
+export const CategoryFacetChildValueLink: FunctionalComponent<
+  CategoryFacetParentValueLinkProps
+> = (props, children) => {
+  return (
+    <CategoryFacetValueLink {...props} isParent={false}>
+      {children}
+    </CategoryFacetValueLink>
+  );
+};
--- a/packages/atomic/src/components/common/facets/category-facet/child-value-link.tsx
+++ b/packages/atomic/src/components/common/facets/category-facet/child-value-link.tsx
@@ -4,11 +4,11 @@ import {
   CategoryFacetValueLinkProps,
 } from './value-link';
 
-interface CategoryFacetParentValueLinkProps
+interface CategoryFacetChildValueLinkProps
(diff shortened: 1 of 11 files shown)


packages/atomic/src/components/common/facets/category-facet/child-value-link.tsx
```diff
@@ -0,0 +1,18 @@
+import {FunctionalComponent, h} from '@stencil/core';
+import {
+  CategoryFacetValueLink,
+  CategoryFacetValueLinkProps,
+} from './value-link';
+
+interface CategoryFacetParentValueLinkProps
```
fbeaudoincoveo, line 7: I believe this interface should be renamed to `CategoryFacetChildValueLinkProps`

packages/atomic/src/components/common/facets/category-facet/children-as-tree-container.tsx
```diff
@@ -0,0 +1,14 @@
+import {FunctionalComponent, h} from '@stencil/core';
+
+interface CategoryFacetChildrenAsTreeContainerProps {
+  className?: string;
+}
+export const CategoryFacetChildrenAsTreeContainer: FunctionalComponent<
+  CategoryFacetChildrenAsTreeContainerProps
+> = ({className}, children) => {
+  return (
+    <ul part="values" class={className ? className : ''}>
```
fbeaudoincoveo, line 10:
```suggestion
    <ul part="values" class={className ?? ''}>
```

packages/atomic/src/components/common/facets/category-facet/parent-as-tree-container.tsx
```diff
@@ -0,0 +1,18 @@
...
+export interface CategoryFacetParentAsTreeContainerProps {
+  isTopLevel: boolean;
+  className?: string;
+}
+export const CategoryFacetParentAsTreeContainer: FunctionalComponent<
+  CategoryFacetParentAsTreeContainerProps
+> = ({isTopLevel, className}, children) => {
+  return (
+    <ul
+      class={className ? className : ''}
```
fbeaudoincoveo, line 12:
```suggestion
      class={className ?? ''}
```

fbeaudoincoveo (approved):

PR #3860, opened 2024-04-25
fix: always include suggestions popup in the DOM

louis-bompart:
But hide them when needed.
This will ensure the DOM element referenced by the aria-controls always exists.

That's (kind of) what Google is doing

KIT-3125

### DIFF
--- a/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
+++ b/packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
@@ -548,11 +548,15 @@ export class AtomicSearchBox implements InitializableComponent<Bindings> {
     );
   }
 
-  private renderSuggestions() {
-    if (!this.suggestionManager.hasSuggestions) {
-      return null;
-    }
+  private get shouldShowSuggestions() {
+    return (
+      this.suggestionManager.hasSuggestions &&
+      this.isExpanded &&
+      !this.isSearchDisabledForEndUser(this.searchBoxState.value)
+    );
+  }
 
+  private renderSuggestions() {
     return (
       <div
         id={`${this.id}-popup`}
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
```diff
@@ -562,11 +566,7 @@ export class AtomicSearchBox implements InitializableComponent<Bindings> {
             : 'suggestions-single-list'
         }`}
         class={`flex w-full z-10 absolute left-0 top-full rounded-md bg-background border border-neutral ${
-          this.suggestionManager.hasSuggestions &&
-          this.isExpanded &&
-          !this.isSearchDisabledForEndUser(this.searchBoxState.value)
-            ? ''
-            : 'hidden'
+          this.shouldShowSuggestions ? '' : 'hidden'
```
olamothe, line 569: Should there be an `aria-hidden` added on the englobing div then ?

olamothe (approved):

packages/atomic/src/components/search/atomic-search-box/atomic-search-box.tsx
louis-bompart: Hm you bring a very good point, my fix is incomplete:
I need the suggestion that `div` to be in the DOM **and** visible but empty.
The rationale is that an Element is not in the accessibility tree in all four of this cases:
- The HTML [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute is present
- The element or the element's ancestor is hidden with [display: none](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
- The element or the element's ancestor is hidden with [visibility: hidden](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)
- The HTML [aria-hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) attribute is set to true.

Source: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

fbeaudoincoveo (approved):

PR #3864, opened 2024-04-26
fix(quantic): fixed issue with glitchy state caused by the standalone search box

mmitiche:
## [SFINT-5526](https://coveord.atlassian.net/browse/SFINT-5526)

The events that were dispatched by the Quantic Search Box input component were not properly stopped, so they kept propagating in the DOM causing several issues with the state in a situation where the Quantic Standalone Search box was used.

### Explanation 

To understand the issue, some knowledge about how the Quantic Standalone Search Box component operates is needed:
- When not being  in a search page:   the Quantic Standalone Search Box component acts like a real standalone search box that init a new engine that we will call the `Standalone Engine`.
- When being  in a search page:   the Quantic Standalone Search Box component acts like a normal search box that uses a another engine used by the rest of the search page  that we will call the `Search Page Engine`.

Since the events dispatched by the Quantic Search Box Input component were not stopped, it meant that these events were caught by some wrong listeners that are defined in the `QuanticStandaloneSearchBox`. these listeners were modifying the state `Standalone Engine` while they should have modified the state of only the `Search Page Engine`.

### Before:
https://github.com/coveo/ui-kit/assets/86681870/f7df2979-888b-4579-a6f1-b1b32d0a28da

### After:

https://github.com/coveo/ui-kit/assets/73316533/1454a8c7-4186-4f70-873e-cf78467ae4c7

[SFINT-5526]: https://coveord.atlassian.net/browse/SFINT-5526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
tested it on my end, works well ! 👍

tedre191 (approved):

PR #3873, opened 2024-04-30
refactor(atomic): genericize numeric-facet-common for commerce

olamothe:
Genericize `numeric-facet-common` to support commerce use case.

KIT-3061


packages/atomic/src/components/common/facets/numeric-facet/formatter.ts
```diff
@@ -0,0 +1,67 @@
...
+  formatter,
+}: FormatFacetValueRange) => {
+  const manualRangeLabel = manualRanges.find((range) =>
+    areRangesEqual(range, facetValue)
+  )?.label;
+  return manualRangeLabel
+    ? getFieldValueCaption(field, manualRangeLabel, i18n)
+    : i18n.t('to', {
+        start: formatNumberLocalized(facetValue.start, i18n, logger, formatter),
+        end: formatNumberLocalized(facetValue.end, i18n, logger, formatter),
```
fbeaudoincoveo, line 37: When the the range is not endInclusive, should we subtract 1 from its end?

fbeaudoincoveo (approved):

packages/atomic/src/components/search/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
```diff
@@ -287,20 +340,173 @@ export class AtomicNumericFacet implements InitializableComponent {
...
+                ></atomic-facet-number-input>
+              ),
+            ]}
+          </FacetContainer>
+        ) : (
+          <FacetPlaceholder
+            isCollapsed={this.isCollapsed}
+            numberOfValues={this.numberOfValues}
+          />
+        )}
```
fpbrault, line 391: I noticed that in atomic-result-list, instead a ternary, we have two guards (ItemDisplayGuard and ResultsPlaceholdersGuard) to control whether the placeholder of the result is displayed. 

Are we using the ternary here because we have a simple condition or is there a deeper reason?

fpbrault (approved):
Looks good to me!

packages/atomic/src/components/common/facets/numeric-facet/formatter.ts
olamothe: Good point, that means this bug was there for a long time :D

packages/atomic/src/components/search/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
olamothe: Hummm for result list, it's a bit more complex because of different list layouts.

PR #3876, opened 2024-04-30
fix(quantic) : Issue with margin bottom of facets cards when DNE is disabled

SimonMilord:
[SFINT-5494](https://coveord.atlassian.net/browse/SFINT-5494)

### The issue:

The margin in the HIP refine modal content is not the same whether DNE is enabled or not. They should be the same.

There is margin for facets with DNE, because the `quanticFacetManager` is adding dynamically the class to add the margin on its items.

ie: `this.itemTemplate.classList.add('slds-var-m-bottom_large');`

### Solution: 
I added a wrapper around the `c-quantic-card-container` with the following slds class: `slds-var-m-bottom_large` to add the same margin bottom as when DNE is enabled. 

This wrapper was already present in the `quanticFacet` component so I added it for the other facet types.

I checked (as you can see in the demo) that is was not breaking the following facets scenarios:

- HIP (refine toggle modal)
- HIP (fullsearch facets)
- HSP (regular HSP facets)
- HSP (refine toggle modal on small screens)

### DEMO: 

https://github.com/coveo/ui-kit/assets/73316533/a74afbd6-cfe6-4fc2-867d-966f6da21a2a

[SFINT-5494]: https://coveord.atlassian.net/browse/SFINT-5494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau (changes requested):
That change is good, but you need to also "remove" the automatically added class by the DNE facets, otherwise we'll have double margins no?

packages/quantic/force-app/main/default/lwc/quanticFacetManager/quanticFacetManager.js
```diff
@@ -129,7 +129,6 @@ export default class QuanticFacetManager extends LightningElement {
 
   resolveItemTemplate() {
     this.itemTemplate = document.createElement('div');
-    this.itemTemplate.classList.add('slds-var-m-bottom_large');
```
SimonMilord, line 132: removed the margin here since we are changing the templates directly in the HTML to support both DNE enabled/disabled

erocheleau (approved):
Looks good now

tedre191 (approved):

PR #3879, opened 2024-04-30
feat(headless/commerce): support continuous ranges on numeric facets

fbeaudoincoveo:
https://coveord.atlassian.net/browse/CAPI-720

Initially when we implemented commerce numeric facets, ranges had to be explicitly specified in the CMH configuration.

It is now also possible to configure numeric facets with automatic discrete ranges, or with a single continuous range in the CMH.

Every facet response now includes a `domain` property indicating the `min` and `max` range values that will return products when a set as a selected range's `start` and `end` values respectively.

The `domain` also includes an `increment` property that indicates the number of "steps" between each value. This will always be `0` if the range is continuous, or if there is no regularity between the discrete ranges (e.g., if the discrete ranges overlap or "jump" steps).

The `domain` property is not absolutely necessary for discrete ranges, although an implementer could leverage it to allow end users to "add" and select valid arbitrary numeric ranges in a facet (e.g., by entering valid start and end values in an input).

The `domain` is absolutely necessary to implement continuous facets (e.g., sliders), however.

In this pull request, I added support for the `domain` on the request and response types (and did some refactoring too...), as well as on the numeric facet controller state (I did not expose the `increment` property in the numeric facet controller state, as I don't see an immediate use for it, though). I also exposed a `setRanges` method on the numeric facet controller.

Corresponding PR in Barca Sports: https://github.com/coveo/barca-sports/pull/243

### DIFF
--- a/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
+++ b/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
@@ -5,11 +5,11 @@ import {
 } from '../../../../app/commerce-engine/commerce-engine';
 import {deselectAllBreadcrumbs} from '../../../../features/breadcrumb/breadcrumb-actions';
 import {commerceFacetSetReducer as commerceFacetSet} from '../../../../features/commerce/facets/facet-set/facet-set-slice';
+import {FacetType} from '../../../../features/commerce/facets/facet-set/interfaces/common';
 import {
   AnyFacetResponse,
   AnyFacetValueResponse,
   BaseFacetValue,
-  FacetType,
 } from '../../../../features/commerce/facets/facet-set/interfaces/response';
 import {toggleSelectCategoryFacetValue} from '../../../../features/facets/category-facet-set/category-facet-set-actions';
 import {facetOrderReducer as facetOrder} from '../../../../features/facets/facet-order/facet-order-slice';
--- a/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
+++ b/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
@@ -50,6 +50,10 @@ export interface Breadcrumb<Value extends BaseFacetValue> {
    */
   facetId: string;
   /**
+   * The display name of the underlying facet.
+   */
+  facetDisplayName: string;
+  /**
    * The field on which the underlying facet is configured.
    */
   field: string;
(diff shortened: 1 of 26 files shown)


packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.test.ts
```diff
@@ -1955,6 +2006,92 @@ describe('commerceFacetSetReducer', () => {
...
+        values: newValues,
+      });
+      const finalState = commerceFacetSetReducer(state, action);
+
+      const targetValues = finalState[facetId]?.request.values;
+      expect(targetValues).toEqual(convertToNumericRangeRequests(newValues));
+    });
+  });
+
+  describe('#updateFacetNumberOfValues', () => {
```
fbeaudoincoveo, line 2067: Out of scope, but we were not testing this case so I added it.

packages/headless/src/controllers/commerce/core/facets/numeric/headless-commerce-numeric-facet.ts
```diff
@@ -28,6 +59,14 @@ export type NumericFacet = CoreCommerceFacet<
   NumericRangeRequest,
   NumericFacetValue
 > & {
+  /**
+   * Replaces the current range values with the specified ones.
+   * @param ranges - The new ranges to set.
+   */
+  setRanges: (ranges: NumericFacetValue[]) => void;
```
Spuffynism, line 66: I would have assumed this would be capped to a length of two, to set the min and the max of the currently selected range.

Why is this an array?

Spuffynism (approved):

packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
```diff
@@ -142,6 +146,7 @@ export function buildCoreBreadcrumbManager(
 
   const createBreadcrumb = (facet: AnyFacetResponse) => ({
     facetId: facet.facetId,
+    facetDisplayName: facet.displayName,
```
fbeaudoincoveo, line 149: Also out of scope, but this was a quick win. We need the facet display name to properly display breadcrumbs in a UI, otherwise they're just going to indicate the selected values with no further context, which can get confusing.

packages/headless/src/controllers/commerce/core/facets/numeric/headless-commerce-numeric-facet.ts
fbeaudoincoveo: Initially, I had made the function accept a single range, but I decided to go with an array because this will allow developers to implement numeric facets into which the user can enter arbitrary ranges (e.g., by using numeric inputs).

When the user adds a new range in this fashion, the facet would call the setRanges method and pass the current ranges + the new range. We can validate that the user-entered range is valid by comparing it to the facet's domain.

olamothe (approved):

PR #3881, opened 2024-04-30
feat(headless): expose cartAction method

samisayegh:
**Context**
The cart controller's `updateItem` currently does two things: a) it updates the state of the cart in headless, and b) logs a cart analytics event. The second behavior makes the `cartAction` event an exception. Whereas all other events are uncoupled (i.e. `productClick`, `productView` and `purchase`), the `cartAction` is coupled.

This design has created some challenges:
1. Conforama wants to use GTM to track events. Coupling meant that it would not be possible to opt-out of headless's behavior, and would result in duplicate cart events being sent by headless and GTM.
2. When Barca Sports initializes the cart controller, it calls [updateItem](https://github.com/coveo/barca-sports/blob/DEV/Components/Cart/cart-controller.ts#L27) to synchronize the state. This emits a burst of `cartAction` events if the cart is populated, even if the user did not add anything to the cart. Initializing the controller using `initialState.items` avoids the issue, but it's easy for an implementer to make this mistake. Worth noting there's a [TODO to expose batch helpers](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts#L156) which might have prevented the issue if it doesn't send analytics events. More on this later.
3. From a documentation perspective, coupling the `cartAction` means it will always be different. The docs will explain how to opt-in to sending click, view and purchase, but we'll need a disclaimer explaining that cart events are handled and it is not possible to opt-out.

**Approach**
The PR decouples the analytics event from the state change, providing separate methods for each.

**Open questions**
- The `quantity` in `cartAction` expects a diff rather than an absolute number. A concern is this could become a recurring point of confusion, where a controller exposes two different approaches to `quantity`.
- Could it be helpful if `updateItem` used a diff api too? Setting quantity to `-1` decreases the amount by one, stopping at 0? Methods like `setItem` or a batch `setItems` would provide an absolute API?

CAPI-809

### DIFF
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
@@ -25,6 +25,8 @@ import {
   totalQuantitySelector,
 } from './headless-cart-selectors';
 
+export type CartActionPayload = Omit<Ec.CartAction, 'currency'>;
+
 export interface CartInitialState {
   items?: CartItemWithMetadata[];
 }
@@ -72,8 +74,7 @@ export interface CartProps {
  */
 export interface Cart extends Controller {
   /**
-   * Creates, updates, or deletes an item in the cart, and emits an `ec.cartAction` analytics event if the `quantity` of
-   * the item in the cart changes.
+   * Creates, updates, or deletes an item in the cart.
    *
    * If an item with the specified `productId` already exists in the cart:
    * - Setting `quantity` to `0` deletes the item from the cart; `name` and `price` have no effect.
@@ -84,19 +85,21 @@ export interface Cart extends Controller {
    * - Setting `quantity` to a positive number creates the item in the cart with the specified `productName`,
    * `pricePerUnit`, and `quantity`.
    *
-   * If the specified `quantity` is equivalent to the current quantity of the item in the cart, no analytics event is
-   * emitted. Otherwise, the method emits an `ec.cartAction` event with the appropriate action (`add` or `remove`).
-   *
    * @param item - The cart item to create, update, or delete.
    */
   updateItem(item: CartItem): void;
 
   /**
    * Sets the quantity of each item in the cart to 0, effectively emptying the cart.
-   *
-   * Emits an `ec.cartAction` analytics event with the `remove` action for each item removed from the cart.
    */
   empty(): void;
+
+  /**
+   * Emits an `ec.cartAction` analytics event.
+   *
+   * @param transaction - The object with the id and the total revenue from the transaction, including taxes, shipping, and discounts.
+   */
+  cartAction(payload: CartActionPayload): void;
 
   /**
    * Emits an `ec.purchase` analytics event and then empties the cart without emitting any additional events.
@@ -158,22 +161,6 @@ export function buildCart(engine: CommerceEngine, props: CartProps = {}): Cart {
     dispatch(setItems(initialState.items));
   }
 
-  function isNewQuantityDifferent(
-    currentItem: CartItem,
-    prevItem: CartItemWithMetadata
-  ) {
-    return prevItem ? prevItem.quantity !== currentItem.quantity : true;
-  }
-
-  function getCartAction(
-    currentItem: CartItem,
-    prevItem: CartItemWithMetadata | undefined
-  ): 'add' | 'remove' {
-    const isCurrentQuantityGreater =
-      !prevItem || currentItem.quantity > prevItem.quantity;
-    return isCurrentQuantityGreater ? 'add' : 'remove';
-  }
-
   function getCurrency(): CurrencyCodeISO4217 {
     return engine.state.commerceContext.currency;
   }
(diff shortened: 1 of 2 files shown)


packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -84,20 +85,22 @@ export interface Cart extends Controller {
...
    * Sets the quantity of each item in the cart to 0, effectively emptying the cart.
-   *
-   * Emits an `ec.cartAction` analytics event with the `remove` action for each item removed from the cart.
    */
   empty(): void;
 
+  /**
+   * Emits an `ec.cartAction` analytics event.
+   *
+   * @param transaction - The object with the id and the total revenue from the transaction, including taxes, shipping, and discounts.
```
Spuffynism, line 100: The documentation for the parameter is not quite right

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -84,20 +85,22 @@ export interface Cart extends Controller {
    * - Setting `quantity` to a positive number creates the item in the cart with the specified `productName`,
    * `pricePerUnit`, and `quantity`.
    *
-   * If the specified `quantity` is equivalent to the current quantity of the item in the cart, no analytics event is
-   * emitted. Otherwise, the method emits an `ec.cartAction` event with the appropriate action (`add` or `remove`).
-   *
    * @param item - The cart item to create, update, or delete.
    */
   updateItem(item: CartItem): void;
```
samisayegh, line 90: Another approach could be to add a second optional `track` parameter, defaulting to `false`.
The key theme to solve the different issues in the PR description is allowing implementers to opt-in to the behavior, whereas today they do not have such control,

Spuffynism: I think it would feel weird to have a boolean flag to optionally trigger relay events. Other events triggered through headless don't have a similar mechanism. I think I prefer splitting the logic in two methods 🤔

Spuffynism (commented):
Could we rename `updateItem` to `setItem`? `setItem` having an absolute API seems better to me than `updateItem` having an absolute API

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
samisayegh: Is it ok to have two different api styles?

If not, we could use two methods, but to name them explicitly. e.g. `updateItem` and `updateItemAndTrackCartAction`
If one needs to sync state, they use the first one, if they want to update state and track, they use the second.

We'd also do the same for `empty`? `emptyAndTrackCartAction` ...

If the naming is too verbose, another idea could be to create property `withAnalytics` that exposes these methods. e.g.

```
cart.withAnalytics.updateItem()
cart.withAnalytics.empty()
``` 
With such a pattern, would `purchase` need to move under the `withAnalytics` object?

sallainCoveo: > With such a pattern, would purchase need to move under the withAnalytics object?

It might be relevant to mention here, but I don't think `purchase` is fully decoupled. If I remember correctly, when you call `purchase`, you empty the cart as well. In that sense, they're would be two patterns the implementer might have to choose when implementing their solution; one with analytics and one without.

samisayegh: Great point, confirming that `purchase` [empties the cart](https://github.com/coveo/ui-kit/blob/master/packages/headless/src/features/commerce/context/cart/cart-slice.ts#L41)

This suggests that the current PR is not right. The supported modes on the cart controller might be:
- state-only
- state + analytics
but not analytics-only

louis-bompart (changes requested):
I disagree/question the premise of this PR.
For GTM, the solution is not client-side, but server-side:
Emit the event to Coveo using EP, then recover the enriched data from DataPlatform using the Reader Account to push'em' in GTM.

The two other points are IMHO too weak to be worth introducing a "you messed up your analytics path"

We should document better than there's no opt-out, and keep the no-opt out in my opinion.

I think tho that PR review is not the right forum for this discussion, @samisayegh let's chat ;)

Leaving a RC review as blocking.

louis-bompart:
I took the liberty of converting this to a draft, as our discussions are inconclusive for now

samisayegh:
Discussed the other week with @BenKandelaars, @Spuffynism, @sallainCoveo and @fbeaudoincoveo and left with two decisions:
1. If a customer is using headless, we will recommend using headless controllers to send analytics rather than GTM. Customers are free to forward events to GTM to send them to other services, but not to send them to Coveo. Conforama has agreed to this approach.
2. Making it clearer that the `updateItem` method should not be used to sync initial state through documentation, but perhaps also by making the cart are a required option when initializing the commerce engine.

Thanks all for sharing your thoughts, closing this PR,

PR #3883, opened 2024-05-02
KIT-3144: mark atomic commerce components as internal

jpmarceau:
Checked with @olamothe and the components shouldn't be publicly documented yet

https://coveord.atlassian.net/browse/KIT-3144


fpbrault (approved):
Thanks for spotting this!

PR #3890, opened 2024-05-02
feat(headless/commerce): make commerce engine state internal

fbeaudoincoveo:
https://coveord.atlassian.net/browse/CAPI-768

@spiracha-coveo I'm adding you as a reviewer on this one for visibility, because this is potentially breaking for the implementation you're working on. The engine's state will no longer be publicly accessible. The only way to interact with the state will be through controllers and manually dispatching actions.


louis-bompart:
FYI @dmgauthier: We will want to use a similar approach to 'hide' the state with QAPI

packages/headless/src/app/commerce-engine/commerce-engine.ts
```diff
@@ -59,18 +60,17 @@ export type CommerceEngineState =
...
-   *
-   * @param analyticsEvent - The analytics event to log in association with the first search. If unspecified, `logInterfaceLoad` will be used.
    */
   executeFirstSearch(): void;
 
   /**
-   * Executes the first search, and logs the analytics event that triggered a redirection from a standalone search box.
-   *
-   * @param analytics - The standalone search box analytics data.
+   * Executes the first search after a redirection from a standalone search box.
```
fbeaudoincoveo, line 73: Note to self: given that we're not logging client side events with the commerce sub-package, this actually does the exact same thing as executeFirstSearch... could probably be removed, although I see that @fpbrault added it, so it's probably there for a reason.

packages/headless/src/controllers/commerce/core/facets/category/headless-commerce-category-facet-search.ts
```diff
@@ -37,11 +38,19 @@ export function buildCategoryFacetSearch(
...
-      executeFacetSearchActionCreator: executeCommerceFacetSearch,
-      executeFieldSuggestActionCreator: executeCommerceFieldSuggest,
-    });
+    buildCoreCategoryFacetSearch(
+      {
+        ...engine,
+        get state() {
+          return engine[stateKey];
+        },
+      },
```
Spuffynism, line 47: Why do we need to do this? Won't the `buildCoreCategoryFacetSearch` access the engine state through the `stateKey` as well?

Spuffynism (approved):

packages/headless/src/controllers/commerce/core/facets/category/headless-commerce-category-facet-search.ts
fbeaudoincoveo: Initially I hesitated to change the type of the "engine" argument to CoreEngine<...> | CoreEngineNext<...>, but I did so, so oddness is now gone 😄

packages/headless/src/app/engine.ts
```diff
@@ -99,6 +100,23 @@ export interface CoreEngine<
...
+   * The readonly internal state of the headless engine.
+   *
+   * @internal
+   */
+  readonly [stateKey]: State & CoreState;
+
+  /**
+   * The readonly global headless engine configuration
+   */
+  readonly configuration: EngineConfiguration;
```
fbeaudoincoveo, line 117: I'm exposing this here because there's not configuration controller.

Also in Atomic, we retrieve information from the configuration state in [query-error-common.tsx](https://github.com/coveo/ui-kit/pull/3890/files#diff-9d7602efb11be67faba116b145f0fc95a945d94657859c4e7d971637672c7a3aL74)

packages/headless/src/app/commerce-engine/commerce-engine.ts
```diff
@@ -113,24 +113,30 @@ export function buildCommerceEngine(
...
-    get state() {
-      return engine.state;
+    get [stateKey]() {
+      return internalEngine.state;
     },
+
+    get configuration() {
+      return internalEngine.state.configuration;
+    },
+
```
dmgauthier, line 131: This is interesting !
why
```
{
    [stateKey]: internalEngine.state,
    configuration: internalEngine.state.configuration
}
```
is not working here ?

fbeaudoincoveo: Because we need to return the current value of the state every time

E.g.,

```javascript
const test = { state: 'old' };

function stagnantStateGetter(arg) {
  return {
    state: arg.state
  };
};

const bad = stagnantStateGetter(test);

console.log(bad.state);
// prints 'old'

test.state = 'new'
console.log(bad.state)
// still prints 'old' :`(

function freshStateGetter(arg) {
  return {
    get state() {
      return arg.state;
    }
  };
}

const good = freshStateGetter(test);

console.log(good.state)
// prints 'new'

test.state = 'newest'
console.log(good.state)
// prints 'newest' :D
```

packages/headless/src/app/commerce-engine/commerce-engine.ts
```diff
@@ -113,24 +113,30 @@ export function buildCommerceEngine(
...
+  const {state: _, ...engine} = internalEngine;
 
   engine.dispatch(setContext(options.configuration.context));
 
   return {
     ...engine,
 
-    get state() {
-      return engine.state;
+    get [stateKey]() {
```
louis-bompart, line 124: TIL you can use [var] with a getter like so. I thought you had to use Object.defineProperty to define a prop with  Symbol as key
But this is much betta

packages/headless/src/app/engine.ts
louis-bompart: Should we have a configuration controller?

packages/headless/src/app/commerce-engine/commerce-engine.ts
dmgauthier: I was the feeling this was something like this. Thanks for the very detailled explanation. :bowing_man: Its very appreciated !

dmgauthier (approved):

Spuffynism (approved):

packages/headless/src/app/engine.ts
fbeaudoincoveo: I don't think so, no; if you need to adjust the engine configuration, you would normally re-initialize the engine.

louis-bompart (approved):

packages/atomic/src/components/common/query-error/query-error-common.tsx
```diff
@@ -74,11 +74,15 @@ export const QueryErrorCommon: FunctionalComponent<QueryErrorCommonProps> = (
...
+    return 'state' in props.bindings.engine
+      ? props.bindings.engine.state.configuration.organizationId
+      : props.bindings.engine.configuration.organizationId;
   };
 
   const getUrl = () => {
-    return props.bindings.engine.state.configuration.platformUrl;
+    return 'state' in props.bindings.engine
+      ? props.bindings.engine.state.configuration.platformUrl
+      : props.bindings.engine.configuration.platformUrl;
```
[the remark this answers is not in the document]
fbeaudoincoveo: If we remove the state from the other engines we'll have to 😄

PR #3891, opened 2024-05-03
chore: bump cypress timeout

louis-bompart:
A naive approach, but well worth a try.


fpbrault (approved):
Let's see if that does the trick!

fbeaudoincoveo (approved):
🤞

PR #3898, opened 2024-05-06
feat(headless/commerce): add commerce triggers actions and reducer

fbeaudoincoveo:
https://coveord.atlassian.net/browse/CAPI-864

I also did some refactoring to keep things nice and tidy and avoid code duplication.

I'll add the commerce redirection triggers controller in a subsequent PR, and then we'll see about adding controllers for other types of triggers (notify, execute, query), but these aren't as much a priority for now.


packages/headless/src/features/commerce/triggers/triggers-slice.ts
```diff
@@ -0,0 +1,34 @@
...
+  applyQueryTriggerModification,
+  updateIgnoreQueryTrigger,
+} from './triggers-actions';
+
+export const commerceTriggersReducer = createReducer(
+  getTriggerInitialState(),
+  (builder) =>
+    builder
+      .addCase(executeSearch.pending, handleFetchItemsPending)
+      .addCase(fetchProductListing.pending, handleFetchItemsPending)
```
Spuffynism, line 21: I think we could leave out the product listing cases out for now.

packages/headless/src/features/triggers/triggers-slice-functions.ts
```diff
@@ -0,0 +1,69 @@
+import {Trigger} from '../../api/common/trigger';
```
fbeaudoincoveo, line 1: These were simply extracted from the original search triggers slice (see https://github.com/coveo/ui-kit/pull/3898/files#diff-1c85a79538d616be9a043b08cbcddcc8801424d7b2be29b46944e24594d29065).

I added return statements only for testing purposes.

Spuffynism (approved):

packages/headless/src/features/triggers/triggers-slice.test.ts
```diff
@@ -1,233 +1,92 @@
 import {buildMockSearch} from '../../test/mock-search';
```
fbeaudoincoveo, line 1: This is almost identical to the commerce triggers-slice-tests (https://github.com/coveo/ui-kit/pull/3898/files#diff-4b3628fa3e0f90d193221c704dc72c90b96b3c72e8e46591942d853f29858099) except there are fewer cases to test (because the Commerce API returns triggers in product listing calls as well)

packages/headless/src/features/triggers/triggers-slice-functions.test.ts
```diff
@@ -0,0 +1,206 @@
+import {Trigger} from '../../api/common/trigger';
```
fbeaudoincoveo, line 1: The commerce and non-commerce triggers reducers share the exact same logic, so I think it makes sense to extract this logic into functions, test those functions thoroughly, call those functions in the reducers, and test that those functions are getting called in the reducers.

fbeaudoincoveo: (this may not be a pattern we want to adopt widely, but I wanted to try it out)

olamothe (approved):

PR #3899, opened 2024-05-06
feat(headless/commerce): add commerce redirection trigger controller

fbeaudoincoveo:
https://coveord.atlassian.net/browse/CAPI-865

This builds on the changes introduced in https://github.com/coveo/ui-kit/pull/3898

### DIFF
--- a/packages/headless/src/controllers/core/triggers/headless-core-redirection-trigger.ts
+++ b/packages/headless/src/controllers/core/triggers/headless-core-redirection-trigger.ts
@@ -1,0 +1,21 @@
+import {Controller} from '../../controller/headless-controller';
+
+/**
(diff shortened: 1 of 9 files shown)


packages/headless/src/controllers/core/triggers/headless-core-redirection-trigger.ts
```diff
@@ -0,0 +1,21 @@
+import {Controller} from '../../controller/headless-controller';
```
fbeaudoincoveo, line 1: We could potentially extract an actual "core" controller, but this is something I'd rather do once https://github.com/coveo/ui-kit/pull/3890 is merged.

fbeaudoincoveo:
After discussion with @olamothe , I realized that even though we have it for the search use case, this controller is not actually useful. Redirect triggers should always be handled from the standalone search box when the `submit` method is called. I'll close this PR and open another one.

For now, the standalone search box will get "preprocess triggers" (i.e., the redirection triggers) from the Commerce /search endpoint. We'll update this to get them from the Commerce API's equivalent of the /plan endpoint as soon as it becomes available.

PR #3900, opened 2024-05-06
fix(quantic): the rephrase buttons were not responsive correctly.

erocheleau:
[SFINT-5509]

Before:
image

After (with multi-line footer)

https://github.com/coveo/ui-kit/assets/2488155/5f0b3520-7512-4e18-8a1c-99c7630f29a1

(Pay no attention to the Sort & Filter button that is misaligned)

After (without multi-line footer, with collapsible):

https://github.com/coveo/ui-kit/assets/2488155/20a17954-4206-4675-9b7d-90179ea1d1cd

[SFINT-5509]: https://coveord.atlassian.net/browse/SFINT-5509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau:
I know I was forced to use absolute values for min and max width of some components, but I feel like they make sense in the context where they are used. And they are simply used as breakpoints for elements who on purpose have a fixed min or max width.

AKA I cannot combine @container media queries to hide labels of rephrase buttons with `flex-shrink: 0` for example to make a real component that has a normal behavior. I would need JS to do that I am pretty sure.

SimonMilord (approved):
LGTM! Demo looks sick 🎉

SimonMilord:
> Very nice solution using the [@container](https://developer.mozilla.org/en-US/docs/Web/CSS/@container#syntax), well done!
> 
> Just to double check, was this tested in other browsers like Safari/ Firefox? it is supposed to be supported, [browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/@container#browser_compatibility), but just to avoid any surprises in Salesforce.
> 
> Also to avoid having the same issue more than once, we wanted to adapt the Example Quantic Insight Panel to have a more similar size to the HIP that is package in the C4SF, and this to spot next any styling issues earlier in the development process. I think doing this in this PR would be suitable.
> 
> Last but not least, this issue seems to me like an important fix, so in the next released version of the C4SF, it will be very nice to have this fix included and avoid having the RGA displayed with this styling issues we are addressing here.

We were also thinking maybe that could be a potential solution for the issues with the pager if we decide not to do the load more

erocheleau:
> Very nice solution using the [@container](https://developer.mozilla.org/en-US/docs/Web/CSS/@container#syntax), well done!
> 
> Just to double check, was this tested in other browsers like Safari/ Firefox? it is supposed to be supported, [browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/@container#browser_compatibility), but just to avoid any surprises in Salesforce.
> 
> Also to avoid having the same issue more than once, we wanted to adapt the Example Quantic Insight Panel to have a more similar size to the HIP that is package in the C4SF, and this to spot next any styling issues earlier in the development process. I think doing this in this PR would be suitable.
> 
> Last but not least, this issue seems to me like an important fix, so in the next released version of the C4SF, it will be very nice to have this fix included and avoid having the RGA displayed with this styling issues we are addressing here.

I did check with a smaller width Insight Panel but yes I compared and the width of the example Insight Panel is in line with the regular space that we have within Salesforce. So I would assume this was discovered because someone was using a smaller screen or a smaller sidebar for the HIP than what is the default.

PR #3902, opened 2024-05-06
feat(Commerce Atomic): add product text component

fpbrault:
This PR adds atomic-product-text and atomic-commerce-text.

https://coveord.atlassian.net/browse/KIT-3150


packages/atomic/src/components/commerce/result-template-components/atomic-product-text/atomic-product-text.tsx
```diff
@@ -0,0 +1,92 @@
...
+    );
+    if (isArray(productValueRaw)) {
+      this.bindings.engine.logger.error(
+        `atomic-product-text cannot be used with multi value field "${this.field}" with values "${productValueRaw}".`,
+        this
+      );
+    }
+  }
+
+  public render() {
```
olamothe, line 54: We should add a story in the epic to support highlighting here. We don't have to do it for GA, but to keep track.

olamothe (approved):

PR #3908, opened 2024-05-07
chore!: remove abab

### LINKED ISSUE #3897: [Bug]: dependency abab has been deprecated from npm

Headless

### product version

2.63.0

### bug description

https://github.com/coveo/ui-kit/blob/master/packages/headless/package.json#L58 the dependecy of headless includes a deprecated module "abab", would you mind consider replace/remove it.

Image 2024-5-6 at 13 28

### Steps to reproduce

please kindly check https://github.com/coveo/ui-kit/blob/master/packages/headless/package.json#L58

### Relevant log output

_No response_

louis-bompart:
https://coveord.atlassian.net/browse/KIT-3160

Fixes #3897


olamothe (approved):
To be "pedantic", I guess this should go in v3. 

Since that's when we will pin node to whatever the latest major is at that point in time.

But, I think the changes of it breaking clients implementation is relatively low...

louis-bompart:
> To be "pedantic", I guess this should go in v3.
> 
> Since that's when we will pin node to whatever the latest major is at that point in time.
> 
> But, I think the changes of it breaking clients implementation is relatively low...

Technically true, yeah. Will put it on draft/hold.

fpbrault (approved):

PR #3910, opened 2024-05-07
feat: add quickview-modal-header-title

louis-bompart:
❓ How to document parts on 'internal' components?

https://coveord.atlassian.net/browse/KIT-3112


louis-bompart:
Discussed internally, we should essentially make the component public but uninitializable by external users if possible, or document "don't use that"

fbeaudoincoveo (approved):

olamothe (approved):
Will need a manual build + commit generated files to fix build

PR #3912, opened 2024-05-07
fix: default on empty str for query when restoring search params

louis-bompart:
CAPI doesn't like empty/missing query.

https://coveord.atlassian.net/browse/KIT-3161


packages/headless/src/features/commerce/query/query-state.ts
```diff
@@ -1,5 +1,5 @@
 export interface CommerceQueryState {
-  query?: string;
+  query: string;
```
louis-bompart, line 2: Essentially thee shall never have no query in your state.

Spuffynism (approved):

PR #3923, opened 2024-05-10
chore: add back textarea

louis-bompart:
let's figure out why it's going haywire later on, eh.

### DIFF
--- a/packages/atomic/src/pages/examples/commerce-website/search.html
+++ b/packages/atomic/src/pages/examples/commerce-website/search.html
@@ -29,7 +29,7 @@
         <atomic-commerce-interface type="search" style="max-width: 1000px; margin: auto">
           <atomic-search-layout>
             <atomic-layout-section section="search">
-              <atomic-commerce-search-box>
+              <atomic-commerce-search-box textarea>
                 <atomic-commerce-search-box-recent-queries></atomic-commerce-search-box-recent-queries>
                 <atomic-commerce-search-box-query-suggestions></atomic-commerce-search-box-query-suggestions>
                 <atomic-commerce-search-box-instant-products


fbeaudoincoveo (approved):
🤷

packages/atomic/src/pages/examples/commerce-website/search.html
```diff
@@ -29,7 +29,7 @@ <h1>Search page</h1>
         <atomic-commerce-interface type="search" style="max-width: 1000px; margin: auto">
           <atomic-search-layout>
             <atomic-layout-section section="search">
-              <atomic-commerce-search-box>
+              <atomic-commerce-search-box textarea>
```
olamothe, line 32: Not necessary, the search box should already be a textarea

fpbrault: I think the problem is that we are still using the css of `atomic-search-box`

```css
@import '../../search/atomic-search-box/atomic-search-box.pcss';
```

if we remove the textarea attribute from the selector we should be good. Alternatively we could add textarea on the host attribute in `atomic-commerce-search-box` but I don't think thats the right solution.
```css
:host([textarea]) {
  @apply relative z-10;
}
```

louis-bompart: Partially, other CSS rules apply only to textarea.

olamothe: https://github.com/coveo/ui-kit/pull/3936/files#diff-91eb577ee22326dd19a78cfc762da30b9eb5b8819bbdaac2fec2adc49bc55b1bR2-R5 

+ the whole PR that adds sections

PR #3924, opened 2024-05-10
feat: improve state lock-out with reflection

louis-bompart:
An improvement/iteration on #3890:
- Hide the state symbol from all methods relying on `[[OwnPropertyKeys]]` [^1]
- Do some error handling to help users if they're facing an error due to a double-load of Headless.

Further improvements to consider for 'next-gen' controller:
 - Add safeguards for cross-contaminations on controller instantiation, see [POC](https://louis-bompart.github.io/automatic-eureka/) ([source](https://github.com/louis-bompart/automatic-eureka/tree/main))
 
 [^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/ownKeys

### DIFF
--- a/packages/headless/src/app/state-key.ts
+++ b/packages/headless/src/app/state-key.ts
@@ -7,7 +7,7 @@ export const redactEngine = <TEngine extends CoreEngineNext>(
 ): TEngine =>
   new Proxy(engine, {
     ownKeys(target) {
-      return Reflect.ownKeys(target).filter((key) => typeof key !== 'symbol');
+      return Reflect.ownKeys(target).filter((key) => key !== stateKey);
     },
     get(target, prop, receiver) {
       if (
--- a/packages/headless/src/app/state-key.ts
+++ b/packages/headless/src/app/state-key.ts
@@ -1,1 +1,24 @@
+import type {CoreEngineNext} from './engine';
+
 export const stateKey = Symbol('state');
+
+export const redactEngine = <TEngine extends CoreEngineNext>(
+  engine: TEngine
+): TEngine =>
+  new Proxy(engine, {
+    ownKeys(target) {
+      return Reflect.ownKeys(target).filter((key) => typeof key !== 'symbol');
+    },
+    get(target, prop, receiver) {
+      if (
+        typeof prop === 'symbol' &&
+        prop.description === 'state' &&
+        prop !== stateKey
+      ) {
+        engine.logger.warn(
+          "You might be loading Headless twice. Please check your setup.\nIf you are trying to access the inner state... Don't"
+        );
+      }
+      return Reflect.get(target, prop, receiver);
+    },
+  });
--- a/packages/headless/src/app/state-key.ts
+++ b/packages/headless/src/app/state-key.ts
@@ -1,6 +1,7 @@
 import type {CoreEngineNext} from './engine';
 
-export const stateKey = Symbol('state');
+const stateKeyDescription = 'state';
+export const stateKey = Symbol(stateKeyDescription);
 
 export const redactEngine = <TEngine extends CoreEngineNext>(
   engine: TEngine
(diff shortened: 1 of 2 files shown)


packages/headless/src/app/state-key.ts
```diff
@@ -1 +1,24 @@
+import type {CoreEngineNext} from './engine';
+
 export const stateKey = Symbol('state');
```
louis-bompart, line 3: We could denature the symbol description and put serialized information, such as the stack trace, using `new Error().stack`...
But I prefer to avoid polluting the Symbol description.
In the medium term, I'd prefer to improve the headless-controller base class instead of introducing what's shown in the POC I linked in the description.

packages/headless/src/app/state-key.ts
```diff
@@ -1 +1,24 @@
...
+
+export const redactEngine = <TEngine extends CoreEngineNext>(
```
louis-bompart, line 5: This method could (and probably will) also directly include the prototype chain shenanigans.

After that, when the prototype chain guards are in place on the controllers, we could remove/reformulate the warning to be more of a "How the heck did you reach this? Please contact Coveo Support."

packages/headless/src/app/state-key.ts
fbeaudoincoveo, line 3:
```suggestion
const stateKeyDescription = 'state';
export const stateKey = Symbol(stateKeyDescription);
```

packages/headless/src/app/state-key.ts
```diff
@@ -1 +1,24 @@
...
+  engine: TEngine
+): TEngine =>
+  new Proxy(engine, {
+    ownKeys(target) {
+      return Reflect.ownKeys(target).filter((key) => key !== stateKey);
+    },
+    get(target, prop, receiver) {
+      if (
+        typeof prop === 'symbol' &&
+        prop.description === 'state' &&
```
fbeaudoincoveo, line 15:
```suggestion
        prop.description === stateKeyDescription &&
```

packages/headless/src/app/state-key.ts
```diff
@@ -1 +1,24 @@
...
+        prop !== stateKey
+      ) {
+        engine.logger.warn(
+          "You might be loading Headless twice. Please check your setup.\nIf you are trying to access the inner state... Don't"
```
fbeaudoincoveo, line 19: 😅

olamothe (commented):
It's elegant, but... I'm not particularly enthusiastic about this complexity. Especially since it's not strictly necessary.

Problem is we want to make something private, but it's hard to do correctly in JS, or not a native feature of the language.

I would personally vote for just a simple leading underscore, which is well understood to mean "do not use directly", or "use at your own risk".

Queue up Python: "Were all consenting adults here" 😉

louis-bompart:
> It's elegant, but... I'm not particularly enthusiastic about this complexity. Especially since it's not strictly necessary.
> 
> Problem is we want to make something private, but it's hard to do correctly in JS, or not a native feature of the language.
> 
> I would personally vote for just a simple leading underscore, which is well understood to mean "do not use directly", or "use at your own risk".
> 
> Queue up Python: "Were all consenting adults here" 😉

I agree that this would be unwarranted complexity... if we did not offer commercial support on the solution and could let the "so-called adults" in their quagmires when they come crying for help after using an internal API 😅 
Those assistance requests, even if we answer "nope, you're on your own" -which rarely happens- incur costs to Coveo.
Let's try to :
- Reduce the complexity for Coveo
- Reduce the risks of SNAFUs

If the code were overly complex, I would not have entertained this idea, but here, we have something quite lean and with a cost/rewards that makes it worth it.

However, I might be wrong and overprotective here. Still, assuming the two positions are equally valid (and I think they indeed are), objectively (I think), I would lean towards the one proposed here because it'll be much easier to go from something that is genuinely internal to something "exposed but you should not use."

packages/headless/src/app/state-key.ts
louis-bompart: I want to remove the funny comment in a future iteration based on the POC I mentioned 😂

louis-bompart:
👍 going forward with this.

PR #3925, opened 2024-05-10
docs(atomic): improve shouldHighlight property docs

telmaq:
I updated the documentation regarding the "shouldHighlight" property. Hopefully it's now less confusing

https://coveord.atlassian.net/browse/DOC-13743


jpmarceau:
Not sure how it happened @telmaq , but your changes disappeared 🤔

louis-bompart:
> Not sure how it happened @telmaq , but your changes disappeared 🤔

https://github.com/coveo/ui-kit/pull/3925/commits/1dbe768a3566cdc37ffd81862f8f6a463c9a504c
Not the right files ;)
Need to modify the JSDOC of the components, not the generated files ;)

telmaq:
I wasn't even aware that I was changing the generated files 😮. Thank you, and it should be fixed now

louis-bompart (approved):

packages/atomic/src/components/search/result-template-components/atomic-result-text/atomic-result-text.tsx
```diff
@@ -32,7 +32,7 @@ export class AtomicResultText implements InitializableComponent {
    */
   @Prop({reflect: true}) public field!: string;
   /**
-   * If this is set to true, it will look for the corresponding highlight property and use it if available.
+   * When this is set to `true`, the component attempts to highlight text based on the highlights properties provided by the search API response.
```
jpmarceau, line 35: I think this works a bit better, grammatically:
```suggestion
   * When this is set to `true`, the component attempts to highlight text based on the highlighting properties provided by the search API response.
```

fbeaudoincoveo (approved):

PR #3933, opened 2024-05-13
fix(atomic): update initializeBindings utility function for commerce custom components

olamothe:
Custom components were previously only possible with classic `Bindings` (ie: headless search engine). We need to make it possible with `CommerceBindings` (headless commerce engine)

Modify the exported function to instead expose the fact that it is generic under the hood.
Make it so that it defaults on `Bindings` (search engine use case) for backward compatibility.

https://coveord.atlassian.net/browse/KIT-3156

### DIFF
--- a/packages/atomic/src/utils/initialization-utils.tsx
+++ b/packages/atomic/src/utils/initialization-utils.tsx
@@ -35,9 +35,9 @@ const initializableElements = [
  * @param event Element on which to dispatch the event, which must be the child of a configured "atomic-search-interface" or "atomic-external" element.
  * @returns A promise that resolves on initialization of the parent "atomic-search-interface" or "atomic-external" element, and rejects when it's not the case.
  */
-export function initializeBindings<SpecificBindings extends AnyBindings>(
-  element: Element
-) {
+export function initializeBindings<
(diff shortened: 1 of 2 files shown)


packages/atomic/src/utils/initialization-utils.tsx
```diff
@@ -31,13 +31,13 @@ const initializableElements = [
 ];
 
 /**
- * Retrieves `Bindings` on a configured parent search interface.
- * @param event Element on which to dispatch the event, which must be the child of a configured "atomic-search-interface" or "atomic-external" element.
- * @returns A promise that resolves on initialization of the parent "atomic-search-interface" or "atomic-external" element, and rejects when it's not the case.
+ * Retrieves `Bindings` or `CommerceBindings` on a configured parent interface.
+ * @param event Element on which to dispatch the event, which must be the child of a configured atomic container element.
```
fbeaudoincoveo, line 35:
```suggestion
 * @param event - The element on which to dispatch the event, which must be the child of a configured Atomic container element.
```

louis-bompart (approved):
Some type of inferring would be nice (i.e. we guess the kind of bindings we return based on the element), but this doesn't seem easy. Oh well!

PR #3934, opened 2024-05-13
fix(headless): improved condition that needs to be satisfied before handling a generated answer stream

mmitiche:
[SFINT-5516](https://coveord.atlassian.net/browse/SFINT-5516)

### The issue

In Salesforce, where the Abort Controller is not available, we currently use a a ponyfill to mimic the behaviour of the the Abort Controller and to create an alternative `abortable fetch`.

However, this workaround we created is not able to really abort and the stop the communication completely.
I tested and confirmed that a simple fetch request can be stopped using our ponyfill. However the server sent communication that we initiate with the `fetch-event-source` library to get the generated answer is not completely stopped, the server is still sending messages even after aborting the request.

Therefor, the following corner case is occurring sometimes when triggering a new search query while a generated answer is still being streamed:      

https://github.com/coveo/ui-kit/assets/86681870/17d1b9ba-d738-4ece-842c-29d0a19a21b2

### The solution:

To solve this, we strengthened the condition that needs to be satisfied before taking a generated answer stream in consideration.

Now we first check if `abortController.signal.aborted` is set to `true`, if it is the case we discard the stream even if its request is not completely aborted.  

https://github.com/coveo/ui-kit/assets/86681870/37cc4fc5-fb65-4f88-8a8d-d15d66e55c11

[SFINT-5516]: https://coveord.atlassian.net/browse/SFINT-5516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
The fix makes sense to me, nice catch!

Also I think you meant Polyfill :p 

Also random question, but your PRs always seem to include a lot of random merge commits not sure why that is?

louis-bompart:
I don't think this is the right solution.
Of course, the true solution would be for Salesforce to porperly support standard web APIs, but well, we can't count or act on that 🙄.

So, because we can't rely on our fetch being aborted properly (sigh...), we should ignore it if its aborted. For that, the state handling is too "high" in the code proposed: this should not be handled at the state level, instead this should be handled at the API level.
i.e. response from aborted request should not be considered at all and not trigger actions even.

So I would suggest to add support for AbortController in the `generated-answer-client` instead.
i.e. if abortCtrler.abort() is called, the listeners.

Here's a patch that should do just that:
```patch
diff --git a/packages/headless/src/api/generated-answer/generated-answer-client.ts b/packages/headless/src/api/generated-answer/generated-answer-client.ts
index 6b22f0b6e..9bdccd011 100644
--- a/packages/headless/src/api/generated-answer/generated-answer-client.ts
+++ b/packages/headless/src/api/generated-answer/generated-answer-client.ts
@@ -132,6 +132,9 @@ export class GeneratedAnswerAPIClient {
           }
         },
         onmessage: (event) => {
+          if (abortController?.signal.aborted) {
+            return;
+          }
           const data: GeneratedAnswerStreamEventData = JSON.parse(event.data);
           if (data.finishReason === 'ERROR') {
             timeoutStateManager.remove(timeout!);
@@ -152,6 +155,9 @@ export class GeneratedAnswerAPIClient {
           }
         },
         onerror: (err) => {
+          if (abortController?.signal.aborted) {
+            return;
+          }
           timeoutStateManager.remove(timeout!);
           if (err instanceof FatalError) {
             abortController?.abort();
```

louis-bompart (changes requested):
see suggestion ☝️

mmitiche:
> I don't think this is the right solution. Of course, the proper solution would be for Salesforce to support standard web APIs properly, but we can't count or act on that 🙄.
> 
> So, because we can't rely on our fetch being aborted properly (sigh...), we should ignore it if it's aborted. For that, the state handling is too "high" in the code proposed: this should not be handled at the state level, instead this should be handled at the API level. i.e., a response from an aborted request should not be considered at all and should not even trigger actions.
> 
> So, I would suggest adding support for AbortController in the `generated-answer-client` instead. i.e. if abortCtrler.abort() is called, the listeners.
> 
> Here's a patch that should do just that:
> 
> ```diff
> diff --git a/packages/headless/src/api/generated-answer/generated-answer-client.ts b/packages/headless/src/api/generated-answer/generated-answer-client.ts
> index 6b22f0b6e..9bdccd011 100644
> --- a/packages/headless/src/api/generated-answer/generated-answer-client.ts
> +++ b/packages/headless/src/api/generated-answer/generated-answer-client.ts
> @@ -132,6 +132,9 @@ export class GeneratedAnswerAPIClient {
>            }
>          },
>          onmessage: (event) => {
> +          if (abortController?.signal.aborted) {
> +            return;
> +          }
>            const data: GeneratedAnswerStreamEventData = JSON.parse(event.data);
>            if (data.finishReason === 'ERROR') {
>              timeoutStateManager.remove(timeout!);
> @@ -152,6 +155,9 @@ export class GeneratedAnswerAPIClient {
>            }
>          },
>          onerror: (err) => {
> +          if (abortController?.signal.aborted) {
> +            return;
> +          }
>            timeoutStateManager.remove(timeout!);
>            if (err instanceof FatalError) {
>              abortController?.abort();
> ```

Better! thank you @louis-bompart.

erocheleau (approved):
Seems elegant and simple ty @louis-bompart

louis-bompart (approved):

PR #3940, opened 2024-05-14
feat(commerce): capi 851 commerce context cart should sent product id instead of SKU

BenKandelaars:
# [CAPI-851](https://coveord.atlassian.net/browse/CAPI-851)

The cart controller stored products according to `sku` value. Sku isn't returned from the Commerce API, nor is used by any Coveo systems. Consequently `sku` has been removed from the keys. This poses an additional problem now, in that the events should contain duplicate `productId` when a product is in the cart at different prices. To address this the cart stores products grouped by price, to cater to this use case.

Updates have been made to Commerce base request builder and QuerySuggest to return cart products with deduped productIds.

[CAPI-851]: https://coveord.atlassian.net/browse/CAPI-851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
@@ -1,9 +1,10 @@
 import {createSelector} from '@reduxjs/toolkit';
 import {itemsSelector} from '../../../../features/commerce/context/cart/cart-selector';
 import {CartState} from '../../../../features/commerce/context/cart/cart-state';
+import {CartItem, createCartKey} from './headless-cart';
 
-export function itemSelector(cartState: CartState, sku: string) {
-  return cartState.cart[sku];
+export function itemSelector(cartState: CartState, item: CartItem) {
+  return cartState.cart[createCartKey(item)];
 }
 
 export const totalQuantitySelector = createSelector(itemsSelector, (items) =>
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
@@ -3,10 +3,8 @@ import {itemsSelector} from '../../../../features/commerce/context/cart/cart-sel
 import {CartState} from '../../../../features/commerce/context/cart/cart-state';
 import {CartItem, createCartKey} from './headless-cart';
 
-export function itemSelector(cartState: CartState, cartItem: CartItem) {
-  return Object.entries(cartState.cart[createCartKey(cartItem)])
-    .filter(([_, item]) => item.price === cartItem.price)
-    .map(([_, item]) => item)[0];
+export function itemSelector(cartState: CartState, item: CartItem) {
+  return cartState.cart[createCartKey(item)];
 }
 
 export const totalQuantitySelector = createSelector(itemsSelector, (items) =>
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
@@ -3,8 +3,10 @@ import {itemsSelector} from '../../../../features/commerce/context/cart/cart-sel
 import {CartState} from '../../../../features/commerce/context/cart/cart-state';
 import {CartItem, createCartKey} from './headless-cart';
 
-export function itemSelector(cartState: CartState, item: CartItem) {
-  return cartState.cart[createCartKey(item)];
+export function itemSelector(cartState: CartState, cartItem: CartItem) {
+  return Object.entries(cartState.cart[createCartKey(cartItem)])
+    .filter(([_, item]) => item.price === cartItem.price)
+    .map(([_, item]) => item)[0];
 }
 
 export const totalQuantitySelector = createSelector(itemsSelector, (items) =>
--- a/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
+++ b/packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
@@ -4,7 +4,7 @@ import {stateKey} from '../../../../app/state-key';
 import {
   purchase,
   setItems,
-  updateItem,
+  updateItemQuantity,
 } from '../../../../features/commerce/context/cart/cart-actions';
 import {
   Transaction,
@@ -84,7 +84,7 @@ export interface Cart extends Controller {
    *
    * @param item - The cart item to create, update, or delete.
    */
-  updateItem(item: CartItem, update: CartItem): void;
+  updateItemQuantity(item: CartItem): void;
 
   /**
    * Sets the quantity of each item in the cart to 0, effectively emptying the cart.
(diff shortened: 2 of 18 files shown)


packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -252,6 +246,12 @@ export function buildCart(engine: CommerceEngine, props: CartProps = {}): Cart {
   };
 }
 
+export type CartKey = string;
+
+export function createCartKey(item: CartItem): CartKey {
+  return `${item.productId},${item.name}`;
```
Spuffynism, line 252: I thought we wanted to key by productId, name and price? 🤔

Spuffynism (approved):

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
```diff
@@ -90,7 +84,7 @@ export interface Cart extends Controller {
    *
    * @param item - The cart item to create, update, or delete.
    */
-  updateItem(item: CartItem): void;
+  updateItemQuantity(item: CartItem): void;
```
samisayegh, line 87: Why rename this method?

packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
```diff
@@ -1,9 +1,12 @@
...
 import {itemsSelector} from '../../../../features/commerce/context/cart/cart-selector';
 import {CartState} from '../../../../features/commerce/context/cart/cart-state';
+import {CartItem, createCartKey} from './headless-cart';
 
-export function itemSelector(cartState: CartState, sku: string) {
-  return cartState.cart[sku];
+export function itemSelector(cartState: CartState, cartItem: CartItem) {
+  return Object.entries(cartState.cart[createCartKey(cartItem)])
+    .filter(([_, item]) => item.price === cartItem.price)
+    .map(([_, item]) => item)[0];
```
sallainCoveo, line 9: I think by adding the stringify price to the key, it would simplify the logic here.

```suggestion
  return cartState.cart[createCartKey(cartItem)]
```

packages/headless/src/features/commerce/common/actions.test.ts
```diff
@@ -72,7 +71,9 @@ describe('commerce common actions', () => {
       state.cart.cartItems = [product.productId];
       state.cart.cart = {
         [product.productId]: {
-          ...product,
+          [product.price]: {
+            ...product,
+          },
```
samisayegh, line 76: I think we ended with going for a flat structure with a three part key, rather than nesting price?

samisayegh (commented):
If we keep the same `updateItem` method name and use a three part key, could that help simplify the PR?

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
BenKandelaars: We are using both productId and name as keys. It makes it clearer that its intended for quantity updates. Name updates don't behave as expected after this refactor

packages/headless/src/controllers/commerce/context/cart/headless-cart-selectors.ts
```diff
@@ -1,9 +1,12 @@
 import {createSelector} from '@reduxjs/toolkit';
 import {itemsSelector} from '../../../../features/commerce/context/cart/cart-selector';
 import {CartState} from '../../../../features/commerce/context/cart/cart-state';
+import {CartItem, createCartKey} from './headless-cart';
 
-export function itemSelector(cartState: CartState, sku: string) {
-  return cartState.cart[sku];
+export function itemSelector(cartState: CartState, cartItem: CartItem) {
+  return Object.entries(cartState.cart[createCartKey(cartItem)])
```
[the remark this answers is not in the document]
BenKandelaars: Using Object.entries I can use filter and map

packages/headless/src/features/commerce/common/actions.test.ts
BenKandelaars: I thought otherwise, maybe my jet lag clouded my recall :>. As we cater to aggregating on `productId` for CAPI, it is easier for consumers to use the nested prices instead of splitting the key. Also, this makes it possible to add other helpers like `listPrices`.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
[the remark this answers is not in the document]
BenKandelaars: As we use `name` in the key, updating the name causes a create operation whilst leaving the existing product in the cart. This extends to anything we add in the key, including price. Keeping the key as simple as possible address's this issue.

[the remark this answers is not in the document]
BenKandelaars: The price actually has the same issue. We can therefore only update quantity, without doing a silent create operation.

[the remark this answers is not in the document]
sallainCoveo: Ah ok, I think I have a better understanding. I think it means we have to change the JSDOC here still, especially on line 75-76

> - Setting `quantity` to a positive number updates the item's `name`, `price`, and `quantity` to the
   * specified values.
   
From my understanding, we don't really update the item's name & price, because the combination of all attributes represent a single item.

packages/headless/src/controllers/commerce/context/cart/headless-cart.ts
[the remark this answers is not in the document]
BenKandelaars: I've revert the PR back to using a single key of productId, name and price.

packages/headless/src/features/commerce/common/actions.test.ts
BenKandelaars: Reverted to flat structure

packages/headless/src/features/commerce/context/cart/cart-slice.ts
```diff
@@ -13,29 +14,32 @@ export const cartReducer = createReducer(
...
         }, getCartInitialState());
 
         setItemsInState(state, cartItems, cart);
       })
-      .addCase(updateItem, (state, {payload}) => {
-        if (!(payload.sku in state.cart)) {
+      .addCase(updateItemQuantity, (state, {payload}) => {
+        const key = createCartKey(payload);
+        if (!(key in state.cart)) {
+          // deleteProductFromCart(payload.item, state);
```
sallainCoveo, line 32:
```suggestion
```

sallainCoveo (approved):
Nice! :100:

Spuffynism (approved):

samisayegh (approved):
Looks good ✅

louis-bompart (approved):

louis-bompart:
@BenKandelaars I think. my suggestion doesn't work/break stuff 😅

PR #3941, opened 2024-05-14
feat(Commerce Atomic): add product rating component

fpbrault:
This PR adds the atomic-product-rating template component. By default, it will use the ec_rating field. It also can display a text value next to the rating. This can be used to display the numerical value of the rating, or the number of ratings.

image

https://coveord.atlassian.net/browse/KIT-3053


louis-bompart (approved):

packages/atomic/src/components/commerce/product-template-components/atomic-product-rating/atomic-product-rating.tsx
```diff
@@ -0,0 +1,133 @@
+import {Product, ProductTemplatesHelpers} from '@coveo/headless/commerce';
```
olamothe, line 1: This PR would have benefitted from a small overhaul/refactor of `atomic-result-rating` component, to extract some common pattern into functional components or shared utility function (similar to what was done past few weeks across many different files/components).

I don't think it would be a huge time sink, and would prevent adding more tech debt.

PR #3942, opened 2024-05-14
feat(Commerce Atomic): add mappable alt text for atomic-product-image

fpbrault:
This PR adds the possibility of setting a field as a source for the alt text of atomic-product-image images.

- If the value of the field is a string, it will be used for all images.
- If the value of the field is an array of strings, each image will use the alt text corresponding to its index. Any extra image will use the default alt text.
- If no field is provided, if the field value is empty or not a string, the default alt text ("Image {index} out of {totalImages} for {productName}") will be used for all images.

https://coveord.atlassian.net/browse/KIT-3180


packages/atomic/src/components/commerce/product-template-components/atomic-product-image/atomic-product-image.tsx
```diff
@@ -0,0 +1,238 @@
+import {Product, ProductTemplatesHelpers} from '@coveo/headless/commerce';
```
olamothe, line 1: The PR needs an update from master/resolve merge conflict before we can review it.

fpbrault: should be good now

packages/atomic/src/components/commerce/product-template-components/atomic-product-image/atomic-product-image.tsx
```diff
@@ -137,6 +164,21 @@ export class AtomicProductImage implements InitializableComponent {
...
 
+  private get imagesAlt() {
+    if (this.imagesAltField) {
+      const value = ProductTemplatesHelpers.getProductProperty(
+        this.product,
+        this.imagesAltField
+      );
+
+      if (Array.isArray(value)) {
+        return value.map((v) => `${v}`.trim());
```
fbeaudoincoveo, line 175: Why are we trimming the values only when it's an array?

fbeaudoincoveo (approved):

packages/atomic/src/components/commerce/product-template-components/atomic-product-image/atomic-product-image.tsx
fpbrault: Missed that, thanks!

PR #3949, opened 2024-05-15
chore(quantic):  fix daily quantic package creation task

mmitiche:
## [SFINT-5513](https://coveord.atlassian.net/browse/SFINT-5513)

The Quantic Package Creation task has been failing repeatedly lately: https://github.com/coveo/ui-kit/actions/workflows/create-quantic-package.yml

The issue started appearing after starting to use the `Setup` action in this workflow. The `Setup` actions uses the NX cache in order to gain time and avoid installing and building everything every time. 

However, I noticed that we are caching only the `node_modules`. and for Quantic to be built correctly it needs the static resources to be created out of the node_modules.

To fix that, I added a step to the `setup` action in order to also cache the `static resources` folder.

Another solution could just be running `npx nx run quantic:build` in the `Create Quantic Package` workflow.
Let me know which solution you prefer or if you have any better ideas.

[SFINT-5513]: https://coveord.atlassian.net/browse/SFINT-5513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/setup/action.yml
```diff
@@ -29,6 +29,10 @@ runs:
           **/node_modules/
           !node_modules/.cache/nx/
         key: npm-${{ hashFiles('package-lock.json') }}-${{ hashFiles('utils/atomic-storybook/package-lock.json') }}
+    - uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4
+      with:
+        path: packages/quantic/force-app/main/default/staticresources
+        key: quantic-${{ hashFiles('package-lock.json') }}-${{ hashFiles('utils/atomic-storybook/package-lock.json') }}
```
SimonMilord, line 35: The setup action seems to be used in other workflows as well like `e2e-quantic.yml`, `prbot.yml`, `release.yml`, etc.

Do you think the current solution could impact those as well by adding the static resources? For this reason I might be tempted to say maybe we should favor running `npx nx run quantic:build` in the failing task instead so we know it only affects one thing. 

Curious to hear what you think, do you think adding a step to build quantic there would make the task that much slower?

mmitiche: This step I'm adding only make use of the NX cache to store the static resources to avoid building them every time. which means that the impact the this step could have is only positive and will win us time.

I don't see any negative impact of using what's already in the cache.

louis-bompart (approved):

.github/actions/setup/action.yml
erocheleau, line 35: Why this key exactly? does it matter?

erocheleau (approved):
I don't know what the key is for, I'm going to assume to generate a unique hash based on the content of the package-lock?

So it doesn't really matter, and package-lock seems like a good way to ensure uniqueness. 👍

SimonMilord (approved):

.github/actions/setup/action.yml
mmitiche: not really, it's just an identifier of that cache, like documented here: https://github.com/actions/cache?tab=readme-ov-file#inputs

From the documentation, using the hashFiles function allows us to create a new cache when dependencies change which is exactly what we want.

PR #3951, opened 2024-05-15
feat(commerce): expose url parameter manager as sub-controller

Spuffynism:
[CAPI-907]

[CAPI-907]: https://coveord.atlassian.net/browse/CAPI-907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
+++ b/packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
@@ -5,12 +5,13 @@ import {
   Unsubscribe,
   UrlManager,
   buildSearch,
-  buildSearchUrlManager,
   getOrganizationEndpoints as getOrganizationEndpointsHeadless,
   updateQuery,
   CommerceEngine,
   CommerceEngineConfiguration,
   buildCommerceEngine,
+  buildProductListing,
+  ProductListing,
 } from '@coveo/headless/commerce';
 import {
   Component,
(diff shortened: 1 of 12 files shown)


packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
```diff
@@ -365,7 +366,11 @@ export class AtomicCommerceInterface
   }
 
   private initSearchStatus() {
-    this.searchStatus = buildSearch(this.engine!);
```
Spuffynism, line 368: @louis-bompart I'm not sure this change is right 🤔 Would love a second pair of eyes on it.

My understanding is that the commerce interface is used for search and listing. Its status should therefore reflect either the search or listing one, not only the search one.

olamothe: This change seems 💯 correct.

olamothe (approved):

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
fbeaudoincoveo: I personally think this is fine and the previous behavior was wrong; otherwise I don't see how we're going to be able to support URL state in PLPs

fbeaudoincoveo (approved):

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
Spuffynism: Now, since this interface is used possibly used for recs as well, should we type the urlManager to be conditionally present?

Spuffynism: I'll leave this as-is for now as a few atomic components use an if/else if with product-listing and search. Hence, if we do what I propose we'll need to apply it in a few places. I propose we revisit this later.

PR #3952, opened 2024-05-15
feat(atomic/commerce): add atomic-product-children component

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3056


packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.pcss
```diff
@@ -0,0 +1,7 @@
+@import '../../../../global/global.pcss';
+
+.active-child {
```
olamothe, line 3: Could set border radius and box sizing directly in `tsx` code as taildwind classes : `box-border rounded`

packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.tsx
```diff
@@ -0,0 +1,188 @@
...
+  @Element() hostElement!: HTMLElement;
+
+  @State() private children: (Product | ChildProduct)[] = [];
+  @State() private currentPage = 0;
+
+  public error!: Error;
+  private originalParentProduct: Product | null = null;
+
+  // TODO: there's a totalNumberOfRelatedProducts property on the Product that could be used to do server-side pagination. However I'm not sure how to (or even if we can) fetch additional related products with the commerce API
+  // TODO: remove this prop; we'll want the carousel to automatically kick in at the proper responsive breakpoint instead.
```
olamothe, line 47: Should remove  the prop right now.

packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.tsx
```diff
@@ -0,0 +1,188 @@
...
+    currentParentProductId: string;
+    originalParentProduct: Product;
+  }>;
+
+  constructor() {
+    this.validateProps();
+  }
+
+  public connectedCallback(): void {
+    this.originalParentProduct = this.isProduct(this.product.children[0])
```
olamothe, line 70: If we do the "swapProduct()" as part of Headless, I guess we could arrange for this logic to go away: The swap() could modify `children` so that it becomes a true `parent` (with an array of children)

packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.tsx
```diff
@@ -0,0 +1,188 @@
...
+        nextPage={() => this.nextPage()}
+        previousPage={() => this.previousPage()}
+        numberOfPages={this.numberOfPages}
+      >
+        {this.subsetOfChildren.map((child) => this.renderChild(child))}
+      </Carousel>
+    );
+  }
+
+  // TODO: get rid of most of the inline styles; in most cases we'll use the generic stuff.
```
olamothe, line 152: I would do it right now, or wait for product sections to guide you before merging this then.
As discussed in person, template components need to take all available space given to them, not limit their own size/width/height themselves (as a general rule of thumb)

It's the responsability of `product-sections` to define the space available.

olamothe (approved):

packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
```diff
@@ -148,6 +157,32 @@ export class AtomicCommerceProductList
...
+      event.detail;
+    const currentParentProductIndex = this.productState.products.findIndex(
+      (product) => product.permanentid === currentParentProductId
+    );
+
+    if (currentParentProductIndex === -1) {
+      return;
+    }
+
+    const stateCopy = JSON.parse(JSON.stringify(this.productState));
```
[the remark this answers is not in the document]
fbeaudoincoveo: Good idea!

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
```diff
@@ -166,7 +166,9 @@ export class AtomicCommerceInterface
   }
```
fbeaudoincoveo, line 166: Out of scope, but these exceptions when the atomic-commerce-interface's type is "recommendations" make me think we may want to have a dedicated simplified interface component for this use case after all.

packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  public error!: Error;
+
+  // TODO: We could leverage the totalNumberOfRelatedProducts property on the Product to show the number of available
+  // children (e.g., ...and 3 more). This would require some UX design though.
+
+  /**
+   * The non-localized label to display for the product children section.
+   */
+  @Prop() public label: string = 'available-in';
```
fbeaudoincoveo, line 54: We could omit that prop, and then the user would be responsible for adding the label themselves with an `atomic-commerce-text` component.

However, including the label is consistent with the UX mockup.

fbeaudoincoveo: I made it so that passing an empty string as a label won't render the label part at all, so it's possible for users to get rid of it this way.

packages/atomic/src/components/common/template-system/row-desktop.pcss
```diff
@@ -188,7 +198,8 @@
     grid-template-columns: minmax(0, min-content) auto 1fr auto;
     grid-template-rows: repeat(7, minmax(0, min-content));
 
-    $(section-visual) {
+    $(section-visual),
+    $(section-children) .product-child {
```
fbeaudoincoveo, line 202: Addresses https://coveord.atlassian.net/browse/KIT-3217

packages/headless/src/controllers/commerce/instant-products/headless-instant-products.ts
```diff
@@ -164,6 +185,17 @@ export function buildInstantProducts(
       );
     },
 
+    promoteChildToParent(childPermanentId, parentPermanentId) {
```
olamothe, line 188: Why isn't the signature just `promoteChildToParent(childProduct: Product)` ?

Headless already knows the current "parent" (and thus it's permanent id)

packages/atomic/custom-elements.json
```diff
@@ -0,0 +1,8811 @@
+{
```
olamothe, line 1: ? what's this file

packages/atomic/src/components/common/template-system/cell-desktop.pcss
```diff
@@ -211,6 +212,13 @@
       width: 100%;
     }
 
+    $(section-children) {
```
olamothe, line 215: Most of these new rules added in (cell-desktop/cell-mobile/row-desktop/row-mobile) etc. should not be added in the generic `common/temnplate-system` folder, but instead in the commerce specific `commerce/atomic-product/atomic-product.pcss`.

The reason is that all these new rules are there for `product-child`, which is not a generic / use case agnostic component.

packages/headless/src/controllers/commerce/instant-products/headless-instant-products.ts
fbeaudoincoveo: Mainly for performance reasons.

If we only pass the child permanent ID (or the full child product), we will have to traverse the children of each product in the controller's state in order to locate the child to promote (O(mn)).

Providing the parent ID in the call allows us to directly traverse the children of the parent product (O(n)).

Also, there wouldn't be an advantage to passing the full product: we only need the permanent id to locate it.

packages/atomic/custom-elements.json
fbeaudoincoveo: I have no clue, seems like it was automatically generated when I added .nx to gitignore:

eda6232

I'll delete it.

packages/headless/src/controllers/commerce/instant-products/headless-instant-products.ts
olamothe: > Also, there wouldn't be an advantage to passing the full product: we only need the permanent id to locate it.

The advantage would be for the implementer: Passing a `Product` interface is more prescriptive/restrictive (in a good way) than "any string will suffice as far as the compiler is concerned, but please use the one named permanent ID which is set as a property on a product". 

Even if internally we end up only needing permanentID to do the job.

> If we only pass the child permanent ID (or the full child product), we will have to traverse the children of each product in the controller's state in order to locate the child to promote (O(mn)).

It's again mostly for implementer POV.  It's more foolproof/harder to do a mistake if the onus is not on the implementer to pass 2 values "correctly linked to one another" (ie: the parent-child relationship is correctly respected by the implementer) VS having Headless handle it internally.

For atomic-product-children, the way the code is written, we have a solid relationship between the parent product and it's children (it's all colocated in the same file), so the advantage are minimal. 

But I can easily see a world where an implementer would build their own version of `atomic-product-children` with no reference to the parent product, or have each individual children product tile only have a reference to an individual child product.

And so you can imagine that the complexity would have to be shifted to the implementer.

For perf: We can add a reference to the parent product on each child product (in Headless reducer) to have it be a simple lookup.

Spuffynism (approved):

packages/headless/src/controllers/commerce/instant-products/headless-instant-products.ts
fbeaudoincoveo: Will address in https://coveord.atlassian.net/browse/KIT-3221

PR #3960, opened 2024-05-16
fix(atomic,headless): fix placeholder display issue for atomic-product-list

olamothe:
https://coveord.atlassian.net/browse/KIT-3178

### DIFF
--- a/packages/headless/src/features/commerce/pagination/pagination-slice.ts
+++ b/packages/headless/src/features/commerce/pagination/pagination-slice.ts
@@ -114,5 +114,5 @@ function getEffectiveSlice(
 }
 
 function handlePaginationReset(state: CommercePaginationState) {
-  state.principal = getCommercePaginationInitialSlice();
+  state.principal.page = getCommercePaginationInitialSlice().page;
 }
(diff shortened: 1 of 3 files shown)


packages/headless/src/features/commerce/pagination/pagination-slice.ts
```diff
@@ -114,5 +114,5 @@ function getEffectiveSlice(
 }
 
 function handlePaginationReset(state: CommercePaginationState) {
-  state.principal = getCommercePaginationInitialSlice();
+  state.principal.page = getCommercePaginationInitialSlice().page;
```
olamothe, line 117: About this change: This was causing issues in atomic, when we need to check the "what's the number of products as reported by pagination" to know "are there products available, if there's no product we should hide this section".

It seems wrong to reset to original state; Instead I think we should be only resetting the "current page" to 0, not full pagination state.

Spuffynism: Seems similar to what is done in the search pagination reducer :+1: https://github.com/coveo/ui-kit/blob/702a04a919cce2df64e4555df5acdc6e787b0bc5/packages/headless/src/features/pagination/pagination-slice.ts#L163

Spuffynism (approved):

fpbrault (approved):

PR #3963, opened 2024-05-17
fix(SFINT-5536): Standalone searchbox issue for smaller screens

SimonMilord:
[SFINT-5536](https://coveord.atlassian.net/browse/SFINT-5536)

## ISSUE:
- On smaller screens (smaller than 768px), the standalone searchbox collapses and its display is weird. I figured out the weird display was due to the `position: absolute` that we need on the searchbox floating container in order to have the searchbox expand without pushing the page content down the page.

ie: 
image

## SOLUTION:
- I added a `position: relative` to the parent of the floating container (searchbox__container-wrapper). This allows the searchbox to be properly displayed while still supporting the expandable searchbox to cover the page content instead of pushing the content down.
- I also added a `min-width: 300px` to make sure that even on small screens the searchbox isnt too small/collapsed. I set it to 300px so that it matches the width of other searchboxes on small screens such as Community Searchbox and also so that it still provides sufficient space to write queries.
- This solution works and does not impact the searchbox when on larger screens and when the searchbox is not expandable. (default searchbox)

## DEMO (expandable searchbox):

https://github.com/coveo/ui-kit/assets/73316533/e10da323-a697-4553-906c-17d4ca64c759

## DEMO (default searchbox):

https://github.com/coveo/ui-kit/assets/73316533/9edf9604-4933-4647-b582-3879a66b9969

[SFINT-5536]: https://coveord.atlassian.net/browse/SFINT-5536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau (approved):
Seems good

mmitiche (approved):

PR #3964, opened 2024-05-17
fix(Commerce Atomic): return responseId from Engine rather than state

y-lakhdar:
Changes
* Added a new selector: `responseIdSelectorFromEngine`.
* The selector accepts a `CommerceEngine` instance (instead of the state) and retrieves the `responseId` property from the internal engine state.

Eventually, we will want to change all selector to accept an engine since the state is now considered internal

https://coveord.atlassian.net/browse/KIT-3197

### DIFF
--- a/packages/atomic/src/components/commerce/atomic-commerce-interface/store.ts
+++ b/packages/atomic/src/components/commerce/atomic-commerce-interface/store.ts
@@ -1,9 +1,9 @@
 import {
   CommerceEngine,
+  Selectors,
   NumericFacetValue,
   DateFacetValue,
   SortCriterion,
-  responseIdSelectorFromEngine,
 } from '@coveo/headless/commerce';
 import {DEFAULT_MOBILE_BREAKPOINT} from '../../../utils/replace-breakpoint';
 import {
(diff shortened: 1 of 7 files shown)


packages/atomic/src/components/commerce/atomic-commerce-interface/store.ts
```diff
@@ -78,5 +80,9 @@ export function createAtomicCommerceStore(): AtomicCommerceStore {
         makeDesktopQuery(commonStore.state.mobileBreakpoint)
       ).matches;
     },
+
+    getUniqueIDFromEngine(engine: CommerceEngine): string {
+      return responseIdSelectorFromEngine(engine);
```
Spuffynism, line 85: The responseId can come from either the search, product listing or recommendations state. Why are we only using the product listing's state here?

packages/headless/src/commerce.index.ts
```diff
@@ -33,6 +33,9 @@ export * from './features/commerce/query-suggest/query-suggest-actions-loader';
 export * from './features/configuration/configuration-actions-loader';
 export * from './features/commerce/query/query-actions-loader';
 
+// Selectors
+export {responseIdSelectorFromEngine} from './features/commerce/product-listing/product-listing-selectors';
```
olamothe, line 37: We should discuss how to organize those selector exports.

Just exporting them all at root like this will balloon up to hundreds of exported symbols very quickly, which is not good for implementers discoverability when your are in an IDE and see a huge list of available functions from `@coveo/headless/commerce`.

How about namespacing those export for now ?

packages/atomic/src/components/commerce/atomic-commerce-interface/store.ts
olamothe: But... this is the atomic-store for `atomic-commerce-interface`, which handle both search and listing. 

I don't understand why we must not differentiate between the two types when we need to get a `uniqueIDFromEngine` 🤔

olamothe (approved):

Spuffynism (approved):

PR #3965, opened 2024-05-17
feat(IPX): add atomic-ipx-recs-list

dguerinCoveo:
For IPX, we need to add a pageView event in the actions history when we click on the grid containing an result list item. 

To do that, I copied the `atomic-recs-list` component and made a `atomic-ipx-recs-list` component. Its exactly the same except that it modifies the `onSelect` method of the built `interactiveResult`. 

Another way to go would be to modify the origin atomic-recs-list to accept custom `buildRecsInteractiveResult` methods but its not the way we usually do for IPX components (as usually we dont modify origin atomic components to fit the needs of the experimental IPX components)

Tested locally ✅


packages/atomic/src/components/recommendations/atomic-recs-result-template/atomic-recs-result-template.tsx
```diff
@@ -72,7 +72,7 @@ export class AtomicRecsResultTemplate {
       setError: (err) => {
         this.error = err;
       },
-      validParents: ['atomic-recs-list'],
+      validParents: ['atomic-recs-list', 'atomic-ipx-recs-list'],
```
dguerinCoveo, line 75: This is the only change in a common atomic component that I needed to do.

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -369,35 +369,70 @@
...
+                              stroke="currentColor"
+                              stroke-linecap="round"
+                              stroke-linejoin="round"
+                            />
+                          </svg>
+                        </span>
+                      </p>
+                      <a slot="attributes" target="_blank"></a>
+                    </atomic-ipx-result-link>
+                  </atomic-result-section-bottom-metadata>
```
dguerinCoveo, line 432: Took the opportunity to freshen the look of the IPX recs result template

PR #3966, opened 2024-05-17
feat(IPX): add atomic-ipx-recs-list

dguerinCoveo:
For IPX, we need to add a pageView event in the actions history when we click on the grid containing an result list item.  

To do that, I copied the atomic-recs-list component and made a atomic-ipx-recs-list component. Its exactly the same except that it modifies the onSelect method of the built interactiveResult.

Another way to go would be to modify the origin atomic-recs-list to accept custom buildRecsInteractiveResult methods but its not the way we usually do for IPX components (as usually we dont modify origin atomic components to fit the needs of the experimental IPX components)

Tested locally ✅

Note: We already have a mechanism to add `pageView` events to `actionsHistory` for search and result items. We were just missing the same mechanism when the click is on the container itself (`result-list-grid-clickable-container`) rather than the item link


packages/atomic/src/components/recommendations/atomic-recs-result-template/atomic-recs-result-template.tsx
```diff
@@ -72,7 +72,7 @@ export class AtomicRecsResultTemplate {
       setError: (err) => {
         this.error = err;
       },
-      validParents: ['atomic-recs-list'],
+      validParents: ['atomic-recs-list', 'atomic-ipx-recs-list'],
```
dguerinCoveo, line 75: This is the only change in a common atomic component that I needed to do.

packages/atomic/src/pages/examples/ipx.html
```diff
@@ -369,35 +369,70 @@
...
+                              stroke="currentColor"
+                              stroke-linecap="round"
+                              stroke-linejoin="round"
+                            />
+                          </svg>
+                        </span>
+                      </p>
+                      <a slot="attributes" target="_blank"></a>
+                    </atomic-ipx-result-link>
+                  </atomic-result-section-bottom-metadata>
```
dguerinCoveo, line 432: Took the opportunity to freshen the look of the IPX recs result template

packages/atomic/src/components/ipx/atomic-ipx-recs-list/atomic-recs-list/atomic-ipx-recs-list.tsx
```diff
@@ -0,0 +1,446 @@
...
+  }
+
+  private getPropsForAtomicRecsResult(recommendation: RecsResult) {
+    const interactiveResult = buildRecsInteractiveResult(this.bindings.engine, {
+      options: {result: recommendation},
+    });
+    const originalSelect = interactiveResult.select;
+    interactiveResult.select = () => {
+      this.onSelect(recommendation, originalSelect);
+    };
```
dguerinCoveo, line 346: This is the main change. We override the interactiveResult.onSelect method to call our custom onSelect method on line 325

mikegron (commented):
Ouin c'est quand meme plate la quantité de copy paste stuff 😕 Il y a pas moyen "d'hériter" de l'autre component or something?

dguerinCoveo:
Je refer a nos DX overlords @louis-bompart @olamothe

olamothe (approved):
In general, we'll need to think about refactoring a lot of stuff for IPX.

The way it is currently done is * not * long term viable, where we sometimes piggy `search` components, or either we end up with a lot of copy pasting.

The correct approach will be to extract "pure" (no headless package dependency) common functional components, and do code reuse that way.

This is the approach we've taken for insight panel, and for everything we are doing for atomic-commerce currently.

Unfortunately, pretty much the whole team is focusing on delivering commerce related features for the short term, so we don't have any bandwidth...

We could help you with this sometimes in the mid term...

dguerinCoveo:
Yeah, we have it in our backburner somewhere to eventually make a real, full fledge library of the IPX components too. We're not there yet though.

pathiery (approved):

PR #3974, opened 2024-05-21
fix(headless): send referrer as null when originLevel3 is default

samisayegh:
**Context**
The server event schema was recently tightened to require document referrer to be a url if specified. The schema definition is available [here](https://github.com/coveo-platform/analytics_schema/blob/master/schemas/common/types.json#L289).

Headless' behavior is to forward the originLevel3 value, which by default is set to the string `default`.
This causes all server events sent by headless to be marked as invalid.

**Approach**
When the `analyticsMode` is `next` and originLevel3 is `default`, headless will send `null` instead of `default`.

### DIFF
--- a/packages/headless/src/api/search/search-api-params.ts
+++ b/packages/headless/src/api/search/search-api-params.ts
@@ -119,7 +119,7 @@ export interface AnalyticsParam {
     deviceId?: string;
     pageId?: string;
     clientTimestamp: string;
-    documentReferrer: string;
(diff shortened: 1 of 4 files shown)


sallainCoveo (approved):

olamothe (approved):
There's some UT failing in `analytics-migration.test.ts` which will need to be adjusted.

They are tripping up because `documentReferrer` is not behaving the same as before.

PR #3975, opened 2024-05-21
fix(atomic-react): output only commerce components in commerce react folder

olamothe:
Previously, we would output commerce components in both `@coveo/atomic-react` and `@coveo/atomic-react/commerce`.

We would also output non commerce specific component in `@coveo/atomic-react/commerce`. 

This PR add a new output target in `stencil.config.ts` which, use `excludeComponents` + generate list of use case specific components based on file names.

Also update `CommerceSearchPage` in atomic-react/sample, to leverage new components and making sure they all work properly.

https://coveord.atlassian.net/browse/KIT-3202


packages/atomic/stencil.config.ts
```diff
@@ -20,6 +20,41 @@ import {generateAngularModuleDefinition as angularModule} from './stencil-plugin
...
+  {
+    recursive: true,
+  }
+)
+  .map((fileName) => /(atomic-[a-z-]+)\.tsx$/.exec(fileName.toString()))
+  .filter((m) => m !== null)
+  .flatMap((m) => m![1]);
+
+console.log(commerceComponents, searchComponents);*/
+
```
fbeaudoincoveo, line 57: remove?

fbeaudoincoveo (approved):

louis-bompart (approved):

packages/atomic/stencil.config.ts
louis-bompart:
```suggestion
```

PR #3978, opened 2024-05-21
chore: update nx & co

louis-bompart:
- Update NX
- Remove plugin of uncertain usefulness
  - Checked dists https://github.com/coveo/nx-update-atomic-artifact-comparaison/commit/1438646dbf1a975764ceff74f20be4ce7b0ee07f#diff-5979a120314b63bc56751f08c61ceb01c898b4e6f198ebffe0fff1633958ee46


olamothe (commented):
Without being a nx expert, I am lacking context into what has changed with the new version that makes it so we don't need those `nx-plugin` files anymore.

Can you provide some explanation/details about the changes ?

louis-bompart:
I think the plugin might have outlive its usefulness, or at the very least I'm not able to confidently migrate/understand what it does.

So I think removing it a

> Without being a nx expert, I am lacking context into what has changed with the new version that makes it so we don't need those `nx-plugin` files anymore.
> 
> Can you provide some explanation/details about the changes ?

I think the plugin might have outlive its usefulness, or at the very least I'm not able to confidently migrate/understand what it does.

So I think removing it is the right move.
I compared the artifacts, lgtm:
https://github.com/coveo/nx-update-atomic-artifact-comparaison/commit/1438646dbf1a975764ceff74f20be4ce7b0ee07f#diff-5979a120314b63bc56751f08c61ceb01c898b4e6f198ebffe0fff1633958ee46

PR #3983, opened 2024-05-23
fix(atomic): add label on atomic-pager radiogroup

fpbrault:
This PR adds a label to the role="radiogroup" element inside PagerPageButtons. This label is required on elements with this role.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role#associated_wai-aria_roles_states_and_properties:~:text=aria%2Dlabelledby%20/,aria%2Dlabel.

https://coveord.atlassian.net/browse/KIT-3211

### DIFF
--- a/packages/atomic/src/components/commerce/atomic-commerce-pager/atomic-commerce-pager.tsx
+++ b/packages/atomic/src/components/commerce/atomic-commerce-pager/atomic-commerce-pager.tsx
@@ -112,17 +112,17 @@ export class AtomicCommercePager
         hasResults={this.pagerState.totalPages > 1}
         isAppLoaded={this.bindings.store.isAppLoaded()}
       >
-        <PagerNavigation label={this.bindings.i18n.t('pagination')}>
+        <PagerNavigation i18n={this.bindings.i18n}>
           <PagerPreviousButton
             icon={this.previousButtonIcon}
             disabled={this.pagerState.page === 0}
-            ariaLabel={this.bindings.i18n.t('previous')}
+            i18n={this.bindings.i18n}
             onClick={() => {
               this.pager.previousPage();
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/common/pager/pager-buttons.tsx
```diff
@@ -73,7 +73,12 @@ export const PagerPageButton: FunctionalComponent<PagerPageButtonProps> = (
 
 export const PagerPageButtons: FunctionalComponent = (_, children) => {
   return (
-    <div part="page-buttons" role="radiogroup" class="contents">
+    <div
+      part="page-buttons"
+      role="radiogroup"
+      aria-label="Pagination"
```
olamothe, line 79: Should use `i18n.t('pagination')`

olamothe (approved):

packages/atomic/src/components/commerce/atomic-commerce-pager/atomic-commerce-pager.tsx
```diff
@@ -122,7 +122,7 @@ export class AtomicCommercePager
               this.focusOnFirstResultAndScrollToTop();
             }}
           />
-          <PagerPageButtons>
+          <PagerPageButtons label={this.bindings.i18n.t('pagination')}>
```
olamothe, line 125: Nitpick: 
Instead of doing it like this, modify `PagerPageButtonProps` to receive `i18n`, and do `i18n.t('pagination')` directly in the functional components.

Rationale is that it's easier to update the label everywhere at once if necessary, and it's not the type of value that needs to be unique per use case (ie: all use case should use the exact same label description)

olamothe (approved):

dmgauthier (approved):

PR #4002, opened 2024-05-24
feat(atomic/commerce): add dedicated commerce recommendation interface

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3214


packages/atomic/src/components/commerce/atomic-commerce-recommendation-interface/atomic-commerce-recommendation-interface.tsx
```diff
@@ -0,0 +1,270 @@
...
+      return;
+    }
+
+    scrollContainerElement.scrollIntoView({behavior: 'smooth'});
+  }
+
+  /**
+   * Initializes the connection with an already preconfigured [headless commerce engine](https://docs.coveo.com/en/headless/latest/reference/commerce/).
+   */
+  @Method() public initializeWithEngine(engine: CommerceEngine) {
```
fbeaudoincoveo, line 197: I only included the initializeWithEngine method on the atomic-commerce-recommendation-interface because otherwise we could end up with multiple engines / states, which I believe we should absolutely avoid.

It think it's OK to have the regular initialize method on the atomic-commerce-interface, though, because there should only be one PLP or search interface displayed at a time. However, there could be multiple recommendation interfaces.

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
```diff
@@ -186,10 +191,6 @@ export class AtomicCommerceInterface
 
   @Watch('analytics')
   public toggleAnalytics() {
-    if (!this.commonInterfaceHelper.engineIsCreated(this.engine)) {
```
fbeaudoincoveo, line 189: Already checked in commonInterfaceHelper.onAnalyticsChange()

PR #4014, opened 2024-05-27
feat(knowledgeAPI): api slice

dmgauthier:
## Overview
This update introduces a client to manage the new AnswerApi stream, serving RGA model answers. The client leverages [Redux Toolkit Query](https://redux-toolkit.js.org/rtk-query/overview).

## Tech
`baseKnowledgeAPISlice.ts`
This file contains the baseQuery, which can be used for simple API implementations, and the main slice definition, which can be used to create the Redux entry in the state.

`answerApiSlice.ts`
This file defines the endpoint that will manage the streaming, completing the Redux API slice.

While the `dynamicBaseQuery` pattern sets up a basic structure for adding new knowledge endpoints in the future, the code-splitting pattern is useful for maintaining a separation of concerns between generic code and endpoint-specific code.

## The Feature
This feature creates an entry in the state, forming the foundation of the new answer controller for the answer generation stream. It manages everything that is currently manually dispatched regarding API-level activity, such as `isStreaming`, `isLoading`, `answer`, and `citations`.

The RTK/Query library already handles the multi-query pattern, eliminating the need for complex ID management through a custom random ID implementation.

image

**Edit** Now works with the new AnswerApi endpoint
image

### DIFF
--- a/packages/headless/src/api/knowledge/answer-slice.ts
+++ b/packages/headless/src/api/knowledge/answer-slice.ts
@@ -1,0 +1,45 @@
+import {
+  BaseQueryFn,
+  createApi,
+  FetchArgs,
+  fetchBaseQuery,
+  FetchBaseQueryError,
+  retry,
+} from '@reduxjs/toolkit/query';
+import {ConfigurationSection} from '../../state/state-sections';
+
+type StateNeededByAnswerSlice = ConfigurationSection;
+
+/**
+ * `dynamicBaseQuery` is passed to the baseQuery of the createApi,
+ * but note that the baseQuery will not be used if a queryFn is provided in the createApi endpoint
+ */
+const dynamicBaseQuery: BaseQueryFn<
+  string | FetchArgs,
+  unknown,
+  FetchBaseQueryError
+> = async (args, api, extraOptions) => {
+  const state = api.getState() as StateNeededByAnswerSlice;
+  const {accessToken, organizationId, platformUrl} = state.configuration;
+  const updatedArgs = {
+    ...(args as FetchArgs),
+    headers: {
+      ...((args as FetchArgs)?.headers || {}),
+      Authorization: `Bearer ${accessToken}`,
+    },
+  };
+  try {
+    const data = fetchBaseQuery({
+      baseUrl: `${platformUrl}/rest/organizations/${organizationId}`,
+    })(updatedArgs, api, extraOptions);
+    return {data};
+  } catch (error) {
+    return {error: error as FetchBaseQueryError};
+  }
+};
+
+export const answerSlice = createApi({
+  reducerPath: 'answer',
+  baseQuery: retry(dynamicBaseQuery, {maxRetries: 3}),
+  endpoints: () => ({}),
+});
--- a/packages/headless/src/api/knowledge/stream-answer-api.ts
+++ b/packages/headless/src/api/knowledge/stream-answer-api.ts
@@ -1,0 +1,287 @@
+import {
+  ArrayValue,
+  BooleanValue,
+  RecordValue,
+  Schema,
+  StringValue,
+} from '@coveo/bueno';
+import {
+  EventSourceMessage,
+  fetchEventSource,
+} from '@microsoft/fetch-event-source';
+import {
+  GeneratedAnswerStyle,
+  GeneratedContentFormat,
+} from '../../features/generated-answer/generated-response-format';
+import {selectPipeline} from '../../features/pipeline/select-pipeline';
+import {selectQuery} from '../../features/query/query-selectors';
+import {QueryState} from '../../features/query/query-state';
+import {selectSearchHub} from '../../features/search-hub/search-hub-selectors';
+import {
+  ConfigurationSection,
+  DebugSection,
+  GeneratedAnswerSection,
+  SearchSection,
+} from '../../state/state-sections';
+import {GeneratedAnswerCitation} from '../generated-answer/generated-answer-event-payload';
+import {SearchRequest} from '../search/search/search-request';
+import {answerSlice} from './answer-slice';
+
+type StateNeededByAnswerAPI = ConfigurationSection &
+  GeneratedAnswerSection &
+  SearchSection &
+  DebugSection & {answer: ReturnType<typeof answerApi.reducer>};
+
+interface ErrorPayload {
+  message?: string;
+  code?: number;
+}
+
+class FatalError extends Error {
+  constructor(public payload: ErrorPayload) {
+    super(payload.message);
+  }
+}
+
+interface GeneratedAnswerStream {
+  answerStyle: GeneratedAnswerStyle | undefined;
+  contentFormat: GeneratedContentFormat | undefined;
+  answer: string | undefined;
+  citations: GeneratedAnswerCitation[] | undefined;
+  generated: boolean;
+  isStreaming: boolean;
+  isLoading: boolean;
+}
+interface HeaderMessage {
+  answerStyle: GeneratedAnswerStyle;
+  contentFormat: GeneratedContentFormat;
+}
+
+type PayloadType =
+  | 'genqa.headerMessageType'
+  | 'genqa.messageType'
+  | 'genqa.citationsType'
+  | 'genqa.endOfStreamType';
+
+const headerMessageSchema = new Schema<HeaderMessage>({
+  answerStyle: new StringValue(),
+  contentFormat: new StringValue(),
+});
+
+const messageSchema = new Schema({
+  textDelta: new StringValue(),
+});
+
+const citationsSchema = new Schema({
+  citation: new ArrayValue(
+    new RecordValue({
+      values: {
+        clickUri: new StringValue(),
+        id: new StringValue(),
+        permanentid: new StringValue(),
+        text: new StringValue(),
+        title: new StringValue(),
+        uri: new StringValue(),
+      },
+    })
+  ),
+});
+
+const validateHeaderMessage = (headerMessage: HeaderMessage) => {
+  headerMessageSchema.validate(headerMessage);
+};
+
+const validateMessage = (message: {textDelta: string}) => {
+  messageSchema.validate(message);
+};
+
+const validateCitationsMessage = (citations: {
+  citation: GeneratedAnswerCitation[];
+}) => {
+  citationsSchema.validate(citations);
+};
+
+const validateEndOfStream = (endOfStream: {answerGenerated: boolean}) => {
+  new Schema({
+    answerGenerated: new BooleanValue(),
+  }).validate(endOfStream);
+};
+
+const handleHeaderMessage = (
+  draft: GeneratedAnswerStream,
+  payload: HeaderMessage
+) => {
+  validateHeaderMessage(payload);
+  const {answerStyle, contentFormat} = payload;
+  draft.answerStyle = answerStyle;
+  draft.contentFormat = contentFormat;
+  draft.isStreaming = true;
+  draft.isLoading = false;
+};
+
+const handleMessage = (
+  draft: GeneratedAnswerStream,
+  payload: {textDelta: string}
+) => {
+  validateMessage(payload);
+  if (draft.answer === undefined) {
+    draft.answer = payload.textDelta;
+  } else {
+    draft.answer = draft.answer.concat(payload.textDelta);
+  }
+};
+
+const handleCitations = (
+  draft: GeneratedAnswerStream,
+  payload: {citation: GeneratedAnswerCitation[]}
+) => {
+  validateCitationsMessage(payload);
+  draft.citations = payload.citation;
+};
+
+const handleEndOfStream = (
+  draft: GeneratedAnswerStream,
+  payload: {answerGenerated: boolean}
+) => {
+  validateEndOfStream(payload);
+  draft.generated = payload.answerGenerated;
+  draft.isStreaming = false;
+};
+
+const updateCacheWithEvent = (
+  event: EventSourceMessage,
+  draft: GeneratedAnswerStream
+) => {
+  const message: {payloadType: PayloadType; payload: string} = JSON.parse(
+    event.data
+  );
+  const parsedPayload = JSON.parse(message.payload);
+  switch (message.payloadType) {
+    case 'genqa.headerMessageType':
+      handleHeaderMessage(draft, parsedPayload);
+      break;
+    case 'genqa.messageType':
+      handleMessage(draft, parsedPayload);
+      break;
+    case 'genqa.citationsType':
+      handleCitations(draft, parsedPayload);
+      break;
+    case 'genqa.endOfStreamType':
+      handleEndOfStream(draft, parsedPayload);
+      break;
+  }
+};
+
+const onOpenStream = async (response: Response) => {
+  if (
+    response.ok &&
+    response.headers.get('content-type')?.includes('text/event-stream')
+  ) {
+    return;
+  }
+
+  const isClientSideError =
+    response.status >= 400 && response.status < 500 && response.status !== 429;
+
+  if (isClientSideError) {
+    throw new FatalError({
+      message: 'Error opening stream',
+      code: response.status,
+    });
+  } else {
+    throw new Error();
+  }
+};
+
+const onError = (err: Error) => {
+  if (err instanceof FatalError) {
+    throw err;
+  }
+};
+
+export const answerApi = answerSlice.injectEndpoints({
+  overrideExisting: true,
+  endpoints: (builder) => ({
+    getAnswer: builder.query<GeneratedAnswerStream, Partial<SearchRequest>>({
+      queryFn: () => ({
+        data: {
+          answerStyle: undefined,
+          contentFormat: undefined,
+          answer: undefined,
+          citations: undefined,
+          generated: false,
+          isStreaming: true,
+          isLoading: true,
+        },
+      }),
+      async onCacheEntryAdded(
+        args,
+        {getState, cacheDataLoaded, updateCachedData}
+      ) {
+        await cacheDataLoaded;
+        /**
+         * createApi has to be called prior to creating the redux store and is used as part of the store setup sequence.
+         * It cannot use the inferred state used by Redux, thus the casting.
+         * https://redux-toolkit.js.org/rtk-query/usage-with-typescript#typing-dispatch-and-getstate
+         */
+        const {configuration} = getState() as unknown as StateNeededByAnswerAPI;
+        const {platformUrl, organizationId, accessToken, knowledge} =
+          configuration;
+        await fetchEventSource(
+          `${platformUrl}/rest/organizations/${organizationId}/answer/v1/configs/${knowledge.answerConfigurationId}/generate`,
+          {
+            method: 'POST',
+            body: JSON.stringify(args),
+            headers: {
+              Authorization: `Bearer ${accessToken}`,
+              Accept: 'application/json',
+              'Content-Type': 'application/json',
+              'Accept-Encoding': '*',
+            },
+            fetch,
+            onopen: onOpenStream,
+            onmessage: (event) => {
+              updateCachedData((draft) => {
+                updateCacheWithEvent(event, draft);
+              });
+            },
+            onerror: onError,
+          }
+        );
+      },
+    }),
+  }),
+});
+
+export const fetchAnswer = (
+  state: StateNeededByAnswerAPI & {
+    knowledge: ReturnType<typeof answerApi.reducer>;
+    query?: QueryState;
+    searchHub?: string;
+    pipeline?: string;
+  }
+) => {
+  const query = selectQuery(state)?.q;
+  const searchHub = selectSearchHub(state);
+  const pipeline = selectPipeline(state);
+
+  return answerApi.endpoints.getAnswer.initiate({
+    q: query,
+    ...(searchHub?.length && {searchHub}),
+    ...(pipeline?.length && {pipeline}),
+  });
+};
+
+export const selectAnswer = (
+  state: StateNeededByAnswerAPI & {
+    knowledge: ReturnType<typeof answerApi.reducer>;
+    query?: QueryState;
+    searchHub?: string;
+    pipeline?: string;
+  }
+) =>
+  answerApi.endpoints.getAnswer.select({
+    q: selectQuery(state)?.q,
+    ...(selectSearchHub(state)?.length && {searchHub: selectSearchHub(state)}),
+    ...(selectPipeline(state)?.length && {pipeline: selectPipeline(state)}),
+  })(state);
--- a/packages/headless/src/features/configuration/configuration-state.ts
+++ b/packages/headless/src/features/configuration/configuration-state.ts
@@ -51,6 +51,10 @@ export interface ConfigurationState {
    * The global headless engine Usage Analytics API configuration.
    */
   analytics: AnalyticsState;
+  /**
+   * The global headless engine Knowledge configuration.
+   */
+  knowledge: KnowledgeState;
 }
 
 export interface AnalyticsState {
@@ -148,6 +152,10 @@ export interface AnalyticsState {
 export const searchAPIEndpoint = '/rest/search/v2';
 export const analyticsAPIEndpoint = '/rest/ua';
 
+interface KnowledgeState {
+  answerConfigurationId: string;
+}
+
 export const getConfigurationInitialState: () => ConfigurationState = () => ({
   organizationId: '',
   accessToken: '',
@@ -172,5 +180,8 @@ export const getConfigurationInitialState: () => ConfigurationState = () => ({
     trackingId: '',
     analyticsMode: 'legacy',
     source: {},
+  },
+  knowledge: {
+    answerConfigurationId: '',
   },
 });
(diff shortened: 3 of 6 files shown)


packages/headless/src/api/knowledge/baseKnowledgeAPISlice.ts
```diff
@@ -0,0 +1,53 @@
...
+    return {data};
+  } catch (error) {
+    return {error: error as FetchBaseQueryError};
+  }
+};
+
+export const baseKnowledgeAPISlice = createApi({
+  reducerPath: 'knowledgeApi',
+  baseQuery: retry(dynamicBaseQuery, {maxRetries: 3}),
+  endpoints: () => ({
```
olamothe, line 44: How about : 
```suggestion
  endpoints: () => ({})
```

This seems to be what RTK recommend: https://redux-toolkit.js.org/rtk-query/usage/code-splitting

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
```diff
@@ -0,0 +1,126 @@
+import {
```
olamothe, line 1:
```
create-answer-api-slice.ts
```

packages/headless/src/api/knowledge/baseKnowledgeAPISlice.ts
```diff
@@ -0,0 +1,53 @@
+import {
```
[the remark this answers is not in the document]
olamothe: Alternatively, naming could be : 
```
knowledge-api
knowledge-api-answer
knowledge-api-insight
knowledge-api-ipx
```
etc.  Keyword "base" seems superfluous

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
```diff
@@ -0,0 +1,126 @@
...
+  EventSourceMessage,
+  fetchEventSource,
+} from '@microsoft/fetch-event-source';
+import {ConfigurationSection} from '../../state/state-sections';
+import {createAbortController} from '../../utils/abort-controller-polyfill';
+import {baseKnowledgeAPISlice} from './baseKnowledgeAPISlice';
+
+type StateNeededByKnowledgeAPI = ConfigurationSection;
+
+interface ErrorPayload {
```
olamothe, line 11: There are already error types defined in `platform-client.ts` and related: Need to unify those and not create more error interface.

Idem for `FatalError`
I know they come from `generated-answer-client`, but... I would like if we could think about unifying those a bit. There is no reasons to have as many different error class and types.

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
olamothe, line 1: Kebab:
```
create-answer-api-slice
```
(Or renaming following my above advice)

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
```diff
@@ -0,0 +1,126 @@
...
+  code?: number;
+}
+
+class FatalError extends Error {
+  constructor(public payload: ErrorPayload) {
+    super(payload.message);
+  }
+}
+
+const abortController = createAbortController();
```
olamothe, line 22: Having this a global module variable is problematic.

There is an inherent statefulness related to aborting request, which you don't want to handle at a global level.

Also, I'm not sure I understand how the abort controller is used (if at all?) by RTK ? 

Where does this `abortController` gets passed to the `Fetch` request ? I see we reference it `onError`, but for it to do anything, it must be passed to the underlying fetch.

I have a hunch it does not do anything right now.

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
```diff
@@ -0,0 +1,126 @@
...
+let retryCount = 0;
```
olamothe, line 23: Same as above regarding the global nature of this stateful variable.

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const data = {
```
olamothe, line 25: `data` (and `type Data`) is a poor name.

How about `GeneratedAnswerStream` ?

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
```diff
@@ -0,0 +1,126 @@
...
+  answerStyle: '',
+  contentFormat: '',
+  answer: '',
+  citations: [],
+  generated: false,
+  isStreaming: true,
+  isLoading: true,
+};
+
+type Data = typeof data;
```
olamothe, line 35: Do we really need the `variable` data ? Can't we just use an interface ?
```suggestion
interface GeneratedAnswerStream { 
[...]
}
```

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
```diff
@@ -0,0 +1,126 @@
...
+  constructor(public payload: ErrorPayload) {
+    super(payload.message);
+  }
+}
+
+const abortController = createAbortController();
+let retryCount = 0;
+
+const data = {
+  answerStyle: '',
```
olamothe, line 26: Most (all?) of these properties can be strongly typed, and are already defined in the project in `api/generated-answer`.

We should refactor and reuse as much as possible.

olamothe (changes requested):

louis-bompart (commented):
This is the kind of code that would benefit from some form of contract testing 🤔

dmgauthier:
> This is the kind of code that would benefit from some form of contract testing 🤔

You are right. Clients might be the perfect candidates.

packages/headless/src/api/knowledge/baseKnowledgeAPISlice.ts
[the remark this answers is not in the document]
dmgauthier: Thanks, this is exactly what I will do :+1:

packages/headless/src/api/knowledge/baseKnowledgeAPISlice.ts
```diff
@@ -0,0 +1,53 @@
...
+import {ConfigurationSection} from '../../state/state-sections';
+
+type StateNeededByKnowledgeAPI = ConfigurationSection;
+
+/**
+ * `dynamicBaseQuery` is passed to the baseQuery of the createApi,
+ * but note that the baseQuery will not be used if a queryFn is provided in the createApi endpoint
+ */
+const dynamicBaseQuery: BaseQueryFn<
+  string | FetchArgs,
```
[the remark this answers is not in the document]
dmgauthier: Because I became confused down the rabbit hole trying to comply with those types and this lost the common sense. Thanks for pointing that out. It had no use.

packages/headless/src/api/knowledge/baseKnowledgeAPISlice.ts
dmgauthier: Yup. No need to declare the endpoint firsthand if we use the returning object from the injection. Thanks again.

At first I was pulling on the original definition and got a problem with `(Unfortunately, it cannot modify the types for the original definition.)`

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
dmgauthier: Fair enough :pray:

dmgauthier: Looking a this a bit more closely. I am not sure what to do.
First of all, I did not question to much how the error management for the actual stream was made. I rather wanted to keep is as close at its actually handling. Maybe this is the right decision, maybe not, but all in all, this bring me to the following point.

As you certainly noticed, the `ErrorPayload` interface and the `FatalError` class are used to make trigger different behaviors according to the error code coming from the API. There might be a better pattern to respect this behavior, but I cannot see how to keep the current logic without them. And I do not see anything in `platform-client.ts` that I can use to handle a stream. 
I do understand the unification concern. But what would you suggest to avoid declaring those here ?

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
dmgauthier: I removed the polyfilled implementation of the `abortController` since the MVP of the new API flow will not be used in Quantic/Saleforce where it might be required. We'll cross the bridge when we get there.

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
dmgauthier: data IS a poor name. :+1:  thanks for pointing it out

packages/headless/src/api/knowledge/createAnswerAPISlice.ts
dmgauthier: The variable is used as the initial state. following your other global variables comments, I just created an interface and passed the initial state object directly in the `queryFn` since its not used anywhere else

dmgauthier:
@olamothe  I resolved some of the comment you made. Feel free to unresolve any of them or reopen new comments if anything.

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+  citations: GeneratedAnswerCitation[] | undefined;
+  generated: boolean;
+  isStreaming: boolean;
+  isLoading: boolean;
+}
+
+const updateCacheWithEvent = (event: EventSourceMessage, draft: Data) => {
+  const message = JSON.parse(event.data);
+  const payload = JSON.parse(message.payload);
+  if (message.payloadType === 'genqa.headerMessageType') {
```
fbeaudoincoveo, line 50: Given that `message.payloadType` can only be strictly equal to one value, I believe we should return at the end of each "if" clause.

Also, I think we should have stricter typing here. Essentially, `message` and `payload` are inferred as `any` which is not great. We should define the expected types of `event.data` and `message.payload`, and use type guard functions to do the conversion.

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+class FatalError extends Error {
+  constructor(public payload: ErrorPayload) {
+    super(payload.message);
+  }
+}
+
+interface GeneratedAnswerStream {
+  answerStyle: string | undefined;
```
fbeaudoincoveo, line 28: Is there no stricter type than string for this?

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+  let retryCount = 0;
+
+  if (err instanceof FatalError) {
+    throw err;
+  }
+
+  if (++retryCount > 3) {
+    const error = {
+      message: 'Failed to complete stream.',
+      code: 1,
```
fbeaudoincoveo, line 106: Should we send the response.status instead?

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+};
+
+const onError = (err: Error) => {
+  let retryCount = 0;
+
+  if (err instanceof FatalError) {
+    throw err;
+  }
+
+  if (++retryCount > 3) {
```
fbeaudoincoveo, line 103: I'd define the max retry count as a constant

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+          getState() as unknown as StateNeededByKnowledgeAPI;
+
+        await fetchEventSource(
+          `${configuration.platformUrl}/rest/organizations/${configuration.organizationId}/answer/v1/configs/${configuration.knowledge.configurationId}/generate`,
+          {
+            method: 'POST',
+            body: JSON.stringify(arg),
+            headers: {
+              Authorization: `Bearer ${configuration.accessToken}`,
+              Accept: '*/*',
```
fbeaudoincoveo, line 147: I believe this should be 'application/json', as this is what the client expects

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+              'Accept-Encoding': '*',
```
fbeaudoincoveo, line 148: This is equivalent to not specifying the header at all, might as well remove it.

packages/headless/src/api/knowledge/knowledge-api.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * `dynamicBaseQuery` is passed to the baseQuery of the createApi,
+ * but note that the baseQuery will not be used if a queryFn is provided in the createApi endpoint
+ */
+const dynamicBaseQuery: BaseQueryFn<
+  string | FetchArgs,
+  unknown,
+  FetchBaseQueryError
+> = async (args, api, extraOptions) => {
+  const state = api.getState() as StateNeededByKnowledgeAPI;
```
fbeaudoincoveo, line 22: You could also destructure `accessToken`, `platformUrl`, and `organizationId` and simplify your string interpolations:

```typescript
const {accessToken, platformUrl, organizationId} = state.configuration;
```

packages/headless/src/api/knowledge/knowledge-api-answer.ts
dmgauthier: yes :+1:

packages/headless/src/api/knowledge/knowledge-api-answer.ts
dmgauthier: this code is useless here. It will be logged in the `onOpenStream` when the API returns an errorCode.  I will remove it from here

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+  answerStyle: string | undefined;
+  contentFormat: GeneratedContentFormat | undefined;
+  answer: string | undefined;
+  citations: GeneratedAnswerCitation[] | undefined;
+  generated: boolean;
+  isStreaming: boolean;
+  isLoading: boolean;
+}
+
+interface Data {
```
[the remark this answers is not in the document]
dmgauthier: It was a useless left over. Good catch

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+  if (message.payloadType === 'genqa.endOfStreamType') {
+    draft.generated = Boolean(payload.answerGenerated);
+    draft.isStreaming = false;
+  }
+};
+
+const onOpenStream = async (response: Response) => {
+  if (
+    response.ok &&
+    response.headers.get('content-type')?.includes('text/event-stream')
```
pathiery, line 78: don't we wanna store the strings as constants at the top of the file ? We could probably also use (or already have) a lib for getting response headers if that's an option 🤔

packages/headless/src/api/knowledge/knowledge-api-answer.ts
pathiery, line 103: if we set the retryCount on line 97, how can this ever be true ?

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+          isLoading: true,
+        },
+      }),
+      async onCacheEntryAdded(
+        arg,
+        {getState, cacheDataLoaded, updateCachedData}
+      ) {
+        await cacheDataLoaded;
+        /**
+         * createApi has to be called prior to create the redux store and is used as part of the store setup sequence.
```
pathiery, line 133:
```suggestion
         * createApi has to be called prior to creating the redux store and is used as part of the store setup sequence.
```

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+export const answerAPI = knowledgeApi.injectEndpoints({
+  overrideExisting: true,
+  endpoints: (builder) => ({
+    getAnswer: builder.query<GeneratedAnswerStream, Partial<SearchRequest>>({
+      queryFn: () => ({
+        data: {
+          answerStyle: undefined,
+          contentFormat: undefined,
+          answer: undefined,
+          citations: undefined,
```
pathiery, line 121: why define the `undefined` ?

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+        {getState, cacheDataLoaded, updateCachedData}
+      ) {
+        await cacheDataLoaded;
+        /**
+         * createApi has to be called prior to create the redux store and is used as part of the store setup sequence.
+         * It cannot use the inferred state used by Redux, thus the casting.
+         * https://redux-toolkit.js.org/rtk-query/usage-with-typescript#typing-dispatch-and-getstate
+         */
+        const {configuration} =
+          getState() as unknown as StateNeededByKnowledgeAPI;
```
pathiery, line 138: should be answer api

packages/headless/src/api/knowledge/knowledge-api.ts
```diff
@@ -0,0 +1,44 @@
...
+    const data = fetchBaseQuery({
+      baseUrl: `${state.configuration.platformUrl}/rest/organizations/${state.configuration.organizationId}`,
+    })(updatedArgs, api, extraOptions);
+    return {data};
+  } catch (error) {
+    return {error: error as FetchBaseQueryError};
+  }
+};
+
+export const knowledgeApi = createApi({
```
pathiery, line 40: should be answer api

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+
+        await fetchEventSource(
+          `${configuration.platformUrl}/rest/organizations/${configuration.organizationId}/answer/v1/configs/${configuration.knowledge.configurationId}/generate`,
+          {
+            method: 'POST',
+            body: JSON.stringify(arg),
+            headers: {
+              Authorization: `Bearer ${configuration.accessToken}`,
```
pathiery, line 146: how come you're not reusing the definition from the base query ?

packages/headless/src/api/knowledge/knowledge-api-answer.ts
dmgauthier: This is the initial object the query needs. And at first those values are undefined

packages/headless/src/api/knowledge/knowledge-api-answer.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  if (message.payloadType === 'genqa.citationsType') {
+    draft.citations = payload.citations;
+  }
+
+  if (message.payloadType === 'genqa.endOfStreamType') {
+    draft.generated = Boolean(payload.answerGenerated);
+    draft.isStreaming = false;
+  }
```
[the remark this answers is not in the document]
dmgauthier: I always prefered `If` to switchcases, thats why I do not have the reflex to use them. But since we better `return` at each of them, I will use a switch case to `break`

packages/headless/src/api/knowledge/knowledge-api-answer.ts
dmgauthier: I did not see any function already created for it.  
Oli mentionned the loose variable outside a scope (aka at the top of the file) could be problematic.
In this case I do not see why we should extract or complicate more this simple condition.

packages/headless/src/api/knowledge/knowledge-api.ts
dmgauthier: I do not think so. I defeat the purpose of reusing a single slice to handle many apis.
If down the road we have many apis to handle under the knowledge use case, the `answerApi` will only be one of them.

KnowledgeApi is only the name of the React Toolkit Query Wrapper. The real endoint calling the answerApi is declared [here](https://github.com/coveo/ui-kit/pull/4014/files#diff-c08f41924a6308395c6062336daabdfbd4717d924f8439b03d64b6f1eda80388R112)

The container here is used to handle the base query. And the common shenanigans that we could want to make generic. Its pretty barebone right now.

In the future, it can be a possibility to have this slice just called `api` without the need to `knowledge` at all. And use a single instance to control all the coveo APIs used by headless

packages/headless/src/api/knowledge/knowledge-api-answer.ts
dmgauthier: because the base query is not used by a `queryFn` 

I could have omited completely the `baseQuery` implementation here and the actual `answerApi` endpoint implementation would have worked exactly the same.

packages/headless/src/api/knowledge/knowledge-api-answer.ts
dmgauthier: Good question. It will never be true even with just 3.  Since the number of retry is actually limited to 3 directly at the root of the slice.
https://github.com/coveo/ui-kit/pull/4014/files#diff-2a2048d0918dc62fa9b3767828d61914422ff71fc7d8b395cf163c2b0afc1e19R42

I will remove the check here.

pathiery (approved):

packages/headless/src/features/configuration/configuration-state.ts
```diff
@@ -51,6 +51,10 @@ export interface ConfigurationState {
    * The global headless engine Usage Analytics API configuration.
    */
   analytics: AnalyticsState;
+  /**
+   * The global headless engine Knowledge API configuration.
```
pathiery, line 55:
```suggestion
   * The global headless engine Answer API configuration.
```

or 
```suggestion
   * The global headless engine Knowledge configuration.
```

packages/headless/src/api/knowledge/answer-slice.ts
```diff
@@ -0,0 +1,45 @@
...
+  FetchArgs,
+  fetchBaseQuery,
+  FetchBaseQueryError,
+  retry,
+} from '@reduxjs/toolkit/query';
+import {ConfigurationSection} from '../../state/state-sections';
+
+type StateNeededByAnswerSlice = ConfigurationSection;
+
+/**
```
olamothe, line 13: Here, we are not in the realm of "public function that implementers need to know about", so JSDoc not necessary (only used by UI-KIT dev)

Personally find these kind of comments problematic. 

We typically should not need to explain "here's how this function X fit with this other one Y to create end result  Z". They do not evolve well over time, and tend to get out of sync pretty quickly with how the software actually works when you fast forward some months/years.

You can keep it if you prefer. 

But I think given that a contributor must know how RTK query works at a base level (you have to read about it a bit if you are going to be touching these files anyway) what this function does is perfectly understandable without the comment.

packages/headless/src/api/knowledge/answer-slice.ts
```diff
@@ -0,0 +1,45 @@
...
+  const updatedArgs = {
+    ...(args as FetchArgs),
+    headers: {
+      ...((args as FetchArgs)?.headers || {}),
+      Authorization: `Bearer ${accessToken}`,
+    },
+  };
+  try {
+    const data = fetchBaseQuery({
+      baseUrl: `${platformUrl}/rest/organizations/${organizationId}`,
```
olamothe, line 33: A bit unrelated for now, but please do keep in your radar when working on the actual knowledge engine itself: 

We'll want to only support/use organizationEndpoints : https://docs.coveo.com/en/mcc80216/manage-an-organization/organization-endpoints

The code is already there in Headless, it's a matter of configuring it correctly when initializing the engine.

packages/headless/src/api/knowledge/stream-answer-api.ts
```diff
@@ -0,0 +1,287 @@
...
+const headerMessageSchema = new Schema<HeaderMessage>({
+  answerStyle: new StringValue(),
+  contentFormat: new StringValue(),
+});
+
+const messageSchema = new Schema({
+  textDelta: new StringValue(),
+});
+
+const citationsSchema = new Schema({
```
olamothe, line 75: Hum, interesting that all the endpoint output gets validated with Bueno... 🤔 

Typically we use it to validate implementers code (ie: validate arguments types/shape that they pass to Headless functions), and not what our API return.

I can see it will make it easier to "detect that something went horribly wrong from the API", but.... we'll still just end up showing an irrecoverable error to the end user, just in a different manner.

And nothing that an implementer/client can do to fix if that happen either, except wait for us to fix.

So, all in all, interesting, but I'm dubious about what tangible improvement it will bring to our users.

olamothe (approved):

packages/headless/src/api/knowledge/stream-answer-api.ts
dmgauthier: I share your uncertainty about the dubious value it will add to our end user. 

Being new working in the ui-kit, after seeing this comment https://github.com/coveo/ui-kit/pull/4014#discussion_r1654943794 I just thought I would overkill it. 

I welcome any suggestion to make the type check more relevant.

jelmedini (approved):
Nice work !!! (sorry for the late review 🙈 )

PR #4015, opened 2024-05-27
feat(quantic): rich text formatting for quantic RGA

erocheleau:
SFINT-5459

Rich Text Formatting supported within the Quantic RGA component.

Very similar in essence to what @lbergeron did.

Some tweaks and changes were made to the styling of the rendered HTML based on Salesforce design system.

Unit tests were made for the new components and code, but i can't find a good way to create e2e tests for this feature.
It heavily depends on first receiving an EventStream message to set the answerType, and THEN renders markdown as HTML. We don't have the setup to be able to mock two consecutive RGA EventStream events, only one.

Some examples of rendered HTML:

Code blocks
image

Table: has a maximum height and is scrollable horizontally and vertically
image

Different levels of headers, all headers are repleaced with <div> elements instead of using `h1,h2,h3,etc.` elements in order to not break headings hierarchy in the page and break accessibility rules:
image

Inline code 
image

With Collapse = true
image

Added some Cypress tests too!

I am covering most of the "irregular" cases where we differ from the simple behavior of the `marked` library.

image


packages/quantic/copy-static-resources.js
```diff
@@ -12,7 +12,26 @@ const copy = async (source, dest) => {
...
+
+  await mkdir('./force-app/main/default/staticresources/marked', {
+    recursive: true,
+  });
+
+  await copy(
+    '../../node_modules/marked/marked.min.js',
+    './force-app/main/default/staticresources/marked/marked.min.js'
+  );
+};
```
lbergeron, line 31: I would also add a trace after Marked is copied, just to match Headless.

```suggestion
const copyMarked = async () => {
  console.info('Begin copy Marked.');

  await mkdir('./force-app/main/default/staticresources/marked', {
    recursive: true,
  });

  await copy(
    '../../node_modules/marked/marked.min.js',
    './force-app/main/default/staticresources/marked/marked.min.js'
  );
  
  console.info('Marked copied.');
};
```

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,62 @@
...
+        .catch((error) => {
+          console.error('Error loading marked', error);
+        });
+    }
+  }
+
+  updateHtmlContent() {
+    const newHTMLContent =
+      // @ts-ignore
+      window.marked && transformMarkdownToHtml(this.answer, window.marked);
```
lbergeron, line 43: :thinking: Suppose `window.marked` is falsy, then the element will be set to `false`... I guess, in this case we should raise an error.

lbergeron (commented):
That's very nice!

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/__tests__/quanticGeneratedAnswerContent.test.js
```diff
@@ -0,0 +1,120 @@
...
+  describe('text/plain answer content', () => {
+    it('should render a simple text/plain answer when not streaming', async () => {
+      const element = createTestComponent({
+        ...defaultOptions,
+        answer: 'Hello, world!',
+      });
+      await flushPromises();
+
+      const answerContent = element.shadowRoot.querySelector(
+        'span.generated-answer-content__answer'
```
mmitiche, line 58: I suggestion creating a `selectors` object at the top of this file to avoid re-writing these multiple times.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/__tests__/quanticGeneratedAnswerContent.test.js
mmitiche, line 1: I would add a reactivity test for this component to confirm that in the situation where `answerContentFormat ` is first set to `text/plain` then updated to be `markdown/plain`, the component reacts properly by loading the Marked library and formatting the markdown and to confirm that component does not stick with its initial behaviour.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/__tests__/quanticGeneratedAnswerContent.test.js
```diff
@@ -0,0 +1,120 @@
...
+      document.body.removeChild(document.body.firstChild);
+    }
+    jest.clearAllMocks();
+  }
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  describe('text/plain answer content', () => {
```
mmitiche, line 49: Test idea:
We can also test that the Marked library was not loaded in this situation.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/__tests__/quanticGeneratedAnswerContent.test.js
```diff
@@ -0,0 +1,120 @@
...
+      );
+
+      expect(answerContent.textContent).toBe('Hello, world!');
+      expect(answerContent.className).toContain(
+        'generated-answer-content__answer--streaming'
+      );
+    });
+  });
+
+  describe('text/markdown answer content', () => {
```
mmitiche, line 84: We can also check in the other hand the marked library was loaded in this situation.

mmitiche (commented):
Nice job 👍

packages/quantic/force-app/main/default/lwc/quanticUtils/markdownUtils.js
```diff
@@ -0,0 +1,88 @@
+const unclosedElement = /(\*{1,3}|`)($|\w[\w\s]*$)/;
```
SimonMilord, line 1: It would be nice to add a comment to detail this regex

packages/quantic/force-app/main/default/lwc/quanticUtils/markdownUtils.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * complete unclosed markdown elements in a string
```
SimonMilord, line 4:
```suggestion
 * Closes unclosed markdown elements in a string
```

SimonMilord, line 4:
```suggestion
 * Completes unclosed markdown elements in a string
```

packages/quantic/force-app/main/default/lwc/quanticUtils/markdownUtils.js
```diff
@@ -0,0 +1,88 @@
...
+    return `<pre><code>${escapeHtml(code)}</code></pre>`;
+  },
+
+  /**
+   * @param {string} text
+   * @param {string} level
+   */
+  heading(text, level) {
+    return `<div data-level="answer-heading-${level}" aria-label="${text}">${text}</div>`;
+  },
```
SimonMilord, line 46:
```suggestion
  /**
   * Custom Marked renderer for heading elements.
   * @param {string} text
   * @param {string} level
   */
  heading(text, level) {
    return `<div data-level="answer-heading-${level}" aria-label="${text}">${text}</div>`;
  },
```

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
```diff
@@ -45,7 +45,12 @@
           class="generated-answer__content slds-var-m-top_medium"
         >
           <div data-cy="generated-answer__answer" class={generatedAnswerClass}>
```
SimonMilord, line 47:
```suggestion
          <div data-cy="generated-answer__answer" class={generatedAnswerContentClass}>
```

To avoid confusion since we are already using `generatedAnswerClass` in the `quanticGeneratedAnswerContent` component?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+
+  connectedCallback() {
+    if (this.answerContentFormat === 'text/markdown') {
+      loadScript(this, MARKED_JS + '/marked.min.js')
+        .then(() => {
+          this._ready = true;
+          this.updateHtmlContent();
+        })
+        .catch((error) => {
+          console.error('Error loading marked', error);
```
SimonMilord, line 36:
```suggestion
          console.error('Error loading the Marked.JS library', error);
```

SimonMilord: If marked doesn't load we still display the answer?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+        });
+    }
+  }
+
+  updateHtmlContent() {
+    const newHTMLContent =
+      // @ts-ignore
+      (window.marked && transformMarkdownToHtml(this.answer, window.marked)) ||
+      '';
```
SimonMilord, line 45:
```suggestion
    const newHTMLContent =
      // @ts-ignore
      (window.marked && transformMarkdownToHtml(this.answer, window.marked)) || '';
```

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+    const answerContainer = this.template.querySelector(
+      '.generated-answer-content__answer'
+    );
```
SimonMilord, line 48:
```suggestion
    const answerContainer = this.template.querySelector('.generated-answer-content__answer');
```

SimonMilord (commented):
very nice! 🎉

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+import {transformMarkdownToHtml} from 'c/quanticUtils';
+// @ts-ignore
+import {loadScript} from 'lightning/platformResourceLoader';
+import {LightningElement, api} from 'lwc';
+// @ts-ignore
+import generatedMarkdownContentTemplate from './templates/generatedMarkdownContent.html';
+// @ts-ignore
+import generatedTextContentTemplate from './templates/generatedTextContent.html';
+
+export default class QuanticGeneratedAnswerContent extends LightningElement {
```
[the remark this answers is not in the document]
erocheleau: good point

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/__tests__/quanticGeneratedAnswerContent.test.js
erocheleau: We don't actually load the library in the tests, I'm the one mocking it, so we can't test that the library was not loaded.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/__tests__/quanticGeneratedAnswerContent.test.js
erocheleau: No we don't load the actual library in this test.

It just tests that we render the correct HTML container basically for the HTML content.

I'll add an expectation that it has the property `dom:manual` though.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/__tests__/quanticGeneratedAnswerContent.test.js
erocheleau: That's not really something that happens though.
The answerContent component only gets rendered when we've received an answer.
And the first message of the answer is the answerType, so it doesn't first gets rendered as text and then changes to markdown.

If you were to do one query, get a response as text, and then do another query, get another response as markdown after. 
The component doesn't "remain" between the queries, it gets destroyed, and then re-created when the 2nd answer comes in.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+    if (this.answerContentFormat === 'text/markdown') {
+      this.updateHtmlContent();
+    }
+  }
+
+  _ready = false;
+  _answer;
+
+  connectedCallback() {
+    if (this.answerContentFormat === 'text/markdown') {
```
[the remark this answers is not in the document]
erocheleau: The component doesn't stay "connected" in between rendering of different answers.

Since the `QuanticGeneratedAnswer` component doesn't render the `QuanticGeneratedAnswerContent` component when there's no answer, or when you change your query to a different one, the `QuanticGeneratedAnswerContent` component completely gets disconnected and re-connected in the page.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+      loadScript(this, MARKED_JS + '/marked.min.js')
+        .then(() => {
+          this._ready = true;
```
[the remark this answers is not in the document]
erocheleau: Yea I think it was left from a previous thing I tried I guess.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+
+  get generatedAnswerClass() {
+    return `generated-answer-content__answer ${this.isStreaming ? 'generated-answer-content__answer--streaming' : ''}`;
+  }
+
+  get content() {
+    return this._answer;
+  }
```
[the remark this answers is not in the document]
erocheleau: I actually wanted to make it clear that we were using something different in Text/Plain than in Text/HTML.

I'll rename the getter to `contentAsText`

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
erocheleau: This is prettier, not me.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
erocheleau: Also prettier, not me.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,68 @@
...
+        .catch((error) => {
+          console.error('Error loading marked', error);
+        });
+    }
+  }
+
+  updateHtmlContent() {
+    const newHTMLContent =
+      // @ts-ignore
+      (window.marked && transformMarkdownToHtml(this.answer, window.marked)) ||
```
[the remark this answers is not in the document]
erocheleau: The utils code imported cannot "load" the markedJs library. 
https://developer.salesforce.com/docs/platform/lwc/guide/js-third-party-library.html

It has to also import the lightning platform resource loader, which I'm pretty sure can only be done inside a real LWC.

And the first param of the `loadScript` has to be an instance of a `LightningElement`. 

So either the lwc passes the instance of `window.marked` to the utils file, or it has to pass `this` in order to load the script.

Also this way, with dependency injection, helps greatly with the unit tests ;)

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.css
```diff
@@ -7,8 +7,8 @@
 
 .generated-answer__answer {
   white-space: pre-wrap;
-  line-height: 1.5rem;
-  font-size: 15px;
+  line-height: 1.25rem;
+  font-size: 0.8125rem;
```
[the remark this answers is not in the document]
erocheleau: Maybe not, it does change the styling a little bit to remove them though.

Overall only a tiny difference so I'll remove it

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/templates/generatedAnswer.html
erocheleau: There's not really any confusion of two variable names in two different files?

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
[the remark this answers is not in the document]
mmitiche: But at this point the library is already loaded and stored in the window object and accessible via `window.marked`  so the utility code does not need to load it again, it could just call window.marked and access whatever method available in this library.

packages/quantic/copy-static-resources.js
```diff
@@ -12,7 +12,28 @@ const copy = async (source, dest) => {
...
+
+  await copy(
+    '../../node_modules/marked/marked.min.js',
+    './force-app/main/default/staticresources/marked/marked.min.js'
+  );
+
+  console.log('Marked copied.');
+};
+
+const copyHeadless = async () => {
```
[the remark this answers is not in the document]
erocheleau: Alright

SimonMilord (approved):

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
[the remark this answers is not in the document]
louis-bompart: ⚠️ Cypress tests are banned, no exclusion

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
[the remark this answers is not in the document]
erocheleau: Yes but there's no way for the utility code to "know" or wait for the lib to be loaded.

By requiring it as a parameter you enforce that it's already loaded before this method (which doesn't work without the lib loaded) has access to `window.marked`.

If I were to simply do a `if (window.marked)` for example I would have to do nothing in case the lib wasn't loaded and you can't handle this error case in any way, you'd have to just do nothing and say "well please load marked before calling me. Which is not really a good practice.

With dependency injection for example, we can 1. mock it entirely for testing purposes which is a great plus.

SimonMilord (approved):

mmitiche (commented):
Looks good to me ✅
Consider it approved on my side, I will be just waiting for adding the Dom purify before re-reviewing and approving.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
erocheleau: Should be handled now

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,89 @@
...
+import {loadScript} from 'lightning/platformResourceLoader';
+import {LightningElement, api} from 'lwc';
+// @ts-ignore
+import generatedMarkdownContentTemplate from './templates/generatedMarkdownContent.html';
+// @ts-ignore
+import generatedTextContentTemplate from './templates/generatedTextContent.html';
+
+/**
+ * The `QuanticGeneratedAnswerContent` component displays the generated answer content.
+ * @category Search
```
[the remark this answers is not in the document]
erocheleau: Alright makes sense

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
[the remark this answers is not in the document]
erocheleau: Yea I prefer it that way it's explicit rather than trusting that there will be a "global" object be defined

SimonMilord (approved):
Nice job on the tests!

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
```diff
@@ -0,0 +1,116 @@
...
+        // @ts-ignore
+        (window.marked &&
+          // @ts-ignore
+          transformMarkdownToHtml(this.answer, window.marked)) ||
+        '';
+      if (answerContainer) {
+        try {
+          // @ts-ignore
+          // eslint-disable-next-line @lwc/lwc/no-inner-html
+          answerContainer.innerHTML = DOMPurify.sanitize(newHTMLContent);
```
mmitiche, line 89: What about a test to make sure that the HTML was really purified?

mmitiche (approved):
Nice job 👍
It would be nice if we could have a test covering the DOM purification logic.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswerContent/quanticGeneratedAnswerContent.js
erocheleau: There's not much use in testing the DOMPurity library, we know it works.

louis-bompart (approved):
RSLGTM: @coveo/search was included 'cause of the lockfile, so I did not review the code.

fbeaudoincoveo (approved):

PR #4016, opened 2024-05-28
feat(Commerce Headless): change the way selectors are exposed

y-lakhdar:
The initial idea was to refactor all selectors in a way that the arguments of the selector function would be the commerce engine rather than the state.

However, this would require a significant amount of effort for the current need. Since we only need to use a couple of selectors from all the existing ones (and it is unlikely that we will need to use the majority of them outside of headless), it is not wise IMO to spend all the time changing all the selectors and their unit tests.

Instead, I have simply wrapped all the selectors that we need outside of headless in a function that accepts an engine.

WDYT?

https://coveord.atlassian.net/browse/KIT-3199


Spuffynism (approved):

packages/headless/src/selectors/commerce-selectors.index.ts
```diff
@@ -1,12 +1,22 @@
...
+  export const responseIdSelector = wrap(SearchSelectors.responseIdSelector);
 }
 
 export namespace ProductListing {
-  export const responseIdSelectorFromEngine =
-    ProductListingSelectors.responseIdSelectorFromEngine;
+  export const responseIdSelector = wrap(PLSelectors.responseIdSelector);
+}
+
+function wrap<
```
olamothe, line 14:
```suggestion
function wrapWithAccessFromEngine<
```

olamothe (approved):

fbeaudoincoveo (approved):

PR #4020, opened 2024-05-28
chore: add deprecation note for DNT

samisayegh:
The PR adds a deprecation note to remove DNT as part of the next major release.
The decision is captured [here](https://coveord.atlassian.net/browse/LENS-1502?focusedCommentId=764668). The main drivers were:
- DNT is an outdated standard. Newer standards like GPC have been introduced. We don't believe it's Coveo's role to update libraries to adhere to the latest standards?
- Privacy standards differ around the world. Baking a single privacy stance into Coveo's libraries and apis will be overly restrictive in certain geographies.

**Aside**
I tried to adjust the logic in headless to only apply DNT if the analyticsMode is `legacy`. However, this required changing an action to an thunk. The signature change created multiple typescript errors, and so seems like it's breaking. I changed course to add a deprecation note instead. With the v3 release around the corner, it seems less risky to simply remove the `doNotTrack` function rather than making a signature change.


sallainCoveo (approved):

jpmarceau (approved):

fbeaudoincoveo (approved):

samisayegh:
Made the changes @olamothe. Removing `doNotTrack` from the action simplifies the problem, but is a change in behavior. Are we ok with this?

olamothe (approved):

packages/headless/src/app/engine.ts
```diff
@@ -198,7 +198,7 @@ function getUpdateAnalyticsConfigurationPayload(
     apiBaseUrl,
   };
 
-  if (doNotTrack()) {
+  if (payloadWithURL.analyticsMode !== 'next' && doNotTrack()) {
```
louis-bompart, line 201:
```suggestion
  // TODO KIT-2844
  if (payloadWithURL.analyticsMode !== 'next' && doNotTrack()) {
```

packages/headless/src/utils/utils.ts
```diff
@@ -48,6 +48,11 @@ const doNotTrackValues = new Set(['1', 1, 'yes', true]);
```
louis-bompart, line 48:
```suggestion
// TODO KIT-2844

```

PR #4023, opened 2024-05-29
feat(headless): add method to set range on Date Facet

y-lakhdar:
* Add `setRanges()` method to easily apply a date range filter
* Add missing getter on the state

https://coveord.atlassian.net/browse/KIT-3227


packages/headless/src/features/commerce/facets/facet-set/facet-set-slice.test.ts
```diff
@@ -20,6 +20,7 @@ import {
   buildMockCommerceRegularFacetValue,
```
fbeaudoincoveo, line 20: Someday I'd like to split this into a few files, say one per facet type or something. This test file is getting insane 😂

fbeaudoincoveo (approved):

Spuffynism (approved):

PR #4025, opened 2024-05-29
docs(atomic): DOC-15095-atomic-ipx-recs-list-component-should-be-internal

telmaq:
added @internal to the `atomic-ipx-recs-list` component


dguerinCoveo (approved):
thanks!

jpmarceau (approved):

PR #4026, opened 2024-05-29
feat(atomic commerce): add interactive product controller support

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3149
https://coveord.atlassian.net/browse/KIT-3165

### DIFF
--- a/packages/headless/src/api/commerce/common/product.ts
+++ b/packages/headless/src/api/commerce/common/product.ts
@@ -1,10 +1,10 @@
 export type ChildProduct = Omit<
-  RawProduct,
+  BaseProduct,
   'children' | 'totalNumberOfChildren'
 >;
 
 // TODO: KIT-3164 update based on https://coveord.atlassian.net/browse/DOC-14667
-export interface RawProduct {
+export interface BaseProduct {
   /**
    * The SKU of the product.
    */
@@ -109,7 +109,7 @@ export interface RawProduct {
   totalNumberOfChildren: number;
 }
 
-export interface Product extends RawProduct {
+export interface Product extends BaseProduct {
   /**
    * The 1-based product's position across the non-paginated result set.
    *
--- a/packages/headless/src/api/commerce/common/response.ts
+++ b/packages/headless/src/api/commerce/common/response.ts
@@ -5,12 +5,12 @@ import {
   SearchAPIErrorWithStatusCode,
 } from '../../search/search-api-error-response';
 import {Pagination} from './pagination';
-import {RawProduct} from './product';
+import {BaseProduct} from './product';
 import {Sort} from './sort';
 
 export interface BaseCommerceSuccessResponse {
   responseId: string;
-  products: RawProduct[];
+  products: BaseProduct[];
   pagination: Pagination;
   triggers: Trigger[];
 }
--- a/packages/headless/src/commerce.index.ts
+++ b/packages/headless/src/commerce.index.ts
@@ -26,7 +26,7 @@ export type {LogLevel, LoggerOptions} from './app/logger';
 export type {NavigatorContext} from './app/navigatorContextProvider';
 
 export type {
-  RawProduct,
+  BaseProduct,
   Product,
   ChildProduct,
 } from './api/commerce/common/product';
--- a/packages/headless/src/features/commerce/instant-products/instant-products-slice.test.ts
+++ b/packages/headless/src/features/commerce/instant-products/instant-products-slice.test.ts
@@ -2,7 +2,7 @@ import {SearchCommerceSuccessResponse} from '../../../api/commerce/search/respon
 import {
   buildMockChildProduct,
   buildMockProduct,
-  buildMockRawProduct,
+  buildMockBaseProduct,
 } from '../../../test/mock-product';
 import {
   fetchInstantProducts,
@@ -202,7 +202,7 @@ describe('instant products slice', () => {
         const action = fetchInstantProducts.fulfilled(
           {
             response: {
-              products: [buildMockRawProduct()],
+              products: [buildMockBaseProduct()],
               pagination: {totalEntries: 123},
               responseId: 'someid',
             } as unknown as SearchCommerceSuccessResponse,
@@ -246,7 +246,7 @@ describe('instant products slice', () => {
         const action = fetchInstantProducts.fulfilled(
           {
             response: {
-              products: [buildMockRawProduct()],
+              products: [buildMockBaseProduct()],
               pagination: {totalEntries: 123},
               responseId: 'someid',
             } as unknown as SearchCommerceSuccessResponse,
@@ -289,8 +289,8 @@ describe('instant products slice', () => {
           {
             response: {
               products: [
-                buildMockRawProduct({ec_name: 'product1'}),
-                buildMockRawProduct({ec_name: 'product2'}),
+                buildMockBaseProduct({ec_name: 'product1'}),
+                buildMockBaseProduct({ec_name: 'product2'}),
               ],
               pagination: {totalEntries: 22},
               responseId: 'someid',
--- a/packages/headless/src/features/commerce/instant-products/instant-products-slice.ts
+++ b/packages/headless/src/features/commerce/instant-products/instant-products-slice.ts
@@ -1,5 +1,5 @@
 import {createReducer} from '@reduxjs/toolkit';
-import {Product, RawProduct} from '../../../api/commerce/common/product';
+import {Product, BaseProduct} from '../../../api/commerce/common/product';
 import {
   clearExpiredItems,
   fetchItemsFulfilled,
@@ -104,7 +104,7 @@ export const instantProductsReducer = createReducer(
   }
 );
 
-function preprocessProduct(product: RawProduct, position: number): Product {
+function preprocessProduct(product: BaseProduct, position: number): Product {
   const isParentAlreadyInChildren = product.children.some(
     (child) => child.permanentid === product.permanentid
   );
--- a/packages/headless/src/features/commerce/product-listing/product-listing-slice.test.ts
+++ b/packages/headless/src/features/commerce/product-listing/product-listing-slice.test.ts
@@ -2,7 +2,7 @@ import {buildMockCommerceRegularFacetResponse} from '../../../test/mock-commerce
 import {
   buildMockChildProduct,
   buildMockProduct,
-  buildMockRawProduct,
+  buildMockBaseProduct,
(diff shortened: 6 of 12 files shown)


packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
```diff
@@ -193,12 +194,48 @@ export class AtomicCommerceProductList
...
+    } else {
+      parentController = buildSearch(this.bindings.engine);
+    }
+
+    const {name, price, productId, warning} =
+      ProductTemplatesHelpers.getRequiredProductPropertiesForAnalytics(product);
+
+    const pagination = parentController.pagination();
+    const position =
+      pagination.state.page * pagination.state.pageSize + index + 1;
```
fbeaudoincoveo, line 214: Expected position is 1-based (see https://github.com/coveo-platform/analytics_schema/blob/0dc47fa9ad35aa119e7a2f22592e09974e860685/schemas/client/api.event.ec.productClick/1/api.event.ec.productClick.1.request.json#L27)

Also confirmed with @samisayegh that position must account for pagination offset. This will be clarified in the documentation (see [LENS-1900](https://coveord.atlassian.net/browse/LENS-1900))

packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
```diff
@@ -193,12 +194,48 @@ export class AtomicCommerceProductList
...
+
+    const controller =
+      name === undefined || price === undefined || productId === undefined
+        ? undefined
```
fbeaudoincoveo, line 218: I think it's preferable to completely avoir building the interactive product controller when some required properties are missing. This is why I made it an optional prop on the atomic-product component.

Another option would be to initialize the controller regardless, and use placeholder values.

packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
```diff
@@ -218,31 +255,40 @@ export class AtomicCommerceProductList
...
           }}
-          cancelPendingSelect={function (): void {
-            throw new Error('Function not implemented. TODO KIT-3149');
+          cancelPendingSelect={() => {
+            analyticsWarning
+              ? this.logWarning(analyticsWarning)
+              : interactiveProduct.controller?.cancelPendingSelect();
           }}
         >
-          <atomic-product {...this} {...propsForAtomicProduct}></atomic-product>
```
fbeaudoincoveo, line 245: @fpbrault was there any legitimate reason to pass {...this} to the atomic-product?

fpbrault: I took this from atomic-result/atomic-result-list, but I'm looking at it and can't find anything that wouldn't be in propsForAtomicProduct or already in AtomicProduct..!

packages/headless/src/features/commerce/product-templates/product-templates-helpers.ts
```diff
@@ -105,8 +150,88 @@ const getFieldValuesFromProduct = (
...
+
+  if (value === undefined) {
+    return value;
+  }
+
+  return value as number;
+};
+
+export const getProductNameFromStandardLookupFields = (product: Product) => {
+  const lookupFields = ['ec_name', 'ec_product_id', 'permanentid'];
```
fbeaudoincoveo, line 194: At the moment I'm using a hardcoded, opinionated list of lookup values that will be verified in the specified order (i.e., in this case if ec_name isn't defined on a product, or if its value isn't a string, then we will fallback to ec_product_id, and then to permanentid).

Eventually, customers might come up with valid use cases for specifying their own lookup fields (which would probably be set in additionalFields). If this happens, it will be straightforward to add options on the getRequiredProductPropertiesForAnalytics helper function.

We could even image cases where, for instance, determining the field to use for the price could be done by a function rather than a simple list of prioritized lookup fields. Again, having a single function to get the required product properties for analytics allows us to add support for this as needed.

For now, though, I think we should keep it simple 😁

Spuffynism (approved):

Spuffynism (approved):

olamothe (approved):

fbeaudoincoveo:
@Spuffynism , @olamothe this PR has changed significantly enough that I think it would be worth your time to review it again.

packages/atomic/src/pages/examples/commerce-website/homepage.html
```diff
@@ -12,10 +12,12 @@
...
       (async () => {
-        await customElements.whenDefined('atomic-commerce-interface');
-        const commerceInterface = document.querySelector('atomic-commerce-recommendation-interface');
+        await customElements.whenDefined('atomic-commerce-recommendation-interface');
+        const recommendationInterfaces = document.querySelectorAll('atomic-commerce-recommendation-interface');
 
-        await commerceInterface.initializeWithEngine(commerceEngine);
+        for (const recommendationInterface of recommendationInterfaces) {
+          await recommendationInterface.initializeWithEngine(commerceEngine);
+        }
```
louis-bompart, line 20: Nit:

```suggestion
        await Promise.all(recommendationInterfaces.map(interface => interface.initializeWithEngine(commerceEngine))
```

packages/atomic/src/pages/examples/commerce-website/cart.html
```diff
@@ -12,9 +12,12 @@
...
       (async () => {
-        await customElements.whenDefined('atomic-commerce-interface');
-        const commerceInterface = document.querySelector('atomic-commerce-recommendation-interface');
-        await commerceInterface.initializeWithEngine(commerceEngine);
+        await customElements.whenDefined('atomic-commerce-recommendation-interface');
+        const recommendationInterfaces = document.querySelectorAll('atomic-commerce-recommendation-interface');
+
+        for (const recommendationInterface of recommendationInterfaces) {
+          await recommendationInterface.initializeWithEngine(commerceEngine);
+        }
```
louis-bompart, line 20: Nit:

```suggestion
        await Promise.all(recommendationInterfaces.map(interface => interface.initializeWithEngine(commerceEngine))
```

packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
```diff
@@ -217,12 +217,26 @@ export class AtomicCommerceProductList
...
+    }
+  }
+
+  private getInteractiveProduct(product: Product) {
+    let parentController: ProductListing | Search;
+    if (this.bindings.interfaceElement.type === 'product-listing') {
+      parentController = this.productListing;
+    } else {
+      parentController = this.search;
+    }
```
louis-bompart, line 232:
```suggestion
    const parentController =
      this.bindings.interfaceElement.type === "product-listing"
        ? this.productListing
        : this.search;
```

packages/atomic/src/components/commerce/atomic-commerce-recommendation-list/atomic-commerce-recommendation-list.tsx
```diff
@@ -359,20 +343,32 @@ export class AtomicCommerceRecommendationList
...
+        cancelPendingSelect={() => {
+          this.logWarningIfNeeded(interactiveProduct.warningMessage);
+          interactiveProduct.cancelPendingSelect();
+        }}
         setRef={(element) =>
           element && this.itemListCommon.setNewResultRef(element, i)
         }
       >
-        <atomic-product {...atomicProductProps}></atomic-product>
+        <atomic-product {...propsForAtomicProduct}></atomic-product>
```
louis-bompart, line 371: is InteractiveProduct being used in the atomic-product?

If not, see my suggestion above L346-347.
If yes, dismiss it

packages/atomic/src/components/commerce/atomic-commerce-recommendation-list/atomic-commerce-recommendation-list.tsx
```diff
@@ -359,20 +343,32 @@ export class AtomicCommerceRecommendationList
   }
 
   private renderAsGrid(product: Product, i: number) {
-    const atomicProductProps = this.getAtomicProductProps(product);
+    const propsForAtomicProduct = this.getAtomicProductProps(product);
+    const {interactiveProduct} = propsForAtomicProduct;
```
louis-bompart, line 347:
```suggestion
    const {interactiveProduct, ...propsForAtomicProduct} = this.getAtomicProductProps(product);
```

louis-bompart (approved):
only some nits

packages/atomic/src/components/commerce/atomic-commerce-recommendation-list/atomic-commerce-recommendation-list.tsx
fbeaudoincoveo: Dismissing since interactiveProduct must be passed as a prop to atomic-product anyways

packages/headless/src/api/commerce/common/product.ts
```diff
@@ -1,7 +1,10 @@
-export type ChildProduct = Omit<Product, 'children' | 'totalNumberOfChildren'>;
+export type ChildProduct = Omit<
+  RawProduct,
+  'children' | 'totalNumberOfChildren'
+>;
 
 // TODO: KIT-3164 update based on https://coveord.atlassian.net/browse/DOC-14667
-export interface Product {
+export interface RawProduct {
```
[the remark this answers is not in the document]
fbeaudoincoveo: The idea is to make the transition to the new interactive product smoother.

By keeping "Product" as the name of the augmented interface, we ensure that the few current implementations that are already using the headless commerce sub-package will import it if they were previously importing the non-augmented one. Thus, they'll automatically have access to the `position` property, and they'll be able to simply pass a product object when building an interactive product controller.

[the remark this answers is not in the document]
fbeaudoincoveo: For using BaseProduct instead of RawProduct, sure!

packages/atomic/src/pages/examples/commerce-website/homepage.html
fbeaudoincoveo: It's a NodeList, so as discussed won't do (we'd have to use Array.from, to be able to call map)

PR #4028, opened 2024-05-29
fix(quantic): Sort dropdown and tooltip partially hidden by Salesforce container 

SimonMilord:
[SFINT-5552](https://coveord.atlassian.net/browse/SFINT-5552)

- For all issues: The main culprit was the component container in Salesforce. We don't have control over that and it also makes it virtually impossible to precisely know where the end of the container is since Salesforce doesn't seem to allow us to target it. So we can't quite use JS to get coordinates of elements and do some math to figure out when the component is hidden or not.... I therefore had to find crafty solutions...

### Before:
image

### After:
Screenshot 2024-05-29 at 4 17 18 PM

### Solution:
- The `lightning-combobox` had the class: `slds-dropdown_fluid` automatically added by Salesforce which was adding the following properties: `min-width 12rem`, `max-width: 100%` and `width: 100%` which was causing the dropdown to overflow on the container.
- The solution here is essentially to create a CSS class `sort__combobox` which overrides the properties set by `slds-dropdown_fluid`. I could not just remove the min-width, because then the labels would be truncated.

### Before:
image

### After:
Screenshot 2024-05-29 at 12 31 50 PM

Screenshot 2024-05-29 at 12 32 06 PM

### Solution:
- The solution found here was to change the label to something shorter so that it doesn't get hidden. I replaced the old label by ON/OFF since its already pretty obvious we are talking of the generatedAnswer component.
- Why I think this is acceptable: We will eventually remove this toggle button so why overthink this and also the copy to clipboard label : `copy` is also small enough to not get hidden when we do remove the toggle.

## ISSUE 3: Tooltips getting partially hidden on the top
Will be fixed in a seperate PR in the C4SF package

[SFINT-5552]: https://coveord.atlassian.net/browse/SFINT-5552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -296,7 +296,8 @@
   </labels>
   <labels>
     <fullName>quantic_NoEndpointsDesc</fullName>
-    <value>You will need to add sources in your index, or wait for the created sources to finish indexing.</value>
+    <value>You will need to add sources in your index, or wait for the created sources to finish
+      indexing.</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Add sources in your index, or wait for the created sources to finish indexing.</shortDescription>
@@ -376,11 +377,13 @@
     <value>You may want to try using different keywords, or checking for spelling mistakes.</value>
     <language>en_US</language>
     <protected>false</protected>
-    <shortDescription>You may want to try using different keywords, or checking for spelling mistakes.</shortDescription>
+    <shortDescription>You may want to try using different keywords, or checking for spelling
+      mistakes.</shortDescription>
   </labels>
   <labels>
     <fullName>quantic_NoResultsWithFilters</fullName>
-    <value>You may want to try using different keywords, checking for spelling mistakes or select fewer filters.</value>
+    <value>You may want to try using different keywords, checking for spelling mistakes or select
+      fewer filters.</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Check for spelling mistakes or select fewer filters.</shortDescription>
@@ -1136,7 +1139,8 @@
   </labels>
   <labels>
     <fullName>quantic_InvalidNumberOfAutoOpenedDocuments</fullName>
-    <value>The number of automatically opened document suggestions must be an integer greater that 0.</value>
+    <value>The number of automatically opened document suggestions must be an integer greater that
+      0.</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Invalid number of automatically opened document suggestions.</shortDescription>
@@ -1409,14 +1413,14 @@
   </labels>
   <labels>
     <fullName>quantic_GeneratedAnswerOn</fullName>
-    <value>Generated answer ON</value>
+    <value>ON</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Generated answer ON</shortDescription>
   </labels>
   <labels>
     <fullName>quantic_GeneratedAnswerOff</fullName>
-    <value>Generated answer OFF</value>
+    <value>OFF</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Generated answer OFF</shortDescription>
--- a/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
+++ b/packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
@@ -1413,14 +1413,14 @@
   </labels>
   <labels>
     <fullName>quantic_GeneratedAnswerOn</fullName>
-    <value>ON</value>
+    <value>Generated answer ON</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Generated answer ON</shortDescription>
   </labels>
   <labels>
     <fullName>quantic_GeneratedAnswerOff</fullName>
-    <value>OFF</value>
+    <value>Generated answer OFF</value>
     <language>en_US</language>
     <protected>false</protected>
     <shortDescription>Generated answer OFF</shortDescription>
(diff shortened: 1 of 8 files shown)


mmitiche (commented):
Solutions 1 and 2 look good to me, however I disagree with solution 3.

I don't think we should change the behaviour/design of the tooltip because of a padding issue, I would qualify this as a minor issue that I think we could easily find ways around it without making significant changes in our components.

In the Quantic components, one of the goals is to create components that have the same look and feel as native Salesforce components.
Thus, when building/modifying Quantic components our reference is the SLDS design system and not Atomic.

In the SLDS design system, the tooltip is presented as the following: https://www.lightningdesignsystem.com/components/tooltips/
We should not get far from the tooltip behaviour inside Salesforce by removing the arrow.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.css
```diff
@@ -1,6 +1,7 @@
 .generated-answer__card-header {
   justify-content: space-between;
   align-items: center;
+  margin-top: 0.6rem;
```
mmitiche, line 4: This issue is only happening with the hosted search page and only when the integrated search box is not being used, so this is not the problem of this component internally.

Thus, to add this space, it's better to add it in the container around your component.
and it's better to use SLDS CSS classes.

This would be something doable in the C4SF package,  in the component that displays the HSP, and when the  integrated search box is not used, you can don this:

```
<div class="slds-m-top_x-small">
 RGA Component...
</div>
```

SimonMilord: ok I will create a seperate PR in C4SF package for that then!

packages/quantic/force-app/main/default/lwc/quanticSort/quanticSort.html
```diff
@@ -10,7 +10,7 @@
           <lightning-formatted-text value={labels.sortBy}></lightning-formatted-text>
         </lightning-layout-item>
         <lightning-layout-item>
-          <lightning-combobox name="sort" value={value} variant="label-hidden" label={labels.sortBy} placeholder={labels.relevancy} options={options} onchange={handleChange}>
+          <lightning-combobox class="sort__combobox" name="sort" value={value} variant="label-hidden" label={labels.sortBy} placeholder={labels.relevancy} options={options} onchange={handleChange}>
```
mmitiche, line 13: I'm not sure how is this CSS class is overriding the Salesforce CSS class defined inside the lightning-combobox component since this class you  are adding is in a higher level than the SF class. are you sure it is overriding it?

mmitiche: In another note I suggest trying the following solution to see if it works:

The `lightning-combobox` component has a public property called: `dropdown-alignment` it specifies where the drop-down list is aligned. One of the values that this property could take is `auto` which will let the component determine where to open the list based on space available.
So it's worth trying to see if this component by it self is able to adapt according to the SF boundaries.

ref: https://developer.salesforce.com/docs/component-library/bundle/lightning-combobox/specification

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -157,15 +158,17 @@
     <name>quantic_DisconnectedTitle</name>
   </customLabels>
   <customLabels>
-    <label>Votre requête n'a pas plus être envoyée. Peut-être que vous avez perdu votre connexion internet.</label>
+    <label>Votre requête n'a pas plus être envoyée. Peut-être que vous avez perdu votre connexion
```
mmitiche, line 161:
```suggestion
    <label>Votre requête n'a pas pu être envoyée.. Peut-être que vous avez perdu votre connexion
```

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -283,8 +287,7 @@
   <customLabels>
     <label>Aperçu non disponible</label>
     <name>quantic_NoPreviewAvailable</name>
-  </customLabels>quantic_Colon
-  <customLabels>
+  </customLabels>quantic_Colon <customLabels>
```
mmitiche, line 290:
```suggestion
  </customLabels>quantic_Colon<customLabels>
```

packages/quantic/force-app/main/default/lwc/quanticSort/quanticSort.html
SimonMilord: 1) Yes I checked it is overriding it.

2) I already tried that when I was experimenting with the potential idea to make it dynamically aligned-right and auto doesnt work :( It was still overflowing... I think it might not account properly for the side of the container or something

erocheleau: Yea I would suggested replacing simply the `dropdown-alignment` to set it to `dropdown-alignment="right"` It makes more sense that the dropdown fits by being right aligned instead of limiting the width etc.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1409,14 +1413,14 @@
   </labels>
   <labels>
     <fullName>quantic_GeneratedAnswerOn</fullName>
-    <value>Generated answer ON</value>
+    <value>ON</value>
```
erocheleau, line 1416: We can't change the values just because they'd fit better xD

Also please don't run prettier on this file it looks stupid to have text on multiple lines.

erocheleau (changes requested):
The thing that's hiding the overflow btw is the `.hosted-search-page_layout[c-hostedSearchPageRenderer_hostedSearchPageRenderer]` This is the class that has the `overflow: hidden` on it, so I don't know why we can't just change that?

PR #4032, opened 2024-05-30
tests: setup playwright

louis-bompart:
Initial setup of playwright

See #4033 for a test demoing the "result" and good practice.


packages/atomic/src/hello-world.e2e.ts
louis-bompart, line 1: Removed in #4033

.github/actions/playwright-atomic/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: Run Playwright tests
+      run: npx playwright test
+      working-directory: packages/atomic
+      shell: bash
+    - uses: actions/upload-artifact@v4
+      if: always()
+      with:
+        name: atomic-playwright-report
+        path: packages/atomic/playwright-report/
+        retention-days: 30
```
fbeaudoincoveo, line 18: If this does what I think it does, that might be a bit much.

I've personally never had to go back in test reports from more than a day ago. I think a retention of 5 days would be more than sufficient (I'd even go with 1 TBH).

fbeaudoincoveo (approved):

.github/actions/playwright-atomic/action.yml
louis-bompart: It's the artifact on the github run, but yeah, true, let's kick it down
```suggestion
        retention-days: 5
```

PR #4033, opened 2024-05-30
test: atomic-commerce-searchbox (playwright demo)

louis-bompart:


packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.tsx
```diff
@@ -92,7 +94,7 @@ import {SelectChildProductEventArgs} from '../product-template-components/atomic
  * @part instant-results-show-all - The clickable suggestion to show all items for the current instant results search rendered by an `atomic-commerce-search-box-instant-products` component.
  * @part instant-results-show-all-button - The button inside the clickable suggestion from the `atomic-commerce-search-box-instant-products` component.
  *
- * @internal
+ * @alpha
```
louis-bompart, line 97: @jpmarceau would there be an easy way to 'hide' or put annotation on doc pages when there's some specific JSDOC tag?

 Otherwise, I'll try again to filter'em'out differently on my side

packages/atomic/storybookUtils/commerce-interface-wrapper.tsx
```diff
@@ -0,0 +1,109 @@
...
+        },
+        {
+          productId: 'SP04236_00005',
+          quantity: 1,
+          name: 'Eco-Brave Rashguard',
+          price: 33,
+        },
+      ],
+    },
+  });
```
louis-bompart, line 58: Could/should be improved by having a similar method in Coveo Headless (and thus an org that we are OK with customer touching)

packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.tsx
jpmarceau: Morning @louis-bompart 
Good question, thanks for checking.
Yeah, I think we could implement that fairly quickly. Feel free to use the `alpha` annotation and I'll handle them docs-side for the next release 👍

olamothe: What's the reason for the tag change ?

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/atomic-commerce-search-box.e2e.ts
```diff
@@ -0,0 +1,234 @@
...
+      test('should collapse the suggested queries', async ({searchBox}) => {
+        await expect(searchBox.searchSuggestions().first()).not.toBeVisible();
+      });
+    });
+  });
+});
+
+test.describe('with instant results & query suggestions', () => {
+  test.beforeEach(async ({page}) => {
+    await page.goto(
```
olamothe, line 54: Since this type of URL will be used for every test, we could have a utility function, that accept a component name or tag, and arguments for the component props (strongly typed? ... derived from the component HTML type 🤔 ?).

And navigate to the page or simply return the built string.

olamothe (approved):
Pretty clean !

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/pageObject.ts
```diff
@@ -0,0 +1,48 @@
+import type {Page} from '@playwright/test';
```
olamothe, line 1: Lets use kebab case like (most) of the repo

`page-object`

packages/atomic/src/components/commerce/facets/atomic-commerce-facets/e2e/pageObject.ts
```diff
@@ -0,0 +1,18 @@
+import type {Page} from '@playwright/test';
```
olamothe, line 1: `page-object`

packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.tsx
louis-bompart: This is due to the generation of the Custom Element Manifest.
`@internal` impacts the visibility modifier (private/public) of the AST handling of the tool. I struggled to wrangle the tool to make it generate the manifest for specific internal tags.

Without a proper manifest, Storybook params handling ain't right, which in turn make the param setting in the test not right either

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/atomic-commerce-search-box.e2e.ts
louis-bompart: I thought about it!

Let's see that in a future iteration.

louis-bompart: https://coveord.atlassian.net/browse/KIT-3241

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/atomic-commerce-search-box.e2e.ts
```diff
@@ -0,0 +1,234 @@
...
+      });
+
+      test('should collapse the suggested queries', async ({searchBox}) => {
+        await expect(searchBox.searchSuggestions().first()).not.toBeVisible();
+      });
+    });
+  });
+});
+
+test.describe('with instant results & query suggestions', () => {
```
mmitiche, line 52: shouldn't we include also the recent queries that can be displayed with the search box as well in this test?

mmitiche (approved):
Very neat in my opinion, clear and simple! Liked the accessibility tests that are added in a very simple fashion 👌

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/atomic-commerce-search-box.e2e.ts
louis-bompart: Tests are partial here, as the goal is more to demonstrate/have an example to create new tests

PR #4041, opened 2024-05-31
feat(quantic): recent query suggestions feature added to quantic search box

mmitiche:
[SFINT-5429](https://coveord.atlassian.net/browse/SFINT-5429)

- Added the feature that displays the recent queries among the query suggestions presented for the user.
- Now the Quantic Search Box by default, includes the recent queries as query suggestions. This feature could be disabled with the new public property: `disableRecentQuerySuggestions `.
- The number of (query suggestions  + recent query suggestions) can be controller by the public property `numberOfSuggestions` and I set by default to be `7`.
- When displaying query suggestion, there are now some considerations:

  1.  When a value is both a query suggestions and a recent query  it should only be displayed a recent query.
  2.  When a value is written in the search box, only the recent queries starting with that value should be displayed.

- When a recent query is click the UA event `recentQueriesClick` is logged.
- Note that with this changes the recent queries will be by default available in the search box. 

PS: the accessibility of this new feature and the component search box in general needs to be improved as a whole and this would require some important changes in that  Quantic Search Box, so I'm planing to work in a second iteration on fixing all the accessibility issues. 

#### The Quantic Search box before:

https://github.com/coveo/ui-kit/assets/86681870/9941cef1-6f6e-475b-a388-8eb60d26aa66

#### The Quantic Search box after:

https://github.com/coveo/ui-kit/assets/86681870/0645effd-2a0a-4a3d-898c-f665de38317d

[SFINT-5429]: https://coveord.atlassian.net/browse/SFINT-5429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/examples/main/lwc/exampleQuanticSearchBox/exampleQuanticSearchBox.js
```diff
@@ -0,0 +1,64 @@
+import {LightningElement, track} from 'lwc';
+
+export default class ExampleQuanticSearchBox extends LightningElement {
+  @track config = {};
+  isConfigured = false;
+
+  pageTitle = 'Quantic Search Box';
+  pageDescription =
+    'The `QuanticSearchBox` component creates a search box with built-in support for query suggestions.';
```
SimonMilord, line 9:
```suggestion
    'The `QuanticSearchBox` component creates a search box with built-in support for query suggestions and query history.';
```

packages/quantic/force-app/examples/main/lwc/exampleQuanticSearchBox/exampleQuanticSearchBox.js
```diff
@@ -0,0 +1,64 @@
...
+  options = [
+    {
+      attribute: 'engineId',
+      label: 'Engine id',
+      description: 'The ID of the engine instance the component registers to.',
```
SimonMilord, line 14:
```suggestion
      description: 'Use Engine ID to connect components with each other.',
```

I think we should add the new description we use for engine ID following etienne's PR

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -46,9 +48,9 @@ export default class QuanticSearchBox extends LightningElement {
    * The maximum number of suggestions to display.
    * @api
    * @type {number}
-   * @defaultValue 5
+   * @defaultValue 7
    */
-  @api numberOfSuggestions = 5;
+  @api numberOfSuggestions = 7;
```
SimonMilord, line 53: we increase this to 7 because numberOfSuggestions = # of suggestions + # of recent queries?
If that is the case I would update at least the description to something like `The maximum number of suggestions and recent search queries to display.`

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -126,13 +166,35 @@ export default class QuanticSearchBox extends LightningElement {
...
 
+  updateRecentQueriesListState() {
+    if (this.recentQueriesList.state?.queries) {
+      this.recentQueries = this.recentQueriesList.state.queries;
+      setItemInLocalStorage(
+        this.localStorageKey,
+        this.recentQueriesList.state.queries
+      );
+    }
+  }
```
SimonMilord, line 184:
```suggestion
updateRecentQueriesListState() {
  const { recentQueries } = this.recentQueriesList.state || {};

  if (recentQueries) {
    this.recentQueries = recentQueries;
    setItemInLocalStorage(this.localStorageKey, recentQueries);
  }
}
```

what do you think of this instead?

packages/quantic/force-app/examples/main/lwc/exampleQuanticSearchBox/exampleQuanticSearchBox.js
mmitiche: This description is used in the documentation page of all the Quantic component, so I would not change this in this PR.

If we wanna change, we will need to change that for all the components at the same time in a separate PR.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
mmitiche: yeah sure the description needs to be updated.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
mmitiche: you probably meant to write `queries` instead of `recentQueries`.
 I'm not really a fan of trying to destructure  a property from a possibly empty object `{}`.

This suggestion could be better, but it's very minor, it's just a coding style preference that does not necessarily have an impact

```
  updateRecentQueriesListState() {
    if (this.recentQueriesList.state?.queries) {
      const {queries} = this.recentQueriesList.state;
      this.recentQueries = queries;
      setItemInLocalStorage(this.localStorageKey, queries);
    }
  }
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.html
```diff
@@ -1,34 +1,34 @@
...
-                    onmouseover={handleSuggestionHover}
-                    onmousedown={preventDefault}>
-                    <template if:true={suggestion.isSelected}>
-                        <div class="slds-media slds-listbox__option slds-listbox__option_plain slds-media_small slds-has-focus" role="option" aria-selected="true">
-                            <span class="slds-media__figure slds-listbox__option-icon"></span>
-                            <span class="slds-media__body">
-                                <lightning-formatted-rich-text class="slds-truncate" value={suggestion.value}></lightning-formatted-rich-text>
+            <template for:each={allOptions} for:item="option">
+                <template lwc:if={option.isClearRecentQueryButton}>
+                    <li data-cy="clear-recent-queries" key={option.key} role="presentation" class="slds-listbox__item"
```
mmitiche, line 6:
```suggestion
                    <li data-cy="clear-recent-queries" key={option.key} role="button" class="slds-listbox__item"
```

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,514 @@
+import {configure} from '../../../page-objects/configurator';
```
SimonMilord, line 1: do we have a playwright setup in quantic? If we are doing the transition from cypress to playwright do we not want to write the tests in playwright instead?

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    logClearRecentQueries: () => {
+      cy.wait(InterceptAliases.UA.ClearRecentQueries)
+        .then((interception) => {
+          const analyticsBody = interception.request.body;
+          expect(analyticsBody).to.have.property('eventType', 'recentQueries');
+        })
+        .logDetail("should log the 'clearRecentQueries' UA event");
+    },
+  };
```
SimonMilord, line 76: what about analytics if we select a recent query or suggestion?

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -88,6 +101,23 @@ export default class QuanticSearchBox extends LightningElement {
...
+    if (
+      !this.disableRecentQuerySuggestions &&
+      this.headless.buildRecentQueriesList
+    ) {
+      this.recentQueriesList = this.headless.buildRecentQueriesList(engine, {
+        initialState: {
+          queries: getItemFromLocalStorage(this.localStorageKey) ?? [],
+        },
+        options: {
+          maxLength: 100,
```
SimonMilord, line 114: why 100 here if we display 7 by default? if the user sets the option to 100 it would display 100 recent queries? That seems a lot x)

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.css
```diff
@@ -1,3 +1,18 @@
...
   margin-top: 0;
+}
+
+.option {
+  align-items: center;
+}
+
+.suggestion-option__icon {
+  display: flex;
+  justify-content: flex-start;
```
SimonMilord, line 11: here why not set `slds-grid` & something like `slds-grid_align-start`?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -87,18 +124,100 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+        icon: option.isRecentQuery ? 'utility:clock' : 'utility:search',
+      }));
+      return [
+        {
+          key: 0,
+          isSelected: this.selectionIndex === 0,
+          isClearRecentQueryButton: true,
+        },
+        ...options,
+      ].slice(0, this.maxNumberOfSuggestions + 1);
```
SimonMilord, line 146: ok the index+ 1 here is because of the recent searches label + clear button option? 
nice!

SimonMilord (commented):
A few comments but otherwise looks great! Nice job.

Mostly wondering if we want to write the e2e in playwright since we are doing the transition anyways in quantic

SimonMilord (approved):

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: Nope we don't have the setup yet in Quantic, we should do that before considering making tests in playwright plus this PR was made before the decision to start using playwright x)

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
mmitiche: That's a very good point 👍 I will update accordingly

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
mmitiche: Basically that's the max number of recent queries that could be considered. But the thing is that not all 100 will be displayed:

- The maximum number of suggestions is set with the public property `numberOfSuggestions` which defaults to 7.
- A lot of recent queries will be filtered out based on the query written on the search box, for example if the user writes `a`, only the recent queries starting with this letter will appear.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
mmitiche: indeed.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
SimonMilord: yeah that makes sense, just wanted to know if instead of 7 the user sets 100 they could see 100 displayed?

mmitiche: with the current logic yes, if the user has 100 recent queries and set the property to 100 yeah 100 recent queries will be displayed x) we can limit the value of this property if we want, but in the same time we are only realizing the user's wish to see 100 recent queries so are not necessarily doing anything wrong x)

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.html
```diff
@@ -1,34 +1,34 @@
...
-                    data-rawvalue={suggestion.rawValue}
-                    onclick={handleSuggestionSelection}
-                    onmouseover={handleSuggestionHover}
-                    onmousedown={preventDefault}>
-                    <template if:true={suggestion.isSelected}>
-                        <div class="slds-media slds-listbox__option slds-listbox__option_plain slds-media_small slds-has-focus" role="option" aria-selected="true">
-                            <span class="slds-media__figure slds-listbox__option-icon"></span>
-                            <span class="slds-media__body">
-                                <lightning-formatted-rich-text class="slds-truncate" value={suggestion.value}></lightning-formatted-rich-text>
+            <template for:each={allOptions} for:item="option">
```
erocheleau, line 4: Why not a 

```
<template lwc:if={hasRecentQueries}>
  <!-- Button to clear recent queries -->
  <template for:each={allOptions} for:item="option">
    <...>
  </template>
</template>
```

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,526 @@
...
+                }
+              );
+              Expect.logClickSuggestion(
+                exampleQuerySuggestions[clickedSuggestionIndex]
+              );
+            });
+          });
+        });
+
+        describe('when a custom value is set for the property numberOfSuggestions', () => {
```
erocheleau, line 405:
```suggestion
        describe('the property #numberOfSuggestions', () => {
```
Are we not doing this kind of pattern anymore?

erocheleau (commented):
Some questions and suggestions

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -88,6 +101,23 @@ export default class QuanticSearchBox extends LightningElement {
         },
       },
     });
+
+    if (
+      !this.disableRecentQuerySuggestions &&
+      this.headless.buildRecentQueriesList
```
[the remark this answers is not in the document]
mmitiche: In the case of the insight use case, the recent queries feature is not yet supported in that use case.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -126,13 +166,35 @@ export default class QuanticSearchBox extends LightningElement {
...
       })) ?? [];
   }
 
+  updateRecentQueriesListState() {
+    if (this.recentQueriesList.state?.queries) {
+      this.recentQueries = this.recentQueriesList.state.queries;
+      setItemInLocalStorage(
+        this.localStorageKey,
+        this.recentQueriesList.state.queries
+      );
```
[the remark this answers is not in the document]
mmitiche: For your first question: `This completely overrides the current existing recent queries in the Local Storage if some exists, is that expected`

No this will no override the existing queries, in other words no recent query will be lost because the headless state contains the same piece of information that can be found in the local storage because in the initialization we do the folowing:
```
      this.recentQueriesList = this.headless.buildRecentQueriesList(engine, {
        initialState: {
          queries: getItemFromLocalStorage(this.localStorageKey) ?? [],
        },
       ...
```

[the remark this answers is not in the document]
mmitiche: For your second question about the Standalone search box and how recent queries are handled with this component:

for now now that is what is supported:

- If you perform a query with  standalone searchbox, it will be added to the recent queries and it will be displayed as a recent query in the search box of the search page:

https://github.com/coveo/ui-kit/assets/86681870/221c9706-d960-4028-a907-84e571f78451


Otherwise, the standalone search box will NOT display  recent queries.

To make this happen, it's fairly complicated technically, we will have to always  keep the state of of the Standalone engine similar to the state of the search engine used in the search page. We can do it but it will not be pretty it will over complicate the QuanticSearchBox component and the QuanticStandaloneSearchBox component.

So in my opinion it will not be worth it to support that, especially that naturally the standalone search box does not need recent queries as usually it is used to perform the first query, then it redirects to the search way where you will  start seeing the recent queries as expected in the search box.  

We cam evaluate of course and decide to add it, but in all cases I would consider this in another iteration as the solution to accomplish this will be complicated/hackish when I imagine it x) and for sure it will come with some corner cases that we will need to pay attention to so it will need to be very heavily tested, that's why I don't see it in this PR.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -126,13 +166,35 @@ export default class QuanticSearchBox extends LightningElement {
...
+   * Clears the recent queries.
+   */
+  clearRecentQueries = (event) => {
+    event.stopPropagation();
+    this.recentQueriesList.clear();
+  };
+
+  get localStorageKey() {
+    return `${this.engineId}_quantic-recent-queries`;
+  }
```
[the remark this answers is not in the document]
mmitiche: we can of course but I don't see what this would add comparing to a getter?
for me they are exactly the same with one small advantage  for the  getter which is that it is more easy to notice when reading the code.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -171,8 +233,14 @@ export default class QuanticSearchBox extends LightningElement {
    */
   selectSuggestion = (event) => {
     event.stopPropagation();
-    const selectedSuggestion = event.detail.selectedSuggestion;
-    this.searchBox?.selectSuggestion(selectedSuggestion);
+    const {value, isRecentQuery} = event.detail;
+    if (isRecentQuery) {
+      this.recentQueriesList.executeRecentQuery(
+        this.recentQueries.indexOf(value)
+      );
```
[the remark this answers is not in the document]
mmitiche: since the QuanticSearchBox received an event for a recent query selection,  `this.recentQueries` could not be undefined. so no it cant be undefined.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -92,6 +92,19 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
+   * @type {String[]}
+   */
+  @api recentQueries;
+  /**
+   * The maximum number of suggestions to display.
+   * @api
+   * @type {number}
+   * @defaultValue 7
+   */
+  @api maxNumberOfSuggestions;
```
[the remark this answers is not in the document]
mmitiche: true!

Btw what do you think of this 7 default value? is it fine?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -87,26 +124,111 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+        icon: option.isRecentQuery ? 'utility:clock' : 'utility:search',
+        clickHandler: () => {
+          this.handleSuggestionSelection(option.rawValue, option.isRecentQuery);
+        },
+      }));
+      return [
+        {
+          key: 0,
+          isSelected: this.selectionIndex === 0,
+          isClearRecentQueryButton: true,
```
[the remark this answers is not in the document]
mmitiche: Mainly because the Quantic Search Box Suggestions List has the logic to control what suggestion to highlight when clicking the up and down arrow of the keyboard. also to select a suggestion using the keyboard. and all this rely on indexes.

so if we put  "clearRecentQueries" buttons outside the `template:foreach` we will be then obliged to re-work this logic that handles navigating and highlighting the different options when using the keyboard,  and it will be kinda hard to deal with the indexes as all the suggestions will be indexed from 0 to n  but the  "clearRecentQueries" buttons are not being considered in this so we will need to add conditions like the following:

- if index === n and down arrow key is selected, highlight  "clearRecentQueries" buttons and not the suggestion at index 0.
- if index === 0 and up arrow key is selected, highlight  "clearRecentQueries" buttons and not the suggestion at index n.

For that reason and to avoid that, I saw it will be better to include  "clearRecentQueries" buttons in the forEach so all the logic based on the index will continue to work naturally.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.html
mmitiche: answered ☝️

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: It's a good pattern  I will update

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -87,26 +124,111 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
     }
   }
 
-  get suggestionsToRender() {
-    return this.suggestions?.map((s, i) => ({
-      ...s,
-      isSelected: this.selectionIndex === i,
-    }));
+  get allOptions() {
```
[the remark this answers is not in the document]
mmitiche: For your code snippet suggestion I think it will not work properly because the the indexes will be duplicated we will have more than one option with index 0 for example, one for the first recent query and one for the first suggestion.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -87,26 +124,111 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+            (recentQuery) => recentQuery.rawValue === suggestion.rawValue
+          )
+      ) || []
+    );
   }
 
-  emitSuggestionHighlighted() {
-    const highlightChangeEvent = new CustomEvent('highlightchange', {
-      detail: this.suggestions[this.selectionIndex],
+  get filteredRecentQueries() {
```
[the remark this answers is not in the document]
mmitiche: sure!

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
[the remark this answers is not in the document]
SimonMilord: personally I have no problem with 7

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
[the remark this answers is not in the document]
erocheleau: Well what I'm saying is that it might be a better idea to rely on the localStorage more than to rely on the state of Headless for that exact reason.

For example too, I know customers have multiple searchboxes. One on the main page, one inline in the search page as well as one in the header too only visible outside of the home page and search page.

I believe it would make more sense to have the recent queries all show up in these searchboxes no?

Already that the recent queries are not available on the standalone searchbox outside of the search page is a bit worrying. It will generate questions.

I wonder how Atomic does it with a standalone searchbox? Do you know?

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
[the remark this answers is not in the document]
erocheleau: A getter is code executed every time the property is called, so everytime you do `this.localStorageKey` you are creating a function, executing it, evaluating `this.engineId`, etc. to return a string that won't change after initialization.

It's a question of efficiency of code AFAIK.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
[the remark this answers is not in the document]
erocheleau: Seems arbitrary but I don't mind :D

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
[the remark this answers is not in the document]
erocheleau: Ahh yea that makes sense, it does make the code more complex though :(

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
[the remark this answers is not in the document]
erocheleau: Yea my bad, then you need to calculate the `index` at the end 😂 

```
return [
  ...recentQueriesOptions,
  ...querySuggestions
].map((suggestion, index) => ({...suggestion, key: index}).slice(0, this.maxNumberOfSuggestions + 1);
```

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -57,6 +59,13 @@ export default class QuanticSearchBox extends LightningElement {
    * @defaultValue false
    */
   @api textarea = false;
+  /**
+   * Whether to disable rendering the recent queries as suggestions.
+   * @api
+   * @type {boolean}
+   * @defaultValue false
+   */
+  @api disableRecentQuerySuggestions = false;
```
erocheleau, line 68: We're sure we want it as "default: on" right? Is that what Atomic is doing?

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -71,6 +80,10 @@ export default class QuanticSearchBox extends LightningElement {
   suggestions = [];
   /** @type {boolean} */
   hasInitializationError = false;
+  /** @type {RecentQueriesList} */
+  recentQueriesList;
+  /** @type {Array} */
```
erocheleau, line 85: Do you want to defined a type above for this?

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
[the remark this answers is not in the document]
erocheleau: ah interesting ty

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -118,6 +154,10 @@ export default class QuanticSearchBox extends LightningElement {
       'quantic__selectsuggestion',
       this.selectSuggestion
     );
+    this.removeEventListener(
```
erocheleau, line 157: You gave me the same comment on my pr recently :D 

If you check for `if (!this.disableRecentQuerySuggestions)` to add the event listener, why don't you also check to remove the event listener? :D

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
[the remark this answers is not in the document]
erocheleau: I'm wondering if we shouldn't instead have some kind of `uniqueList` of recent queries in localStorage, no matter where they are coming from.

So that any standalone searchbox can also display them.

They would indeed be updated based on this `updateRecentQueriesListState` but the `localStorageKey` could be shared across all Coveo Quantic components in your community for example.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -71,7 +71,7 @@ export default class QuanticSearchBoxInput extends LightningElement {
    */
   @api
   get inputValue() {
-    return this.input.value;
+    return this.input?.value;
```
erocheleau, line 74: Good catch

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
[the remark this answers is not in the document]
erocheleau: We have to simplify this, it's really hard to read.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -87,26 +124,126 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+            (recentQuery) => recentQuery.rawValue === suggestion.rawValue
+          )
+      ) || []
+    );
+  }
+
+  /**
+   * Returns the recent queries that start with the query currently typed by the end user.
+   */
+  get recentQueriesThatStartWithCurrentQuery() {
```
erocheleau, line 186: This should really be in Headless no?

We should be able to pass the current search query and headless already contains the state that has the recent queries, and it should return only the queries that starts with?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -87,26 +124,126 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+        })) || []
+    );
+  }
+
+  /**
+   * Highlights a recent query based on the letters that match the current query.
+   * @param {String} value
+   * @returns {String}
+   */
+  formatRecentQuery(value) {
```
erocheleau, line 207: Please put this in a Utils file, it might be useful elsewhere too no?

Something like `Utils.highlightString(...)`

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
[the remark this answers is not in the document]
erocheleau: Do we REALLY need to have the clear recent queries as part of that though?

Could we perhaps have 

```
searchbox [down]
suggestions [up/down]
clear button [down] <- but displayed as the first element because of flex-ordering for example?
```

As in always have that one at the end instead of being the first one? Would that simplify things?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -87,26 +124,126 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+          length: value.length - this.query.length,
+        },
+      ],
     });
-    this.dispatchEvent(highlightChangeEvent);
+    return highlightedValue;
+  }
+
+  emitSuggestionHighlighted() {
+    if (!(this.shouldDisplayRecentQueries && this.selectionIndex === 0)) {
```
erocheleau, line 223: You see this is where you're moving complexity because you have to account for the fact that the "clear recent queries" is counted in the indices of recent queries.

Would it be acceptable in terms of a11y perspective to have the "clear" be the last element instead of the first? But displayed as the first maybe?

erocheleau (changes requested):
I think we can simplify this further because 

1. the clear button is making this super complicated when it shouldn't be.
2. the merging of the two suggestions is causing issues and making the code very complex as well :(

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
mmitiche: Currently in Headless we have a state/controller for just recent queries, and not for recent queries that will be displayed as suggestions, thus currently in Headless there is no logic to filter out recent queries based on the current query.

It can be a nice improvement for Headless, but for now it is not supported.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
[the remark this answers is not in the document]
mmitiche: Good point 👍

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
mmitiche: Yes in atomic it's by default ON.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
mmitiche: Sure! I will add that in #4120 .

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
mmitiche: I felt like removing wouldn't harm, if it's there it will be removed, if it's not, then it's not there, trying to remove wouldn't cause a problem.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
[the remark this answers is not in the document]
mmitiche: As discussed previously, like you mentioned the only way to support recent queries in the standalone search box to have the `localStorageKey` shared across all Coveo Quantic components and that's how Atomic does it.

In our case, I would challenge this decision because we are likely to have multiple engines running in the same time, in the HIP for example with multiple cases, it can become messy if qll recent queries are shared.

so I think the best thing to do now to separate this concern.
We start by not support recent queries in the standalone search box, and we evaluate the needs/risks of supporting it separately in a second iteration.

PR #4043, opened 2024-05-31
fix: calculate limits based on domain and input values

y-lakhdar:
Ensure the maximum and minimum input limit take into account the domain as well as the other input value

https://coveord.atlassian.net/browse/KIT-3225


fbeaudoincoveo (approved):

packages/atomic/src/components/commerce/facets/facet-number-input/atomic-commerce-facet-number-input.tsx
```diff
@@ -59,8 +59,20 @@ export class FacetNumberInput {
     ]);
   }
 
+  private get minimumInputValue() {
+    const {domain} = this.facet.state;
+    const min = Number.MIN_SAFE_INTEGER;
+    return Math.max(this.start || min, domain?.min || min);
```
olamothe, line 65: I think the falsy check will stumble if `this.start` or `domain.min` is 0.
We should do strict null or undefined check instead (bueno has a utility function for this)

packages/atomic/src/components/commerce/facets/facet-number-input/atomic-commerce-facet-number-input.tsx
```diff
@@ -59,8 +59,20 @@ export class FacetNumberInput {
...
+  }
+
+  private get maximumInputValue() {
+    const {domain} = this.facet.state;
+    const min = Number.MAX_SAFE_INTEGER;
+    return Math.min(this.end || min, domain?.max || min);
```
olamothe, line 71: Same as above.

PR #4045, opened 2024-06-03
chore(atomic): remove unreachable condition

y-lakhdar:
The condition is already present in the FacetGuard component
https://coveord.atlassian.net/browse/KIT-3238


olamothe:
I think the PR is polluted with another branch 🤔

olamothe (approved):

PR #4046, opened 2024-06-03
feat(atomic): add excerpt and product name highlights

y-lakhdar:
Support highlights for `atomic-product-text`.

https://coveord.atlassian.net/browse/KIT-3152

### DIFF
--- a/packages/atomic/src/components/commerce/product-template-components/atomic-product-text/atomic-product-text.tsx
+++ b/packages/atomic/src/components/commerce/product-template-components/atomic-product-text/atomic-product-text.tsx
@@ -69,7 +69,8 @@ export class AtomicProductText
           host={this.host}
           logger={this.bindings.engine.logger}
           defaultValue={this.default}
-          itemValueRaw={productValueAsString}
+          item={this.product}
+          getProperty={ProductTemplatesHelpers.getProductProperty}
         >
           <atomic-commerce-text
             value={getFieldValueCaption(
--- a/packages/atomic/src/components/common/item-text/item-text-fallback.tsx
+++ b/packages/atomic/src/components/common/item-text/item-text-fallback.tsx
@@ -1,20 +1,22 @@
 import {isUndefined} from '@coveo/bueno';
-import {Fragment, FunctionalComponent, h} from '@stencil/core';
+import {Fragment, FunctionalComponent, VNode, h} from '@stencil/core';
 import {possiblyWarnOnBadFieldType} from './field-warning';
 
-export interface ItemTextProps {
+export interface ItemTextProps<T> {
   logger: Pick<Console, 'error'>;
   host: HTMLElement;
   defaultValue: string | undefined;
-  itemValueRaw: unknown;
   field: string;
+  item: T;
+  getProperty: (result: T, property: string) => unknown;
 }
 
-export const ItemTextFallback: FunctionalComponent<ItemTextProps> = (
-  {field, host, logger, defaultValue, itemValueRaw},
-  children
-) => {
-  possiblyWarnOnBadFieldType(field, itemValueRaw, host, logger);
+export const ItemTextFallback = <T,>(
+  {field, host, logger, defaultValue, item, getProperty}: ItemTextProps<T>,
+  children: VNode[]
+): FunctionalComponent<ItemTextProps<T>> | null => {
+  const raw = getProperty(item, field);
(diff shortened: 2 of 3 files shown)


packages/atomic/src/components/commerce/product-template-components/atomic-product-text/atomic-product-text.tsx
```diff
@@ -51,6 +82,11 @@ export class AtomicProductText
     }
   }
 
+  private get fieldForHighlights() {
+    const cleanedField = this.field.replace(/^ec_/, '');
```
Spuffynism, line 86: nit: we _could_ make this function simpler by explicitly making it handle the `ec_name` and `excerpt` fields.

Otherwise, the function gives the impression it can highlight other fields.

Spuffynism (approved):

packages/atomic/src/components/commerce/product-template-components/atomic-product-text/atomic-product-text.tsx
```diff
@@ -33,11 +37,39 @@ export class AtomicProductText
...
+   * When this is set to `true`, the component attempts to highlight text based on the highlighting properties provided by the search API response.
+   * This property only works for the document excerpt and the ec_name field.
+   */
+  @Prop({reflect: true}) public shouldHighlight = true;
   /**
    * The locale key for the text to display when the configured field has no value.
    */
   @Prop({reflect: true}) public default?: string;
 
+  private renderWithHighlights(
```
olamothe, line 50: This could be extracted in a utility function in `/common` (or create functional component(s) for `atomic-result-text` and `atomic-product-text` since there is more overlap for the warning on invalid field). 

 To make sure there is no `@coveo/headless/commerce` import in the `/common` folder, `HighlightUtils.highlightString` can be abstracted into an interface in that utility function.

fpbrault: I think the second option makes more sense here

packages/atomic/src/components/commerce/product-template-components/atomic-product-text/atomic-product-text.tsx
```diff
@@ -51,6 +83,10 @@ export class AtomicProductText
     }
   }
 
+  public isFieldSupportedForHighlighting() {
+    return ['ec_name', 'excerpt'].includes(this.field);
+  }
+
   public render() {
     const productValueAsString = getStringValueFromProductOrNull(
```
fpbrault, line 91: Not sure how I missed that when I made this component but this conditional is useless so you could remove it
```typescript
productValueAsString ?? this.default,
```

packages/atomic/src/pages/examples/commerce-website/search.html
```diff
@@ -94,7 +94,7 @@ <h1>Search page</h1>
                         <atomic-product-price currency="USD"></atomic-product-price>
                       </atomic-product-section-emphasized>
                       <atomic-product-section-description>
-                        <atomic-product-description></atomic-product-description>
+                        <atomic-product-text field="excerpt" class="block text-neutral-dark"></atomic-product-text>
```
fpbrault, line 97: Why not use atomic-product-description and have it use atomic-product-text internally? We lose the expandable description otherwise...

fpbrault (approved):

packages/atomic/src/pages/examples/commerce-website/search.html
fpbrault: I do think it makes a lot of sense to also allow using the excerpt in `atomic-product-description`, good idea!

packages/atomic/src/components/commerce/product-template-components/atomic-product-text/atomic-product-text.tsx
```diff
@@ -33,22 +38,22 @@ export class AtomicProductText
    * This will look in the Product object first, and then in the product.additionalFields object for the fields.
    */
   @Prop({reflect: true}) public field!: string;
+  /**
+   * When this is set to `true`, the component attempts to highlight text based on the highlighting properties provided by the search API response.
+   * This property only works for the document excerpt and the ec_name field.
```
Spuffynism, line 43:
```suggestion
   * This property only works for the product excerpt and the ec_name field.
```

Spuffynism (approved):

PR #4047, opened 2024-06-03
fix(atomic): reset input range on facet clear

y-lakhdar:
https://coveord.atlassian.net/browse/KIT-3240


olamothe (commented):
Humm so we don't tie the range input to any state at all in Headless for commerce ? Why is that 🤔

That's a bit strange, and why we end up with having to handle a stateful property like this in Atomic.

PR #3850, opened 2024-04-24
refactor(atomic): refactor facet-common

olamothe:
Cumulative PR for both (already reviewed) https://github.com/coveo/ui-kit/commit/fb28e0d6d7daf1a85b72c1caf0fab6c03bc578ff and https://github.com/coveo/ui-kit/commit/fb28e0d6d7daf1a85b72c1caf0fab6c03bc578ff

This time, targeting master.

https://coveord.atlassian.net/browse/KIT-3031

PR #4052, opened 2024-06-05
fix(atomic) Fix dependent facet regression

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3242

The regression was introduced in https://github.com/coveo/ui-kit/pull/3850.

`atomic-insight-timeframe-facet` and  `atomic-timeframe-facet` were not affected because they use the `timeframeFacetCommon`'s `disconnectedCallback` which has a guard to return when `this.props.host.isConnected` is `true`

Similarly, `atomic-color-facet`, `atomic-category-facet`, `atomic-rating-facet`, and `atomic-rating-range-facet` were not affected by the refactoring because they didn't rely on the `facetCommon`'s `disconnectedCallback` in the first place, and their respective `disconnectedCallback` methods already had a proper guard clause in place.

### DIFF
--- a/packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
+++ b/packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
@@ -182,6 +182,9 @@ export class AtomicInsightFacet
   }
 
   public disconnectedCallback() {
+    if (this.host.isConnected) {
+      return;
+    }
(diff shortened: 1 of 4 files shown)


olamothe (approved):

fbeaudoincoveo:
Created https://coveord.atlassian.net/browse/KIT-3292 to improve the e2e tests.

PR #4053, opened 2024-06-05
feat(headless,atomic): fix issues for commerce did you mean + handle query trigger for commerce

olamothe:
Add a new `search-actions-thunk-processor` file for commerce search actions (which mimics what we have for search use case)

It is in charge of processing the search response specifically (not product listing) since there can be a mismatch between what Headless think was the query executed (what the user submitted in the search box) and what was actually executed in the backend (query correction/query trigger)

When there is a mismatch, we need to reconcile the search boxes (query set slice) and the query itself (query slice) by dispatching actions. Otherwise it causes bug where the end user will see that did you mean component says "query `foo` was executed" but the search box will display `bar`.

Since it can be a complicated piece of code to push into the thunk function itself, the `processor` file is there to isolate and make it easier to test the behaviour.

https://coveord.atlassian.net/browse/KIT-3224


Spuffynism (approved):

fbeaudoincoveo (approved):

fbeaudoincoveo:
I think it could be theoretically possible to get in an infinite loop with the combination of automatic query correction + query trigger.

E.g.,

1. "foo" query is executed
2. did you mean corrects query to "bar" and automatically executes
3. configured query trigger replaces "bar" with "foo" and frontend executes execute, looping back to 1.

This is a total edge case, but maybe we should consider protecting ourselves against this if this is not already done.

louis-bompart (approved):

PR #4055, opened 2024-06-05
fix(atomic): no break word for product-price + truncate/wrap

olamothe:
JIRA ticket describes the problem pretty well.

Truncate  + no word break on very small screen: 

image

https://coveord.atlassian.net/browse/KIT-3232


fpbrault (approved):
Isn't it a problem for the users if the price gets truncated?

louis-bompart (approved):

PR #4056, opened 2024-06-05
feat(atomic): add alt text field on atomic-result-image

fpbrault:
This PR adds a field on atomic-result-image that will be used as the value of the alt text, similar to atomic-product-image.

https://coveord.atlassian.net/browse/KIT-3243


packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -27,6 +27,13 @@ export class AtomicResultImage implements InitializableComponent {
    */
   @Prop({reflect: true}) field!: string;
 
+  /**
+   * The result field that contains the alt text for the image. This will look for the field in the Result object first, then in the Result.raw object
+   *
+   * If the field is not specified, or does not contain a valid value, the alt text will be set to "Image for {productName}".
+   */
+  @Prop({reflect: true}) imageAltField?: string;
```
fpbrault, line 35: Wondering if I should rename `imagesAltField` to `imageAltField` on atomic-product-image for consistency between these two components

louis-bompart: It would be a breaking change because the attribute is part of the public API.

louis-bompart (approved):

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -42,6 +49,25 @@ export class AtomicResultImage implements InitializableComponent {
     return Array.isArray(value) ? value[0] : value;
   }
 
+  private get altText() {
```
fbeaudoincoveo, line 52: As it is, this will return `any` and I think it should always return `string`.

If the specified field contains anything but a string or a string[] we should return the default alt (and maybe show a warning, although that warning would likely be shown for every result in the page...)

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
```diff
@@ -42,6 +49,25 @@ export class AtomicResultImage implements InitializableComponent {
...
+      if (Array.isArray(value)) {
+        return value[0];
+      }
+
+      if (value !== null) {
+        return value;
+      }
+    }
+
+    return `Image for ${this.result.title}`;
```
fbeaudoincoveo, line 68: We should localize this.

fbeaudoincoveo (changes requested):

packages/atomic/src/components/search/result-template-components/atomic-result-image/atomic-result-image.tsx
fpbrault: Good point

louis-bompart (approved):

PR #4057, opened 2024-06-05
feat(commerce): toggle `capture` from analytics configuration

Spuffynism:
~Allow toggling analytics capture through the `capture` property. It's true by default and is configurable through the engine configuration (as part of the context), or through the context controller.~ Instead of this approach, we forward the `capture` property from the `configuration.analytics.enabled` property.

[CAPI-1003]

[CAPI-1003]: https://coveord.atlassian.net/browse/CAPI-1003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/api/commerce/commerce-api-client.test.ts
```diff
@@ -46,6 +46,7 @@ describe('commerce api client', () => {
         url: '',
```
samisayegh, line 46: The current approach works, but it's not consistent with the other engines, where the value of `capture` is set based on whether analytics are enabled or not. Is there a need to have direct control over the capture flag with commerce engine, or could we use a similar pattern?

Spuffynism: Doh! That's a much better approach. I don't think there's a need for direct control of the `capture` property. I'll make the change.

Spuffynism:
@louis-bompart @samisayegh I documented the tracking id in https://github.com/coveo/ui-kit/pull/4057/commits/25be6e6c9f53afdf5a3ff6c141540f24de01d5f3, since we had some "TBD" comments hanging around. I copied over and adapted the catalog service's definition. Let me know if that's ok.

samisayegh (approved):
Nice ✅

packages/headless/src/app/engine-configuration.ts
```diff
@@ -156,7 +156,7 @@ export interface AnalyticsConfiguration {
    */
   documentLocation?: string;
   /**
-   * TBD
+   * Specifies the unique identifier of the tracking target.
```
fbeaudoincoveo, line 159:
```suggestion
   * The unique identifier of the tracking target.
```

packages/headless/src/features/configuration/configuration-actions.ts
```diff
@@ -153,7 +153,7 @@ export interface UpdateAnalyticsConfigurationActionCreatorPayload {
    */
   documentLocation?: string;
   /**
-   * TBD
+   * Specifies the unique identifier of the tracking target.
```
fbeaudoincoveo, line 156:
```suggestion
   * The unique identifier of the tracking target.
```

packages/headless/src/features/configuration/configuration-state.ts
```diff
@@ -128,7 +128,7 @@ export interface AnalyticsState {
    */
   documentLocation: string;
   /**
-   * TBD trackingId
+   * Specifies the unique identifier of the tracking target.
```
fbeaudoincoveo, line 131:
```suggestion
   * The unique identifier of the tracking target.
```

packages/headless/src/features/commerce/common/actions.test.ts
```diff
@@ -189,6 +190,16 @@ describe('commerce common actions', () => {
       });
     });
 
+    it('sets the capture property from the analytics configuration', async () => {
+      const request = await Actions.buildCommerceAPIRequest(state);
+
+      expect(mockedBuildBaseCommerceAPIRequest).toHaveBeenCalledWith(state);
+
+      expect(request.context.capture).toEqual(
```
fbeaudoincoveo, line 198: Since it defaults to true in the initial state, I think we should test that it's correctly updated if you set it to false.

fbeaudoincoveo (approved):

packages/headless/src/features/commerce/common/actions.test.ts
Spuffynism: Done in https://github.com/coveo/ui-kit/pull/4057/commits/23d27b5c73e0db89db75dd6a021ee038f4eb9e1b

PR #4058, opened 2024-06-05
fix(atomic): fix grid system for commerce when no results or error

olamothe:
Grid system (search-layout/commerce-layout) relies on specific css classes being toggled on the container interface when there are no products/results or when there is an error, in order to align the columns properly.

https://coveord.atlassian.net/browse/KIT-3239

### DIFF
--- a/packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
+++ b/packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
@@ -14,6 +14,10 @@ import {
   ProductListing,
   Context,
   buildContext,
+  buildSearchSummary,
+  buildListingSummary,
+  SearchSummary,
+  ListingSummary,
 } from '@coveo/headless/commerce';
 import {
   Component,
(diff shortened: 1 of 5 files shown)


packages/atomic/src/components/commerce/atomic-commerce-layout/commerce-layout.ts
```diff
@@ -1,5 +1,11 @@
 import {buildSearchLayoutCommon} from '../../common/atomic-layout-section/search-layout';
 
+export const layoutWebComponentTagName = 'atomic-commerce-layout';
+export const containerWebComponentTagName = 'atomic-commerce-interface';
```
fbeaudoincoveo, line 4: This makes me realize: now that we have a dedicated `atomic-commerce-recommendation-interface`, we need to make some adjustments to make sure it also supports the `atomic-commerce-layout` (or create a dedicated `atomic-commerce-recommendation-layout` component).

fbeaudoincoveo (approved):

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
```diff
@@ -365,29 +377,54 @@ export class AtomicCommerceInterface
...
         this.store.hasLoadingFlag(FirstSearchExecutedFlag)
       ) {
         this.store.unsetLoadingFlag(FirstSearchExecutedFlag);
       }
     });
   }
 
+  private initSummary() {
+    this.summary =
+      this.type === 'product-listing'
```
louis-bompart, line 397: the type smell like "magic string", should we refactor it?

louis-bompart (approved):

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
olamothe: `this.type` is strongly typed. 
You can't do `if this.type=='foo'`, TypeScript will complain.

So I don't think there's anything to refactor.

PR #4059, opened 2024-06-05
chore: fix several release issues

louis-bompart:
A good bundle of 3 issues.

## The NXpocalypse

Originally, my removal of the plugin caused that, but I still think the plugin was a bit of a code-smell hiding some other issue, such as the following one.
For some obscure reason, I did not succeed in deciphering `release:phase2`(before my changes), which systematically refuses to run its dependents, even when they are defined.
So, I scrapped it and replaced it with a single task, `release:phase1` (after my changes). It does the same thing, but it's a bit simpler.

## Whoopsie daisy in https://github.com/coveo/semantic-monorepo-tools/commit/916471193fafd59c82472af562a91dd1042b35d8

So, fun fact: The command works as intended on Windows and returns the correct stuff.
On Linux/macOS, the double quotes are not interpreted nicely or the same way, and every commit ends with `format:`.
That ain't a good semantic commit message, so since we updated https://github.com/coveo/ui-kit/commit/eb3542e592888dd14196bbcbec230a58cb6e7d86, every release has been a patch...

Will be fixed in coveo/semantic-monorepo-tools#240, but we can remove the patch when we update, it'll be about the same.

## Reify choke on packages installing themselves as a dependency.

It's probably a way to make it work, but I decided to replace the version by ' file:.`. It's a dev dep, so it has no impact on the consumers and saves me some headaches.

## Was it tested?

Yes: I used some admin magic to be able to run the Prerelease enviro on this branch until it worked. This is now revoked/back to normal.


packages/atomic/package.json
```diff
@@ -67,7 +65,7 @@
   "devDependencies": {
     "@axe-core/playwright": "4.9.1",
     "@babel/core": "7.24.5",
-    "@coveo/atomic": "2.65.5",
+    "@coveo/atomic": "file:.",
```
fbeaudoincoveo, line 68: Makes me wonder, why do we have a dev dependency on @coveo/atomic in the atomic package.json?

patches/@coveo+semantic-monorepo-tools+2.4.31.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@coveo/semantic-monorepo-tools/dist/git/getCommits.js b/node_modules/@coveo/semantic-monorepo-tools/dist/git/getCommits.js
+index 302c25a..1bbd782 100644
+--- a/node_modules/@coveo/semantic-monorepo-tools/dist/git/getCommits.js
++++ b/node_modules/@coveo/semantic-monorepo-tools/dist/git/getCommits.js
+@@ -13,7 +13,7 @@ export default async function (projectPath, from, to = "HEAD") {
+     const delimiter = `<--- ${randomBytes(64).toString("hex")} --->`;
+     const gitParams = [
+         "log",
+-        `--pretty="format:%B%n-hash-%n%H ${delimiter}"`,
++        `--format=%B%n-hash-%n%H%n${delimiter}`,
```
fbeaudoincoveo, line 10: Will we still need a patch after https://github.com/coveo/semantic-monorepo-tools/pull/240 is merged?

fbeaudoincoveo (approved):

internal-docs/release-process.md
```diff
@@ -98,12 +94,18 @@ This task will create a new "version bump" commit, which will contain:
...
+> [!WARNING]
+> Currently disabled.
 
-This task will do the same thing as `release:phase2`, but is intended to publish the NPM package to GitHub Packages instead.
+This task will do the same thing as `release:phase1`, sub-phase 3, but is intended to publish the NPM package to GitHub Packages instead.
 
 # Deploying
 
+> [!NOTE]
+> Probably out-of-date
```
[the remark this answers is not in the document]
louis-bompart: We should update it, but right now I want to fix the release process post haste 

Also the diagram is not related to this code change

.github/workflows/masterbot.yml
```diff
@@ -206,7 +206,8 @@ jobs:
       - name: Release
         run: |
           npm run nx:graph
-          npm run release:phase1-3
+          npm run release:phase1
+          npm run release:phase2
```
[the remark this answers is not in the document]
louis-bompart: Yes 👍 
Although mind the new releases process where the phases numeration is a but shifted

packages/atomic/package.json
louis-bompart: Yes, we do for the tests/storybook

patches/@coveo+semantic-monorepo-tools+2.4.31.patch
louis-bompart: No, just to update

PR #4061, opened 2024-06-06
test(commerce): add tests for atomic-commerce-text

Spuffynism:
TODO:
- [x] Test for pluralization

https://coveord.atlassian.net/browse/KIT-3255


packages/atomic/src/components/commerce/atomic-commerce-text/e2e/page-object.ts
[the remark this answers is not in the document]
alexprudhomme: This cant really be used here since the dom is only made of an <atomic-commerce-text>.

packages/atomic/src/components/commerce/atomic-commerce-text/atomic-commerce-text.tsx
```diff
@@ -32,6 +32,7 @@ export class AtomicCommerceText
   @Prop({reflect: true}) public value!: string;
   /**
    * The count value used for plurals.
+   * @type {number}
```
alexprudhomme, line 35: This is necessary for the url manager of storybook to work 🤔. We will have to add jsdoc to every prop.

louis-bompart: Good JSDOC ain't a bad thin :) 
It helps with Intellisense
Custom Element Manifest may also be packed up to help.

tldr: good jsdoc is good.

louis-bompart (approved):

PR #4062, opened 2024-06-06
Feat svcc 3827 knowledge headless controller

dmgauthier:
# Controller using the knowledge-api-client

## Summary

Refactored the current generatedAnswer controllers to maximize re-usability between the current stream pattern using a stream ID from the searchAPI and the new AnswerAPI streaming directly from a query. 

### headless-core-generated-answer controller

Removed the logic listening to the search to make it more generic. All the state related to the UI component management is left untouched.

### headless-searchapi-generated-answer controller

Is now the main controller used by all the implementation prior to the knowledgeAPI stream. And whenever the interface is not configured to use the new client, it will still use the `headless-searchapi-generated-answer` controller.

### headless-knowledge-generated-answer controller

Its the controller used to connect the new client stream with the `generated-answer` state. I decided to keep the same state slice that was already used by the atomic and quantic implementation. I only mapped the required state keys from the client to the controller state with listeners dispatching the right state updating action when required.

### Testing

I decided to duplicate a lot of test because I did not want any change made to the `headless-core-generated-answer` controller to affect the other controller extending it. 

I did not find a way to properly test what is happening inside the listener, unfortunately


dmgauthier:
This is now obsolete

PR #4068, opened 2024-06-10
feat: update search action cause to browseResults in insight-search and pagination

louis-bompart:
The search action cause `pagerScrolling` has been updated to `browseResults` in the `insight-search` and `pagination` features. This change ensures consistent logging of user navigation actions.

https://coveord.atlassian.net/browse/KIT-2983

### DIFF
--- a/packages/headless/src/controllers/insight/pager/headless-insight-pager.ts
+++ b/packages/headless/src/controllers/insight/pager/headless-insight-pager.ts
@@ -1,10 +1,6 @@
 import {InsightEngine} from '../../../app/insight-engine/insight-engine';
 import {fetchPage} from '../../../features/insight-search/insight-search-actions';
-import {
-  pagerNext,
-  pagerNumber,
-  pagerPrevious,
-} from '../../../features/pagination/pagination-analytics-actions';
+import {browseResults} from '../../../features/pagination/pagination-analytics-actions';
 import {
   logPageNumber,
   logPageNext,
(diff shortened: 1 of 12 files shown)


packages/headless/src/features/pagination/pagination-insight-analytics-actions.ts
louis-bompart, line 1: @mmitiche not sure abt this

packages/headless/src/controllers/insight/pager/headless-insight-pager.ts
```diff
@@ -1,10 +1,6 @@
 import {InsightEngine} from '../../../app/insight-engine/insight-engine';
```
samisayegh, line 1: Could we also adjust `pagerNext` and similar action causes? We see this error quite often on [Barca](https://platformdev.cloud.coveo.com/admin/#/barcasportsmcy01fvu/usage/health/event-problems)

image

samisayegh (approved):

packages/headless/src/controllers/insight/pager/headless-insight-pager.ts
louis-bompart: oops, missed a file.
Fixed in cc6834e

olamothe (approved):

mmitiche (approved):
LGTM 👍

This single line will need to be updated accordingly to allow the Quantic E2E tests to pass successfully: https://github.com/coveo/ui-kit/blob/35e151707c87fe36d0168d5d6d2d7b0f49bd12f7/packages/quantic/cypress/e2e/default-1/examples/insight-panel/insight-panel.cypress.ts#L90

packages/headless/src/features/pagination/pagination-insight-analytics-actions.ts
mmitiche, line 1: The changes in this file need to reverted as the analytics actions in this file are only dispatched and used when using legacy analytics, and we do not wanna update the action cause in that case, so these needs to remain `SearchPageEvents.pagerNumber`, `SearchPageEvents.pagerNext` and `SearchPageEvents.pagerPrevious` respectively.

PR #4069, opened 2024-06-10
feat(headless): improve Generate Answer documentation

lbergeron:
[SVCC-3877](https://coveord.atlassian.net/browse/SVCC-3877)

This PR clarifies how to set the `contentFormat` attribute for the Generated Answer controller.

[SVCC-3877]: https://coveord.atlassian.net/browse/SVCC-3877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/features/generated-answer/generated-response-format.ts
```diff
@@ -23,6 +23,8 @@ export interface GeneratedResponseFormat {
   answerStyle: GeneratedAnswerStyle;
   /**
    * The content formats that are supported for rendering the answer. If no values are provided, `text/plain` is used.
+   * The answer is generated using the first format supported by the front-end component, from left to right.
```
fbeaudoincoveo, line 26: This seems a tad silly since there are only two supported formats at the moment, one of which will, if I understand correctly, always be supported 😅 

In essence, you can/should either:
1. leave this empty, or
2. pass `['text/markdown', 'text/plain']`

because:
- passing `['text/plain', 'text/markdown']` is silly (`text/plain` will always be used)
- passing `['text/markdown']` is not recommended (you should add the fallback)
- passing `['text/plain']` is useless (it's the same as not passing anything)

Still, granted that we may eventually add support for additional formats, I understand why using an array makes sense.

That being said, this makes me wonder: instead of recommending to always include `text/plain` as the last value in the array as a fallback, should we just fallback to `text/plain` regardless?

packages/headless/src/features/generated-answer/generated-response-format.ts
```diff
@@ -23,6 +23,8 @@ export interface GeneratedResponseFormat {
...
+   * The request fails when no matching format is found. It is recommended to include `text/plain` as the last value in the array to act as a fallback format.
```
fbeaudoincoveo, line 27: > when no matching format is found

found by what?

lbergeron: It is the first match between formats supported by the front-end component, and formats supported by the model.

packages/headless/src/features/generated-answer/generated-response-format.ts
lbergeron: The `contentFormat` attribute is not meant to request a specific format. It represents the capabilities of the front-end component. It is the platform that decides which format to return.

It was implemented that way so the feature rollout could be controlled from the platform.

packages/headless/src/features/generated-answer/generated-response-format.ts
fbeaudoincoveo: Sorry for the super late reply, @lbergeron !

How about:

> The request fails when none of the specified formats are supported by the front-end component or by the model.

packages/headless/src/features/generated-answer/generated-response-format.ts
fbeaudoincoveo: It still seems to me like it would be less error-prone to fallback to `text/plain` even if it's not specified as the last value in the array.

As it is, `contentFormat` is optional, and yet, if I understand correctly, the requests will fail when no value is specified. Given that we have a  strong opinion of what he fallback format should be (i.e., `It is recommended to include `text/plain` as the last value in the array to act as a fallback format`), I think we might as well just fallback to this value, and explain the behavior in the documentation.

Unless there is a reason why someone would actually _want_ requests to fail... but this seem dubious to me.

lbergeron:
Closing this PR and will open a new one to add the automatic fallback to `text/plain` and update the documentation.

PR #4071, opened 2024-06-10
test(atomic): start tests on atomic-commerce-facets 

olamothe:
Mostly: Add new `base-page-object` to inherit from for each tests, which should simplify/streamline the "load this story for this component with these arguments in the URL" repetitive pattern.

https://coveord.atlassian.net/browse/KIT-3256


packages/atomic/playwright-utils/base-page-object.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  async load(args?: Component, story: string = 'default') {
+    if (args) {
+      await this.page.goto(
+        `${this.urlRoot}?id=${this.tag}--${story}&args=${buildArgsParam(undefined, this.camelToKebab(args))}`
+      );
+    } else {
+      await this.page.goto(`${this.urlRoot}?id=${this.tag}--${story}`);
+    }
+    await this.hydrated.waitFor();
```
louis-bompart, line 30: That's my opinion, but I see the "wait on hydration" as an escape hatch and an escape hatch only: I think it should be used when we have no other means of awaiting, not as a "standard."

Hydration is an internal concept and does not reflect "what the user sees and does" like the other locators do.
I would remove:
```suggestion
```

And add it as needed, not everywhere.

louis-bompart (commented):
I'm not fond of those tests/utils if I'm perfectly honest.
I think we should try to adhere more closely to the modern test philosophy shared across all modern test frameworks, which entails testing user-visible behaviour.
https://playwright.dev/docs/best-practices#test-user-visible-behavior
https://testing-library.com/docs/guiding-principles


What does that mean for this PR?
1. `packages/atomic/src/components/commerce/facets/atomic-commerce-facets/e2e/page-object.ts` should have locators built around the facet labels
2. The utils should make lighter uses of the hydrated selector.

packages/atomic/playwright-utils/base-page-object.ts
```diff
@@ -0,0 +1,43 @@
+import type {Page} from '@playwright/test';
+import {buildArgsParam} from '@storybook/router';
+import {JSX} from '../dist/types/components';
```
louis-bompart, line 3: I'm still uncertain we'll be able to uses this everywhere :tm: and that the JSX will map always to what's available in Storybook, but we'll see!

olamothe: I tried to use `custom-elements.json` thingy. 

Main problem is that it's a (weakly typed) JSON, and so we end up with way more complex Typescript in this file to just try to have have some type/interface to represent "an actual Atomic web component available with its tag name and its prop". The end goal is just to get intellisense completion when writing a new test.

And for now in test, we don't need to configure slots, JSDoc, events, shadow parts, etc. Nor did we need to with the older Cypress tests: just components properties. So it just seems like over engineering.

Furthermore: Nothing stop us from writing a `page.goto(....)` manually if absolutely necessary.

louis-bompart: > And for now in test, we don't need to configure slots, JSDoc, events, shadow parts, etc. Nor did we need to with the older Cypress tests: just components properties. So it just seems like over engineering.

I'm pretty sure the slot will be required. It's been used in the Cypress test using the `appendChild` (i.e., look for `appendChild` in `cypress`; most references are "slot" in disguise).

However, most slots could/should be tested by adding a new story.
We can cross that road if/when we reach it.

PR #4072, opened 2024-06-10
fix(atomic): clear recent searches on keyboard enter keypress

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3234


packages/atomic/src/components/common/suggestions/suggestion-manager.ts
```diff
@@ -105,7 +105,7 @@ export class SuggestionManager<SearchBoxController> {
   }
 
   public onSubmit() {
-    this.updateActiveDescendant();
+    this.clickOnActiveElement();
```
fbeaudoincoveo, line 108: `this.clickOnActiveElement` calls `this.updateActiveDescendant`

olamothe (commented):
Code change seems fine after testing locally. 
It means we are missing a cypress test for this though.

It should be done by default when fixing any kind of bugs, and only an exception to not add one (ie: very hard to reproduce/setup/automate), with a PR explanation as to why it was not added !

PR #4073, opened 2024-06-10
fix(atomic): remove boundary validation on numeric inputs

y-lakhdar:
The boundary limit considers only the domain returned by the API and the input value. However, if the product price is determined by multiple fields (e.g., ec_price, ec_promo_price, ec_something_price, etc.), the boundary validation will fail (see image below).

To prevent this issue, I removed the validation that checks against the domain range returned by the API.

image

https://coveord.atlassian.net/browse/KIT-3301

### DIFF
--- a/packages/atomic/src/components/commerce/facets/facet-number-input/atomic-commerce-facet-number-input.tsx
+++ b/packages/atomic/src/components/commerce/facets/facet-number-input/atomic-commerce-facet-number-input.tsx
@@ -56,8 +56,14 @@ export class FacetNumberInput {
     ]);
   }
 
-  private get minimumInputValue() {
-    return isUndefined(this.start) ? Number.MIN_SAFE_INTEGER : this.start;
+  private get absoluteMinimum(): number {
+    const {field} = this.facet.state;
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/commerce/facets/facet-number-input/atomic-commerce-facet-number-input.tsx
```diff
@@ -57,18 +57,11 @@ export class FacetNumberInput {
   }
 
   private get minimumInputValue() {
-    const {domain} = this.facet.state;
-    const ensureValue = (num?: number) =>
-      isUndefined(num) ? Number.MIN_SAFE_INTEGER : num;
-    return Math.max(ensureValue(this.start), ensureValue(domain?.min));
+    return isUndefined(this.start) ? Number.MIN_SAFE_INTEGER : this.start;
```
olamothe, line 60: Negative price is weird.... 
Could check if `this.field === ec_price || ec_promo_price` and if so, set minimum to 0 instead of `Number.MIN_SAFE_INTEGER`

olamothe (approved):

PR #4077, opened 2024-06-11
fix(quantic): fix unexpected back browser button behaviour with standalone search box

mmitiche:
## [SFINT-5539](https://coveord.atlassian.net/browse/SFINT-5539)

There were mainly two issues(issues recorded from the staging package): 

## Issue 01:

When doing an empty search from the standalone search box, clicking the back button one time would not take you to the home page:

https://github.com/coveo/ui-kit/assets/86681870/08461748-bd25-4f0b-8fac-d9592c9f6b54

To solve this issue, I'm simply making the standalone url redirect to  `redirectUrl` instead of `redirectUrl?#={query}` when the query is empty.

https://github.com/coveo/ui-kit/assets/86681870/e774ffb7-8d12-4415-b208-36f771174269

## Issue 02:

After doing multiple queries and clicking the back button several times, there re situations where SF will disconnect the component and connect it again, disconnecting the component causes the removal for the event listener of hash change which makes the component completely loose control of the changes happening in the URL:
 
https://github.com/coveo/ui-kit/assets/86681870/0d0a5ebf-10f3-4409-a4ae-5adba07e186b

To solve this issue, I'm setting `this.initialized` to `false` in the `QuanticSearchInterface` component to make sure we redo the initialization when the component disconnects and reconnect again, the reinitialization would allow adding back the event listener to hash changes so our component will not loose track.
Additionally when the component get's connected after being  disconnected I'm calling the `synchronize` method to make sure the search parameters are always up to date with what we have in the hash.

https://github.com/coveo/ui-kit/assets/86681870/7a0ef4f5-4aef-40d8-b945-493f7151a061

### Try it yourself: https://momentum-force-3625-dev-ed.scratch.my.site.com/s/

[SFINT-5539]: https://coveord.atlassian.net/browse/SFINT-5539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -174,6 +175,9 @@ export default class QuanticSearchInterface extends LightningElement {
     this.urlManager = CoveoHeadless.buildUrlManager(engine, {
       initialState: {fragment: this.fragment},
     });
+    // Make sure sure to re-synchronize the search interface when the component gets discoconnected and reconnected again.
```
mmitiche, line 178:
```suggestion
    // Make sure to re-synchronize the search interface when the component gets disconnected and reconnected again.
```

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -125,6 +125,7 @@ export default class QuanticSearchInterface extends LightningElement {
   }
 
   disconnectedCallback() {
+    this.initialized = false;
```
SimonMilord, line 128: Ok so just to confirm, it was not calling the connectedCallback in `quanticSearchInterface` because it was calling the connectedCallback in the `HostedSearchInterface` which extends the `quanticSearchInterface` ?

And this works because in the `HostedSearchInterface` intialize method we call:


`  initializeHostedInterface = (engine) => {
    super.initialize(engine);
  };`
which calls the initialize in `quanticSearchInterface` and would do:



`    if (this.initialized) {
      return;
    }` 

and just return instead of initializing properly correct?

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
```diff
@@ -241,7 +241,7 @@ export default class QuanticStandaloneSearchBox extends NavigationMixin(
       {
         type: 'standard__webPage',
         attributes: {
-          url: `${this.redirectUrl}#q=${encodeURIComponent(value)}`,
+          url: `${this.redirectUrl}${value ? `#q=${encodeURIComponent(value)}` : ''}`,
```
SimonMilord, line 244: @erocheleau at least we had this part 😄

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
```diff
@@ -174,6 +175,9 @@ export default class QuanticSearchInterface extends LightningElement {
     this.urlManager = CoveoHeadless.buildUrlManager(engine, {
       initialState: {fragment: this.fragment},
     });
+    // Make sure to re-synchronize the search interface when the component gets disconnected and reconnected again.
+    this.urlManager.synchronize(this.fragment);
```
SimonMilord, line 179: yeah I tried adding the synchronize here, but I was missing the `this.initialized = false` in disconnectedCallback for this to work properly

SimonMilord (approved):
It seem to work well, I tried multiple "edge cases" with back/forward buttons, empty queries, etc.

Nice catch on the `this.initialized=false` overlooked that

SimonMilord (approved):

packages/quantic/force-app/main/default/lwc/quanticSearchInterface/quanticSearchInterface.js
mmitiche: Exactly yeah, the  `connectedCallback` of `quanticSearchnterface` is not called because this method, the `connectedCallback` of the `quanticSearchInterface` is being overridden by the   the `connectedCallback` of `hostedSearchInterface`.

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
erocheleau: haha

erocheleau (approved):
Looks good, I'll trust you on it I didn't test it personally.

Maybe a test for it?

mmitiche:
> Looks good, I'll trust you on it I didn't test it personally.
> 
> Maybe a test for it?

Since currently we are not equipped to properly be able to test that in Quantic, I think it would be best to test this in the C4SF package where we are equipped with an env that allow to test this which a standalone search box that redirects to a search page where a quantic Search interface is used.

SimonMilord:
> > Looks good, I'll trust you on it I didn't test it personally.
> > Maybe a test for it?
> 
> Since currently we are not equipped to properly be able to test that in Quantic, I think it would be best to test this in the C4SF package where we are equipped with an env that allow to test this which a standalone search box that redirects to a search page where a quantic Search interface is used.

Yeah opened the PR to add the E2E tests for that

PR #4078, opened 2024-06-11
fix(quantic): issue occurring with result title and quick view in salesforce service console addressed

mmitiche:
### The issue: 

An issue was occurring in the Service Console when trying to open a result or its  Quickview:

https://github.com/coveo/ui-kit/assets/86681870/7d3048d9-61aa-4e1d-a72c-e79fbf9a6e5e

This issue is duo to the Proxy object that Salesforce wraps the result object with, this proxy object is causing issues when the Headless library tries to use the result object.

### The solution:

Instead of using `Object.create(this.result)` that we used to use to create a copy of the object and that apparently is no longer useful, I'm now creating a new object after distructuring  the result object and this is doing the job preventing the errors from occuring:

https://github.com/coveo/ui-kit/assets/86681870/c71ea442-e183-4f04-91af-d2bc8cb71f25

[SFINT-5571]: https://coveord.atlassian.net/browse/SFINT-5571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SimonMilord (approved):
Looks fine to me, did you also test in the quickview?

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -121,7 +121,7 @@ export default class QuanticResultLink extends NavigationMixin(
     this.engine = engine;
     ResultUtils.bindClickEventsOnResult(
       this.engine,
-      this.result,
+      {...this.result, raw: {...this.result.raw}},
```
erocheleau, line 124: Why do we need to de-structure the object here?

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -185,7 +185,9 @@ export default class QuanticResultQuickview extends LightningElement {
     const {pushRecentResult} = this.headless.loadRecentResultsActions(
       this.engine
     );
-    this.engine.dispatch(pushRecentResult(Object.create(this.result)));
+    this.engine.dispatch(
+      pushRecentResult({...this.result, raw: {...this.result.raw}})
+    );
```
erocheleau, line 190: Ooof I am not sure of this solution...

IIRC we want to create a copy of the result object to make sure the recentResult object cannot modify by reference the result itself.

IIRC we started with first doing a `JSON.parse(JSON.stringify(this.result))` but this was causing lag because of Salesforce's Locker too having to run a function on each of the result properties.

So I think this was changed to `Object.create` which is normally an alternative and a better way to "clone" a JS object without keeping the references.

So no the issue is that `Object.create` when used on the Proxy given to us by Salesforce's Locker is causing issues?

I am afraid that de-structuring the result is not enough to completely kill all the references to the Result Object.

Also do we really need the `recentResult` feature in the Service Console? Is it only for the Hosted Search Page? Because I think we don't use it on purpose for the InsightPanel?

This was the previous bug which led us to use the Object.create
https://github.com/coveo/ui-kit/pull/3016

Really not sure this is the fix for this...

erocheleau (commented):
Not certain this is the fix...

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
mmitiche: Because Headless is not dealing correctly with the Proxy  object.
As we know Salesforce converts object to Proxy objects, so with this procedure, the de-structuring, I'm converting the Proxy object back to a normal object.

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
mmitiche: Actually, our goal in the past we have done JSON.parse(JSON.stringify(...)) or Object.Create() is not to delete the references to this object but simply because Headless always find difficulties dealing with Proxy Object instead of a normal Object.
So all these procedures had one goal which is converting the Proxy to a normal Object.

The problem with `JSON.parse(JSON.stringify(...))` is that it became super slow at a certain moment, probably because SF changed how they do their transfer from an object to a proxy.
We replaced it with Object.Create()` that worked well for a while but it's no longer working as expected now, doing for example Object.Create() of a Proxy is now returning an empty object which break the recent queries feature.
and Again I think `Object.Create()` stopped working because SF changed the logic they use to build a Proxy out of an Object.

Finally, de-structuring seems now like the method that is doing the job properly, the job of converting a Proxy to an object, which means that Headless is working as expected and the recent queries are working as expected with no errors that were reported for this issue.

Also, this issue is not happening only in the Service Console, it happens everywhere in Salesforce, even in the Community the same error was noticed and the same recent results issue was noticed.

So basically, that's the logic behind it, it's not about killing the references to a Result object, but it is about avoiding passing a Proxy to Headless.

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
erocheleau: And it works by de-structuring? You were able to reproduce before and can no longer reproduce after?

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
erocheleau: Ok well put a comment in the code describing what to do the next time Salesforce changes something that breaks this new spread ;)

erocheleau (approved):

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
mmitiche: Exactly like you can see in the recordings included in the PR description the error we are trying to solve is the `TypeError isExtensible....`

That's an example of doing console.log(this.result) amd console.log({...this.result, raw: {...this.result.raw}}):
Screenshot 2024-06-12 at 9 48 00 AM

packages/quantic/force-app/main/default/lwc/quanticResultLink/quanticResultLink.js
```diff
@@ -121,7 +121,8 @@ export default class QuanticResultLink extends NavigationMixin(
     this.engine = engine;
     ResultUtils.bindClickEventsOnResult(
       this.engine,
-      this.result,
+      // Destructuring transfers the Proxy object created by Salesforce to a normal object so no unexpected behaviour will occur with the Headless library.
```
mmitiche, line 124:
```suggestion
      // Destructuring transforms the Proxy object created by Salesforce to a normal object so no unexpected behaviour will occur with the Headless library.
```
```suggestion
      // Destructuring transfers the Proxy object created by Salesforce to a normal object so no unexpected behaviour will occur with the Headless library.
```

packages/quantic/force-app/main/default/lwc/quanticResultQuickview/quanticResultQuickview.js
```diff
@@ -185,7 +185,11 @@ export default class QuanticResultQuickview extends LightningElement {
     const {pushRecentResult} = this.headless.loadRecentResultsActions(
       this.engine
     );
-    this.engine.dispatch(pushRecentResult(Object.create(this.result)));
+
+    // Destructuring transfers the Proxy object created by Salesforce to a normal object so no unexpected behaviour will occur with the Headless library.
```
mmitiche, line 189:
```suggestion
    // Destructuring transforms the Proxy object created by Salesforce to a normal object so no unexpected behaviour will occur with the Headless library.
```

PR #4080, opened 2024-06-11
fix(atomic): fix facet and product list loading state

olamothe:
* Set the `loadingFlag` in `atomic-commerce-interface`: We need this to be set, otherwise components wrongly assume that "everything is loaded and ready to go". For end user, this fixes an issue with `atomic-product-list` removing placeholders too quickly without anything to render except a big blank empty state.

* Use a property `numberOfPlaceholders` to decide the number of placeholder to display for the product list: using the number of products in headless state means its always 0, because the products are still being loaded/not yet returned from the backend yet. Making the placeholder useless otherwise.

* Add a `<FacetPlaceholder />` display in `atomic-commerce-facets` while the app is being loaded (ie: while the product list is still rendering a placeholder). This reduces the (annoying) layout shifts.

* Remove `random()` usage from `<FacetPlaceholder />`: This would cause weird flicker issue with the placeholder `width` changing on every render loop.

https://coveord.atlassian.net/browse/KIT-3311

### DIFF
--- a/packages/atomic/src/components/commerce/facets/atomic-commerce-facets/atomic-commerce-facets.tsx
+++ b/packages/atomic/src/components/commerce/facets/atomic-commerce-facets/atomic-commerce-facets.tsx
@@ -13,12 +13,13 @@ import {
   ListingSummary,
   SearchSummary,
 } from '@coveo/headless/commerce';
-import {Component, h, Element, Host, State, Prop} from '@stencil/core';
+import {Component, h, Element, State, Prop, Fragment} from '@stencil/core';
 import {
   BindStateToController,
   InitializableComponent,
   InitializeBindings,
 } from '../../../../utils/initialization-utils';
+import {FacetPlaceholder} from '../../../common/facets/facet-placeholder/facet-placeholder';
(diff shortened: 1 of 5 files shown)


louis-bompart (approved):

packages/atomic/src/components/commerce/facets/atomic-commerce-facets/atomic-commerce-facets.tsx
```diff
@@ -35,6 +36,7 @@ import {CommerceBindings as Bindings} from '../../atomic-commerce-interface/atom
 export class AtomicCommerceFacets implements InitializableComponent<Bindings> {
   @InitializeBindings() public bindings!: Bindings;
   public facetGenerator!: FacetGenerator;
+  public summary!: ListingSummary | SearchSummary;
```
fbeaudoincoveo, line 39: Just the heads up, ListingSummary / SearchSummary will soon become sub-controllers:

https://github.com/coveo/ui-kit/pull/4082

fbeaudoincoveo (approved):

PR #4081, opened 2024-06-11
fix(atomic): collapsed answer gradient colour

jpmarceau:
https://coveord.atlassian.net/browse/KIT-3313

Right now, the linear-gradient is hardcoded to white rather than to the background atomic colour. This will create issues when clients modify their background colour to something other than white, like we doc on the docs site when in dark mode.

Screenshot 2024-06-11 at 3 34 27 PM

It doesn't seem possible for clients to change the linear-gradient. Probably because it requires targeting a pseudo-element of a part with a certain class. So, in this PR, I use the atomic-background colour instead.

### DIFF
--- a/packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
+++ b/packages/atomic/src/components/common/generated-answer/styles/generated-answer.pcss
@@ -105,7 +105,7 @@
   }
   &.answer-collapsed:before {
     @apply h-full w-full absolute top-0 left-0 content-[''];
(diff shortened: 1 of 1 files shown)


jpmarceau:
Oh, and if you want to have a look, you can do so with this branch: https://github.com/coveo/ui-kit/compare/KIT-3313-test
Screenshot 2024-06-11 at 3 39 52 PM

olamothe (approved):

PR #4083, opened 2024-06-11
test(atomic): atomic-commerce-pager

y-lakhdar:
https://coveord.atlassian.net/browse/KIT-3250


packages/atomic/src/components/commerce/atomic-commerce-pager/e2e/atomic-commerce-pager.e2e.ts
```diff
@@ -0,0 +1,144 @@
...
+} from './pager-assertions';
+
+test.describe('Default', () => {
+  test.beforeEach(async ({page}) => {
+    await page.goto(
+      'http://localhost:4400/iframe.html?id=atomic-commerce-pager--default&viewMode=story'
+    );
+  });
+
+  assertEnabled('nextButton');
```
olamothe, line 18: ehhh I dont particularly like this pattern we had for the old tests... I'd prefer just "normal" test directly in the file instead of those.

 What does everybody think ?

louis-bompart: We should adhere as much as possible to the practice of the test framework we're using as we do with code formatting with prettier.

Now for this specific issue/question:

My read from Playwright is that refactor and DRY come into three forms:
 - [Page Object](https://playwright.dev/docs/pom)
 - [Fixture ](https://playwright.dev/docs/test-fixtures)
 - [Test parametrization](https://playwright.dev/docs/test-parameterize) 
 
The first two work hand-in-hand for refactoring specific page aspects: they refactor **actions** and **locators** into smaller subsets as we do for our Headless controller. However, they do not refactor/encompass assertions.

The last one wraps and refactors the expected value used by assertions but not assertions.

So, in short, in my opinion
 - Locators should be refactored (more often than not)
 - Actionable may be refactored (if frequently used)
 - `test` and `expect` are barred from cross-file refactoring or being embedded in functions.
 
 > What I find frustrating is navigating through many functions just to get the assertion due to nested logic. However, if we keep assertions straightforward and flat within a separate file, I don't have an issue with it.
 
 Seeing `test` or `describe` as functions from an AST perspective is correct, but to interpret them as such as developers would be a mistake, I think.
 It's best to see a test suite as a novel:
 - It should be a readable piece (preferably pleasingly)
 - `describe` & `it` represents the headings that structure the novel.
 
 
 TL;DR:
 - Thee shall not embed assertions.
 - Thee shall not embed describe or test, with the exception of test parametrization _in a same file_ (think describe.each from Jest)

packages/atomic/src/components/commerce/atomic-commerce-pager/e2e/atomic-commerce-pager.e2e.ts
```diff
@@ -0,0 +1,232 @@
...
+  test('should display a next button', async ({pager}) => {
+    await expect(pager.nextButton).toBeVisible();
+  });
+
+  test.describe('after clicking the button next', () => {
+    test.beforeEach(async ({pager}) => {
+      await pager.nextButton.click();
+    });
+
+    test('pager button 2 should be selected', async ({pager}) => {
```
fpbrault, line 47: Since these are radio buttons, what about just using .toBeChecked() instead of checking the attribute?

packages/atomic/src/components/commerce/atomic-commerce-pager/e2e/atomic-commerce-pager.e2e.ts
```diff
@@ -0,0 +1,232 @@
...
+});
+
+test.describe('with an invalid page in the hash', () => {
+  test.beforeEach(async ({page}) => {
+    await page.goto(
+      'http://localhost:4400/iframe.html?id=atomic-commerce-pager--default&viewMode=story#page=8k3'
+    );
+  });
+
+  test('pager button 9 should be selected', async ({pager}) => {
```
fpbrault, line 180: Seems like on first load, the page fragment gets discarded if it has an invalid value... weird!

packages/atomic/src/components/commerce/atomic-commerce-pager/e2e/atomic-commerce-pager.e2e.ts
```diff
@@ -0,0 +1,232 @@
...
+      'http://localhost:4400/iframe.html?id=atomic-commerce-pager--default&viewMode=story&args=attributes-number-of-pages:3'
+    );
+  });
+
+  test('should display 3 pages', async ({pager}) => {
+    await expect(pager.pages).toHaveCount(3);
+  });
+});
+
+test.describe('with numberOfPages=-5', () => {
```
fpbrault, line 206: Not really related to this component directly, but I noticed that on first load of a commerce search page, with an invalid value for page or number-of-pages, the interface fails to load..!

fpbrault (approved):

packages/atomic/src/components/commerce/atomic-commerce-pager/e2e/atomic-commerce-pager.e2e.ts
fpbrault: Isn't it a bad experience for the user, let's say they use a url that has an invalid value for whatever reason, the search interface will not load and they may just leave thinking the search is broken. I don't think it will happen often but it makes more sense to me to just ignore the invalid value instead of preventing the interface from loading..

PR #4084, opened 2024-06-11
chore: hackpatch storybook to support attributes/css/slot with identical names

louis-bompart:
Essentially: Furl args with their category, unfurl when displaying, well, anything.
CSS shadow parts receive a special treatment to be shorter (css-shadow-parts-next-button-icon was quite a mouthful).


packages/atomic/src/components/search/atomic-pager/atomic-pager.tsx
```diff
@@ -73,7 +73,7 @@ export class AtomicPager implements InitializableComponent {
    * - Use a value that starts with `assets://`, to display an icon from the Atomic package.
    * - Use a stringified SVG to display it directly.
    */
-  @Prop({reflect: true}) previousButtonIcon = ArrowLeftIcon;
+  @Prop({reflect: true}) previousButtonIcon: string = ArrowLeftIcon;
```
louis-bompart, line 76: bonus, this helps set the attribute of the pager more easily

olamothe (approved):

PR #4085, opened 2024-06-12
test(atomic): add tests for atomic-commerce-search-box

fpbrault:
This PR adds more tests for atomic-commerce-search-box

https://coveord.atlassian.net/browse/KIT-3246


packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.new.stories.tsx
```diff
@@ -11,7 +11,7 @@ const {decorator, play} = wrapInCommerceInterface({skipFirstSearch: true});
 
 const meta: Meta = {
   component: 'atomic-commerce-search-box',
-  title: 'Atomic-Commerce/Searchbox',
+  title: 'Atomic-Commerce/Interface Components/atomic-commerce-search-box',
```
olamothe, line 14: Why this name change ?

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/page-object.ts
```diff
@@ -14,14 +14,44 @@ export class AtomicCommerceSearchBoxLocators {
     return this.page.getByRole('combobox', {name: 'Search'});
   }
 
+  get clearButton() {
+    return this.page.getByLabel('Clear', {exact: true});
+  }
+
+  get ariaLive() {
+    return this.page.locator('div[id*="aria-live-"][id$="-search-box"]');
```
olamothe, line 22: Could it be more simply: 
```suggestion
    return this.page.locator('aria-live');
```
?

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/page-object.ts
```diff
@@ -14,14 +14,44 @@ export class AtomicCommerceSearchBoxLocators {
...
+  }
+
+  get clearRecentQueriesButton() {
+    return this.page.getByLabel('Clear recent searches.');
+  }
+
+  get numberOfQueries() {
```
olamothe, line 29: This is a bit flawed since it check the attribute on the search box, not the actual number of suggestion shown to the end user. ie: the attribute/property could be bugged and the selector would be invalid.

How about trying to `getByRole('button')` with some filter on the button attributes to snipe only suggestions ?

packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.new.stories.tsx
fpbrault: Follows the nomenclature in the "Storybook Commerce Hierarchy" document shared by @louis-bompart

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/page-object.ts
fpbrault: This would select the full atomic-aria-live element but what I need is the div specific to the searchbox

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/page-object.ts
fpbrault: I'm using this to validate that the aria-live message contains the smallest value between the count of suggestions+recent queries and the maximum number of suggestions as set on the search box.

packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.new.stories.tsx
```diff
@@ -36,6 +36,14 @@ export const RichSearchBox: Story = {
...
 };
 
+export const StandaloneSearchBox: Story = {
+  name: 'Standalone search box',
+  args: {
+    'attributes-redirection-url':
+      './iframe.html?id=atomic-commerce-search-box--in-page&viewMode=story&args=enable-query-syntax:!true;suggestion-timeout:5000',
+  },
+};
+
```
fpbrault, line 46: @louis-bompart This is what I was talking about, I added the "attributes-" prefix here, but it does not appear on my atomic-commerce-search-box component. If I don't add the prefix, I do see it, but as just "url" instead of "redirection-url"...

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/page-object.ts
```diff
@@ -14,14 +14,44 @@ export class AtomicCommerceSearchBoxLocators {
...
+  }
+
+  recentQueries({
+    index,
+    total,
+    listSide,
+  }: {index?: number; total?: number; listSide?: 'Left' | 'Right'} = {}) {
+    return this.page.getByLabel(
+      new RegExp(
+        `.* recent query\\.(?: Button\\.)? ${index ?? '\\d'} of ${total ?? '\\d'}\\.${this.listSideAffix(listSide)}`
```
fbeaudoincoveo, line 54:
```suggestion
        `recent query\\.(?: Button\\.)? ${index ?? '\\d'} of ${total ?? '\\d'}\\.${this.listSideAffix(listSide)}`
```

fbeaudoincoveo (approved):

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/atomic-commerce-search-box.e2e.ts
```diff
@@ -27,6 +28,136 @@ test.describe('default', () => {
       await expect(searchBox.searchSuggestions().first()).toBeVisible();
     });
 
+    test('should update aria-live message', async ({searchBox}) => {
+      await searchBox.searchSuggestions().first().waitFor({state: 'visible'});
+      await expect(searchBox.ariaLive).toContainText(
+        (await searchBox.searchSuggestions().count()).toString()
```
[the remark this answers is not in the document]
fpbrault: That's right... ty!

PR #4086, opened 2024-06-12
fix(commerce): forward analytics source on `context.source`

Spuffynism:
I also made it so the `requestMetadata` is populated on commerce api requests in https://github.com/coveo/ui-kit/commit/dd3b3698ef001883db7344fbdea4c7c0bd063150#. This enables hooking into requests with `preprocessRequest` on a commerce api `method`, instead of matching on an url path.

[CAPI-1053]

[CAPI-1053]: https://coveord.atlassian.net/browse/CAPI-1053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/headless/src/api/commerce/commerce-api-client.test.ts
+++ b/packages/headless/src/api/commerce/commerce-api-client.test.ts
@@ -98,6 +98,7 @@ describe('commerce api client', () => {
         language: request.language,
         currency: request.currency,
       },
+      requestMetadata: {method: 'listing'},
     });
   });
 
@@ -130,6 +131,7 @@ describe('commerce api client', () => {
         language: request.language,
         currency: request.currency,
       },
+      requestMetadata: {method: 'search'},
     });
   });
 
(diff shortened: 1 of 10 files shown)


packages/headless/src/api/commerce/commerce-api-client.test.ts
```diff
@@ -190,6 +196,7 @@ describe('commerce api client', () => {
         language: request.language,
         currency: request.currency,
       },
+      requestMetadata: {method: 'search/querySuggest'},
```
fbeaudoincoveo, line 199: Shouldn't we also have an it for search/productSuggest?

fbeaudoincoveo (approved):

olamothe (approved):

packages/headless/src/api/commerce/commerce-api-client.test.ts
Spuffynism: Done in https://github.com/coveo/ui-kit/pull/4086/commits/91b6b2b7aed99c128f87a54efb9333f197749002

PR #4088, opened 2024-06-12
feat(commerce): make commerce sub-package generally available

Spuffynism:
Opening as draft because I'd like some advice on the message we include to indicate deprecation of the `product-listing` sub-package. I was thinking of something like:
> @deprecated The `product-listing` sub-package is deprecated. Prefer using the `commerce` sub-package instead.

@anthonydelage @jpmarceau 

Also, is there something else that needs to be done in headless apart from removing the comments?

[CAPI-98]

[CAPI-98]: https://coveord.atlassian.net/browse/CAPI-98?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jpmarceau:
How about: "@deprecated The `product-listing` sub-package is deprecated. Use the `commerce` sub-package instead."

As far as what's required for the new headless controllers, if we want to remove the `beta` comment, I think we'll just have to sync so that I remove analogous warnings on docs.coveo.com starting from the target version 👍

Spuffynism:
I've also deprecated the product recommendations sub-package as it's also replace by the commerce one!

Spuffynism:
Let's keep this as draft until we have remaining [Commerce GA](https://coveord.atlassian.net/browse/CAPI-718) tasks complete.

jpmarceau:
I was curious and tried something I should have tried earlier: generating docs using the `@deprecated` tag. I noticed that it doesn't currently do anything, as far as documentation goes 🙃 It would be nice to add support for the tag, but for now I think we could just use plain text and the deprecation would be clear in the docs.
I still think that having the `@internal` tag can be useful for intellisense though. So, how about having both:
```
Deprecated. Use the commerce sub-package instead of the deprecated product-listing sub-package.
@deprecated
```

Spuffynism:
@jpmarceau did you mean to include `@internal` in your suggestion? I'm a bit confused? 🤔

jpmarceau:
> @jpmarceau did you mean to include `@internal` in your suggestion? I'm a bit confused? 🤔

Yes, my thinking was to include `@internal` on its own line, for intellisense without interfering with docs. If you put the `@internal` tag on the same line as a doc string, that line isn't extracted for docs.

Spuffynism:
@jpmarceau @spiracha-coveo I've swapped `@internal` for `@deprecated` since the docs wouldn't build with `@internal`. I think it's because `@internal` tells the docs to not include the documented item in the doc.

packages/headless/src/api/search/search/product-recommendation.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * Deprecated. The `product-recommendation` sub-package is deprecated. Use the `commerce` sub-package instead.
```
olamothe, line 2: Nitpick, but these should be in this format:
```
@deprecated  The `product-recommendation` sub-package is deprecated. Use the `commerce` sub-package instead.
```

On 1 line instead of 2 lines. It makes it nicer to read in most IDE for implementers.

olamothe (approved):

agillesCoveo (approved):
Give me my 1$!

anthonydelage (approved):
:ketsgo:

packages/headless/src/api/search/search/product-recommendation.ts
Spuffynism: @olamothe I intentionally split this in half to have the string show up in the documentation: https://github.com/coveo/ui-kit/pull/4088#issuecomment-2196794791

Still, I created https://coveord.atlassian.net/browse/CAPI-1124 to address this (cc @jpmarceau, @spiracha-coveo , feel free to transfert this to your board if you think it's more appropriate!)

PR #4091, opened 2024-06-13
fix(atomic): use currency from context instead of prop for atomic-product-price

olamothe:
Remove `currency` as a property from `atomic-product-price`, as it it mandatory to configure it from Headless, and it should be the source of truth in all scenarios.

https://coveord.atlassian.net/browse/KIT-3317


louis-bompart (approved):

louis-bompart:
react sample commerce to update ;)

PR #4094, opened 2024-06-14
fix(atomic): fallback language to commerce engine value instead of always defaulting to english

olamothe:
* Remove the default `@Prop` value for `language` in `atomic-commerce-interface` and `atomic-commerce-recommendation-interface`. Instead fallback on the value set in commerce engine.

* Modify `CommonAtomicInterfaceHelper` so that `language` is an optional property, and add the fallback to `en` there instead (ie: limit the amount of code changes for other types of interfaces that are not commerce). 

https://coveord.atlassian.net/browse/KIT-3318

### DIFF
--- a/packages/atomic/src/components.d.ts
+++ b/packages/atomic/src/components.d.ts
@@ -356,9 +356,9 @@ export namespace Components {
          */
         "initializeWithEngine": (engine: CommerceEngine) => Promise<void>;
         /**
-          * the commerce interface language.
+          * the commerce interface language.  Will default to the value set in the Headless engine context if not provided.
          */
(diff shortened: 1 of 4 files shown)


fbeaudoincoveo (approved):

olamothe:
> Would it make sense to do the same with the current in `atomic-product-price`?
> 
> ```ts
> @Prop({reflect: true}) public currency: string = 'USD';
> ```
> 
> And fallback on the currency defined in the engine ?
> 
> ... just to stay consistent with the rest of the logic

Its in another PR: https://github.com/coveo/ui-kit/pull/4091

PR #4095, opened 2024-06-14
fix(commerce): remove `executeFirstSearch` & friend from commerce engine

Spuffynism:
As the commerce engine is solution-type agnostic, it should not expose search-specific methods on its engine.

I adjusted the atomic commerce interface to take the change into consideration. However, I think we should remove `executeFirstSearch` from it as well since it's used for search and listings. Thoughts?


packages/headless/src/integration-tests/search-app.test.ts
```diff
@@ -27,9 +27,7 @@ import {
   buildRelevanceSortCriterion,
```
Spuffynism, line 27: I was cleaning up the commerce tests, so I cleaned up the search ones as well. Instead of running in 20s, they run in 7ms now.

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
```diff
@@ -287,14 +287,14 @@ export class AtomicCommerceInterface
       );
```
Spuffynism, line 287: @fbeaudoincoveo @louis-bompart should we even expose an `executeFirstSearch` on the interface? If this component is solution-type agnostic, it would make more sense for it _not_ to expose an `executeFirstSearch`.

Alternatively, if having this method available on the interface is useful, I'd propose we make it trigger a listing refresh or a search call, depending on the solution type.

olamothe: The `execute first request` type of function on atomic should stay on the interface. It can be named differently (I however dont have a nice generic name to propose...)

It should however be adapted to the solution type (ie: call a different Headless function), as you point out.

olamothe (approved):

louis-bompart (approved):

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
Spuffynism: Did this in https://github.com/coveo/ui-kit/pull/4095/commits/40cbd94bdee6340b59c9b672438e831f6ed03546

packages/atomic-react/src/components/commerce/CommerceInterfaceWrapper.tsx
```diff
@@ -2,19 +2,19 @@ import type {JSX, i18n} from '@coveo/atomic';
...
  * The properties of the AtomicCommerceInterface component
  */
 interface WrapperProps
   extends Omit<JSX.AtomicCommerceInterface, 'i18n' | 'pipeline' | 'searchHub'> {
   /**
-   * An optional callback function that can be used to control the execution of the first query.
+   * An optional callback function that can be used to control the execution of the first request.
    *
-   * If not provided, a default function will be used, which execute the first query immediately after initialization.
+   * If not provided, a default function will be used, which execute the request query immediately after initialization.
```
fbeaudoincoveo, line 15:
```suggestion
   * If not provided, a default function will be used, which execute the request immediately after initialization.
```

fbeaudoincoveo (approved):

packages/atomic-react/src/components/commerce/CommerceInterfaceWrapper.tsx
Spuffynism: Fixed to
> If not provided, a default function will be used, which executes the first request immediately after initialization.

in https://github.com/coveo/ui-kit/pull/4095/commits/fe9ac79cf66175479b331eae7330394cbdd7e586

PR #4096, opened 2024-06-17
fix(atomic): fix atomic-commerce-search-box needlessly requesting suggestions when input is disabled

olamothe:
Calling `searchBox.updateText()` triggers a query suggestion request in Headless: This is fine, and generally what implementers requires the search box to do.

However, in atomic, we've "split" the suggestion outside the searchbox component itself into a `SuggestionManager` utility class, which should be the one in charge of requesting suggestions when necessary.

To fix, exported `loadQuerySetActions` from Headless commerce, and use this action directly to update the state of the search box, without requesting `/querySuggest`.

https://coveord.atlassian.net/browse/KIT-3320


packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.tsx
```diff
@@ -650,6 +654,16 @@ export class AtomicCommerceSearchBox
       : this.bindings.i18n.t('query-suggestions-unavailable');
   }
 
+  private updateQueryWithoutQuerySuggestionTrigger(query: string) {
```
fbeaudoincoveo, line 657: An alternative would be to add an optional `fetchQuerySuggestions` parameter (which would default to `true`) on the search box controller's `updateText` method.

packages/headless/src/features/commerce/query-set/query-set-actions-loader.ts
```diff
@@ -0,0 +1,24 @@
+import {CommerceEngine} from '../../../app/commerce-engine/commerce-engine';
```
fbeaudoincoveo, line 1: I think we should rather create / use / export commerce-prefixed actions.

You can cherry-pick this commit if you want: https://github.com/coveo/ui-kit/commit/6c45fc583193bc9ffa5c0c9e9a3ce7e1fd107743

fbeaudoincoveo (changes requested):

packages/atomic/src/components/commerce/atomic-commerce-search-box/atomic-commerce-search-box.tsx
olamothe: I do not like this alternative: It adds complexity into a (IMO) perfectly fine function.

Dispatching actions directly is what we recommend implementers to do when a controller does not fit exactly their needs: just applying what we preach ;)

packages/atomic/src/components/commerce/atomic-commerce-search-box/e2e/atomic-commerce-search-box.e2e.ts
```diff
@@ -124,7 +136,14 @@ test.describe('with disable-search=true and minimum-query-length=1', () => {
...
 
 test.describe('with minimum-query-length=4', () => {
+  let querySuggestionRequestPerformed = false;
   test.beforeEach(async ({page}) => {
+    querySuggestionRequestPerformed = false;
+    page.on('request', (request) => {
+      if (request.url().includes('/querySuggest')) {
+        querySuggestionRequestPerformed = true;
+      }
+    });
```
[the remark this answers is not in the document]
olamothe: Im confused by "`we move that to the test suite?`" (collection of test). 

You seem to mean "a single test function" ?

louis-bompart (approved):

fbeaudoincoveo (approved):

PR #4104, opened 2024-06-19
fix(headless): update product interface to reflect how the commerce api behaves

olamothe:
This is less changes that what I had originally thought.

The `BaseProduct` interface should now properly reflect what's returned by the commerce API, following the source of truth (the code of the service itself). 

* Most of the change are about changing from `foo | undefined` to `foo | null`.
* Array values are never null, but might be an empty array (due to how the service is coded)
* ec_category is always an array, not a string
* Added couple missing properties (non breaking)

There is only "one lie" in this change: `excerpts, nameHighlights, excerptsHighlights` are "always defined when in a search use case", and "always undefined when in a listing or recommendation use case".

I thought about splitting and creating multiple interface `SearchProduct, ListingProduct, RecommendationProduct`, and remove `Product` as it exists right now.

However this would be a very noisy change, for existing deployments. `Product` is an *important* interface, which is nearly always referenced by any implementer using TypeScript.

So I made the judgment call to "lie a little bit, for the greater good" 😂  They are marked as possibly undefined for all use cases.

https://coveord.atlassian.net/browse/KIT-3164


packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.tsx
```diff
@@ -144,7 +144,7 @@ export class AtomicProductChildren
         <img
           class="aspect-square p-1"
           src={this.getImageUrl(child)}
-          alt={child.ec_name}
+          alt={child.ec_name || undefined}
```
buggyvelarde, line 147: can't help wondering whether a helper function to safely try to extract these things from an object would be useful - should we really be doing

```obj.something || underfined ```

or could we be doing something like

```getProductString(object, key)```

which would encapsulate this

Comment (olamothe) [reply packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.tsx:147]: The TypeScript compiler is the safety here: Since everything is strict, and `alt` must be `string | undefined` (it's a basic HTML property), the compiler is doing the protection of `string | null does not map perfectly to string | undefined`, and refuse to build otherwise.

Comment (buggyvelarde) [reply packages/atomic/src/components/commerce/product-template-components/atomic-product-children/atomic-product-children.tsx:147]: I guess you don't feel the need to because the type definitions in `ChildProduct` force you to have those checks

File: packages/headless/src/api/commerce/common/product.ts [hunk 2]
```diff
@@ -18,81 +20,90 @@ export interface BaseProduct {
    *
    * From the `ec_name` field.
    */
-  ec_name?: string;
+  ec_name: string | null;
```
buggyvelarde, line 23: @Spuffynism are we sure these appear as null in the JSON returned?

Spuffynism: Yup, for all of these properties, if the mapped field is not found in the Search API response, the casting to a `String` or a `Double` will result in the field being `null`.

Spuffynism (approved):

buggyvelarde (approved):
Nice types!

louis-bompart (approved):

PR #4105, opened 2024-06-19
feat(headless): update rga and smart snippets feedback events

jelmedini:
[SVCC-3817](https://coveord.atlassian.net/browse/SVCC-3817)

- Bump `coveo.analytics` to  `V2.30.6`
- Bump `@coveo/relay-event-types` to `V8.0.0`
- Add new version of analytics event for RGA new feedback modal: `logGeneratedAnswerFeedbackV2 `
- Update `like` and `dislike` events fro smart snippets and genqa, from  `Qna.SubmitFeedback` to `qna.answerAction`.
- Update smart snippet feedback event from `Qna.SubmitFeedback` to `Qna.SubmitSmartSnippetFeedback`
- Update existing genqa submit event from `Qna.SubmitFeedback` to `Qna.SubmitRgaFeedback` and remove the EP events, as they are no longer available and this function will be deprecated and removed after V3.

[SVCC-3817]: https://coveord.atlassian.net/browse/SVCC-3817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -84,6 +87,11 @@ export interface GeneratedAnswer extends Controller {
    * @param details - Details on why the generated answer was not relevant.
    */
   sendDetailedFeedback(details: string): void;
+  /**
```
olamothe, line 90: Should the old function be annotated as `@deprecated` in JS doc ?

Another way to go about this would be to change `sendFeedback(feedback: GeneratedAnswerFeedback | GeneratedAnswerFeedbackV2)`, and at runtime, detect the shape of the feedback object (is it v1 or v2) and dispatch the correct action from there.

olamothe: I guess, just in general, the `v2` name on a public facing function is problematic, and will need to coordinate to deprecate in the next major version.

epic for V3: https://coveord.atlassian.net/browse/KIT-2157

jelmedini: @olamothe Yes i love the first solution, it's way better, I will update according to thaat!

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -316,7 +324,20 @@ export function buildCoreGeneratedAnswer(
     },
 
     sendFeedback(feedback) {
-      dispatch(analyticsClient.logGeneratedAnswerFeedback(feedback));
+      if ((feedback as GeneratedAnswerFeedbackV2).helpful !== undefined) {
```
fbeaudoincoveo, line 327: Do we really need two actions?

I think we could always dispatch the same action, but the `feedback` argument would be `GeneratedAnswerFeedback | GeneratedAnswerFeedbackV2`. Then in the action's body, you'd make / submit the proper action based on the received `feedback` type.

E.g.,

```typescript
function isGeneratedAnswerFeedbackV2(
  feedback: GeneratedAnswerFeedback | GeneratedAnswerFeedbackV2
): feedback is GeneratedAnswerFeedbackV2 {
  return typeof feedback === 'object';
}

export const logGeneratedAnswerFeedback = (
  feedback: GeneratedAnswerFeedback | GeneratedAnswerFeedbackV2
): CustomAction =>
  makeAnalyticsAction({
    prefix: 'analytics/generatedAnswer/sendFeedback',
    __legacy__getBuilder: (client, state) => {
      const generativeQuestionAnsweringId =
        generativeQuestionAnsweringIdSelector(state);
      if (!generativeQuestionAnsweringId) {
        return null;
      }
      return isGeneratedAnswerFeedbackV2(feedback)
        ? client.makeGeneratedAnswerFeedbackSubmitV2({
            generativeQuestionAnsweringId,
            ...feedback,
          })
        : client.makeGeneratedAnswerFeedbackSubmit({
            generativeQuestionAnsweringId,
            reason: feedback,
          });
    },
  });
```

This way in V3, instead of removing an analytics action, you'd be restricting the type this action accepts as an argument.

In your tests, you'd just be checking that the action gets dispatched with the expected payload.

jelmedini: indeed much better! thank you

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -78,7 +81,9 @@ export interface GeneratedAnswer extends Controller {
    * Sends feedback about why the generated answer was not relevant.
    * @param feedback - The feedback that the end user wishes to send.
    */
-  sendFeedback(feedback: GeneratedAnswerFeedback): void;
+  sendFeedback(
+    feedback: GeneratedAnswerFeedback | GeneratedAnswerFeedbackV2
+  ): void;
```
mmitiche, line 86: With the v3, we will update the parameter `feedback` here to be only of type `GeneratedAnswerFeedbackV2` right?
In that case I think we should add a `TODO` comment here with the jira ticket in order to update that when we start the v3.

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -17,8 +17,26 @@ export type GeneratedAnswerFeedback =
   | 'notAccurate'
   | 'outOfDate'
   | 'harmful';
+
+export type GeneratedAnswerFeedbackOption = 'yes' | 'unknown' | 'no';
+
+export type GeneratedAnswerFeedbackV2 = {
```
mmitiche, line 23: Will we rename this type to become `GeneratedAnswerFeedback` starting from the v3?
In that case we need another `TODO` comment

packages/headless/src/features/generated-answer/generated-answer-insight-analytics-actions.ts
```diff
@@ -16,8 +16,26 @@ export type GeneratedAnswerFeedback =
...
+
+export type GeneratedAnswerFeedbackV2 = {
+  helpful: boolean;
+  documented: GeneratedAnswerFeedbackOption;
+  correctTopic: GeneratedAnswerFeedbackOption;
+  hallucinationFree: GeneratedAnswerFeedbackOption;
+  readable: GeneratedAnswerFeedbackOption;
+  details?: string;
+  documentUrl?: string;
+};
```
mmitiche, line 30: I think we should import these types from the `generated-answer-analytics-actions.ts` file instead of redeclaring them

packages/headless/src/features/generated-answer/generated-answer-insight-analytics-actions.ts
```diff
@@ -16,8 +16,26 @@ export type GeneratedAnswerFeedback =
...
 const RGAType = 'RGA';
 
+export function isGeneratedAnswerFeedbackV2(
+  feedback: GeneratedAnswerFeedback | GeneratedAnswerFeedbackV2
+): feedback is GeneratedAnswerFeedbackV2 {
+  return typeof feedback === 'object';
+}
+
```
mmitiche, line 38: This as well could be imported from the other file.

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -67,15 +67,13 @@ export const logLikeSmartSnippet = (): CustomAction =>
       return client.makeLikeSmartSnippet();
     },
     analyticsType: 'Qna.SubmitFeedback',
```
mmitiche, line 69: These analyticsType should become 'Qna.AnswerAction'.
Snapshots to be updated accordingly.

packages/headless/src/features/question-answering/question-answering-analytics-actions.ts
```diff
@@ -187,15 +183,14 @@ export const logSmartSnippetFeedback = (
       return client.makeSmartSnippetFeedbackReason(feedback);
     },
     analyticsType: 'Qna.SubmitFeedback',
```
mmitiche, line 185: analyticsType value to be updated.

packages/headless/src/features/question-answering/question-answering-insight-analytics-actions.ts
```diff
@@ -92,15 +90,13 @@ export const logDislikeSmartSnippet = (): InsightAction =>
       );
     },
     analyticsType: 'Qna.SubmitFeedback',
```
mmitiche, line 92: here as well ☝️

erocheleau (commented):
Not yet done reviewing, but some comments that might help?

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -151,22 +167,20 @@ export const logDislikeGeneratedAnswer = (): CustomAction =>
...
           type: RGAType,
         },
-        feedback: {
-          liked: false,
-        },
       };
     },
   });
 
 export const logGeneratedAnswerFeedback = (
```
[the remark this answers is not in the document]
jelmedini: @erocheleau that s what i didi first but i got feedbacks from @fbeaudoincoveo that suggestes this, and I tend to agree with him, This way in V3, instead of removing an analytics action, we will be restricting the type this action accepts as an argument.

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -17,8 +17,26 @@ export type GeneratedAnswerFeedback =
...
+  readable: GeneratedAnswerFeedbackOption;
+  details?: string;
+  documentUrl?: string;
+};
 const RGAType = 'RGA';
 
+export function isGeneratedAnswerFeedbackV2(
+  feedback: GeneratedAnswerFeedback | GeneratedAnswerFeedbackV2
+): feedback is GeneratedAnswerFeedbackV2 {
+  return typeof feedback === 'object';
```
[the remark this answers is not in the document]
jelmedini: i think it's pretty strait forward here to do it like this,  as the other type is not an object

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -176,26 +190,52 @@ export const logGeneratedAnswerFeedback = (
...
+          });
     },
+    analyticsType: isGeneratedAnswerFeedbackV2(feedback)
+      ? 'Qna.SubmitRgaFeedback'
+      : undefined,
+    analyticsPayloadBuilder: isGeneratedAnswerFeedbackV2(feedback)
+      ? (state): Qna.SubmitRgaFeedback => {
+          const {search} = state;
+          const {response} = search || {};
+          const responseId = response?.searchUid || '';
```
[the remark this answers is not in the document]
jelmedini: 🤷 the state is like this all over the code

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -661,7 +661,8 @@ describe('quantic-generated-answer', () => {
                 visitGeneratedAnswer({useCase: param.useCase});
               });
 
-              it('should properly display the like button', () => {
+              //Skipping this test for now, it will be adapted to the new logic in another ticket
```
olamothe, line 664: Add ticket number in the comment please !

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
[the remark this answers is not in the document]
olamothe: Its because technically, an implementer could try to dispatch any analytics action from any app, even one where the necessary data is not in the store.

ie: Nothing stop someone from calling `dispatch(logGeneratedAnswerFeedback())` from a page that has no gen ai answer component/controller loaded at all. 

It would obviously not do much or be very useful, but it should no crash the app.

As for the syntax, what Etienne propose is easier to read.

olamothe (approved):

fbeaudoincoveo (approved):
LGTM!

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
mmitiche, line 664: Created the ticket to adapt this:
```suggestion
              // TODO: SFINT-5608 Skipping this test for now, it will be adapted to the new logic in another ticket
```

mmitiche (approved):
The changes look good to me, nice job!
I would just do a sanity check by trying the events in an Atomic/Quantic interface and test the payload sent with the validate endpoint: https://platformdev.cloud.coveo.com/docs?urls.primaryName=Event#/Event%20API/rest_organizations_paramId_events_v1_validate_post

To make sure that the events sent according to this logic are valid.

PR #4107, opened 2024-06-19
feat(atomic-commerce): products-per-page

louis-bompart:
- refactor:`PageGuard` to be "generic" (results->items)
- refactor: `PageGuard` for `atomic-per-page`
- feat: atomic-commerce-products-per-page
- tests: e2e for ☝️ 

https://coveord.atlassian.net/browse/KIT-3407


packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.tsx
```diff
@@ -0,0 +1,179 @@
...
+ * @part active-button - The active product per page button.
+ *
+ * @alpha
+ */
+@Component({
+  tag: 'atomic-commerce-products-per-page',
+  styleUrl: 'atomic-commerce-products-per-page.pcss',
+  shadow: true,
+})
+export class AtomicResultsPerPage implements InitializableComponent<Bindings> {
```
olamothe, line 37: There's too much copy paste between this and `atomic-results-per-page`. 

The most important thing will be to extract functional components for everything visual (ie: all `renderXYZ()` function) at the very least. This is mandatory to make sure parts, styling (css classes), i18n strings, and how things behave for end users stays consistent.

Then there's some other utility type function (ie: the `validateXYZ()` functions, or code that control end user focus for accessibility) can be shared exactly one to one, and so we should aim to refactor and extract those. 

In general however, those utility function are "less important" to refactor, if compared to the rendering code itself.

olamothe (changes requested):

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.new.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+            ${story()}
+          </atomic-layout-section>
+        </atomic-layout-section>
+      </atomic-commerce-layout>`,
+  ],
+  play: async (context) => {
+    await play(context);
+    await playExecuteFirstSearch(context);
+  },
+};
```
alexprudhomme, line 102: This is annoying but because of this regex in Storybook via the `buildArgsParam` in our `load` function, I was unable to sent via the url the choicesDisplayed attribute since it contains commas `,`. Since I couldnt just send a attribute via the url, I had to create a whole story for it.
https://github.com/storybookjs/storybook/blob/4f9108cc05621d913aff9163243c3af8ba80b9f2/code/core/src/router/utils.ts#L66

packages/headless/src/controllers/commerce/core/pagination/headless-core-commerce-pagination.ts
```diff
@@ -170,6 +177,10 @@ export function buildCorePagination(
       dispatch(props.fetchProductsActionCreator());
     },
 
+    setPageSizeWithoutFetch(pageSize: number) {
+      dispatch(setPageSize({slotId, pageSize}));
+    },
+
```
alexprudhomme, line 183: I am unsure what to do here. The problem is that `setPageSize` (expand the file to see it just above) sets the page size and ALSO fetches products via `dispatch(props.fetchProductsActionCreator());`. This causes a problem in atomic because `setPageSize` is used to set the `initialChoice` value. On the first load, a second search is triggered at the same time of the first (only one with the correct pageSize value) which causes big problems with the url manager and race conditions. 

Two options :
 - Add a new `setPageSizeWithoutFetch` that only sets the headless state value without fetching products.
 - Remove the  `dispatch(props.fetchProductsActionCreator());` from the setPageSize and manually do `setPageSize` + `fetchMoreProducts` in Atomic. (Possibly breaking on the headless side)

alexprudhomme: I could also call it `SetInitialPageSize`

fbeaudoincoveo: Another option would be to dispatch the `setPageSize` action directly in Atomic to set the `initialChoice` value.

alexprudhomme: No, that is precisely the problem. In the `initialize` function, we set the intialChoice using setPageSize which triggers a second search. We need a option to setPageSize without triggering a search for this use case in the initialize function that happens before the `executeFirstRequest` is triggered.

image

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.tsx
```diff
@@ -0,0 +1,125 @@
...
+    validateInitialChoice(this, this.choices);
+
+    const controller =
+      this.bindings.interfaceElement.type === 'search'
+        ? buildSearch(this.bindings.engine)
+        : buildProductListing(this.bindings.engine);
+
+    this.summary = controller.summary();
+    this.pagination = controller.pagination();
+    this.pagination.setInitialPageSize(this.initialChoice!);
```
fbeaudoincoveo, line 88: Here, instead of calling the controller method, which has the undesirable side effect of triggering a refresh / search, you could dispatch the `setPageSize` action directly.

It would look something like:

```
const paginationAction = loadPaginationActions(this.bindings.engine);
this.bindings.engine.dispatch(this.paginationActions.setPageSize(this.initialChoice));
```

packages/headless/src/controllers/commerce/core/pagination/headless-core-commerce-pagination.ts
alexprudhomme: image

alexprudhomme: my mistake, I can just enginge.dispatch 😅

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.tsx
alexprudhomme: thanks 😅

alexprudhomme:
> Something is wrong with the URL management. Whenever the search page loads with the `#perPage=` attribute in the URL, 2 search queries are triggered. The second being the value in the `perPage` value from the URL image
> 
> However, when first loading a listing page, the perPage value does not have any effect on the component.

fixed with this ! 

 `   this.bindings.engine.dispatch(setPageSize({pageSize: this.initialChoice!}));`
instead of 
`    this.pagination.setPageSize(this.initialChoice!)`

because of this 

https://github.com/coveo/ui-kit/pull/4107#discussion_r1681257270

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/e2e/atomic-commerce-products-per-page.e2e.ts
```diff
@@ -0,0 +1,80 @@
...
+      ).toBeVisible();
+    });
+
+    test.describe('when selecting a different number of products', () => {
+      test.beforeEach(async ({productsPerPage}) => {
+        await productsPerPage.choice(25).click();
+      });
+
+      test('should render the component with the right choice selected', async ({
+        productsPerPage,
```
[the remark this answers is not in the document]
alexprudhomme: I don't think that is relevant in that component since there is nothing in the logic referring to either a listing page or a search page. If there is a problem here due to it's commerce-interface type, there would be a problem in every other component.

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.tsx
alexprudhomme: I extracted the rendering logic in 2 seperate common component and added the validate logic to a common file.

packages/atomic/src/components/common/items-per-page/validate.ts
```diff
@@ -0,0 +1,32 @@
...
+    const parsedChoice = parseInt(choice);
+    if (isNaN(parsedChoice)) {
+      const errorMsg = `The choice value "${choice}" from the "choicesDisplayed" option is not a number.`;
+      context.bindings.engine.logger.error(errorMsg, context);
+      throw new Error(errorMsg);
+    }
+
+    return parsedChoice;
+  });
+}
```
louis-bompart, line 17: How about having this function takes choicesDisplayed, instead of the whole context?
Make it throw a dedicated Error (e.g. ChoiceIsNaNError), catch it, log it, rethrow it.

packages/atomic/src/components/common/items-per-page/validate.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function validateInitialChoice(
+  context: AtomicResultsPerPage | AtomicCommerceProductsPerPage,
+  choices: number[]
+) {
+  if (!context.initialChoice) {
+    context.initialChoice = choices[0];
```
fbeaudoincoveo, line 24: I don't think the function should have a side effect like that. Instead, it should return a value that indicates what the initial choice will be.

Then, we can remove the `mutable: true` property on the `initialChoice` prop

packages/atomic/src/components/common/items-per-page/validate.ts
```diff
@@ -0,0 +1,32 @@
...
+      const errorMsg = `The choice value "${choice}" from the "choicesDisplayed" option is not a number.`;
+      context.bindings.engine.logger.error(errorMsg, context);
+      throw new Error(errorMsg);
+    }
+
+    return parsedChoice;
+  });
+}
+
+export function validateInitialChoice(
```
fbeaudoincoveo, line 19: I'd call this function "getInitialPageSize" instead.

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.tsx
```diff
@@ -0,0 +1,127 @@
...
+  );
+
+  /**
+   * A list of choices for the number of products to display per page, separated by commas.
+   */
+  @Prop({reflect: true}) choicesDisplayed = '10,25,50,100';
+  /**
+   * The initial selection for the number of product per page. This should be part of the `choicesDisplayed` option. By default, this is set to the first value in `choicesDisplayed`.
+   */
+  @Prop({mutable: true, reflect: true}) initialChoice?: number;
```
fbeaudoincoveo, line 71: I messed around a bit to try to understand why your e2e tests are failing, and I found out that as it is, in storybook, the `initialChoice` property is typed as an object (which makes no sense!)

 if you set a default value (say 0) to the initialChoice, then everything works fine (you don't even have to adjust the check in validate.ts, cause !context.initialChoice will be truthy when initialChoice === 0).

packages/atomic/src/components/common/items-per-page/validate.ts
alexprudhomme: This makes the initialize() function in the corresponding atomic components very verbose. I prefer it the way it is now. What exactly is the problem with passing the context ?

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    const controller =
+      this.bindings.interfaceElement.type === 'search'
+        ? buildSearch(this.bindings.engine)
+        : buildProductListing(this.bindings.engine);
+
+    this.summary = controller.summary();
+    this.pagination = controller.pagination();
+    const {setPageSize} = loadPaginationActions(this.bindings.engine);
+    this.bindings.engine.dispatch(setPageSize({pageSize: this.initialChoice!}));
```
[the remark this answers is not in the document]
alexprudhomme: thanks !

packages/atomic/src/components/common/items-per-page/validate.ts
alexprudhomme: Thanks, instead I removed the logic from the validate function and added this just before instead

`      this.initialChoice = this.initialChoice ?? this.choices[0];`

packages/atomic/src/components/common/items-per-page/validate.ts
alexprudhomme: agree !

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.new.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  title: 'Atomic-Commerce/ProductsPerPage',
+  id: 'atomic-commerce-products-per-page',
+
+  render: renderComponent,
+  decorators: [decorator],
+  parameters,
+  play: async (context) => {
+    await play(context);
+    await playExecuteFirstSearch(context);
+  },
```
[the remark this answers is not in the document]
alexprudhomme: Agreed

packages/atomic/src/components/common/items-per-page/validate.ts
alexprudhomme: instead I removed the side effect !

packages/atomic/src/components/commerce/atomic-commerce-products-per-page/atomic-commerce-products-per-page.tsx
alexprudhomme: Yea it is weird. I found a fix where adding a @type {number} in the jsdoc fixes it !

PR #4108, opened 2024-06-20
fix(atomic): adjust product-list placeholder

y-lakhdar:
* Add ItemListGuard to prevent the  placeholders to be displayed in case no results is returned on the initial page load
* Added the minimum E2E tests to cover the bug fix

https://coveord.atlassian.net/browse/KIT-3329

### DIFF
--- a/packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
+++ b/packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
@@ -6,6 +6,9 @@ import {
   SearchState,
   Search,
   Product,
+  Summary,
+  ProductListingSummaryState,
+  SearchSummaryState,
 } from '@coveo/headless/commerce';
 import {
   Component,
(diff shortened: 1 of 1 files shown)


packages/atomic/src/components/commerce/atomic-commerce-product-list/atomic-commerce-product-list.tsx
```diff
@@ -183,29 +183,39 @@ export class AtomicCommerceProductList
       : this.searchState;
   }
 
+  get summaryState() {
+    return this.bindings.interfaceElement.type === 'product-listing'
+      ? this.productListing.summary().state
+      : this.search.summary().state;
+  }
```
louis-bompart, line 190: I don't think this is the right way of doing that 🤔 
I think the state is not reactive that way.

You should probably initialize and store the summary controller in a private property of the class and use our decorators to map its state.

I did it recently in #4083 for example

olamothe (approved):
All good, apart from `summaryState()` not being reactive, as pointed out by Louis

PR #4111, opened 2024-06-21
feat(headless commerce): retrieve clientId from Relay instead of coveoua when building commerce API requests

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3335

### DIFF
--- a/packages/headless/src/features/commerce/common/actions.ts
+++ b/packages/headless/src/features/commerce/common/actions.ts
@@ -33,12 +33,12 @@ export interface QueryCommerceAPIThunkReturn {
   response: CommerceSuccessResponse;
 }
 
-export const buildCommerceAPIRequest = async (
+export const buildCommerceAPIRequest = (
   state: ListingAndSearchStateNeededByQueryCommerceAPI,
   relay: Relay
-): Promise<CommerceAPIRequest> => {
+): CommerceAPIRequest => {
   return {
-    ...(await buildBaseCommerceAPIRequest(state, relay)),
+    ...buildBaseCommerceAPIRequest(state, relay),
     facets: getFacets(state),
     ...(state.commerceSort && {
       sort: getSort(state.commerceSort.appliedSort),
@@ -46,11 +46,11 @@ export const buildCommerceAPIRequest = async (
   };
 };
 
-export const buildBaseCommerceAPIRequest = async (
+export const buildBaseCommerceAPIRequest = (
   state: StateNeededByQueryCommerceAPI,
   relay: Relay,
   slotId?: string
-): Promise<BaseCommerceAPIRequest> => {
+): BaseCommerceAPIRequest => {
   const {view, user, ...restOfContext} = state.commerceContext;
   return {
     accessToken: state.configuration.accessToken,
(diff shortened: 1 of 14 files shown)


packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -47,6 +48,7 @@ export const buildCommerceAPIRequest = async (
 
 export const buildBaseCommerceAPIRequest = async (
```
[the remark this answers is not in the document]
fbeaudoincoveo: Funny I remember making the exact same comment when @louis-bompart made similar changes elsewhere 😅

Spuffynism:
Just noticed the query suggest request was forgotten: https://github.com/coveo/ui-kit/blob/c9822ed0710644c0973546e8e2c2cdb3609554b9/packages/headless/src/features/commerce/query-suggest/query-suggest-actions.ts#L67-L79

packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -34,10 +34,11 @@ export interface QueryCommerceAPIThunkReturn {
 }
 
 export const buildCommerceAPIRequest = async (
-  state: ListingAndSearchStateNeededByQueryCommerceAPI
+  state: ListingAndSearchStateNeededByQueryCommerceAPI,
+  relay: Relay
```
olamothe, line 38: Why do we add `relay` as a function parameter (and thus, impact all upstream code calling the function, as we can see in this PR) instead of just `createRelay(....extractStuffFromTheState)` directly in the body of this function ?

olamothe: Or, another alternative, add a new function in the relay package that does only the clientId extraction, without creating a full client.

olamothe: And then, more generally, should this be replicated to **all** API request builder function across the project, when analytics mode is set to `next`. ie: making sure the fix is correctly applied to all use cases, not just commerce.

fbeaudoincoveo: > Or, another alternative, add a new function in the relay package that does only the clientId extraction, without creating a full client.

I kind of like this idea.

On the other hand, I don't think there's anything wrong with what we're doing right now: the Relay client is initialized once in the core engine, and then passed down as needed.

olamothe: > On the other hand, I don't think there's anything wrong with what we're doing right now:

What I don't like is the "shotgun effect" it creates in the codebase (one change === multiple file changes).

"Passing it down as needed" is complexity to figure out where something actually comes from/is created (multiple level upstream), vs just a local variable/function (easier to reason about).

If would be one thing if the builder function was really lacking data to get what it needs from relay, but from my understanding it's already all available, so it's a bit silly IMO to do it like this PR propose.

olamothe (approved):

fbeaudoincoveo:
> Just noticed the query suggest request was forgotten:
> 
> https://github.com/coveo/ui-kit/blob/c9822ed0710644c0973546e8e2c2cdb3609554b9/packages/headless/src/features/commerce/query-suggest/query-suggest-actions.ts#L67-L79

Good catch! Fixing

packages/headless/src/features/commerce/common/actions.ts
samisayegh: Not opposed to it, but I'd like to see a design from the perspective of an implementer first and then we can decide on the next steps. There's a similar discussion happening [here](https://github.com/coveo/ui-kit/pull/4113#discussion_r1653008709),

PR #4113, opened 2024-06-21
fix(commerce): populate `context.user.userAgent` with `navigatorContext`

Spuffynism:
headless was populating the `context.user.userAgent` property with what was set through the `setUser`. With this PR, we extract the `userAgent` and the `referrer` from the relay instance and use these values instead. This makes using the headless commerce controllers less error-prone.

⚠️ This means we can remove the `setUser` action (and its sibling method on the context controller), and `setView`'s `referrer` parameter, which I do in this PR. Do you think this makes sense? Is there a use case for allowing manual control over these parameters? @samisayegh @louis-bompart @fbeaudoincoveo ?

[CAPI-1091]

[CAPI-1091]: https://coveord.atlassian.net/browse/CAPI-1091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/headless/src/app/commerce-engine/commerce-engine-configuration.ts
```diff
@@ -57,7 +57,6 @@ export function getSampleCommerceEngineConfiguration(): CommerceEngineConfigurat
       currency: 'USD',
```
samisayegh, line 57: Hey, Louis recently added a function called `navigatorContext` that returns values for `userAgent`, `referrer` and `url`.
The benefit for using the function instead of relay is to allow implementers to pass values to headless in a server-side rendering (SSR) scenario. If we only use relay internally, it would not be possible to configure in that context,

Spuffynism: Should we not retrieve the clientId from relay in that case, as is done in https://github.com/coveo/ui-kit/pull/4111?

samisayegh: That's a great point, indeed the same argument would apply there. The default `navigatorContext` used in the browser could return the different properties using relay, while still allowing one to specify them in an SSR scenario,

fbeaudoincoveo:
> Is there a use case for allowing manual control over these parameters? 

If we use the navigator context provider to set the `userAgent` and `referrer` as suggested by @samisayegh , then I don't think there would be any need to allow further manual control over these parameters. The implementer could have complete control over them by passing a custom `navigatorContextProvider` when initializing the engine.

packages/headless/src/app/commerce-engine/commerce-engine-configuration.ts
fbeaudoincoveo: @samisayegh , does this mean than in an SSR scenario (which is the default scenario we will ultimately want to push for, especially for commerce implementations), Relay will not be able to extract the client ID?

packages/headless/src/features/commerce/pagination/pagination-slice.ts
```diff
@@ -9,7 +9,7 @@ import {
   toggleExcludeNumericFacetValue,
   toggleSelectNumericFacetValue,
 } from '../../facets/range-facets/numeric-facet-set/numeric-facet-actions';
-import {setContext, setUser, setView} from '../context/context-actions';
+import {setContext, setView} from '../context/context-actions';
```
fbeaudoincoveo, line 12: Kinda stating the obvious here, but this is a breaking change 😅

packages/headless/src/app/commerce-engine/commerce-engine-configuration.ts
samisayegh: That's correct, SSR is completely different to the browser. A `clientId` would either:
1. Read from the request header if present.
2. Need to be generated on the server and returned to the browser e.g., via a response header

The clientId needs to be given to headless in an SSR context

We can export functions from relay (or a different package) to help. It'll be better to design how it looks like first though.

Naively, 

```
app.get('/', (req, res) => {
  const clientId = getOrGenerateClientId(req.headers) // provided by relay?

  const engine = buildCommerceEngine({
    ...,
    navigatorContext: () => {
     clientId,
     userAgent: req.userAgent,
     etc..
    }
  })
  ...

  headlers = enrichHeadersWithClientId(headers) // provided by relay?
  res.headers(headers) 
})
```
If headless SSR accepts and returns headers, we could simplify things quite a bit 🤞

Spuffynism:
Re-requesting reviews since I've changed the approach to use the navigator context across commerce requests to fetch the user agent and referrer!

packages/headless/src/features/commerce/facets/facet-search-set/category/commerce-category-facet-search-request-builder.ts
```diff
@@ -34,7 +36,7 @@ export const buildCategoryFacetSearchRequest = async (
     clientId,
     context,
     ...restOfCommerceAPIRequest
-  } = buildCommerceAPIRequest(state, relay);
+  } = buildCommerceAPIRequest(state, relay, navigatorContext);
```
samisayegh, line 39: Passing both `relay` and `navigatorContext` is good for now, and we can revisit in the future.

Have we logged a jira to take another look at this though, maybe in the context of headless commerce SSR?

samisayegh (approved):
✅

packages/headless/src/features/commerce/facets/facet-search-set/category/commerce-category-facet-search-request-builder.ts
louis-bompart: Headess Commerce SSR is to be considered for investigation in Q3. For now I'll add a link to the Epic (that I'll create now)

samisayegh: Thanks,

pvanool (approved):

packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -49,19 +51,32 @@ export const buildCommerceAPIRequest = (
 export const buildBaseCommerceAPIRequest = (
   state: StateNeededByQueryCommerceAPI,
   relay: Relay,
+  navigatorContext: NavigatorContext,
   slotId?: string
 ): BaseCommerceAPIRequest => {
-  const {view, user, ...restOfContext} = state.commerceContext;
+  const {view, ...restOfContext} = state.commerceContext;
+  const meta = relay.getMeta('');
```
fbeaudoincoveo, line 58: I'm not against this, but this seems to be the only place where we're doing this (usually we just do `clientId: relay.getMeta('').clientId`.

fbeaudoincoveo (approved):

packages/headless/src/features/commerce/common/actions.ts
Spuffynism: I originally did this because I was re-using it for the other properties. I'll inline it!

Spuffynism: Done in https://github.com/coveo/ui-kit/pull/4113/commits/4ac9380f5842bd413277a8f55bc48b5c3dd0d90f

PR #4116, opened 2024-06-25
fix(atomic): escape searchQuery in regex

y-lakhdar:
https://coveord.atlassian.net/browse/KIT-3337


louis-bompart (approved):
LGTM I think 🤔

PR #4120, opened 2024-06-26
feat(quantic): quantic search box improvements

mmitiche:
## [SFINT-5576](https://coveord.atlassian.net/browse/SFINT-5576)

In this PR:

- Search box recent queries feature added to the Quantic Search Box.
- Several bugs fixed with the Quantic Search Box.
- A11y of the Quantic Search Box improved.
- Cleaned the Quantic Search Box component and all of its internal components for a more simplicity.

Quantic Search Box autopsy: https://coveord.atlassian.net/wiki/spaces/~60c8f25b2bd2140069e4cdbc/pages/4208459786/Quantic+Search+Box+Autopsy

[SFINT-5576]: https://coveord.atlassian.net/browse/SFINT-5576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


erocheleau (commented):
I still need to review the search-box.cypress.ts file.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1442,4 +1442,46 @@
...
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>Recent query,</shortDescription>
+  </labels>
+  <labels>
+    <fullName>quantic_QuerySuggestionAriaLabel</fullName>
+    <value>Query suggestion,</value>
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>Query suggestions,</shortDescription>
```
SimonMilord, line 1485: It does indeed feels weird to have a "," in the label here also. IMO labels should be independent of each other or any other text, meaning it should not "assume" there is more text coming either before or after. Better for reusability

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1442,4 +1442,46 @@
...
+  <labels>
+    <fullName>quantic_SuggestionFound</fullName>
+    <value>{{0}} search suggestion found, to navigate use up and down arrows.</value>
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>1 search suggestion found</shortDescription>
+  </labels>
+  <labels>
+    <fullName>quantic_SuggestionFound_plural</fullName>
+    <value>{{0}} search suggestions found, to navigate use up and down arrows.</value>
```
SimonMilord, line 1461: What do you think of a seperate label for `to navigate use up and down arrows.` ? Would we ever want to use the first part of the label without the navigation sentence?

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1442,4 +1442,46 @@
     <protected>false</protected>
     <shortDescription>Show less</shortDescription>
   </labels>
+  <labels>
+    <fullName>quantic_SearchFieldWithSuggestions</fullName>
+    <value>Search field with suggestions. Suggestions may be available under this field. To send, press Enter.</value>
```
SimonMilord, line 1447: Question here. This label reads a bit odd if not in a aria context. Do we want to create "aria-specific" labels? Meaning labels only meant for screen readers/accessibility purposes or we want to keep everything uniform? I was thinking we could maybe name the labels for aria purposes something like: `quantic_Aria_SearchFieldWithSuggestions`

packages/quantic/cypress/e2e/default-2/search-box/search-box-selectors.ts
```diff
@@ -0,0 +1,43 @@
...
+    SearchBoxSelectors.get().find('[data-cy="quantic-search-box-input"]'),
+  input: (textarea = false) =>
+    SearchBoxSelectors.get().find(
+      textarea
+        ? 'c-quantic-search-box-input [data-cy="search-box-textarea"]'
+        : 'c-quantic-search-box-input [data-cy="search-box-input"]'
+    ),
+  suggestionList: () =>
+    SearchBoxSelectors.get().find(
+      'c-quantic-search-box-input c-quantic-search-box-suggestions-list [data-cy="suggestion-list"]'
```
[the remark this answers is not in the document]
mmitiche: not really indeed we can remove it.

packages/quantic/cypress/e2e/default-2/search-box/search-box-actions.ts
```diff
@@ -0,0 +1,44 @@
+import {SearchBoxSelector, SearchBoxSelectors} from './search-box-selectors';
+
+const standaloneSearchBoxActions = (selector: SearchBoxSelector) => {
+  return {
+    typeInSearchBox: (query: string, textarea = false) => {
+      selector
+        .input(textarea)
+        .then((searchbox) => {
+          let updateText = '';
+          query.split('').forEach((letter) => {
```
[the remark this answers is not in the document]
mmitiche: I will try the type and see, as far as I remember this method didn't really trigger the keyboard events that we are listening too, I will retry it and come back with a confirmation.

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
```diff
@@ -0,0 +1,142 @@
...
+        .logDetail(
+          `${should(display)} display the 'clear recent queries' button`
+        );
+    },
+    querySuggestionsEquals: (querySuggestions: String[]) => {
+      querySuggestions.forEach((querySuggestion, index) => {
+        selector
+          .querySuggestion(index)
+          .invoke('attr', 'data-rawvalue')
+          .should('equal', querySuggestion)
```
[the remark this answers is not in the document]
mmitiche: the reason we don't relay on the actual text is because the text is displayed using a rich text formatter as it contains some bold parts. 

but I agree that this could be improved for more assurance, I will try to improve it.

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
```diff
@@ -0,0 +1,142 @@
...
+          .logDetail(
+            'should display the query suggestions properly and in the correct order'
+          );
+      });
+    },
+    searchWithQuery: (
```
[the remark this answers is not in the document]
mmitiche: Indeed that's true.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -120,31 +151,38 @@ export default class QuanticSearchBox extends LightningElement {
     );
   }
 
+  get searchBoxValue() {
+    return this.searchBox?.state.value || '';
+  }
```
SimonMilord, line 156: we no longer want to manage the value in the `quanticSearchBoxInput`? Why pass it as props?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +93,200 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+    if (this.selectionIndex >= this.allOptions.length) {
       this.selectionIndex = 0;
     }
+    return this.allOptionsHTMLElements[this.selectionIndex].getAttribute('id');
   }
 
   /**
    * Return the currently selected suggestion.
-   * @returns {Suggestion}
+   * @returns {Object}
```
SimonMilord, line 104: why not `@returns {Suggestion}` here?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +93,200 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+      detail: listboxId,
+      bubbles: true,
+      composed: true,
     });
-    this.dispatchEvent(highlightChangeEvent);
+    this.dispatchEvent(suggestionListEvent);
+  }
+
+  /**
+   * Returns all the options to be displayed inside the suggestion list.
```
SimonMilord, line 175:
```suggestion
   * Returns all the options (recent queries + query suggestions) to be displayed inside the suggestion list.
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +93,200 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+      const clearRecentQueriesOption = {
+        key: 0,
+        id: 'selection-0',
+        isSelected: this.selectionIndex === 0,
+        isClearRecentQueryButton: true,
+        onClick: (event) => {
+          this.handleSelection(event, 0);
+        },
+      };
+      options.unshift(clearRecentQueriesOption);
```
SimonMilord, line 197: Nice!

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -1,16 +1,35 @@
...
+
+const optionCSSClass =
+  'slds-media slds-listbox__option slds-listbox__option_plain slds-media_small slds-grid option';
 
 /**
  * @typedef Suggestion
  * @property {number} key
  * @property {string} value
  * @property {string} rawValue
+ * @property {false} isRecentQuery
```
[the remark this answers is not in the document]
SimonMilord: also if we need it do we not want to add it to the `quanticSearchBoxInput` Suggestion definition also?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -197,69 +198,81 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
    */
-  onKeyup(event) {
+  onKeyDown(event) {
     switch (event.key) {
+      case keys.ESC:
+        this.input.removeAttribute('aria-activedescendant');
+        this.input.blur();
+        break;
+
+      case keys.ARROWRIGHT:
```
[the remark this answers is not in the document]
SimonMilord: yeah I think the expected behaviour here would be to use left and right arrows to correct or navigate the input to change the query and not to navigate through the dropdown menu

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
```diff
@@ -241,7 +240,9 @@ export default class QuanticStandaloneSearchBox extends NavigationMixin(
       {
         type: 'standard__webPage',
         attributes: {
-          url: `${this.redirectUrl}${value ? `#q=${encodeURIComponent(value)}` : ''}`,
+          url: `${this.redirectUrl}${
+            value ? `#q=${encodeURIComponent(value)}` : ''
+          }`,
```
SimonMilord, line 245: was that due to the linter? seems weirder imo

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -180,24 +190,75 @@ describe('c-quantic-search-box-input', () => {
...
+            expect(suggestionsListItems.length).toEqual(mockSuggestions.length);
           });
-          await flushPromises();
+        });
 
-          const suggestionsList = element.shadowRoot.querySelector(
-            selectors.searchBoxSuggestionsList
-          );
-          expect(suggestionsList).not.toBeNull();
+        describe('when both query suggestions and recent queries are displayed', () => {
```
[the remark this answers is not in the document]
SimonMilord: "enabled"

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -180,24 +190,75 @@ describe('c-quantic-search-box-input', () => {
...
+            const clearRecentQueriesButton =
+              suggestionsList.shadowRoot.querySelector(
+                selectors.clearRecentQueryButton
+              );
+            expect(suggestionsListItems).not.toBeNull();
+            expect(clearRecentQueriesButton).not.toBeNull();
+            expect(suggestionsListItems.length).toEqual(
+              mockSuggestions.length + exampleRecentQueries.length
+            );
+          });
```
SimonMilord, line 261:
```suggestion
            const suggestionsListItems =
              suggestionsList.shadowRoot.querySelectorAll(
                selectors.suggestionOption
              );
            expect(suggestionsListItems).not.toBeNull();

            const clearRecentQueriesButton =
              suggestionsList.shadowRoot.querySelector(
                selectors.clearRecentQueryButton
              );
            expect(clearRecentQueriesButton).not.toBeNull();
            
            expect(suggestionsListItems.length).toEqual(
              mockSuggestions.length + exampleRecentQueries.length
            );
          });
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -250,6 +311,68 @@ describe('c-quantic-search-box-input', () => {
...
+            await input.focus();
+
+            const suggestionsList = element.shadowRoot.querySelector(
+              selectors.searchBoxSuggestionsList
+            );
+            expect(suggestionsList).not.toBeNull();
+
+            const querySuggestionIndex = 0;
+            const firstSuggestion =
+              suggestionsList.shadowRoot.querySelectorAll('li')[
```
SimonMilord, line 330:
```suggestion
              suggestionsList.shadowRoot.querySelectorAll(selectors.suggestionOption)[
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -259,16 +382,71 @@ describe('c-quantic-search-box-input', () => {
...
+            await input.focus();
+
+            const suggestionsList = element.shadowRoot.querySelector(
+              selectors.searchBoxSuggestionsList
+            );
+            expect(suggestionsList).not.toBeNull();
+
+            const recentQueryIndex = 0;
+            const firstSuggestion =
+              suggestionsList.shadowRoot.querySelectorAll('li')[
```
SimonMilord, line 433:
```suggestion
              suggestionsList.shadowRoot.querySelectorAll(selectors.suggestionOption)[
```

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -259,16 +382,71 @@ describe('c-quantic-search-box-input', () => {
...
+                recentQueryIndex + 1
+              ];
+            expect(firstSuggestion).not.toBeNull();
+
+            firstSuggestion.dispatchEvent(new CustomEvent('mousedown'));
```
SimonMilord, line 438:
```suggestion
            const firstRecentQuery =
              suggestionsList.shadowRoot.querySelectorAll(selectors.suggestionOption)[
                recentQueryIndex + 1
              ];
            expect(firstRecentQuery).not.toBeNull();

            firstRecentQuery.dispatchEvent(new CustomEvent('mousedown'));
```

packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box-actions.ts
```diff
@@ -15,7 +15,7 @@ const standaloneSearchBoxActions = (selector: StandaloneSearchBoxSelector) => {
             updateText += letter;
             cy.wrap(searchbox)
               .invoke('val', updateText)
-              .trigger('keyup', {which: letter.charCodeAt(0)});
+              .trigger('input', {which: letter.charCodeAt(0)});
```
SimonMilord, line 18: wut?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,750 @@
...
+  {variantName: 'default', textarea: false},
+  {variantName: 'expandable', textarea: true},
+];
+
+function setRecentQueriesInLocalStorage(recentQueries: String[]) {
+  window.localStorage.setItem(
+    recentQueriesLSKey,
+    JSON.stringify(recentQueries)
+  );
+}
```
SimonMilord, line 37: question here, when we use the local storage in tests, does the storage memory of the browser resets every time we run the tests?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,750 @@
...
+              Actions.focusSearchBox(textarea);
+              cy.wait(InterceptAliases.QuerySuggestions);
+
+              Expect.displaySuggestionList(true);
+              Expect.displayClearRecentQueriesButton(true);
+              Expect.numberOfQuerySuggestions(expectedQuerySuggestions.length);
+              Expect.querySuggestionsEquals(expectedQuerySuggestions);
+            });
+
+            scope('when selecting a recent query', () => {
```
SimonMilord, line 130: why this scope block here if we only test that the recent queries/suggestions are displayed in the right order?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,750 @@
...
+              Actions.clickQuerySuggestion(clickedSuggestionIndex);
+              Expect.searchWithQuery(
+                exampleQuerySuggestions[clickedSuggestionIndex],
+                {
+                  LSkey: recentQueriesLSKey,
+                  queries: [exampleQuerySuggestions[clickedSuggestionIndex]],
+                }
+              );
+            });
+          });
```
SimonMilord, line 304: here should we add a scope to check if you focus again on the searchbox after clearing that the recent queries are no longer displayed?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,750 @@
...
+        describe('when using the keyboard to select a recent query', () => {
+          const exampleRecentQueries = ['foo', 'bar'];
+
+          beforeEach(() => {
+            setRecentQueriesInLocalStorage(exampleRecentQueries);
+            visitSearchBox({...defaultOptions, textarea});
+            mockQuerySuggestions(exampleQuerySuggestions);
+          });
+
+          it('should properly navigate and select the right suggestion', () => {
```
SimonMilord, line 316:
```suggestion
          it('should properly navigate and select the right recent query', () => {
```

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,750 @@
...
+              Expect.numberOfQuerySuggestions(customNumberOfSuggestions);
+              Expect.querySuggestionsEquals(expectedQuerySuggestions);
+            });
+          });
+        });
+
+        describe('when a the property #disableRecentQuerySuggestions is set to true', () => {
+          const exampleRecentQueries = ['foo', 'bar'];
+
+          describe('when the local storage contains recent query suggestions', () => {
```
SimonMilord, line 452: maybe overkill but do we want to also test when local storage does not contain recent queries? or no point here?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,750 @@
...
+
+        describe('the aria-activedescendant attribute', () => {
+          describe('when only query suggestions are displayed', () => {
+            beforeEach(() => {
+              setRecentQueriesInLocalStorage([]);
+              visitSearchBox({...defaultOptions, textarea});
+              mockQuerySuggestions(exampleQuerySuggestions);
+            });
+
+            it('the search input should have the id of the heighlighted suggestion as its value for the ariaActivedescendant attribute', () => {
```
SimonMilord, line 567:
```suggestion
            it('the search input should have the id of the highlighted suggestion as its value for the ariaActivedescendant attribute', () => {
```

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,750 @@
...
+          describe('when both query suggestions and recent queries are displayed', () => {
+            const exampleRecentQueries = ['foo'];
+
+            beforeEach(() => {
+              setRecentQueriesInLocalStorage(exampleRecentQueries);
+              visitSearchBox({...defaultOptions, textarea});
+              mockQuerySuggestions(exampleQuerySuggestions);
+            });
+
+            it('the search input should have the id of the heighlighted suggestion as its value for the ariaActivedescendant attribute', () => {
```
SimonMilord, line 612:
```suggestion
            it('the search input should have the id of the highlighted suggestion as its value for the ariaActivedescendant attribute', () => {
```

packages/quantic/cypress/e2e/default-2/search-box/search-box-selectors.ts
```diff
@@ -0,0 +1,43 @@
+import {ComponentSelector, CypressSelector} from '../../common-selectors';
+
+export const standaloneSearchBoxComponent = 'c-quantic-search-box';
+
+export interface SearchBoxSelector extends ComponentSelector {
+  input: (textarea?: boolean) => CypressSelector;
```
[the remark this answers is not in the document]
mmitiche: I think input is a more precise name.

packages/quantic/cypress/e2e/default-2/search-box/search-box-selectors.ts
```diff
@@ -0,0 +1,43 @@
...
+  quanticSearchBoxInput: () => CypressSelector;
```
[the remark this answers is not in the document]
mmitiche: I'm trying to be as precise as I can to avoid any confusion.

packages/quantic/cypress/e2e/default-2/search-box/search-box-actions.ts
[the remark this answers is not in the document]
mmitiche: `.type()` worked like a charm.

packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box-actions.ts
```diff
@@ -38,7 +38,7 @@ const standaloneSearchBoxActions = (selector: StandaloneSearchBoxSelector) => {
     },
     pressEnter: (textarea = false) => {
       selector.input(textarea).then((searchbox) => {
-        cy.wrap(searchbox).trigger('keyup', {key: 'Enter'});
+        cy.wrap(searchbox).trigger('keydown', {key: 'Enter'});
```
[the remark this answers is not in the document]
mmitiche: previously before, this PR, we had  complicated/messy logic that listens to multiple of these events, it was listening for both keydown and keyup events and it was a bit hackish and this is actually one of the factors that made this component complicated.

now I simplified that logic and made our component listen to only keydown events to make it more simple and more natural.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1442,4 +1442,46 @@
...
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>There are no search suggestions.</shortDescription>
+  </labels>
+  <labels>
+    <fullName>quantic_RecentQueryAriaLabel</fullName>
+    <value>Recent query,</value>
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>Recent query,</shortDescription>
```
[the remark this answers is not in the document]
mmitiche: yes this label is used in the icon alternative so the screen reader will read the text depending `recent query` or `query suggestion` on the icon.

and the `,` is made to make the screen reader do a pause while reading the whole thing:
when what's read is `Query suggestion, My suggestion` the screenreader will do a small pause between reading `query suggestion` and `My suggestion`.

when we don't have the `,`, the two sections will be read without pause which sounds a bit weird.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/expandableSearchBoxInput.html
```diff
@@ -27,17 +24,20 @@
...
-                role="textbox"
+                role="combobox"
+                aria-label={labels.searchFieldWithSuggestions}
+                aria-expanded={shouldDisplaySuggestions}
+                aria-autocomplete="list"
+                aria-haspopup="listbox"
                 style="white-space: nowrap"
                 data-cy="search-box-textarea"
               >
+              {value}
```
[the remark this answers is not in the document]
mmitiche: Actually I am programmatically setting `textarea_element.value` in the rendered callback.
this can be removed from the HTML.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -88,6 +101,24 @@ export default class QuanticSearchBox extends LightningElement {
...
+      !this.disableRecentQuerySuggestions &&
+      this.headless.buildRecentQueriesList
+    ) {
+      this.localStorageKey = `${this.engineId}_quantic-recent-queries`;
+      this.recentQueriesList = this.headless.buildRecentQueriesList(engine, {
+        initialState: {
+          queries: getItemFromLocalStorage(this.localStorageKey) ?? [],
+        },
+        options: {
+          maxLength: 100,
```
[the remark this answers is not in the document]
mmitiche: the default in the Headless recent queries is 10 but this does't suit this current use case.
Because we are not just displaying recent queries, but we are displaying recent queries based on the query written in the search box which means we need a larger number of recent queries cause a lot of them will be filtered out based on the current query.

the choice of a 100 is just a large number.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -197,69 +198,81 @@ export default class QuanticSearchBoxInput extends LightningElement {
   }
 
   handleValueChange() {
-    this.sendInputValueChangeEvent(this.input.value, false);
+    this.sendInputValueChangeEvent(this.input.value);
```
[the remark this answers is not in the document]
mmitiche: Actually we were, but it was useless in a way.

That's part of the simplification, now the same logic is offered with a more simple and natural way.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -197,69 +198,81 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
       case keys.ARROWDOWN:
-        this.suggestionListElement?.selectionDown();
+        event.preventDefault();
+        this.ariaActiveDescendant = this.suggestionListElement?.selectionDown();
+        this.input.setAttribute(
+          'aria-activedescendant',
+          this.ariaActiveDescendant
+        );
         break;
       default:
```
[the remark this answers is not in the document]
mmitiche: You are we need to reset the selection somewhere, and I do reset it in `renderedCallback` of the `QuanticSearchBoxSuggestions` component whenever the query changes.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +93,200 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
-        bubbles: true,
-        composed: true,
-      });
-      this.dispatchEvent(suggestionListEvent);
+      this.suggestionsAriaLiveMessage = AriaLiveRegion(
+        'suggestions',
+        this,
+        true
+      );
+      this.sendSuggestionListIdToInput();
```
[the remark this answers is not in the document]
mmitiche: Actually this component renders multiple times, like if there are suggestions we display it otherwise it's completely removed from the DOM.
And yes that's the only way we have so far to deal with this.

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
```diff
@@ -286,8 +283,10 @@ export default class QuanticStandaloneSearchBox extends NavigationMixin(
    */
   selectSuggestion = (event) => {
     event.stopPropagation();
-    const selectedSuggestion = event.detail.selectedSuggestion;
-    this.standaloneSearchBox?.selectSuggestion(selectedSuggestion);
+    const {
+      selectedSuggestion: {value},
+    } = event.detail;
+    this.standaloneSearchBox?.selectSuggestion(value);
```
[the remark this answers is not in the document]
mmitiche: I'm de-structuring `value` from `selectedSuggestion`

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +93,200 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+      ...this.recentQueriesThatStartWithCurrentQuery,
+      ...this.querySuggestionsNotInRecentQueries,
+    ]
+      ?.map(this.buildOption)
+      .slice(0, this.maxNumberOfSuggestions);
+
+    if (this.shouldDisplayRecentQueries) {
+      const clearRecentQueriesOption = {
+        key: 0,
+        id: 'selection-0',
```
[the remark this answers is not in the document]
mmitiche: Of course, so it can be targeted by `aria-activedescendant` and a11y wise screen readers will know that this is the currently selected option.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +93,200 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+  /**
+   * Augments a suggestion with the necessary information needed to display the suggestion as an option in the suggestion list
+   */
+  buildOption = (suggestion, index) => {
+    const optionIndex = this.shouldDisplayRecentQueries ? index + 1 : index;
+    const optionIsSelected = this.selectionIndex === optionIndex;
+
+    return {
+      ...suggestion,
+      id: `selection-${optionIndex}`,
```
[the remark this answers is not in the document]
mmitiche: No it is not, SF will add a suffix and we will use the whole final id (id value + SF suffix) as the aria-activedescendant attribute of the search box input element.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.html
```diff
@@ -1,34 +1,33 @@
...
-                    </template>
-                    <template if:false={suggestion.isSelected}>
-                        <div class="slds-media slds-listbox__option slds-listbox__option_plain slds-media_small" role="option">
-                            <span class="slds-media__figure slds-listbox__option-icon"></span>
-                            <span class="slds-media__body">
-                                <lightning-formatted-rich-text class="slds-truncate" value={suggestion.value}></lightning-formatted-rich-text>
+                    </li>
+                </template>
+                <template lwc:else>
+                    <li id={option.id} key={option.key} onmousedown={option.onClick} tabindex="-1" data-cy="suggestions-option" role="option"
```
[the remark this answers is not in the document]
mmitiche: no it was not doing much in the previous iteration.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -65,50 +67,51 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
   }
+
   disconnectedCallback() {
     this.removeEventListener(
       'suggestionlistrender',
       this.handleSuggestionListEvent
     );
   }
 
+  renderedCallback() {
```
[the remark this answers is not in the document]
mmitiche: This is to keep the value of the input always in sync  with the value in the state of the search box.

if there is an initial value in the state, the value will be automatically reflected in the input.
when a keyboard event is made on the input, an event is dispatched to update the state, when  the state updates this renderedCallback runs and reflect the new query in the input.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -197,69 +198,81 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
-  onKeyup(event) {
+  onKeyDown(event) {
     switch (event.key) {
+      case keys.ESC:
+        this.input.removeAttribute('aria-activedescendant');
+        this.input.blur();
+        break;
+
+      case keys.ARROWRIGHT:
+        // eslint-disable-next-line no-case-declarations
```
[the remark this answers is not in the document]
mmitiche: es-link doesn't like it when you declare a constant inside a `case` section of a `switch` block.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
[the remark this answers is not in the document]
mmitiche: when navigation in the query suggestions, clicking the right arrow will complete the input with the query suggestion you are currently selecting


https://github.com/coveo/ui-kit/assets/86681870/b1424150-77fd-4ab6-bb85-1bfd5fe92175

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -197,69 +198,81 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
+        this.input.setAttribute('aria-activedescendant', selectedOption.id);
+        break;
+
       case keys.ENTER:
         this.handleEnter(event);
         break;
+
       case keys.ARROWUP:
-        this.suggestionListElement?.selectionUp();
+        event.preventDefault();
```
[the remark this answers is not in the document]
mmitiche: I'm specifically adding preventDefault in ArrowUp and Arrowdown because the default behaviour on an input  receiving a keydown event with these two keys is the following:


https://github.com/coveo/ui-kit/assets/86681870/1bfc4206-0630-402b-9a01-81ffa73fc9a7

Notice how the cursor is moving from the start of the input to its end, that's what I am preventing.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
```diff
@@ -171,8 +209,18 @@ export default class QuanticSearchBox extends LightningElement {
    */
   selectSuggestion = (event) => {
     event.stopPropagation();
-    const selectedSuggestion = event.detail.selectedSuggestion;
-    this.searchBox?.selectSuggestion(selectedSuggestion);
+    const {
+      selectedSuggestion: {value, isRecentQuery, isClearRecentQueryButton},
```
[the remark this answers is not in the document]
mmitiche: Yes you are right

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -250,6 +311,68 @@ describe('c-quantic-search-box-input', () => {
...
+
+            const eventData =
+              functionsMocks.exampleSelectSuggestion.mock.calls[0][0];
+            const expectedFirstSuggestionSelected = {
+              isClearRecentQueryButton: undefined,
+              isRecentQuery: undefined,
+              value: mockSuggestions[querySuggestionIndex].rawValue,
+            };
+
+            // @ts-ignore
```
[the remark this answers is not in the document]
mmitiche: removed 👍

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
mmitiche: I would do this separation when we face the need of using one part separately, cause it's very possible that there will be no case where we will  need that.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
mmitiche: While it can be a nice idea to do that I don't necessarily see the value it will add versus the naming rule that will be added up.

packages/quantic/force-app/main/default/lwc/quanticSearchBox/quanticSearchBox.js
mmitiche: who said we no longer want to manage it x) We do want to manage the value in quanticSearchBoxInput, and that's why it's passed a prop as the value in  the `quanticSearchBoxInput` needs always to match the value we have in the state.

i just simplified how we manage this value.

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
mmitiche: yes.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: Nope it doesn't reset automatically.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: to make sure that if we click the first option in that oder we will click a suggestion and long the correct analytics.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: sure good idea!

packages/quantic/cypress/e2e/default-2/standalone-search-box/standalone-search-box-actions.ts
mmitiche: the input component no longer listens to keyup events, that's part of the simplification I added.

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/templates/standaloneSearchBox.html
```diff
@@ -8,6 +8,7 @@
       disable-state-in-url
     >
       <c-quantic-search-box-input
+        value={searchBoxValue}
```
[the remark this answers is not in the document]
mmitiche: https://github.com/coveo/ui-kit/pull/4120#discussion_r1665627615

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
SimonMilord: ok then should we not clear the storage before setting the item?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +93,200 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+    const suggestionSelectedEvent = new CustomEvent('selection', {
+      detail: {selection},
     });
     this.dispatchEvent(suggestionSelectedEvent);
+  };
+
+  /**
+   * Returns the recent queries that start with the query currently typed by the end user.
+   */
+  get recentQueriesThatStartWithCurrentQuery() {
```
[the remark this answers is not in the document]
mmitiche: This is used in multiple places so it becomes heavy to do in multiple places something like:
```
this.myProperty = this.buildMyProperty()
doSomethingWithMyProperty(this.myProperty)
```
With the getter, the step of build is done automatically for us so less code to write.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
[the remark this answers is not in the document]
SimonMilord: IMO a better behaviour here would be to update the value in the searchbox right when you navigate through the suggestions. It feels more intuitive and it is how many searchboxes behave. I dont think the right arrow should be doing anything when navigating through suggestions.

In short:
UP/DOWN ARROWS: navigate through suggestions
LEFT ARROW: go back on the query value in the input to modify query for example
RIGHT ARROW: do nothing
ENTER: Select a suggestion/recent query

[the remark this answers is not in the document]
SimonMilord: @erocheleau @mmitiche what do you guys think?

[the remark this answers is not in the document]
mmitiche: this makes sense!
Other than the behaviour expected for the `TAB`and `shift+ TAB`, clicking on these keys would take you to the next/previous tab-able element, and suggestions are not tab-able elements. 

So clicking on tab/Shift+Tab will completely remove the focus from the search box which means the suggestions are no longer displayed.

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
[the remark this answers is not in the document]
erocheleau: if you take the .text attribute, I think you should be able to get the text and not the formatting tags.

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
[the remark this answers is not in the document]
erocheleau: What kind of timing issues would you expect?

The localStorage should "reset" whenever we want it to reset? for example if you had a 

```
search(query)
...
intercept, validate property `q`;
...
validate localStorage
```
If you make a new test after that that navigates back to the page, the browser context shouldn't be kept AFAIK.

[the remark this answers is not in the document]
erocheleau: Anyway it's not a big deal, but I'm curious why it's so timing sensitive that we have to put them together.

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
```diff
@@ -0,0 +1,146 @@
...
+        .should(display ? 'exist' : 'not.exist')
+        .logDetail(`${should(display)} display the query suggestions list`);
+    },
+    numberOfQuerySuggestions: (value: number) => {
+      selector
+        .querySuggestions()
+        .should('have.length', value)
+        .logDetail(`should display ${value} query suggestions`);
+    },
+    searchInputHasAriaOwnsValue: (textarea = false) => {
```
erocheleau, line 69: Isn't this more of a unit test honnestly?
Do we really need a e2e test for this?

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
```diff
@@ -0,0 +1,146 @@
...
+            const actualValue = inputElement[0].getAttribute('aria-owns');
+            expect(actualValue).to.equal(suggestionListElement[0].id);
+          });
+          return;
+        })
+        .logDetail(
+          'The search input should have the id of the suggestion list as its value for the aria-owns attribute'
+        );
+    },
+    searchInputHasAriaActiveDescendantValue: (
```
erocheleau, line 83: Isn't this more of a unit test honnestly?
Do we really need a e2e test for this?

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
```diff
@@ -0,0 +1,146 @@
...
+              );
+              expect(actualValue).to.equal(suggestionOptionElement[0].id);
+            });
+          return;
+        })
+        .logDetail(
+          `the aria-activedescendant attribute of the searchbox input should have the id of the ${index}-th suggestion as its value`
+        );
+    },
+    searchInputHasNoAriaActiveDescendantValue: (textarea = false) => {
```
erocheleau, line 104: Isn't this more of a unit test honnestly?
Do we really need a e2e test for this?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
erocheleau: Cypress absolutely **should** clear localStorage between tests?
Are you sure it doesn't?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+          const exampleRecentQueries = ['ABC'];
+          const exampleQuerySuggestions = ['ABC'];
+
+          beforeEach(() => {
+            setRecentQueriesInLocalStorage(exampleRecentQueries);
+            visitSearchBox({...defaultOptions, textarea});
+            mockQuerySuggestions(exampleQuerySuggestions);
+          });
+
+          it('should not display the same suggestion twice', () => {
```
erocheleau, line 185: This should be a unit test IMO

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+            ...exampleQuerySuggestions,
+          ];
+
+          beforeEach(() => {
+            setRecentQueriesInLocalStorage(allRecentQuerySuggestions);
+            visitSearchBox({...defaultOptions, textarea});
+            mockQuerySuggestions(exampleQuerySuggestions);
+          });
+
+          it('should only display the recent query suggestions that starts with the value written in the search box input', () => {
```
erocheleau, line 224: This should be a unit test IMO

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+
+            scope('when clicking on the clear recent queries button', () => {
+              Actions.clickClearRecentQueriesButton();
+
+              Expect.logClearRecentQueries();
+              Expect.displayClearRecentQueriesButton(false);
+              Expect.displaySuggestionList(false);
+            });
+
+            scope('when selecting a query suggestion', () => {
```
erocheleau, line 288: This is the kind of stuff we should be writing e2e tests for.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+              Actions.focusSearchBox(textarea);
+              cy.wait(InterceptAliases.QuerySuggestions);
+
+              Expect.displaySuggestionList(true);
+              Expect.displayClearRecentQueriesButton(true);
+              Expect.numberOfQuerySuggestions(expectedQuerySuggestions.length);
+              Expect.querySuggestionsEquals(expectedQuerySuggestions);
+            });
+
+            scope('when selecting a recent query with the keyboard', () => {
```
erocheleau, line 336: This should be the only thing we do in this e2e.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+                }
+              );
+              Expect.logClickRecentQueries(
+                exampleRecentQueries[clickedSuggestionIndex]
+              );
+            });
+          });
+        });
+
+        describe('when using the keyboard to select a query suggestion', () => {
```
erocheleau, line 357: This test should test that when selecting a query suggestion it will execute that search and that's all?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+                }
+              );
+              Expect.logClickSuggestion(
+                exampleQuerySuggestions[clickedSuggestionIndex]
+              );
+            });
+          });
+        });
+
+        describe('when a custom value is set for the property #numberOfSuggestions', () => {
```
erocheleau, line 413: This should also be only a unit test?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+
+              Expect.displaySuggestionList(true);
+              Expect.displayClearRecentQueriesButton(true);
+              Expect.numberOfQuerySuggestions(customNumberOfSuggestions);
+              Expect.querySuggestionsEquals(expectedQuerySuggestions);
+            });
+          });
+        });
+
+        describe('when a the property #disableRecentQueries is set to true', () => {
```
erocheleau, line 449: The only thing we should potentially test with e2e here, 

we should execute a query, and check that it's not showing up in the list when focusing on the searchbox.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+              setRecentQueriesInLocalStorage([]);
+              visitSearchBox({
+                ...defaultOptions,
+                textarea,
+                disableRecentQueries: true,
+              });
+              mockQuerySuggestions(exampleQuerySuggestions);
+            });
+
+            it('should not add the query to the local storage after executing a new search', () => {
```
erocheleau, line 496: Yea only this test, and we should be mostly just looking at the suggestions that pops up when we click on the searchbox, **maybe** the localStorage, but not required.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+        const exampleQuerySuggestions = ['ABC', 'EFG'];
+
+        describe('the aria-owns attribute', () => {
+          beforeEach(() => {
+            setRecentQueriesInLocalStorage([]);
+            visitSearchBox({...defaultOptions, textarea});
+            mockQuerySuggestions(exampleQuerySuggestions);
+          });
+
+          it('the search input should have the id of the suggestion list as its value for the ariaOwns attribute', () => {
```
erocheleau, line 543: Why can't this simply be unit tested?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+              Actions.focusSearchBox(textarea);
+              cy.wait(InterceptAliases.QuerySuggestions);
+
+              Expect.displaySuggestionList(true);
+              Expect.searchInputHasAriaOwnsValue(textarea);
+            });
+          });
+        });
+
+        describe('the aria-activedescendant attribute', () => {
```
erocheleau, line 559: Why can't this simply be unit tested?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,738 @@
...
+
+              scope('when the suggestions list is closed', () => {
+                Actions.blurSearchBox(textarea);
+                Expect.searchInputHasNoAriaActiveDescendantValue(textarea);
+              });
+            });
+          });
+        });
+
+        describe('aria live', () => {
```
erocheleau, line 640: That one is a good one to e2e test because it's independent components.

erocheleau (changes requested):
I feel like there's way too many things being tested as e2e tests here that should be simpler unit tests instead?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -65,50 +67,51 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
-   * @type {string}
+   * @type {String}
    */
-  @api
-  get inputValue() {
-    return this.input.value;
-  }
-  set inputValue(newValue) {
-    this.input.value = newValue;
-  }
```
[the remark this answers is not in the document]
mmitiche: Yes if it was an exposed component, it would be a breaking change. but this component was created just to reduce the complexity of the main QuanticSearchBox and is purely internal.

The current logic we have been following lately is that these internal component are purely for us to better organize  the code and the component composition so changing them would not be considered breaking.

Other wise this would limit us a lot and would discourage us to reduce the complexity of the exposed components by using smaller internal component.

and for this change, yeah I changed the name yeah but that's not the main point I wanted to improve here, the main point is that we had a getter and a setter for this property which were useless, just a normal property can do the same job. that's the simplification I'm doing here.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -65,50 +67,51 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
+  @api value;
   /**
-   * The blur function.
+   * The list containing the recent query suggestions.
    * @api
-   * @type {VoidFunction}
+   * @type {String[]}
    */
-  @api blur() {
```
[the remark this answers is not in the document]
mmitiche: That's a discussion we can have of course, but so far we were considering the internal components things we are free to change as they are not supposed to be used by the user.

and removed this method as part of the simplification as it was also useless,

packages/quantic/force-app/main/default/lwc/quanticStandaloneSearchBox/quanticStandaloneSearchBox.js
```diff
@@ -253,13 +254,9 @@ export default class QuanticStandaloneSearchBox extends NavigationMixin(
...
     event.stopPropagation();
-    const updatedValue = event.detail.newInputValue;
-    const isSelectionReset = event.detail.resetSelection;
-    if (this.standaloneSearchBox?.state?.value !== updatedValue) {
-      if (isSelectionReset) {
-        this.quanticSearchBoxInput.resetSelection();
-      }
-      this.standaloneSearchBox.updateText(updatedValue);
+    const newValue = event.detail.value;
+    if (this.standaloneSearchBox?.state?.value !== newValue) {
```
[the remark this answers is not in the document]
mmitiche: we do need reseting the selections and we are still doing that, just the logic now is directly in the `QuanticSearchBoxInput`  component internally  instead of what we had before having to call `this.quanticSearchBoxInput.resetSelection()`

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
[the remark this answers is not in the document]
mmitiche: made it work with the `.innerText`👍

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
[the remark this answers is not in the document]
mmitiche: The thing is that the Cypress commands are asynchronous, so when you do
```
Expect.myFirstExpectation() // Executes a cypress command to validate something
Expect.mySecondExpectation() // Executes a cypress command to validate something else
```
it doesn't necessarily mean that `myFirstExpectation` will be executed before  `mySecondExpectation`.

in our case, we first start by waiting for the request, then using the `.then()` we validate that the search query was made with the expected parameters and we confirm that the local storage was updated successfully.

if the local storage validation was in another command, it means that, because cypress commands are asynchronous, the validation of local storage could happen before that the request was made which will make the test fail.

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
mmitiche: it can be, mainly why this was not made an e2e is because it includes an interaction between two components and not just one that could be unit tested.

The QuanticSuggestionsList sends a custom events with the `id` to the `QuanticSearchBoxInut` that then sets it as the `aria-owns`.

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
mmitiche: Same previous answer, it's because it's involves more than one component.
But I'm exploring  transforming that to unit tests tho.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: yes, just using the the keyboard.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: if we unit test the situations where we provide custom properties to the Quantic component every single Quantic component test should be updated x)

I see it tho, it's totally legit and makes sense  to unit test that, however the current logic we have with the Quantic tests goes with the opposite direction.

It's a good discussion to have tho, especially with the coming potential migration to playwright, we could take this occasion to adjust our logic to test Quantic components.

In my opinion, unit tests are the most essential tests to have, they are easier to have, robust and very stable.
Previously we were not preferring unit tests, because having always to mock headless was a pain, but we did that in several components in the C4SF and it was not that bad tbh, especially if we take some time make  this logic to mock headless more general and easy to use.

This way we could have much more unit tests and much less e2e tests, e2e tests will only focused on how a user interacts with the components in a search page and sending analytics and search requests.

mmitiche (commented):
pr feedback addressed 👍

packages/quantic/cypress/e2e/default-2/search-box/search-box-expectations.ts
```diff
@@ -0,0 +1,104 @@
...
+          const requestBody = interception.request.body;
+          expect(requestBody).to.have.property('q', expectedQuery);
+          if (expectedRecentQueriesInLocalStorage) {
+            const {LSkey, queries} = expectedRecentQueriesInLocalStorage;
+            const recentQueries = window.localStorage.getItem(LSkey);
+            expect(recentQueries).to.equal(JSON.stringify(queries));
+          }
+        })
+        .logDetail(
+          `should make a search request with the query ${expectedQuery}`
```
erocheleau, line 43: maybe add another log for "expectedQueriesInLocalStorage"?

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,634 @@
...
+  function visitSearchBox(
+    options: Partial<StandaloneSearchBoxOptions> = defaultOptions
+  ) {
+    interceptSearch();
+    cy.visit(pageUrl);
+    configure(options);
+    cy.wait(InterceptAliases.Search);
+  }
+
+  variants.forEach(({variantName, textarea}) => {
```
erocheleau, line 55:
```suggestion
  const variants = [
    {variantName: 'default', textarea: false},
    {variantName: 'expandable', textarea: true},
  ];
  variants.forEach(({variantName, textarea}) => {
```
Do you mind putting the variants declaration right next to its usage? 
It makes it hard to understand having to go up to find what the different `variants` are.

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,634 @@
...
+                {
+                  LSkey: recentQueriesLSKey,
+                  queries: [exampleQuerySuggestions[clickedSuggestionIndex]],
+                }
+              );
+            });
+          });
+        });
+
+        describe('when using the keyboard to select a recent query', () => {
```
erocheleau, line 307: Not a huge deal, but personally I would've built these tests this way.
You don't have to change it.

```
describe('when using the keyboard', () => {
  it('can select recent queries', () => {});
  it('can select query suggestions', () => {});
});
```

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
```diff
@@ -0,0 +1,634 @@
...
+
+        describe('aria live', () => {
+          describe('when only query suggestions are displayed', () => {
+            beforeEach(() => {
+              setRecentQueriesInLocalStorage([]);
+              visitSearchBox({...defaultOptions, textarea});
+              mockQuerySuggestions(exampleQuerySuggestions);
+            });
+
+            it('should have the aria live message the properly the number of query suggestions available', () => {
```
erocheleau, line 544: I think I said before, but this should've been a unit test instead of e2e.
The unit test should test that we're correctly "sending" the message to the aria-live region.
Since we already have unit tests on aria-live that it displays messages it receives, we're covered.

I think it's a bit overkill to test e2e the behavior of both together.
But don't remove it, just keep that in mind for the future.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -65,50 +67,51 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
+   * @type {String}
    */
-  @api
-  get inputValue() {
-    return this.input.value;
-  }
-  set inputValue(newValue) {
-    this.input.value = newValue;
-  }
+  @api value;
```
erocheleau, line 74: Still need to change this back to `inputValue` to not be a breaking change without justification.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
erocheleau, line 84: Don't delete it just for fun.
If possible still make it work to blur the input when this is called.
And mark it by a comment that it's `@deprecated` and create a jira to remove it in v3.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
```diff
@@ -65,50 +67,51 @@ export default class QuanticSearchBoxInput extends LightningElement {
...
+  @api recentQueries;
   /**
-   * The reset selection function.
+   * The maximum number of suggestions to display.
    * @api
-   * @type {VoidFunction}
+   * @type {number}
+   * @defaultValue 7
    */
-  @api resetSelection() {
```
erocheleau, line 92: Same as above.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -180,24 +191,75 @@ describe('c-quantic-search-box-input', () => {
...
+            expect(suggestionsListItems.length).toEqual(mockSuggestions.length);
           });
-          await flushPromises();
+        });
 
-          const suggestionsList = element.shadowRoot.querySelector(
-            selectors.searchBoxSuggestionsList
-          );
-          expect(suggestionsList).not.toBeNull();
+        describe('with both query suggestions and recent queries available', () => {
```
erocheleau, line 225: This is where you could also have a unit test to test that if you have the same query in recent and in query suggest, it's not displayed twice

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
```diff
@@ -250,27 +312,147 @@ describe('c-quantic-search-box-input', () => {
...
+            )[querySuggestionIndex];
             expect(firstSuggestion).not.toBeNull();
 
-            firstSuggestion.click();
+            firstSuggestion.dispatchEvent(new CustomEvent('mousedown'));
+            expect(
+              functionsMocks.exampleSelectSuggestion
+            ).toHaveBeenCalledTimes(1);
+
+            /** @type{object} */
```
erocheleau, line 340: why do you need this?

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
erocheleau, line 1: Amazing unit tests added here! GG

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
```diff
@@ -41,87 +89,205 @@ export default class QuanticSearchBoxSuggestionsList extends LightningElement {
...
+      };
     }
-    return undefined;
-  }
-
-  /**
-   * Reset the current selection.
-   */
-  @api
-  resetSelection() {
```
erocheleau, line 65: If we can keep this it would be great. Even if we don't use it ourselves.

Mark as @deprecated, and make a jira to remove it for v3

erocheleau (changes requested):
Really almost there, I would accept it as that if you re-add the breaking @api changes.
Some suggestions if you want to accept them.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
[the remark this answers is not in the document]
erocheleau: It's more that even for an internal component, I wouldn't do a breaking change just to rename a prop for fun.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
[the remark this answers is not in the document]
erocheleau: I commented else where, but I would do it in two step.

Keep it for now, tag it with `@deprecated` in the jsdoc comment.

And in v3 remove it.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxSuggestionsList/quanticSearchBoxSuggestionsList.js
mmitiche: added back and marked as deprecated.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
mmitiche: added back and marked as deprecated.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/quanticSearchBoxInput.js
mmitiche: added back.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
mmitiche: In order to avoid having typing issues when trying to check the payload of the event.
Improved that the type here to make it more precise and more representative about the expected payload.

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
mmitiche: Indeed, this will be part of this: https://coveord.atlassian.net/browse/SFINT-5645

packages/quantic/cypress/e2e/default-2/search-box/search-box.cypress.ts
mmitiche: I will also try to address that when adding more unit tests to this components 👍

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/__tests__/quanticSearchBoxInput.test.js
erocheleau: I mean, stating it's a "object" is pretty much the same as disabling ts checking for this object :D 

It's all good though.

erocheleau (approved):
Very nice! Finally we got there ;) GG

packages/quantic/force-app/main/default/lwc/quanticSearchBoxInput/templates/expandableSearchBoxInput.html
```diff
@@ -27,14 +24,16 @@
...
-                onkeydown={onKeydown}
-                onkeyup={onKeyup}
+                onkeydown={onKeyDown}
                 onfocus={onFocus}
                 onblur={onBlur}
                 oninput={onTextAreaInput}
-                aria-autocomplete="list"
                 autocomplete="off"
-                role="textbox"
+                role="combobox"
```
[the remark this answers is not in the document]
mmitiche: It can work but it comes with an additional complexity to the component.

PR #4121, opened 2024-06-26
feat(commerce): add atomic-commerce-breadbox 

fpbrault:
This PR adds the `atomic-commerce-breadbox` component. It also refactors `atomic-breadbox` to use the same functional components added for `atomic-commerce-breadbox`.

image

https://coveord.atlassian.net/browse/KIT-3196


packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
```diff
@@ -0,0 +1,435 @@
...
+    return (
+      this.breadcrumbManagerState.facetBreadcrumbs.filter(
+        ({facetId, type}) =>
+          this.bindings.store.state.facets[facetId] && type === 'regular'
+      ) as Breadcrumb<RegularFacetValue>[]
+    )
+      .map(({facetId, field, values, type}) =>
+        values.map((value) => ({value, facetId, field, type}))
+      )
+      .flat()
```
Spuffynism, line 247: I think we could merge `map` with `flat` into one `flatMap` call

packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
```diff
@@ -0,0 +1,435 @@
...
+    return [
+      ...this.facetBreadcrumbs,
+      ...this.categoryFacetBreadcrumbs,
+      ...this.numericFacetBreadcrumbs,
+      ...this.dateFacetBreadcrumbs,
+    ];
+  }
+
+  private renderBreadcrumbs(allBreadcrumbs: IBreadcrumb[]) {
+    const sortedBreadcrumbs = allBreadcrumbs.sort((a, b) => {
```
Spuffynism, line 345: Instead of building all breadcrumbs by type, and then sorting them, could we iterate over `this.breadcrumbManagerState.facetBreadcrumbs` and build them on the fly? I don't see the private `facetBreadcrumbs`, `categoryFacetBreadcrumbs`, `numericFacetBreadcrumbs` and `dateFacetBreadcrumbs` used elsewhere.

It's somewhat how we render breadcrumbs in barca: https://github.com/coveo/barca-sports/blob/672c05c55c02324d3db2381a5da8c1a32496dc8f/Components/Commerce/Breadcrumb.tsx#L42

packages/atomic/src/components/common/breadbox/breadbox.tsx
```diff
@@ -0,0 +1,241 @@
+import {FacetValueState} from '@coveo/headless';
```
olamothe, line 1: For the sake of uniformity, if you compare with other recently refactor functional components, they are all put into separate files that matches the component name.

We should have the same structure here.

```
common/breadcrumb/container.tsx
     export const BreadcrumbContainer: FunctionalComponent<...>

common/breadcrumb/showless.tsx
     export const BreadcrumbShowLess: FunctionalComponent<...>
```

It also makes it so that for a new use case, we could pick and chose "only some" functional components, and replace some other by more specific ones. By putting them all in the same files, its less obvious.

packages/atomic/src/components/common/breadbox/breadbox.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+export interface BreadcrumbButtonProps {
+  pathLimit: number;
+  onSelectBreadcrumb: () => void;
+  focusTargets: BreadboxFocusTargets;
+  lastRemovedBreadcrumbIndex: number | undefined;
+  firstExpandedBreadcrumbIndex: number | undefined;
+  breadcrumb: IBreadcrumb;
+  index: number;
+  bindings: AnyBindings;
```
olamothe, line 47: Again for uniformity, other recently refactor component only requires `i18n`, not the whole "bindings" object.

Apart from uniformity, requiring a "large, complex type" (as in bindings is a lot of different sub object and types), when we end up only using one property (i18n) causes portability issue.

ie: It's harder to reuse a functional component in a different setup when that requires a lot of complex props, as opposed to only sniping those absolutely necessary.

packages/atomic/src/components/common/breadbox/breadbox.tsx
```diff
@@ -0,0 +1,241 @@
+import {FacetValueState} from '@coveo/headless';
+import {h, VNode, FunctionalComponent} from '@stencil/core';
+import CloseIcon from '../../../images/close.svg';
+import {FocusTargetController} from '../../../utils/accessibility-utils';
+import {Button} from '../button';
+import {AnyBindings} from '../interface/bindings';
+
+export interface IBreadcrumb {
```
olamothe, line 8: We don't normally prefix interface with `I` in ui-kit.

packages/atomic/src/components/common/breadbox/breadbox.tsx
olamothe, line 1: This is not something we can do: Importing from `@coveo/headless` in a common component.

I will touch upon this in a separate comment. It's related to the interfaces for functional components, and them doing too much.

packages/atomic/src/components/common/breadbox/breadbox.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  if (pathLimit === 1 && path.length > 1) {
+    return [ELLIPSIS, path[path.length - 1]].join(SEPARATOR);
+  }
+
+  const ellipsedPath = [path[0], ELLIPSIS, ...path.slice(-(pathLimit - 1))];
+  return ellipsedPath.join(SEPARATOR);
+}
+
+export interface BreadcrumbButtonProps {
```
olamothe, line 39: This interface is very large: This points to a problem that this component (the button) is doing too much.

* `focusTargets` is un-necessary: what it can do instead is resurface the `ref` prop to the parent.
* Related to the above, `lastremoved`/`firstExpanded` then does not need to leak here, and can be handled by the top level web component.
* The content should be a separate component, and button could just output it's `children`.
* `facetId` does not seem to be used anywhere
* `deselect` does not seem to be used anywhere

olamothe: This comment potentially applies to all components in this file;

olamothe (changes requested):
We need to take great care when doing those refactor, and making sure the interface for all common/functional component is clean.

Interfaces should be small and focused, and it's better to "explode" functional component into lots of smaller focused ones that can be embedded into one another.

Another way to think about it is that functional components should be "as flat as possible" in an individual manner; It is by embedding them into one another that complexity can be achieved.

Otherwise it makes the reuse/adaptation/tweaking of those components very difficult in the future, as was the case when the initial work for commerce started.

packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
```diff
@@ -0,0 +1,420 @@
...
+            breadcrumb.facetId
+          ]?.label(),
+        deselect: value.deselect,
+        formattedValue: [
+          this.valueForFacetType(
+            breadcrumb.type,
+            breadcrumb.facetId,
+            value as BreadcrumbValue<AnyFacetValue>
+          ),
+        ],
```
alexprudhomme, line 308: The bug referenced in my other comment is caused by this part of the code. You need to map all the `path` values together to form `formattedValue`

image

alexprudhomme (approved):

olamothe (approved):
💯 GG !

Spuffynism (approved):

PR #4123, opened 2024-06-26
test(atomic): add atomic-commerce-facet tests

olamothe:
Add tests for `atomic-commerce-facet`.

You will see some shenanigan in story setup to "keep only a single facet".

As you know, atomic-commerce-facet are generated/internal: So we need to trick storybook a bit to output `atomic-commerce-facets` (with an `s`), and then "snipe" components that are not of the correct type (ie: everything that is not an `atomic-commerce-facet`).

This works pretty well for the test themselves. We get a page with only a specific type of facet, and we can run test as normal. 

There will be work needed still to make storybook work correctly with the Shadow parts.

The shadow parts are properly documented; It's just that when they get customized in Storybook UI, the styling output is not reflected correctly in the story.

However, I prefer to continue with testing and revisit that at a later time.

https://coveord.atlassian.net/browse/KIT-3257


packages/atomic/src/components/commerce/facets/atomic-commerce-facet/atomic-commerce-facet.new.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <atomic-commerce-facets></atomic-commerce-facets>
+      </div>`;
+    },
+  ],
+  play: async (context) => {
+    await play(context);
+    await playExecuteFirstSearch(context);
+    playKeepOnlyFirstFacetOfType('atomic-commerce-facet', context);
+  },
+};
```
alexprudhomme, line 39: This component should never be used by users right ? The only reason to make a story for this is to test the individual component that gets automatically rendered and to show the  shadow parts. I think this is confusing for the documentation to have both `Facets/atomic-automatic-facets` and `Facet/atomic-automatic-facet`. I think we should both put them and in the same `Facets` folder and name this story "Shadow Parts" or something.

packages/atomic/storybookUtils/commerce-interface-wrapper.tsx
```diff
@@ -7,7 +7,7 @@ import {Decorator, StoryContext} from '@storybook/web-components';
 import {html} from 'lit/static-html.js';
 import type * as _ from '../src/components.d.ts';
 
-const getSampleCommerceEngineConfiguration =
+export const getSampleCommerceEngineConfiguration =
```
alexprudhomme, line 10: Why is this exported ?

packages/atomic/src/components/commerce/facets/atomic-commerce-facet/atomic-commerce-facet.new.stories.tsx
olamothe: We can tag some story "test only"

fpbrault: @alexprudhomme btw atomic-automatic-facets is [supposed](https://docs.google.com/document/d/1oXGc9dJqBSQjlw7pLD6eolEuPvCyMUcAlyOhedbr78E/edit) to end up in `Interface Components/atomic-automatic-facets` later instead of in the Facets folder.

packages/atomic/storybookUtils/commerce-interface-wrapper.tsx
olamothe: Not needed; Forgot to cleanup

packages/atomic/src/components/commerce/facets/atomic-commerce-facet/atomic-commerce-facet.new.stories.tsx
alexprudhomme: All right, I understand we will review the documentation side later. This is focused on tests.

alexprudhomme (approved):
lgtm

PR #4126, opened 2024-06-27
feat(headless): retrieve client ID from navigator context provider

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3351

We now instead rely on the default browser navigator context provider to retrieve the client ID from Relay in a client-side rendering scenario.

In a server-side rendering scenario, the implementer would be responsible for retrieving the client ID from the request headers (or generating one), returning this value in their navigator context provider function, as well as in the response headers.

This reverts most of the changes I made in https://github.com/coveo/ui-kit/pull/4111.

I also applied the same logic to the search engine, which reverts many of the changes made by @louis-bompart in https://github.com/coveo/ui-kit/pull/3820. These changes were made in a separate commit, however, so we can easily revert them and keep the scope to headless/commerce if we prefer.


Spuffynism (approved):
LGTM! Not sure about the non-commerce use cases though

louis-bompart (approved):

PR #4124, opened 2024-06-26
feat(headless commerce): create use & export commerce-specific actions & loaders

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3207

This PR is starting to grow huge, so I'll put it in review.

I linked another Jira in KIT-3207. I'll open a few other PRs to do the rest of the work.

PR #4128, opened 2024-06-27
fix(commerce): expose child product directly on `promoteChildToParent`

Spuffynism:
Instead of mutating the received products from the API, I propose we dynamically find the corresponding parent and children in the reducer. This _seems_ less error-prone to me, because we don't need to mutate products in every place we fulfill a request containing `products`, at the cost of being less performant, as we iterate on every product every time a children is promoted. Thoughts?

An other option to consider would be to mutate the products directly in the commerce api client.

https://coveord.atlassian.net/browse/KIT-3221

### DIFF
--- a/packages/atomic/src/components/commerce/atomic-commerce-recommendation-interface/store.ts
+++ b/packages/atomic/src/components/commerce/atomic-commerce-recommendation-interface/store.ts
@@ -1,3 +1,4 @@
+import {ChildProduct} from '@coveo/headless/commerce';
 import {DEFAULT_MOBILE_BREAKPOINT} from '../../../utils/replace-breakpoint';
 import {
   AtomicCommonStore,
@@ -9,7 +10,7 @@ import {makeDesktopQuery} from '../atomic-commerce-layout/commerce-layout';
 export interface AtomicStoreData extends AtomicCommonStoreData {
   mobileBreakpoint: string;
   currentQuickviewPosition: number;
-  activeProductChild: {parentPermanentId: string; childPermanentId: string};
+  activeProductChild: ChildProduct | undefined;
 }
 
 export interface AtomicCommerceRecommendationStore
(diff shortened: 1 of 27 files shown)


packages/headless/src/features/commerce/instant-products/instant-products-slice.ts
```diff
@@ -69,28 +73,23 @@ export const instantProductsReducer = createReducer(
         }
         const products = cache.products;
 
-        const currentParentIndex = products.findIndex(
-          (product) => product.permanentid === action.payload.parentPermanentId
-        );
+        let childToPromote;
+        const currentParentIndex = products.findIndex((product) => {
```
fbeaudoincoveo, line 77: My concern with this approach would be that in cases where lots of products are being displayed (which I think can happen in commerce as the user "browses" a list with the show more button), this could get slower and slower.

I do like the simplicity of this approach, however.

I think we can go ahead with this for now, and keep this on our radar as a potential spot for performance optimization.

packages/atomic/src/components/commerce/atomic-commerce-recommendation-interface/store.ts
```diff
@@ -9,7 +10,7 @@ import {makeDesktopQuery} from '../atomic-commerce-layout/commerce-layout';
 export interface AtomicStoreData extends AtomicCommonStoreData {
   mobileBreakpoint: string;
   currentQuickviewPosition: number;
-  activeProductChild: {parentPermanentId: string; childPermanentId: string};
```
fbeaudoincoveo, line 12: You'll have to adjust the listener in atomic-commerce-recommendation-list.tsx

packages/headless/src/controllers/commerce/search/headless-search.ts
```diff
@@ -62,13 +62,9 @@ export interface Search extends Controller, SearchSubControllers {
    * **Note:** In the controller state, a product that has children will always include itself as its own child so that
    * it can be rendered as a nested product, and restored as the parent product through this method as needed.
    *
-   * @param childPermanentId The permanentid of the child product that will become the new parent.
-   * @param parentPermanentId The permanentid of the current parent product of the child product to promote.
+   * @param child The product that will become the new parent.
```
fbeaudoincoveo, line 65:
```suggestion
   * @param child The child product that will become the new parent.
```

fbeaudoincoveo: Same comment everywhere this docs is duplicated

fbeaudoincoveo (approved):
Nice!

I'll take care of exposing the actions in the proper loaders once https://github.com/coveo/ui-kit/pull/4124 is merged

PR #4129, opened 2024-06-27
feat(answer): knowledge controller preparation

dmgauthier:
## Summary

This change is just moving logic around; this increment is not modifying any functionalities.

The change intents to prepare the addition of a new answer controller that will leverage the new answerApi client. The end goal is to separate the legacy logic of using the streamId from the searchApi, from the new logic of the answerApi stream.  

## Why is this needed

Since both logics (search-api VS answer-api) share a strong common root, the current change will permit us to reuse the current headless-core-generated-answer controller in both cases.

## Testing

Those change must be tested to ensure that the current functionalities stay unchanged


packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
```diff
@@ -34,8 +35,8 @@ export function buildGeneratedAnswer(
   props: GeneratedAnswerProps = {}
 ): GeneratedAnswer {
   const {dispatch} = engine;
-  const controller = buildCoreGeneratedAnswer(
-    engine,
+  const controller = buildSearchAPIGeneratedAnswer(
+    engine as unknown as SearchEngine, // TODO: types
```
dmgauthier, line 39: This cast is disgracious. a better solution seems necessary.

pathiery (approved):

lvu285 (approved):

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
dmgauthier: Cannot find any better solution than to cast here.

alexprudhomme: are we eventually going to remove the buildSearchApiGeneratedAnswer and only use the answer api ?

alexprudhomme (approved):

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
dmgauthier: This is a good question. I think the end goal is yes. But lets admit that this may take quite a while since we will need all users leave the current flow. Which probably mean all our customer migrate to V3, it may (and likely will) take quite a while.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,171 @@
...
+export function buildSearchAPIGeneratedAnswer(
+  engine: SearchEngine,
+  analyticsClient: SearchAPIGeneratedAnswerAnalyticsClient,
+  props: GeneratedAnswerProps = {}
+): SearchAPIGeneratedAnswer {
+  if (!loadGeneratedAnswerReducer(engine)) {
+    throw loadReducerError;
+  }
+
+  const {dispatch} = engine;
```
fbeaudoincoveo, line 125:
```suggestion
```

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,171 @@
...
+    dispatch(setId({id: genQaEngineId}));
+    subscribeStateManager.engines[genQaEngineId] = {
+      abortController: undefined,
+      lastRequestId: '',
+      lastStreamId: '',
+    };
+  }
+
+  subscribeStateManager.subscribeToSearchRequests(engine);
+
```
fbeaudoincoveo, line 140:
```suggestion

  const isSearchEngine = (
    engine: SearchEngine | InsightEngine
  ): engine is SearchEngine => {
    return 'executeFirstSearchAfterStandaloneSearchboxRedirect' in engine;
  };
```

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,171 @@
...
+    get state() {
+      return getState().generatedAnswer;
+    },
+    retry() {
+      dispatch(
+        executeSearch({
+          legacy: analyticsClient.logRetryGeneratedAnswer(),
+        })
+      );
+    },
```
fbeaudoincoveo, line 153:
```suggestion
    retry() {
      if (!isSearchEngine(engine)) {
        return;
      }
      engine.dispatch(
        executeSearch({
          legacy: analyticsClient.logRetryGeneratedAnswer(),
        })
      );
    },
```

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    rephrase(responseFormat: GeneratedResponseFormat) {
+      controller.rephrase(responseFormat);
+      dispatch(
+        executeSearch({
+          legacy: analyticsClient.logRephraseGeneratedAnswer(responseFormat),
+        })
+      );
+    },
```
fbeaudoincoveo, line 162:
```suggestion
    rephrase(responseFormat: GeneratedResponseFormat) {
      controller.rephrase(responseFormat);
      if (!isSearchEngine(engine) {
        return;
      }
      engine.dispatch(
        executeSearch({
          legacy: analyticsClient.logRephraseGeneratedAnswer(responseFormat),
        })
      );
    },
```

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Creates a `GeneratedAnswer` controller instance using the search api stream pattern.
+ *
+ * @param engine - The headless engine.
+ * @param props - The configurable `GeneratedAnswer` properties.
+ * @returns A `GeneratedAnswer` controller instance.
+ */
+export function buildSearchAPIGeneratedAnswer(
+  engine: SearchEngine,
```
fbeaudoincoveo, line 117:
```suggestion
  engine: SearchEngine | InsightEngine,
```

and import it

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
fbeaudoincoveo: With the suggestions I made above I think you should be able to get rid of the cast.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
fbeaudoincoveo: If you apply those suggestions, make sure to adjust the tests accordingly!

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
```diff
@@ -241,8 +139,29 @@ export interface GeneratedAnswerAnalyticsClient {
   logGeneratedAnswerCollapse: () => CustomAction;
 }
 
+export interface GeneratedAnswerProps {
+  initialState?: {
```
fbeaudoincoveo, line 143: I think we should define the type of `initialState` in a separate interface (e.g., `GeneratedAnswerPropsInitialState`) rather than inline here.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
dmgauthier: Thanks a lot for the narrowing @fbeaudoincoveo . This is indeed exactly what I needed

packages/headless/src/controllers/insight/generated-answer/headless-insight-generated-answer.ts
dmgauthier: This narrowing is way better than this ugly cast. 

thanks a lot.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
dmgauthier: Since its the searchEngine that is mocked, all the test are passing. But even if it was not so, I think no tests would fail. 

The tests for the original controller, before the refactor, were already not testing that the search was triggered and I havent added any myself. I suppose there was things that were difficult to test, if this was not covered in the first place. And as the narrowing will not prevent the main controller action but only the search request from being executed, there is nothing impacting the tests as they are already implemented.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,184 @@
...
+    };
+    return engine.subscribe(strictListener);
+  },
+};
+
+export interface SearchAPIGeneratedAnswerAnalyticsClient
+  extends GeneratedAnswerAnalyticsClient {}
+
+/**
+ * Creates a `GeneratedAnswer` controller instance using the search api stream pattern.
```
fbeaudoincoveo, line 111:
```suggestion
 * Creates a `GeneratedAnswer` controller instance using the search API stream pattern.
```

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
fbeaudoincoveo: You resolved this comment but I don't see the change 😅

fbeaudoincoveo (approved):

packages/headless/src/controllers/core/generated-answer/headless-core-generated-answer.ts
dmgauthier: https://github.com/coveo/ui-kit/pull/4129/files#diff-7409be0b72c2931d8b6f67a543483583ca497601bafd32145eaebab92ec8bc45R142

Did I resolve before my push ?

That would be just like me.

dmgauthier: Or else you meant doing it another way.

pathiery (approved):

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,184 @@
...
+export function buildSearchAPIGeneratedAnswer(
+  engine: SearchEngine | InsightEngine,
+  analyticsClient: SearchAPIGeneratedAnswerAnalyticsClient,
+  props: GeneratedAnswerProps = {}
+): SearchAPIGeneratedAnswer {
+  if (!loadGeneratedAnswerReducer(engine)) {
+    throw loadReducerError;
+  }
+
+  // const {dispatch} = engine;
```
mmitiche, line 126: Do we need this comment?

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,184 @@
...
+    retry() {
+      if (!isSearchEngine(engine)) {
+        return;
+      }
+      engine.dispatch(
+        executeSearch({
+          legacy: analyticsClient.logRetryGeneratedAnswer(),
+        })
+      );
+    },
```
mmitiche, line 163: out of curiosity, Is there a particular reason why we are using this pattern of treating the case of the searchEngine in the core controller and returning null for other use cases or it's just a coding preference?

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,184 @@
...
+    retry() {
+      if (!isSearchEngine(engine)) {
+        return;
+      }
+      engine.dispatch(
+        executeSearch({
+          legacy: analyticsClient.logRetryGeneratedAnswer(),
+        })
+      );
```
mmitiche, line 162: I don't see in the unit test of the headless-seaerchapi-generated-answer that we are testing this case where the engine is a search engine and the `executeSearch` action needs to be dispatched.
I think it would be useful to have such a test.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
```diff
@@ -0,0 +1,184 @@
...
+    rephrase(responseFormat: GeneratedResponseFormat) {
+      controller.rephrase(responseFormat);
+      if (!isSearchEngine(engine)) {
+        return;
+      }
+      engine.dispatch(
+        executeSearch({
+          legacy: analyticsClient.logRephraseGeneratedAnswer(responseFormat),
+        })
+      );
```
mmitiche, line 174: same thing for the rephrase action, do we wanna unit-test that?

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
mmitiche, line 1: Why are we removing these tests? I see that the `retry` and `rephrase` methods were removed from the headless-generated-answer controller however the behaviour when these methods are called didn't change as the logic is now defined in the core controller.

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -580,7 +580,8 @@ describe('quantic-generated-answer', () => {
               visitGeneratedAnswer({useCase: param.useCase});
             });
 
-            it(`should send a new search query with the rephrase option ${option} as a parameter`, () => {
+            // TODO: unskip [SVCC-3956](https://coveord.atlassian.net/browse/SVCC-3956)
+            it.skip(`should send a new search query with the rephrase option ${option} as a parameter`, () => {
```
mmitiche, line 584: why are we skipping this test?

mmitiche (commented):
The refactor looks good!
I just have some questions and maybe some tests to add.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
dmgauthier: No, I wanted to remove it in a next PR to avoid re running the CI, but might as well do it right now.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
dmgauthier: this is type narrowing to avoid having to cast the engine type.  I would prefer not doing so while fighting with typescript but the `as unknown as engineType` is even more ugly than this pattern.

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
dmgauthier: Yes. I will add a line to the rephrase test.
``` 
 it('dispatches a rephrase action', () => {
    const generatedAnswer = createGeneratedAnswer();
    const responseFormat: GeneratedResponseFormat = {answerStyle: 'step'};
    generatedAnswer.rephrase(responseFormat);
    expect(updateResponseFormat).toHaveBeenCalledWith(responseFormat);
    expect(executeSearch).toHaveBeenCalled();
  });
```

packages/headless/src/controllers/core/generated-answer/headless-searchapi-generated-answer.ts
dmgauthier: Yes, I did add a whole test for `retry` too. (reply and rephrase are tested for executeSearch now)

packages/headless/src/controllers/generated-answer/headless-generated-answer.test.ts
dmgauthier: If you look well, the main core controller still handle a rephrase and a retry, but will not systematically execute search while retrying or rephrasing. The new knowledge controller will not handle the request that way. So there are some instance where the top level `headless-generated-answer` (whenever a answer config ID is passed) will not execute a search.

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
dmgauthier: Because of the type narrowing. The test will not have the proper engine and the test will never execute the search query.

To my defence the cypress setup is so convoluted and hard to navigate that I am afraid to not reach the deadline to merge everything while trying to understand a deprecated setup (we are migrating the tests to playwright).
Before going GA with the feature, I will make sure to give everything the coverage it deserves.

SimonMilord (approved):

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
mmitiche: my concern is more about the feature tho and not the tests, will the rephrase feature be broken after making the changes in the PR since there are no longer a search query executed here?

My concern is mainly about the following, in the `headless-searchapi-generated-answer.ts` you have the following:
```
    rephrase(responseFormat: GeneratedResponseFormat) {
      controller.rephrase(responseFormat);
      if (!isSearchEngine(engine)) {
        return;
      }
      engine.dispatch(
        executeSearch({
          legacy: analyticsClient.logRephraseGeneratedAnswer(responseFormat),
        })
      );
    },
```

we first execute `controller.rephrase(responseFormat);` , then if the Engine is not a searchEngine we stop
other wise if the engine is a search engine we dispatch `executeSearch`.

So coming back to the test, the test is running with a search engine, meaning that the `executeSearch` should be dispatched, which also means that theoretically everything  will keep working as it was before and the e2e mentioned above wouldn't fail and wouldn't need to be skipped.

dmgauthier: This is all legitimate. And turns out the tests shouldn't be skipped. As you just said, they are using the search engine, so there was no reason for the tests to fail, or put it another way, they failed for a good reason but I had had no confidence in them. I was wrong.

There was a typo in the type check function, turns out it worked to get rid of the casting, but did not work in the real implementation :grimacing: 

Long story short, I fixed the typo and un-skipped the tests.

mmitiche: Perfect! 👌just waiting your latest commit to approve!

mmitiche (approved):
nice work!

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
dmgauthier: Amended !

PR #4130, opened 2024-06-28
feat(headless): add the capability to pass a answerConfigurationId to the insight engine 

dmgauthier:
## Summary

Addition of the capability to pass a `answerConfigurationId` to the engine when initiating it. When the `answerConfigurationId` is passed, the configuration state will be updated with it. 

## Why is this needed

When an interface will initiate the engine with an `answerConfigurationId`, the engine must dispatch the id to the configuration state.
- To be used by the AnswerApi
- To flip the switch between the legacy Generated Answer Client and the new one
- To flip the switch between the controller leveraging the Search API pattern and the Controller using the new Stream

### DIFF
--- a/packages/headless/src/app/search-engine/search-engine.ts
+++ b/packages/headless/src/app/search-engine/search-engine.ts
@@ -1,6 +1,7 @@
 import {StateFromReducersMapObject} from '@reduxjs/toolkit';
 import {Logger} from 'pino';
-import {GeneratedAnswerAPIClient} from '../../api/generated-answer/generated-answer-client';
+import {LegacyGeneratedAnswerAPIClient} from '../../api/generated-answer/generated-answer-client';
+import {answerApi} from '../../api/knowledge/stream-answer-api';
 import {NoopPreprocessRequest} from '../../api/preprocess-request';
 import {SearchAPIClient} from '../../api/search/search-api-client';
 import {
@@ -18,6 +19,7 @@ import {
 } from '../../features/analytics/analytics-actions';
 import {LegacySearchAction} from '../../features/analytics/analytics-utils';
 import {
+  updateAnswerConfiguration,
   updateSearchConfiguration,
   UpdateSearchConfigurationActionCreatorPayload,
 } from '../../features/configuration/configuration-actions';
@@ -111,20 +113,29 @@ export interface SearchEngineOptions
 export function buildSearchEngine(options: SearchEngineOptions): SearchEngine {
   const logger = buildLogger(options.loggerOptions);
   validateConfiguration(options.configuration, logger);
-
   const searchAPIClient = createSearchAPIClient(options.configuration, logger);
-  const generatedAnswerClient = createGeneratedAnswerAPIClient(logger);
 
   const thunkArguments = {
     ...buildThunkExtraArguments(options.configuration, logger),
     apiClient: searchAPIClient,
-    streamingClient: generatedAnswerClient,
+    ...(options.configuration.answerConfigurationId
+      ? {}
+      : {streamingClient: createLegacyGeneratedAnswerAPIClient(logger)}),
   };
 
-  const augmentedOptions: EngineOptions<SearchEngineReducers> = {
+  type EngineOptionsWithAnswerMiddlewares = Omit<
+    EngineOptions<SearchEngineReducers>,
+    'middlewares'
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  > & {middlewares?: any};
+
+  const augmentedOptions: EngineOptionsWithAnswerMiddlewares = {
     ...options,
     reducers: searchEngineReducers,
     crossReducer: jwtReducer(logger),
+    ...(options.configuration.answerConfigurationId
+      ? {middlewares: [answerApi.middleware]}
+      : {}),
   };
 
   const engine = buildEngine(augmentedOptions, thunkArguments);
@@ -133,6 +144,14 @@ export function buildSearchEngine(options: SearchEngineOptions): SearchEngine {
 
   if (search) {
     engine.dispatch(updateSearchConfiguration(search));
+  }
+
+  if (options.configuration.answerConfigurationId) {
+    engine.dispatch(
+      updateAnswerConfiguration({
+        answerConfigurationId: options.configuration.answerConfigurationId,
+      })
+    );
   }
 
   return {
@@ -208,8 +227,8 @@ function createSearchAPIClient(
   });
 }
 
-function createGeneratedAnswerAPIClient(logger: Logger) {
-  return new GeneratedAnswerAPIClient({
+function createLegacyGeneratedAnswerAPIClient(logger: Logger) {
+  return new LegacyGeneratedAnswerAPIClient({
     logger,
   });
 }
--- a/packages/headless/src/features/configuration/configuration-actions.ts
+++ b/packages/headless/src/features/configuration/configuration-actions.ts
@@ -37,6 +37,13 @@ export interface UpdateBasicConfigurationActionCreatorPayload {
   platformUrl?: string;
 }
 
+interface UpdateAnswerConfigurationActionCreatorPayload {
+  /**
+   * Specifies the unique identifier for a  answer configuration.
+   */
+  answerConfigurationId: string;
+}
+
 export const updateBasicConfiguration = createAction(
   'configuration/updateBasicConfiguration',
   (payload: UpdateBasicConfigurationActionCreatorPayload) =>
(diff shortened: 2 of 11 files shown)


packages/headless/src/app/search-engine/search-engine.ts
```diff
@@ -111,20 +112,26 @@ export interface SearchEngineOptions
...
   };
 
   const augmentedOptions: EngineOptions<SearchEngineReducers> = {
     ...options,
     reducers: searchEngineReducers,
     crossReducer: jwtReducer(logger),
+    // The middleware for the answer API is only added when the knowledgeConfigurationId is provided
+    ...(options.configuration.answerConfigurationId
+      ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        {middlewares: [answerApi.middleware as any]}
```
dmgauthier, line 133: I Need help here. The `middlewares` type is inferred from `SearchEngineReducers` and obviously I cannot modify the latter.

packages/headless/src/app/engine-configuration.ts
```diff
@@ -50,6 +50,11 @@ export interface EngineConfiguration<
    * The access token to use to authenticate requests against the Coveo Cloud endpoints. Typically, this will be an API key or search token that grants the privileges to execute queries and push usage analytics data in the target Coveo Cloud organization.
    */
   accessToken: string;
+  /**
```
olamothe, line 53: This cannot be added to the "global engine config". 

This is the engine config that every Coveo use case supported by Headless will inherit from. 

This config needs to be colocated to where to only where this is actually needed. 

In other words, this PR should not be touching any files related to `engine-configuration`, `engine.ts`, and possibly `configuration` (TBD...). 

This could possibly be solely located in the existing question answering slice ?

olamothe (commented):
Apart from where the config should be located, im skeptical about the purpose of this flag.

It is used both the control "which endpoints will be used" and possibly also will have other end user facing impact ? 

So its like a flag that does many different things ? 
Will we want everyone to eventually use the new endpoint ? 
If so, how will we make it happen with an optional flag ?
Should we instead have a fully dedicated "genqa" configuration section on the specific engine that requires it ? (insight, search)

I'd like to know more before approving these changes. It's not necessarily a bad idea, but I need more context.

dmgauthier:
> Apart from where the config should be located, im skeptical about the purpose of this flag.
> 
> It is used both the control "which endpoints will be used" and possibly also will have other end user facing impact ?
> 
> So its like a flag that does many different things ? Will we want everyone to eventually use the new endpoint ? If so, how will we make it happen with an optional flag ? Should we instead have a fully dedicated "genqa" configuration section on the specific engine that requires it ? (insight, search)
> 
> I'd like to know more before approving these changes. It's not necessarily a bad idea, but I need more context.

Thanks for your concerns :raised_hands: 
Roughly said, when a interface associated with an answer configuration, we want that interface to use the answer API instead of the search UI stream ID pattern. 
This is the way I see how this can be done properly: By adding the possibility to initiate an interface with an Answer Config Id, we can accomplish a couple of things. First of, It can easily be "templated" from the search-interface-service. From there, the configurationId can easily be dispatched to the state, and then used by the endpoint to call the proper config. Also, it can be leveraged by the end user components to use the proper behavior, because this change of pattern will differ from letting the search response return a stream ID; some components, like the search box, will have to request a stream directly, which wasn't the case.

In the end, we might want to deprecate the current searchApi streamID pattern completely, but it hasn't been decided yet of course, and in the meanwhile we want all the customers without an AnswerManager configuration to keep the exact same behavior as before. Hence using the configId a bit like a flag.

Should we instead have a fully dedicated "genqa" configuration section on the specific engine that requires it ? Well maybe, but I am afraid that this might be overkill and only serve to host an "answerConfigId" in the end. I do see though that if we do not create one now, but realize in the future we'd better create one, it will introduce a breaking change.

packages/headless/src/app/engine-configuration.ts
dmgauthier: I get your very relevant concern.
Down the road I think that all the engines that use RGA will leverage this Answer configuration ID, but this doesn't means we absolutely need to modify the core interfaces. Or the core engine. I'll see what I can do.

dmgauthier: I removed the change I made to `engine-configuration` and and `engine`. But kept the ones in configuration-slice and configuration-actions

packages/headless/src/app/search-engine/search-engine.ts
```diff
@@ -111,20 +113,24 @@ export interface SearchEngineOptions
...
+      : {streamingClient: createLegacyGeneratedAnswerAPIClient(logger)}),
   };
 
   const augmentedOptions: EngineOptions<SearchEngineReducers> = {
     ...options,
     reducers: searchEngineReducers,
     crossReducer: jwtReducer(logger),
+    ...(options.configuration.answerConfigurationId
+      ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        {middlewares: [answerApi.middleware as any]} // TODO: types
```
dmgauthier, line 132: This needs to be fixed before mergine

packages/headless/src/features/configuration/configuration-actions.ts
```diff
@@ -37,6 +37,13 @@ export interface UpdateBasicConfigurationActionCreatorPayload {
   platformUrl?: string;
 }
 
+interface UpdateAnswerConfigurationActionCreatorPayload {
+  /**
+   * The Answer Configuration ID to use.
```
dmgauthier, line 42: This description is pretty poor

packages/headless/src/app/search-engine/search-engine.ts
```diff
@@ -111,20 +113,29 @@ export interface SearchEngineOptions
...
+      ? {}
+      : {streamingClient: createLegacyGeneratedAnswerAPIClient(logger)}),
   };
 
-  const augmentedOptions: EngineOptions<SearchEngineReducers> = {
+  type EngineOptionsWithAnswerMiddlewares = Omit<
+    EngineOptions<SearchEngineReducers>,
+    'middlewares'
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  > & {middlewares?: any};
```
dmgauthier, line 130: This is not a lot better than casting the middlewares type at line 137. But I did not find any better way to pass the answerApi.middleware in the options.

packages/headless/src/app/engine-configuration.ts
dmgauthier: > This could possibly be solely located in the existing question answering slice ?

Since the usage of the `answerConfiguration` will influence components outside the question answering, like the searchBox, and to ease to configuration of the interfaces using the answer manager. I think we have more advantage to set its entry point at the engine initialization level.

olamothe:
So, after re-reviewing and discussing with other in the team, really think we should not be adding this ID as anything global.

Certainly not for search engine, and not necessary for insight engine either (although not as clear cut as search engine).

If a config is to be global, then it must be truly because every single element you can think of in the interface will need it. For example `organizationId`, `how to reach the platform (url, org endpoints)`, `analytics`.

In the case of commerce, we add stuff like `the context of the user` because this is ubiquitous in the context of commerce (your storefront powered by Coveo literally cannot function without that information)

For Search engine, most clients don't use RGA. For insight engine (insight panel) it's the same thing. For a search page, facets, pagination, result components etc. won't need to ever know about the answerConfigurationId.


I think this should belong entirely in the `generated-answer-slice`, and that implementer should configure this through the `headless-generated-answer` and `headless-insight-generated-answer` controller (an option when creating the controller).

> First of, It can easily be "templated" from the search-interface-service. 

The accompanying atomic and quantic component will need to be updated to have a `answerConfigurationId` property. And they will be in charge of passing that down to the controller. So the templating will happen at the component level in charge of displaying the generated answer.

> Since the usage of the answerConfiguration will influence components outside the question answering, like the searchBox

If the searchbox requires that slice to perform it's job, then we simply add it to the loaded reducer in `headless-search-box`, `headless-insight-search-box`. As opposed to making it global everywhere.

And the reason I'm a stickler for this kind of thing, is that those kind of issue about "where should an implementer configure this particular thing" have long lasting effect (in terms on years), and are hard to change because of breaking changes management. 

Typically better to have stuff as local as possible otherwise we quickly end up with a "godlike" config object.

dmgauthier:
> And the reason I'm a stickler for this kind of thing, is that those kind of issue about "where should an implementer configure this particular thing" have long lasting effect (in terms on years), and are hard to change because of breaking changes management.
> 
> Typically better to have stuff as local as possible otherwise we quickly end up with a "godlike" config object.

That sells it.  I will close the current PR and use the "more local" route.

PR #4132, opened 2024-06-28
fix(commerce): only expose breadcrumbs with values

Spuffynism:
By only exposing breadcrumbs with values, breadcrumbs without facet values will not be included. Since we also filter out breadcrumbs without values, the `hasBreadcrumbs` check will be `false` instead of `true` when there are facets with no values, which is the desired state.

[CAPI-1089]

### DIFF
--- a/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
+++ b/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
@@ -236,7 +236,10 @@ export function buildCoreBreadcrumbManager(
           .map((facetId) =>
             options.facetResponseSelector(engine[stateKey], facetId)
           )
-          .filter((facet): facet is AnyFacetResponse => facet !== undefined)
+          .filter(
+            (facet): facet is AnyFacetResponse =>
+              facet !== undefined && facet.values.length > 0
+          )
           .map(createBreadcrumb) ?? [];
       return {
         facetBreadcrumbs: breadcrumbs,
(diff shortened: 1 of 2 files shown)


packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
```diff
@@ -236,7 +236,10 @@ export function buildCoreBreadcrumbManager(
           .map((facetId) =>
             options.facetResponseSelector(engine[stateKey], facetId)
           )
-          .filter((facet): facet is AnyFacetResponse => facet !== undefined)
+          .filter(
+            (facet): facet is AnyFacetResponse =>
+              facet !== undefined && facet.values.length > 0
```
samisayegh, line 241: Looks good, only question I'd have is does it mirror the behavior of non-commerce breadcrumb manager controllers?

samisayegh (approved):
Nice ✅

packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
Spuffynism: Yup, the non-commerce one excludes breadcrumbs with no values: https://github.com/coveo/ui-kit/blob/3330e49f180b7b997861446cf23bad2ea28cab51/packages/headless/src/controllers/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts#L249

alexprudhomme (approved):

fbeaudoincoveo (approved):

PR #4134, opened 2024-06-28
fix(SFINT-5598): Always send contentFormat even with rephrase RGA

erocheleau:
Fix a bug where after using the rephrase buttons, the contentFormat of the requested RGA response was not persisting to sending `text/markdown`.

Now on every search, we do send `['text/markdown', 'text/plain']` as the accepted format alongside the `answerStyle` parameter.


mmitiche (commented):
the fix looks good 👍
I just think that a small e2e test could be added to avoid fixing this more than once.

SimonMilord (approved):
Looks good!

PR #4135, opened 2024-07-01
feat(answer): knowledge answer controller

dmgauthier:
## Summary

This increment adds the Controller that will map the state of the answerApi Slice to the Generated Answer State used by the front end components. 

## Why this is needed

Since we must for a while support the current working flow with the API stream, while adding the new flow from the AnswerAPI, it was way simpler to make sure both implementation rely on the same state. Cuz having to rework all the components to align them with the AnswerApi slice directly would be breaking and would not cover all the actual functionalities, like rephrasing.

## How does it work

The front end components will use the same controller in both the flows, the only thing changing, at a DX level, is to pass the configuration Id in the props of the controller. 
By doing so, the first level controller will then build and extend the right sub controller to handle the proper GenAI flow.
Both the legacy controller working with the searchUi and the new controller working with the Answer API share the same state.
The configuration Id will be passed to the client seamlessly. 
In most cases, the answer generation will be identical in both the flows. The main difference is that the flow using the answerConfigurationId unlocks the answer management capabilities.

### DIFF
--- a/packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
+++ b/packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
@@ -9,6 +9,7 @@ import {
   GeneratedAnswerPropsInitialState,
 } from '../core/generated-answer/headless-core-generated-answer';
 import {buildSearchAPIGeneratedAnswer} from '../core/generated-answer/headless-searchapi-generated-answer';
+import {buildKnowledgeGeneratedAnswer} from '../knowledge/generatedAnswer/headless-knowledge-generated-answer';
 
 export type {
   GeneratedAnswerCitation,
@@ -21,7 +22,7 @@ export type {
 
 /**
  * Creates a `GeneratedAnswer` controller instance.
- *
+
  * @param engine - The headless engine.
  * @param props - The configurable `GeneratedAnswer` properties.
  * @returns A `GeneratedAnswer` controller instance.
@@ -30,11 +31,17 @@ export function buildGeneratedAnswer(
   engine: SearchEngine,
   props: GeneratedAnswerProps = {}
 ): GeneratedAnswer {
-  const controller = buildSearchAPIGeneratedAnswer(
-    engine,
-    generatedAnswerAnalyticsClient,
-    props
-  );
+  const controller = props.answerConfigurationId
+    ? buildKnowledgeGeneratedAnswer(
+        engine,
+        generatedAnswerAnalyticsClient,
+        props
+      )
+    : buildSearchAPIGeneratedAnswer(
+        engine,
+        generatedAnswerAnalyticsClient,
+        props
+      );
 
   return {
     ...controller,
--- a/packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.test.ts
+++ b/packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.test.ts
@@ -1,0 +1,129 @@
+import {answerApi, fetchAnswer} from '../../../api/knowledge/stream-answer-api';
+import {
+  resetAnswer,
+  updateAnswerConfigurationId,
+  updateResponseFormat,
+} from '../../../features/generated-answer/generated-answer-actions';
+import {generatedAnswerAnalyticsClient} from '../../../features/generated-answer/generated-answer-analytics-actions';
+import {
+  GeneratedAnswerState,
+  getGeneratedAnswerInitialState,
+} from '../../../features/generated-answer/generated-answer-state';
+import {queryReducer} from '../../../features/query/query-slice';
+import {
+  buildMockSearchEngine,
+  MockedSearchEngine,
+} from '../../../test/mock-engine-v2';
+import {createMockState} from '../../../test/mock-state';
+import {
+  GeneratedAnswerProps,
+  GeneratedResponseFormat,
+} from '../../generated-answer/headless-generated-answer';
+import {buildKnowledgeGeneratedAnswer} from './headless-knowledge-generated-answer';
+
+jest.mock('../../../features/generated-answer/generated-answer-actions');
+jest.mock(
+  '../../../features/generated-answer/generated-answer-analytics-actions'
+);
+jest.mock('../../../features/search/search-actions');
+jest.mock('../../../api/knowledge/stream-answer-api', () => {
+  const originalStreamAnswerApi = jest.requireActual(
+    '../../../api/knowledge/stream-answer-api'
+  );
+  return {
+    ...originalStreamAnswerApi,
+    fetchAnswer: jest.fn(),
+  };
+});
+
+describe('knowledge-generated-answer', () => {
+  it('should be tested', () => {
+    expect(true).toBe(true);
+  });
+
+  let engine: MockedSearchEngine;
+
+  const createGeneratedAnswer = (props: GeneratedAnswerProps = {}) =>
+    buildKnowledgeGeneratedAnswer(
+      engine,
+      generatedAnswerAnalyticsClient,
+      props
+    );
+
+  const buildEngineWithGeneratedAnswer = (
+    initialState: Partial<GeneratedAnswerState> = {}
+  ) => {
+    const state = createMockState({
+      generatedAnswer: {
+        ...getGeneratedAnswerInitialState(),
+        ...initialState,
+      },
+    });
+    return buildMockSearchEngine(state);
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    engine = buildEngineWithGeneratedAnswer();
+  });
+
+  it('initializes', () => {
+    const generatedAnswer = createGeneratedAnswer();
+    expect(generatedAnswer).toBeTruthy();
+  });
+
+  it('it adds the correct reducers to engine', () => {
+    createGeneratedAnswer();
+    expect(engine.addReducers).toHaveBeenCalledWith({
+      [answerApi.reducerPath]: answerApi.reducer,
+      query: queryReducer,
+    });
+  });
+
+  it('dispatches the configuration id upon initialization', () => {
+    const answerConfigurationId = 'answerConfigurationId';
+    createGeneratedAnswer({answerConfigurationId});
+    expect(updateAnswerConfigurationId).toHaveBeenCalledWith(
+      answerConfigurationId
+    );
+  });
+
+  it('exposes its state', () => {
+    const generatedAnswer = createGeneratedAnswer();
+
+    expect(generatedAnswer.state).toEqual({
+      ...engine.state.generatedAnswer,
+      answer: undefined,
+      answerContentFormat: 'text/plain',
+      error: {message: undefined},
+    });
+  });
+
+  it('dispatches a rephrase action', () => {
+    const generatedAnswer = createGeneratedAnswer();
+    const responseFormat: GeneratedResponseFormat = {answerStyle: 'step'};
+    generatedAnswer.rephrase(responseFormat);
+    expect(updateResponseFormat).toHaveBeenCalledWith(responseFormat);
+  });
+
+  it('initialize the format', () => {
+    const responseFormat: GeneratedResponseFormat = {answerStyle: 'concise'};
+    createGeneratedAnswer({
+      initialState: {responseFormat},
+    });
+
+    expect(updateResponseFormat).toHaveBeenCalledWith(responseFormat);
+  });
+
+  it('dispatches a retry action', () => {
+    const generatedAnswer = createGeneratedAnswer();
+    generatedAnswer.retry();
+    expect(fetchAnswer).toHaveBeenCalledTimes(1);
+  });
+
+  it('dispatches a reset action', () => {
+    const generatedAnswer = createGeneratedAnswer();
+    generatedAnswer.reset();
+    expect(resetAnswer).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
+++ b/packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
@@ -1,0 +1,101 @@
+import {
+  answerApi,
+  fetchAnswer,
+  selectAnswer,
+} from '../../../api/knowledge/stream-answer-api';
+import {SearchEngine} from '../../../app/search-engine/search-engine';
+import {
+  resetAnswer,
+  updateAnswerConfigurationId,
+} from '../../../features/generated-answer/generated-answer-actions';
+import {queryReducer as query} from '../../../features/query/query-slice';
+import {
+  GeneratedAnswerSection,
+  QuerySection,
+} from '../../../state/state-sections';
+import {loadReducerError} from '../../../utils/errors';
+import {
+  buildCoreGeneratedAnswer,
+  GeneratedAnswer,
+  GeneratedAnswerAnalyticsClient,
+  GeneratedAnswerProps,
+  GeneratedResponseFormat,
+} from '../../core/generated-answer/headless-core-generated-answer';
+
+export interface KnowledgeGeneratedAnswer extends GeneratedAnswer {
+  /**
+   * Resets the last answer.
+   */
+  reset(): void;
+}
+
+interface KnowledgeGeneratedAnswerProps extends GeneratedAnswerProps {}
+
+export interface SearchAPIGeneratedAnswerAnalyticsClient
+  extends GeneratedAnswerAnalyticsClient {}
+
+/**
+ * Creates a `GeneratedAnswer` controller instance using the Answer API stream pattern.
+ *
+ * @param engine - The headless engine.
+ * @param props - The configurable `GeneratedAnswer` properties.
+ * @returns A `GeneratedAnswer` controller instance.
+ */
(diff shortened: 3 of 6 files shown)


pathiery (approved):

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -25,14 +29,20 @@ export type {
  * @returns A `GeneratedAnswer` controller instance.
  */
 export function buildGeneratedAnswer(
-  engine: SearchEngine,
+  engine: SearchEngine | EngineNeededForKnowledgeGeneratedAnswer,
```
fbeaudoincoveo, line 32: If you apply the comments I left here https://github.com/coveo/ui-kit/pull/4129/files#diff-644461440e8a66297a77b95e54021c1e4840bad4c475302dea37bfc17945091cR117, then this would be

```suggestion
  engine: SearchEngine | InsightEngine | EngineNeededForKnowledgeGeneratedAnswer,
```

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,106 @@
...
+  buildCoreGeneratedAnswer,
+  GeneratedAnswer,
+  GeneratedAnswerAnalyticsClient,
+  GeneratedAnswerProps,
+  GeneratedResponseFormat,
+} from '../../core/generated-answer/headless-core-generated-answer';
+
+export interface KnowledgeGeneratedAnswer extends GeneratedAnswer {
+  /**
+   * resets the last answer.
```
fbeaudoincoveo, line 28:
```suggestion
   * Resets the last answer.
```

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export type EngineNeededForKnowledgeGeneratedAnswer = CoreEngine<
+  GeneratedAnswerSection &
+    ConfigurationSection &
+    SearchSection &
+    DebugSection & {answer: ReturnType<typeof answerApi.reducer>}
+>;
+
+/**
+ * Creates a `GeneratedAnswer` controller instance using the search api stream pattern.
```
fbeaudoincoveo, line 46:
```suggestion
 * Creates a `GeneratedAnswer` controller instance using the search API stream pattern.
```

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,106 @@
...
+        error: {
+          message: answerApiState?.error?.message,
+          statusCode: answerApiState?.error?.code,
+        },
+        isLoading: answerApiState?.isLoading ?? false,
+        isStreaming: answerApiState?.isStreaming ?? false,
+        answerContentFormat: answerApiState?.contentFormat ?? 'text/plain',
+        isAnswerGenerated: answerApiState?.generated ?? false,
+      };
+    },
```
dmgauthier, line 87: I know that we intend to move from exposing the state to expose selectors. Should we do something different here ?

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -27,14 +32,23 @@ export type {
  * @returns A `GeneratedAnswer` controller instance.
  */
 export function buildGeneratedAnswer(
-  engine: SearchEngine,
+  engine:
```
olamothe, line 35: General questions about this change here: 

What do we expect implementers to actually pass as an argument here ? Really any of those 3 different types ?

How does that jive with `headless-insight-generated-answer` ? 

It seems we should instead doing these changes in `headless-core-generated-answer` instead 🤔 

Because that one is not "user facing", it's internal. And so the internal function can be polymorphic/generic as you've done here, but the top level facing one stays exactly the same. Otherwise it's confusing to manage for implementers.

If not, and you prefer to keep that approach, then we must not expand the type here; Instead `SearchEngine` type should be modified in a non breaking manner so that it fits your needs downstream. 

But I think `SearchEngine` already specifies that it optionally contains a `GeneratedAnswerSection` as part of it's state, so there should not be much (if anything?) to modify on the interface.

dmgauthier: I reverted the change, but I had to cast downstream.

Opening the subject about the types in Headless, I'm not sure if I'm missing something, but I find them unnecessarily difficult to work with for no apparent reason.

For example, the Redux state appears to be the same across all instances, but I often find myself resorting to hacks like StateFromReducersMapObject<Reducers> or manually creating StateRequiredForThisAndThat. This feels unnecessarily cumbersome and frustrating without providing significant additional confidence.

I believe having one big state type that we can use everywhere would be more... honest ? While offering the intelliSense support, and prevent constant issues with TypeScript errors.

Overall, I would be interested to know the motivation behind this type of architecture.

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export interface SearchAPIGeneratedAnswerAnalyticsClient
+  extends GeneratedAnswerAnalyticsClient {}
+
+export type EngineNeededForKnowledgeGeneratedAnswer = CoreEngine<
+  GeneratedAnswerSection &
+    ConfigurationSection &
+    SearchSection &
+    DebugSection & {answer: ReturnType<typeof answerApi.reducer>}
+>;
```
alexprudhomme, line 45: I don't think this type is necessary.  Its only used as a cast in the test file

dmgauthier: quite right, thanks !

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -30,11 +31,17 @@ export function buildGeneratedAnswer(
...
+    ? buildKnowledgeGeneratedAnswer(
+        engine,
+        generatedAnswerAnalyticsClient,
+        props
+      )
+    : buildSearchAPIGeneratedAnswer(
+        engine as SearchEngine,
+        generatedAnswerAnalyticsClient,
+        props
+      );
```
mmitiche, line 44: Why are we not having such a logic for the insight generated answer controller?
is the feature of the new `buildKnowledgeGeneratedAnswer` controller not yet supported for insight?

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
```diff
@@ -30,11 +31,17 @@ export function buildGeneratedAnswer(
...
-    props
-  );
+  const controller = props.answerConfigurationId
+    ? buildKnowledgeGeneratedAnswer(
+        engine,
+        generatedAnswerAnalyticsClient,
+        props
+      )
+    : buildSearchAPIGeneratedAnswer(
+        engine as SearchEngine,
```
mmitiche, line 41: why the `as SearchEngine` is needed here? the engine is already of `SearchEngine` in this controller,

mmitiche (approved):
The changes look good 👍

I have just a question about why the new way of streaming answers using the AnswerAPI is not implemented as well in the Insight Generated Answer controller?

Also, after this PR,  the next one to review is the following: https://github.com/coveo/ui-kit/pull/4148/files#diff-40f9b6094183b67c164540da56e5747ca8dec42711e580891dd77cd7a682a008 right?

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
fbeaudoincoveo: Exposing state of a controller is still perfectly fine; it's the engine state what will eventually become private (it already is in the commerce engine).

Eventually we'll do some refactoring, but for now this looks good to me!

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (!loadAnswerApiReducer(engine)) {
+    throw loadReducerError;
+  }
+
+  const {rephrase: coreRephrase, ...controller} = buildCoreGeneratedAnswer(
+    engine,
+    analyticsClient,
+    props
+  );
+  const getState = () => engine.state as unknown as StateNeededByAnswerAPI;
```
fbeaudoincoveo, line 60:
```suggestion
  const getState = () => engine.state;
```

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.test.ts
```diff
@@ -0,0 +1,285 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+    engine = buildEngineWithGeneratedAnswer();
+  });
+
+  it('initializes', () => {
+    const generatedAnswer = createGeneratedAnswer();
+    expect(generatedAnswer).toBeTruthy();
+  });
+
```
fbeaudoincoveo, line 71: I would add a test to ensure it loads the answerApi reducer, and another test to assert that it dispatches the `updateConfigurationId` action with the correct answerConfigurationId as a payload upon initialization.

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.test.ts
```diff
@@ -0,0 +1,285 @@
...
+  it('initialize the format', () => {
+    const responseFormat: GeneratedResponseFormat = {answerStyle: 'concise'};
+    createGeneratedAnswer({
+      initialState: {responseFormat},
+    });
+
+    expect(updateResponseFormat).toHaveBeenCalledWith(responseFormat);
+  });
+
+  it('dispatches a like action', () => {
```
fbeaudoincoveo, line 99: I don't think we necessarily need to retest everything that the core controller does here, since it's already tested in headless-core-generated-answer.test.ts!

you could simply test, the `state` getter, and the `rephrase`, `retry`, and `reset` methods.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
fbeaudoincoveo: > For example, the Redux state appears to be the same across all instances

This is not quite accurate: the shape of the state ultimately depends on which reducers are added to the engine.

For instance, in an atomic search page, if you have the following markup:

```
 <atomic-search-interface>
   <atomic-result-list></atomic-result-list>
 </atomic-search-interface>
```

which is pretty much as basic as it gets, you'll have a state that looks like this:

Screenshot 2024-07-11 at 3 35 13 PM

and then as you add components, which rely on controllers that load additional reducers, the state will change accordingly, e.g.,:

```
<atomic-search-interface>
  <atomic-facet field="author"></atomic-facet>
  <atomic-result-list></atomic-result-list>
</atomic-search-interface>
```

would yield a state like:

Screenshot 2024-07-11 at 3 36 57 PM

In short, you're not guaranteed that the state will contain the properties you need unless you load the corresponding reducers in your controllers.

dmgauthier:
> The changes look good 👍

Thanks !

> I have just a question about why the new way of streaming answers using the AnswerAPI is not implemented as well in the Insight Generated Answer controller?

Not yet !


> Also, after this PR, the next one to review is the following: https://github.com/coveo/ui-kit/pull/4148/files#diff-40f9b6094183b67c164540da56e5747ca8dec42711e580891dd77cd7a682a008 right?

Yes !

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
dmgauthier: This was not in the radar for the MVP and for the sake of being able to give the feature to some MVP customers, we only support the basic interface usecase for now.  No doubt we will have to support insight at some point though.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
dmgauthier: I see.

And I get that for the sake of optimization, it does makes a lot of sense to not load everything each time. 

But unless there is some core element I did not understand in the Typescript patterns that are used, its quite hard to modify or anything without having to fight with typing issues.

A PR review might not be the right place to talk about this though.  Thanks for the detailed answer.

packages/headless/src/controllers/generated-answer/headless-generated-answer.ts
dmgauthier: actually we need it because the `SearchAPIGeneratedAnswer` controller will be used by the insight engine. We pass the insight engine to the `buildGeneratedAnswer` but I had to cast the `searchEngine` when passing it to the `buildSearchAPIGeneratedAnswer`

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.test.ts
```diff
@@ -0,0 +1,131 @@
+import {answerApi, fetchAnswer} from '../../../api/knowledge/stream-answer-api';
+import {
+  resetAnswer,
+  updateAnswerConfigurationId,
+  updateResponseFormat,
+} from '../../../features/generated-answer/generated-answer-actions';
+import {generatedAnswerAnalyticsClient} from '../../../features/generated-answer/generated-answer-analytics-actions';
+// import {generatedAnswerReducer} from '../../../features/generated-answer/generated-answer-slice';
```
dmgauthier, line 8: :weary:

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,110 @@
...
+   */
+  reset(): void;
+}
+
+interface KnowledgeGeneratedAnswerProps extends GeneratedAnswerProps {}
+
+export interface SearchAPIGeneratedAnswerAnalyticsClient
+  extends GeneratedAnswerAnalyticsClient {}
+
+export type EngineNeededForKnowledgeGeneratedAnswer = SearchEngine<
```
fbeaudoincoveo, line 40: This type represents the shape that the engine will have _after_ the controller is built, not the shape that the engine must have when the `buildKnowledgeGeneratedAnswer` function is called.

The controller should simply accept a `SearchEngine` object for the `engine` parameter. Then, you should use the `loadApiAnswerReducer` function to ensure that the required reducers / state sections are added to the engine. This is why the `loadApiAnswerReducer` function is implemented as a type guard: once you've added the desired reducers, you know that the engine's state will match the desired shape. This implies that the `loadApiAnswerReducer` implementation is very important.

You know that any `SearchEngine` object will already have the following reducers / state sections:
- `ConfigurationSection` (because `configuration` is part of the core reducers loaded for every engine)
- `SearchSection` (because `search` is part of the core search reducers)
- `DebugSection` (for the same reason)
- `GeneratedAnswerSection` (because the `generatedAnswer` reducers is added in `buildCoreGeneratedAnswer`)

So you don't need to add those again in the `loadAnswerApiReducer`  function (although doing so would have no significant adverse effect). The reducers / state sections you're actually missing are `{answer: ReturnType<typeof answerApi.reducer>}` and `QuerySection`. Right now, you're only adding `{answer: ReturnType<typeof answerApi.reducer>}` in the `loadAnswerApiReducer` function, so you also need to load the query reducer (I'll add a comment).

In short: you don't need this type!

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,110 @@
...
+  };
+}
+
+function loadAnswerApiReducer(
+  engine: CoreEngine
+): engine is SearchEngine<
+  GeneratedAnswerSection &
+    QuerySection & {answer: ReturnType<typeof answerApi.reducer>}
+> {
+  engine.addReducers({[answerApi.reducerPath]: answerApi.reducer});
```
fbeaudoincoveo, line 108:
```suggestion
  engine.addReducers({[answerApi.reducerPath]: answerApi.reducer, query});
```

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -0,0 +1,110 @@
...
+} from '../../../state/state-sections';
+import {loadReducerError} from '../../../utils/errors';
+import {
+  buildCoreGeneratedAnswer,
+  GeneratedAnswer,
+  GeneratedAnswerAnalyticsClient,
+  GeneratedAnswerProps,
+  GeneratedResponseFormat,
+} from '../../core/generated-answer/headless-core-generated-answer';
+
```
fbeaudoincoveo, line 27:
```suggestion
import {queryReducer as query} from '../../../features/query/query-slice';
```

fbeaudoincoveo (changes requested):

fbeaudoincoveo (approved):

PR #4136, opened 2024-07-02
feat(headless): quick view event for insight use case made an InsightPanel.ItemAction

mmitiche:
## [SFINT-5581](https://coveord.atlassian.net/browse/SFINT-5581)

- Quick view event sent in the insight use case updated to be an `InsightPanel.ItemClick` event according to the latest schemas of the `@coveo/relay-event-types` library.
- Unit tests added.

[SFINT-5581]: https://coveord.atlassian.net/browse/SFINT-5581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olamothe (approved):

packages/headless/src/features/result-preview/result-preview-insight-analytics-actions.ts
```diff
@@ -0,0 +1,34 @@
...
+  ClickAction,
+  analyticsEventItemMetadata,
+  documentIdentifier,
+  makeInsightAnalyticsActionFactory,
+  partialDocumentInformation,
+  validateResultPayload,
+} from '../analytics/analytics-utils';
+import {analyticsEventCaseContext} from '../analytics/insight-analytics-utils';
+
+export const logDocumentQuickview = (result: Result): ClickAction => {
```
erocheleau, line 13: I don't understand?
What was it before this PR? Just a regular ItemClick with Quickview as the "cause" ?

erocheleau (approved):
Just that question but otherwise everything looks perfect

packages/headless/src/features/result-preview/__snapshots__/result-preview-insight-analytics-actions.test.ts.snap
```diff
@@ -0,0 +1,24 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`#logDocumentQuickview when analyticsMode is \`next\` should call relay.emit properly 1`] = `
+[
+  "InsightPanel.ItemAction",
+  {
+    "action": "preview",
```
fbeaudoincoveo, line 7: Should we also send the `sourceEntityId`?

https://github.com/coveo-platform/analytics_schema/blob/master/schemas/client/api.event.insightPanel.itemAction/1/api.event.insightPanel.itemAction.1.request.json#L41

fbeaudoincoveo (approved):

packages/headless/src/features/result-preview/result-preview-insight-analytics-actions.ts
mmitiche: Indeed, before this PR it was a regular itemClick exactly like we send for the search use case.

packages/headless/src/features/result-preview/__snapshots__/result-preview-insight-analytics-actions.test.ts.snap
mmitiche: good question, since it exists in the schema I think we should yeah, however I see that we are not sending in all the insight panel events, so I will create a jira to investigate and updated this, we basically need to figure out what to send as the value of this key and add it to all the events where it should be sent.

packages/headless/src/features/result-preview/__snapshots__/result-preview-insight-analytics-actions.test.ts.snap
```diff
@@ -0,0 +1,24 @@
...
+    "context": {
+      "caseNumber": "5678",
```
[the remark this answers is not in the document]
mmitiche: yes indeed, however we need to complete the following jira before being able to add it properly: https://coveord.atlassian.net/browse/SFINT-5420.

This ticket will allow us to save the caseSubject properly in the Headless state and include it in such events. I'm linking this comment to the mentioned Jira so we tackle this when we address the ticket.

PR #4140, opened 2024-07-02
feat(headless commerce): expose additional actions

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3350


packages/headless/src/commerce.index.ts
```diff
@@ -52,11 +52,14 @@ export * from './features/commerce/facets/date-facet/date-facet-actions-loader';
 export * from './features/commerce/facets/numeric-facet/numeric-facet-actions-loader';
 export * from './features/commerce/query-set/query-set-actions-loader';
 export * from './features/commerce/query-suggest/query-suggest-actions-loader';
-export * from './features/configuration/configuration-actions-loader';
+export * from './features/commerce/configuration/configuration-actions-loader';
```
fbeaudoincoveo, line 55: This is breaking, please take time to review this file (and configuration-actions.ts) carefully!

packages/headless/src/features/commerce/configuration/configuration-actions.ts
```diff
@@ -0,0 +1,44 @@
...
+  'commerce/configuration/updateBasicConfiguration',
+  (payload: UpdateBasicConfigurationPayload) =>
+    validatePayload(payload, {
+      accessToken: nonEmptyString,
+      organizationId: nonEmptyString,
+      platformUrl: nonEmptyString,
+    })
+);
+
+export type UpdateAnalyticsConfigurationPayload = Pick<
```
fbeaudoincoveo, line 22: This is a significant change, here's my reasoning:

- `originContext`, `originLevel2`, and `originLevel3` are not needed beacause they are not leveraged by the commerce API (i.e., server-side events) or by Relay (i.e., client-side events), so they don't seem relevant to me. Arguably, `originLevel3` would normally be the referrer, but that's retrieved through the navigator context provider.
- `apiBaseUrl` / `nextApiBaseUrl` is not needed because it would be the same as for search requests (for server-side events), or it would be managed by Relay (for client-side events). I might be wrong about that though.
- I suppose `runtimeEnvironment` should be inferred (or explicitely specified) in the navigator context provider.
- `anonymous` not supported by the commerce API or Relay AFAIK
- `deviceId` I believe this is not useful because the clientId is sufficient
- `userDisplayName` not supported by commerce API or Relay AFAIK.
- `documentLocation` this would be the view.url in the context of commerce.
- `analyticsMode` only next is supported in commerce

Bear in mind I may very well be wrong about any of these, so please challenge this ;)

packages/headless/src/features/commerce/recent-queries/recent-queries-analytics-actions.ts
```diff
@@ -1,6 +0,0 @@
-import {SearchPageEvents} from '../../analytics/search-action-cause';
```
fbeaudoincoveo, line 1: Renamed

packages/headless/src/features/commerce/search/search-actions-thunk-processor.ts
```diff
@@ -17,7 +17,6 @@ import {
   CommerceDidYouMeanSection,
   TriggerSection,
 } from '../../../state/state-sections';
-import {logQueryError} from '../../search/search-analytics-actions';
```
fbeaudoincoveo, line 20: Irrelevant in commerce

packages/headless/src/features/configuration/configuration-slice.ts
```diff
@@ -156,3 +171,23 @@ export const configurationReducer = createReducer(
...
+  if (payload.organizationId) {
+    state.organizationId = payload.organizationId;
+  }
+  if (payload.platformUrl) {
+    state.platformUrl = payload.platformUrl;
+    state.search.apiBaseUrl = `${payload.platformUrl}${searchAPIEndpoint}`;
+    state.analytics.apiBaseUrl = legacyAnalyticsUrlFromPlatformUrl(
+      payload.platformUrl,
+      state.organizationId
+    );
```
Spuffynism, line 191: Do we actually need to set these urls when handling `updateCommerceBasicConfiguration`? I don't think they're used for commerce

Spuffynism (approved):

packages/headless/src/features/commerce/configuration/configuration-actions.ts
fbeaudoincoveo: I'm hesitant mostly on the following properties:

- `nextApiBaseUrl`
- `runtimeEnvironment`

@louis-bompart, a penny for your thoughts 😄

packages/headless/src/features/configuration/configuration-slice.ts
fbeaudoincoveo: On the other hand, there's no harm in setting those (it's not costly by any means in terms of performance).

Spuffynism: Yeah, I don't mind leaving this in

packages/headless/src/features/commerce/configuration/configuration-actions.ts
fbeaudoincoveo: As discussed in, worst case scenario we can always add some properties back.

louis-bompart (approved):

PR #4146, opened 2024-07-04
feat(answer): error handling

dmgauthier:
## Summary

Updates to the Answer Api client to make sure the error handling is proper. Some little extras needed for it to work proper that will make the subsequent review a bit easier.

## Why

The Answer Api is almost just a proxy for the stream coming from reveal. And for some technical reasons, the call to the answer api will return an error message but will not return downright error status codes like 400s when the stream fails.
That said, we can rely on the stream end message in case the stream is not working properly, and since its better to give the user or the devs all the information we can, this increment makes sure that the error will appear in the logs and in the state in case we want to leverage the error message in the front end component.

### DIFF
--- a/packages/headless/src/api/knowledge/stream-answer-api.ts
+++ b/packages/headless/src/api/knowledge/stream-answer-api.ts
@@ -1,14 +1,8 @@
 import {
-  ArrayValue,
-  BooleanValue,
-  RecordValue,
-  Schema,
-  StringValue,
-} from '@coveo/bueno';
-import {
   EventSourceMessage,
   fetchEventSource,
 } from '@microsoft/fetch-event-source';
+import {selectFieldsToIncludeInCitation} from '../../features/generated-answer/generated-answer-selectors';
 import {
   GeneratedAnswerStyle,
   GeneratedContentFormat,
@@ -30,31 +24,32 @@ import {answerSlice} from './answer-slice';
 type StateNeededByAnswerAPI = ConfigurationSection &
   GeneratedAnswerSection &
   SearchSection &
-  DebugSection & {answer: ReturnType<typeof answerApi.reducer>};
-
-interface ErrorPayload {
-  message?: string;
-  code?: number;
-}
-
-class FatalError extends Error {
-  constructor(public payload: ErrorPayload) {
-    super(payload.message);
-  }
-}
+  DebugSection & {
+    answer: ReturnType<typeof answerApi.reducer>;
+    query?: QueryState;
+    searchHub?: string;
+    pipeline?: string;
+  };
 
 interface GeneratedAnswerStream {
-  answerStyle: GeneratedAnswerStyle | undefined;
-  contentFormat: GeneratedContentFormat | undefined;
-  answer: string | undefined;
-  citations: GeneratedAnswerCitation[] | undefined;
-  generated: boolean;
+  answerStyle?: GeneratedAnswerStyle;
+  contentFormat?: GeneratedContentFormat;
+  answer?: string;
+  citations?: GeneratedAnswerCitation[];
+  generated?: boolean;
   isStreaming: boolean;
   isLoading: boolean;
+  error?: {message: string; code: number};
 }
-interface HeaderMessage {
-  answerStyle: GeneratedAnswerStyle;
-  contentFormat: GeneratedContentFormat;
+
+interface StreamPayload
+  extends Pick<
+    GeneratedAnswerStream,
+    'answerStyle' | 'contentFormat' | 'citations'
+  > {
+  textDelta?: string;
+  padding?: string;
+  answerGenerated?: boolean;
 }
 
 type PayloadType =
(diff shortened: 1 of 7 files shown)


packages/headless/src/api/knowledge/stream-answer-api.ts
```diff
@@ -172,33 +191,6 @@ const updateCacheWithEvent = (
...
-    throw new Error();
-  }
-};
-
-const onError = (err: Error) => {
-  if (err instanceof FatalError) {
-    throw err;
-  }
-};
-
```
dmgauthier, line 201: This was useless

packages/headless/src/api/knowledge/stream-answer-api.ts
```diff
@@ -224,11 +217,11 @@ export const answerApi = answerSlice.injectEndpoints({
...
          */
-        const {configuration} = getState() as unknown as StateNeededByAnswerAPI;
-        const {platformUrl, organizationId, accessToken, knowledge} =
-          configuration;
+        const {configuration, generatedAnswer} =
+          getState() as unknown as StateNeededByAnswerAPI;
+        const {platformUrl, organizationId, accessToken} = configuration;
         await fetchEventSource(
-          `${platformUrl}/rest/organizations/${organizationId}/answer/v1/configs/${knowledge.answerConfigurationId}/generate`,
+          `${platformUrl}/rest/organizations/${organizationId}/answer/v1/configs/${generatedAnswer.answerConfigurationId}/generate`,
```
dmgauthier, line 224: The `answerConfigurationId` is now handle locally, not in the configuration slice

packages/headless/src/api/knowledge/stream-answer-api.ts
```diff
@@ -239,49 +232,46 @@ export const answerApi = answerSlice.injectEndpoints({
...
-            onopen: onOpenStream,
             onmessage: (event) => {
               updateCachedData((draft) => {
                 updateCacheWithEvent(event, draft);
               });
             },
-            onerror: onError,
+            onerror: (error) => {
+              throw error;
+            },
```
dmgauthier, line 242: I the rare occurence were the answerAPI itself will return an error. This will likely never be used by the stream itself.

alexprudhomme (approved):

fpbrault (approved):

packages/headless/src/features/generated-answer/generated-answer-selectors.ts
```diff
@@ -15,3 +16,9 @@ export function generativeQuestionAnsweringIdSelector(
 ) {
   return state.search?.response?.extendedResults?.generativeQuestionAnsweringId;
 }
+
+export const selectFieldsToIncludeInCitation = createSelector(
+  (state: Partial<GeneratedAnswerSection>) =>
+    state.generatedAnswer?.fieldsToIncludeInCitations,
+  (fieldsToInclude) => fieldsToInclude
+);
```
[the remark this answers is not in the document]
dmgauthier: I could leave the campfire spot in a better shape than when I came, like a boy Scout. But I did not dare to do so.

The simple answer is that the two other selectors did not leverage `createSelector` at first. But I think the cashing would go no hurt. If you want, I can update the code here and modify the selectors to make them use `createSelector`

pathiery (approved):

packages/headless/src/api/knowledge/stream-answer-api.ts
```diff
@@ -148,14 +144,37 @@ const handleEndOfStream = (
...
 
+interface MessageType {
+  payloadType: PayloadType;
+  payload: string;
+  finishReason?: string;
+  errorMessage?: string;
+  code?: number;
+}
+
+const handleError = (draft: GeneratedAnswerStream, message: MessageType) => {
```
fbeaudoincoveo, line 155: If you're 100% sure all `MessageType` properties are going to be defined on the `message`  argument when `handleError` is called (which seems like a fair assumption!) you could do:

```
const handleError = (draft: GeneratedAnswerStream, message: Required<MessageType>) => {
```

and then omit the `!` when setting `draft.error`.

packages/headless/src/api/knowledge/stream-answer-api.ts
```diff
@@ -148,14 +144,37 @@ const handleEndOfStream = (
...
-  const message: {payloadType: PayloadType; payload: string} = JSON.parse(
-    event.data
-  );
-  const parsedPayload = JSON.parse(message.payload);
+  const message: MessageType = JSON.parse(event.data);
+  if (message.finishReason === 'ERROR' && message.errorMessage) {
+    handleError(draft, message);
+  }
+
+  const parsedPayload = message.payload.length
```
fbeaudoincoveo, line 175: Better typing would be nice here 😄

fbeaudoincoveo (approved):

packages/headless/src/api/knowledge/stream-answer-api.ts
dmgauthier: I cannot believe I never used the Required generic.

packages/headless/src/api/knowledge/stream-answer-api.ts
```diff
@@ -63,55 +58,10 @@ type PayloadType =
...
-}) => {
-  citationsSchema.validate(citations);
-};
-
-const validateEndOfStream = (endOfStream: {answerGenerated: boolean}) => {
-  new Schema({
-    answerGenerated: new BooleanValue(),
-  }).validate(endOfStream);
-};
-
```
dmgauthier, line 109: Instead of using shemas to type check function that arent client facing, I used simpler conditions

PR #4148, opened 2024-07-04
feat(headless): trigger request answer from search request

dmgauthier:
## Summary

Addition of a listener that will call the answer API when a new search request ID is detected.

## Why is it needed

The AnswerApi stream will not be automatically handled by the search like the current flow is. The api must be queried explicitly.

## How it works

When a request ID is updated in the state by the regular search pattern, the knowledge answer api will trigger a call for a stream with the query.
[simplescreenrecorder-2024-07-15_16.38.46.webm](https://github.com/user-attachments/assets/f7316535-9307-4943-bcb1-156f33adefbf)

### DIFF
--- a/packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
+++ b/packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
@@ -2,6 +2,8 @@ import {
   answerApi,
   fetchAnswer,
   selectAnswer,
+  selectAnswerTriggerParams,
+  StateNeededByAnswerAPI,
 } from '../../../api/knowledge/stream-answer-api';
 import {SearchEngine} from '../../../app/search-engine/search-engine';
 import {
@@ -34,6 +36,25 @@ interface KnowledgeGeneratedAnswerProps extends GeneratedAnswerProps {}
 export interface SearchAPIGeneratedAnswerAnalyticsClient
   extends GeneratedAnswerAnalyticsClient {}
 
+const subscribeToSearchRequest = (
+  engine: SearchEngine<StateNeededByAnswerAPI>
+) => {
+  let lastTriggerParams: ReturnType<typeof selectAnswerTriggerParams>;
+  const strictListener = () => {
+    const state = engine.state;
+    const triggerParams = selectAnswerTriggerParams(state);
+    if (triggerParams.requestId === undefined) {
+      return;
+    }
+    if (JSON.stringify(triggerParams) === JSON.stringify(lastTriggerParams)) {
+      return;
+    }
+    lastTriggerParams = triggerParams;
+    engine.dispatch(fetchAnswer(state));
+  };
+  engine.subscribe(strictListener);
+};
+
 /**
  * Creates a `GeneratedAnswer` controller instance using the Answer API stream pattern.
  *
@@ -58,6 +79,8 @@ export function buildKnowledgeGeneratedAnswer(
   const getState = () => engine.state;
 
   engine.dispatch(updateAnswerConfigurationId(props.answerConfigurationId!));
+
+  subscribeToSearchRequest(engine as SearchEngine<StateNeededByAnswerAPI>);
 
   return {
     ...controller,
--- a/packages/headless/src/features/generated-answer/generated-answer-selectors.ts
+++ b/packages/headless/src/features/generated-answer/generated-answer-selectors.ts
@@ -1,23 +1,28 @@
 import {createSelector} from '@reduxjs/toolkit';
+import {GeneratedAnswerCitation} from '../../controllers/generated-answer/headless-generated-answer';
 import {SearchAppState} from '../../state/search-app-state';
 import {GeneratedAnswerSection} from '../../state/state-sections';
+import {selectQuery} from '../query/query-selectors';
+
+export const generativeQuestionAnsweringIdSelector = (
+  state: Partial<SearchAppState>
+) => state.search?.response?.extendedResults?.generativeQuestionAnsweringId;
+
+export const selectFieldsToIncludeInCitation = (
+  state: Partial<GeneratedAnswerSection>
+) => state.generatedAnswer?.fieldsToIncludeInCitations;
 
 export const citationSourceSelector = createSelector(
-  (state: Partial<GeneratedAnswerSection>, citationId: string) =>
-    state.generatedAnswer?.citations?.find(
-      (citation) => citation.id === citationId
-    ),
-  (citation) => citation
+  (state: Partial<GeneratedAnswerSection>) => state.generatedAnswer?.citations,
+  (_state: Partial<GeneratedAnswerSection>, citationId: string) => citationId,
+  (citations, citationId) =>
+    citations?.find(
+      (citation: GeneratedAnswerCitation) => citation.id === citationId
+    )
 );
 
-export const generativeQuestionAnsweringIdSelector = createSelector(
-  (state: Partial<SearchAppState>) =>
-    state.search?.response?.extendedResults?.generativeQuestionAnsweringId,
-  (generativeQuestionAnsweringId) => generativeQuestionAnsweringId
-);
-
-export const selectFieldsToIncludeInCitation = createSelector(
-  (state: Partial<GeneratedAnswerSection>) =>
-    state.generatedAnswer?.fieldsToIncludeInCitations,
-  (fieldsToInclude) => fieldsToInclude
+export const selectAnswerTriggerParams = createSelector(
+  (state) => selectQuery(state)?.q,
+  (state) => state.search.requestId,
+  (q, requestId) => ({q, requestId})
 );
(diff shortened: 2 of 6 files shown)


packages/headless/src/features/search/search-actions.ts
```diff
@@ -110,6 +114,12 @@ export const executeSearch = createAsyncThunk<
   'search/executeSearch',
   async (searchAction: TransitiveSearchAction, config) => {
     const state = config.getState();
+
+    if (
+      (state as StateNeededByAnswerAPI).generatedAnswer.answerConfigurationId
```
fbeaudoincoveo, line 119: You could use a type guard function here I think

olamothe (commented):
> The goal is to insert, in the current search flow, the request for an answer from the AnswerApi stream. Using the least amount of change possible in the code.

Can you explain why that is the goal ? This is a temporary change ? Could it be done instead in : 
```
headless-searchapi-generated-answer.ts : subscribeToSearchRequests()
```

Something like:  
```
if(state.generatedAnswer.answerConfigurationId) {
   engine.dispatch(fetchAnswer())
} else {
   engine.dispatch(streamAnswer();
}
```

fbeaudoincoveo (approved):

olamothe (changes requested):
Due to my comment above, this is not something we should do, but instead look into `headless-searchapi-generated-answer`

By dispatching the fetch stream "inside" the body of this function (and not letting the reducer do its job of actually processing the outcome of the search request), there's a bunch of bugs linked to this.

For example, there could be an unhandled search error that would happen downstream, but we would incorrectly launch a `fetchAnswer()` request.

The `state` variable (that you pass to `fetchAnswer()` is actually incorrect at that point in time. The state does not actually reflect the content of the interface (for example what's in the search box, or any of the results returned by the search API), because the reducer did not have the time to do its job (nor all the code that happen downstream in the `processor()`)

So, this approach cannot work.

packages/headless/src/features/search/search-actions.ts
```diff
@@ -110,6 +114,12 @@ export const executeSearch = createAsyncThunk<
   'search/executeSearch',
   async (searchAction: TransitiveSearchAction, config) => {
     const state = config.getState();
+
+    if (
```
fbeaudoincoveo, line 118: Could you do something like this instead? 

https://github.com/coveo/ui-kit/pull/4129/files#diff-644461440e8a66297a77b95e54021c1e4840bad4c475302dea37bfc17945091cR70

pathiery:
> Due to my comment above, this is not something we should do, but instead look into `headless-searchapi-generated-answer`

Hmm, I'm a bit confused by this: (please excuse my limited knowledge of this part of the UI-KIT)

> 
> By dispatching the fetch stream "inside" the body of this function (and not letting the reducer do its job of actually processing the outcome of the search request), there's a bunch of bugs linked to this.
> 

We don't need nor want the reducer to process the outcome of the search request (and why would we ?). The whole point of the Answer API, as described in the [RFC](https://docs.google.com/document/d/1jCOd0arSHLmXqI5w8x8W_zuTpr98GEFsf32Z_D3e1uI/edit) and the [Design doc](https://coveord.atlassian.net/wiki/spaces/PRODUCT/pages/4053860622/Answers+API+Design+Document) is that the search request and the answer request are decoupled.

> For example, there could be an unhandled search error that would happen downstream, but we would incorrectly launch a `fetchAnswer()` request.
> 

Yup, that's fine. The answer request will then probably generate the same error which the answer controller/component (not sure which layer exactly but you get my point) should handle separately. 

> The `state` variable (that you pass to `fetchAnswer()` is actually incorrect at that point in time. The state does not actually reflect the content of the interface (for example what's in the search box, or any of the results returned by the search API), because the reducer did not have the time to do its job (nor all the code that happen downstream in the `processor()`)
> 

Once again, we don't want to wait for results or any search api response to send an answer request. I might even argue that we want to send the answer request before the search request since we need less info and the process to get an answer is longer than to get results, but that's a conversation for another time.
I'm curious about how the state doesn't reflect "what's in the search box" when a search request can be built from the state, so it does have the query ? When testing this, the query was effectively passed to the underlying API call, so I'm guessing it's available ? I'd be happy to get a better understanding of that one 🤔


> So, this approach cannot work.

With all this, I'm challenging your idea that this cannot work. I think the process of generating an answer was not made clear enough to the DX team, because you seem to have assumptions that aren't applicable to the AnswerAPI.

I'd be happy to get your response to all this !

fbeaudoincoveo (changes requested):
Changing my review to request changes because I agree with @olamothe 's comments

fbeaudoincoveo:
> I'm curious about how the state doesn't reflect "what's in the search box" when a search request can be built from the state, so it does have the query ? When testing this, the query was effectively passed to the underlying API call, so I'm guessing it's available ? I'd be happy to get a better understanding of that one 🤔

One thing to consider is that the auto correction feature can kick, or a query pipeline rule (e.g., query trigger statement) might modify the original query, which would then be correctly reflected in the state only after the execute search action is fulfilled.

packages/headless/src/features/search/search-actions.ts
dmgauthier: I will remove changes to this file entirely

packages/headless/src/features/search/search-actions.ts
dmgauthier: I will remove all changes here

pathiery:
> Changing my review to request changes because I agree with @olamothe 's comments

I invite you to read my comment, read the linked RFC and Design Doc and discuss this with @olamothe. I had an extensive discussion with him that I will not duplicate here, but none of these concerns actually apply to the answer api.

@olamothe's last remaining concern was that we're shipping answer api support to all implementations by doing this like so, which is raising the bundle size and somewhat a not great separation of concerns. He did mention we should update our `headless-answerapi-generated-answer` controller to subscribe to the engine's `executeSearch` event. (which is kind of funny considering that that's something @louis-bompart advised Danny not to do)

This would alleviate the remaining concerns.

pathiery (approved):

olamothe (approved):

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
```diff
@@ -42,6 +42,25 @@ export type EngineNeededForKnowledgeGeneratedAnswer = CoreEngine<
     DebugSection & {answer: ReturnType<typeof answerApi.reducer>}
 >;
 
+const subscribeToSearchRequest = (
+  engine: EngineNeededForKnowledgeGeneratedAnswer
+) => {
+  let lastRequestId: string;
```
olamothe, line 48: This will retrigger a new `fetchAnswer()` whenever `requestID` changes: 

Should it instead be "whenever `requestID` + `q` changes?"

For example, if the user "loads more results" (pager), but don't change the intent of the query (`q` does not change), I don't think it's meaningful to resend a request to `fetchAnswer()` ?

Here I talk about pager, but it could be facetting change, sorting changes etc.

olamothe: This can be done later by the way;

I have a hunch the best approach will be to have a shortlist of parameters that, even if they do end up requesting a new request to search API when they change, that they should not automatically 1 to 1 retrigger a gen ai request.

fbeaudoincoveo (approved):

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
louis-bompart: ^ I would advise creating a (memoized) selector that returns those parameters in an object (or a hash, tho I think the former would be more efficient) and do a reference (=== or Object.is) comparison on the previous value returned from the selector and the new one.

You'll be able just to add parameters as required as you go easily.

packages/headless/src/features/generated-answer/generated-answer-selectors.ts
```diff
@@ -1,17 +1,23 @@
...
-    (citation) => citation.id === citationId
-  );
-}
+export const citationSourceSelector = createSelector(
+  (state: Partial<GeneratedAnswerSection>, citationId: string) =>
+    state.generatedAnswer?.citations?.find(
+      (citation) => citation.id === citationId
+    ),
+  (citation) => citation
+);
```
louis-bompart, line 11: https://redux.js.org/usage/deriving-data-selectors#writing-memoized-selectors-with-reselect

You swapped things around: the first parameters, the input selector must be 'light-weight' and the output selector should do the heavy lifting

> "input selectors" should usually just extract and return values, and the "output selector" should do the transformation work

dmgauthier: I knew this...  of course I knew this, right :badpokerface:

packages/headless/src/controllers/knowledge/generatedAnswer/headless-knowledge-generated-answer.ts
dmgauthier: I will apply the suggestion and create a selector with the triggerParams because you guys are right about it.

It is good to know though that a requestId change with the same `q` wouldn't have called the endpoint again because of the infamous caching mechanism of RTK query.

PR #4153, opened 2024-07-05
test(atomic): rework the interface of "load" in base-page-object

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-3367

I kept running into issues like this :

image


packages/atomic/src/components/commerce/atomic-commerce-load-more-products/e2e/page-object.ts
```diff
@@ -29,4 +29,7 @@ export class LoadMoreProductsPageObject extends BasePageObject<'atomic-commerce-
   get progressValue() {
     return this.page.locator('[part="progress-bar"] > div');
   }
+  get loadMoreButton() {
+    return this.page.getByRole('button', {name: 'Load more results'});
+  }
```
alexprudhomme, line 34: This is random but it was somehow missing and test were failing because of it ? This getter was called in the `atomic-search-box.e2e.ts`

olamothe (approved):

PR #4154, opened 2024-07-05
feat(atomic): clickable "More matches for {query}" in facetSearch

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-3361

Still need to implement the feature for the commerce use case in a future PR.

BEFORE
image

AFTER
image
increments like this :
numberOfValues += initalNumberOfValues 
image


packages/atomic/src/components/common/facets/facet-search/facet-search-matches.tsx
```diff
@@ -22,13 +24,25 @@ function matchesFound(
...
+  i18n: i18n
+) {
+  return i18n.t(key, {
+    query: `<span class="font-bold italic" part="matches-query">${escape(
+      query
+    )}</span>`,
+    interpolation: {escapeValue: false},
+  });
+}
+
```
alexprudhomme, line 39: This is temporary while before https://coveord.atlassian.net/browse/KIT-3368 (implementation for commerce) is done

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -59,7 +53,6 @@ export class AtomicInsightFacet
   @InitializeBindings() public bindings!: InsightBindings;
   public facet!: InsightFacet;
   public searchStatus!: InsightSearchStatus;
-  public withSearch = false;
```
alexprudhomme, line 62: The insight facet never had support for facet search. This is a relic of the past before the refactor of facet-common. 

Delete !!!! Less bundle size 😋

packages/atomic/src/components/search/facets/atomic-category-facet/e2e/atomic-category-facet.e2e.ts
```diff
@@ -0,0 +1,30 @@
...
+    await facet.getFacetSearch.pressSequentially('p');
+    await expect
+      .poll(async () => {
+        return await facet.getFacetValue.count();
+      })
+      .toBeGreaterThanOrEqual(2);
+
+    await facet.facetSearchMoreMatchesFor.click();
+    await expect
+      .poll(async () => {
```
fbeaudoincoveo, line 18: From what I gather, the reason why we're polling here is that we don't have a visual indicator (e.g., a loading spinner) that the network request is in progress. Otherwise we could base our tests on the visibility of that element.

I think it would be better if we had something like that, although it probably falls outside the scope of this PR.

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
```diff
@@ -18,7 +18,6 @@ import {
   AriaLiveRegion,
   FocusTargetController,
 } from '../../../utils/accessibility-utils';
-import {getFieldCaptions} from '../../../utils/field-utils';
```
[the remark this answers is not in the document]
olamothe: Those are okay and "required" though. We should not touch them.

packages/atomic/src/components/insight/atomic-insight-facet/atomic-insight-facet.tsx
olamothe: Insight panel components exist only for the preview feature in the admin (ie: kinda like a demo if you will)

They are not "production ready" for customers, or actually in use to power an insight panel for end user. This might happen later this year, when the service team invest in it.

I would remove the import for now.

olamothe (approved):

dmgauthier (approved):

PR #4162, opened 2024-07-09
feat(quantic): new feedback modal for RGA

erocheleau:
Mockups: https://www.figma.com/design/l63i3ZXWrVxldOR4WUBKS1/CRGA-Component?node-id=5514-14647&t=pOZ5lDKywVYJCqgI-0

SFINT-5555:

New Quantic RGA Feedback modal

image

The top 4 questions are required, and you have an example of how we tell the user when they are required and missing.

The bottom two text fields are optional.

When you press send feedback there will be the same screen as before thanking you for your feedback.

If you press skip it closes the modal. And nothing you've selected is "remembered" if you re-open it.

The modal now opens on both like and dislike (thumbs up and thumbs down), and as part of the event we send, we specify if the feedback was positive or negative. We still send a separate event already when you press thumbs up or down.

You can "change your mind", even if you send feedback, you can still change your thumbs up to a thumbs down. but it will NOT re-open the modal if you've already submitted feedback once. 
Also if you click a 2nd time on thumbs down for example we won't re-open the modal.

Everything is tested through unit tests since this is only the "content" of the modal, and not the feedback process itself that we're testing.


mmitiche:
> I changed it so that you can "change your mind", even if you send feedback, you can still change your thumbs up to a thumbs down. and it will re-open the modal. But if you click a 2nd time on thumbs down for example we won't re-open the modal.

Is this the same behaviour we have in Atomic?

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1442,4 +1463,74 @@
...
+  <labels>
+    <fullName>quantic_FeedbackQuestionTopic</fullName>
+    <value>Is the answer about the right topic?</value>
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>Feedback Question Topic</shortDescription>
+  </labels>
+  <labels>
+    <fullName>quantic_FeedbackQuestionHallucination</fullName>
+    <value>Is the answer free of hallucinated content?</value>
```
SimonMilord, line 1475:
```suggestion
    <value>Is the answer free of hallucinated content?</value>
```

Is this not too "technical" for users? Seems like a weird way to just say something like: "Is the answer accurate?"
Hallucinations are LLM/Generative AI jargon IMO so I would be tempted to keep it simple. What do you think?

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1442,4 +1463,74 @@
...
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>Feedback Question Hallucination</shortDescription>
+  </labels>
+  <labels>
+    <fullName>quantic_FeedbackQuestionDocumented</fullName>
+    <value>Can the question be answered by the documentation?</value>
```
SimonMilord, line 1482: What is the purpose of this question? 😂  I feel like most users who will ask questions to the model might not check the doc X)

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
```diff
@@ -1442,4 +1463,74 @@
...
+  <labels>
+    <fullName>quantic_FeedbackAnswerEvaluation</fullName>
+    <value>Answer evaluation</value>
+    <language>en_US</language>
+    <protected>false</protected>
+    <shortDescription>Feedback Answer Evaluation</shortDescription>
+  </labels>
+  <labels>
+    <fullName>quantic_FeedbackDocumentUrl</fullName>
+    <value>Link to correct answer</value>
```
SimonMilord, line 1510:
```suggestion
    <value>Link to the correct answer</value>
```

packages/quantic/force-app/main/translations/fr.translation-meta.xml
```diff
@@ -796,4 +808,36 @@
     <label>Voir moins</label>
     <name>quantic_GeneratedAnswerShowLess</name>
   </customLabels>
+  <customLabels>
+    <label>La réponse concerne-t-elle le bon sujet?</label>
+    <name>quantic_FeedbackQuestionTopic</name>
+  </customLabels>
+  <customLabels>
+    <label>La réponse est-elle exempte de contenu halluciné?</label>
```
SimonMilord, line 816: same comment here as above in the english file

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
erocheleau: Nope the questions have been decided by ML. The target audience is a lot more "Subject matter experts" than general users AFAIK, even if we present it to everyone.
Accurate is already pretty much covered by the other questions too.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
erocheleau: It's for subject matter experts to tell us.

packages/quantic/force-app/main/default/labels/CustomLabels.labels-meta.xml
erocheleau: That's coming from the mockup, not my decision.

packages/quantic/force-app/main/translations/fr.translation-meta.xml
erocheleau: Same answer :)

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -297,16 +315,16 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
-    this.generatedAnswer.dislike?.();
-    if (!this.feedbackSubmitted) {
+    if (!this._disliked) {
+      this._disliked = true;
+      this._liked = false;
+      this.generatedAnswer.dislike?.();
       // @ts-ignore
-      await FeedbackModal.open({
-        label: this.labels.feedback,
+      await FeedbackModalQna.open({
```
SimonMilord, line 323: if we are no longer opening the feedbackModal, but the FeedbackModalQna here shouldn't we remove the options getter and associated labels?

erocheleau:
> > I changed it so that you can "change your mind", even if you send feedback, you can still change your thumbs up to a thumbs down. and it will re-open the modal. But if you click a 2nd time on thumbs down for example we won't re-open the modal.
> 
> Is this the same behaviour we have in Atomic?

Yea you were right to ask, I updated the PR.

If you've submitted the feedback once, thumbs up or down, you won't see the modal again until after you do a new query.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,192 @@
...
+   * @type {function}
+   */
+  @api handleSubmit;
+
+  /**
+   * @api
+   * If the feedback modal was opened on the thumbs up. (otherwise it was on a thumbs down).
+   * @type {boolean}
+   */
+  @api liked;
```
SimonMilord, line 72:
```suggestion
  /**
   * @api
   * The function that will be executed when the feedback is submitted.
   * @type {function}
   */
  @api handleSubmit;
  /**
   * @api
   * If the feedback modal was opened on the thumbs up. (otherwise it was on a thumbs down).
   * @type {boolean}
   */
  @api liked;
```

SimonMilord, line 72:
```suggestion
  /**
   * @api
   * The function that will be executed when the feedback is submitted.
   * @type {function}
   */
  @api handleSubmit;
  /**
   * @api
   * If the feedback modal was opened on the thumbs up. (otherwise it was on a thumbs down).
   * @type {boolean}
   */
  @api liked;
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,192 @@
...
+      documentUrl: this.documentUrlValue,
+      details: this.detailsValue,
+      helpful: this.liked,
+    });
+    this.isFeedbackSubmitted = true;
+  }
+
+  /**
+   * Returns the right HTML template.
+   */
```
SimonMilord, line 185:
```suggestion
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,192 @@
...
+
+  get feedbackQuestionsRadios() {
+    return this.template.querySelectorAll('lightning-radio-group');
+  }
+
+  get feedbackFormIsValid() {
+    let valid = true;
+    const feedbackQuestions = this.feedbackQuestionsRadios;
+    feedbackQuestions.forEach((question) => {
+      if (question?.checkValidity() !== true) {
```
SimonMilord, line 145: where is `checkValidity()` defined? Did I miss something?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,192 @@
...
+   */
+  closeModal() {
+    this.close();
+  }
+
+  get feedbackQuestionsRadios() {
+    return this.template.querySelectorAll('lightning-radio-group');
+  }
+
+  get feedbackFormIsValid() {
```
SimonMilord, line 141: nice!

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+  questionHallucinationRadio:
+    'lightning-radio-group[data-name="hallucinationFree"]',
+  questionDocumentedRadio: 'lightning-radio-group[data-name="documented"]',
+  questionReadableRadio: 'lightning-radio-group[data-name="readable"]',
+  documentUrlInput: 'lightning-input[data-name="documentUrl"]',
+  additionalDetailsTextarea: 'lightning-textarea[data-name="details"]',
+  skipButton: 'lightning-button.skip',
+  submitButton: 'lightning-button.submit-feedback',
+  requiredMessages: 'slds-form-element__help',
+  successMesage: '.feedback-modal-body__success-message',
```
SimonMilord, line 34:
```suggestion
  successMessage: '.feedback-modal-body__success-message',
```

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+  questionTopicRadio: 'lightning-radio-group[data-name="correctTopic"]',
+  questionHallucinationRadio:
+    'lightning-radio-group[data-name="hallucinationFree"]',
+  questionDocumentedRadio: 'lightning-radio-group[data-name="documented"]',
+  questionReadableRadio: 'lightning-radio-group[data-name="readable"]',
+  documentUrlInput: 'lightning-input[data-name="documentUrl"]',
+  additionalDetailsTextarea: 'lightning-textarea[data-name="details"]',
+  skipButton: 'lightning-button.skip',
+  submitButton: 'lightning-button.submit-feedback',
+  requiredMessages: 'slds-form-element__help',
```
SimonMilord, line 33: this selector doesn't seem used anywhere, is it extra or are you missing something in a test?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+    );
+    expect(additionalDetailsInput).not.toBeNull();
+
+    const skipButton = element.shadowRoot.querySelector(selectors.skipButton);
+    expect(skipButton).not.toBeNull();
+
+    const submitButton = element.shadowRoot.querySelector(
+      selectors.submitButton
+    );
+    expect(submitButton).not.toBeNull();
```
SimonMilord, line 126: I would use an array of selectors here to just check with a forEach()

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+  });
+
+  describe('when submitting the feedback', () => {
+    describe('with missing required answers', () => {
+      it('should not submit the feedback and not show the success screen', async () => {
```
SimonMilord, line 131: As mentionned I would add a check for the required message here

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+            details: expectedDetails,
+          })
+        );
+      });
+    });
+  });
+  describe('when clicking the skip button', () => {
+    it('should not submit the feedback and not execute the handleSubmit function', async () => {
+      const element = createTestComponent();
+      await flushPromises();
```
SimonMilord, line 269:
```suggestion
      });
    });
  });
  
  describe('when clicking the skip button', () => {
    it('should not submit the feedback and not execute the handleSubmit function', async () => {
      const element = createTestComponent();
      await flushPromises();
```

mmitiche:
I'm noticing some differences with the mockups we have, while they maybe are not all adjustable because we are using SF components, some others I think are adjustable:

Mockups:
Screenshot 2024-07-15 at 10 40 10 AM

Real compomnent:
Screenshot 2024-07-15 at 10 43 47 AM

- The label `Answer Evaluation` is not with the right color, it currently a weak text color while in the mockup it's not.
- The label of the two text inputs should be bold and with a black color, it's currently not bold and with a weak text color.
- The label in the modal bottom should be `Required fields` instead of `This field is required`. not a big deal but it's better to stay consistent

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
erocheleau: yea no you're correct, it was before I realized the inputs were just stubs in tests (so even if they are invalid they don't display their required message) I was trying to make a test to make sure they were displayed, I'll remove it.

mmitiche:
one of the other differences is that the end of the text inputs is not well aligned with the end of the radio buttons input like in the mokups, this happens in narrower screens, maybe we could find a better way to align this:

Screenshot 2024-07-15 at 10 55 37 AM
Screenshot 2024-07-15 at 10 56 06 AM


https://github.com/user-attachments/assets/3fc377d3-d2bd-48d2-b29f-4ae17c36c6bf

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
erocheleau: I can't. Jest lightning-* components don't actually "exist". They are only stubs in the DOM with some public attributes accessible to tests.

Meaning they don't actually react to display a required message while in unit tests.

erocheleau:
> one of the other differences is that the end of the text inputs is not well aligned with the end of the radio buttons input like in the mokups, this happens in narrower screens, maybe we could find a better way to align this:
> 
> Screenshot 2024-07-15 at 10 55 37 AM Screenshot 2024-07-15 at 10 56 06 AM
>  Screen.Recording.2024-07-15.at.10.57.54.AM.mov

I'll check for narrower screens thanks. 
Didn't think anything I was using would be impacted by that.

erocheleau:
> I'm noticing some differences with the mockups we have, while they maybe are not all adjustable because we are using SF components, some others I think are adjustable:
> 
> Mockups: Screenshot 2024-07-15 at 10 40 10 AM
> 
> Real compomnent: Screenshot 2024-07-15 at 10 43 47 AM
> 
> * The label `Answer Evaluation` is not with the right color, it currently a weak text color while in the mockup it's not.
> * The label of the two text inputs should be bold and with a black color, it's currently not bold and with a weak text color.
> * The label in the modal bottom should be `Required fields` instead of `This field is required`. not a big deal but it's better to stay consistent

For the labels colors, it's expected and validated with Vyka already.

The things is the "Answer Evaluation" is a style from slds as a title/sub-title.
The other two (link to correct answer and additional notes) are input labels (a styling I cannot control.

I think this is simply due to the fact that we don't have a "Quantic" version of the mockup.

Good catch on the text for "Required fields" I'll update it.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+describe('c-quantic-feedback-modal-qna', () => {
+  afterEach(() => {
+    // The jsdom instance is shared across test cases in a single file so reset the DOM
+    while (document.body.firstChild) {
+      document.body.removeChild(document.body.firstChild);
+    }
+    jest.clearAllMocks();
+  });
+
+  it('loads correctly', async () => {
```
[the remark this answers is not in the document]
erocheleau: It already says the name of the component in the describe above.

But I changed it to `should properly render` if you prefer that.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
erocheleau: No, because the tests if an element fail to render, would fail inside the loop and we would have no idea which element didn't render correctly.

If we keep it like this, we know exactly which expect failed and which element it was.

In tests, you want to be explicit even if it means writing more code.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,188 @@
...
+    });
+    return valid;
+  }
+
+  documentUrlChangeHandler(event) {
+    this.documentUrlValue = event?.target?.value;
+  }
+
+  detailsChangeHandler(event) {
+    this.detailsValue = event?.target?.value;
```
mmitiche, line 158: same thing here.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,188 @@
...
+  ];
+
+  answerOptions = [
+    {label: this.labels.answerYes, value: 'yes'},
+    {label: this.labels.answerNotSure, value: 'unknown'},
+    {label: this.labels.answerNo, value: 'no'},
+  ];
+
+  /** @type {string} */
+  error;
```
mmitiche, line 128: I don't see that this is being used anywhere?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/success.html
mmitiche, line 1: usually we add the templates in a folder named `templates` for a better organization, it would be nice to keep following that pattern.

mmitiche: also I don't see the success screen in the mockups, so just want to double check, do we really want it? it exists in Atomic as well?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,190 @@
...
+   * The function that will be executed when the feedback is submitted.
+   * @type {function}
+   */
+  @api handleSubmit;
+  /**
+   * @api
+   * If the feedback modal was opened on the thumbs up. (otherwise it was on a thumbs down).
+   * @type {boolean}
+   */
+  @api liked;
```
mmitiche, line 73: I don't think that this property is really necessary for this component as I see that it is only used when calling the `handleSubmit` method which passed from the QuanticGeneratedAnswer component, and in the QuanticGeneratedAnswer we have the information about whether the answer is liked or not, so no need it to pass it here.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -286,9 +292,21 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
+      this._liked = true;
+      this._disliked = false;
+      this.generatedAnswer.like?.();
+      // @ts-ignore
+      await FeedbackModalQna.open({
+        size: 'small',
+        liked: true,
+        label: this.labels.feedbackHelpUsImprove,
+        handleSubmit: this.submitFeedback.bind(this),
+      });
```
mmitiche, line 307:
```suggestion
      await FeedbackModalQna.open({
        size: 'small',
        label: this.labels.feedbackHelpUsImprove,
        handleSubmit: (payload) => {this.submitFeedback(...payload, helpful: true)},
      });
```
you may also need to transform `submitFeedback` to an arrow function to avoid the `bind(this)`

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -297,16 +315,16 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
-        label: this.labels.feedback,
+      await FeedbackModalQna.open({
         size: 'small',
-        description: this.labels.feedback,
-        options: this.options,
+        liked: false,
+        label: this.labels.feedbackHelpUsImprove,
         handleSubmit: this.submitFeedback.bind(this),
-        optionsLabel: this.labels.whyGeneratedAnswerWasNotHelpful,
       });
```
mmitiche, line 328: We could isolate the logic to open the modal in a separate function so we can just call it for the like handler and the dislike handler instead of having to write this same logic twice.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+function flushPromises() {
+  return new Promise((resolve) => setTimeout(resolve, 0));
+}
+
+function lightningElementValueChange(element, selector, value) {
+  const lightningElement = element.shadowRoot.querySelector(selector);
+  // All base components in Jest tests are stubs. So we have to set the value manually.
+  // Including to the checkValidity method.
+  lightningElement.value = value;
+  lightningElement.validity = true;
```
mmitiche, line 64: is it necessary to set `validity` to true? forcing `checkValidity` to return `true` should be enough I think

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+
+// Helper function to wait until the microtask queue is empty.
+function flushPromises() {
+  return new Promise((resolve) => setTimeout(resolve, 0));
+}
+
+function lightningElementValueChange(element, selector, value) {
+  const lightningElement = element.shadowRoot.querySelector(selector);
+  // All base components in Jest tests are stubs. So we have to set the value manually.
+  // Including to the checkValidity method.
```
mmitiche, line 62: I think this comment should be adjusted,

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
```diff
@@ -0,0 +1,299 @@
...
+        );
+        lightningElementValueChange(
+          element,
+          selectors.questionDocumentedRadio,
+          'yes'
+        );
+        lightningElementValueChange(
+          element,
+          selectors.questionHallucinationRadio,
+          'yes'
```
mmitiche, line 175: I wouldn't put all the expected answers to `yes` to also check that our component can return other values as well and not always `yes`  whatever what option you select.

mmitiche (commented):
Very nice job! just few adjustments suggested👍

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: Interesting, this implies the feedback modal doesn't know if it was triggered by like or dislike

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: I don't see much value in doing this, especially when we have at least 1 difference

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
```diff
@@ -317,11 +335,12 @@ export default class QuanticGeneratedAnswer extends LightningElement {
...
-      this.generatedAnswer.sendDetailedFeedback(feedbackPayload.details);
-    } else if (feedbackPayload?.value) {
-      this.generatedAnswer.sendFeedback(feedbackPayload.value);
-    }
+    this.generatedAnswer.sendFeedback(feedbackPayload);
+    // if (feedbackPayload?.details) {
+    //   this.generatedAnswer.sendDetailedFeedback(feedbackPayload.details);
+    // } else if (feedbackPayload?.value) {
+    //   this.generatedAnswer.sendFeedback(feedbackPayload.value);
+    // }
```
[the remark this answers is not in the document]
erocheleau: huh? that was gone for a while when you reviewed this no?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.css
```diff
@@ -0,0 +1,7 @@
+:host {
+    --slds-c-textarea-sizing-min-height: 100px;
+}
+
+.feedback-modal-qna__radio-group {
+    min-width: 186px;
```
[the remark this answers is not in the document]
erocheleau: yea, I can calculate it

Not sure what I want to happen if we increase the font size by a lot though

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,188 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
erocheleau: Damn I forgot I put that, it is in the quanticFeedbackModal.js file as well xD

I'll see if I can fix it.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,188 @@
...
+import successTemplate from './success.html';
+
+/** @typedef {"yes"|"unknown"|"no"} GeneratedAnswerFeedbackOption */
+
+/**
+ * The `QuanticFeedbackModalQna` component overlays a message modal on top of the current app window, the modal contains a form that allows the user to give feedback on a CRGA response.
+ *
+ * Under the hood, the component relies on a [`lightningModal`](https://developer.salesforce.com/docs/component-library/bundle/lightning-modal/documentation) component.
+ * For an example of how to use the `QuanticFeedbackModal` component, see the [`quanticSmartSnippet`](https://github.com/coveo/ui-kit/blob/master/packages/quantic/force-app/main/default/lwc/quanticSmartSnippet/quanticSmartSnippet.js) implementation.
+ *
```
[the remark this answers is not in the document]
erocheleau: Updated to what?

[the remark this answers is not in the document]
mmitiche: we need to talk about the `QuanticFeedbackModalQna` instead of talking about `QuanticFeedbackModal`.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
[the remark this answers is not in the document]
mmitiche: I know that we skipped the tests, but the logic was not removed before this pr

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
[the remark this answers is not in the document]
erocheleau: ohhh yea

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
erocheleau: wut?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
erocheleau: yea good simplification!

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,188 @@
...
+   */
+  closeModal() {
+    this.close();
+  }
+
+  get feedbackQuestionsRadios() {
+    return this.template.querySelectorAll('lightning-radio-group');
+  }
+
+  get feedbackFormIsValid() {
```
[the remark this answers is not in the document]
erocheleau: Yea that's true I'll change it

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,188 @@
...
+    feedbackQuestions.forEach((question) => {
+      if (question?.checkValidity() !== true) {
+        valid = false;
+      }
+    });
+    return valid;
+  }
+
+  documentUrlChangeHandler(event) {
+    this.documentUrlValue = event?.target?.value;
```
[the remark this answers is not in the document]
erocheleau: I guess not

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
erocheleau: yea it's in the HTML on the `onclick` of the close button or skip button

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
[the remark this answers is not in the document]
erocheleau: the `feedbackFormIsValue` is an actual property of the feedback form.

I could change it to a function as well and it would be perfectly fine to me as well.

That one was taken from an example in the salesforce doc, so it's more a property than a function for me.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
SimonMilord: Are you sure? Im pretty sure I tried it somewhere before and despite the loop it was still logging in error the component in the array that was faulty...

did that in the `hostedSearchPageRenderer.spec.ts`

image

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
SimonMilord: ok

SimonMilord (approved):

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.css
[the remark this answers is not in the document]
erocheleau: This still feels like black magic :D but it's better now.
And this was also the cause of the alignment being weird on small screens.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
erocheleau: ah because you put the `it` in the loop, your example creates 1 test per component.

My test tests all "default" components in 1 test, I don't want to have 1 test per component, otherwise it will re-create and render the component X times, instead of loading once and checking for the components to be there.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/quanticFeedbackModalQna.js
```diff
@@ -0,0 +1,188 @@
...
+      changeFn: (event) => {
+        this.correctTopicValue = event.target.value;
+      },
+      id: 'correctTopic',
+    },
+    {
+      question: this.labels.questionHallucination,
+      changeFn: (event) => {
+        this.hallucinationFreeValue = event.target.value;
+      },
```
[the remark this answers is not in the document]
erocheleau: Sure I can try that.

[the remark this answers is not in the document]
erocheleau: I tried and it doesn't fix it.
I can reproduce the same thing with the default `lightning-radio-group` even in the Salesforce lightning component examples, so I think it's on them...

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/success.html
erocheleau: Confirmed yes Atomic has it.

But I will change the message on it, since it was to be changed.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
erocheleau: 🤷 not a huge issue to have both huh?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
erocheleau: To what?

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
erocheleau: It doesn't matter which value I put, I'm the one setting the value.

So even if I put 'abracadabra', in the end I'm the one that says `value='abracadabra'` and then executing `get value` so essentially I can only test that the code gets the value I set. Changing the string doesn't add anything to the test.

Keep in mind, in reality I am not clicking, nor selecting an option, I am setting the `lightning-radio-group` stub's `value` attribute to any string I want. And when the code of `quanticFeedbackModalQna` checks for the value, it simply returns the value I've set on the stub.

packages/quantic/force-app/main/default/lwc/quanticGeneratedAnswer/quanticGeneratedAnswer.js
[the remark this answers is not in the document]
mmitiche: Ah I see I see x) all good then x)

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
mmitiche: It's just that `Including to the checkValidity method` doesn't seem like a correct english sentence x)

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
mmitiche: That's where we feel the value of E2E tests to come and complement the unit tests.

packages/quantic/force-app/main/default/lwc/quanticFeedbackModalQna/__tests__/quanticFeedbackModalQna.test.js
mmitiche: just so to slightly reduce the complexity, if we don't need it, why adding it.

mmitiche (approved):
Well done 👍

erocheleau:
Following @lvu285 review:

1. I changed it so that if you don't submit your feedback (click on vote button, but click skip or X), if you click on the same vote button, it re-opens the modal.
2. Even if you submit your feedback, you can change your vote, but cannot re-open the modal, even if you change your vote.
3. Also changed the look on really smaller screens:

image

mmitiche (approved):
latest changes look good to me 👍

PR #4171, opened 2024-07-10
fix: add atomic version in the analytics initial config for Headless

louis-bompart:
Also, append instead of prepend Headless version to replace Headless-Atomic-Relay order by Atomic-Headless-Relay

https://coveord.atlassian.net/browse/KIT-3308

### DIFF
--- a/packages/atomic/src/components/common/interface/analytics-config.ts
+++ b/packages/atomic/src/components/common/interface/analytics-config.ts
@@ -66,7 +66,7 @@ export function getNextAnalyticsConfig(
     analyticsConfiguration,
     augmentAnalyticsConfigWithAtomicVersion()
   );
-  Object.assign(configuration, searchEngineConfig.analytics);
+  Object.assign(configuration, analyticsConfiguration);
 
   return configuration;
 }
--- a/packages/atomic/src/components/common/interface/analytics-config.ts
+++ b/packages/atomic/src/components/common/interface/analytics-config.ts
@@ -55,17 +55,18 @@ export function getNextAnalyticsConfig(
   searchEngineConfig: EngineConfiguration,
   enabled: boolean
 ): AnalyticsConfiguration {
-  const defaultConfiguration: AnalyticsConfiguration = {
+  const configuration: AnalyticsConfiguration = {
     enabled,
     documentLocation: document.location.href,
     ...(document.referrer && {originLevel3: document.referrer}),
   };
 
-  if (searchEngineConfig.analytics) {
-    return {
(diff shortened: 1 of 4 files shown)


packages/atomic/src/components/common/interface/analytics-config.ts
```diff
@@ -55,17 +55,18 @@ export function getNextAnalyticsConfig(
...
-      ...searchEngineConfig.analytics,
-    };
-  }
-  return defaultConfiguration;
+  const analyticsConfiguration = searchEngineConfig.analytics ?? {};
+  Object.assign(
+    analyticsConfiguration,
+    augmentAnalyticsConfigWithAtomicVersion()
+  );
+  Object.assign(configuration, searchEngineConfig.analytics);
```
alexprudhomme, line 69: why object.assign and no spread operator ?

packages/atomic/src/components/common/interface/analytics-config.ts
```diff
@@ -55,17 +55,18 @@ export function getNextAnalyticsConfig(
...
   };
 
-  if (searchEngineConfig.analytics) {
-    return {
-      ...defaultConfiguration,
-      ...searchEngineConfig.analytics,
-    };
-  }
-  return defaultConfiguration;
+  const analyticsConfiguration = searchEngineConfig.analytics ?? {};
```
alexprudhomme, line 64: Also isnt analyticsConfiguration useless here ? It is assigned and then unused

Spuffynism (approved):

packages/atomic/src/components/common/interface/analytics-config.ts
fbeaudoincoveo, line 69:
```suggestion
  Object.assign(configuration, analyticsConfiguration);
```

fbeaudoincoveo (approved):

packages/atomic/src/components/common/interface/analytics-config.ts
louis-bompart: Taste mostly. To use the spread operator here, I would need to have something like:
```
{
  ...defaultConfiguration,
  analytics: {
     ...analyticsConfiguration,
     ...augmentAnalyticsConfigWithAtomicVersion()
  }
}
```
I think it's less readable than the Object.assign

packages/atomic/src/components/common/interface/analytics-config.ts
louis-bompart: resolved w/ fred suggestion

PR #4172, opened 2024-07-10
test(atomic): add tests for atomic-commerce-interface

fpbrault:
This PR adds tests for atomic-commerce-interface

https://coveord.atlassian.net/browse/KIT-3375


packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
```diff
@@ -116,7 +117,7 @@ export class AtomicCommerceInterface
    *
    * Will default to the value set in the Headless engine context if not provided.
    */
-  @Prop({reflect: true, mutable: true}) public language?: string;
+  @Prop({reflect: true, mutable: true}) public language: string = '';
```
fpbrault, line 120: @louis-bompart Had to make this prop not optional otherwise I can't set a value for it in storybook. Do you know what could be causing this and if there is an alternative?

louis-bompart: So, there's two ways: the one you did, which can make sense tbh, as a 'reflect' param undefined doesn't make much sense imo, as this doesn't translate well in the HTML. (String, number, bools are the only one that realy does)

The other one is to tweak the argTypes in your story to override the argument definition in Storybook.
See https://storybook.js.org/docs/api/arg-types

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.new.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+    );
+    await searchInterface!.executeFirstRequest();
+  },
+};
+
+export const WithProductList: Story = {
+  tags: ['commerce', 'test'],
+  argTypes: {
+    'attributes-language': {control: 'text', table: {category: 'attributes'}},
+  },
```
louis-bompart, line 51: I see you used the argTypes already, good 👍 
However:
1. You should define it in the meta to reuse it in all the stories
2. Here's the one you need to be able to keep the optionality of the param:
```ts
  argTypes: {
    'attributes-language': {
      name: 'language',
      type: 'string',
    },
  },
  ```

Comment (louis-bompart) [inline packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.new.stories.tsx:51]:
```suggestion
```

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.new.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+  id: 'atomic-commerce-interface',
+  render: renderComponent,
+  parameters,
+  play: async (context) => {
+    await initializeSearchInterface(context.canvasElement);
+    const searchInterface = context.canvasElement.querySelector(
+      'atomic-commerce-interface'
+    );
+    await searchInterface!.executeFirstRequest();
+  },
```
louis-bompart, line 28:
```suggestion
  },
  argTypes: {
    'attributes-language': {
      name: 'language',
      type: 'string',
    },
  },
```

packages/atomic/src/components/commerce/atomic-commerce-interface/atomic-commerce-interface.tsx
fpbrault: Thank you 💯 !

packages/atomic/src/components/commerce/atomic-commerce-interface/e2e/atomic-commerce-interface.e2e.ts
```diff
@@ -0,0 +1,147 @@
...
+        await page.goto(
+          'http://localhost:4400/iframe.html?id=atomic-commerce-interface--with-product-list&viewMode=story&args=attributes-reflect-state-in-url:false'
+        );
+
+        const facetValueLabel = commerceInterface.getFacetValue('Nike');
+
+        await facetValueLabel.click();
+
+        await page.waitForTimeout(1000);
+
```
louis-bompart, line 140: Try to wait on 'something' instead than an arbitrary timeout.
For example, you could wait for something to appear in the page (I know, it'll be easier when we have a stable org/source, but still)

louis-bompart (changes requested):
LGTM, tho I'll be really pick with time-based waits in tests and we should try to avoid them

louis-bompart (approved):

packages/atomic/src/components/commerce/atomic-commerce-interface/e2e/atomic-commerce-interface.e2e.ts
```diff
@@ -0,0 +1,149 @@
...
+      });
+
+      await expect(commerceInterface.interface()).toContainText('Produits');
+    });
+    test('should revert to english after changing the language to an invalid value', async ({
+      page,
+      commerceInterface,
+    }) => {
+      await page.goto(
+        'http://localhost:4400/iframe.html?id=atomic-commerce-interface--with-product-list&viewMode=story&args=attributes-language:fr'
```
alexprudhomme, line 97: cant you do `commerceInterface.load({story:"with-product-list", args: [{language: "fr"}])` ?

alexprudhomme (approved):

PR #4175, opened 2024-07-10
fix(headless commerce): correctly determine hasBreadcrumbs value for category facets

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3385

### DIFF
--- a/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
+++ b/packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
@@ -230,23 +230,32 @@ export function buildCoreBreadcrumbManager(
     ];
   };
 
+  const hasActiveValue = (
+    facet: AnyFacetResponse | undefined
+  ): facet is AnyFacetResponse => {
+    if (!facet) {
+      return false;
+    }
+    if (facet.values.length === 0) {
+      return false;
+    }
+    if (facet.type === 'hierarchical') {
+      return findActiveValueAncestry(facet.values).length > 0;
+    }
+
+    return facet.values.some((value) => value.state !== 'idle');
+  };
+
   const commerceFacetSelector = createSelector(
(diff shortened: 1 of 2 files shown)


packages/headless/src/controllers/commerce/core/breadcrumb-manager/headless-core-breadcrumb-manager.ts
```diff
@@ -238,11 +238,13 @@ export function buildCoreBreadcrumbManager(
...
-              facet !== undefined &&
-              facet.values.length > 0 &&
-              facet.values.some((value) => value.state !== 'idle')
+          .filter((facet): facet is AnyFacetResponse =>
+            facet !== undefined &&
+            facet.values.length > 0 &&
+            facet.type !== 'hierarchical'
+              ? facet.values.some((value) => value.state !== 'idle')
+              : findActiveValueAncestry((facet as CategoryFacetResponse).values)
+                  .length > 0
```
louis-bompart, line 247: This isn't easy to read.
You should extract the filter callback function into a const:
```ts
  const hasActiveValue = (facet): facet is AnyFacetResponse => {
    if (!facet) {
      return false;
    }
    if (facet.values.length === 0) {
      return false;
    }
    return facet.type === 'hierarchical'
      ? findActiveValueAncestry(facet.values).length > 0
      : facet.values.some((value) => value.state !== 'idle');
  };
```

It will make the array functions chain more readable.

louis-bompart: Second read: breadCrumbs should be the result of a single flatMap call.
We have an iteration galore here, and that's often a smell.

fbeaudoincoveo: Much more readable indeed! Thanks!

I applied your suggestion with some adjustments

Spuffynism (approved):

PR #4176, opened 2024-07-11
fix: remove pageId from analyticsPayload

louis-bompart:
https://coveord.atlassian.net/browse/KIT-3386

### DIFF
--- a/packages/headless/src/features/configuration/analytics-params.ts
+++ b/packages/headless/src/features/configuration/analytics-params.ts
@@ -1,6 +1,5 @@
 import {EventDescription} from 'coveo.analytics';
 import {getAnalyticsSource} from '../../api/analytics/analytics-selectors';
-import {getPageID} from '../../api/analytics/search-analytics';
 import {AnalyticsParam} from '../../api/search/search-api-params';
 import {NavigatorContext} from '../../app/navigatorContextProvider';
 import {AnalyticsState} from './configuration-state';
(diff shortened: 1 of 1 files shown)


alexprudhomme (approved):

lvu285:
@louis-bompart if we remove this one, how it will affect InsightPanel & IPX, as I see the analytics from those products send pageId?

louis-bompart:
> @louis-bompart if we remove this one, how it will affect InsightPanel & IPX, as I see the analytics from those products send pageId?

We have 2 central place where we build the common analytics section for every and any search calls.
Here, where I did the change, and the legacy counterpart, for coveoua.

So, tldr, the change should propagate everywhere :tm:

louis-bompart:
Manually tested:
 - Open ipx.html on localhost.
 - Click a result
 - Do a search
 - Check the analytics part of the search payload
 - Without changes: pageId is in there. With my changes, pageId isn't here.

louis-bompart:
There's concern that it might impact IPX.
As far as we know, IPX is not used/ready to be used with the Event Protocol.

If the pageId is required for a use case, the lob owner/developer should first ensure the schema of existing events (or new events) fit their needs, then trickle it down to the implementation.

From the headless/atomic perspective, The Schema is the sole truth.

fbeaudoincoveo (approved):

dguerinCoveo (approved):
We don't specifically use `pageId` anywhere in the code, so no problem for us there. Though I am wondering if this will affect the analytics reports of customer using the old analytics. Is it possible that they have some reporting based on the `pageId`, which will no longer be accessible?

PR #4178, opened 2024-07-11
fix(atomic/commerce): fix breadbox labels / values

fbeaudoincoveo:
https://coveord.atlassian.net/browse/KIT-3383

### DIFF
--- a/packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
+++ b/packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
@@ -276,33 +276,10 @@ export class AtomicCommerceBreadbox
   };
 
   private buildBreadcrumb(breadcrumb: Breadcrumb<AnyFacetValue>) {
-    let facetStateName:
-      | 'facets'
-      | 'categoryFacets'
-      | 'numericFacets'
-      | 'dateFacets';
-
-    switch (breadcrumb.type) {
-      case 'hierarchical':
-        facetStateName = 'categoryFacets';
-        break;
-      case 'numericalRange':
-        facetStateName = 'numericFacets';
-        break;
-      case 'dateRange':
-        facetStateName = 'dateFacets';
-        break;
-      default:
-        facetStateName = 'facets';
-    }
-
     return breadcrumb.values.map((value: BreadcrumbValue<AnyFacetValue>) => {
       return {
         facetId: breadcrumb.facetId,
-        label:
-          this.bindings.store.state[facetStateName][
-            breadcrumb.facetId
-          ]?.label(),
+        label: breadcrumb.facetDisplayName,
         deselect: value.deselect,
         formattedValue: this.valueForFacetType(
           breadcrumb.type,
--- a/packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
+++ b/packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
@@ -17,6 +17,7 @@ import {
 } from '@coveo/headless/commerce';
 import {Component, h, State, Element, Prop} from '@stencil/core';
 import {FocusTargetController} from '../../../utils/accessibility-utils';
+import {parseDate} from '../../../utils/date-utils';
 import {getFieldValueCaption} from '../../../utils/field-utils';
 import {
   InitializableComponent,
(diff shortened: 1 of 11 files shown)


packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
```diff
@@ -249,15 +252,25 @@ export class AtomicCommerceBreadbox
     switch (type) {
       case 'numericalRange':
         return [
-          this.bindings.store.state.numericFacets[field].format(
```
fbeaudoincoveo, line 252: This approach works when the facets are already initialized, but it doesn't work when we're restore the state from the URL.

olamothe: Could you explain why that is ?
And so would this point to some issue in `registerFacet()` function for all types of commerce-facets ? 

If it's not a reliable property that should not be used because of a good reason, then we should ensure no other code relies on it currently, and that no other will be written in the future to rely on it.

This would mean making sure there is no `registerFacet` function (or `facets` property) available on the store specifically for `atomic-commerce-store`

olamothe (approved):

packages/atomic/src/components/commerce/facets/atomic-commerce-category-facet/atomic-commerce-category-facet.tsx
```diff
@@ -124,14 +123,19 @@ export class AtomicCategoryFacet implements InitializableComponent<Bindings> {
   private showMoreFocus?: FocusTargetController;
   private headerFocus?: FocusTargetController;
   private activeValueFocus?: FocusTargetController;
+  private unsubscribeFacetController?: () => void;
 
   @AriaLiveRegion('facet-search')
   protected facetSearchAriaMessage!: string;
 
   public initialize() {
-    if (!this.facetState) {
```
fbeaudoincoveo, line 132: This condition was never satisfied

packages/atomic/src/components/commerce/atomic-commerce-breadbox/atomic-commerce-breadbox.tsx
fbeaudoincoveo: > Could you explain why that is ?

Relying on the registered facet in the Stencil store means implies an order of operations: we assume that the `initialize` method of the facet controller has been executed before the breadcrumbs have been rendered.

However, if I add console logs in the `registerFacet` method of the `AtomicCommerceFacet` component, and in the `render` function of the `AtomicCommerceBreadbox` component, we can see that this is not the case:

logs

I believe this is caused by the fact that the commerce facet components are generated rather than being directly included in the markup configuration.

I notice that we're doing something similar to what I'm suggesting here for automatic facet breadcrumbs (and only for those) in `atomic-breadbox.tsx` (see https://github.com/coveo/ui-kit/blob/master/packages/atomic/src/components/search/atomic-breadbox/atomic-breadbox.tsx#L286). This seems to confirm what I'm saying.

> And so would this point to some issue in registerFacet() function for all types of commerce-facets ?

If what I stated above is accurate, yes, it means we should not rely on facets registered in the Stencil store at all in the commerce use case. I'll do some more in-depth testing before eliminating this completely, though.

alexprudhomme: From my understanding the store is used for the refine-modal and the breadcrumbs. The solution for the refine-modal was to create a second FacetGenerator inside of the atomic-refine-modal instead of cloning using the atomic store. For the breadcrumbs/searchParameter, the solution was to add a bunch of custom funky code for all cases to match (https://github.com/coveo/ui-kit/blob/095ab4b7ca97b21dc01fc488b9f91ee12f9705f1/packages/headless/src/features/facets/automatic-facet-set/automatic-facet-set-slice.ts#L62). Not sure what the best thing to do for commerce is. I think I did all that mostly because there was no client side "label" for automatic facets.

alexprudhomme: For me there was a lot of problem with forward/back, well have to test the same thing on commerce facets.

https://github.com/coveo/ui-kit/pull/3062

fbeaudoincoveo: The label is not an issue for commerce facets, since the labels are pre-configured in the CMH. We include the facet label is included in the breadcrumb state, so we definitely don't need the store for that.

We also don't use the store in the atomic-commerce-refine-modal: we essentially append an `atomic-commerce-facets` to the host when the refine modal is opened, which renders the facets.

I believe we can safely get rid of this in Atomic commerce.

fbeaudoincoveo: > For me there was a lot of problem with forward/back, well have to test the same thing on commerce facets.

Tested locally after removing the store registration and it's still working seamlessly

PR #4180, opened 2024-07-12
ci(atomic): split playwright tests into shards

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-3396

simply split into shards instead of browser + shards. Shards only is simpler. We like simpler.

~60 min playwright report -> ~6 min


.github/actions/merge-playwright-reports/action.yml
```diff
@@ -0,0 +1,21 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Download blob reports from GitHub Actions Artifacts
+      uses: actions/download-artifact@v4
+      with:
+        path: packages/atomic/all-blob-reports
+        pattern: blob-report-*
+        merge-multiple: true
+    - name: Merge into JSON Report
```
fpbrault, line 12: So we will a single merged report instead of one per shard right? Nice 🚀

fpbrault (approved):
LGTM

PR #4183, opened 2024-07-15
fix(headless): separate manual numeric facet in a separate slice for commerce

olamothe:
This PR should fix various issues that were reported by QA surrounding "manual numeric facets"; ie: the input in Atomic that allows end user to select an arbitrary range. The issues were linked to the fact that we tied together "the ranges returned by the API/computed by the index" with those entered "manually" by the end user.

This PR creates a new slice ( `manual-numeric-facet-slice` ) and record those entry in the URL as `mnf` (for manual numeric facet).

It also removes the `range` stateful property in Atomic, to instead rely on a stateful value provided by  the Headless controller itself.

KIT-3240


packages/headless/src/features/commerce/facets/numeric-facet/numeric-facet-actions.ts
```diff
@@ -57,3 +62,15 @@ export const updateNumericFacetValues = createAction(
...
+
+export const updateManualNumericFacetRange = createAction(
+  'commerce/facets/numericFacet/updateManualRange',
+  (payload: UpdateManualNumericFacetRangePayload) => {
+    try {
+      validateManualNumericRanges({currentValues: [payload]});
+      return {payload, error: null};
+    } catch (error) {
+      return {payload, error: serializeSchemaValidationError(error as Error)};
+    }
```
Spuffynism, line 74: Why do we need this try/catch logic?

Spuffynism (approved):

packages/headless/src/features/commerce/facets/numeric-facet/numeric-facet-actions.ts
olamothe: It's a copy of the above function; But it's un-necessary in this case. I'll simplify.

packages/headless/src/features/commerce/parameters/parameters-actions.ts
```diff
@@ -22,6 +22,10 @@ export interface Parameters {
    * A record of the numeric facets, where the key is the facet id, and value is an array containing the numeric ranges to request.
    */
   nf?: Record<string, NumericRangeRequest[]>;
+  /**
+   * A record of manual numeric facets, where the key is the facet id, and value is an array containing the numeric ranges to request.
```
fbeaudoincoveo, line 26:
```suggestion
   * A record of manual numeric facets, where the key is the facet ID, and the value is an array containing the numeric ranges to request.
```

packages/headless/src/features/search-parameters/search-parameter-actions.ts
```diff
@@ -57,6 +57,10 @@ export interface SearchParameters {
    * A record of the numeric facets, where the key is the facet id, and value is an array containing the numeric ranges to request.
    */
   nf?: Record<string, NumericRangeRequest[]>;
+  /**
+   * A record of the manual numeric facets, where the key is the facet id, and value is the selected numeric range.
```
fbeaudoincoveo, line 61:
```suggestion
   * A record of the manual numeric facets, where the key is the facet ID, and the value is the selected numeric range.
```

fbeaudoincoveo (approved):

packages/headless/src/controllers/commerce/core/facets/numeric/headless-commerce-numeric-facet.ts
```diff
@@ -89,13 +96,22 @@ export function buildCommerceNumericFacet(
...
 
   return {
     ...coreController,
 
+    setManualRange(range: NumericRangeRequest) {
+      dispatch(updateManualNumericFacetRange({facetId, ...range}));
+      dispatch(fetchProductsActionCreator());
+    },
+
     setRanges(ranges: NumericRangeRequest[]) {
```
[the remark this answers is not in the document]
olamothe: I could also rewrite `setRanges` and change its behavior. 
Probably simpler.

packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -104,8 +110,32 @@ function getFacets(state: ListingAndSearchStateNeededByQueryCommerceAPI) {
...
+) {
+  if (!state.manualNumericFacetSet) {
+    return [];
+  }
+
+  return Object.entries(state.manualNumericFacetSet!)
+    .filter(
+      ([_, manualNumericFacet]) => manualNumericFacet.manualRange !== undefined
+    )
+    .map(([facetId, manualNumericFacet]) => ({
```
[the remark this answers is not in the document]
olamothe: Don't really see a need to: It gets added to the request like any other `AnyFacetRequest`, which is already a defined type and correctly handled/protected by TS compiler.

What I can do is manually specify the return type of the function if that makes it clearer.

packages/headless/src/controllers/commerce/search/headless-search.ts
```diff
@@ -145,6 +146,7 @@ function loadBaseSearchReducers(
     configuration,
     commerceSearch,
     commerceQuery,
+    manualNumericFacetSet,
```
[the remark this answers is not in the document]
olamothe: Good catch; Not needed; Was an artefact forgot to cleanup

PR #4186, opened 2024-07-15
fix(atomic): breaking change in `SearchBoxSuggestionsEvent`

alexprudhomme:
https://coveord.atlassian.net/browse/KIT-3388


fbeaudoincoveo (approved):
As discussed in person, we should make the search box controller type mandatory V3. Can you create a Jira and add it to the V3 epic?

PR #4189, opened 2024-07-16
fix: use browseResults for results-per-page

louis-bompart:
https://coveord.atlassian.net/browse/KIT-3387

### DIFF
--- a/packages/headless/src/controllers/results-per-page/headless-results-per-page.ts
+++ b/packages/headless/src/controllers/results-per-page/headless-results-per-page.ts
@@ -1,6 +1,9 @@
 import {configuration} from '../../app/common-reducers';
 import {SearchEngine} from '../../app/search-engine/search-engine';
-import {logPagerResize} from '../../features/pagination/pagination-analytics-actions';
+import {
+  browseResults,
+  logPagerResize,
+} from '../../features/pagination/pagination-analytics-actions';
 import {paginationReducer as pagination} from '../../features/pagination/pagination-slice';
 import {fetchPage} from '../../features/search/search-actions';
 import {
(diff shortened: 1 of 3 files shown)


packages/headless/src/controllers/results-per-page/headless-results-per-page.ts
```diff
@@ -52,7 +55,7 @@ export function buildResultsPerPage(
 
     set(num: number) {
       coreController.set(num);
-      dispatch(fetchPage({legacy: logPagerResize()}));
+      dispatch(fetchPage({legacy: logPagerResize(), next: browseResults()}));
```
fbeaudoincoveo, line 58: I think we should adjust the tests accordingly

fbeaudoincoveo (approved):

packages/headless/src/controllers/results-per-page/headless-results-per-page.ts
louis-bompart: All tests are green tho 🙃 ?

fbeaudoincoveo: Yes, but the fact that you can change the `fetchPage` action payload and the tests are still green seem to indicate that the test is incomplete

louis-bompart: Fair, I added a basic ut to cover the change in parameters at the ut level.
More advanced/comprehensive testing might be wise, but to be done properly should probably be done either from an e2e (true e2e, that includes the backend systems) perspective, or thru contract testing.
The former would be sluggish and not really setup, the second one would probably be the most adequate, but contract testing is not something done at Coveo yet and span across multiple teams, so out-of-scope.

PR #4195, opened 2024-07-17
feat(headless): add tab manager controller

fpbrault:
This PR adds the tab manager controller, which will be used for the atomic-tab-manager component.

https://coveord.atlassian.net/browse/CDX-1582

### DIFF
--- a/packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
+++ b/packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
@@ -1,5 +1,4 @@
 import {configuration} from '../../../app/common-reducers';
-import {registerTab} from '../../../features/tab-set/tab-set-actions';
 import {tabSetReducer as tabSet} from '../../../features/tab-set/tab-set-slice';
 import {
   buildMockSearchEngine,
@@ -26,14 +25,6 @@ describe('Core Tab Manager', () => {
     expect(engine.addReducers).toHaveBeenCalledWith({
       configuration,
       tabSet,
-    });
-  });
-
-  describe('initialization', () => {
-    it('dispatches #registerTabManager', () => {
-      initTabManager();
-
-      expect(registerTab).toHaveBeenCalledWith();
     });
   });
 });
--- a/packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
+++ b/packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
@@ -1,4 +1,3 @@
-import {configuration} from '../../../app/common-reducers';
 import {tabSetReducer as tabSet} from '../../../features/tab-set/tab-set-slice';
 import {
   buildMockSearchEngine,
@@ -21,9 +20,25 @@ describe('Core Tab Manager', () => {
     initTabManager();
   });
 
+  it('returns the current active tab', () => {
+    expect(engine.state.tabSet).toEqual({});
+
+    const tabId = '1';
+    const expectedTabSet = {
+      [tabId]: {
+        id: tabId,
+        expression: 'test',
+        isActive: true,
+      },
+    };
+
+    engine.state.tabSet = expectedTabSet;
+
+    expect(engine.state.tabSet).toEqual(expectedTabSet);
+  });
+
   it('adds the correct reducers to engine', () => {
     expect(engine.addReducers).toHaveBeenCalledWith({
-      configuration,
       tabSet,
     });
   });
--- a/packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
+++ b/packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
@@ -1,0 +1,39 @@
+import {configuration} from '../../../app/common-reducers';
+import {registerTab} from '../../../features/tab-set/tab-set-actions';
+import {tabSetReducer as tabSet} from '../../../features/tab-set/tab-set-slice';
+import {
+  buildMockSearchEngine,
+  MockedSearchEngine,
+} from '../../../test/mock-engine-v2';
+import {createMockState} from '../../../test/mock-state';
+import {buildCoreTabManager} from './headless-core-tab-manager';
+
+jest.mock('../../../features/tab-set/tab-set-actions');
+
+describe('Core Tab Manager', () => {
+  let engine: MockedSearchEngine;
+
+  function initTabManager() {
+    buildCoreTabManager(engine);
+  }
+
+  beforeEach(() => {
+    engine = buildMockSearchEngine(createMockState());
+    initTabManager();
+  });
+
+  it('adds the correct reducers to engine', () => {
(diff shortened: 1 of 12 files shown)


alexprudhomme (approved):
lgtm.

packages/headless/src/controllers/core/tab/headless-core-tab.ts
```diff
@@ -126,6 +135,11 @@ export function buildCoreTab(engine: CoreEngine, props: TabProps): Tab {
     ...controller,
 
     select() {
+      if (props.options.clearFiltersOnTabChange) {
+        dispatch(deselectAllBreadcrumbs());
+        dispatch(deselectAllNonBreadcrumbs());
+        dispatch(updateQuery({q: ''}));
+      }
```
louis-bompart, line 142: Wondering if we should refactor and use `prepareForSearchWithQuery`
https://github.com/coveo/ui-kit/blob/2698979fa47de81882269389e505b023696ebe0c/packages/headless/src/features/search/search-actions.ts#L76-L77

louis-bompart (approved):
LGTM overall, nitpik maybe refactor and use/create an action for this 'prep up query without filters'

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns A `TabManager` controller instance.
+ */
+export function buildCoreTabManager(engine: CoreEngine): TabManager {
+  if (!loadTabReducers(engine)) {
+    throw loadReducerError;
+  }
+
+  const controller = buildController(engine);
+
+  const getCurrentTab = () => {
```
fbeaudoincoveo, line 38: Not super easy to read, and could be replaced by a redux selector to benefit from memoization! 😄 

```typescript
  const currentTab = createSelector(
    (state: typeof engine.state) => state.tabSet,
    (state) => {
      const activeTab = Object.values(state).find((tab) => tab.isActive);
      return activeTab?.id ?? '';
    }
  );
```

I don't think you need to fallback to `{}` if `engine.state.tabSet` is undefined, because adding the `tabSet` reducer to the engine (which you do in `loadTabReducers` implies that `tabSet` is going to be present in the engine's state, and that it's going to be set to the initial state (i.e., `{}`) if it's not already present.

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.ts
```diff
@@ -0,0 +1,60 @@
...
+        ?.id ?? ''
+    );
+  };
+
+  return {
+    ...controller,
+
+    get state() {
+      const activeTab = getCurrentTab();
+      return {activeTab};
```
fbeaudoincoveo, line 50: If you appy my above comment, then here you'd do:

```suggestion
      return {activeTab: currentTab(engine.state)};
```

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.ts
```diff
@@ -0,0 +1,60 @@
...
+    },
+  };
+}
+
+function loadTabReducers(
+  engine: CoreEngine
+): engine is CoreEngine<ConfigurationSection & TabSection> {
```
fbeaudoincoveo, line 57:
```suggestion
): engine is CoreEngine<TabSection> {
```

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.ts
```diff
@@ -0,0 +1,60 @@
...
+  engine.addReducers({configuration, tabSet});
```
fbeaudoincoveo, line 58:
```suggestion
  engine.addReducers({tabSet});
```

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.ts
fbeaudoincoveo: From what I can tell, you're not explicitly using the configuration section in the controller or in the slice. Also, the configuration section is included by default on the CoreEngine state, so this is superfluous.

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  beforeEach(() => {
+    engine = buildMockSearchEngine(createMockState());
+    initTabManager();
+  });
+
+  it('adds the correct reducers to engine', () => {
+    expect(engine.addReducers).toHaveBeenCalledWith({
+      configuration,
```
fbeaudoincoveo, line 26:
```suggestion
```

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+jest.mock('../../../features/tab-set/tab-set-actions');
+
+describe('Core Tab Manager', () => {
+  let engine: MockedSearchEngine;
+
+  function initTabManager() {
+    buildCoreTabManager(engine);
+  }
+
```
fbeaudoincoveo, line 18: You should also test that the state returns the currently active tab.

packages/headless/src/controllers/core/tab/headless-core-tab.ts
```diff
@@ -1,7 +1,12 @@
 import {BooleanValue, Schema} from '@coveo/bueno';
```
fbeaudoincoveo, line 1: A bit out of scope, maybe, but there's an opportunity here to refactor the "isActive" part of the state into a Redux selector as I suggested for the tab manager's `currentTab`.

packages/headless/src/controllers/tab-manager/headless-tab-manager.ts
```diff
@@ -0,0 +1,20 @@
+import {SearchEngine} from '../../app/search-engine/search-engine';
+import {
+  TabManager,
+  TabManagerState,
+  buildCoreTabManager,
+} from '../core/tab-manager/headless-core-tab-manager';
+
+export type {TabManager, TabManagerState};
+
+export function buildTabManager(engine: SearchEngine): TabManager {
```
fbeaudoincoveo, line 10: Should be documented

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.ts
```diff
@@ -0,0 +1,60 @@
...
+  activeTab: string;
+}
+
+/**
+ * Creates a `TabManager` controller instance.
+ *
+ * @param engine - The headless engine.
+ * @returns A `TabManager` controller instance.
+ */
+export function buildCoreTabManager(engine: CoreEngine): TabManager {
```
fbeaudoincoveo, line 31: Doesn't really need to be documented, and if it is it should be marked as @internal

fbeaudoincoveo (approved):

PR #4196, opened 2024-07-17
feat(atomic): add atomic-tab-manager component

fpbrault:
This PR adds the `atomic-tab-manager` and `atomic-tab` components. It only includes the new components without adding support for controlling visibility of other components, as this will be added in another PR.

image
image
image

https://coveord.atlassian.net/browse/CDX-1580


packages/atomic/src/components/common/tab-manager/tab-button.tsx
```diff
@@ -0,0 +1,24 @@
+import {FunctionalComponent, h} from '@stencil/core';
```
olamothe, line 1: I would not advocate to change it for this PR; However food for thoughts for everyone:

The approach of splitting functional components is mandatory when there's re-use for multiple use case (ie: we need a similar component for commerce/search/insight). Otherwise we risk introducing inconsistencies/bugs for our users.

Here however, there is currently no re-use.

There * might be * in the future, but as of today it's not the case. 
So, the code is slightly more complex (split across multiple files) for no tangible gain "today".
But, * if * we need it in the future, the work is already done, so that's nice.

So, just something to keep in mind, wonder what other think !

packages/atomic/src/components/common/tab-manager/tab-button.pcss
```diff
@@ -0,0 +1,21 @@
+:host {
+  border-radius: 60px;
```
olamothe, line 2: Most of those could be set with tailwind classes, no ? 
Or am I mistaken ?

Also those arbitrary variable (height/radius) should be controlled by --atomic variables

olamothe (approved):
Much simpler with the manager !

packages/atomic/src/components/common/tab-manager/tab-button.pcss
fpbrault: Yep, It now uses tailwind classes. Also instead setting atomic variables for the tab bar thickness/position, I use normal tailwind classes, and have added a css part to the button container so that it is easy to style the bar.

packages/headless/src/controllers/core/tab-manager/headless-core-tab-manager.ts
```diff
@@ -35,26 +32,24 @@ export function buildCoreTabManager(engine: CoreEngine): TabManager {
...
-      Object.values(engine.state.tabSet ?? {}).find((tab) => tab.isActive)
-        ?.id ?? ''
-    );
-  };
+  const currentTab = createSelector(
+    (state: typeof engine.state) => state.tabSet,
+    (state) => {
+      const activeTab = Object.values(state).find((tab) => tab.isActive);
+      return activeTab?.id ?? '';
+    }
```
louis-bompart:
```suggestion
    (tabSet) => 
      Object.values(tabSet).find((tab) => tab.isActive)?.id ?? ''
```

packages/atomic/src/components/common/tab-manager/tab-button.tsx
louis-bompart: I like what's done here.
Notwithstanding the reusability of the components, I think it helped a great lot to structure the code:
You have the manager that encapsulate the logic and show what it'll do in 'broad lines' without getting too much in the nitty gritty of the CSS and whatnot.
Then, we have the functional components that are dumb, and display stuff nice and proper.

I think it's nicely done!

The only point I have mixed feelings about (but nothing I'd change), is the location of the code itself. /common/ when there are no other uses than /search/ might be an overreach, especially given we can refactor them whenever we need

louis-bompart (approved):
LGTM, nice job!

I'm guessing the storybook/playwright are scheduled in future PRs.

alexprudhomme (approved):

PR #4198, opened 2024-07-17
feat(headless SSR): define the SSR Commerce Engine

y-lakhdar:
In this PR:
* Defining a new `SSRCommerceEngine` interface supporting two solution types (listing, search)
* Implementing a `defineCommerceEngine()` method for commerce engine configuration (to stay consistent with the non-commerce SSR package)

> Note:
> Only supporting listing page in this PR.
> The support for search page will be tackled in KIT-3394

https://coveord.atlassian.net/browse/KIT-3390

### DIFF
--- a/packages/headless/src/app/commerce-engine/commerce-engine.ssr.ts
+++ b/packages/headless/src/app/commerce-engine/commerce-engine.ssr.ts
@@ -86,6 +86,7 @@ export interface CommerceEngineDefinition<
   > {}
 
 /**
+ * @internal
  * Initializes a Commerce engine definition in SSR with given controllers definitions and commerce engine config.
  * @param options - The commerce engine definition
  * @returns Three utility functions to fetch the initial state of the engine in SSR, hydrate the state in CSR,


packages/headless/src/app/commerce-engine/commerce-engine.ssr.ts
```diff
@@ -0,0 +1,221 @@
...
+  buildCommerceEngine,
+} from './commerce-engine';
+
+export interface SSRCommerceEngineOptions extends CommerceEngineOptions {
+  /**
+   * The type of the solution the engine will be used for
+   * - 'search': Indicates that the engine is used for Search.
+   * - 'product-listing': Indicates that the engine is used for Product listing.
+   */
+  solutionType: 'search' | 'listing';
```
Spuffynism, line 36: Why do we need this? The commerce engine should work with listings, search and recs (all at the same time). Is this due to a limitation with SSR?

packages/headless/src/app/commerce-engine/commerce-engine.ssr.ts
```diff
@@ -0,0 +1,221 @@
...
+}
+
+/**
+ * The SSR commerce engine.
+ */
+export interface SSRCommerceEngine extends CommerceEngine {
+  /**
+   * Waits for the search to be completed and returns a promise that resolves to a `SearchCompletedAction`.
+   */
+  waitForSearchCompletedAction(): Promise<SearchCompletedAction>;
```
Spuffynism, line 46: `SearchCompletedAction` doesn't seem to be the right type, as it seems to be a type that works for non-commerce responses

Spuffynism, line 46: To better reflect that this works with listings, where we don't execute searches, I would propose we rename this
```suggestion
  waitForRequestCompletedAction(): Promise<SearchCompletedAction>;
```

fbeaudoincoveo (approved):

packages/headless/src/app/commerce-engine/commerce-engine.ssr.ts
```diff
@@ -0,0 +1,201 @@
...
+export interface CommerceEngineDefinition<
+  TControllers extends ControllerDefinitionsMap<SSRCommerceEngine, Controller>,
+> extends EngineDefinition<
+    SSRCommerceEngine,
+    TControllers,
+    CommerceEngineOptions
+  > {}
+
+/**
+ * Initializes a Commerce engine definition in SSR with given controllers definitions and commerce engine config.
```
fbeaudoincoveo, line 89: Maybe we should clearly mark this as internal / not ready for use for now.

alexprudhomme (approved):

PR #4202, opened 2024-07-18
test(quantic): e2e tests adapted to new rga feedback modal

mmitiche:
## [SFINT-5608](https://coveord.atlassian.net/browse/SFINT-5608)

- Adapted tests to the latest analytics schema when submitting rga feedback.
- Adapted the e2e to the new rga feedback modal by supporting the new questions and testing the validity of the feedback form.
-  These tests are still skipped because of the flakiness of the lightning modal but now, if we run them, they pass not like before where they were only expected to work with the previous feedback modal.

https://github.com/user-attachments/assets/c6c3f101-49e5-40ed-9978-244d15c97c87

[SFINT-5608]: https://coveord.atlassian.net/browse/SFINT-5608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -772,107 +832,25 @@ describe('quantic-generated-answer', () => {
...
-                          secondStreamId,
-                          {
-                            reason: otherOption,
-                            details: exampleDetails,
-                          }
-                        );
-                      }
 
+                      Actions.clickFeedbackSubmitButton();
                       Actions.clickFeedbackDoneButton();
```
SimonMilord, line 854: are you not expecting something here? why do all these actions to expect nothing?

SimonMilord (approved):
just a question, but rest looks good to me!

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
mmitiche: No cause mainly these actions are to prepare for the next step of the test and the expectations that need to be checked after submitting a feedback are already check in another test so no need to re add them here.

PR #4203, opened 2024-07-18
test(atomic): add tests for product link

olamothe:
https://coveord.atlassian.net/browse/KIT-3269


packages/atomic/src/components/commerce/product-template-components/atomic-product-link/atomic-product-link.new.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+export default meta;
+
+export const Default: Story = {
+  name: 'atomic-product-link',
+};
+
+export const WithSlotsAttributes: Story = {
+  name: 'opens in a new browser tab',
+  decorators: [
+    (_) => {
```
fpbrault, line 28: I am wondering why the _ instead of nothing there?

packages/atomic/src/utils/object-utils.tsx
```diff
@@ -0,0 +1,17 @@
...
+  for (const k of keys) {
+    if (current && typeof current === 'object' && k in current) {
+      current = (current as Record<string, unknown>)[k];
+    } else {
+      return undefined;
+    }
+  }
+
+  return typeof current === 'string' ? current : undefined;
+}
```
alexprudhomme, line 17: It was actually bugged in commerce because the readFromObject function was different than the one in search. I made a common function for the two of them here with the one from search.

PR #4207, opened 2024-07-19
fix(ep): do not send actionCause w/ breadcrumbResetAll

louis-bompart:
https://coveord.atlassian.net/browse/KIT-3403


fbeaudoincoveo (approved):

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
```diff
@@ -160,15 +160,20 @@ describe('quantic-generated-answer', () => {
...
-          it('should perform a search query with the default rephrase button', () => {
-            cy.wait(InterceptAliases.Search);
-            Expect.searchQueryContainsCorrectRephraseOption(
-              defaultRephraseOption,
-              param.useCase === 'search' ? 'interfaceLoad' : 'searchboxSubmit',
-              ['text/markdown', 'text/plain']
-            );
-          });
+          // SFINT-5652
+          (analyticsMode === 'legacy' ? it : it.skip)(
```
mmitiche, line 164: Why is this test skipped while it's not related to the main changes in this PR where we stopped sending the analytics action cause for `breadcrumbResetAll`, `FacetShowLess` and `FacetShowMore`?

mmitiche (approved):

packages/quantic/cypress/e2e/default-2/generatedAnswer/generated-answer.cypress.ts
louis-bompart: As we approach GA for EP, safety guards for unmigrated events are now removed.
Meaning that, when using Event Protocol:
• If an event is not migrated, it is not sent
• If an actionCause for a search is not migrated, no actionCause is sent
During the removal of the safety rails, the following test has failed when using the Event Protocol, so I skipped it so we can decide later on what to do with it

PR #4208, opened 2024-07-19
fix(ep): remove rga non-schematized search cause

louis-bompart:
https://coveord.atlassian.net/browse/KIT-3404


alexprudhomme (approved):
What happens whenever a 'next' event is not specified ?

louis-bompart:
> What happens whenever a 'next' event is not specified ?

Good question: no actionCause will be sent.
This is the conclusion we reached with the DataPlatform team

mmitiche (approved):
👍

packages/headless/src/features/generated-answer/generated-answer-analytics-actions.ts
```diff
@@ -408,10 +408,6 @@ export const retryGeneratedAnswer = (): SearchAction => ({
   actionCause: SearchPageEvents.retryGeneratedAnswer,
 });
 
-export const rephraseGeneratedAnswer = (): SearchAction => ({
-  actionCause: SearchPageEvents.rephraseGeneratedAnswer,
-});
-
```
SimonMilord, line 414: do we not want to remove the retry as well here?

louis-bompart:
Following the slack discussion, @mmitiche @SimonMilord , I rerequest your review.
I think this PR is ready to be scrubbed maybe 🤔

mmitiche:
@louis-bompart  Following the same slack discussion: https://coveo.slack.com/archives/C02ET4TQDD4/p1721921559026459?thread_ts=1721918413.953499&cid=C02ET4TQDD4, ServiceData/DataPlatform gang agreed to add the `retryGeneratedAnswer` and `rephraseGeneratedAnswer` to the schemas. 

In that case wouldn't better to avoid removing these two causes now just to re-add them in like one week when they will be added to the accepted schemas?

louis-bompart:
> @louis-bompart Following the same slack discussion: https://coveo.slack.com/archives/C02ET4TQDD4/p1721921559026459?thread_ts=1721918413.953499&cid=C02ET4TQDD4, ServiceData/DataPlatform gang agreed to add the `retryGeneratedAnswer` and `rephraseGeneratedAnswer` to the schemas.
> 
> In that case wouldn't better to avoid removing these two causes now just to re-add them in like one week when they will be added to the accepted schemas?

yes.
Pending update, will reloop on the PR after the changes in the schema.

louis-bompart:
Trying to dust off this PR @mmitiche, so essentially I should revert the changes that touches RGA, right?

mmitiche:
> Trying to dust off this PR @mmitiche, so essentially I should revert the changes that touches RGA, right?

Exactly yeah @louis-bompart , the RGA related search causes have been added to the schemas 👍

louis-bompart:
Third read, pretty certain everything has been standardized.
Closing.

PR #4210, opened 2024-07-19
feat(atomic): add tab-manager support for facet components

fpbrault:
This PR allows facets to be disabled/enabled based on the currently active tab in the tab manager.

https://coveord.atlassian.net/browse/CDX-1555


packages/atomic/src/components/search/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -253,7 +297,7 @@ export class AtomicSegmentedFacet implements InitializableComponent {
         <div
           part="placeholder"
           aria-hidden
-          class="h-8 w-48 bg-neutral animate-pulse rounded"
+          class="w-48 h-8 rounded bg-neutral animate-pulse"
```
louis-bompart, line 300: ? :D

packages/atomic/src/components/search/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
```diff
@@ -265,7 +309,7 @@ export class AtomicSegmentedFacet implements InitializableComponent {
     return (
       <div
         part="segmented-container"
-        class="flex whitespace-nowrap h-10 items-center"
+        class="flex items-center h-10 whitespace-nowrap"
```
louis-bompart, line 312: Why the reorder

louis-bompart (commented):
What if I'm a goofball and I put the same tab in both included and excluded?

packages/atomic/src/components/search/facets/atomic-segmented-facet/atomic-segmented-facet.tsx
fpbrault: I happen to have [Headwind](https://marketplace.visualstudio.com/items?itemName=heybourn.headwind) installed! It sorts the tailwind classes in a consistent order. A better solution is to use the [tailwind prettier plugin](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier)

fpbrault:
> What if I'm a goofball and I put the same tab in both included and excluded?

Covered in this doc comment

```typescript
  /*
  * If you don't set this property, or set it to `'[]'`, the facet can be displayed on any tab. Otherwise, the facet can only be displayed on the specified tabs. In either case, the facet won't be displayed on any of the tabs specified in the `tabs-excluded` property (exclusion takes precedence).
   */
  @ArrayProp()
  @Prop({reflect: true, mutable: true})
  public tabsIncluded: string[] | string = '[]'
```

louis-bompart:
> > What if I'm a goofball and I put the same tab in both included and excluded?
> 
> Covered in this doc comment
> 
> ```ts
>   /*
>   * If you don't set this property, or set it to `'[]'`, the facet can be displayed on any tab. Otherwise, the facet can only be displayed on the specified tabs. In either case, the facet won't be displayed on any of the tabs specified in the `tabs-excluded` property (exclusion takes precedence).
>    */
>   @ArrayProp()
>   @Prop({reflect: true, mutable: true})
>   public tabsIncluded: string[] | string = '[]'
> ```

Should we consider this an error of implementation and throw a warning?

fpbrault:
> > > What if I'm a goofball and I put the same tab in both included and excluded?
> > 
> > 
> > Covered in this doc comment
> > ```ts
> >   /*
> >   * If you don't set this property, or set it to `'[]'`, the facet can be displayed on any tab. Otherwise, the facet can only be displayed on the specified tabs. In either case, the facet won't be displayed on any of the tabs specified in the `tabs-excluded` property (exclusion takes precedence).
> >    */
> >   @ArrayProp()
> >   @Prop({reflect: true, mutable: true})
> >   public tabsIncluded: string[] | string = '[]'
> > ```
> 
> Should we consider this an error of implementation and throw a warning?

I think a warning could be nice, yes!

fbeaudoincoveo:
> What if I'm a goofball and I put the same tab in both included and excluded?

Exclusion wins!

packages/atomic/src/components/search/facets/atomic-category-facet/atomic-category-facet.tsx
```diff
@@ -224,7 +258,13 @@ export class AtomicCategoryFacet implements InitializableComponent {
   protected facetSearchAriaMessage!: string;
 
   public initialize() {
+    if (this.tabsIncluded.length > 0 && this.tabsExcluded.length > 0) {
+      console.warn(
+        'Values for both "tabs-included" and "tabs-excluded" have been provided. This is could lead to unexpected behavior.'
```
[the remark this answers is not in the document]
fbeaudoincoveo: Also, maybe we should update the documentation a bit

Instead of saying "this property complements ...", we should say something like "To avoid unexpected behaviors, do not set both this property and ..."

fbeaudoincoveo (approved):

packages/atomic/src/components/search/facets/atomic-facet/atomic-facet.tsx
```diff
@@ -128,11 +136,38 @@ export class AtomicFacet implements InitializableComponent {
...
+   *
+   * Set this property as a stringified JSON array, e.g.,
+   * ```html
+   *  <atomic-facet tabs-included='["tabIDA", "tabIDB"]'></atomic-facet>
+   * ```
+   * If you don't set this property, or set it to `'[]'`, the facet can be displayed on any tab. Otherwise, the facet can only be displayed on the specified tabs. In either case, the facet won't be displayed on any of the tabs specified in the `tabs-excluded` property (exclusion takes precedence).
+   */
+  @ArrayProp()
+  @Prop({reflect: true, mutable: true})
+  public tabsIncluded: string[] | string = '[]';
```
alexprudhomme, line 150: What's the reasoning behind  _If you don't set this property, or set it to `'[]'`_ ? Why would they prefer to write `[]` instead of nothing ?

Why can't it just be an array of string with an empty default value ?
```
  @ArrayProp()
  @Prop({reflect: true, mutable: true})
  public tabsIncluded: string[] = [];
```

fpbrault: The idea was to allow providing just a tab name when a single tab is included/excluded.

packages/atomic/src/components/search/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
```diff
@@ -197,13 +231,20 @@ export class AtomicNumericFacet implements InitializableComponent {
   }
 
   public initialize() {
+    if (this.tabsIncluded.length > 0 && this.tabsExcluded.length > 0) {
```
louis-bompart, line 234: Mh should the warning be when tabsIncluded & tabsExcluded are both specified?
What I had in mind was more if there's an intersection between the two sets/arrays.

louis-bompart (commented):
getting there!

packages/atomic/src/components/search/facets/atomic-numeric-facet/atomic-numeric-facet.tsx
fpbrault: @louis-bompart I don't think there's any situation where it would make sense to both include and exclude, since only one tab can be active at a time...

packages/atomic/src/utils/tab-utils.ts
```diff
@@ -0,0 +1,24 @@
...
+    return true;
+  }
+
+  const isIncluded =
+    includeTabs.length === 0 || includeTabs.includes(activeTab);
+  const isNotExcluded =
+    excludeTabs.length === 0 || !excludeTabs.includes(activeTab);
+
+  return isIncluded && isNotExcluded;
+}
```
alexprudhomme, line 24: If this is only used in facet-tab-utils.ts, why not make it a private function there and write tests for `updateFacetVisibilityForActiveTab` instead ?

alexprudhomme (approved):

packages/atomic/src/utils/tab-utils.ts
fpbrault: @alexprudhomme it will be used elsewhere, just not in this PR

PR #4215, opened 2024-07-22
docs(headless): commerce controllers

jpmarceau:
Expose reference documentation for more commerce actions, and also various commerce reference documentation improvements.

https://coveord.atlassian.net/browse/KIT-3416


packages/headless/src/features/commerce/product-listing-parameters/product-listing-parameters-actions-loader.ts
```diff
@@ -30,11 +30,7 @@ export interface ProductListingParametersActionCreators {
  * Loads the commerce facet order, facet set, pagination, and sort reducers and returns the available product listing parameters action creators.
  *
  * @param engine - The headless commerce engine.
-<<<<<<< HEAD
- * @returns An object holding the action creators.
-=======
  * @returns An object holding the product listing parameters action creators.
->>>>>>> ce6fb7ee5e23f10840b4cc654544e7dde05d80d2
```
alexprudhomme, line 37: Bad merge

packages/headless/src/controllers/commerce/core/interactive-product/headless-core-interactive-product.ts
jpmarceau, line 1: Question for @fbeaudoincoveo or @Spuffynism here: This (sub?) controller is under `/core` and we sometimes referred to it elsewhere as a sub-controller, but it has no state nor subscribe function. Is that expected?

Spuffynism: Yeah I think that's ok since it doesn't handle any redux-backed state. There's nothing to subscribe to. And I don't think we have a concept of "controller with no state" in headless 🤷.

Spuffynism (approved):

packages/headless/src/commerce.index.ts
```diff
@@ -216,6 +216,8 @@ export type {
   RecentQueriesState,
 } from './controllers/recent-queries-list/headless-recent-queries-list.ts';
 
+export type {RegisterRecentQueriesCreatorPayload} from './features/recent-queries/recent-queries-actions';
```
fbeaudoincoveo, line 219: Instead of exporting this type directly in `commerce.index.ts`, we could export it in `recent-queries-actions-loader.ts`. Then since we're already doing

```
export * from './features/commerce/recent-queries/recent-queries-actions-loader';
```

on line 61, we'd be exporting the type at the same time!

packages/headless/src/commerce.index.ts
```diff
@@ -290,6 +292,8 @@ export type {
 } from './controllers/core/triggers/headless-core-query-trigger';
 export {buildQueryTrigger} from './controllers/commerce/triggers/headless-commerce-query-trigger';
 
+export type {updateIgnoreQueryTrigger} from './features/commerce/triggers/triggers-actions';
```
fbeaudoincoveo, line 295: Similar comment here

fbeaudoincoveo (approved):

PR #4216, opened 2024-07-23
test(atomic): add tests for atomic-tab-manager

fpbrault:
This PR adds tests for the atomic-tab-manager component. It includes tests for facet visibility based on the active tab. Tests for visibility of other components will be added in their respective PRs.

https://coveord.atlassian.net/browse/CDX-1579


alexprudhomme:
Looks like this PR needs updates from the base branch

packages/atomic/src/components/search/tabs/atomic-tab-manager/atomic-tab-manager.new.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <atomic-tab name="all" label="All" tab></atomic-tab>
+      <atomic-tab name="article" label="Articles"></atomic-tab>
+      <atomic-tab name="documentation" label="Documentation"></atomic-tab>
+    `,
+  },
+};
+
+export default meta;
+
+export const Default: Story = {
```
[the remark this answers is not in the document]
fpbrault: Yes thats a good point

packages/atomic/src/components/common/facets/facet-tabs/facet-tabs-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export function updateFacetVisibilityForActiveTab(
+  tabsIncluded: string[] | string,
+  tabsExcluded: string[] | string,
+  activeTab: string,
+  facet?: AnyFacetType
+): void {
+  if (
+    !facet ||
+    (tabsIncluded.length === 0 && tabsExcluded.length === 0) ||
```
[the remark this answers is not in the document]
fpbrault: Right now that will only happen if someone does something like `tabs-excluded='"[]"'` instead of `tabs-excluded="[]"` since all the components that use this function use @Arrayprop for the tabs-excluded/tabs-included properties. 

@fbeaudoincoveo should I just remove the possibility of providing a single string as a value for tabsExcluded/tabsIncluded? It would prevent this problem from happening at all..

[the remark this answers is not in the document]
fpbrault: Nevermind, @ArrayProp needs either an array or a stringified array, not a single string value. Found a better way to handle this:

```typescript
   updateFacetVisibilityForActiveTab(
      [...this.tabsIncluded],
      [...this.tabsExcluded],
      this.tabManagerState?.activeTab,
      this.facet
    );
```
and updateFacetVisibilityForActiveTab now requires a string array

packages/atomic/src/components/common/tab-manager/tab-button.tsx
```diff
@@ -9,18 +9,23 @@ export interface TabButtonProps {
...
 export const TabButton: FunctionalComponent<TabButtonProps> = (props) => {
   const activeTabClass = props.isActive
-    ? "relative after:content-[''] after:block after:w-full after:h-1 after:absolute after:-bottom-0.5 after:bg-primary after:rounded"
+    ? "active-tab relative after:content-[''] after:block after:w-full after:h-1 after:absolute after:-bottom-0.5 after:bg-primary after:rounded"
     : '';
   const activeTabTextClass = props.isActive ? '' : 'text-neutral-dark';
   return (
-    <div part="button-container" class={activeTabClass}>
+    <li
+      aria-current={props.isActive ? 'true' : false}
```
alexprudhomme, line 17: This was broken and will be required for the active tab selector.
```suggestion
      aria-current={props.isActive ? 'true' : 'false'}
```

packages/atomic/src/components/common/tab-manager/tab-button.tsx
```diff
@@ -9,18 +9,23 @@ export interface TabButtonProps {
 
 export const TabButton: FunctionalComponent<TabButtonProps> = (props) => {
   const activeTabClass = props.isActive
-    ? "relative after:content-[''] after:block after:w-full after:h-1 after:absolute after:-bottom-0.5 after:bg-primary after:rounded"
+    ? "active-tab relative after:content-[''] after:block after:w-full after:h-1 after:absolute after:-bottom-0.5 after:bg-primary after:rounded"
```
alexprudhomme, line 12: adding an empty class for tests is usually not the best solution. Maybe you could check the aria-current property instead !

packages/atomic/src/components/search/tabs/atomic-tab-manager/e2e/atomic-tab-manager.e2e.ts
```diff
@@ -0,0 +1,184 @@
...
+    ]);
+  });
+
+  test.describe('when clicking on tab button', () => {
+    test.beforeEach(async ({tabManager}) => {
+      await tabManager.tabButtons('Articles').click();
+    });
+
+    test('should change active tab', async ({tabManager}) => {
+      await expect(tabManager.tabButtons('Articles')).toHaveClass(/active-tab/);
```
alexprudhomme, line 40: nitpick 
```suggestion
      await expect(tabManager.activeTab).toHaveText('Articles');
```

alexprudhomme (commented):
tests are still failing

packages/atomic/src/components/common/tab-manager/tab-button.tsx
fpbrault: I don't think there's a locator for that... I'll look for a better solution!

fpbrault: Actually I think can just do `.locator('[aria-current="true"]')`

packages/atomic/src/components/search/tabs/atomic-tab-manager/e2e/atomic-tab-manager.e2e.ts
```diff
@@ -0,0 +1,182 @@
...
+  test('should not display tabs dropdown', async ({tabManager}) => {
+    await expect(tabManager.tabDropdown).not.toBeVisible();
+  });
+
+  test('should display tab buttons for each atomic-tab elements', async ({
+    tabManager,
+  }) => {
+    await expect(tabManager.tabDropdown).not.toBeVisible();
+
+    expect(await tabManager.tabButtons().allTextContents()).toEqual([
```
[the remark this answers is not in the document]
fpbrault: Yep, I think I forgot to change that one!

louis-bompart (approved):

PR #4217, opened 2024-07-23
fix(headless/commerce): send clientId only when analytics are enabled

jpmarceau:
https://coveord.atlassian.net/browse/KIT-3312

Legally, one can only send data persisted in the browser (like the clientId) when the user agrees to tracking (even for search queries, not just for UA).

Most cases across the ui-kit project were already covered. I think these are the only two spots left, but let me know 👍


packages/headless/src/features/commerce/common/actions.ts
```diff
@@ -65,7 +65,9 @@ export const buildBaseCommerceAPIRequest = (
     organizationId: state.configuration.organizationId,
     trackingId: state.configuration.analytics.trackingId,
     ...restOfContext,
-    clientId: navigatorContext.clientId,
+    clientId: state.configuration.analytics.enabled
```
fbeaudoincoveo, line 68: @Spuffynism does the commerce API now allow for the empty string to be passed as a clientId? I know this was making requests fail previously...

Spuffynism: Yeah this is fine since https://github.com/coveo-platform/commerce-service/pull/2069

Spuffynism (approved):
I was going to propose solving this in the navigator context when fetching the `clientId` from relay, but that could still lead to a `clientId` being sent to CAPI if a navigator context provides one even if analytics are disabled.
https://github.com/coveo/ui-kit/blob/master/packages/headless/src/app/engine.ts#L292-L300

LGTM!

packages/headless/src/features/commerce/common/actions.ts
samisayegh: Would it be cleaner/more consistent to use the pattern on line 72, not sending a key with an empty string? Same comment for the query suggest,

```
...(state.configuration.analytics.enabled ? {clientId: navigatorContext.clientId} : {})
```

packages/headless/src/features/commerce/query-suggest/query-suggest-actions.ts
```diff
@@ -129,7 +129,9 @@ export const buildQuerySuggestRequest = (
     trackingId: state.configuration.analytics.trackingId,
     query: state.querySet[id],
     ...restOfContext,
-    clientId: navigatorContext.clientId,
+    clientId: state.configuration.analytics.enabled
```
samisayegh, line 132: Is this change also covered by a UT?

samisayegh (approved):
Looks good ✅

packages/headless/src/features/commerce/common/actions.ts
jpmarceau: You're right, that would be cleaner @samisayegh  👍 
FYI @fbeaudoincoveo , @Spuffynism , that does mean that I'll make the `clientId` optional (see `commerce-api-params.ts`), let me know if that's an issue.
Update incoming :loading:

Spuffynism (approved):

PR #4219, opened 2024-07-24
fix(refine-modal): include facets from atomic-external

louis-bompart:
Consider all atomic-search-interface & atomic-external for the facet acquisitions of the refine-modal.

Rough algo
- If a facet is not in a section, put it in the 'orphan bucket'
- Sort facet by interface
- In a same interface, put vertical facet first, horizontal facet second
- Once all non-orphaned facets are sorted out, sort the orphan facets in DOM order, and add'em.

KIT-3428


packages/atomic/src/components/common/facets/facet-header/facet-header.tsx
```diff
@@ -34,7 +34,7 @@ export const FacetHeader: FunctionalComponent<FacetHeaderProps> = (props) => {
       style="text-transparent"
       part="label-button"
       class="flex font-bold justify-between w-full py-1 px-2 text-lg rounded-none"
-      title={props.isCollapsed ? expandFacet : collapseFacet}
+      ariaLabel={props.isCollapsed ? expandFacet : collapseFacet}
```
louis-bompart, line 37: ⚠️ Small but interesting change.

there's an order of things for an 'accessible name' (as defined by ARIA).
https://russmaxdesign.github.io/accessible-forms/accessible-name-button01.html

So, the title was not used by screen readers notably.

I elected to replace the ariaLabel by title so that screenreaders -and my tests- pick up the string

packages/atomic/src/components/search/atomic-refine-toggle/atomic-refine-toggle.new.stories.tsx
```diff
@@ -40,3 +40,56 @@ export const Default: Story = {
     `,
   ],
 };
+
+export const WithAtomicExternals: Story = {
+  name: 'atomic-refine-toggle',
+  decorators: [
+    (story) => html`
+      <atomic-search-interface id="foo" data-testid="root-interface">
```
louis-bompart, line 48: It's purposefully mixed up

fbeaudoincoveo: Maybe add a comment to ensure that no one accidentally boy-scouts it clean 😅

alexprudhomme (commented):
Looks good, I tested with the facet-manager and also with automatic facets. Two things : 

1-The facets in the facet manager seem to be in random order. 
```
   < atomic-layout-section section="facets">
          <atomic-facet field="author" label="facet1"></atomic-facet>
          <atomic-facet-manager>
            <atomic-facet field="author" label="facet10"></atomic-facet>
            <atomic-facet field="author" label="facet11"></atomic-facet>
            <atomic-facet field="author" label="facet12"></atomic-facet>
          </atomic-facet-manager>
        </atomic-layout-section>
```
image


2-The automatic facets are being put at the bottom.

```
<atomic-layout-section section="facets">
          <atomic-facet field="author" label="facet1"></atomic-facet>
          <atomic-automatic-facet-generator
            desired-count="3"
          ></atomic-automatic-facet-generator>
        </atomic-layout-section>
```

image

This sorting stuff is so annoying 😢

packages/atomic/src/components/common/facets/facet-header/facet-header.tsx
fpbrault: The only issue with adding both is that screen reader that support both may read the aria-label value and then the title value. For example I tried with voiceover utility and it does read both.

packages/atomic/src/components/search/atomic-external/atomic-external.tsx
```diff
@@ -30,19 +31,23 @@ export class AtomicExternal {
...
 
-  private get interface() {
-    const element = document.querySelector(this.selector);
-    if (!element) {
-      throw new Error(
-        `Cannot find interface element with selector "${this.selector}"`
-      );
+  @Prop({mutable: true}) boundInterface?: AtomicInterface;
+
+  get #interface() {
```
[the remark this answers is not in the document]
fpbrault: Makes it a private property https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

packages/atomic/src/components/search/atomic-refine-toggle/e2e/atomic-refine-toggle.e2e.ts
```diff
@@ -0,0 +1,33 @@
...
+        'facet1',
+        'facet2',
+        'facet3',
+        'facet4',
+        'facet5',
+        'facet6',
+        'facet7',
+        'facet8',
+        'facet9',
+      ]);
```
fpbrault, line 30: Shouldn't the facets be displayed in the same order as on the page, instead of in alphabetical order?

fpbrault (approved):

louis-bompart:
> Looks good, I tested with the facet-manager and also with automatic facets. Two things :
> 
> 1-The facets in the facet manager seem to be in random order.
> 
> ```
>    < atomic-layout-section section="facets">
>           <atomic-facet field="author" label="facet1"></atomic-facet>
>           <atomic-facet-manager>
>             <atomic-facet field="author" label="facet10"></atomic-facet>
>             <atomic-facet field="author" label="facet11"></atomic-facet>
>             <atomic-facet field="author" label="facet12"></atomic-facet>
>           </atomic-facet-manager>
>         </atomic-layout-section>
> ```
> 
> first load image second load image

We are having facets on all the same fields, so yeah, suboptimal for the facet manager 😆
Facet manager order facets like they want, then we follow for those

> 2-The automatic facets are being put at the bottom.
> 
> ```
> <atomic-layout-section section="facets">
>           <atomic-facet field="author" label="facet1"></atomic-facet>
>           <atomic-automatic-facet-generator
>             desired-count="3"
>           ></atomic-automatic-facet-generator>
>         </atomic-layout-section>
> ```
> 
> image

If you put the automatic facet after the static facet, yeah, that make sense 😅

> This sorting stuff is so annoying 😢

Complicated indeed, let's go with baby steps, I think it's better than what's currently in place :)

alexprudhomme (approved):

packages/atomic/src/components/common/facets/facet-header/facet-header.tsx
louis-bompart: > The only issue with adding both is that screen reader that support both may read the aria-label value and then the title value. For example I tried with voiceover utility and it does read both.

Should not tho according to W3C:
1. https://www.w3.org/TR/html-aam-1.0/#input-type-text-input-type-password-input-type-number-input-type-search-input-type-tel-input-type-email-input-type-url-and-textarea-element-accessible-name-computation
2. https://www.w3.org/TR/accname-1.2/#computation-steps

It should follow 2Dii resolution path.

packages/atomic/src/components/search/atomic-refine-toggle/atomic-refine-toggle.new.stories.tsx
```diff
@@ -40,3 +40,56 @@ export const Default: Story = {
     `,
   ],
 };
+
+export const WithAtomicExternals: Story = {
+  name: 'atomic-refine-toggle',
```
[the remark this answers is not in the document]
louis-bompart: Yes for the name, no for test: it's a good demo.

packages/atomic/src/components/search/atomic-refine-toggle/atomic-refine-toggle.new.stories.tsx
louis-bompart: They can try, the tests will bite them 😈

packages/atomic/src/components/search/atomic-external/atomic-external.tsx
```diff
@@ -30,19 +31,23 @@ export class AtomicExternal {
...
     );
   }
 
-  private get interface() {
-    const element = document.querySelector(this.selector);
-    if (!element) {
-      throw new Error(
-        `Cannot find interface element with selector "${this.selector}"`
-      );
+  @Prop({mutable: true}) boundInterface?: AtomicInterface;
```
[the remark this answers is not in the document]
louis-bompart: Comment yes, tests... Meh.

PR #4222, opened 2024-07-24
feat(atomic): add tab support for atomic-sort-dropdown

fpbrault:
This PR allows atomic-sort-dropdown to be visible/hidden based on the currently active tab in the tab manager.

https://coveord.atlassian.net/browse/CDX-1554


packages/atomic/src/components/search/atomic-sort-dropdown/atomic-sort-dropdown.tsx
```diff
@@ -102,6 +124,20 @@ export class AtomicSortDropdown implements InitializableComponent {
     option && this.sort.sortBy(option.criteria);
   }
 
+  public componentShouldUpdate(): void {
+    if (
+      !this.options
```
fbeaudoincoveo, line 129: Not a big deal, but I think you could instead use a guard clause to exit ASAP:

```typescript
if (this.options.some((option) => option.expression === this.sortState.sortCriteria)) {
  return;
}
```

fbeaudoincoveo (commented):
LGTM (aside from the failing new e2e tests of course 😄 )

fbeaudoincoveo (approved):

alexprudhomme (approved):

